PR #344, opened 2015-11-11
Support VS 2015 Update 1

SamB:
Some project.json changes needed to build with VS 2015 Update 1 RC.

(Can anyone find documentation for the change in VS/NuGet that forces this?  It seems kind of like it ought to be in the Update 1 release notes...)


rainersigwald:
I have an extremely similar change on my machine I was about to send out :+1: 

Mine is different because I just changed all the `"win-"` to `"win"`.  Did you do differently on purpose?  I think I'd prefer having only the one entry if that still works downlevel.  I'm going to send my PR to see if that works on the CI machine . . .

rainersigwald:
Mine failed, requiring `"win-"` on the CI machine.

LGTM.

SamB:
@rainersigwald: Yeah, I included both `"win-"` and `"win"` because I'd seen some mention (perhaps in a commit to Roslyn?) of needing the former for VS 2015 RTM and the latter for Update 1.

cdmihai:
LGTM. 

I tried to track down the issue in Nugget, to see if there's a better solution to this. From what I can tell the build needs for Nuget.exe and the Nuget msbuild code to both agree on the moniker name (e.g. `win-x64`). Prior to Update1 the Nuget build code ($NuGetRuntimeIdentifier in Microsoft.Nuget.targets) had a bug which resolved it to `win-` instead of `win`, so the project.json files also had to have `win-` in order for the two to agree. 

There are other possible workarounds that work for both Update1 and RTM that involve manually setting the properties that the Nuget build code uses in our `dir.props`, but I think @SamB's solution is cleaner because it does not touch Nuget dependencies in weird ways. :)

rainersigwald:
We also got an internal email promising documentation on this in NuGet (but it doesn't exist yet).

PR #345, opened 2015-11-11
Update NuGet project.json for Update 1

rainersigwald:
Note:  this is mostly to test whether this or @SamB's #344 is the minimal change to be compatible both before and after Update 1.  If this doesn't work we should clearly prefer the other.

These changes were required to avoid errors of the form:

C:\Program Files
(x86)\MSBuild\Microsoft\NuGet\Microsoft.NuGet.targets(109,5): error :
Your project.json doesn't have a runtimes section. You should add
'"runtimes": { "win": { } }' to your project.json and then re-run NuGet
restore. [O:\msbuild\src\XMakeBuildEngine\Microsoft.Build.csproj]

after installing VS2015 Update 1 (and its Windows SDK update).


rainersigwald:
Ok, looks like #344 is the way to go, since this failed complaining that `"win-"` wasn't present.

PR #346, opened 2015-11-11
Avoid ambient node reuse when not requested

### LINKED ISSUE #339: A build with /nodeReuse:false will reuse nodes that were left running by a previous build
Consider this:
```cmd
msbuild /m /nr:true someBigProject.proj
msbuild /m /nr:false someOtherProject.proj
```

Most people would expect the first build to leave child processes running (that being the point of the `/nodeReuse` flag).  The second shouldn't leave any new processes running.

What's surprising is that the second build will reuse processes left over from the first build.  That causes problems if you have per-node state, or if you're looking at I/O from the process cone of MSBuild.

This appears to be intentional but I can't justify the behavior.  I think we should change `/nr:false` to always spawn new processes.

rainersigwald:
Resolves #339 by requiring that any node that will be used as a worker have the same value for `nodeReuse` as the coordinator.

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
+++ b/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
@@ -69,21 +69,21 @@ namespace Microsoft.Build.BackEnd
 
         /// <summary>
         /// Magic number sent by the host to the client during the handshake.
-        /// Derived from the binary timestamp to avoid mixing binary versions.
+        /// Derived from the binary timestamp to avoid mixing binary versions,
+        /// Is64BitProcess to avoid mixing bitness, and enableNodeReuse to
+        /// ensure that a /nr:false build doesn't reuse clients left over from
+        /// a prior /nr:true build.
         /// </summary>
-        internal static long GetHostHandshake()
+        /// <param name="enableNodeReuse">Is reuse of build nodes allowed?</param>
+        internal static long GetHostHandshake(bool enableNodeReuse)
         {
             long baseHandshake = Constants.AssemblyTimestamp;
-            if (Environment.Is64BitProcess)
-            {
-                unchecked
-                {
-                    baseHandshake = baseHandshake ^ (long) 0x8b8b8b8b8b8b8b8b;
-                }
-            }
 
-            baseHandshake = CommunicationsUtilities.GenerateHostHandshakeFromBase(baseHandshake, GetClientHandshake());
-            return baseHandshake;
+            baseHandshake = baseHandshake*17 + Environment.Is64BitProcess.GetHashCode();
+
+            baseHandshake = baseHandshake*17 + enableNodeReuse.GetHashCode();
+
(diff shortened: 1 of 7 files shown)


src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -69,39 +69,33 @@ public int AvailableNodes
...
-            {
-                // Mask out the first byte. That's because old
-                // builds used a single, non zero initial byte,
-                // and we don't want to risk communicating with them
-                return (Constants.AssemblyTimestamp ^ Int64.MaxValue) & 0x00FFFFFFFFFFFFFF;
-            }
+            // Mask out the first byte. That's because old
+            // builds used a single, non zero initial byte,
+            // and we don't want to risk communicating with them
+            return (Constants.AssemblyTimestamp ^ long.MaxValue) & 0x00FFFFFFFFFFFFFF;
```
cdmihai: Int64 might actually be more self-documenting here, since it mentions the mask size.Not worth changing the commit though.

src/XMakeBuildEngine/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -234,15 +234,28 @@ internal static bool IsOutOfProcNode
...
+        }
+
+
+        /// <summary>
+        /// Starts up the node and processes messages until the node is requested to shut down.
+        /// </summary>
+        /// <param name="enableReuse">Whether this node is eligible for reuse later.</param>
+        /// <param name="shutdownException">The exception which caused shutdown, if any.</param>
+        /// <returns>The reason for shutting down.</returns>
+        public NodeEngineShutdownReason Run(bool enableReuse, out Exception shutdownException)
```
cdmihai: Doesn't anyone have to call this method with enableReuse set to true, when node reuse is actually desired? With the current code it looks like nodeReuse is always set to false.

rainersigwald: This gets called at https://github.com/rainersigwald/msbuild/blob/3c015edee424f4eb3535f2235af6ec15f0b57e7f/src/XMakeCommandLine/XMake.cs#L1956.  That'll get set to true when the command line parameter (of the child) is set to true.

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
rainersigwald: Good point.

AndyGerlicher:
LGTM

src/XMakeBuildEngine/BackEnd/Node/OutOfProcNode.cs
cdmihai: Ah, alright, I missed that diff.

cdmihai:
LGTM

PR #347, opened 2015-11-13
Fix Exec tasks & tests for *nix

ValMenn:
* Use source (.) to run shell scripts -- no need for exec permission
* Fix encoding on Linux so that scripts don't have a BOM
* Since exit code on failure is always 1 on Linux added additional checks.
* A test that checks the exit code is fixed to account for the fact that Linux exit codes are byte-wide

### DIFF
--- a/src/XMakeTasks/UnitTests/Exec_Tests.cs
+++ b/src/XMakeTasks/UnitTests/Exec_Tests.cs
@@ -68,8 +68,12 @@ namespace Microsoft.Build.UnitTests
             bool result = exec.Execute();
 
             Assert.Equal(false, result);
-            Assert.Equal(4, exec.ExitCode);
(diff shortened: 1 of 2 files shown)


rainersigwald:
LGTM

src/XMakeTasks/UnitTests/Exec_Tests.cs
```diff
@@ -151,6 +155,7 @@ public void NonUNCWorkingDirectoryUsed()
         }
 
         [Fact]
+        [PlatformSpecific(Xunit.PlatformID.Windows)]   // UNC is Windows-Only
```
cdmihai: Cool, I did not know about the PlatformSpecific attribute

cdmihai:
LGTM

PR #363, opened 2015-11-20
Add MSBuildRuntimeType built-in property

dsplaisted:
Add an `MSBuildRuntimeType` built-in property.  This will be set to `Core` in the .NET Core version of MSBuild, and `Full` in the full Framework version.  This will allow projects to ie load a target for the correct runtime type.

Feedback on the name of the property and what the values should be is welcome.

This PR also deletes the project.lock.json files.  Current guidance from @davidfowl is that these should not be checked in.


rainersigwald:
Hooray getting rid of project.lock.json!

LGTM.  My only slight concern is about the naming of the values, but this seems fine and I bet you know the .NET Core branding guidelines better than I do.

cdmihai:
+1 for getting rid of the lock files. I have a script that deletes them each time before building msbuild on osx :)

Regarding the built-in netcore property, is it supposed to replace the current `NetCoreBuild` property or live along with it? (https://github.com/Microsoft/msbuild/blob/xplat/dir.props#L127). Might get confusing with two of them.

rainersigwald:
@dazinator It wasn't, so I submitted https://github.com/Microsoft/visualstudio-docs/pull/113.

dazinator:
Nice one, thanks.

PR #369, opened 2015-11-24
OSX build

cdmihai:
The bulk of the changes included bumping up versions, adding new dependencies to osx specific nuget packages, and upgrading build code to take osx into consideration.


src/.nuget/packageLoad.targets
```diff
@@ -129,15 +130,18 @@
...
     <PropertyGroup>
       <ProjectDirectory>%(ProjectJsonFile.RootDir)%(ProjectJsonFile.Directory)</ProjectDirectory>
     </PropertyGroup>
 
     <Message Importance="High" Text="Restoring in $(ProjectDirectory)" />
 
-    <!-- Restore packages for this directory -->
-    <Exec Command="$(NugetRestoreCommand) $(NuGetPackageRestoreOptions) -PackagesDirectory &quot;$(HOME)/.nuget/packages&quot; &quot;@(ProjectJsonFile)&quot;" />
+    <!-- Restore packages for this directory using Nuget-->
+    <!-- <Exec Command="$(NugetRestoreCommand) $(NuGetPackageRestoreOptions) -PackagesDirectory &quot;$(HOME)/.nuget/packages&quot; &quot;@(ProjectJsonFile)&quot;" />  -->
```
rainersigwald: Uncomment?

src/.nuget/packages.OSX.config
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rainersigwald: Why XML?

src/XMakeCommandLine/UnitTests/project.json
```diff
@@ -7,28 +7,25 @@
...
   "frameworks": {
     "net451": {
-      "dependencies": {
-        "xunit.runner.visualstudio": "2.1.0"
-      }
+      "dependencies": {}
     },
     "net46": {
-      "dependencies": {
-        "xunit.runner.visualstudio": "2.1.0"
```
rainersigwald: Why remove this dependency?

src/.nuget/packageLoad.targets
cdmihai: This target is only run on non Windows.
Nuget restore does not quite work momentarily, so I am running restore with DNU.
I left the Nuget line commented out because we'll have to re-enable when net core Nuget comes out.

Another option is to split this target in 2 targets (restoreDNU and restoNuget) and then have a master target select between the two. But I thought that might be overkill.

rainersigwald: Ok, leave it commented out but add a note in the previous line about why/when to uncomment it.

src/.nuget/packages.OSX.config
rainersigwald: "because the rest of them are".  Works for me.

src/XMakeCommandLine/UnitTests/project.json
cdmihai: From what I remember, @ValMenn said that MONO MSbuild would not compile because of a missing transitive dependency for xunit.runner.visualstudio under net46 on linux.

Since xunit test discovery within VS does not work on net core, we concluded that the easiest way to get unstuck is to remove it. 

We can work on getting it back up when test discovery works on net core (hopefully, the whole transitive dependency closure will also be available cross-platform by then). I can file an issue to add this back later.

src/.nuget/packages.OSX.config
cdmihai: @ValMenn 

We run Nuget Install on the dependencies under /src/.nuget to get some of the required bootstrapping tools. My current understanding is that Nuget Install works with the xml format.

src/XMakeCommandLine/UnitTests/project.json
rainersigwald: I'm worried about this since I'm planning to turn full-framework builds back on for this branch, and they'll need this.  We can punt it for a while, though.

dsplaisted: The `xunit.runner.visualstudio` package is only used for the VS Test Explorer integration.  You can still run tests during the build without it, either using the console runner (which is pretty much what we do for .NET Core), or the [xunit.runner.msbuild](http://xunit.github.io/docs/running-tests-in-msbuild.html) package, which has an MSBuild task you can use.

rainersigwald: We do tests for full framework with the console runner, too.  But having them in VS is pretty nice . . .

Agree that this isn't blocking.

PR #370, opened 2015-11-24
Improve CI diagnosability

rainersigwald:
I was looking at enabling CI for full-framework builds in xplat and noticed again that we don't reliably fail the script on intermediate step failures.  This has caused some confusion in the past and it's clearly just a bad idea.

### DIFF
--- a/BuildAndCopy.cmd
+++ b/BuildAndCopy.cmd
@@ -25,14 +25,25 @@ echo ** Additional Build Parameters:%AdditionalBuildCommand%
 echo.
 :: Build MSBuild
 call "%~dp0build.cmd" /t:Rebuild %AdditionalBuildCommand%
+set BUILDERRORLEVEL=%ERRORLEVEL%
 
 :: Kill Roslyn, which may have handles open to files we want
 taskkill /F /IM vbcscompiler.exe
 
+if %BUILDERRORLEVEL% NEQ 0 (
+    echo.
+    echo Failed to build with errorlevel %BUILDERRORLEVEL% 1>&2
+    exit /b %BUILDERRORLEVEL%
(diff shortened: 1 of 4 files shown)


BuildAndCopy.cmd
```diff
@@ -25,14 +25,25 @@ echo ** Additional Build Parameters:%AdditionalBuildCommand%
...
 :: Build MSBuild
 call "%~dp0build.cmd" /t:Rebuild %AdditionalBuildCommand%
+set BUILDERRORLEVEL=%ERRORLEVEL%
 
 :: Kill Roslyn, which may have handles open to files we want
 taskkill /F /IM vbcscompiler.exe
 
+if %BUILDERRORLEVEL% NEQ 0 (
+    echo.
+    echo Failed to build with errorlevel %BUILDERRORLEVEL% 1>&2
```
cdmihai: Since there's three runs, would it be useful to also echo the current log file?

rainersigwald: What do you mean "current log file"?  Something like "For more details see %MSBUILDLOGPATH%"?

cdmihai: Yes. It may cut down the time required to figure out which build failed.

rainersigwald: I added the log path to the last message from `build.cmd` (the common script).  That work?

rainersigwald:
Looks like this will continue to fail until the update to `netci.groovy` is pushed.  Does it look ok, @cdmihai?

cdmihai:
Yup, It should work :)
OSX and linux just produce "msbuild.log" for now, but that will get captured by "msbuild*log"

PR #372, opened 2015-11-25
Don't use frameworkAssembly references in NuGet packages for .NET Framework

dsplaisted:
Currently the NuGet packages have '<frameworkAssembly/>' elements in them which causes the in-box reference assembly version of the DLL to be referenced instead of the one in the NuGet package when compiling on .NET Framework.

This wasn't working when compiling with the .NET CLI, so this PR changes the NuGet packages so that the DLLs in the packages will now be used when compiling for .NET Framework.

### DIFF
--- a/src/dirs.proj
+++ b/src/dirs.proj
@@ -1,5 +1,6 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="dir.props" />
+  <Import Project="BuildValues.props" />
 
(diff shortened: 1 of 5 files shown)


cdmihai:
LGTM

BTW, aren't the nuspec files "supposed" to be generated and not versioned, just like the lock files?

rainersigwald:
LGTM

dsplaisted:
@cdmihai No, generally nuspec files aren't generated automatically.  The version number is passed on the command line when creating the package and overrides what's in the .nuspec file.

Issue #360, opened 2015-11-18
MSBuild accidentally enumerates *.sln* when searching for solutions


rainersigwald:
This one's a combination of layers of bad behavior.

* When invoked on the command line with no project argument, we [search](https://github.com/Microsoft/msbuild/blob/07a15cfe9974dc1f558b8b0346a15ab375ee93a4/src/XMakeCommandLine/XMake.cs#L2061) for `.sln` files in the current working directory.
* That search is performed using [`Directory.GetFiles`](https://msdn.microsoft.com/en-us/library/wz42302f(v=vs.110).aspx), which uses Windows APIs to do wildcard expansion on `"*.sln"`.
* Those Windows APIs (as documented in that link) have "interesting" legacy behavior--if *exactly three characters* are specified after a dot at the end of the pattern, that pattern will match *any extension that starts with those three characters* and otherwise matches the pattern.

As a result, @zjrunner encountered a bug when there was a file named `something.slnhash` in the current directory.  MSBuild tried to build it, then failed to parse.

We should be defensive about this and double-check the results of `Directory.GetFiles` to make sure we're really only getting `*.sln` in the literal/UNIX sense.

tverboon:
Created a pull request for this: #373

rainersigwald:
Thanks, @tverboon.  The fix is in master as of b1f05bb.

PR #373, opened 2015-11-25
#360 - MSBuild accidentally enumerates *.sln* when searching for solutions

tverboon:
Extra check if file is really a solution file.
Because the Windows API has legacy behavior, the searchpattern *.sln will be converted to *.sln*.


rainersigwald:
An elegant solution.  Thanks!

LGTM

cdmihai:
LGTM

PR #375, opened 2015-11-27
Update to .NET Core RC1 packages

dsplaisted:
The .NET Core RC1 packages are versioned `-beta-23516` and are available on NuGet.org.


rainersigwald:
IIRC @cdmihai ran into some problems on OS X when doing this (?) update.

cdmihai:
Bumping System.IO.* beyond -beta-23406 brings in a corefx bug (https://github.com/dotnet/corefx/issues/4274) that fails one of our tests.

However, the CI appears to fail before running tests.

dsplaisted:
I had to explicitly add System.Linq.Expressions to runtime.project.json to avoid errors like the following:

    System.Linq.Expressions 4.0.11-beta-23516 provides a compile-time reference assembly for System.Linq.Expressions on DNXCore,Version=v5.0, but there is no run-time assembly compatible with win7-x86

@ericstj You previously helped me with a similar issue with NuGet where indirect references to packages with runtime-specific components didn't work.  Is this tracked as an issue for NuGet somewhere?

rainersigwald:
LGTM

PR #379, opened 2015-12-02
Portable PDBs need to be copied to output

jaredpar:
The test for _DebugSymbolsProduced was not taking into account the new portable PDB setting.


jaredpar:
See the following issue for context:

https://github.com/dotnet/roslyn/issues/6668

rainersigwald:
LGTM

ctaggart:
Thank you for fixing this. My understanding is that the fix will probably ship with Visual Studio 2015 Update 2. Is that correct? I currently have Update 1 installed and I am able to apply the fix manually.

``` ps
PS C:\cs\roslyn> msbuild /version
Microsoft (R) Build Engine version 14.0.24720.0

14.0.24720.0
```

``` ps
PS C:\cs\roslyn> cmd /c where msbuild
C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe
C:\Program Files (x86)\MSBuild\14.0\Bin\amd64\MSBuild.exe
```

Depending on which MSBuild.exe I'm using, I have to apply the fix to either:
```
C:\Program Files (x86)\MSBuild\14.0\Bin\Microsoft.Common.CurrentVersion.targets
C:\Program Files (x86)\MSBuild\14.0\Bin\amd64\Microsoft.Common.CurrentVersion.targets
```

rainersigwald:
@ctaggart That's correct, and those are the right files to edit if you want to apply the fix manually to an earlier version.

PR #380, opened 2015-12-02
Override ToString for TaskItem

### LINKED ISSUE #378: TaskItem.ToString is returning the wrong thing on CoreCLR
I notice that this returns "Microsoft.Build.Utilities.TaskItem" instead of the ItemSpec when called on ITaskItem.  This is due to the type not overriding ToString.  /cc @dsplaisted

rainersigwald:
This appears to have been accidentally conditionalized away for .NET Core
builds in 70aebd1.

Fixes #378.

### DIFF
--- a/src/Utilities/TaskItem.cs
+++ b/src/Utilities/TaskItem.cs
@@ -408,11 +408,7 @@ namespace Microsoft.Build.Utilities
         /// Gets the item-spec.
         /// </summary>
         /// <returns>The item-spec string.</returns>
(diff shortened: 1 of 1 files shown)


rainersigwald:
@ValMenn It looks like the `#if` was added in your change 70aebd1.  Everything seems to work fine without it--can you confirm that it wasn't needed for something else?

ValMenn:
I changed it, because compilation failed at the time. Seems to work fine now. LGTM

PR #383, opened 2015-12-03
Update to NuGet 3.3

rainersigwald:


dsplaisted:
How about we take the opportunity to fix the issue where if you've already got NuGet.exe in your local enlistment it won't update?  Here's a commit that does that: https://github.com/dsplaisted/msbuild/commit/b8ced26b54fc6667a2262db45493ff12c824ccbe

cdmihai:
LGTM. @dsplaisted's PR #384 also includes this change. Should we wait for that PR or go ahead with this change?

I tried restoring with the new Nuget version on OSX but it still freezes midway.

rainersigwald:
I've been trying to pull b8ced26 into this but running into some merge issues because it depends on some infrastructure from xplat that doesn't exist here.  I do think we should do this for both master and xplat, and they're different enough that it needs to be separate.  But I like the idea of fixing the stale-nuget.exe problems, so I'll keep plugging.

cdmihai:
@dotnet-bot test this please

cdmihai:
Yea, the easiest way might be to just redo the changes, and then solve the conflict when we unify master and xplat.

rainersigwald:
Cherry-picking that change turned out to be a real pain, so I'm going to check this in as-is and rely on the improved behavior coming in from xplat when we re-merge it.  Since the update is nice but not required for full-framework builds, I hope that we can avoid any problems.  If any arise, we can revisit this decision.

PR #385, opened 2015-12-03
No universal indent size

SamB:
So, I was looking at the XML schema files today, and I noticed that the indentation wasn't terribly consistent, which may have something to do with the fact that your .editorconfig file has `indent_size = 4` in the `[*]` section, which blindly overrides ALL editor defaults for anything not speciifcally listed later in the file (and `.xsd` files aren't).

So I polished off a PR I made to avoid that in Roslyn, dotnet/roslyn#6943, and then I re-indented your XML schema files to 2-space indents (as VS defaults to), and I manually re-aligned your multi-line comments (which for some reason VS doesn't touch inside when reformatting -- not even to move subsequent lines the same number of columns left or right that the comment-start sequence was moved).

If for some reason you really *did* want them to have 4-space indentation, and the parts with 2-space were in error, I can revise the commits appropriately.

(I also noticed that a number of your MSBuild files -- targets files and/or props files -- are inexplicably also indented in 4-space increments, despite being explicitly assigned `indent_size = 2` further down.  But they have a lot more multiline comments, and weren't even my initial itch, so I decided to leave them for later.)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,13 +4,16 @@ http://EditorConfig.org
 # top-most EditorConfig file
 root = true
 
-# Default settings:
-# Use 4 spaces as indentation
+# Don't use tabs for indentation.
 [*]
 indent_style = space
-indent_size = 4
+# (Please don't specify an indent_size here; that has too many unintended consequences.)
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+# Code files
+[*.cs,*.csx,*.vb,*.vbx]
+indent_size = 4
 
 # Xml project files
 [*.{csproj,vbproj,vcxproj,vcxproj.filters,proj,projitems,shproj}]
@@ -18,4 +21,8 @@ indent_size = 2
 
 # Xml config files
 [*.{props,targets,ruleset,config,nuspec,resx,vsixmanifest,vsct}]
+indent_size = 2
+
+# JSON files
+[*.json]
 indent_size = 2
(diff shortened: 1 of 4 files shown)


cdmihai:
@dotnet-bot test this please

SamB:
@cdmihai: Are you trying to teach the CI system not to attempt Ubuntu/OSX builds on non-xplat commits?  Otherwise, I'm puzzled as to what you'd hope to learn by triggering a retest.

rainersigwald:
I think the root cause of the error here is a combination of:
* We haven't edited these files since they were forked into GitHub, and
* None of us use an editor that respects `.editorconfig` regularly.

Do you/most folks use the VS plugin?

I don't object to the intent of the changes here, but it looks like something in your reworking inserted newlines--look at [the ignore-whitespace diff](https://github.com/SamB/msbuild/commit/92133b4f56fb499e535b3c1d9ce7bd3c7e4b4b52?w=1).  Can you clean that up?

I'm also a bit unsure whether this should go into master or xplat.  A cosmetic change (good though it is) seems misplaced in master, which is currently our "vs2015 servicing" branch.  I'd have absolutely no objection to putting it in our "vnext" branch, though, which is/will be xplat (I hope to merge xplat back into master).  Thoughts either way?

cdmihai:
@dotnet-bot test this please

cdmihai:
Sorry about that. I am testing the CI builds and your PR was a good test case since it was targeting master :)

SamB:
@rainersigwald: Oh, sorry about those; I didn't realize VS would do/had done that.  While I would *hope* that your translation tooling would tolerate such changes without fuzzying everything, the lack of line breaks there was clearly intentional and I can fix that... just as soon as I complete my research into the laziest way to get it done.

rainersigwald:
@SamB I agree that those changes *should* be harmless but I'm ~~paranoid~~experienced enough to not want to test it.

AndyGerlicher:
Thanks for the contribution. After we merge xplat with master we would be open to changes like this. Right now it's just too difficult to manage the changes across branches to take. When we're done with The Great Merge we will want to do additional code formatting.

PR #387, opened 2015-12-03
Update README for xplat

cdmihai:

PR #388, opened 2015-12-03
Update build status badge

rainersigwald:
Looks like this job changed names when we brought the multi-platform builds online for the xplat branch.

@cdmihai


rainersigwald:
Never mind, Let's do #387 instead.

PR #389, opened 2015-12-03
Only archive CI tests for windows

cdmihai:
Linux and OSX do not yet run tests, they just compile.
Moved the line instead of passing skipIfNoTestFiles  = true to
addXunitDotNetResults as a sanity measure to ensure we do not loose the
test phase by mistake and not notice it. :)

@mmitche: would merging this to master cause the Jenkins projects to be silently recreated/updated or will they get duplicated? (if they get duplicated, is there a method we could call in the netci script to clean previous projects?)


rainersigwald:
LGTM

PR #404, opened 2015-12-16
Updates for CI

### LINKED ISSUE #390: Ubuntu and OSX CIs fail at DNU restore
The current bar for Ubuntu and OSX CIs is to pass compilation without running any tests. However, both CIs fail at DNU restore.

On my OSX instance (10.11) DNU restore succeeds. I also tried Nuget 3.3.0, but it still freezes while restoring.

cdmihai:
- set HOME if not set (HOME not set on CI machines)
- use unzip instead of tar (tar failed on linux)
- bump up compilers (the version we used was removed form myget)
- reuse nuget cache variable

Fixes #390

### DIFF
--- a/cibuild.sh
+++ b/cibuild.sh
@@ -8,20 +8,32 @@ usage()
     echo "  --scope <scope>                Scope of the build (Compile / Test)"
 }
 
+setHome(){
+    if [ -z ${HOME+x}]
+    then
+        MONO_COMMAND="( export HOME=$HOME_DEFAULT ; $MONO_COMMAND )"
+        mkdir -p $HOME_DEFAULT
+    fi
+}
+
 downloadMSBuild(){
     if [ ! -e "$MSBUILD_EXE" ]
     then
         mkdir -p $PACKAGES_DIR # Create packages dir if it doesn't exist.
+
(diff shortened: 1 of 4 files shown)


cibuild.sh
```diff
@@ -32,12 +44,16 @@ THIS_SCRIPT_PATH="`dirname \"$0\"`"
 PACKAGES_DIR="$THIS_SCRIPT_PATH/packages"
 MSBUILD_EXE="$PACKAGES_DIR/mono-msbuild/bin/Unix/Debug-MONO/MSBuild.exe"
 MSBUILD_DOWNLOAD_URL="https://github.com/Microsoft/msbuild/releases/download/mono-hosted-msbuild-v0.1/mono-msbuild.zip"
+MSBUILD_ZIP="$PACKAGES_DIR/msbuild.zip"
+HOME_DEFAULT="/tmp/msbuild-CI-home"
```
cdmihai: @mmitche We're setting HOME if it's not set. Does this location look alright to you? Is it persisted from one run to the other? If it's not, it would increase the job time, since nuget will restore without a cache each time.

rainersigwald:
LGTM.

SUPER exciting to see a CI build pass on OS X!

Issue #331, opened 2015-11-04
Convert tests to use PlatformSpecific attributes


rainersigwald:
After adopting `xunit.netcore.extensions`, we have a nice way to specify that tests are platform-specific,  with the `[PlatformSpecific]` attribute.  We should adopt that everywhere and avoid using early-returns like

```C#
            if (NativeMethodsShared.IsUnixLike)
            {
                return;
            }
```

where possible.

PR #412, opened 2015-12-18
Ignore windows specific tests when running on OSX

radical:


rainersigwald:
I'm trying to model this off "the CoreFX way", which has an attribute to model this:

```C#
[PlatformSpecific(PlatformID.Windows)]
```

There's an open issue to do that elsewhere (#331). Can you move to that? That should eliminate the need to add to `dir.targets`.

radical:
Sure, updated.

cdmihai:
LGTM

PR #418, opened 2016-01-06
Make WriteCodeFragment deterministic

### LINKED ISSUE #408: WriteCodeFragment should be deterministic and not include timestamp in comments
The `WriteCodeFragment` task includes a comment in the generated file that has the value `DateTime.Now` of when the file was generated.  This interferes with determinism compiles because it means every build of a project that uses `WriteCodeFragment` is guaranteed to have a file with a different checksum.  

Can we provide a way to disable this timestamp?  Or really can it just be removed by default from the task?  I don't think it provides much value.  It's a generated file that ideally no one is looking at.  If they are looking at it the timestamp is of questionable value because knowing when it was generated is less valuable than knowing the inputs that generated it (at least IMHO).  

Note: in many cases it doesn't directly affect determinism because it often doesn't include executable code, just attributes.  But even in those cases it makes it difficult to track down problems.

jaredpar:
Removes the use of DateTime.Now in the output of WriteCodeFragment.  The
presence of DateTime.Now made the output non-deterministic and prevented
deterministic builds of projects that took advantage of this task.

closes #408


jaredpar:
CC @rainersigwald

rainersigwald:
LGTM, thanks!

PR #419, opened 2016-01-07
Remove invalid check in CommandLineBuilder

cameron314:
When escaping quotes, a check was done to ensure that the number of quotes that will be escaped is even. This doesn't make sense, since there's plenty of command line parsers (including Windows's own `CreateProcess`) that correctly process any number of embedded quotes (as long as they're escaped properly).

The effects of this check can be seen, by the way, by attempting to define a macro in the project properties for a C++ file which has in its body an odd number of quotes (e.g. a string with an embedded quote). Add `FOO="\""` to the C++ Preprocessor Definitions in the property pages and observe the MSB6001 error that results when you attempt to build.

### DIFF
--- a/src/Utilities/CommandLineBuilder.cs
+++ b/src/Utilities/CommandLineBuilder.cs
@@ -302,9 +302,6 @@ namespace Microsoft.Build.Utilities
                 }
                 if (literalQuotes > 0)
                 {
-                    // Command line parsers typically break if you attempt to pass in an odd number of
-                    // escaped double quotes. We can only error if there isn't an even number.
-                    ErrorUtilities.VerifyThrowArgument(((literalQuotes % 2) == 0), "General.StringsCannotContainOddNumberOfDoubleQuotes", unquotedTextToAppend);
                     // Replace any \" sequences with \\"
                     unquotedTextToAppend = unquotedTextToAppend.Replace("\\\"", "\\\\\"");
                     // Now replace any " with \"


AndyGerlicher:
Hi. Thanks for the contribution. We agree with the intention of this fix, but would like to see a test case added that would fail otherwise without this change. Also since we're removing the only usage of StringsCannotContainOddNumberOfDoubleQuotes it should be removed from the strings resource file as well.

cameron314:
Fair enough. I'll rustle up a test and remove that string when I have a moment :-)

jasonwilliams200OK:
:+1: 
Does the CI supports C++ tests? Or can the test be written with arbitrary macro?

cameron314:
Not sure what you're referring to?
The exception can be triggered by building a command line for any program which contains a (single) embedded double-quote. This is easy to produce artificially in any test environment, but can occur also naturally in the current release of Visual Studio by defining a macro in the project properties containing an odd number of quotes.

jasonwilliams200OK:
@cameron314, I was wondering how would the test look like and whether we need a real C macro in the test. Thanks for explaining.

rainersigwald:
As a policy question, we'd rather avoid C++ since it will complicate the cross-platform build.

AndyGerlicher:
Haven't heard from @cameron314 in a while. We would still be willing to accept this with unit tests. Thanks for the contribution!

cameron314:
Yeah, sorry, haven't had the time. Next time I've got a free hour ;-)
(The test itself is trivial, just one line, but setting up the environment, etc. takes a while for people like me who don't hack on .NET very often.)

Issue #426, opened 2016-01-15
MSBuild.exe /? still shows /debug option


rainersigwald:
We removed support for `/debug` in #163, but we didn't remove the help text that mentions it. We shouldn't tease people with a cut feature.

nikson:
@rainersigwald, is it reasonable to remove `enum` value `ParameterlessSwitch.Debugger` from the set?

rainersigwald:
@nikson My first inclination is to leave the enum value (and the string resource) while we have the code `#if`ed out. When we commit to removal we can clean it all up together.

rainersigwald:
Fixed via #430. Thanks, @nikson.

PR #430, opened 2016-01-16
Removed help message for /debug in #426

nikson:
New build is not showing the debug message, it's handled by a condition  in `ShowHelpMessage()`.

I have removed the condition and debug message.

### DIFF
--- a/src/XMakeCommandLine/XMake.cs
+++ b/src/XMakeCommandLine/XMake.cs
@@ -1802,7 +1802,9 @@ namespace Microsoft.Build.CommandLine
                         preprocessWriter = ProcessPreprocessSwitch(commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.Preprocess]);
                     }
 
+#if FEATURE_MSBUILD_DEBUGGER
                     debugger = commandLineSwitches.IsParameterlessSwitchSet(CommandLineSwitches.ParameterlessSwitch.Debugger);
+#endif
                     detailedSummary = commandLineSwitches.IsParameterlessSwitchSet(CommandLineSwitches.ParameterlessSwitch.DetailedSummary);
 
                     // figure out which loggers are going to listen to build events
--- a/src/XMakeCommandLine/CommandLineSwitches.cs
+++ b/src/XMakeCommandLine/CommandLineSwitches.cs
@@ -48,7 +48,9 @@ namespace Microsoft.Build.CommandLine
             OldOM,
 #endif
             DistributedFileLogger,
+#if FEATURE_MSBUILD_DEBUGGER
             Debugger,
+#endif
             DetailedSummary,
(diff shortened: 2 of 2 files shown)


src/XMakeCommandLine/XMake.cs
```diff
@@ -2983,14 +2983,8 @@ private static void ShowHelpMessage()
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_19_IgnoreProjectExtensionsSwitch"));
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_24_NodeReuse"));
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_25_PreprocessSwitch"));
-
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_26_DetailedSummarySwitch"));
-
```
rainersigwald: For consistency with the rest of the disabled feature, bracket this with

```c#
#if FEATURE_MSBUILD_DEBUGGER
...
#endif
```

instead of removing it entirely. We're still trying to gauge the reaction from people about removing the debugger.

src/XMakeCommandLine/Strings.resx
```diff
@@ -565,13 +565,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     </value>
   </data>
-  <data name="HelpMessage_27_DebuggerSwitch" UESanitized="false" Visibility="Public">
-   <value>  /debug
-                     Causes a debugger prompt to appear immediately so that 
-                     Visual Studio can be attached for you to debug the 
-                     MSBuild XML and any tasks and loggers it uses.
-   </value>
-  </data>
+
```
rainersigwald: Leave this (until `FEATURE_MSBUILD_DEBUGGER` is entirely removed).

rainersigwald:
Thanks for the contribution! This will help reduce confusion, but it should be disabled in the same way as the rest of the debugger code.

nikson:
Updated according your suggestion

AndyGerlicher:
Looks good, but can you please squash to a single commit for this? Once that's done we'll merge the change. Thanks for the contribution!

rainersigwald:
Actually, and sorry to have told you wrong earlier: I think you were on the right track with the removing-enum question. It looks like the best place to put the `#if` is around the definition of `ParameterlessSwitch.Debugger`, and then `CommandLineSwitches.IsParameterlessSwitch("debug")` will take care of the display parameters.

Sorry again for the confusion!

nikson:
Failed case is 0 but some `msbuild` error in CI log.

rainersigwald:
@dotnet-bot test this please

rainersigwald:
Definitely confused by the failed test, but it passed on retry. LGTM.

rainersigwald:
Thank you for the contribution!

PR #431, opened 2016-01-19
Sync NuGet package versions.

AndyGerlicher:
I went through all the NuGet packages we referenced and got them in sync. Horrible manual process unfortunately. I also removed the daily feeds so it wouldn't look there since it shouldn't be needed right now.

This now builds on my machine, hopefully the CI will pass.

### DIFF
--- a/Samples/PortableTask/PortableTask.csproj
+++ b/Samples/PortableTask/PortableTask.csproj
@@ -10,6 +10,7 @@
     <AssemblyName>PortableTask</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <UseProductOutputPath>true</UseProductOutputPath>
+	<CopyNuGetImplementations>false</CopyNuGetImplementations>
   </PropertyGroup>
   <PropertyGroup Condition="'$(NetCoreBuild)' == 'true'">
     <StartAction Condition="'$(StartAction)'==''">Program</StartAction>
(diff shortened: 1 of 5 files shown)


Samples/PortableTask/PortableTask.csproj
```diff
@@ -10,6 +10,7 @@
     <AssemblyName>PortableTask</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <UseProductOutputPath>true</UseProductOutputPath>
+	<CopyNuGetImplementations>false</CopyNuGetImplementations>
```
rainersigwald: Looks suspicious.

Samples/TaskUsageLogger/TaskUsageLogger.csproj
```diff
@@ -9,6 +9,7 @@
     <AssemblyName>TaskUsageLogger</AssemblyName>
     <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+	<CopyNuGetImplementations>false</CopyNuGetImplementations>
```
rainersigwald: Ditto

rainersigwald:
LGTM

AndyGerlicher:
This is failing because of registry access which is no longer available on the Windows build. @rainersigwald is working on disabling those tests.

PR #432, opened 2016-01-19
Comment out registry access in our targets

rainersigwald:
Tests that used `Microsoft.Common.targets` were failing after 0437195
because the common targets/props had some registry properties. For now,
I'm just commenting those parts of the files out in this branch, but this
will require a long-term fix so that we can restore the registry lookups
for full-framework-MSBuild-requiring projects.


dsplaisted:
We added an `MSBuildRuntimeType` built-in variable you could use to detect whether you're running on the .NET Core version of MSBuild, ie:

    <PropertyGroup Condition="'$(MSBuildRuntimeType)'=='Core'">

rainersigwald:
Just switched to that, @dsplaisted. Sorry, never hit "go" on the comment I wrote saying I was going to try it.

AndyGerlicher:
LGTM. Finally a pass for Windows!

PR #433, opened 2016-01-19
Do not create satellite assemblies in .NET Core

rainersigwald:
There is not currently an `al.exe` that works on Linux, so satellite
assemblies are not creatable there. Fortunately, they're not needed, so we
can just stop trying to create them when running .NET Core MSBuild.

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -3277,7 +3277,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Target Name="GenerateSatelliteAssemblies"
           Inputs="$(MSBuildAllProjects);@(_SatelliteAssemblyResourceInputs);$(IntermediateOutputPath)$(TargetName)$(TargetExt)"
           Outputs="$(IntermediateOutputPath)%(Culture)\$(TargetName).resources.dll"
-          Condition="'@(_SatelliteAssemblyResourceInputs)' != ''">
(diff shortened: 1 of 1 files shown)


rainersigwald:
@joperezr I didn't do exactly what we talked about, but this should accomplish the same goals. Look reasonable to you?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -3342,7 +3342,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <ItemGroup>
       <IntermediateSatelliteAssembliesWithTargetPath Include="$(IntermediateOutputPath)%(EmbeddedResource.Culture)\$(TargetName).resources.dll"
-                                                     Condition="'%(EmbeddedResource.Culture)' != ''">
+                                                     Condition="'%(EmbeddedResource.Culture)' != '' and '$(MSBuildRuntimeType)' != 'Core'">
```
joperezr: Will this change prevent the Copy to output directory target to try to copy the satellite assemblies?

joperezr: Oh I just looked at the code and looks like this should work.

joperezr:
LGTM

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Yeah, since it should keep the list empty so the existing condition on the `Copy` task should skip it.

PR #442, opened 2016-01-21
Update Exec task to only use UTF8 when required.

### LINKED ISSUE #397: .NET applications invoked through <Exec> tasks may send malformed UTF-8 to their children
This was reported internally in reference to the Windows Driver Kit.

Visual Studio 2015 Update 1 shipped with a fix (#143) to allow tools to run with paths that included non-ANSI Unicode characters.  It did this by changing the Windows codepage within the `.cmd` script emitted by `Exec` to one that supports UTF-8.

This causes problems when a child process spawns another child process and attempts to write to its standard input.  From the excellent internal repro:
```xml
    <Target Name="Repro">
        <Exec Command="repro.exe" />
    </Target>
</Project>
```

Where `repro.exe` is compiled from
```C#
using System;
using System.Diagnostics;
using System.IO;

namespace Repro
{
    class Repro
    {
        {
            using (Process cmdProcess = new Process())
            {
                ProcessStartInfo cmdProcessStartInfo = new ProcessStartInfo("cmd.exe");
                cmdProcessStartInfo.UseShellExecute = false;
                cmdProcessStartInfo.RedirectStandardInput = true;
                cmdProcess.StartInfo = cmdProcessStartInfo;
                cmdProcess.Start();

                StreamWriter inputWriter = cmdProcess.StandardInput;
                inputWriter.AutoFlush = true;
                inputWriter.WriteLine("set CL=");
                inputWriter.WriteLine("exit");
                cmdProcess.WaitForExit();

                Console.WriteLine("Exit code: " + cmdProcess.ExitCode);
            }
            Console.Write("Repro Exiting");
        }
    }
}
```

Produces this output:
```
Microsoft (R) Build Engine version 14.0.24720.0

Build started 12/7/2015 5:00:43 PM.
Project "d:\work\sdv_encoding\project.vcxproj" on node 1 (default targets).
Repro:
  repro.exe
  Microsoft Windows [Version 10.0.10240]
  (c) 2015 Microsoft Corporation. All rights reserved.

  d:\work\sdv_encoding>���set CL=
  '���set' is not recognized as an internal or external command,
  operable program or batch file.

  d:\work\sdv_encoding>exit
  Exit code: 9009
  Repro Exiting
Done Building Project "d:\work\sdv_encoding\project.vcxproj" (default targets).
```

Note the super-sketchy "unknown Unicode character" glyphs that `cmd.exe` is complaining about not understanding when prepended to `set`.

AndyGerlicher:
Proposed change for #397. This should revert the change to the existing behavior (will not change the code page) in almost all cases, but still allow for those who do have non-ANSI characters in their Command. This solution seems a little unfortunate, and we are open to additional input if anyone has a better idea. I tried to implement it with an escape hatch (setting Never or Always on the task) so the user has more control if absolutely necessary.

Exec task was previously changed to use UTF8 code page because of an issue
running commands that contained non-ANSI characters. This turned out to be
problematic with some older tools that did not correctly support the cmd
code page change. With this change, we will only change the code page when
the command the Exec task is running cannot be represented by the current
code page. In that case, we will switch to UTF8.

Added a parameter to allow the user to override this behavior and Always
or Never use UTF8 encoding. The default value if not specified is Detect
which will use UTF8 only when it would otherwise fail.

Closes #397

### DIFF
--- a/src/XMakeTasks/Exec.cs
+++ b/src/XMakeTasks/Exec.cs
@@ -33,16 +33,20 @@ namespace Microsoft.Build.Tasks
             // Console-based output uses the current system OEM code page by default. Note that we should not use Console.OutputEncoding
             // here since processes we run don't really have much to do with our console window (and also Console.OutputEncoding
             // doesn't return the OEM code page if the running application that hosts MSBuild is not a console application).
-            // Note: 8/12/15 - Changed encoding to use UTF8 when OS version is greater than or equal to 6.1 (Windows 7)
-            _standardOutputEncoding = GetEncodingWithOsFallback();
-            _standardErrorEncoding = GetEncodingWithOsFallback();
+            // If the cmd file contains non-ANSI characters encoding may change.
+            _standardOutputEncoding = EncodingUtilities.CurrentSystemOemEncoding;
+            _standardErrorEncoding = EncodingUtilities.CurrentSystemOemEncoding;
         }
 
         #endregion
 
         #region Fields
 
-        // Are the ecodings for StdErr and StdOut streams valid
+        private const string UseUtf8Always = "ALWAYS";
+        private const string UseUtf8Never = "NEVER";
+        private const string UseUtf8Detect = "DETECT";
+
+        // Are the encodings for StdErr and StdOut streams valid
         private bool _encodingParametersValid = true;
         private string _workingDirectory;
         private ITaskItem[] _outputs;
@@ -70,7 +74,7 @@ namespace Microsoft.Build.Tasks
         /// Enable the pipe of the standard out to an item (StandardOutput).
         /// </summary>
         /// <Remarks>
-        /// Even thought this is called a pipe, it is infact a Tee.  Use StandardOutputImportance to adjust the visibility of the stdout.
+        /// Even thought this is called a pipe, it is in fact a Tee.  Use StandardOutputImportance to adjust the visibility of the stdout.
         /// </Remarks>
         public bool ConsoleToMSBuild { get; set; }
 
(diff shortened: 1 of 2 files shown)


AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:
@jaredpar look ok to you? Any concerns?

src/XMakeTasks/Exec.cs
```diff
@@ -33,15 +33,19 @@ public Exec()
...
+            // If the cmd file contains non-ANSI characters encoding may change.
+            _standardOutputEncoding = EncodingUtilities.CurrentSystemOemEncoding;
+            _standardErrorEncoding = EncodingUtilities.CurrentSystemOemEncoding;
         }
 
         #endregion
 
         #region Fields
 
+        private const string UseUft8Always = "always";
```
rainersigwald: Typo: uft

src/XMakeTasks/Exec.cs
```diff
@@ -561,19 +580,72 @@ protected override MessageImportance StandardOutputLoggingImportance
...
 
-            return Environment.OSVersion.Version >= windows7
-                ? s_utf8WithoutBom
-                : EncodingUtilities.CurrentSystemOemEncoding;
+            if (Environment.OSVersion.Version < windows7)
+            {
+                useUtf8 = UseUtf8Never;
+            }
+
+            switch (useUtf8.ToLowerInvariant())
```
rainersigwald: Probably doesn't matter for this case since we're only looking at 3 strings, but isn't the usual advice to uppercase things?

src/XMakeTasks/Exec.cs
```diff
@@ -124,6 +128,14 @@ protected override Encoding StandardErrorEncoding
         }
 
         /// <summary>
+        /// Whether or not to use UTF encoding for the cmd file and console window.
+        /// Values: Always, Never, Detect
+        /// If set to Detect, UTF8 will only be used when the Command / WorkingDirectory
+        /// cannot be encoded with the current code page.
```
rainersigwald: Something about this reads awkward to me. Maybe "If set to Detect, the current code page will be used unless it cannot represent the string. In that case, UTF-8 is used."?

rainersigwald:
LGTM

jaredpar:
LGTM (sorry been out for a few days). 

Is the MSBuild snippet to force Exec to not use UTF8 the following? 

``` xml
<Exec UseUtf8Output="never"> 
```

AndyGerlicher:
Thanks for taking a look. And yes, that would never use UTF8 even when it would fail. Or similarly `<Exec UseUtf8Output="always">`.

Issue #368, opened 2015-11-24
Item Functions are not parsed correctly in conditions


rainersigwald:
This came up on an internal mailing list.  A user was trying to emit a message if an item didn't contain a certain element, and I recommended using the [Item Function](https://msdn.microsoft.com/en-us/library/ee886422.aspx) `AnyHaveMetadataValue`.  But that produces an error:

```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <MainEmbeddedFiles Include="foo.baml" />
    <MainEmbeddedFiles Include="UserControl1.baml" />
  </ItemGroup>

  <Target Name="b">
    <Error Text="Missing expected BAML file!!!"
           Condition="!@(MainEmbeddedFiles->AnyHaveMetadataValue('Filename', 'UserControl1'))" />
  </Target>
</Project>
```

```console
d:\play>msbuild test.proj /v:q /nologo
d:\play\test.proj(9,12): error MSB4092: An unexpected token ")" was found at character position 71 in condition "!@(MainEmbeddedFiles->AnyHaveMetadataValue('Filename', 'UserControl1'))".
```

This can be worked around by creating a property:
```xml
  <Target Name="b">
    <PropertyGroup>
      <HasFile>@(MainEmbeddedFiles->AnyHaveMetadataValue('Filename', 'UserControl1'))</HasFile>
    </PropertyGroup>
    <Error Text="Missing expected BAML file!!! @(MainEmbeddedFiles->AnyHaveMetadataValue('Filename', 'UserControl1'))"
           Condition="!$(HasFile)" />
  </Target>
```

I see no reason that we should behave differently in a `Condition`.

nikson:
Is it still open for grab?

rainersigwald:
@nikson Yes, this is still up for grabs. However, it's in a part of the code that hasn't seen changes in a while, so we'd like to be cautious. If you'd like to take a look, can you post a quick summary of the changes you think you'll make before going all the way through the fix? (Of course, if it's easier to just send a PR with a proposed fix, that's fine, but I'd like to potentially save some of your time by considering an approach first.)

PR #447, opened 2016-01-25
#368 fixed Scanner to consume correct close paren.

heXelium:
I suppose there is a bug in a method ```ParseInternalItemList```. It just reads token to the first closing paren, which is causing creation of incorrect token ```@(MainEmbeddedFiles->AnyHaveMetadataValue('Filename', 'UserControl1')``` instead ```@(MainEmbeddedFiles->AnyHaveMetadataValue('Filename', 'UserControl1'))```. 

I made a fix and add some unit tests. @rainersigwald can you take a look on my solution?


rainersigwald:
Not quite as scary a change as I feared. Thank you!

LGTM

AndyGerlicher:
LGTM. Thanks!

PR #448, opened 2016-01-25
Audit uses of IsUnix

rainersigwald:


src/Shared/UnitTests/FileUtilities_Tests.cs
```diff
@@ -840,5 +841,24 @@ public void GenerateTempBatchFileWithBadDirectory()
             }
            );
         }
+
+        [Fact]
+        [PlatformSpecific(PlatformID.AnyUnix)]
+        public void AbsolutePathLooksLikeUnixPathOnUnix()
+        {
+            var absolutePathToThisAssembly = Path.GetFullPath(typeof (FileUtilities_Tests).GetTypeInfo().Assembly.Location);
```
rainersigwald: I felt a little dirty about this, but couldn't think of a better way to get an absolute path fragment that has elements that are guaranteed to exist. Open to better ideas . . .

AndyGerlicher:
LGTM

PR #451, opened 2016-01-27
Fix some issues with OSX build

ValMenn:
1. Remove some v2 sources from NuGet.Config
2. Update dnx package to rc2-16357 (the old package is no longer available)
3. Use 'ulimit -n 2500' to avoid 'Too many open files' error from dnx
4. Make sure mono compilers that come with the mono MSBuild bootstrap package have executable permission.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -10,11 +10,7 @@
     <add key="myget.org buildtools v3" value="https://www.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <add key="myget.org nugetbuild" value="https://www.myget.org/F/nugetbuild/api/v3/index.json" />
     <add key="myget.org dotnet-core v3" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
-    <add key="myget.org buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
-    <add key="myget.org dotnet-corefx" value="https://www.myget.org/F/dotnet-corefx/" />
-    <add key="myget.org dotnet-corefxtestdata" value="https://www.myget.org/F/dotnet-corefxtestdata/" />
-    <add key="myget.org dotnet-core" value="https://www.myget.org/F/dotnet-core/" />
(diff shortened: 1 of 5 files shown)


ValMenn:
@dotnet-bot test this please

NuGet.Config
```diff
@@ -10,11 +10,7 @@
...
     <add key="myget.org nugetbuild" value="https://www.myget.org/F/nugetbuild/api/v3/index.json" />
     <add key="myget.org dotnet-core v3" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
-    <add key="myget.org buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
-    <add key="myget.org dotnet-corefx" value="https://www.myget.org/F/dotnet-corefx/" />
-    <add key="myget.org dotnet-corefxtestdata" value="https://www.myget.org/F/dotnet-corefxtestdata/" />
-    <add key="myget.org dotnet-core" value="https://www.myget.org/F/dotnet-core/" />
-    <add key="myget.org aspnet vnext" value="https://www.myget.org/F/aspnetvnext/api/v3/index.json" />
-    <add key="nuget.org v2" value="https://nuget.org/api/v2/" />
+    <add key="myget.org aspnet vnext v3" value="https://www.myget.org/F/aspnetvnext/api/v3/index.json" />
+    <add key="myget.org aspnet vnext" value="https://www.myget.org/F/aspnetvnext/" />
```
rainersigwald: Should this include `/api/v3/index.json`?

rainersigwald:
LGTM

NuGet.Config
rainersigwald: Oh, I see--my question is wrong. It's actually "why leave the v2 reference here when it's not for any of the other sources?"

ValMenn: I had to keep both v2 and v3 sources for aspnetvnex, else dnx would not consistently download the latest version from v3 (it could be temporary, but this source is only used to download dnx, so it does not have any effect on the other packages).

rainersigwald: Fair enough.

PR #453, opened 2016-01-28
Fix build on Mac OS X

mfilippov:
I fix build on Mac OS X. I update DNX version and remove invalid feed from NuGet.config.


ValMenn:
@mfilippov Looks like I made similar changes a few hours before you did. Since I made some more changes, it may make sense for you to review my PR instead. Thanks.

mfilippov:
@ValMenn Yes you are right. When I made PR I have not seen your PR.

Issue #438, opened 2016-01-21
CI build script doesn't look for the right msbuild.log


rainersigwald:
Every CI build includes this (benign) string:
```
FINDSTR: Cannot open d:\j\workspace\Microsoft_msbuild\_Windows_NT_prtest\msbuild.log
```

That's misleading and we should fix it. I probably didn't update the search-the-log part of the script when I switched to disambiguated log files.

PR #454, opened 2016-01-29
Make the static state of ToolLocationHelper thread safe

cdmihai:
Some fields were not protected by the lock. Added the lock.

This branch was made out of the update 2 branch, therefore this PR is also a merge of update2 into master

### DIFF
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -853,30 +853,33 @@ namespace Microsoft.Build.Utilities
                 string registryRoot
             )
         {
-            if (s_cachedTargetPlatformReferences == null)
+            lock (s_locker)
             {
-                s_cachedTargetPlatformReferences = new Dictionary<string, string[]>(StringComparer.OrdinalIgnoreCase);
-            }
+                if (s_cachedTargetPlatformReferences == null)
+                {
+                    s_cachedTargetPlatformReferences = new Dictionary<string, string[]>(StringComparer.OrdinalIgnoreCase);
+                }
 
-            string cacheKey = String.Join("|", sdkIdentifier, sdkVersion, targetPlatformIdentifier, targetPlatformMinVersion, targetPlatformVersion, diskRoots, registryRoot);
+                string cacheKey = String.Join("|", sdkIdentifier, sdkVersion, targetPlatformIdentifier, targetPlatformMinVersion, targetPlatformVersion, diskRoots, registryRoot);
 
-            string[] targetPlatformReferences = null;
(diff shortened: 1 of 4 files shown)


rainersigwald:
Can you reference #438 in the commit that you fixed it with? (I'm fighting to never be too lazy to _file_ the bug, even if I'm too lazy to fix it)

rainersigwald:
Made a lightweight audit of the static caches and they all appear to be under lock now :+1: 

Do you think it's worth filing an issue to change them all to ConcurrentDictionaries like we discussed?

LGTM

cdmihai:
Added the #438 to the commit message.

Regarding the concurrent collections, do you have a sense of how often is the tool location helper called? Also, from what I saw, all those caches contain a very well defined and finite set of possible values. If the contention is not high (e.g. one thread calls into it once per build) then I do not think it's worth it.

rainersigwald:
I doubt contention is particularly significant and I'm willing for profiling to show otherwise. I suspect it's on the order of ~10 calls per project, and these lookups should go pretty quickly after the initial population, so I'm fine with doing nothing else now.

src/XMakeTasks/UnitTests/GetReferencePaths_Tests.cs
```diff
@@ -157,16 +157,51 @@ public void TestGeneralFrameworkMonikerNonExistent()
             MockEngine engine = new MockEngine();
             GetReferenceAssemblyPaths getReferencePaths = new GetReferenceAssemblyPaths();
             getReferencePaths.BuildEngine = engine;
-            // Make a framework which does not exist, intentional mispelling of framework
+            // Make a framework which does not exist, intentional misspelling of framework
             getReferencePaths.TargetFrameworkMoniker = ".NetFramewok, Version=v99.0";
-            getReferencePaths.Execute();
+            bool success = getReferencePaths.Execute();
```
kivancmuslu: This is a test comment, please ignore.

src/XMakeTasks/UnitTests/GetReferencePaths_Tests.cs
```diff
@@ -157,16 +157,51 @@ public void TestGeneralFrameworkMonikerNonExistent()
             MockEngine engine = new MockEngine();
```
kivancmuslu: Another test comment, please ignore.

src/Utilities/ToolLocationHelper.cs
```diff
@@ -853,30 +853,33 @@ public static string[] GetTargetPlatformReferences
                 string registryRoot
             )
         {
-            if (s_cachedTargetPlatformReferences == null)
+            lock (s_locker)
```
kivancmuslu: Just a test comment, please ignore.

AndyGerlicher:
LGTM

PR #458, opened 2016-01-30
Option for using ANSI-Colors that works with redirection

assafnativ:
Eliminat some code duplication, SetColor and ResetColor were already
defined in BaseConsoleLogger.cs


AndyGerlicher:
Hi. First I do want to say thanks @assafnativ for getting involved and making a contribution! However in this case I'm going to have to ask you to make an Issue here and discuss the motivations behind this change. As it's not something that was on our road map I think the discussion and formal motivation is important (i.e. what environment/OS is this change needed for?). See the [Contributing Code](https://github.com/Microsoft/msbuild/wiki/Contributing%20Code)) wiki entry for some additional information. 

We discussed this in our Triage yesterday and in the current state we like the removal of code duplication, but without proper motivation for the ANSI color codes change we would not accept the PR. Again, open to discussion in an Issue.

PR #462, opened 2016-02-03
Revert "Copy .mdb files when they are generated by the compiler instead"

### LINKED ISSUE #459: .Net Core MSBuild shouldn't try to copy .mdb files when DebugType = Portable in Linux
When using .Net Core MSBuild to build corefx repo, I'm hitting this issue when I set DebugType = Portable. The reason is that with portable pdb generation enabled, a _.pdb file will be created, instead of a *.mdb file(_.mdb files are only generated when using Mono MSBuild). The failure I'm getting is that according to Microsoft.Common.CurrentVersion.targets located on the Microsoft.Build.Targets nuget package, whenever DebugType is set to portable in a non-Windows environment, it will try to copy the PDB file as well as the MDB file which doesn't exist.
Instead of doing this, MSBuild should notice if it is running Mono MSBuild or .Net Core MSBuild, and depending on which one it is, then choose to copy the PDB or the MDB but not both.

Code in Microsoft.Common.CurrentVersion.targets looks like this:

``` xml
<_DebugSymbolsIntermediatePathMDB Include="$(IntermediateOutputPath)$(TargetName)$(TargetExt).mdb" Condition="'$(OS)' != 'Windows_NT' and '$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
<_DebugSymbolsIntermediatePath Include="@(_DebugSymbolsIntermediatePathPDB);@(_DebugSymbolsIntermediatePathMDB)" Condition="'$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
```

@rainersigwald provided a workaround that can be applied into corefx directly for now, but it would be nice to have the right fix in MSBuild instead.

rainersigwald:
This reverts commit 658fc55ea8af591299d7d59bc7a4fd3908b39f6d.

That change detected "is building using the Mono compiler" through the
condition "is not on Windows_NT", which isn't accurate since the Roslyn
compiler works fine on Linux and OS X now. That caused errors to find
`.mdb` files when building with Roslyn and creating debug symbols on
Unices--so this fixes #459.

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -348,9 +348,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <ItemGroup Condition="'$(_DebugSymbolsProduced)' == 'true'">
     <_DebugSymbolsIntermediatePath Include="$(IntermediateOutputPath)$(TargetName).compile.pdb" Condition="'$(OutputType)' == 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
-    <_DebugSymbolsIntermediatePathPDB Include="$(IntermediateOutputPath)$(TargetName).pdb" Condition="'$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
-    <_DebugSymbolsIntermediatePathMDB Include="$(IntermediateOutputPath)$(TargetName)$(TargetExt).mdb" Condition="'$(OS)' != 'Windows_NT' and '$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
-    <_DebugSymbolsIntermediatePath Include="@(_DebugSymbolsIntermediatePathPDB);@(_DebugSymbolsIntermediatePathMDB)" Condition="'$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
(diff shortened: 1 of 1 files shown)


rainersigwald:
@joperezr I confirmed that overwriting the local copy of `Microsoft.Common.CurrentVersion.targets` post-`init-tools.sh` with this copy gets a successful build on your repro branch of corefx.

rainersigwald:
@akoeplinger Do you think this will cause problems with actual Mono builds? We talked it over internally and the assumption was that those builds would use Mono targets that should already handle `.mdb` creation and placement, but we weren't sure.

joperezr:
@rainersigwald awesome, thanks for the help. Please let me know when you have a new package ready with this fix.

AndyGerlicher:
LGTM

akoeplinger:
@rainersigwald can we somehow detect if portable pdbs are used on non-Windows? Shipping Roslyn with Mono is planned and I suppose we'll have a period where we expect both .pdb from Roslyn and .mdb from mcs to be produced. We haven't integrated MSBuild yet though, so it doesn't really impact us right now.

akoeplinger:
@radical do you think this has an impact?

rainersigwald:
@akoeplinger There's a [setting](https://github.com/Microsoft/msbuild/blob/639470bfa680e246af02ef11240603b7c1eca657/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L149) that should allow us to determine whether a Portable PDB is in use, `'$(DebugType)'=='portable'`. But I don't love the idea of detecting "mono debug symbol outputs" with "isn't Windows and isn't portable"--I'd prefer something along the lines of an explicit "is using mcs".

Since this is keeping our friends in CoreFX from building debug symbols, I'd like to move forward with this change while we figure out the best medium/long-term fix in #463.

rainersigwald:
@joperezr This should be in our packages version 00017 and up.

joperezr:
@rainersigwald Thanks! trying it out now.

PR #314, opened 2015-10-23
Generate resources cache fix

AndyGerlicher:
This fixes the scenario of switching builds using RTM and Update 1. Normal usage this should only be for an uninstall scenario since MSBuild is typically in the GAC, but in cases of a checked in toolset or using the open source version we should be backwards compatible. This is just renaming the fields back for the cache classes.

PR #466, opened 2016-02-04
ResGen Cache Serialization Fix

AndyGerlicher:
### Update ResGen Serialization Compat for RTM and U1
_602201c_ When reverting the CodeFormatter private field name changes (#314), the Dependencies class was missed. This causes an incompatibility going from  Update 1 back to RTM and causes a null ref. This change fixes the compat issue (rename the field) and updates the serialization version to ensure U1 -> U2 works as expected.

### Rename field for internal build.
_c863538_ Looks like the field was renamed for STANDALONEBUILD but not otherwise.Correcting this.


rainersigwald:
LGTM, except you misspelled "serialization" in the commit message.

PR #468, opened 2016-02-05
Show full class and method name in output results

jdom:
This is so that the Jenkins plugin that gathers xUnit test results shows the correct test names, because the converter trims the first portion of the test name, assuming that it is prefixed with the full class name

### DIFF
--- a/src/xunit.runner.json
+++ b/src/xunit.runner.json
@@ -2,5 +2,5 @@
   "shadowCopy": false,
   "maxParallelThreads": 1,
   "parallelizeAssembly": false,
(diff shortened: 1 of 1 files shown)


rainersigwald:
Oh, is THAT what was wrong? Awesome! Thanks.

LGTM

jdom:
yeah, we had the same issue in Orleans, and while trying to search for other usages I stumbled upon your repo too. We fixed it here: https://github.com/dotnet/orleans/pull/1382

AndyGerlicher:
LGTM

Issue #460, opened 2016-02-02
Memory leaks when unloading Projects


rainersigwald:
There is an ancient thread about this at http://social.msdn.microsoft.com/Forums/vstudio/en-US/36381eaf-4118-4f10-9101-69021619d3ba/memory-leak-in-solutionloadstandaloneproject. As far as I know this hasn't been fixed.

According to another internal report from long ago,

> You can confirm this by creating a simple project that just loads and disposes projects; it will eat up all of your memory.

This seems likely to be related to memory-usage complaints from another internal team that has a similar load-a-bunch-of-projects workflow.

Sarabeth-Jaffe-Microsoft:
@heXelium did some good analysis on this issue in #474, which presents a couple of options.

OniBait:
I'd love to make a PR to fix this issue but seeing as how #474 was closed because there was no clear direction on the fix I'm a bit hesitant to. There's several ways I can think of to fix it:
1) Use `String.Intern()` instead of `ProjectStringCache` -- the projects aren't getting unloaded properly anyway, so the reference counting going on doesn't happen.
2) Don't try and pool the strings at all. They should already be getting pooled in an `XmlNameTable` already, so pooling them again seems like a waste.
3) Override the equality/hashcode members on `XmlDocumentWithLocation` so that the reference cache works properly.
4) Use a different key for `ProjectStringCache._documents (the document location, the hashcode of the object, something...)
5) Use something like a `WeakReference<>` as the key and a custom key comparer for the dictionary so that the XmlDocument isn't prevented from being GC'd.

This bug in combination with the `GC.Collect` inside the loop in `BuildRequestEngine.CheckMemoryUsage` can lead to really horrible performance. A key example would be switching a branch in git and clicking "Reload All Projects" in Visual Studio -- the memory pressure goes up due to the projects not getting unloaded and then you start getting huge > 90 second pauses in Visual Studio because it is trying to make room for more memory (glad to see that https://github.com/Microsoft/msbuild/pull/1351 made it into the VS15-rtw branch)

jeromelaban:
I've hit this issue when dealing with Roslyn's `MSBuildWorkspace` (repeatedly creating workspaces in the same process) and here's how I'm currently dealing with it:

```csharp
if (typeof(Microsoft.Build.Construction.ElementLocation).Assembly.GetType("Microsoft.Build.Construction.ProjectStringCache") is { } stringCacheType)
{
[...]

PR #474, opened 2016-02-09
#460 - Memory leak fix.

heXelium:
I investigated issue described in the provided thread. It's easy to reproduce it by creating ```Project``` in a following way (this code is copied from Roslyn and than modified to show the point):
```csharp
public async Task Foo()
{
     string path = @"...some project link";
     var properties = new Dictionary<string, string>();

     for (int i = 0; i < 100; i++)
     {
          var xmlReader = XmlReader.Create(await ReadFileAsync(path, CancellationToken.None).ConfigureAwait(false), s_xmlSettings);
          var xml = ProjectRootElement.Create(xmlReader, collection);
          xml.FullPath = path;

          var project = new Project(
                    xml,
                    properties,
                    toolsVersion: null,
                    projectCollection: collection);
     }
}
```
Whole memory will be consumed by cached items in ```ProjectStringCache```. There no mechanism of freeing cache data automatically if the references to instances of ```ProjectCollection``` and ```Project``` classes will be lost. The only way to handle this is to call ```ProjectCollection.UnloadProject()``` or ```ProjectCollection.UnloadAllProjects()``` manually. 

It seems that in Roslyn they ended up with some kind of a workaround by wrapping ```Project``` in ```ProjectFile``` class with destructor:
```csharp
~ProjectFile()
{
     try
     {
          // unload project so collection will release global strings
          _loadedProject.ProjectCollection.UnloadAllProjects();
     }
     catch
     {
     }
}
```

So solution that I propose is pretty much simple. I just implemented Dispose pattern for ```ProjectCollection``` to clear all this stuff on GC.

### DIFF
--- a/src/XMakeBuildEngine/Definition/ProjectCollection.cs
+++ b/src/XMakeBuildEngine/Definition/ProjectCollection.cs
@@ -308,6 +308,14 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
+        /// Finalizer. During GC all items that were saved in strong cache during project loading will be removed.
+        /// </summary>
+        ~ProjectCollection()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
         /// Handler to receive which project got added to the project collection.
         /// </summary>
         [SuppressMessage("Microsoft.Design", "CA1034:NestedTypesShouldNotBeVisible", Justification = "This has been API reviewed")]
@@ -1489,6 +1497,9 @@ namespace Microsoft.Build.Evaluation
                 ShutDownLoggingService();
                 Tracing.Dump();
             }
+
+            //releasing cached items from strong reference cache for projects loaded into current collection
+            ProjectRootElementCache.Clear();
         }
 
         /// <summary>


AndyGerlicher:
This looks great, and thank you for putting in the effort to try to fix it.

However, it also looks very scary! Particularly the [comment](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs#L55-L57):
```
/// Because of the strongly typed list, some ProjectRootElement's will be held onto indefinitely. This is an acceptable price to pay for
/// being able to provide a commonly used ProjectRootElement immediately it's needed. It is mitigated by the list being finite and small, and
/// because we allow ProjectCollection.UnloadAllProjects to hint to us to clear the list.
```

Because we don't know the implications, this is definitely out of scope for Update 2 (or probably 3) unless we have a really solid understanding of 1) why is this list not `finite and small` as suggested and 2) what implications calling `ProjectCollection.UnloadAllProjects` will have when in a long running process (like Visual Studio IDE) that may still need project references even after the finalizer may run. This change also seems to undo a lot of the logic defined in the ProjectRootElementCache, and it looks like a lot of thought went into it. Is there an issue with the weak/strong reference implementation? Would it be better simply without the cache or that particular implementation? Without knowing the answer to those questions, we can't take this change for VS2015. We would like to have this issue addressed in Dev15 though, so I will leave this open until we can do some further analysis.

heXelium:
@AndyGerlicher thank you for quick and clear answer. Despite my fix currently affects ```ProjectRootElementCache``` the real issue is in [```ProjectStringCache```](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Evaluation/ProjectStringCache.cs). So I tried to find a safer solution. And after doing more testing I see some option how to resolve this.
```ProjectStringCache``` purpose is to intern strings for loaded ```XmlDocument```s to use less memory. But in fact it uses more than need. See below screenshots from memory profiler. I executed example code provided above to load same project file 1 and 100 times in a loop.
image
image
On second screenshot you can see that strings is really interned but it also keeps references to every instance of created ```XmlDocument``` (even loaded xml doc is same). The reason of this because ```XmlDocument.GetHashCode()``` implementation return hashcode of object instance, it's not overridden.
When I removed string internation for testing purposes, obviously memory leak was resolved. And also it works faster while loading and unloading projects simultaneously.
So the issue can be resolved quickly by removing internation and as second option override ```XmlDocumentWithLocation.GetHashCode()``` to compare xml entirely. Not sure what is better option here from perf perspective.
What do you think about it? Can we go with one of them? Or you have other thoughts.

Sarabeth-Jaffe-Microsoft:
@heXelium Thanks for looking in to this! For now, we're going to close this pull request since we don't have a clear direction on this fix.

OniBait:
I looked into this and ran into a similar situation where Visual Studio ends up taking more and more memory due to this.

Taking memory snapshots I'm seeing where 500MB of memory was being used purely by the `ProjectStringCache._documents` dictionary.

The problem is indeed that the projects aren't getting unloaded and ejected from this cache properly and the XmlDocument (used as the key) still has a reference to it.

So the end result is that every build causes a memory leak -- with build imports like `Microsoft.Common.CurrentVersion.targets` utilizing 247k, that adds up pretty quickly (and completely destroys the benefit of interning the strings).

I think a better solution might be to use `XmlDocumentWithLocation.FullPath` as the key to the dictionary (and not interning strings where that is null to account for when the project hasn't been loaded from disk) or to use something like `ConditionalWeakTable<>` to fix this huge memory leak.

Issue #439, opened 2016-01-21
Remove registry package references for .NET Core


rainersigwald:
I recently (#427) turned off the use-registry flag entirely for our .NET Core flavor. But I didn't remove the `Microsoft.Win32.Registry` package references, and the (generated) nuspecs still declare a dependency on it.

rainersigwald:
Looks like this was resolved between #475 and 18b79b28eaba635d7621bf5b8ddca68706e341bc.

PR #475, opened 2016-02-09
#439 - Removed usage "Microsoft.Win32.Registry" from project.json.

heXelium:
As advertised just removed usage of "Microsoft.Win32.Registry" from ...\msbuild\src\Utilities\project.json and ...\msbuild\src\XMakeTasks\project.json (#439).


rainersigwald:
Seems like a good step, but `Microsoft.Win32.Registry` is used in several other `project.json`s. Is there a reason this is a subset?

heXelium:
Heh, you are right. I did solution wide search but it did not return all results on first search for some reason. Removing it from other ```project.json```'s.

rainersigwald:
Thanks! LGTM

AndyGerlicher:
LGTM as well. Can you please squash into a single commit though?

heXelium:
Sure. Done.

PR #478, opened 2016-02-11
Add support for resolving $(MSBuildExtensionsPath) with fallback sear…

radical:
…ch paths.

In Mono's xbuild, $(MSBuildExtensionsPath) is $mono_prefix/lib/mono/xbuild.
On OS X, this is:

    `/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild`

Consequently, when resolving an Import for
$(MSBuildExtensionsPath)\Foo\Abc.targets, the following path will be checked:

    `/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild/Foo/Abc.targets`

a *new* /Library/Frameworks/Mono.framework/Versions/@VERSION@ directory
"lost". The file is still there, but xbuild will no longer find it.

To address this upgrade scenario, Mono's xbuild also checks a fallback path:

    `/Library/Frameworks/Mono.framework/External/xbuild`

Unfortunately, this means that $(MSBuildExtensionsPath) on xbuild is no longer
a single directory; it is a set of directories, and xbuild will check each
directory to see if the file exists:

/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild/Foo/Abc.targets
/Library/Frameworks/Mono.framework/External/xbuild/Foo/Abc.targets
...

Mono would like to eventually migrate to MSBuild instead of xbuild, which
requires bringing xbuild's fallback support to MSBuild.

These search paths can now be specified in the app.config like:

      <toolset toolsVersion="14.1">

        <msbuildExtensionsPathSearchPaths>
            <searchPaths os="osx">
                <property name="MSBuildExtensionsPath" value="/Library/Frameworks/Mono.framework/External/xbuild/;/tmp/foo"/>
                <property name="MSBuildExtensionsPath32" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                <property name="MSBuildExtensionsPath64" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
            </searchPaths>
        </msbuildExtensionsPathSearchPaths>

      </toolset>

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -10,6 +10,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Text;
 using Microsoft.Build.Construction;
 #if FEATURE_MSBUILD_DEBUGGER
 using Microsoft.Build.Debugging;
@@ -1891,26 +1893,21 @@ namespace Microsoft.Build.Evaluation
             }
 #endif
 
-            if (EvaluateConditionCollectingConditionedProperties(importElement, ExpanderOptions.ExpandProperties, ParserOptions.AllowProperties, _projectRootElementCache))
+            List<ProjectRootElement> importedProjectRootElements = ExpandAndLoadImports(directoryOfImportingFile, importElement);
+
+            foreach (ProjectRootElement importedProjectRootElement in importedProjectRootElements)
             {
-                string importExpressionEscaped = _expander.ExpandIntoStringLeaveEscaped(importElement.Project, ExpanderOptions.ExpandProperties, importElement.ProjectLocation);
+                _data.RecordImport(importElement, importedProjectRootElement, importedProjectRootElement.Version);
 
-                List<ProjectRootElement> importedProjectRootElements = ExpandAndLoadImports(directoryOfImportingFile, importExpressionEscaped, importElement);
-
-                foreach (ProjectRootElement importedProjectRootElement in importedProjectRootElements)
-                {
-                    _data.RecordImport(importElement, importedProjectRootElement, importedProjectRootElement.Version);
-
-                    // This key should be unique, as duplicate imports were already discarded
+                // This key should be unique, as duplicate imports were already discarded
 #if FEATURE_MSBUILD_DEBUGGER
-                    if (DebuggerManager.DebuggingEnabled)
-                    {
-                        _importRelationships.Add(importedProjectRootElement, importElement.ContainingProject);
-                    }
+                if (DebuggerManager.DebuggingEnabled)
+                {
+                    _importRelationships.Add(importedProjectRootElement, importElement.ContainingProject);
+                }
 #endif
 
-                    PerformDepthFirstPass(importedProjectRootElement);
-                }
+                PerformDepthFirstPass(importedProjectRootElement);
             }
 
 #if FEATURE_MSBUILD_DEBUGGER
@@ -2026,16 +2023,173 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
+        /// If the project import references a $(MSBuildExtensionsPath*)
+        /// then this tries to find project files corresponding to that, using various search
+        /// paths. It relies on the ExpandAndLoadImportsFromUnescapedImportExpressionConditioned
+        /// to do the actual loading and parsing.
+        /// This is explained in detail in a comment with the method code.
+        ///
+        /// If the project import does not reference the $(MSBuildExtensionsPath*) property, then
+        /// it falls back to the default behavior.
+        ///
+        /// Caches the parsed import into the provided collection, so future
+        /// requests can be satisfied without re-parsing it.
+        /// </summary>
+        private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportingFile, ProjectImportElement importElement)
+        {
+            var refKindInProject = MSBuildExtensionsPathReferenceKind.FindIn(importElement.Project);
+            var fallbackExtensionPaths = _data.Toolset.GetMSBuildExtensionsPathSearchPathsFor(refKindInProject);
+
+            // no reference or we need to lookup only the default path,
+            // so, use the Import path
+            if (fallbackExtensionPaths.Count == 0)
+            {
+                List<ProjectRootElement> projects;
+                var result = ExpandAndLoadImportsFromUnescapedImportExpressionConditioned(directoryOfImportingFile, importElement, importElement.Project, out projects);
+                return projects;
+            }
+
+            // $(MSBuildExtensionsPath*) usually resolves to a single value, single default path. This can be overridden
+            // by the usual property overriding techniques.
+            //
+            // Eg. <Import Project='$(MSBuildExtensionsPath)\foo\extn.proj' />
+            //
+            // But this feature allows that when it is used in an Import element, it will behave as a "search path", meaning
+            // that the relative project path "foo\extn.proj" will be searched for, in more than one location.
+            // Essentially, we will try to load that project file by trying multiple values (search paths) for the
+            // $(MSBuildExtensionsPath*) property.
+            //
+            // The various paths tried, in order are:
+            //
+            // 1. The value of the MSBuildExtensionsPath* property
+            //
+            // 2. Search paths available in the current toolset (via toolset.MSBuildExtensionsPathSearchPathsTable).
+            //    That may be loaded from app.config with a definition like:
+            //
+            //    <toolset .. >
+            //      <msbuildExtensionsPathSearchPaths>
+            //          <searchPaths os="osx">
+            //              <property name="MSBuildExtensionsPath" value="/Library/Frameworks/Mono.framework/External/xbuild/;/tmp/foo"/>
+            //              <property name="MSBuildExtensionsPath32" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
+            //              <property name="MSBuildExtensionsPath64" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
+            //          </searchPaths>
+            //      </msbuildExtensionsPathSearchPaths>
+            //    </toolset>
+            //
+            // This is available only when used in an Import element and it's Condition. So, the following common pattern
+            // would work:
+            //
+            //      <Import Project="$(MSBuildExtensionsPath)\foo\extn.proj" Condition="'Exists('$(MSBuildExtensionsPath)\foo\extn.proj')'" />
+            //
+            // The value of the MSBuildExtensionsPath* property, will always be "visible" with it's default value, example, when read or
+            // referenced anywhere else. This is a very limited support, so, it doesn't come in to effect if the explcit reference to
+            // the $(MSBuildExtensionsPath) property is not present in the Project attribute of the Import element. So, the following is
+            // not supported:
+            //
+            //      <PropertyGroup><ProjectPathForImport>$(MSBuildExtensionsPath)\foo\extn.proj</ProjectPathForImport></PropertyGroup>
+            //      <Import Project='$(ProjectPathForImport)' />
+            //
+
+            // Adding the value of $(MSBuildExtensionsPath*) property to the list of search paths
+            fallbackExtensionPaths.Insert(0, _data.GetProperty(refKindInProject.StringRepresentation).EvaluatedValue);
+
+            string extensionPropertyRefAsString = refKindInProject.MSBuildPropertyName;
+
+            _loggingService.LogComment(_buildEventContext, MessageImportance.Low, "SearchPathsForMSBuildExtensionsPath",
+                                        extensionPropertyRefAsString,
+                                        String.Join(Path.PathSeparator.ToString(), fallbackExtensionPaths));
+
+            bool atleastOneExactFilePathWasLookedAtAndNotFound = false;
+
+            // Try every extension search path, till we get a Hit:
+            // 1. 1 or more project files loaded
+            // 2. 1 or more project files *found* but ignored (like circular, self imports)
+            foreach (var extensionPath in fallbackExtensionPaths)
+            {
+                if (!Directory.Exists(extensionPath))
+                {
+                    continue;
+                }
+
+                var newExpandedCondition = importElement.Condition.Replace(extensionPropertyRefAsString, extensionPath);
+                if (!EvaluateConditionCollectingConditionedProperties(importElement, newExpandedCondition, ExpanderOptions.ExpandProperties, ParserOptions.AllowProperties,
+                            _projectRootElementCache))
+                {
+                    continue;
+                }
+
+                var newExpandedImportPath = importElement.Project;
+                newExpandedImportPath = newExpandedImportPath.Replace(extensionPropertyRefAsString, extensionPath);
+
+                _loggingService.LogComment(_buildEventContext, MessageImportance.Low, "TryingExtensionsPath", newExpandedImportPath, extensionPath);
+
+                List<ProjectRootElement> projects;
+                var result = ExpandAndLoadImportsFromUnescapedImportExpression(directoryOfImportingFile, importElement, newExpandedImportPath,
+                                        false, out projects);
+
+                if (result == LoadImportsResult.ProjectsImported)
+                {
+                    return projects;
+                }
+
+                if (result == LoadImportsResult.FoundFilesToImportButIgnored)
+                {
+                    // Circular, Self import cases are usually ignored
+                    // Since we have a semi-success here, we stop looking at
+                    // other paths
+                    return projects;
+                }
+
+                if (result == LoadImportsResult.TriedToImportButFileNotFound)
+                {
+                    atleastOneExactFilePathWasLookedAtAndNotFound = true;
+                }
+                // else if (result == LoadImportsResult.ImportExpressionResolvedToNothing) {}
+            }
+
+            // Found at least one project file for the Import, but no projects were loaded
+            // atleastOneExactFilePathWasLookedAtAndNotFound would be false, eg, if the expression
+            // was a wildcard and it resolved to zero files!
+            if (atleastOneExactFilePathWasLookedAtAndNotFound && (_loadSettings & ProjectLoadSettings.IgnoreMissingImports) == 0)
+            {
+                ThrowForImportedProjectFromExtensionsPathNotFound(refKindInProject, importElement);
+            }
+
+            return new List<ProjectRootElement>();
+        }
+
+        /// <summary>
+        /// Load and parse the specified project import, which may have wildcards,
+        /// into one or more ProjectRootElements, if it's Condition evaluates to true
+        /// Caches the parsed import into the provided collection, so future
+        /// requests can be satisfied without re-parsing it.
+        /// </summary>
+        private LoadImportsResult ExpandAndLoadImportsFromUnescapedImportExpressionConditioned(string directoryOfImportingFile, ProjectImportElement importElement, string unescapedExpression,
+                                            out List<ProjectRootElement> projects, bool throwOnFileNotExistsError=true)
+        {
+            if (!EvaluateConditionCollectingConditionedProperties(importElement, ExpanderOptions.ExpandProperties, ParserOptions.AllowProperties, _projectRootElementCache))
+            {
+                projects = new List<ProjectRootElement>();
+                return LoadImportsResult.ConditionWasFalse;
+            }
+
+            return ExpandAndLoadImportsFromUnescapedImportExpression(directoryOfImportingFile, importElement, importElement.Project, throwOnFileNotExistsError, out projects);
+        }
+
+        /// <summary>
         /// Load and parse the specified project import, which may have wildcards,
         /// into one or more ProjectRootElements.
         /// Caches the parsed import into the provided collection, so future 
         /// requests can be satisfied without re-parsing it.
         /// </summary>
-        private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportingFile, string importExpressionEscaped, ProjectImportElement importElement)
+        private LoadImportsResult ExpandAndLoadImportsFromUnescapedImportExpression(string directoryOfImportingFile, ProjectImportElement importElement, string unescapedExpression,
+                                            bool throwOnFileNotExistsError, out List<ProjectRootElement> imports)
         {
+            string importExpressionEscaped = _expander.ExpandIntoStringLeaveEscaped(unescapedExpression, ExpanderOptions.ExpandProperties, importElement.ProjectLocation);
             ElementLocation importLocationInProject = importElement.Location;
 
-            List<ProjectRootElement> imports = new List<ProjectRootElement>();
+            bool atleastOneImportIgnored = false;
+            imports = new List<ProjectRootElement>();
             string[] importFilesEscaped = null;
 
             try
@@ -2090,6 +2244,7 @@ namespace Microsoft.Build.Evaluation
                 if (String.Equals(_projectRootElement.FullPath, importFileUnescaped, StringComparison.OrdinalIgnoreCase) /* We are trying to import ourselves */)
                 {
                     _loggingService.LogWarning(_buildEventContext, null, new BuildEventFileInfo(importLocationInProject), "SelfImport", importFileUnescaped);
+                    atleastOneImportIgnored = true;
 
                     continue;
                 }
@@ -2114,6 +2269,7 @@ namespace Microsoft.Build.Evaluation
                         }
 
                         // Ignore this import and no more further processing on it.
+                        atleastOneImportIgnored = true;
                         continue;
                     }
                 }
@@ -2169,6 +2325,7 @@ namespace Microsoft.Build.Evaluation
                         }
 
                         // Since we have already seen this we need to not continue on in the processing.
+                        atleastOneImportIgnored = true;
                         continue;
                     }
                     else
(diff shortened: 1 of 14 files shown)


AndyGerlicher:
Just to give a heads up we do want this (perhaps even in master). We just need time to review and make sure it's an approach we're comfortable with and give feedback.

radical:
Thanks for the heads up! I can wait :)

AndyGerlicher:
Overall, here are our comments. Hopefully you're able to address them and we can get this committed. 

- Remove the feature flag. We'd like to move this to master and if it's going to be enabled in all our configs might as well remove it. 
- Few minor code changes (I'll comment on them individually)
- We need a better way to implement search paths and not hard code anything environment specific. Our proposal:
  - Use / read from `MSBuildExtensionsPathSearchPaths{OS}`
  - Move the `MSBuildExtensionsPathSearchPaths{OS}` property to the [toolsets section in the App.config](https://github.com/Microsoft/msbuild/blob/master/src/XMakeCommandLine/app.config#L38-L40).
  - This would allow us to have `MSBuildExtensionsPathSearchPathsOSX` set to `/Library/Frameworks/Mono.framework/External/xbuild` and `MSBuildExtensionsPathSearchPathsWindows` set to `C:\Program Files (x86)\MSBuild` (see below)

And of course feel free to propose something else if you can think of something better. Our main goal is we want to be able to take this in master in a generic way. We had discussed doing something like this as helping our OSS version work when it doesn't live in `Program Files` but other things do.

Thanks!

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+                    return s_pathsLists.Value.Path64;
+                case MSBuildExtensionsPathReferenceKind.None:
+                    return new List<string> ();
+                default:
+                    // new enum value added?
+                    throw new NotImplementedException();
+            }
+        }
+
+        private static List<string> ComputeFallbackExtensionsPathsFor(MSBuildExtensionsPathReferenceKind refKind)
```
AndyGerlicher: This method isn't called

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            var extnPaths = new List<string>();
+            var refKindAsString = ReferenceKindToString(refKind);
+            var envVar = Environment.GetEnvironmentVariable(refKindAsString + "SearchPaths");
+
+            var allPaths = String.Join(Path.PathSeparator.ToString(),
+                                        //FIXME: IsOSX returns false :/
+                                        NativeMethodsShared.IsUnix ? s_osxExternalXBuildDir : String.Empty,
```
AndyGerlicher: Did you mean IsUnixLike? Either way with the solution I proposed hopefully this should go away and be completely generic here.

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+                var refKindAsString = ReferenceKindToString(refKind);
+                var envVar = Environment.GetEnvironmentVariable(refKindAsString + "SearchPaths");
+
+                var allPaths = String.Join(Path.PathSeparator.ToString(),
+                                            //FIXME: IsOSX returns false :/
+                                            NativeMethodsShared.IsUnix ? s_osxExternalXBuildDir : String.Empty,
+                                            envVar ?? String.Empty);
+
+                var pathsTable = new Dictionary<string, string>();
+                foreach(var extnPath in allPaths.Split(new char[]{Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries)) {
```
AndyGerlicher: Style: { on a new line and space after foreach and if. Also we used to run StyleCop which would also require a {} around every if. It's not a hard requirement, but this would be the only place not conforming so you might as well. The CodeFormatter tool is great for auto correcting these style things, but we haven't ran that in a while it might have a lot of hits.

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+        ///<summary>
+        /// Helper method for resetting the lists for testing, since we need to run the
+        /// tests with different values for the *SearchPaths
+        ///</summary>
+        private static void ResetPathsListsForTesting()
+        {
+            s_pathsLists = new Lazy<ExtensionsPathsLists>(true);
+        }
+
+        class ExtensionsPathsLists
```
AndyGerlicher: Please be explicit on visibility of types / members (in the test class too).

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+                    pathsTable[extnPath] = extnPath;
+                    extnPaths.Add(extnPath);
+                }
+
+                return extnPaths;
+            }
+        }
+    }
+
+    enum MSBuildExtensionsPathReferenceKind {
```
AndyGerlicher: Please add /// doc and explicit `internal`

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+
+                var pathsTable = new Dictionary<string, string>();
+                foreach(var extnPath in allPaths.Split(new char[]{Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries)) {
+                    if(pathsTable.ContainsKey(extnPath))
+                        continue;
+
+                    if(!Directory.Exists(extnPath))
+                        continue;
+
+                    pathsTable[extnPath] = extnPath;
```
cdmihai: Looks a bit weird to have the key and value be the same thing. If you want to check for unique entries, can you do it with a HashSet instead, and thus also skip the first if in the loop?

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+                                            //FIXME: IsOSX returns false :/
+                                            NativeMethodsShared.IsUnix ? s_osxExternalXBuildDir : String.Empty,
+                                            envVar ?? String.Empty);
+
+                var pathsTable = new Dictionary<string, string>();
+                foreach(var extnPath in allPaths.Split(new char[]{Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries)) {
+                    if(pathsTable.ContainsKey(extnPath))
+                        continue;
+
+                    if(!Directory.Exists(extnPath))
```
cdmihai: Why not let callers check for existence? What if the caller code wants to do some funky lazy creation of the paths?

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
cdmihai: You could convert this enum into a struct with private constructor and four static values. Then, you can move all the enum specific static methods in the struct. With this you can also convert the ReferenceKindToString method into a an extra field of the struct (like StringRepresentation).

src/XMakeBuildEngine/Evaluation/MSBuildExtensionsPathHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            s_pathsLists = new Lazy<ExtensionsPathsLists>(true);
+        }
+
+        class ExtensionsPathsLists
+        {
+            /// <summary>
+            /// Cached list of extension paths to be used as search paths for $(MSBuildExtensionsPath)
+            /// </summary>
+            public readonly List<string> Default;
```
cdmihai: You could convert these 3 fields into a single Dictionary<MSBuildExtensionsPathReferenceKind, List<string>> and have GetFallbackExtensionsPathsFor delegate here instead of having that switch statement

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2062,6 +2206,7 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             foreach (string importFileEscaped in importFilesEscaped)
             {
+                atleastOneImportProcessed = true;
```
cdmihai: Can you move this to line 2324, right after it gets added in the imports list? Then, it really means that at least one got processed. Otherwise rename boolean flag to atLeastOneFound.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2026,15 +2021,164 @@ private bool EvaluateWhenOrOtherwiseChildren(IEnumerable<ProjectElement> childre
...
-        private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportingFile, string importExpressionEscaped, ProjectImportElement importElement)
+        /// <param name="atleastOneImportFoundAndIgnored">true, if one or more files were considered for import, but ignored (eg. circular or self imports)</param>
+        /// <param name="atleastOneImportProcessed">true, if one or more files were considered for import. This would be false if a wildcard resolved to zero file paths</param>
+        private List<ProjectRootElement> ExpandAndLoadImportsFromUnescapedImportExpression(string directoryOfImportingFile, ProjectImportElement importElement, string unescapedExpression,
+                                            bool throwOnFileNotExistsError, out bool atleastOneImportFoundAndIgnored, out bool atleastOneImportProcessed)
         {
+            string importExpressionEscaped = _expander.ExpandIntoStringLeaveEscaped(unescapedExpression, ExpanderOptions.ExpandProperties, importElement.ProjectLocation);
             ElementLocation importLocationInProject = importElement.Location;
 
+            atleastOneImportFoundAndIgnored = false;
```
cdmihai: Rename to atleastOneImportIgnored.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2026,15 +2021,164 @@ private bool EvaluateWhenOrOtherwiseChildren(IEnumerable<ProjectElement> childre
...
+            // not supported:
+            //
+            //      <PropertyGroup><ProjectPathForImport>$(MSBuildExtensionsPath)\foo\extn.proj</ProjectPathForImport></PropertyGroup>
+            //      <Import Project='$(ProjectPathForImport)' />
+            //
+
+            // Making a local copy for this instance of Evaluator, to add the the value of
+            // $(MSBuildExtensionsPath*) property that we have here, instead of keeping it
+            // in the static/shared MSBuildExtensionsPathHelper
+            var allExtensionPaths = new List<string>(fallbackExtensionPaths);
```
cdmihai: Make the helper be proactive and clone its lists before returning.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2026,15 +2021,164 @@ private bool EvaluateWhenOrOtherwiseChildren(IEnumerable<ProjectElement> childre
...
+            allExtensionPaths.Insert(0, _data.GetProperty(MSBuildExtensionsPathHelper.ReferenceKindToString(refKindInProject)).EvaluatedValue);
+
+            string extensionPropertyRefAsString = String.Format("$({0})", MSBuildExtensionsPathHelper.ReferenceKindToString(refKindInProject));
```
cdmihai: Move this to the future struct for refKind. Something like refKindInProject.ToMSBuildPropertyName()

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2026,15 +2021,164 @@ private bool EvaluateWhenOrOtherwiseChildren(IEnumerable<ProjectElement> childre
...
+
+                if(atleastOneImportWasIgnored) {
+                    // Circular, Self import cases are usually ignored
+                    // Since we have a semi-success here, we stop looking at
+                    // other paths
+                    return new List<ProjectRootElement>();
+                }
+
+                // Nothing was ignored, but if @atleastOneImportProcessed==true, then we did try
+                // to look at a file path and found it missing!
```
cdmihai: Why the conclusion that it was missing? atleastOneImportProcessed is set to true right at the beginning of the loop, and it may also mean that that path was found and processed into a ProjectRootElement

radical:
@AndyGerlicher Thanks for the feedback! `app.config` approach sounds good. Do you want the properties to be per toolset? Like:

      <toolset toolsVersion="14.1">   
        <property name="MSBuildToolsPath" value="." />   

        <property name="MSBuildExtensionsPathSearchPathsWindows" value="...." />
        <property name="MSBuildExtensionsPath32SearchPathsWindows" value="...." />
        <property name="MSBuildExtensionsPath64SearchPathsWindows" value="...." />

        <property name="MSBuildExtensionsPathSearchPathsOSX" value="...." />
        <property name="MSBuildExtensionsPath32SearchPathsOSX" value="...." />
        <property name="MSBuildExtensionsPath64SearchPathsOSX" value="...." />
      </toolset>

radical:
Or:  

      <toolset toolsVersion="14.1">   

         <msbuildExtensionsPathSearchPaths os="windows">
            <property name="MSBuildExtensionsPath" value="...." />
            <property name="MSBuildExtensionsPath32" value="...." />
            <property name="MSBuildExtensionsPath64" value="...." />
         </extensionPaths>

         <msbuildExtensionsPathSearchPaths os="osx">
            <property name="MSBuildExtensionsPath" value="...." />
            <property name="MSBuildExtensionsPath32" value="...." />
            <property name="MSBuildExtensionsPath64" value="...." />
         </extensionPaths>
      </toolset>

AndyGerlicher:
The second one looks better, but it might be overly complicated to implement. I would say go for the first:
```XML
<toolset toolsVersion="14.1">   
    <property name="MSBuildToolsPath" value="." />   

    <property name="MSBuildExtensionsPathSearchPathsWindows" value="...." />
    <property name="MSBuildExtensionsPath32SearchPathsWindows" value="...." />
    <property name="MSBuildExtensionsPath64SearchPathsWindows" value="...." />
```

radical:
I already implemented this:


    <msbuildToolsets default="14.1">
      <toolset toolsVersion="14.1">
        <property name="MSBuildToolsPath" value="." />
        <allMSBuildExtensionsPathSearchPaths>
            <msbuildExtensionsPathSearchPaths os="osx">
                <property name="MSBuildExtensionsPath" value="/Library/Frameworks/Mono.framework/External/xbuild/;/tmp/testing"/>
                <property name="MSBuildExtensionsPath32" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                <property name="MSBuildExtensionsPath64" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
            </msbuildExtensionsPathSearchPaths>
            <msbuildExtensionsPathSearchPaths os="windows">
                <property name="MSBuildExtensionsPath" value="C:\\foo\\bar"/>
                <property name="MSBuildExtensionsPath64" value="C:\\foo\\bar"/>
            </msbuildExtensionsPathSearchPaths>
        </allMSBuildExtensionsPathSearchPaths>
      </toolset>
    </msbuildToolsets>

Open to suggestions for the new element names.
FYI, I added this to be part of the ToolsetElement and Toolset class. And added the "read" part of
it to ToolsetConfigurationReader. It seemed like a more logical place for this.

I'm just fixing tests now and will try to update this PR tomorrow for review.

AndyGerlicher:
Sounds good, thanks!

src/XMakeBuildEngine/Evaluation/Evaluator.cs
radical: I have tried to make the code a little clearer in the updated patch.

radical:
@AndyGerlicher @cdmihai Thanks for the detailed feedback! I have tried to incorporate your suggestions in the updated PR. I have split the changes into logical self contained commits.

radical:
Should the value of OS specified in ` <searchPaths os="osx">` be case insensitive? It isn't currently.

radical:
BTW, there is a fix required for IsOSX on mono, before this will work on OSX. I will open a new PR for that. But this can reviewed without that.

radical:
Updated to show error location in the config file.

radical:
BTW, PR #518 should be merged before this, else the tests will fail on OSX.

radical:
Any feedback on this?

AndyGerlicher:
Sorry to take a while to get back! This is a decently big feature I wanted to take the time to try it out in a few scenarios. Here some feedback.

1. If you import `$(MSBuildExtensionsPath)\*.props`, it will only import from the first folder that has a hit. I thought that in your previous iteration it imported the union of all props in all search paths? Did this change? Or am I imagining things? My first instinct says it should be the union of all the results that match that are distinct file names, but I'm not sure that's right. Thoughts on that?

2. The error message could be better. 

Previously:
```
D:\test\fallback.proj(3,2): error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\14.0\notfound.props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
```
Now:
```
D:\test\fallback.proj(3,10): error MSB4019: The imported project "$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\notfound.props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
```

It really should note the normal expanded path, and ideally where else it looked. Debugging could end up being really tricky if not done.

Other than that looks good. If you can address # 2 I'd like to get this in. Thanks!

radical:
No problem. Thanks for taking the time and effort to review and give feedback :)

re:1, No, it stops at the first hit by design. Even the previous iteration had the same behavior. If you noticed otherwise, then that is a bug.

Thinking about the union behavior, it sounds like it might be useful for the `*.props` case, but ..

a. Security issue? Being able to inject targets, modify the build in subtle ways.

b. It would create dependence on the "environment", like random props file getting imported unintentionally. Wildcards might "catch"/import too much. It might be difficult to reason about stuff without being on the build machine.

IMHO, I don't think we should go for this.

re:2, error message. The details can actually be seen in a `/v:diag` log, like:

```
Search paths being used for $(MSBuildExtensionsPath) are /Users/ankit/dev/misc/lib/mono/xbuild:/Library/Frameworks/Mono.framework/External/xbuild/
Trying to import /Users/ankit/dev/misc/lib/mono/xbuild/non-existant.proj using extensions path /Users/ankit/dev/misc/lib/mono/xbuild
Trying to import /Library/Frameworks/Mono.framework/External/xbuild//non-existant.proj using extensions path /Library/Frameworks/Mono.framework/External/xbuild/
```

But this assumes that the user is aware that `$(MSBuildExtensionsPath)` is backed by multiple paths and they can see this with `/v:diag` , which might not be the case. And if the user is unaware, then s/he is lost! How about something like:

```
  /Users/ankit/test/msbuild/foo.proj(5,10): error MSB4019: The imported project "$(MSBuildExtensionsPath)/Bar/non-existant.proj" was not found. Tried to find the project file path `Bar/non-existant.proj` under the search paths - /Users/ankit/dev/misc/lib/mono/xbuild and /Library/Frameworks/Mono.framework/External/xbuild/ .  Confirm that the path in the <Import> declaration is correct, and that the file exists on disk in one of the search paths.
```

Maybe also mention that these are set in `/path/to/MSBuild.exe.config` ?

radical:
I think you were suggesting the error message to be like:
```
/Users/ankit/test/msbuild/foo.proj(5,10): error MSB4019: The imported project "/Users/ankit/dev/misc/lib/mono/xbuild/Bar/non-existant.proj" was not found. Also tried to find the project file path `Bar/non-existant.proj` under the search paths - /Library/Frameworks/Mono.framework/External/xbuild/ .  Confirm that the path in the <Import> declaration is correct, and that the file exists on disk in one of the search paths.
```

This shows the normal expanded path, expanded with the value of the property `$(MSBuildExtensionsPath)`. Which one is preferable?

rainersigwald:
@radical I think I like the form with the explicit default path, since that's most like the existing error. I might go a bit further though:

```
/Users/ankit/test/msbuild/foo.proj(5,10): error MSB4019: The imported project "/Users/ankit/dev/misc/lib/mono/xbuild/Bar/non-existant.proj" was not found. Since it was defined using $(MSBuildExtensionsPath), `Bar/non-existant.proj` was also seareched for in - /Library/Frameworks/Mono.framework/External/xbuild/ . Those paths are defined in MSBuild.exe.config.  Confirm that the path in the <Import> declaration is correct, and that the file exists on disk in one of the search paths.
```

The important addition being `Since it was defined using $(MSBuildExtensionsPath),`.

That's awfully wordy though.

radical:
> The important addition being Since it was defined using $(MSBuildExtensionsPath),.

Cool! I was thinking of mentioning that too, but I thought maybe I was being too cautious :)

How about - `... not found. Also, tried to find 'Bar/non-existant.proj' in the fallback search paths for $(MSBuildExtensionsPath) - /foo/bar and /xyz/abc . These search paths are defined in '/path/to/MSBuild.exe.config' . Confirm ...` ?

The *fallback* bit immediately suggests to the user the *order* of the search also. And it's consistent with the `Also, tried to..` .
I think the user can connect that `$(MSBuildExtensionsPath)` was relevant here because the import expression references it.

AndyGerlicher:
Sounds good @radical, not quite as wordy and conveys the search order I like that. And small subtlety since a lot of these paths are like `$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\...`, make sure that the rest of the path is expanded so the error doesn't have `$(MSBuildToolsVersion)` in it.

radical:
For an import `<Import Project="$(MSBuildExtensionsPath32)\$(Version)\non-existant.proj"/>`, the error message is:

  /Users/ankit/test/msbuild/foo.proj(8,10): error MSB4226: The imported project "/Users/ankit/dev/misc/lib/mono/xbuild/v0.1/non-existant.proj" was not found. Also, tried to find "v0.1/non-existant.proj" in the fallback search path(s) for $(MSBuildExtensionsPath32) - "/Library/Frameworks/Mono.framework/External/xbuild/" . These search paths are defined in "/Users/ankit/dev/msbuild/bin/Debug-MONO/OSX_Deployment/MSBuild.exe.config". Confirm that the path in the <Import> declaration is correct, and that the file exists on disk in one of the search paths.

I have not removed the diagnostic messages:

```
Search paths being used for $(MSBuildExtensionsPath32) are /Users/ankit/dev/misc/lib/mono/xbuild:/Library/Frameworks/Mono.framework/External/xbuild/
Trying to import /Users/ankit/dev/misc/lib/mono/xbuild/$(Version)/non-existant.proj using extensions path /Users/ankit/dev/misc/lib/mono/xbuild
Trying to import /Library/Frameworks/Mono.framework/External/xbuild//$(Version)/non-existant.proj using extensions path /Library/Frameworks/Mono.framework/External/xbuild/
```

Should these be removed now? I thought they might be useful in debugging, if say there is a crash before that error message shows up.

Another thing to note: I am using `AppDomain.CurrentDomain.SetupInformation.ConfigurationFile` to get the config file path, but I am not sure whether this will always be correct for our case.

radical:
Oh, .. `AppDomain` won't be in coreclr! I'll fix.

radical:
Another thing to note is that if no extension paths are found, then the error shown is the existing one. That is because in that case the code just executes the existing (regular import) code path.

And for CoreCLR case, the error message is missing the `These search paths ... .config .` part.

Build should be fixed now.

AndyGerlicher:
LGTM

src/XMakeBuildEngine/UnitTests/Definition/ToolsetConfigurationReader_Tests.cs
```diff
@@ -488,6 +500,182 @@ public void PropertyTest_GetValueByName()
...
+                       </msbuildExtensionsPathSearchPaths>
+                     </toolset>
+                   </msbuildToolsets>
+                 </configuration>"));
+
+            Configuration config = ToolsetConfigurationReaderTestHelper.ReadApplicationConfigurationTest();
+            ToolsetConfigurationSection msbuildToolsetSection = config.GetSection(s_msbuildToolsets) as ToolsetConfigurationSection;
+
+            Assert.Equal(msbuildToolsetSection.Default, "2.0");
+            Assert.Equal(1, msbuildToolsetSection.Toolsets.Count);
```
rainersigwald: I was going to complain that the order of arguments is expected, actual--but this whole file is wrong. So leave things like this and maybe we can come along and fix it in the future.

radical: Oops :( Sure, I can fix it after this is merged. Was there anything else wrong with the file?

radical: Oh, the existing stuff is wrong too :/

rainersigwald: Exactly . . . no point worrying about it now.

rainersigwald: (and better to be consistent with the rest of the file than correct, in this case)

rainersigwald:
:+1:

radical:
Yay! Thanks for taking the time with the reviews and feedback :)

PR #483, opened 2016-02-14
dir.props: Do not disable debug file generation on Mono

radical:


AndyGerlicher:
LGTM

AndyGerlicher:
Can you move the DebugSymbols property under the Debug-MONO specific property group `<PropertyGroup Condition="'$(Configuration)' == 'Debug-MONO'">`?

We think `UseRoslynCompilers` property is stale and the entire block can be deleted. Can you remove that [block](https://github.com/radical/msbuild/blob/mono-debug/dir.props#L20-L22) please?

radical:
Updated

rainersigwald:
I'm still a bit suspicious about why we're not producing symbols when we're building with CoreCLR/Roslyn. We can fix that later, though. LGTM.

PR #484, opened 2016-02-14
CommandLineTest: Use correct path to 'cat' on !Windows

radical:
Environment.SpecialFolder.System resolves to "" on Mono, so, instead
hardcode the path to "/bin/cat" on !Windows


rainersigwald:
LGTM, thanks.

PR #485, opened 2016-02-14
XMakeAppTests: Copy MSBuild.exe dependencies also, when trying to run

radical:
from an arbitrary path. This will ensure that the test is actually
running against the built MSBuild, instead of the installed MSBuild
assemblies (via GAC).

This fixes the two tests on OSX.


AndyGerlicher:
Looks good. This would probably help with some of the tests in master as well that are currently ignored. The test suite these came from assumed MSBuild was GACed to run the tests, there are still some left that need this :(.

Thanks!

PR #486, opened 2016-02-14
Expander_Tests: Use Path.Combine to construct paths instead of

radical:
hardcoding like "foo\\bar".
Fixes some tests on OSX.


AndyGerlicher:
LGTM

rainersigwald:
@radical Can you look at the test failures? It looks like they're happening in tests that you touched.

radical:
Sorry about that, fixed now.

PR #487, opened 2016-02-14
On Mono, use PEReader in AssesmblyInformation.GetFrameworkName for

radical:
getting the TargetFrameworkAttribute custom attribute.

Say, it is trying to get the attribute for Abc.dll, but that assembly
has a custom attribute from an assembly Xyz.dll. Mono will try to look
for Xyz.dll, and crash if it cannot find that. This was encountered
while trying to build some projects.

GetFrameworkName has an alternate implementation which uses
System.Reflection.Metadata.PEReader to get the metadata. So, choose that
implementation for MONO builds.

Also, this required adding a reference to System.Collections.Immutable
in project.json .


AndyGerlicher:
LGTM

PR #488, opened 2016-02-15
Disable windowsonly tests when running on !Windows platforms

radical:
These changes were in the code earlier, but seem to have been lost
in the nunit->xunit migration. But they are still relevant, so, adding them
again.

### DIFF
--- a/src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
+++ b/src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
@@ -56,6 +56,7 @@ namespace Microsoft.Build.UnitTests
         }
 
         [Fact]
+        [PlatformSpecific(Xunit.PlatformID.Windows)]
         public void GetApiContractReferencesFindsWinMDs()
         {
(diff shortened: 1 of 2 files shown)


radical:
Updated the PR to fix an issue (use -notrait .. instead of -exclude) with my earlier patch.

radical:
Changed the category `WindowsOnly` to `windowsonly` to be consistent with other categories.

cdmihai:
Can you please use the [PlatformSpecific] attribute? (example: https://github.com/Microsoft/msbuild/blob/xplat/src/Shared/UnitTests/FileUtilities_Tests.cs#L855)

It's an xunit extension msbuild inherits from the buildtools package: https://github.com/dotnet/buildtools/blob/f6361d1f2d53a64ae6c0971be959981390f9d8a2/src/xunit.netcore.extensions/Discoverers/PlatformSpecificDiscoverer.cs#L23

radical:
@cdmihai ah, my bad. Fixed.

rainersigwald:
LGTM, thanks.

Issue #491, opened 2016-02-17
Tests broken on xplat, CI server/script issue


radical:
The tests seem to be broken on xplat. The earlier failing log shows (http://dotnet-ci.cloudapp.net/job/Microsoft_msbuild/job/innerloop_xplat_Windows_NT/44/console)

```08:51:44      TwoSecondEmpty [FAIL]
08:51:44         System.ArgumentException : Xml_InvalidCommentChars
08:51:44         Stack Trace:
08:51:44            d:\j\workspace\innerloop_xpl---31ff7452\src\Shared\Compat\XmlTextWriter.cs(646,0): at System.Xml.XmlTextWriter.WriteComment(String text)
08:51:44               at System.Xml.XmlDocument.Save(XmlWriter w)
08:51:44            d:\j\workspace\innerloop_xpl---31ff7452\src\XMakeBuildEngine\Definition\Project.cs(1614,0): at Microsoft.Build.Evaluation.Project.SaveLogicalProject(TextWriter writer)
08:51:44            Evaluation\Preprocessor_Tests.cs(303,0): at Microsoft.Build.UnitTests.Preprocessor.Preprocessor_Tests.TwoSecondEmpty()```

These tests write current path inside an XML comment, but here the path is

`d:\j\workspace\innerloop_xpl---31ff7452\src...`

.. with the invalid `---` !

Older CI builds were using a different path `d:\j\workspace\Microsoft_msbuild\innerloop_xplat_Windows_NT\..` (http://dotnet-ci.cloudapp.net/job/Microsoft_msbuild/job/innerloop_xplat_Windows_NT/43/console)
with no invalid chars.

The tests should be fixed to handle this.

radical:
Fixed now.

radical:
Looks like this isn't the correct fix :/ The failure happens before we can do the string replace in the code. Investigating.

radical:
The remaining issue was that the path was being written out as a comment while saving the project file.

radical:
Fixed!

PR #492, opened 2016-02-17
Fix issue #491 - test breakage on CI

radical:
These tests include the current directory inside an xml comment. If the
path happens to contain `--`, which is not allowed inside xml comments,
then these tests break. This showed up on CI servers where the path did
include these characters!

Issue: https://github.com/Microsoft/msbuild/issues/491


radical:
IIUC, trying to escape the xml won't help, so, doing a string replace.

rainersigwald:
LGTM, thanks.

PR #493, opened 2016-02-18
Show build status in a table

cdmihai:


AndyGerlicher:
Would probably be nice to show master and xplat status in both branches.

rainersigwald:
That was what I was thinking, too, @AndyGerlicher. This is good, though.

PR #497, opened 2016-02-19
[mono] Handle framework assembly path names like `4.5-api`

radical:
The paths used to be of the form

  ../xbuild/4.5/..

but that has changed to

  ../xbuild/4.5-api/..

which breaks the regex. We fix the regex to handle any suffix
after the version numbers. This fixes the build with latest
(master) mono and continues to work with released mono.

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -601,7 +601,7 @@ namespace Microsoft.Build.Shared
             {
                 return s_frameworkPathRegex
                        ?? (s_frameworkPathRegex =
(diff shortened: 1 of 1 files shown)


radical:
After this, a new mono-msbuild.zip will need to be released, since the current one cannot build mono/master.

rainersigwald:
@dotnet-bot test this please

(Should run OS X CoreCLR tests . . .)

akoeplinger:
@radical can you explain to me what the FixFrameworkPath code actually tries to accomplish, I have a hard time following it :smile: 

xbuild/4.5-api/ **does not** exist (if you see that then it's a bug). I've been spending the last two weeks dealing with fallout of the -api rename, so I'm interested in why we need this code here (ping me in Slack if you want more details).

radical:
I discussed this with @akoeplinger and looks like a deeper look at the code is required. Closing this PR. Thanks @akoeplinger

Issue #464, opened 2016-02-03
Flag for forcing output with colors when redirecting stdout


assafnativ:
Currently MSBuild has nice and helpful command line colors. The colors help a great deal when scanning a long output manually.
However, when redirecting the MSBuild output, like when executing MSBuild from Jenkins for instance, the colors are automatically removed. In most cases this is a reasonable thing to do, because most text editors can't parse the ANSI escape characters.
However, there is flag that was made just for that (/clp:DisableConsoleColor).
It seems like the colors are removed only becuase the C# ConsoleColors API is not good enought, it throw an exception when used while STDOUT is redirected to a file.
I think it's not a good idea to make MSBuild spit ANSI escape characters at this points, becuase users are not expecting that when redirecting the output. But there is a room for a new flag that forces the colors.
Therefore, I suggested my Pull Request #458.

AndyGerlicher:
I'm not opposed to this entirely, but I'm a little worried that it's pretty corner case to implement in the standard console logger. Also if you want to write colorized file, it seems like using a logger that was aware of the output file format would be a much better choice than modifying the console logger and redirecting to a file. This would be very easy to implement in that way. Thoughts?

assafnativ:
1. Implementing a new logger requires overriding the SetColor method which is defined as "internal static void SetColor (ConsoleColor c)". The code duploication in the BaseConsoleLogger.cs and ConsoleLogger.cs, makes it even harder.
2. It seems like there is a full support for colors in the default logger. However, the poor C# ConsoleColors API is throwing exception when used with redirection and all of the hard work that was done in order to generate these colors in the first place is lost.
3. The Jenkins + MSBuild environment is a very common one, everyone who uses it need to replace the default logger to get the data that is already in there.
4. It was quite hard for me to follow the trail of compiling to pinpoint what part is to blame for not having colors.
5. Colors are awesome, you want them in MSBuild.
BTW: Your Jenkins builds of MSBuild itself is so colorless for the same reason.

AndyGerlicher:
Sorry for the delay. We discussed this one at length in triage today. We really don't like that it re-implements what should be core .NET functionality and don't want to be in the business of doing that when possible.

However, we agreed that we'd like to take this change in this case. I like that you've refactored and reduced code duplication and made the change optional behind a feature flag. And I do agree this could benefit a lot of people on CI systems such as our Jenkins build.

[...]

PR #505, opened 2016-02-24
Option for using ANSI-Colors that works with redirection

assafnativ:
Eliminat some code duplication, SetColor and ResetColor were already
defined in BaseConsoleLogger.cs

### DIFF
--- a/src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
+++ b/src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
@@ -399,6 +399,48 @@ namespace Microsoft.Build.BackEnd.Logging
         }
 
         /// <summary>
+        /// Sets foreground color to color specified using ANSI escape codes
+        /// </summary>
+        /// <param name="c">foreground color</param>
+        internal static void SetColorANSI(ConsoleColor c)
+        {
+            string colorString = "\x1b[";
+            switch (c)
+            {
+                case ConsoleColor.Black: colorString += "30"; break;
+                case ConsoleColor.DarkBlue: colorString += "34"; break;
+                case ConsoleColor.DarkGreen: colorString += "32"; break;
+                case ConsoleColor.DarkCyan: colorString += "36"; break;
+                case ConsoleColor.DarkRed: colorString += "31"; break;
+                case ConsoleColor.DarkMagenta: colorString += "35"; break;
+                case ConsoleColor.DarkYellow: colorString += "33"; break;
+                case ConsoleColor.Gray: colorString += "37"; break;
+                case ConsoleColor.DarkGray: colorString += "30;1"; break;
+                case ConsoleColor.Blue: colorString += "34;1"; break;
+                case ConsoleColor.Green: colorString += "32;1"; break;
+                case ConsoleColor.Cyan: colorString += "36;1"; break;
+                case ConsoleColor.Red: colorString += "31;1"; break;
+                case ConsoleColor.Magenta: colorString += "35;1"; break;
+                case ConsoleColor.Yellow: colorString += "33;1"; break;
+                case ConsoleColor.White: colorString += "37;1"; break;
+                default: colorString = ""; break;
+            }
+            if ("" != colorString)
+            {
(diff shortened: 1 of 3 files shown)


assafnativ:
#464 related

src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
```diff
@@ -370,7 +370,7 @@ internal bool IsVerbosityAtLeast(LoggerVerbosity checkVerbosity)
         /// <summary>
         /// Sets foreground color to color specified
         /// </summary>
-        internal static void SetColor(ConsoleColor c)
+        public static void SetColor(ConsoleColor c)
```
AndyGerlicher: Is there a reason to make this public? Would like to keep internal if at all possible, doesn't seem like it needs to change.

src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
```diff
@@ -399,13 +399,55 @@ internal static void ResetColor()
         }
 
         /// <summary>
+        /// Sets foreground color to color specified using ANSI escape codes
+        /// </summary>
+        /// <param name="c">foreground color</param>
+        public static void SetColorANSI(ConsoleColor c)
```
AndyGerlicher: Here too, please make this internal

src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
```diff
@@ -399,13 +399,55 @@ internal static void ResetColor()
...
+
+        /// <summary>
         /// Changes the foreground color to black if the foreground is the
         /// same as the background. Changes the foreground to white if the
         /// background is black.
         /// </summary>
         /// <param name="foreground">foreground color for black</param>
         /// <param name="background">current background</param>
-        internal static ConsoleColor TransformColor(ConsoleColor foreground,
+        static ConsoleColor TransformColor(ConsoleColor foreground,
```
AndyGerlicher: Revert this and keep the explicit visibility please.

src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
assafnativ: I had to change it to public to eliminate the code duplication. You had a copy paste of the function in Logging/ConsoleLogger.cs

assafnativ: See: Logging\ConsoleLogger.cs line 84 and 85. Make it a protected method doesn't work either.

AndyGerlicher: Protected definitely wouldn't work since as you mention it's being called from the ConsoleLogger class. However, they're both in the same assembly so keeping it internal should just work. Let me know if I'm missing a scenario though. 

(Just to make sure I cloned your changes and switched to internal, all tests and the build passed with no issues)

assafnativ:
Fixed according to the comments.

AndyGerlicher:
Code LGTM. I'd like to merge this. One last thing though, do you have a recommendation on how to view a file with these codes? I tried running a build with the flag and the command window colors look fine, and redirected to a file I see the escape characters:
[out.txt](https://github.com/Microsoft/msbuild/files/159116/out.txt)

I tried viewing this with Sublime and the Sublime ANSI package but there were no colors. But their example didn't work either so I don't think it's this file. Do you have a viewer you use and recommend? I just want to make sure this is right.

assafnativ:
I use it with the Jenkins AnsiColor plugin.
According to Stackoverflow, Sublime seems to support it using a plugin http://stackoverflow.com/questions/2271785/text-editor-viewer-with-ansi-codes-rendering-support-for-windows
I use VIM on Windows with AnsiColors plugin from here https://github.com/powerman/vim-plugin-AnsiEsc

AndyGerlicher:
@dotnet-bot Test this please

rainersigwald:
LGTM

PR #506, opened 2016-02-26
Add support for multiprocess builds ("/m") in .NET Core MSBuild

### LINKED ISSUE #303: Enable /m in CoreCLR (xplat)
This features was turned off due to named pipe limitations and binary serialization missing from .NET Core. This feature should be enabled and supported.

dsplaisted:
Resolves #303


src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -390,12 +393,19 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
 
             BackendNativeMethods.PROCESS_INFORMATION processInfo = new BackendNativeMethods.PROCESS_INFORMATION();
 
-            string appName = msbuildLocation;
+
+#if RUNTIME_TYPE_NETCORE
+            string coreRunName = NativeMethodsShared.IsUnixLike ? "CoreRun" : "CoreRun.exe";
```
akoeplinger: I think this should be `corerun` on non-Windows (at least according to https://github.com/dotnet/buildtools/blob/c578e891af4dea18b8f7e2524780bd4a9a3f8ce4/src/Microsoft.DotNet.Build.Tasks/PackageFiles/toolruntime.targets#L12)

AndyGerlicher:
LTGM. Do we need to update the handshake so there's no possibility of mixing nodes?

Also, perhaps we should emit a warning that this is experimental use at your own risk until we're comfortable with the security implications?

And we should make this the default in our self-host scenario.

dsplaisted:
OK, I've switch to use a pair of anonymous pipes on .NET Core instead of a single duplex named pipe, which makes multiprocess builds actually work on Linux and also resolves the security implications.  Enough has changed that I think it's worth re-reviewing the whole thing.

src/Shared/NodeEndpointOutOfProcBase.cs
```diff
@@ -312,7 +345,15 @@ private void InitializeAsyncPacketThread()
...
         private void PacketPumpProc()
         {
+#if FEATURE_NAMED_PIPES_FULL_DUPLEX
             NamedPipeServerStream localPipeServer = _pipeServer;
+            PipeStream localWritePipe = _pipeServer;
+            PipeStream localReadPipe = _pipeServer;
+#else
+            PipeStream localWritePipe = _pipeClientToServer;
+            PipeStream localReadPipe = _pipeServerToClient;
+#endif
```
rainersigwald: Nice way to hide this difference later :+1:

src/Shared/NodePacketTranslator.cs
```diff
@@ -341,7 +341,30 @@ public void TranslateDotNet<T>(ref T value)
                 BinaryFormatter formatter = new BinaryFormatter();
                 value = (T)formatter.Deserialize(_packetStream);
             }
+#else
+            public void TranslateBuildEventArgs(ref BuildEventArgs value)
+            {
+                var serializedBuildEventArgs = new SerializedBuildEventArgs();
+                serializedBuildEventArgs.Translate(this);
+                value = SerializedBuildEventArgs.ToEventArgs(serializedBuildEventArgs);
```
rainersigwald: What's the logic of mutating the input?

src/Shared/NodePacketTranslator.cs
```diff
@@ -1027,5 +1070,407 @@ public bool TranslateNullable<T>(T value)
...
+                translator.Translate(ref fieldName);
+
+                if (translator.Mode == TranslationDirection.ReadFromStream)
+                {
+                    LoadedType loadedDclaringType = _typeLoader.Load(declaringTypeFullName, assemblyLoadInfo);
+                    Type declaringType = loadedDclaringType.Type;
+                    Field = declaringType.GetField(fieldName, BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
+                }
+
+                var fieldType = Field.FieldType;
```
rainersigwald: nit: I'd put a blank line after this

rainersigwald:
LGTM

src/XMakeCommandLine/XMake.cs
```diff
@@ -3081,9 +3134,7 @@ private static void ShowHelpMessage()
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_11_LoggerSwitch"));
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_15_ValidateSwitch"));
             Console.WriteLine(AssemblyResources.GetString("HelpMessage_19_IgnoreProjectExtensionsSwitch"));
-#if FEATURE_APPDOMAIN
```
AndyGerlicher: This should still be #if FEATURE_NODE_REUSE, right?

AndyGerlicher:
LGTM

src/Shared/NodePacketTranslator.cs
dsplaisted: I don't follow.  The signature of the method allows for the value to be written to the stream or read from it, and this is the implementation that reads from the stream.

dsplaisted:
@dotnet-bot Test this please

src/Shared/NodePacketTranslator.cs
rainersigwald: I missed context of being in the read mode. With that, this makes perfect sense.

PR #511, opened 2016-03-03
Explicitly specify /nodeReuse in node creation

### LINKED ISSUE #509: Global msbuild.rsp with /nr:false can break /m builds
This was discovered internally by @AArnott. Thanks for the report and for giving us access to that repro machine.

MSBuild ships with an empty `msbuild.rsp` next to `msbuild.exe` in Program Files. If edited, any line in this response file is effectively prepended to the command line of a normal, undecorated `msbuild` invocation. Presumably this is to allow easily setting defaults for MSBuild across the entire machine.

Since #346, if this response file contains `/nr:false` to disable node reuse, multiproc builds will a) actually use only one process while b) continually spawning new instances of `msbuild.exe /nodeMode:1` until that one process completes.

This is because cc051cb9d4fb8507144aabdd7bfbc44669c862d2 changed how the child/client/worker node handled the `/nodeReuse` command line switch. Before, it was ignored and the child process would always live until it was sent a kill signal or it hit its timeout. After, it uses the value of the switch (after response-file inclusion, and defaulting to true) as part of its handshake computation.

In this situation, the main MSBuild spawned a new child process with no `/nodeReuse` flag, which in the absence of a response file is equivalent to `/nodeReuse:true`. Since there *was* a response file, the child actually started with `/nodeReuse:false`, computed its handshake value using that, and rejected the connection attempt that wanted `/nodeReuse:true`. For every attempted child process, in a loop, until the build ended.

AndyGerlicher:
Closes #509

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
+++ b/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
@@ -112,14 +112,12 @@ namespace Microsoft.Build.BackEnd
             }
 
             // Start the new process.  We pass in a node mode with a node number of 1, to indicate that we 
-            // want to start up just a standard MSBuild out-of-proc node. 
-            string commandLineArgs = " /nologo /nodemode:1 ";
-
-            // Disable node re-use if it is not requested (because no argument means enable reuse).
-            if (!ComponentHost.BuildParameters.EnableNodeReuse)
-            {
-                commandLineArgs += "/nodeReuse:false";
-            }
+            // want to start up just a standard MSBuild out-of-proc node.
+            // Note: We need to always pass /nodeReuse to ensure the value for /nodeReuse from msbuild.rsp
+            // (next to msbuild.exe) is ignored.
+            string commandLineArgs = ComponentHost.BuildParameters.EnableNodeReuse ? 
+                "/nologo /nodemode:1 /nodeReuse:true" : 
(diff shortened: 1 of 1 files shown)


src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -112,14 +112,12 @@ public bool CreateNode(int nodeId, INodePacketFactory factory, NodeConfiguration
...
-            // want to start up just a standard MSBuild out-of-proc node. 
-            string commandLineArgs = " /nologo /nodemode:1 ";
-
-            // Disable node re-use if it is not requested (because no argument means enable reuse).
-            if (!ComponentHost.BuildParameters.EnableNodeReuse)
-            {
-                commandLineArgs += "/nodeReuse:false";
-            }
+            // want to start up just a standard MSBuild out-of-proc node.
+            // Note: We need to always pass /nodeReuse to ensure the machine-wide msbuild.rsp setting for
```
rainersigwald: I'd just say "any response file setting for"--we really mean "the one next to msbuild.exe" and the meaning of that changes--for a build from GitHub it's just in the local folder.

rainersigwald:
LGTM

AndyGerlicher:
@rainersigwald better? Updated the comment.

rainersigwald:
Yeah, sounds good now.

PR #513, opened 2016-03-03
Fix the find error

jasonwilliams200OK:
On OS X, it gives the error:

```bash
find: ftsopen: No such file or directory
```

and on Ubuntu:

```bash
find: cannot search `': No such file or directory (TaskId:2)
```
(both errors can be verified on CI run before this PR)

This change fixes that error.

### DIFF
--- a/src/.nuget/packageLoad.targets
+++ b/src/.nuget/packageLoad.targets
@@ -81,7 +81,7 @@
           AfterTargets="_RestoreBuildToolsPackagesConfig"
           Condition="'$(OsEnvironment)'!='Windows_NT'">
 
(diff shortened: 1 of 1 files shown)


jasonwilliams200OK:
Before: http://dotnet-ci.cloudapp.net/job/Microsoft_msbuild/job/_OSX_prtest/193/console

Search on page "find:" -> 1 hit

After: http://dotnet-ci.cloudapp.net/job/Microsoft_msbuild/job/_OSX_prtest/199/console

0 hits

jasonwilliams200OK:
Test Windows_NT Build please.

cdmihai:
Thank you for looking into this!

Can you please simplify the fix by changing DnuToolDir to DnuToolPath and not removing DnuToolPath?
Though it kinda looks superfluous, I kept DnuToolPath because the targets inside Build Tools still use dnu referencing property names (e.g., their nuget restore targets reference DnuRestoreCommand). For the sake of consistency, I then tried to make sure the dnu properties are still there but reference dotnet cli (got the trick from the corefx repo).

jasonwilliams200OK:
Thanks for the review @cdmihai! I have fixed it up. Now the change is confined to just one file. :)

rainersigwald:
LGTM

PR #516, opened 2016-03-08
Support self-building MSBuild from the .NET Core deployment folder

dsplaisted:
With these changes, you should be able to build the MSBuild project for .NET Core, and then use the version you just built in the output folder to rebuild the MSBuild project.  This should make it easier to test changes to .NET Core MSBuild on real projects.

Assuming the path to your repo is `c:\git\msbuild`, you can use the following commands:

```cmd
cd /d c:\git\msbuild

:: build MSBuild for .NET Core
build /p:Configuration=Debug-NetCore /t:Build

:: Use just-built .NET Core MSBuild to build into bin\bin2 folder
"C:\git\msbuild\bin\Debug-NetCore\Windows_NT_Deployment\CoreRun.exe" "C:\git\msbuild\bin\Debug-NetCore\Windows_NT_Deployment\MSBuild.exe" c:\git\msbuild\build.proj /p:Configuration=Debug-NetCore /p:BinDir=c:\git\msbuild\bin\bin2\ /t:build

:: Inception!  Use bin2 version to build a bin3 version
"C:\git\msbuild\bin\bin2\Debug-NetCore\Windows_NT_Deployment\CoreRun.exe" "C:\git\msbuild\bin\bin2\Debug-NetCore\Windows_NT_Deployment\MSBuild.exe" c:\git\msbuild\build.proj /p:Configuration=Debug-NetCore /p:BinDir=c:\git\msbuild\bin\bin3\ /t:build
```


targets/DeployDependencies.proj
```diff
@@ -20,17 +20,40 @@
...
+      <BuildToolsVersionFile>$(RepoRoot)BuildToolsVersion.txt</BuildToolsVersionFile>
+    </PropertyGroup>
+    <ReadLinesFromFile File="$(BuildToolsVersionFile)">
+      <Output TaskParameter="Lines" ItemName="BuildToolsVersion" />
+    </ReadLinesFromFile>
+    <PropertyGroup>
+      <BuildToolsVersion>@(BuildToolsVersion)</BuildToolsVersion>
+    </PropertyGroup>
+    <ItemGroup>
+      <BuildToolsFiles Include="$(ToolPackagesDir)Microsoft.DotNet.BuildTools\$(BuildToolsVersion)\lib\*.*" />
```
cdmihai: Is everything in there needed to be in the msbuild deployment folders? Or only a subset of those files?

cdmihai:
Awesome! 
LGTM with minor comment.

targets/DeployDependencies.proj
```diff
@@ -20,17 +20,40 @@
...
+  </Target>
 
   <Target Name="CopyPackageContent">
     <ItemGroup>
       <!-- We want the list of files *after* restore, which happens
            early in build, so this item can't be defined outside of a
            target. -->
-      <NuspecFilesInPackagesDir Include="$(PackagesDir)\**\*.nupkg" />
-      <PackagesFoldersInPackagesDir Include="@(NuspecFilesInPackagesDir->'%(RootDir)%(Directory)')" />
+      <PackagesFoldersInPackagesDir Include="$(PackagesDir)\Microsoft.Portable.Targets\0.1.1-dev\" />
```
rainersigwald: Maybe move this version string into its own property? We've done that with some other versions.

rainersigwald:
LGTM. In master this is done by `RebuildWithLocalMSBuild.cmd` but in xplat we mangled that to just do a netcore build. We should make it do this and obey its name again (doesn't have to be right now though).

jasonwilliams200OK:
Awesome! :+1: 

Just FYI, we can add remarks using the DOS comment delimiter (`::`) in github markdown (with `````` ```cmd``````):

```cmd
:: some comment
type code.cs
```

or PowerShell variant (with `````` ```posh`````` or `````` ```powershell``````):

```posh
# some comment
cat code.cs
```

dsplaisted:
@jasonwilliams200OK Thanks, I've modified it to use `::` comments.

targets/DeployDependencies.proj
dsplaisted: The build tools package copies all the files.  Initially I tried copying a subset in this target, but got errors because files were missing.  So instead of trying to find the right set of files through trial and error I decided to just copy all the files.

targets/DeployDependencies.proj
dsplaisted: This is the only place this number will occur in a build file, so factoring it out doesn't reduce duplication.  Plus, this will go away entirely once we start using a version of .NET CLI which supports contentFiles.

Issue #503, opened 2016-02-23
Why is NativeMethodsShared.FixFrameworkPath required?


radical:
This method essentially converts paths of the form

`/framework/base/path/vA.B/foo` to `/framework/base/path/xbuild/vA.B/bin/foo` 

.. if A.B > 4.0 .

I am not sure why this is required or when it might be hit. The `xbuild/*/bin` path
contains the Microsoft.Build* assemblies for versions 12.0 and 14.0..
But how would we get a resolved path to these?

Also, this gets applied to results of property/item expansions via `MaybeAdjustFilePath`,
which means that any Property containing such a path gets adjusted. And this can get
applied multiple times, so ..

` <Foo>/Users/ankit/dev/misc/lib/mono/v8.0/asd</Foo>`

.. becomes ..

`/Users/ankit/dev/misc/lib/mono/xbuild/8.0/bin/bin/asd` 

So, my question is why is this useful?

AndyGerlicher:
@ValMenn does this look familiar?

We're not really sure. Probably seemed like a good idea at the time. I don't know of any reason for or against keeping it.

radical:
Thanks!

PR #517, opened 2016-03-08
NativeMethodsShared.FixFrameworkPath: Remove

radical:
This method tries to adjust paths that seem to be framework paths and
seem to have a major version higher than 4, and maps them to a different
path like `../xbuild/A.B/bin/..` .

/framework/base/path/vA.B/foo to /framework/base/path/xbuild/vA.B/bin/foo

It's not clear what this was trying to fix, but it is unnecessary and
incorrect. This gets used in a common code path for processing
properties, so it ends up getting applied to all properties and can be
applied more than once, which results in mangled paths, like:

<Foo>/Users/ankit/dev/misc/lib/mono/v8.0/asd</Foo>

.. becomes ..

/Users/ankit/dev/misc/lib/mono/xbuild/8.0/bin/bin/asd

Issue: #503

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -461,17 +461,6 @@ namespace Microsoft.Build.Shared
                 }
             }
 
-            // See if we have a framework path and adjust it if needed
-            if (checkValue.StartsWith("/"))
-            {
-                var adjustedValue = NativeMethodsShared.FixFrameworkPath(checkValue);
-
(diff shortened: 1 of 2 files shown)


AndyGerlicher:
LGTM

cdmihai:

radical:
Could we release a new build of mono-msbuild for bootstrapping, after we merge this? It would then work the mono/master and the upcoming mono release.

PR #518, opened 2016-03-09
NativeMethodsShared.IsOSX: Fix this for mono

radical:
Due to historical reasons[1], Environment.OSVersion.Platform returns
a value of `Unix` even for OSX. So, instead check for existence of
    /usr/lib/libc.dylib

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -517,7 +517,9 @@ namespace Microsoft.Build.Shared
         {
             get
             {
-#if FEATURE_OSVERSION
+#if MONO
+                return File.Exists("/usr/lib/libc.dylib");
+#elif FEATURE_OSVERSION
                 return Environment.OSVersion.Platform == PlatformID.MacOSX;
 #else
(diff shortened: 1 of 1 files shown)


radical:
re:GetUnixName, it uses PAL_UNIX_NAME, which IIUC is set at configure time, so, we can't use this to detect the OS at runtime. ( [1](https://github.com/dotnet/corefx/blob/967b0b0a576f212bc94a14896e716597ea708e95/src/Native/configure.cmake#L13), [2](https://github.com/dotnet/corefx/blob/61c48b4b3ad0bfa068997faba3ee2e60cdda9397/src/Native/Common/pal_config.h.in#L3) )

Another way would be to pinvoke to `uname(3)`, but [this](http://pubs.opengroup.org/onlinepubs/7908799/xsh/sysutsname.h.html) suggests that the order of the fields is implementation dependent, hence getting `sysname` won't be straight forward.

The reason for using `/usr/lib/libc.dylib` is the `.dylib` part. This file should be present only on OSX, AFAICS!

Feel free to correct me of course.

cdmihai:
@dotnet-bot test this please

rainersigwald:
:+1:

Issue #514, opened 2016-03-07
Cannot build msbuild on MAC


mfilippov:
I try build CI branch on my Mac. I run "sh cibuild.sh --target Mono --host Mono" and I got error:
```
Microsoft.Build.Framework -> /Users/mfilippov/Projects/msbuild/bin/Debug-MONO/Output/Microsoft.Build.Framework.dll
/Users/mfilippov/Projects/msbuild/targets/DeployDependencies.proj(80,5): error MSB4018: The "PrereleaseResolveNuGetPackageAssets" task failed unexpectedly.
/Users/mfilippov/Projects/msbuild/targets/DeployDependencies.proj(80,5): error MSB4018: This is an unhandled exception from a task -- PLEASE OPEN A BUG AGAINST THE TASK OWNER.
/Users/mfilippov/Projects/msbuild/targets/DeployDependencies.proj(80,5): error MSB4018: System.ArgumentNullException: Value cannot be null.
/Users/mfilippov/Projects/msbuild/targets/DeployDependencies.proj(80,5): error MSB4018: Parameter name: name
/Users/mfilippov/Projects/msbuild/targets/DeployDependencies.proj(80,5): error MSB4018:   at System.Resources.ResourceManager.GetString (System.String name, System.Globalization.CultureInfo culture) [0x0011a] in /private/tmp/source-mono-mac-4.2.0-branch-c6sr1/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.2/external/referencesource/mscorlib/system/resources/resourcemanager.cs:1299 
/Users/mfilippov/Projects/msbuild/targets/DeployDependencies.proj(80,5): error MSB4018:   at System.Resources.ResourceManager.GetString (System.String name) [0x00000] in /private/tmp/source-mono-mac-4.2.0-branch-c6sr1/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.2/external/referencesource/mscorlib/system/resources/resourcemanager.cs:1195 
[...]

PR #519, opened 2016-03-09
dir.props: Update mono build to target 4.6 instead of 4.5.1

radical:
This also fixes the build for mono/osx.

The build would break because mcs was getting duplicate references like:

    1. /reference:/Users/ankit/dev/misc/lib/mono/4.5-api/Facades/System.IO.dll
    2. /reference:/Users/ankit/.nuget/packages/System.IO/4.0.10/ref/dotnet/System.IO.dll

Ref(1) was resolved because of `ImplictlyExpandDesignTimeFacades`, and
Ref(2) came from the task `PrereleaseResolveNugetPackages`. This task
reads from the project.lock.json (generated by `dnu restore`), and for
4.5.1 case these assemblies were marked to have the assemblies from the
package be used. This is not the case for 4.6, where these assemblies
instead have entries like `ref/net45/_._`, effectively suggesting that
the assembly should be resolved from the GAC instead.

There are newer version of these package (like System.IO) available,
which seem to fix this, but it seems to be in flux.

Instead, we just target 4.6 completely for mono builds, which is the
same as the build for Windows and CoreCLR.

This also means that you cannot use mono < 4.4.x to build, since .net
4.6 API will become available only with mono 4.4.0 . But that is OK as
we cannot really get a useful build with the older mono anyway.

This does not affect CoreCLR builds that use mono host.

Issue: #514

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,6 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
     <!-- This is to workaround an issue with a dependent assembly (Microsoft.Build.Tasks.CodeAnalysis.dll)
          copying its dependencies (which are in the GAC). -->
     <DoNotCopyLocalIfInGac>true</DoNotCopyLocalIfInGac>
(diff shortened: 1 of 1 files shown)


dir.props
```diff
@@ -180,8 +179,8 @@
   </PropertyGroup>
 
   <!-- .NET Core build support -->
-  <PropertyGroup Condition="'$(NetCoreBuild)' != 'true'">
-    <TargetFrameworkVersion Condition="'$(TargetFrameworkVersion)' == ''">v4.5.1</TargetFrameworkVersion>
```
cdmihai: Can you please also set TargetFrameworkVersion for the windows full framework build (NetCoreBuild == false and MonoBuild == false) to 4.6.? Ideally they the mono build and windows full framework build should be as close as possible.

cdmihai:
LGTM with pending comment.

radical:
Updated so that for `$(NetCoreBuild) != 'true'` (effectively, mono or windows/full framework builds) use `TargetFrameworkVersion=v4.6` and the corresponding `NuGetTargetMoniker`.

PR #529, opened 2016-03-17
[RFC] Add support for fallback search paths for Target frameworks

radical:
Currently, target frameworks are looked up in the default location or a
path specified by $(TargetFrameworkRootPath). It would be useful to be
able to look for these frameworks in more than one location. For
example, in case of Xamarin products on OSX, mobile specific frameworks
are installed in
`/Library/Frameworks/Mono.framework/External/xbuild-frameworks` and the
.NET ones are installed in
`/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks`
.

With support for fallback search paths, a lookup for a target framework will follow the order:

    1. $(TargetFrameworkRootPath) or Default location is $(TargetFrameworkRootPath) is ''
    2. Fallback search paths

These fallback search paths can be set in app.config, per os, like:
```
<msbuildToolsets default="14.1">
  <toolset toolsVersion="14.1">
    <property name="TargetFrameworkRootPathSearchPathsOSX" value="/tmp/foo;/tmp/bar" />
    <property name="TargetFrameworkRootPathSearchPathsWindows" value="C:\tmp\foo;C:\tmp\bar" />
    ...
```

This is a RFC, and not complete as tests for this feature are yet to be added.

### DIFF
--- a/src/Shared/FrameworkLocationHelper.cs
+++ b/src/Shared/FrameworkLocationHelper.cs
@@ -967,55 +967,118 @@ namespace Microsoft.Build.Shared
             string toolPath = null;
 
 #if FEATURE_SYSTEM_CONFIGURATION
-            if (ToolsetConfigurationReaderHelpers.ConfigurationFileMayHaveToolsets())
+            try
             {
-                try
+                ToolsetElement toolset = GetToolsetElementFromConfigFor(toolsVersion);
+                PropertyElement toolsPathFromConfiguration = toolset?.PropertyElements.GetElement(MSBuildConstants.ToolsPath);
+                toolPath = toolsPathFromConfiguration?.Value;
+
+                if (toolPath != null)
                 {
-                    Configuration configuration = FileUtilities.RunningTests
-                                                      ? ConfigurationManager.OpenExeConfiguration(
-                                                          FileUtilities.CurrentExecutablePath)
-                                                      : ConfigurationManager.OpenExeConfiguration(
-                                                          ConfigurationUserLevel.None);
-
-                    ToolsetConfigurationSection configurationSection = ToolsetConfigurationReaderHelpers.ReadToolsetConfigurationSection(configuration);
-
-                    if (configurationSection != null)
+                    if (!FileUtilities.IsRootedNoThrow(toolPath))
                     {
-                        ToolsetElement toolset = configurationSection.Toolsets.GetElement(toolsVersion);
-
-                        if (toolset != null)
-                        {
-                            PropertyElement toolsPathFromConfiguration = toolset.PropertyElements.GetElement(MSBuildConstants.ToolsPath);
-
-                            if (toolsPathFromConfiguration != null)
-                            {
-                                toolPath = toolsPathFromConfiguration.Value;
-
-                                if (toolPath != null)
-                                {
-                                    if (!FileUtilities.IsRootedNoThrow(toolPath))
-                                    {
-                                        toolPath = FileUtilities.NormalizePath(Path.Combine(FileUtilities.CurrentExecutableDirectory, toolPath));
-                                    }
-
-                                    toolPath = FileUtilities.EnsureTrailingSlash(toolPath);
-                                }
-                            }
-                        }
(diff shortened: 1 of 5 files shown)


radical:
Updated to fix the merge issue.

cdmihai:
@dotnet-bot test this please

radical:
The code here needs to get access to the "current toolsversion" for the project, so that it can load the toolset from the config file and get access to the fallback search paths. But it currently uses `ToolLocationHelper.CurrentToolsVersion`, which is a constant and probably not what is required here. I'm not sure how I might be able to get the actual toolsversion for the project.

radical:
As suggested, out of band, I tried to add the fallback search paths in the targets file and use that. It solves the problem for that particular call. But the problem is that the target frameworks lookup is done in more than one place, for example in tasks like `GetReferenceAssemblyPaths`, `AssemblyIdentity` etc. New task properties will need to be surfaced in all of these for supporting the search paths. Anywhere we don't or can't do this, will have inconsistent behavior, namely that it won't be able to find the frameworks installed in the fallback paths.

Adding this support via the app.config, means that all the existing callers get this support, automatically, since it won't depend on the Any code using the `ToolLocationHelper.GetPathToStandardLibraries` and `ToolLocationHelper.GetPathToReferenceAssemblies` will have this support. So, from user POV, the target frameworks installed in the fallback paths would consistently be available (can be looked up) everywhere.

So, I think that we should go with the `app.config` approach. And sticking with:

```
  <toolset toolsVersion="14.1">
    <property name="TargetFrameworkRootPathSearchPathsOSX" value="/tmp/foo;/tmp/bar" />
    <property name="TargetFrameworkRootPathSearchPathsWindows" value="C:\tmp\foo;C:\tmp\bar" />
```

AndyGerlicher:
Discussed this a bit today. Still seems a bit odd to define it as a property in the toolset but then use it in such an usual way (not reading the property but reading the configuration setting). It also seems unnatural to include it with the other fallback search path since they're such different uses (used as fallback replacement in `<Import>` vs ToolLocationHelper method).

One idea we had for this is potentially add `os` to the `toolset` element. So for example:
```XML
<toolset toolsVersion="14.1" os="windows">
    <property name="TargetFrameworkRootPathSearchPaths" value="C:\tmp\foo;C:\tmp\bar" />
  ...
</toolset>
<toolset toolsVersion="14.1" os="osx">
    <property name="TargetFrameworkRootPathSearchPaths" value="/tmp/foo;/tmp/bar" />
  ...
/>
```
In this scenario, ToolLocationHelper should be able to read the property passed in and without having to add OS to the property name. Thoughts on that?

radical:
I agree, it does seem odd as a property or to be specified with the extensions paths search paths. But maybe then we should make it even clearer? Like:

```
<toolset toolsVersion="14.1">
   <targetFrameworkFallbackSearchPaths>
       <searchPaths os="windows" paths="c:\tmp\foo;c:\tmp\bar" />
       <searchPaths os="osx" paths="/tmp/foo;/tmp/bar" />
   </targetFrameworkFallbackSearchPaths>
   ...
```

(with better names in the `<searchPaths ..` element).
- This removes the confusion that it could be a property or something that can be set or overridden anywhere else.
- It doesn't use the `TargetFrameworkRootPath` name, so again no confusion with that property.
- The name `targetFrameworkFallbackSearchPaths` suggests what it is supposed to do too

One issue might be that this now sounds too similar to `<msbuildExtensionsPathSearchPaths>` thus maybe suggesting (to the user/reader) that they are supposed to have the same behavior. For that, maybe we can rename that to `<msbuildExtensionsPathSearchPathsForImports>`. I think, it would suggest that it is meant for a specific case "ForImports", at which point the user can look at the documentation to find out more.

radical:
Any thoughts on this?

AndyGerlicher:
Can this be closed in favor of #627?

radical:
Err.. sorry for the late response! This still needs to be open as it adds the support for *search paths* for target frameworks. #627 was extracted from commits in here, so, I will update this PR to remove those and so it will be left with the commit for search path support. I could open a new PR, but since discussion about this particular feature already happened here, it might be good to use this PR itself.

Issue #531, opened 2016-03-18
[xplat] Mono build broken - FEATURE_PIPE_SECURITY related?


radical:
xplat build is broken. I disabled `FEATURE_PIPE_SECURITY` in dir.props . But then I get this:

```
BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs(343,42): error CS0122: `System.IO.Pipes.PipeSecurity.PipeSecurity(System.Runtime.InteropServices.SafeHandle, System.Security.AccessControl.AccessControlSections)' is inaccessible due to its protection level [/Users/ankit/dev/msbuild/src/XMakeBuildEngine/Microsoft.Build.csproj]
BackEnd/Node/NodeConfiguration.cs(146,24): error CS1061: Type `Microsoft.Build.BackEnd.INodePacketTranslator' does not contain a definition for `TranslateDotNet' and no extension method `TranslateDotNet' of type `Microsoft.Build.BackEnd.INodePacketTranslator' could be found. Are you missing an assembly reference? [/Users/ankit/dev/msbuild/src/XMakeBuildEngine/Mi
crosoft.Build.csproj]
```

.. and the relevant code ...

```
 340 #if FEATURE_PIPE_SECURITY
 341                     PipeSecurity remoteSecurity = nodeStream.GetAccessControl();
 342 #else
 343                     var remoteSecurity = new PipeSecurity(nodeStream.SafePipeHandle, System.Security.AccessControl.AccessControlSections.Access |
 344                         System.Security.AccessControl.AccessControlSections.Owner | System.Security.AccessControl.AccessControlSections.Group);
 345 #endif
```

radical:
Fixed in #533 .

PR #533, opened 2016-03-18
Fix mono/OSX build

radical:
- Disable FEATURE_PIPE_SECURITY for mono builds
- #if out a PipeSecurity use
- s,FEATURE_APPDOMAIN,FEATURE_BINARY_SERIALIZATION,
    for using INodePacketTranslator.TranslateDotNet

Issue #531

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Node/NodeConfiguration.cs
+++ b/src/XMakeBuildEngine/BackEnd/Node/NodeConfiguration.cs
@@ -142,7 +142,7 @@ namespace Microsoft.Build.BackEnd
             translator.Translate(ref _nodeId);
             translator.Translate(ref _buildParameters, BuildParameters.FactoryForDeserialization);
             translator.TranslateArray(ref _forwardingLoggers, LoggerDescription.FactoryForTranslation);
-#if FEATURE_APPDOMAIN
+#if FEATURE_BINARY_SERIALIZATION && FEATURE_APPDOMAIN
             translator.TranslateDotNet(ref _appDomainSetup);
 #endif
         }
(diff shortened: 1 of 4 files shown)


src/XMakeBuildEngine/BackEnd/Node/NodeConfiguration.cs
```diff
@@ -142,7 +142,7 @@ public void Translate(INodePacketTranslator translator)
             translator.Translate(ref _nodeId);
             translator.Translate(ref _buildParameters, BuildParameters.FactoryForDeserialization);
             translator.TranslateArray(ref _forwardingLoggers, LoggerDescription.FactoryForTranslation);
-#if FEATURE_APPDOMAIN
+#if FEATURE_BINARY_SERIALIZATION
```
dsplaisted: What's the reason for this change?  You shouldn't be serializing AppDomainSetup unless AppDomain's are available.  Combining both also seems reasonable:

    #if FEATURE_APPDOMAIN && FEATURE_BINARY_SERIALIZATION

radical: I based this solely on the availability of `INodePacketTranslator.TranslateDotNet`, which is defined for `FEATURE_BINARY_SERIALIZATION` .

radical: And you are correct, _appDomainSetup is available only for FEATURE_APPDOMAIN, so, it should be a combination.

radical:
I had committed a similar "fix" for `INodePacketTranslator.TranslateDotNet` use earlier, fixing that again in #535 .

dsplaisted:
LGTM, though I would just include the one-line change from #535 in this PR instead of keeping it separate.

radical:
Updated.

AndyGerlicher:
LGTM

PR #534, opened 2016-03-18
NativeMethodsShared.OSName: Fix the case of OSX

radical:
OSX is also a Unix, so check for the less generic kind first. So, change

    return IsUnix ? "Unix" : (IsOSX ? "OSX" : "Windows_NT")
to
    return IsOSX ? "OSX" : (IsUnix ? "Unix" : "Windows_NT")


AndyGerlicher:
LGTM

rainersigwald:
LGTM

PR #535, opened 2016-03-18
TaskHostConfiguration: Don't use _appDomainSetup unless FEATURE_APPDO…

radical:
…MAIN is available.

This actually fixes a previous "fix" (by yours truly),
9300223aaa2614bf5aa89b3dfd28da1fde1102f4, which changed the #if guard
from FEATURE_APPDOMAIN to FEATURE_BINARY_SERIALIZATION. But this code
requires both:

    translator.TranslateDotNet(ref _appDomainSetup);


radical:
Merged as part of #533 .

PR #536, opened 2016-03-21
Adding custom UpdateBuildValues target

AndyGerlicher:
- This is in BuildTools > 206, which causes many build issues. We will
   need to revert this change when we're able to take the new package.
 - I need this change because the GitPushRemote variable was added.


rainersigwald:
:+1:

PR #537, opened 2016-03-21
[Do not merge] Mac / Linux debugging support

dsplaisted:
These changes enable debugging MSBuild on Mac/Linux with [VS Code](http://code.visualstudio.com/).  However, until updated tooling for Visual Studio 2015 and the .NET CLI is released, they break the experience in Visual Studio (no references resolve), so this PR probably shouldn't be merged until that tooling is available.

Specifically, this PR:

- Updates to use recent builds of .NET Core, which are required to support [debugging .NET code in VS Code](https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger.md)
- Updates to use a newer version of the [.NET CLI](https://github.com/dotnet/cli), which is required to support the updated package versions
- Updates the bootstrapping scripts to download a new version of the .NET CLI if the version specified in DotnetCLIVersion.txt has changed
- Standardizes capitalization of "Debug-NetCore" configuration
- Enable portable PDB creation when targeting .NET Core
- Applies workarounds to filenames in deployment directory (capitalize .Targets extension for Microsoft.CSharp.Targets, copy corerun to corerun.exe).  The BuildTools package already does this in the init script when creating the Tools directory, this change mirrors this with the deployment directory.
- Fix Copy task so that on non-Windows, source/target path comparison is case sensitive.
- Support waiting for debugger to attach when an InternalErrorException is thrown

With these changes applied, you can use the following instructions to set up VS Code to debug MSBuild on .NET Core: [Instructions for setting up the .NET Core debugger](https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger.md)

Here is a sample configuration which you can put in VS Code's launch.json file to debug MSBuild building itself on Linux:

```json5
        {
            "name": "self-build",
            "type": "coreclr",
            "request": "launch",
            //"preLaunchTask": "build",
            "program": "${workspaceRoot}/bin/Debug-NetCore/Unix_Deployment/corerun",
            "args": [
                "${workspaceRoot}/bin/Debug-NetCore/Unix_Deployment/MSBuild.exe",
                "${workspaceRoot}/build.proj",
                "/p:Configuration=Debug-NetCore",
                "/p:BinDir=bin/bin2/"
            ],
            "cwd": "${workspaceRoot}",
            "stopAtEntry": false
        },
```


cdmihai:
@dotnet-bot test this please

cdmihai:
Re-opened in #552

PR #538, opened 2016-03-22
$(OS) should return only two values - `Windows_NT` and `Unix` .

radical:
Based on the discussion in issue #539, $(OS) property will return
`Windows_NT` whenever running on windows and `Unix` for any other OS.

NativeMethodsShared.OSName is used to set that property, hence this is
being changed.

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -535,7 +535,7 @@ namespace Microsoft.Build.Shared
         {
             get
             {
-                return IsOSX ? "OSX" : (IsUnix ? "Unix" : "Windows_NT");
+                return IsWindows ? "Windows_NT" : "Unix";
             }
         }
 


cdmihai:
@dotnet-bot test this please

cdmihai:
Should we revert this only for Mono?

radical:
No, IMO, we should have uniform behavior w.r.t this. Having different behavior will likely cause unnecessary pain. And it doesn't affect compatibility with existing projects.

If the proposed approach in #539 sounds acceptable, then I'll change this PR accordingly. The proposed approach being:

1. `$(OS)` will have values of `Windows_NT` and `Unix` only. On Windows - `$(OS)==Windows_NT` and only any `IsUnixLike` platform - `$(OS)==Unix`.

2. Introduce a new property `$(OSName)` which would have more specific values like `Windows_NT`, `OSX` and `Unix`. I'm not sure whether `Unix` is enough here or it should be more specific like `Linux`, `FreeBSD` etc.

If the user's requirement is to simply check for "windows or unix-like", then `$(OS)` can be used. If the requirement is to have more specific behavior, then they can use `$(OSName)`.

@akoeplinger @jonpryor ^

If this is acceptable, then I can update the PR to *not* revert the patch, hence retain `NativeMethodsShared.OSName`, set the new property with this and set the `$(OS)` property explicitly.

cdmihai:
@joperezr Does this affect corefx in any way?

joperezr:
> @joperezr Does this affect corefx in any way? Please also take a look at #539

In corefx [we currently do the 'hack' of setting OSX to Unix](https://github.com/dotnet/corefx/blob/master/dir.props#L10) because we were seeing issues for this so I don't think that this would cause issues for now, but I'm not sure until I test it.

AndyGerlicher:
Based on discussion in #539 this should change to only return `Windows_NT` when Windows, else `Unix`. You want to update this PR?

radical:
Sure, I'll do that.

AndyGerlicher:
LGTM

cdmihai:
@dotnet-bot test this please

PR #542, opened 2016-03-25
Fix ToolName for AL and LC - al.exe and lc.exe respectively

radical:
- the name and case matches what mono uses and should be fine for
  windows

### DIFF
--- a/src/XMakeTasks/Al.cs
+++ b/src/XMakeTasks/Al.cs
@@ -292,7 +292,7 @@ namespace Microsoft.Build.Tasks
         {
             get
             {
-                return NativeMethodsShared.IsWindows ? "AL.exe" : "al";
+                return "al.exe";
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/XMakeTasks/Al.cs
```diff
@@ -292,7 +292,7 @@ override protected string ToolName
         {
             get
             {
-                return NativeMethodsShared.IsWindows ? "AL.exe" : "al";
+                return "al.exe";
```
akoeplinger: I think we rely on `al` (i.e. the wrapper script that then calls into al.exe) on Mono/xbuild. Is that not the case here?

radical: No, we use `al.exe` and since mono's `Process.CreateProcess` can handle that, it works fine.

AndyGerlicher:
LGTM

PR #543, opened 2016-03-25
Tasks: Fix overriding `ToolExe` from project files.

radical:
`ToolExe` can be overriden from project files. If not set, then it
falls back to `ToolName`. So, use `ToolExe` in `GenerateFullPathToTool`
instead of `ToolName`.


AndyGerlicher:
LGTM

Issue #502, opened 2016-02-22
Improve understandability of RAR name mismatch


rainersigwald:
When ResolveAssemblyReferences rejects a file because the names don't match, it isn't always super clear about why.

```
Primary reference "System.Runtime". (TaskId:24)
...
  Considered "d:\CxCache-Scope\DotNetFxRefAssemblies.Corext.4.5.2\Facades\System.Runtime.dll", but its name "System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" didn't match. (TaskId:24)
```

It would be nice to have some sort of clearer error, like "version doesn't match", or even just an aligned set of strings for "expected" and "actual", maybe with markers on another line indicating where the diff begins.

[This is a suggestion from an Microsoft dev on an internal alias.]

rainersigwald:
Maybe something like:

```
Considered "d:\CxCache-Scope\DotNetFxRefAssemblies.Corext.4.5.2\Facades\System.Runtime.dll",
  but its name "System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" didn't match the expected value
               "System.Runtime"
```

PR #550, opened 2016-04-03
RAR name mismatch message with Assembly name

nikson:
Update error message with expected, actual and target `Assembly` name

    Considered "....expected....",
        but its name "....actual....".
    Didn't match the expected version  "..assembly name."

Issue #502

### DIFF
--- a/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs
+++ b/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs
@@ -1193,7 +1193,7 @@ namespace Microsoft.Build.Tasks
 
             // If there is a list of assemblyFiles that was considered but then rejected,
             // show information about them.
-            LogAssembliesConsideredAndRejected(reference, importance);
+            LogAssembliesConsideredAndRejected(reference, fusionName, importance);
 
             if (!reference.IsBadImage)
             {
@@ -1549,7 +1549,7 @@ namespace Microsoft.Build.Tasks
         /// </summary>
         /// <param name="reference">The reference.</param>
         /// <param name="importance">The importance of the message.</param>
-        private void LogAssembliesConsideredAndRejected(Reference reference, MessageImportance importance)
+        private void LogAssembliesConsideredAndRejected(Reference reference, string fusionName, MessageImportance importance)
         {
             if (reference.AssembliesConsideredAndRejected != null)
             {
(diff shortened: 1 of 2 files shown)


src/XMakeTasks/Strings.resx
```diff
@@ -1341,8 +1341,10 @@
                 {0} is an assembly name with no version number like 'D, Culture=neutral, PublicKeyToken=aaaaaaaaaaaaaaaa'
     </comment>
   </data>
-  <data name="ResolveAssemblyReference.ConsideredAndRejectedBecauseFusionNamesDidntMatch">
-    <value>Considered "{0}", but its name "{1}" didn't match.</value>
+  <data name="ResolveAssemblyReference.ConsideredAndRejectedBecauseFusionNamesDidntMatch" xml:space="preserve">
+    <value>Considered "{0}",
+		but its name "{1}".
+	Didn't match the expected version "{2}"</value>
```
rainersigwald: Minor rephrase:
```xml
<value>Considered "{0}",
		but its name "{1}"
		didn't match the expected name "{2}".</value>
```

rainersigwald:
Looks like a good change to me; I suggested a minor edit for sentence formation.

src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -1602,7 +1602,7 @@ private void LogAssembliesConsideredAndRejected(Reference reference, MessageImpo
                                 break;
                             }
                         case NoMatchReason.FusionNamesDidNotMatch:
-                            Log.LogMessageFromResources(importance, "ResolveAssemblyReference.EightSpaceIndent", Log.FormatResourceString("ResolveAssemblyReference.ConsideredAndRejectedBecauseFusionNamesDidntMatch", location.FileNameAttempted, location.AssemblyName.FullName));
+                            Log.LogMessageFromResources(importance, "ResolveAssemblyReference.EightSpaceIndent", Log.FormatResourceString("ResolveAssemblyReference.ConsideredAndRejectedBecauseFusionNamesDidntMatch", location.FileNameAttempted, location.AssemblyName.FullName, location.AssemblyName.Name));
```
radical: The new argument, expected name, should probably be `fusionName` available in the caller (`LogReference`). I couldn't see how that might be available via the `reference` argument itself though.

rainersigwald: That's a good point. @nikson, can you include sample output for a mismatch here? That'll help figure out if it's emitting the right thing.

radical: Call chain: `LogResults` -> `LogReference` -> `LogAssembliesConsideredAndRejected`

Looking at the call sites of LogReference, [1](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs#L918) and [2](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs#L930), fusionName is being set to `assemblyName.FullName`.

IMHO, irrespective of what `fusionName` is being set to, up the call stack, `LogAssembliesConsideredAndRejected` should use the `fusionName` that it's caller `LogReference` was passed.

nikson:
Added `fusionName` from `LogReference` and sample output like following  
```			
Considered "C:\Paul\Assemblies\System.Runtime.dll",
	but its name "System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"
	didn't match the expected name "System.Runtime, Version=2.0.0.0".
```  
There is a `EightSpaceIndent` before the message. Trying to write a test case for this but couldn't generate `FusionNamesDidNotMatch`, stuck somewhere before that.

cdmihai:
@dotnet-bot test this please

AndyGerlicher:
Sorry it took so long to make it in. Thanks for your contribution!

PR #551, opened 2016-04-04
Version bump 15.0 -> 15.1

AndyGerlicher:
- We decided to go with 15.1 rather than confuse with the 15.0 version
   that's in VS "15" preview. 15.1 is intended to build/ship from GitHub
   sources (rather than internal source 'mirror') and not be installed or
   put in the GAC.

It was much easier the second time at least! :smile:


src/XMakeCommandLine/app.config
```diff
@@ -13,29 +13,29 @@
...
-          <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.0.0.0" />
+          <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.1.0.0" />
         </dependentAssembly>
       </assemblyBinding>
     </runtime>
     <!-- To define one or more new toolsets, add an 'msbuildToolsets' element in this file. -->
-    <msbuildToolsets default="15.0">
-      <toolset toolsVersion="15.0">
+    <msbuildToolsets default="15.1">
+      <toolset toolsVersion="15.1">
```
rainersigwald: Should we leave the toolset version at 15.0? It's probably a bit more natural of a progression for end users in VS . . .

rainersigwald:
LGTM

src/XMakeCommandLine/app.config
```diff
@@ -13,29 +13,29 @@
...
         <property name="MSBuildToolsPath" value="." />
       </toolset>
       <toolset toolsVersion="12.0">
```
rainersigwald: We should add a fallback for `14.0`, right?

PR #552, opened 2016-04-05
Linux debugging, packaging numbers, and enabling multiproc

cdmihai:
Builds on top of  #537 

In addition:
- update the public keys for the test assemblies
- update msbuils package numbers in response to new buildtools format
- enable /m for the msbuild build


cdmihai:
@dsplaisted

build.cmd
```diff
@@ -1,7 +1,7 @@
 @echo off
 setlocal
 
-set MSBUILD_ARGS="%~dp0build.proj" /verbosity:minimal /fileloggerparameters:Verbosity=diag;LogFile="%~dp0msbuild.log" %*
```
rainersigwald: Why remove the minimal console verbosity?

dir.traversal.targets
```diff
@@ -16,10 +16,8 @@
   </Target>
 
   <Target Name="TestAllProjects">
-    <!-- To Serialize we use msbuild's batching functionality '%' to force it to batch all similar projects with the same identity
-         however since the project names are unique it will essentially force each to run in its own batch -->
-    <MSBuild Targets="Test" Projects="@(Project)" Condition="'$(SerializeProjects)'=='true'" Properties="Dummy=%(Identity)"/>
-    <MSBuild Targets="Test" Projects="@(Project)" Condition="'$(SerializeProjects)'!='true'" BuildInParallel="true" />
+    <!-- Tests are always sequential -->
+    <MSBuild Targets="Test" Projects="@(Project)" Properties="Dummy=%(Identity)"/>
```
rainersigwald: Probably better to just set BuildInParallel="false".

NuGet.Config
```diff
@@ -3,13 +3,14 @@
   <packageRestore>
     <add key="enabled" value="True" />
   </packageRestore>
+  <!-- the commented  nuget feeds had failing mirrors (only non windows). Temporarily disabling them -->
```
rainersigwald: Be explicit about when to undisable please.

src/dir.targets
```diff
@@ -20,6 +20,9 @@
           Condition="'$(NetCoreBuild)' == 'true' 
                       and Exists('$(CSharpPortableTargetsFallback)')
                       and !Exists('$(CSharpPortableTargets)')"/>
+  
+  <Import Project="$(RepoRoot)\targets\OverrideGetReferenceAssemblyPaths.targets"
+          Condition="'$(NetCoreBuild)' == 'true'" />
```
rainersigwald: Trailing whitespace, boo, hiss!

rainersigwald:
LGTM. Love having all 3 builds passing!

build.cmd
cdmihai: Because I conflicted with myself during rebase and I solved the conflict wrong :)

rainersigwald: A tale as old as time.

NuGet.Config
cdmihai: @rainersigwald 

That is a very good question. The weird thing is that restoring from those feeds only fails on non windows. That's why our xplat CI is broken on osx right now, but passes on windows.

If I remove those offending feeds, all the packages get "magically" downloaded from some other feeds, which is a happy coincidence. I would say to leave them commented for a few months, and if the happy coincidence does not burst into an unhappy failed restore, completely remove them.

AndyGerlicher:
LGTM

PR #560, opened 2016-04-08
Fix OSX/Mono build

radical:


src/Utilities/UnitTests/project.json
```diff
@@ -1,6 +1,7 @@
 ﻿{
   "dependencies": {
     "xunit": "2.1.0",
+    "xunit.netcore.extensions": "1.0.0-prerelease-00116",
```
rainersigwald: Should this be here or in netstandard1.3? I would have guessed below.

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -317,7 +317,7 @@ private string GetProcessesToIgnoreKey(long hostHandshake, long clientHandshake,
         private void ValidateRemotePipeSecurityOnWindows(NamedPipeClientStream nodeStream)
         {
             SecurityIdentifier identifier = WindowsIdentity.GetCurrent().Owner;
-#if FEATURE_PIPE_SECURITY
+#if !FEATURE_PIPE_SECURITY
```
rainersigwald: Man that's confusing. There's `PipeSecurity` on both sides! This seems reasonable based on your commit notes though. @dsplaisted objections?

rainersigwald:
LGTM

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
radical: I agree, very confusing! Maybe we should rename that to `FEATURE_USE_LOCAL_COPY_PIPE_SECURITY` or something better?

src/Utilities/UnitTests/project.json
radical: Is the `netstandard1.3` for .netcore builds? AFAICS, only `xunit.netcore.extensions` seems to have the `PlatformSpecific` attribute that we need here, even for !netcore builds . I could be wrong here!

rainersigwald: Ahh . . . that makes sense. It's in the right place, then, since it's needed "everywhere".

`netstandard1.3` is a [platform standard](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md) that implies running on .NET Core, full .NET Framework 4.6+, and Mono. We're primarily using it to indicate targeting .NET Core, though (hence my original comment).

---
In reply to: [59046481](https://github.com/Microsoft/msbuild/pull/560#discussion_r59046481) [](ancestors = 59046481)

radical: I could add this explanation to the commit message, for future readers!

dsplaisted: I heard a rumor that in the future, you might actually be able to add comments *directly in the project.json file*. :astonished:

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
dsplaisted: This code isn't currently used on .NET Core.  Initially I ported the /m support using named pipes, and as part of that added the local copy of PipeSecurity and factored this code out into a separate method.  Subsequently, I switched to using anonymous pipes in .NET Core.  However, if we want to add support for node reuse to .NET Core at some point, then we would probably need to go back to using named pipes.

In any case, I am still confused and I think this may be the wrong change.  `FEATURE_PIPE_SECURITY` was supposed to mean "the Framework includes PipeSecurity", ie the opposite of what `FEATURE_USE_LOCAL_COPY_PIPE_SECURITY` would be.  When PipeSecurity is included in the framework, then you can call GetAccessControl on the stream.  If it's not included, then you have to call the constructor for PipeSecurity (and it will be the local copy of the code).

What was the problem with the Mono build?  Does Mono include PipeSecurity?

radical: The actual build error is:
```
BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs(323,34): error CS0122: `System.IO.Pipes.PipeSecurity.PipeSecurity(System.Runtime.InteropServices.SafeHandle, System.Security.AccessControl.AccessControlSections)' is inaccessible due to its protection level [/Users/ankit/dev/msbuild/src/XMakeBuildEngine
/Microsoft.Build.csproj]
```

Yeah, we do seem to have the classes. `FEATURE_PIPE_SECURITY` is disabled for Mono builds in `dir.props`. If I enable that, then I hit:

```
../Shared/NodeEndpointOutOfProcBase.cs(209,94): error CS0103: The name `AccessControlType' does not exist in the current context [/Users/ankit/dev/msbuild/src/XMakeBuildEngine/Microsoft.Build.csproj]
```

.. which seems to be because `using System.Security.AccessControl` is guarded by `#if FEATURE_SECURITY_PERMISSIONS` . Changing that to `#if FEATURE_SECURITY_PERMISSIONS || FEATURE_PIPE_SECURITY` (maybe only `FEATURE_PIPE_SECURITY` required here?), the build succeeds with all the tests passing.

AndyGerlicher:
LGTM, thanks for fixing that up! Now we need to get a CI build added for Mono.

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
radical: If the above looks correct, then I'll update the PR.

radical: Um this bit is still left! I can open a fresh PR if the changes in https://github.com/Microsoft/msbuild/pull/560#discussion_r59053675 look correct to @dsplaisted .

dsplaisted: @radical Yes, that sounds good.

PR #561, opened 2016-04-08
Fix windows/mono build: Samples/TaskUsageLogger - Target framework v4.6,

radical:
.. same as the main build.


rainersigwald:
LGTM

cdmihai:
I did a quick scan and the tasks unit tests appears to target 4.0 (https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeTasks/XMakeTasksUnitTests/XMakeTasksUnitTests.csproj#L12). Is there a particular reason why that's so, or can we bump that up to 4.6 as well?

Otherwise, LGTM

AndyGerlicher:
That file isn't building. If you look at the bottom `<Import Project="$(_NTDRIVE)$(_NTROOT)\tools\x86\managed\v4.0\Microsoft.CSharp.targets" />`, there's no way that works outside our internal build so it's never been built here.

PR #562, opened 2016-04-08
Fix regression in use of PipeSecurity

radical:
- commit 583a73e21ac36093e285d473eefec701d9f60d5f was incorrect,
  see https://github.com/Microsoft/msbuild/pull/560#discussion_r59051335

  Essentially, FEATURE_PIPE_SECURITY means that `PipeSecurity` is
  available in the framework. Revert the earlier fix.

- this patch also enables FEATURE_PIPE_SECURITY for mono builds

- Fixes windows build broken by the regression


cdmihai:
Hmm, we should have a test about this, since it flew under the CI :)
@dsplaisted How would a minimal test look like? Build with /m and grep for "on node [^1]"?

radical:
This would have shown up as a windows/mono build breakage. Though that passing would not mean that the compiled code was now correct.

cdmihai:
True, nothing asserts for correctness. We can only reasonably assert for constraining the behaviour. :(
We need to add more CI jobs :)

rainersigwald:
LGTM

Issue #539, opened 2016-03-22
[RFC] Detecting current OS


radical:
MSBuild sets the `$(OS)` property to OSX/Unix/Windows at runtime, which can be used in targets to determine the current OS. But `xbuild` has been using `$(OS)==Unix` on OSX and Linux, because of an old mono bug. And it continues to do so for compatibility reasons.

It makes sense for MSBuild to follow the same, since users would want to use it with existing targets. PR #538 reverts back to that behavior.  So, for non-windows case, `$(OS)` property is effectively acting as a `IsUnixLike`. Considering it like that, `$(OS)` should probably have only two values now - `Windows_NT` and `Unix`.

Now, I think it might be useful to be able to differentiate between `OSX` and other specific unices. But maybe we should use some other (new) property for that? `$(OSName)` ? I am not sure how the CoreCLR case is affected by this, but AFAIU, it would be in the same boat.

Thoughts?

akoeplinger:
We just discussed this in Xamarin and we came to the conclusion that MSBuild changing `$(OS)` to OSX is too much of a breaking change for us, so when we adopt MSBuild we need to quirk this to return `Unix` again to prevent breaking existing code.

@rainersigwald you said in https://github.com/Microsoft/msbuild/issues/446#issuecomment-175205322 that you think it's reasonable to update conditions when moving to MSBuild, but is there a chance we can revisit this decision?

jonpryor:
For bad added context, there are [**140** matches on github for `.targets` files which contain both `OS` and `Unix`](https://github.com/search?utf8=✓&q=Condition+OS+Unix+extension%3Atargets&type=Code&ref=searchresults) (my possibly futile effort at capturing constructs similar to `Condition="'$(OS)' == 'Unix'"`).

There are [**1,653** matches](https://github.com/search?q=Condition+OS+Unix+extension%3Acsproj&ref=searchresults&type=Code&utf8=✓) when targeting `.csproj` files.

This isn't necessarily "a lot" -- or even accurate? -- but it does show that plenty of existing code files use `$(OS)` in one form or another, and adding a new value has the potential to *break* all of these existing files.

Changing `$(OS)` to be `OSX` instead of `Unix` is at minimum a possible compatibility break, and I believe shows that the "abstraction" isn't abstract enough. Users would likely be better served with an alternate mechanism that was more conducive to supporting new platforms without requiring auditing all existing use of e.g. `$(OS)`, though I'm not sure what such an abstraction would resemble.

[...]

PR #568, opened 2016-04-11
Add intrinsic property function `IsOSPlatform(string os)`

radical:
Usage:
    $([MSBuild]::IsOSPlatform(Windows))

Based on @dsplaisted's suggestion in issue #539 .

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
+++ b/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
@@ -364,6 +364,17 @@ namespace Microsoft.Build.Evaluation
             return false;
         }
 
+        internal static bool IsOSPlatform(string os)
+        {
+            switch (os.ToLower())
+            {
+                case "windows": return NativeMethodsShared.IsWindows;
+                case "osx": return NativeMethodsShared.IsOSX;
+                case "unix": return NativeMethodsShared.IsUnixLike;
+                default: return false;
+            }
+        }
+
         #region Debug only intrinsics
 
         /// <summary>
--- a/src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
@@ -2826,6 +2826,36 @@ namespace Microsoft.Build.UnitTests.Evaluation
             Assert.Equal((~-43).ToString(), result);
         }
 
+        private void CheckOSPlatform(Func<string, bool> isOSPlatform, bool isWindows, bool isOSX, bool isUnix)
+        {
+            Assert.Equal(isWindows, isOSPlatform("Windows"));
+            Assert.Equal(isOSX, isOSPlatform("OSX"));
+            Assert.Equal(isUnix, isOSPlatform("Unix"));
+        }
+
+        [Fact]
+        public void PropertyFunctionStaticMethodIntrinsicIsOSPlatform()
+        {
+            PropertyDictionary<ProjectPropertyInstance> pg = new PropertyDictionary<ProjectPropertyInstance>();
+            Expander<ProjectPropertyInstance, ProjectItemInstance> expander = new Expander<ProjectPropertyInstance, ProjectItemInstance>(pg);
+
+            Func<string, bool> isOSPlatform = os => bool.Parse(expander.ExpandIntoStringLeaveEscaped(@"$([MSBuild]::IsOSPlatform(" + os + "))",
+                                                                 ExpanderOptions.ExpandProperties, MockElementLocation.Instance));
+
+            if (NativeMethodsShared.IsWindows)
+            {
+                CheckOSPlatform(isOSPlatform, isWindows:true, isOSX:false, isUnix:false);
+            }
+            else if (NativeMethodsShared.IsOSX)
+            {
+                CheckOSPlatform(isOSPlatform, isWindows:false, isOSX:true, isUnix:true);
+            }
+            else
+            {
+                CheckOSPlatform(isOSPlatform, isWindows:false, isOSX:false, isUnix:true);
+            }
+        }
+
         /// <summary>
         /// Expand a property reference that has whitespace around the property name (should result in empty)
         /// </summary>


src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
```diff
@@ -364,6 +364,17 @@ internal static bool DoesTaskHostExist(string runtime, string architecture)
...
         }
 
+        internal static bool IsOSPlatform(string os)
+        {
+            switch (os.ToLower())
+            {
+                case "windows": return NativeMethodsShared.IsWindows;
+                case "osx": return NativeMethodsShared.IsOSX;
+                case "unix": return NativeMethodsShared.IsUnixLike;
+                default: return false;
```
rainersigwald: I think this makes sense for full-framework and Mono builds, since the CoreFX method doesn't exist there. When it does, though, should this be a passthrough to `RuntimeInformation.IsOSPlatform(OSPlatform.Create(os))`?

radical: Sure, but then we need to ensure that they support the same "set".

Looking at [this](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.InteropServices.RuntimeInformation/src/System/Runtime/InteropServices/RuntimeInformation/OSPlatform.cs) and [this](https://github.com/dotnet/corefx/blob/master/src/Native/configure.cmake), the name depends on `PAL_UNIX_NAME` set in the `configure.cmake` using `CMAKE_SYSTEM_NAME`.

[This](https://cmake.org/cmake/help/v3.4/variable/CMAKE_SYSTEM_NAME.html) and [this](https://cmake.org/cmake/help/v3.4/variable/CMAKE_HOST_SYSTEM_NAME.html#variable:CMAKE_HOST_SYSTEM_NAME) suggest that it uses `uname -s`, when available, for this. So, we could use `uname -s` too on non-windows systems. `Darwin`->`OSX` being the special case.

And maybe add a `Unix` too (== `IsUnixLike`) ?

jonpryor:
Since you're adding an intrinsic, could we expand the scope a bit. :-)

There are many "Unix" or Unix-like platforms, e.g. Linux, AIX, FreeBSD, OpenBSD, NetBSD, Solaris... -- which all have their distinct features. Additionally, various versions of Windows also have their own "quirks" and features.

Aside, consider the [**uname**(1)](http://linux.die.net/man/1/uname) program, which allows determine the kernel name (e.g. "Darwin", "Linux"), version number, hardware architecture, and more.

What might be handy is if the `IsOSPlatform()` intrinsic could act as a *query* mechanism for things OS-related. This query could check for everything `uname` provides, with Windows equivalents, for example:

    $([MSBuild]::IsOSPlatform(Unix{Name=Darwin, Machine=x86_64}))

would return True when running on x86_64 OS X and False elsewhere.

AndyGerlicher:
@jonpryor We discussed this today and while I don't disagree it could be helpful it's something the Framework should provide and we want this to pass through and support only what the Framework returns. In the case of full Framework or Mono where `RuntimeInformation.IsOSPlatform` doesn't exist yet it should just emulate it until it does. Definitely don't want to start our own, new, moniker for this if we can help it. So for Net Core we'll want the code for this to be:
 ``` C#
return RuntimeInformation.IsOSPlatform(OSPlatform.Create(os))
```

Make sense? Do you think there will be cases where extra information would be important to the build? Generally it seems to be test for Windows or not Windows.

jonpryor:
> Do you think there will be cases where extra information would be important to the build? 

I have my own possible purposes for such a thing. :-)

In particular, *native libraries*. Visual Studio will generate MSBuild-compatible `.vcxproj` files with `@(ClCompile)` items. (Xamarin Studio, meanwhile, generates `.cproj` files with `@(Compile)` items.)

I would like to be able to use `.vcxproj` files/semantics to build shared libraries on Unix, hacking things up as necessary, and I can imagine needing to make certain properties conditional on the executing operating system, e.g. if on Unix set the compiler to use clang instead of CL.EXE.

This could presumably be handled with `RuntimeInformation.IsOSPlatform()`.

Where things get potentially "wonky" is dealing with OS differences. Not all Unixes are alike (nor are all Windows versions...), so I can imagine that being able to query OS name/version/architecture/etc. would allow default-setting properties/items/etc. so that things can more sanely "Just Work" without needing to explicitly specify configurations or override properties.

Aside: Where does the `@(ClCompile)` and related logic live? Will MSBuild be able to handle `.vcxproj` files in a cross-platform manner?

Clear as mud? Or a  reason to rethink this my fundamental brain damage in the first place? :-)

jonpryor:
> I have my own possible purposes for such a thing. :-)

...and now that we've open-sourced some things, I can provide pointers. :-)

In the absence of a "useful" `IsOSPlatform()`, I've [hacked my own `$(HostOS)` property][hostos]:

    <HostOS Condition=" '$(HostOS)' == '' And '$(OS)' == 'Windows_NT' ">Windows</HostOS>
    <HostOS Condition=" '$(HostOS)' == '' And '$(OS)' == 'Unix' And Exists ('/Applications') ">Darwin</HostOS>
    <HostOS Condition=" '$(HostOS)' == '' And '$(OS)' == 'Unix' ">Linux</HostOS>

[hostos]: https://github.com/xamarin/xamarin-android/blob/1b3a76c6874853049e89bbc113b22bc632ed5ca4/Configuration.props#L8-L10

This *works*, but is kinda ugly in that I need to use a filesystem path to distinguish between OS X and Linux (`/Applications`).

What do I do with that information? [I choose which Android SDK files to install](https://github.com/xamarin/xamarin-android/blob/1b3a76c6874853049e89bbc113b22bc632ed5ca4/build-tools/android-toolchain/android-toolchain.targets#L18-L19):

    <CreateItem
          Include="@(AndroidSdkItem)"
          Condition=" '%(HostOS)' == '$(HostOS)' Or '%(HostOS)' == '' ">
        <Output TaskParameter="Include" ItemName="_PlatformAndroidSdkItem"/>
    </CreateItem>

[I change compiler build options based on the host OS](https://github.com/xamarin/xamarin-android/blob/1b3a76c6874853049e89bbc113b22bc632ed5ca4/build-tools/mono-runtimes/mono-runtimes.projitems#L94-L127)

    <_MonoRuntime Include="host-Darwin" Condition=" '$(HostOS)' == 'Darwin' ">
      ...
    </_MonoRuntime>
    <_MonoRuntime Include="host-Linux" Condition=" '$(HostOS)' == 'Linux' ">
      ...
    </_MonoRuntime>

[...and I change `.zip` extraction techniques based on Host OS](https://github.com/xamarin/xamarin-android/blob/1b3a76c6874853049e89bbc113b22bc632ed5ca4/src/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/UnzipDirectoryChildren.cs#L91-L102), because `System.IO.Compression.ZipFile` doesn't respect `+x` on OS X and Linux, so the command-line `unzip` program *must* be used.

I'm sure there will be more.

AndyGerlicher:
We'd like to hold off on this now. Please reopen if you feel strongly.

Ideally MSBuild doesn't have a list of known operating systems of flavors of those and passes through to the framework. Since this isn't really possible with Mono (since it doesn't exist in full framework yet), there doesn't seem to be a good alternative. I'm not sure that there's a significant ask for it right now either, and an alternative on .NET Core (the method is in the white-list) already exists.

PR #570, opened 2016-04-13
Add a `create_mono_msbuild_zip.sh` script for generating zip of the ..

radical:
.. mono build. And add a `msbuild` wrapper script to the zip file.


rainersigwald:
Seems reasonable to me. Do we prefer to (continue to) distribute the Mono build as a zip file rather than a set of NuGet packages (as we do for CoreCLR)?

radical:
Discussed out-of-band that we'll go with a nuget based solution instead of the zip files.

PR #572, opened 2016-04-13
Read toolsets from msbuild.exe.config file when in VS

AndyGerlicher:
- When we're in devenv process, read the config file from msbuild rather
   than devenv.exe.config.
 - Fixed a bug with the running tests check. Previously this value was
   always true.

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -34,10 +34,18 @@ namespace Microsoft.Build.Shared
                                                   "QTAGENT32", "CONCURRENT", "RESHARPER", "MDHOST", "TE.PROCESSHOST"
                                               };
 
+        private static readonly string[] s_visualStudioProcess = {"DEVENV"};
+
+
         // This flag, when set, indicates that we are running tests. Initially assume it's true. It also implies that
         // the currentExecutableOverride is set to a path (that is non-null). Assume this is not initialized when we
         // have the impossible combination of runningTests = false and currentExecutableOverride = null.
         private static bool s_runningTests = true;
+
+        /// <summary>
+        /// Set to true/false when we know whether or not we're running inside Visual Studio
+        /// </summary>
+        private static bool? s_runningInVisualStudio;
 
         // This is the fake current executable we use in case we are running tests.
         private static string s_currentExecutableOverride = null;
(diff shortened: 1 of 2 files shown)


src/Shared/FileUtilities.cs
```diff
@@ -110,6 +115,17 @@ private static void GetTestExecutionInfo()
...
         /// <summary>
+        /// Returns true if processName appears in the processList
+        /// </summary>
+        /// <param name="processName">Name of the process</param>
+        /// <param name="processList">List of processes to check</param>
+        /// <returns></returns>
+        private static bool IsProcessInList(string processName, string[] processList)
+        {
+            return processList.Any(s => processName?.IndexOf(s, StringComparison.CurrentCultureIgnoreCase) >= 0);
+        }
```
danmosemsft: process name is not culture specific, I think? shouldn't it be InvariantCultureIgnoreCase

AndyGerlicher: Good catch, thanks. It was that before. Didn't intend to change (that part).

src/Shared/FileUtilities.cs
```diff
@@ -930,5 +946,21 @@ private static string CurrentExecutableOverride
...
+        /// Returns true when the entry point application is Visual Studio.
+        /// </summary>
+        internal static bool RunningInVisualStudio
+        {
+            get
+            {
+                // Check if initialized and do so if not yet
+                if (!s_runningInVisualStudio.HasValue)
+                {
+                    GetTestExecutionInfo();
```
radical: Does this need to worry about thread safety or maybe preventing multiple of these running in parallel, from multiple projects being built in parallel? Maybe use `Lazy<T>` if so?

src/Shared/FileUtilities.cs
```diff
@@ -59,19 +67,16 @@ static internal partial class FileUtilities
         private static void GetTestExecutionInfo()
         {
             // Get the executable we are running
-            var program = Path.GetFileNameWithoutExtension(Environment.GetCommandLineArgs()[0]);
+            var processNameCommandLine = Path.GetFileNameWithoutExtension(Environment.GetCommandLineArgs()[0]);
+            var processNameCurrentProcess = Process.GetCurrentProcess().ProcessName;
```
rainersigwald: Why do you need both of these? Probably worth a comment.

AndyGerlicher: Honestly no idea. The previous comments were completely unhelpful for that. I assume the correct mechanism varies depending on how the process was created, but given that this code was 100% wrong before I can't really back that up. The values are generally the same from the tests I did.

src/Shared/FileUtilities.cs
AndyGerlicher: I think the class could definitely use some improvements, but I don't think there are any issues with thread safety. There's very little state and the methods that set state look to be idempotent.

src/Shared/FileUtilities.cs
rainersigwald: Ahh, I see. I missed the `if (!s_runningTests)` block below.

---
In reply to: [59752231](https://github.com/Microsoft/msbuild/pull/572#discussion_r59752231) [](ancestors = 59752231)

src/Shared/FileUtilities.cs
```diff
@@ -59,19 +67,16 @@ static internal partial class FileUtilities
         private static void GetTestExecutionInfo()
```
rainersigwald: Nit: this could be renamed to be more accurate now that it checks for VS too. `GetExecutionInfo`?

rainersigwald:
:+1:

PR #573, opened 2016-04-13
Expand extension path fallback and enable tests.

AndyGerlicher:
- This allows for something like $(MSBuildProgramFiles32)\MSBuild to be a
   fallback search path.
 - Tests were disabled in master w/o the feature defined.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -2099,12 +2099,14 @@ namespace Microsoft.Build.Evaluation
             // 2. 1 or more project files *found* but ignored (like circular, self imports)
             foreach (var extensionPath in fallbackExtensionPaths)
             {
-                if (!Directory.Exists(extensionPath))
+                string extensionPathExpanded = _data.ExpandString(extensionPath);
+
+                if (!Directory.Exists(extensionPathExpanded))
                 {
                     continue;
                 }
 
-                var newExpandedCondition = importElement.Condition.Replace(extensionPropertyRefAsString, extensionPath);
+                var newExpandedCondition = importElement.Condition.Replace(extensionPropertyRefAsString, extensionPathExpanded);
                 if (!EvaluateConditionCollectingConditionedProperties(importElement, newExpandedCondition, ExpanderOptions.ExpandProperties, ParserOptions.AllowProperties,
                             _projectRootElementCache))
                 {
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
```diff
@@ -349,14 +347,6 @@ public void ImportFromExtensionsPathAnd32And64()
                 </Project>
                 ";
 
-            string v2Folder = NativeMethodsShared.IsWindows
```
AndyGerlicher: Note: These 2 variables were defined but not used so I just deleted them (since IsWindows doesn't exist in master).

rainersigwald:
I'm not sure I understand the motivating scenario--can you go into a bit more detail?

rainersigwald:
We talked offline about this. The motivation is to be able to include a property in the fallback search path, so you can say that `$(Flibbertigibbet)\` falls back to `$(SomeOtherPath)\Fallback`. That makes sense and I can see the need. I might appreciate a bit more verbose commit message, though.

LGTM

radical:
Just thinking aloud here:

- should there be a known property like `$(UserDefinedExtensionsPathFallbackSearchPaths)` set in the app.config ?

`<property name="MSBuildExtensionsPath" value="$(UserDefinedExtensionsPathFallbackSearchPaths);c:\foo\bar" />`

- And should such a property be allowed to have more than one path, separated by `;` , thus allowing multiple additional paths to be specified in the targets file?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2099,22 +2099,24 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
             // 2. 1 or more project files *found* but ignored (like circular, self imports)
             foreach (var extensionPath in fallbackExtensionPaths)
             {
-                if (!Directory.Exists(extensionPath))
+                string extensionPathExpanded = _data.ExpandString(extensionPath);
```
radical: The error message emitted when no project file is found (`ThrowForImportedProjectFromExtensionsPathNotFound`) will not show the fully expanded search path, IIUC. `ThrowForImportedProjectFromExtensionsPathNotFound` needs to be updated, if that is desired.

AndyGerlicher: Good call. Added a test and expanded it there too.

PR #577, opened 2016-04-14
Simplify FindMSBuildExe using FileUtilities logic

AndyGerlicher:
- Remove final fall-back that will point to MSBuild in the Framework
   directory.
 - Remove checking loaded assembly locations as framework assemblies (e.g.
   mscorlib) are located next to MSBuild v4.

In both of these cases, finding/running MSBuild.exe from the framework
directory (MSBuild v4) will cause issues.

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
@@ -956,7 +956,7 @@ namespace Microsoft.Build.Execution
                 _detailedSummary = true;
             }
 
-            FindMSBuildExe();
+            _nodeExeLocation = FindMSBuildExe();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -974,31 +974,27 @@ private void EnsureToolsets()
 
         /// <summary>
         /// This method determines where MSBuild.Exe is and sets the NodeExePath to that by default.
+        /// Note - 4/14/16: This used to look at loaded assemblies and in the Framework folder for
+        /// MSBuild.exe. This would, however, find MSBuild.exe v4 that is included in the Framework
+        /// either from loaded assemblies (e.g. mscorlib) or the eventual fall-back to the framework
+        /// path. This was removed as it is certainly never correct.
```
rainersigwald: Not sure this note is worthwhile since removal of "certainly never correct" behavior seems un-noteworthy :)

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -974,31 +974,27 @@ private void EnsureToolsets()
...
+            if (!string.IsNullOrEmpty(location) && CheckMSBuildExeExistsAt(location))
             {
-                return;
+                return location;
             }
 
-            // Use the location specified in the environment.
-            // MSBUILD_EXE_PATH optionally contains the full path to msbuild.exe, and if present
-            // overrides the rest
-            string path = Environment.GetEnvironmentVariable("MSBUILD_EXE_PATH");
```
rainersigwald: Dropping the environment variable intentionally?

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
radical: It should be in the commit message instead :)

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
AndyGerlicher: That's handled in FileUtilities

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -1007,56 +1003,16 @@ private void FindMSBuildExe()
...
+                if (CheckMSBuildExeExistsAt(location))
                 {
-                    _nodeExeLocation = path;
-                    return;
+                    return location;
                 }
             }
 
-            // Search in the location of any assemblies we have loaded.
-            foreach (System.Reflection.Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
```
rainersigwald: This fallback seems like it might possibly be relevant. Have you shown that it's no longer needed somehow?

(I love the idea of getting rid of it! Just want to be cautious.)

AndyGerlicher: The problem here is the first assembly is usually mscorlib.dll. MSBuild.exe exists in that location, so it would always use that. We could filter out those, but I don't think it's needed since it wouldn't have worked since MSBuild was actually shipped in the Framework 8 years ago or whatever :smile:

rainersigwald: Makes sense. Put that in the commit message and I'll be happy.

---
In reply to: [59794311](https://github.com/Microsoft/msbuild/pull/577#discussion_r59794311) [](ancestors = 59794311)

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -1007,56 +999,16 @@ private void FindMSBuildExe()
             EnsureToolsets();
```
rainersigwald: Comment just above this about "shrug and move on" is bogus now, right?

rainersigwald:
LGTM with comment/commit message updates

PR #578, opened 2016-04-14
Add PublicKeyToken to msbuildToolsets in app.config

AndyGerlicher:
Think it didn't have it in there when we weren't using the OSS signing and had no public key. Adding it back as it was causing some issues in VS not playing nice with binding redirects.

Also removed the comment that it should use Microsoft.Build.Engine since we're no longer compatible with Microsoft.Build.Engine anyway.


rainersigwald:
LGTM

PR #579, opened 2016-04-15
Add support for handling .mdb files

radical:


src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -1351,6 +1351,9 @@ private ICollection<P> AddBuiltInProperties()
 
 #if RUNTIME_TYPE_NETCORE
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.msbuildRuntimeType, "Core"));
+#elif MONO
+            builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.msbuildRuntimeType,
+                                                        NativeMethodsShared.IsMono ? "Mono" : "Full"));
```
rainersigwald: Under what circumstances is `MONO` defined but we want to return Full?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -96,6 +96,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <TargetPlatformDisplayName Condition="'$(TargetPlatformDisplayName)' == ''">$([Microsoft.Build.Utilities.ToolLocationHelper]::GetPlatformSDKDisplayName($(TargetPlatformIdentifier), $(TargetPlatformVersion)))</TargetPlatformDisplayName>
   </PropertyGroup>
 
+  <PropertyGroup>
+      <DebugFileExt Condition="'$(DebugFileExt)' == ''">.pdb</DebugFileExt>
```
rainersigwald: I'd say to move this into the following `<PropertyGroup>` where other symbol stuff is defined.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -348,7 +352,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <ItemGroup Condition="'$(_DebugSymbolsProduced)' == 'true'">
     <_DebugSymbolsIntermediatePath Include="$(IntermediateOutputPath)$(TargetName).compile.pdb" Condition="'$(OutputType)' == 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
-    <_DebugSymbolsIntermediatePath Include="$(IntermediateOutputPath)$(TargetName).pdb" Condition="'$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == ''"/>
+    <_DebugSymbolsIntermediatePath Include="$(IntermediateOutputPath)$(TargetName)$(DebugFileExt)" Condition="'$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == '' and '$(DebugFileExt)' == '.pdb'"/>
+    <_DebugSymbolsIntermediatePath Include="$(IntermediateOutputPath)$(TargetName)$(TargetExt)$(DebugFileExt)" Condition="'$(OutputType)' != 'winmdobj' and '@(_DebugSymbolsIntermediatePath)' == '' and '$(DebugFileExt)' != '.pdb'"/>
```
rainersigwald: Maybe a comment that Mono expects names like `foo.dll.mdb`?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
radical: When running mono-msbuild with full framework? Just being careful

rainersigwald:
@AndyGerlicher what's the current and expected mid-term state of who ships the C# and VB targets? These changes look reasonable, and we want to support Mono using msbuild with our targets, but I'm not sure Roslyn would want to take this change into _their_ version of the targets, which would ideally be the canonical one.

rainersigwald:
Generally LGTM if we're not setting up pain later with C#/VB target duplication.

radical:
@rainersigwald Will the `Microsoft.CSharp.CurrentVersion.targets` also be moved to Roslyn? Where do they get a copy of that? They seem to have only `Microsoft.CSharp.Core.targets` in the source.

rainersigwald:
@radical I don't remember the details. Long term, we're trying to get individual SDK owner teams to own their targets so that this repo is only the engine and the very core targets. That's already more or less true for various SDKs (C++ is in its own world, for instance), but C# and VB have traditionally been part of the core (since we all shipped as part of the .NET framework long ago). Untangling that is an ongoing process, and my only concern with this is I don't want to cause additional pain there. But I'm pretty sure it's ok given the CurrentVersion/Core split that you mention.

radical:
Any thoughts on this?

AndyGerlicher:
@agocke @jaredpar Can you take a look at the `CscToolExe` aspect of this? It seems like with Roslyn extracted from our common targets with `Microsoft.CSharp.Core.targets` that mono might need a `Microsoft.CSharp.Mono.targets` provided to extend the functionality to their own compiler? I don't want these two strategies to diverge unnecessarily.

src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
```diff
@@ -46,6 +46,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <SynthesizeLinkMetadata Condition=" '$(SynthesizeLinkMetadata)' == '' and '$(HasSharedItems)' == 'true' ">true</SynthesizeLinkMetadata>
     </PropertyGroup>
 
+    <PropertyGroup>
+        <!-- Select a default C# compiler here, if unspecified, instead of in the Csc task, so that we can also
+             decide on the debug extension to be used. -->
+        <CscToolExe Condition="'$(CscToolExe)' == '' and '$(MSBuildRuntimeType)' != 'Mono'">csc.exe</CscToolExe>
```
agocke: This will disable the compiler server. `CscToolExe` must be unset for the compiler server to be used.

src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
```diff
@@ -46,6 +46,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <CscToolExe Condition="'$(CscToolExe)' == '' and '$(MSBuildRuntimeType)' == 'Mono'">mcs.exe</CscToolExe>
```
agocke: Why? I don't understand what using the Mono runtime has to do with using `mcs`. The compiler should be independent of the runtime.

radical: It should be, but `mcs` is what is installed by default currently, so, this is a good default when running on Mono. Otherwise anyone trying to use msbuild with mono and not having csc.exe will have to set `CscToolExe` on the command line. We could set this in a `msbuild` wrapper script, but then it overrides any value specified in the targets files, like for Xamarin Android/iOS. Also, we need to know the compiler being used to figure out the debug file extension to use.

The reasoning is explained in the commit message too.

src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
radical: Hm I need this here to figure out a debug file extension to use. I could rework this to preserve the original `CscToolExe` in case of !Mono . Would that be acceptable? For Mono, the compiler server is not being used yet.

agocke:
@AndyGerlicher I like your proposal more than this one. I'm reticent to integrate this into a shipping VS, which is, I think, what we should eventually be aiming for. My opinion about the CurrentVersion targets files is that they're basically frozen.

It seems appropriate to add an extensibility point for the Mono compiler toolchain with different targets.

agocke:
For one, I don't see why we would expect `mcs` and `csc` to have feature parity in the future.

radical:
`mcs`'s command line arguments are compatible with `csc`. And eventually, it will be phased out and replaced by `csc`. But until then most people would be using `mcs` with Mono and it is good to that scenario work by default.

So, is the suggestion that we add a `Microsoft.CSharp.Mono.targets` to roslyn and set properties in that? And import that in which targets file?

The `$(DebugFileExt)` stuff will be still be required in the Common targets files though, since that is where it handles the debug files.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -564,11 +567,29 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </AllowedReferenceAssemblyFileExtensions>
 
     <!--
+        These are the extensions that reference resolution will consider when looking for debug files related
+        to resolved references, for .pdb files
+        -->
+    <AllowedReferenceRelatedDebugFileExtensions Condition="'$(DebugFileExt)' == '.pdb'">
+        .pdb
```
cdmihai: can you replace the literal with `$(DebugFileExt)`?

As an overall comment, these .pdb / .mdb string literals appear too many times. Could you extract them out more into constants, so that you define the literal once (or as few times as possible) into a property and then use the properties instead the literals?

AndyGerlicher:
Team code review notes:

1. (required) The default for csc tool needs to stay the same, we definitely need to preserve behavior for the Roslyn case. Can you make sure to not set that?
2. We like the idea of creating a .Mono targets. I think it's reasonable to keep the target here and import it similar to how Roslyn does there and keep Mono specific changes in there. This isn't a must have, I don't think there's very much that would be in it. But it would be good to keep in mind if you want behavior differences since we can't break Roslyn or Core targets.
3. Only other comment is the conditions are a little confusing. If you can think of a better way to express. For example: `'$(DebugFileExt)' == '$(PdbFileExt)'` and the inverse on the next line.

mfilippov:
Any news about this issue?

radical:
I'll update the PR shortly to incorporate Andy's feedback.

radical:
Update:

1. A default compiler, if unset, is set only when running on Mono to `mcs.exe` .
2. Setting default debug file extensions and the above is done in a separate `Microsoft.{CSharp.VisualBasic}.Mono.targets` file .
3. Removed `$(PdbFileExt)` and directly using `.pdb` everywhere. This changes the conditions to look like `$(_DebugFileExt) == .pdb`, which looks more readable. If we use the pdb property then I'm not sure how to make it more readable. Maybe a `$(IsPdbFileExt)` ?
4. Renamed `$(DebugFileExt)` to `$(_DebugFileExt)` since it isn't meant to be overridden/used by the user.

radical:
Any feedback on this?

src/XMakeTasks/Microsoft.CSharp.Mono.targets
```diff
@@ -0,0 +1,36 @@
...
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <PropertyGroup>
+        <!-- Using _CscToolExeTmp instead of CscToolExe directly here, as we don't want to override it
+             for !mono case -->
+        <_CscToolExeTmp Condition="'$(CscToolExe)' != ''">$(CscToolExe)</_CscToolExeTmp>
+        <_CscToolExeTmp Condition="'$(CscToolExeTmp)' == '' and '$(MSBuildRuntimeType)' == 'Mono'">mcs.exe</_CscToolExeTmp>
```
agocke: So, this file basically just sets the compiler to `mcs.exe`. I still don't really understand why this is bound to the Mono runtime. It seems like the runtime shouldn't decide the compiler. For instance, last I checked, Roslyn runs on Mono. Why can't I use Roslyn if I'm running on Mono?

radical: You can, but rosyln is not installed by default with Mono! This provides a fallback. It is not preventing anyone from using Rosyln.

src/XMakeTasks/Microsoft.CSharp.Mono.targets
```diff
@@ -0,0 +1,36 @@
...
+        <CscDebugFileExt Condition="'$(CscDebugFileExt)' == '' and ('$(_CscToolExeTmp)' == 'mcs' or '$(_CscToolExeTmp)' == 'mcs.exe')">.mdb</CscDebugFileExt>
+        <_DebugFileExt Condition="'$(CscDebugFileExt)' != ''">$(CscDebugFileExt)</_DebugFileExt>
+
+        <!-- Select a default C# compiler here, if unspecified, instead of in the Csc task, so that we can also
+             decide on the debug extension to be used. Don't do this for the roslyn/csc case as it depends on the
+             default being "unset".
+             -->
+        <CscToolExe Condition="'$(CscToolExe)' == '' and '$(MSBuildRuntimeType)' == 'Mono'">mcs.exe</CscToolExe>
+
+        <UseSharedCompilation Condition="'$(UseSharedCompilation)' == '' and '$(MSBuildRuntimeType)' == 'Mono'">false</UseSharedCompilation>
```
jaredpar: Why are we disabling the compilation server on Mono?  Like the compiler it should run just fine when used on Windows.  There are issues running it on non-Windows platforms but the compiler will gracefully handle that situation and downgrade to the command line compiler.

radical: There is a bug in msbuild which causes it to try to pinvoke `CreateProcess` on OSX and hence fails. Once that is fixed, the compilation server can be enabled again.

src/XMakeTasks/Microsoft.CSharp.Mono.targets
agocke: @radical Then that sounds like something you should bake into your deployment of Mono, not MSBuild.

src/XMakeTasks/Microsoft.CSharp.Mono.targets
jaredpar: @radical can you link to the bug?  Haven't seen that yet and curious to see the details.

Also should we add a remark about that bug here so we can fix this later?

radical: I haven't filed that yet. I'll do that and update the commit message+comments.

src/XMakeTasks/Microsoft.CSharp.Mono.targets
```diff
@@ -0,0 +1,36 @@
...
+        <_CscToolExeTmp Condition="'$(CscToolExeTmp)' == '' and '$(MSBuildRuntimeType)' == 'Mono'">mcs.exe</_CscToolExeTmp>
+
+        <CscDebugFileExt Condition="'$(CscDebugFileExt)' == '' and ('$(_CscToolExeTmp)' == 'mcs' or '$(_CscToolExeTmp)' == 'mcs.exe')">.mdb</CscDebugFileExt>
+        <_DebugFileExt Condition="'$(CscDebugFileExt)' != ''">$(CscDebugFileExt)</_DebugFileExt>
+
+        <!-- Select a default C# compiler here, if unspecified, instead of in the Csc task, so that we can also
+             decide on the debug extension to be used. Don't do this for the roslyn/csc case as it depends on the
+             default being "unset".
+             -->
+        <CscToolExe Condition="'$(CscToolExe)' == '' and '$(MSBuildRuntimeType)' == 'Mono'">mcs.exe</CscToolExe>
```
jaredpar: The title of the PR is to support `.mdb` files but the implementation here is to essentially default to `mcs.exe` when running under Mono.  That doesn't seem to solve the problem stated in the title.  Consider for example what happens when I specify an `.mdb` file on the desktop framework.  That would still use Roslyn here and hence fail the compilation.

radical: Yep, it would fail in that case! But I consider that a user error, since they are explicitly opting to use `.mdb` as the debug file extension! This is meant for the case where - defaults work, and if the user wants to use some custom compiler then they can and be able to override the debug file extension too.

The compiler selection is required here for being able to select a debug file extension. In the default (non-mono) case, it will work as is - csc.exe/.pdb . In case of mono, if you are not using Rosyln, then it will work as mcs.exe/.mdb . If users want to use Rosyln, then they override the two properties and can use it.

From the commit message:

> - Providing default values for the `*DebugFileExt` requires knowing which
  compiler is being used, but if `$(CscToolExe)` is not set (by the user
  or in the project files), then we need some defaults.

>  So, we try to set reasonable *defaults* for `$(CscToolExe)` based on
  the runtime host. This still allows the user to override that,
  whether in the project/targets file or via the command line. This is
  done in new `Microsoft.{CSharp,VisualBasic}.Mono.targets` file and the
  override is set *only* for the mono case.

agocke:
@radical Sorry, I feel like my focused comments are really obtuse.

Right now, what this PR does is override the compiler based on a runtime-based flag. I would prefer to have this injected during the layout. So here's the idea: you would keep the imports clause for the mono targets, but we wouldn't actually include any Mono targets in MSBuild.

When you deploy or ship a Mono installation, you would set a variable in some parent target or environment variable saying "Use the Mono targets at this path." You would include the Mono targets at some well-defined path in your installation and direct the code to there. Once you're there, you can override whatever properties you want to set the compiler/runtime/whatever to whatever's required for your Mono installation.

This feels like we're trying to do a bottom-up dependency injection when what we want is a top-down dependency injection.

radical:
@agocke I'm not sure I understand the bit about "parent target/environment variable". Did you mean - have a conditional import for the `*Mono.targets` in the msbuild, and with the mono installation, provide that missing targets file?

agocke:
@radical Or import it based on a property or environment variable.

Conditional import based on path is probably OK -- if the targets file is present, import it. We have precedent here in the form of the "Extension" imports of MSBuild targets.

Alternatively, if you have some parent process that drives MSBuild, if you make it a conditional import, you could set an environment variable in your process, then MSBuild could pick up the value from the environment variable and import that.

radical:
Hm for the IDE (Xamarin Studio) case we could set an environment variable. For the command line case, it could be set in the wrapper script. But for any other host using MSBuild libraries, they would have to know this and set it also, it won't work by default for them.

I think the Path based condition would be better. I can handle the layout side, since we are installing msbuild with Mono, in the same installation prefix. And the import can be from `$(MSBuildExtensionsPath)`.

agocke:
@radical Seems reasonable to me.

agocke:
@AndyGerlicher You're more of an expert on MSBuild extension points than I, does this plan sound viable to you?

jaredpar:
Can we get a bit more context on what the driving user scenario is for this change?  Just looking at the code change it seems that we're essentially saying:

> If MSBuild is run under Mono we should use mcs.exe.  

That feels wrong to me.  From my perspective Mono is just another runtime like desktop or CoreCLR.  It seems very odd that we would switch from using csc to msc when swapping out runtimes.  

Imagine for example I wanted to explore taking a project of mine to build cross platform by switching to use `mono msbuild MySolution.sln`.  I would be quite confused if I suddenly started getting errors from `mcs.exe`.  

I feel like if we had a better understanding of the broader scenario here it would help with us giving better feedback on this change.

radical:
From the commit message:

> - Currently, only `.pdb` debug files generated by csc are supported.
  Mono's compilers generate `.mdb` files, which end up not getting copied
  to the output dir. Rebuild also doesn't work then as the debug file is
  one of the inputs for CoreCompile, and the `.pdb` file would always be
  missing when using mono's compilers.

> - Providing default values for the `*DebugFileExt` requires knowing which
  compiler is being used, but if `$(CscToolExe)` is not set (by the user
  or in the project files), then we need some defaults.

>  So, we try to set reasonable *defaults* for `$(CscToolExe)` based on
  the runtime host. This still allows the user to override that,
  whether in the project/targets file or via the command line. This is
  done in new `Microsoft.{CSharp,VisualBasic}.Mono.targets` file and the
  override is set *only* for the mono case.

> - `*DebugFileExt` and `*ToolExe` could be overriden in a wrapper script
  on mono, but that means overriding the values set by the project, if
  any!

--
> Imagine for example I wanted to explore taking a project of mine to build cross platform by switching to use mono msbuild MySolution.sln. I would be quite confused if I suddenly started getting errors from mcs.exe.

Without this patch you will get errors about `csc.exe` not being found, because it is *not* installed with Mono today. So, this patch helps the default case. If a user happens to have Rosyln installed, then they can explicitly switch to that. This can be removed, of course, once mono switches to Rosyln.

jaredpar:
@radical the commit messages don't provide the user scenario I was asking about.

jaredpar:
> Without this patch you will get errors about csc.exe not being found, because it is not installed with Mono today.

Is the case then when Mono installs MSBuild manually?  If that's the case then we should work to let the specific installation drive the correct decision on csc / mcs.  Right now it's a much bigger hammer by using the runtime.

radical:
>  the commit messages don't provide the user scenario I was asking about.

The user scenario is that default builds with msbuild won't work on Mono. And rebuild don't work, because of the pdb/mdb difference when using (default) mono's compilers.

> Is the case then when Mono installs MSBuild manually? If that's the case then we should work to let the specific installation drive the correct decision on csc / mcs. Right now it's a much bigger hammer by using the runtime.

Sure. @agocke's suggestion to have a condition import for the mono targets file in the msbuild targets and then install that missing mono targets file with the mono installation, should solve that. Thoughts?

Sarabeth-Jaffe-Microsoft:
MSBuild triage:
- Mono distributed target with overrides sounds like a good approach.

jaredpar:
> Sure. @agocke's suggestion to have a conditional import for the mono targets file in the msbuild targets and then install that missing mono targets file with the mono installation, should solve that. Thoughts?

I think that's pretty reasonable.  Perhaps though it shouldn't be specific to Mono?  Maybe just a general mechanism where a tool which installs MSBuild can override these targets.

mfilippov:
@radical Any new about this issue?

AndyGerlicher:
Closing this as it's very old and I'd like to cleanup old PRs. Re-open or sync with us when you believe this is ready if it's still the right decision.

mfilippov:
@radical @AndyGerlicher Without this PR MSBuild cannot put mdb in bin dir on Mac and Linux :( In mono fork it's problem solved (with some hacks).

PR #580, opened 2016-04-15
Adding file version using Nerdbank.GitVersioning

AndyGerlicher:
I spent a few minutes quickly doing this. Should I continue down this route? Thoughts? I'm not set on this particular solution, but I do want auto-increasing versions and I want the commit id in the file somewhere.

@AArnott does this work in .NET Core? Or do you have plans to?

Only code change was switching ProductVersion to FileVersion as the former caused an MSBuild crash using the Version class since it's something like `15.1.0+g1a4c5a01a8`. A test project also defined `ThisAssembly`, but it wasn't used so I deleted it.


AArnott:
The changes look good. NBGV does not yet work on CoreCLR because it depends on libgit2sharp, which isn't yet CoreCLR compliant. [I'm working through that](https://github.com/libgit2/libgit2sharp/issues/1293) and am mostly done though.

AndyGerlicher:
In triage yesterday we made the call to commit this. This is better than what we have in master currently (nothing) and will hopefully be a solution we can adopt in xplat as well when it comes to .NET Core.

src/XMakeBuildEngine/Definition/ProjectCollection.cs
```diff
@@ -384,7 +384,7 @@ public static Version Version
                     // Use .CodeBase instead of .Location, because .Location doesn't
                     // work when Microsoft.Build.dll has been shadow-copied, for example
                     // in scenarios where NUnit is loading Microsoft.Build.
-                    s_engineVersion = new Version(FileVersionInfo.GetVersionInfo(FileUtilities.ExecutingAssemblyPath).ProductVersion);
+                    s_engineVersion = new Version(FileVersionInfo.GetVersionInfo(FileUtilities.ExecutingAssemblyPath).FileVersion);
```
rainersigwald: What does ProductVersion get set to with this new system? Do we anticipate a compat impact here?

AndyGerlicher: It was something like `15.1.0+g1a4c5a01a8` which crashed Version ctor. I would hope people aren't using it in that way, but can't really be sure :disappointed:

rainersigwald:
To summarize my concerns:

* Since we're locked to a Major.Minor by being part of the VS release cycle, this will only control the third version part, which makes for problems if we need to ship a hotfix (say for VS "15" RTM) after we've already produced a build for the next update. We talked about mitigating this by changing the minor version number after each release, while keeping the .NET Assembly version the same throughout. But then we've got multiple different versions embedded in the same file.
* I'm generally leery of commit-height approaches since they're an attempt to linearize a graph.

That said, let's give it a shot. LGTM.

src/XMakeBuildEngine/Definition/ProjectCollection.cs
AArnott: AssemblyVersion is always a M.m.b[.r] format.
AssemblyFileVersion is also an M.m.b.r format
AssemblyInformationalVersion has the M.m.b+g1a4daf format.

I don't know what this GetVersionInfo API returns for ProductVersion or FileVersion though.

AArnott:
@AndyGerlicher said:

> In triage yesterday we made the call to commit this. 

yay!

@rainersigwald said:

> Since we're locked to a Major.Minor by being part of the VS release cycle, this will only control the third version part, which makes for problems if we need to ship a hotfix (say for VS "15" RTM) after we've already produced a build for the next update. We talked about mitigating this by changing the minor version number after each release, while keeping the .NET Assembly version the same throughout. But then we've got multiple different versions embedded in the same file.

I would agree with incrementing the major.minor version after each significant release (major release or Update). This leaves the 3rd version integer allocated for representing servicing of those releases and this automatically increments. In other words, major.minor is your marketed version, and the 3rd component is the technical version within that major.minor release. 

NerdBank.GitVersioning's default policy is that only the major.minor components of the version make it into the AssemblyVersion. The AssemblyFileVersion contains the more precise version. You can override this to make AssemblyVersion more precise if you want (which is what corefx does) or you can make AssemblyVersion something else entirely independent of AssemblyFileVersion. All these settings are available in the version.json file.

> I'm generally leery of commit-height approaches since they're an attempt to linearize a graph.

Agreed. The *4th* version component is an integer representation of the first two bytes of the git commit ID, which further distinguishes versions since as you say, source history isn't strictly linear. Obviously the git commit ID itself is seemingly random, so the fourth component isn't useful for sorting purposes, but it does allow you to tie it back to a specific git commit if you ever needed to (which I've found to be quite useful at times).

Personally, I consider the linearity of history to be important for releases. So while product development is not linear, releases are. So you ship 14.0, 14.1, 14.2, etc. At each of these, git height (the 3rd version component) is irrelevant because the minor number increments. If you go back and service a prior release, you service it at the tip of one of these major.minor releases' branches. So you have a master or dev branch, where most feature work happens. You also have a 14.0 branch, and a 14.1 branch, etc. You create these when you're stabilizing for your release. You tag your official release with "v14.0.{githeight}". Then if you service your 14.0 release you tag it again: "v14.0.{githeight}" where the height is larger than before. History is linear for each release branch and the 3rd version component will ensure you recognize the latest release from something newer.

rainersigwald:
MSBuild's biggest issue with changing AssemblyVersion is that there are Task assemblies compiled against us. Changing assembly version means that we need to update the binding redirects that ensure we load the right version in both msbuild.exe and VS, and increase the possibility that someone will distribute a task assembly compiled against a higher version of our assemblies than they really need (especially since the task interface is pretty darn stable at this point).

Those difficulties led to the idea of leaving AssemblyVersion alone per VS release, but having the other attributes increment in the way @AArnott describes so we can have both semi-sane versioning + a stable AssemblyVersion.

I don't have a great read on how bad just bumping AssemblyVersion for update releases would be, though. A lot of concerns might be alleviated if we just publish a 14.0 NuGet package and gently convince everyone to start using that to build their tasks. We even have a stick to go with such a thing: since we won't be in the GAC for version 15, a simple `<Reference Include="Microsoft.Build.Framework" />` will stop working on a dev15-only system.

PR #581, opened 2016-04-15
Update pkgdef to add binding redirection

AndyGerlicher:
This change will tell Visual Studio where to find MSBuild as well as
providing a binding redirect in devenv.exe.config.


rainersigwald:
Would it be accurate to say that this tells VS "where to find MSBuild and to generate a binding redirect in devenv.exe.config."? That phrasing is clearer in my head.

LGTM

PR #583, opened 2016-04-19
Toolset Translation Bug Fix

AndyGerlicher:
Quick fix for a bug I found in the fallback search paths. We can revisit the implementation, but I need to get this in to continue testing.

Basically converted the type to Dictionary<string, List<string>> since translation for those types already existed (added a simplified Dictionary translate method). I also added to the Toolset translation test functionality to include the fallback paths.

### DIFF
--- a/src/XMakeBuildEngine/Definition/Toolset.cs
+++ b/src/XMakeBuildEngine/Definition/Toolset.cs
@@ -205,6 +205,11 @@ namespace Microsoft.Build.Evaluation
         private string _defaultSubToolsetVersion;
 
         /// <summary>
+        /// Map of MSBuildExtensionsPath properties to their list of fallback search paths
+        /// </summary>
+        private Dictionary<string, List<string>> _msBuildSearchPathsTable;
+
+        /// <summary>
         /// Constructor taking only tools version and a matching tools path
         /// </summary>
         /// <param name="toolsVersion">Name of the toolset</param>
@@ -349,17 +354,21 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
-        /// Returns a copy of the list of search paths for a MSBuildExtensionsPath* property kind.
+        /// Returns a copy of the list of search paths for a MSBuildExtensionsPath* property.
+        /// <returns>List of search paths, else null</returns>
         /// </summary>
-        internal IList<string> GetMSBuildExtensionsPathSearchPathsFor(MSBuildExtensionsPathReferenceKind refKind)
+        internal List<string> GetMSBuildExtensionsPathSearchPathsFor(string refKind)
         {
-            IList<string> paths;
+            if (string.IsNullOrEmpty(refKind))
+                return null;
+
+            List<string> paths;
             if (MSBuildExtensionsPathSearchPathsTable != null && MSBuildExtensionsPathSearchPathsTable.TryGetValue(refKind, out paths))
             {
                 return new List<string>(paths);
             }
 
-            return new List<string>();
+            return null;
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


rainersigwald:
LGTM

src/XMakeBuildEngine/Definition/ToolsetConfigurationReader.cs
```diff
@@ -245,46 +245,39 @@ protected override IEnumerable<ToolsetPropertyDefinition> GetSubToolsetPropertyD
...
 
         /// <summary>
         /// Returns a list of the search paths for a given MSBuildExtensionsPathReferenceKind
         /// </summary>
-        private List<string> ComputeDistinctListOfFallbackPathsFor(MSBuildExtensionsPathReferenceKind refKind, ToolsetElement.PropertyElementCollection propertyCollection)
+        private Dictionary<string, List<string>> ComputeDistinctListOfFallbackPathsFor(ToolsetElement.PropertyElementCollection propertyCollection)
         {
-            var extnPaths = new List<string>();
-            var pathsFromConfig = propertyCollection.GetElement(refKind.StringRepresentation)?.Value;
+            var paths = new Dictionary<string, List<string>>();
```
radical: This should probably be named `pathsTable`

src/XMakeBuildEngine/Definition/ToolsetConfigurationReader.cs
```diff
@@ -245,46 +245,39 @@ protected override IEnumerable<ToolsetPropertyDefinition> GetSubToolsetPropertyD
...
+                if (string.IsNullOrEmpty(property.Name) || string.IsNullOrEmpty(property.Value))
+                {
+                    continue;
+                }
 
-            var pathsTable = new HashSet<string>();
-            foreach (var extnPath in pathsFromConfig.Split(new char[]{_separatorForExtensionsPathSearchPaths}, StringSplitOptions.RemoveEmptyEntries))
-            {
-                if (!pathsTable.Contains(extnPath))
+                var pathsTable = new List<string>();
```
radical: And this should be named `paths`.

src/XMakeBuildEngine/Definition/Toolset.cs
```diff
@@ -351,15 +356,18 @@ private Toolset(INodePacketTranslator translator)
         /// <summary>
         /// Returns a copy of the list of search paths for a MSBuildExtensionsPath* property kind.
```
radical: Maybe update the comment, since the behavior has changed to returning a List of paths or null.

src/XMakeBuildEngine/Definition/ToolsetConfigurationReader.cs
```diff
@@ -245,46 +245,39 @@ protected override IEnumerable<ToolsetPropertyDefinition> GetSubToolsetPropertyD
...
+            var paths = new Dictionary<string, List<string>>();
 
-            //FIXME: handle ; in path on Unix
-            if (String.IsNullOrEmpty(pathsFromConfig))
+            foreach (ToolsetElement.PropertyElement property in propertyCollection)
             {
-                return extnPaths;
-            }
+                //FIXME: handle ; in path on Unix
+                if (string.IsNullOrEmpty(property.Name) || string.IsNullOrEmpty(property.Value))
```
radical: Maybe we can also verify the property name is valid with `MSBuildExtensionsPathReferenceKind.FindIn` ?

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -681,24 +680,24 @@ private MSBuildExtensionsPathReferenceKind(string value)
         /// <summary>
         /// Tries to find a reference to MSBuildExtensionsPath* property in the given string
         /// </summary>
-        public static MSBuildExtensionsPathReferenceKind FindIn(string expression)
+        public static string FindIn(string expression)
```
radical: This is used only in `ExpandAndLoadImports`, which should be fine with this method returning `MSBuildExtensionsPathReferenceKind`.

radical:
I don't know about the *Translate* stuff, but other than that it looks fine. When you are looking at it again next week, we can think talk about what to do with the `MSBuildExtensionsPathReferenceKind`.

PR #586, opened 2016-04-20
Make the extensions path search/fallback generic.

AndyGerlicher:
Changed the implementation of the fallback search paths to be more
generic. It will now replace and fallback any property defined in the
config file.
 - Lots of renames to remove reference to MSBuildExtensionPath
 - Some small refactoring that bothered me while reviewing a lot of the
   code :)

I did this because ASP.NET has the following in their xproj files:
```XML
<PropertyGroup>
    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">15.0</VisualStudioVersion>
    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
  </PropertyGroup>
  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
```

I tried going down other routes that would fallback because `VSToolsPath` contained `MSBuildExtensionsPath32` but they ended up being way too messy. This kept the solution simple and similar to what @radical implemented, just more generic and allowed me to add fallback to `VSToolsPath`.

### DIFF
--- a/src/XMakeBuildEngine/Definition/ProjectImportPathMatch.cs
+++ b/src/XMakeBuildEngine/Definition/ProjectImportPathMatch.cs
@@ -1,0 +1,41 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Build.Evaluation
+{
+    /// <summary>
+    /// Struct representing a reference to a project import path with property fall-back
+    /// </summary>
+    internal struct ProjectImportPathMatch
+    {
+        /// <summary>
+        /// ProjectImportPathMatch instance representing no fall-back
+        /// </summary>
+        public static readonly ProjectImportPathMatch None = new ProjectImportPathMatch(string.Empty, new List<string>());
+
+        internal ProjectImportPathMatch(string propertyName, IEnumerable<string> searchPaths)
+        {
+            PropertyName = propertyName;
+            SearchPaths = searchPaths;
+            MsBuildPropertyFormat = $"$({PropertyName})";
+        }
+
+        /// <summary>
+        /// String representation of the property reference - eg. "MSBuildExtensionsPath32"
+        /// </summary>
+        public string PropertyName { get; }
+
+        /// <summary>
+        /// Returns the corresponding property name - eg. "$(MSBuildExtensionsPath32)"
+        /// </summary>
+        public string MsBuildPropertyFormat { get; }
+
+        /// <summary>
+        /// Enumeration of the search paths for the property.
+        /// </summary>
+        public IEnumerable<string> SearchPaths { get; }
+    }
+}
--- a/src/XMakeBuildEngine/Definition/Toolset.cs
+++ b/src/XMakeBuildEngine/Definition/Toolset.cs
@@ -174,7 +174,7 @@ namespace Microsoft.Build.Evaluation
         private DirectoryGetFiles _getFiles;
 
         /// <summary>
-        /// Delegate to check to see if a direcotry exists
+        /// Delegate to check to see if a directory exists
         /// </summary>
         private DirectoryExists _directoryExists = null;
 
@@ -205,9 +205,9 @@ namespace Microsoft.Build.Evaluation
         private string _defaultSubToolsetVersion;
 
         /// <summary>
-        /// Map of MSBuildExtensionsPath properties to their list of fallback search paths
+        /// Map of project import properties to their list of fall-back search paths
         /// </summary>
-        private Dictionary<string, List<string>> _msBuildSearchPathsTable;
+        private Dictionary<string, List<string>> _propertySearchPathsTable;
 
         /// <summary>
         /// Constructor taking only tools version and a matching tools path
@@ -278,6 +278,7 @@ namespace Microsoft.Build.Evaluation
             _environmentProperties = environmentProperties;
             _overrideTasksPath = msbuildOverrideTasksPath;
             _defaultOverrideToolsVersion = defaultOverrideToolsVersion;
+            
         }
 
         /// <summary>
@@ -289,33 +290,25 @@ namespace Microsoft.Build.Evaluation
         /// Properties that should be associated with the Toolset.
         /// May be null, in which case an empty property group will be used.
         /// </param>
-        internal Toolset(string toolsVersion, string toolsPath, PropertyDictionary<ProjectPropertyInstance> buildProperties, PropertyDictionary<ProjectPropertyInstance> environmentProperties, PropertyDictionary<ProjectPropertyInstance> globalProperties, IDictionary<string, SubToolset> subToolsets, string msbuildOverrideTasksPath, string defaultOverrideToolsVersion)
+        internal Toolset(string toolsVersion, string toolsPath, PropertyDictionary<ProjectPropertyInstance> buildProperties, PropertyDictionary<ProjectPropertyInstance> environmentProperties, PropertyDictionary<ProjectPropertyInstance> globalProperties, IDictionary<string, SubToolset> subToolsets, string msbuildOverrideTasksPath, string defaultOverrideToolsVersion, Dictionary<string, List<string>> importSearchPathsTable = null)
             : this(toolsVersion, toolsPath, environmentProperties, globalProperties, msbuildOverrideTasksPath, defaultOverrideToolsVersion)
         {
             if (_properties == null)
             {
-                if (null != buildProperties)
-                {
-                    _properties = new PropertyDictionary<ProjectPropertyInstance>(buildProperties);
-                }
-                else
-                {
-                    _properties = new PropertyDictionary<ProjectPropertyInstance>();
-                }
+                _properties = buildProperties != null
+                    ? new PropertyDictionary<ProjectPropertyInstance>(buildProperties)
+                    : new PropertyDictionary<ProjectPropertyInstance>();
             }
 
             if (subToolsets != null)
             {
                 Dictionary<string, SubToolset> subToolsetsAsDictionary = subToolsets as Dictionary<string, SubToolset>;
+                _subToolsets = subToolsetsAsDictionary ?? new Dictionary<string, SubToolset>(subToolsets);
+            }
 
-                if (subToolsetsAsDictionary != null)
-                {
-                    _subToolsets = subToolsetsAsDictionary;
-                }
-                else
-                {
-                    _subToolsets = new Dictionary<string, SubToolset>(subToolsets);
-                }
+            if (importSearchPathsTable != null)
+            {
+                _propertySearchPathsTable = importSearchPathsTable;
             }
         }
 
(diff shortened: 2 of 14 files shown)


src/XMakeBuildEngine/Definition/ProjectImportPathMatch.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// String representation of the property reference - eg. "MSBuildExtensionsPath32"
+        /// </summary>
+        public string PropertyName { get; }
+
+        public IEnumerable<string> SearchPaths { get; }
+
+        /// <summary>
+        /// Returns the corresponding property name - eg. "$(MSBuildExtensionsPath32)"
+        /// </summary>
+        public string MsBuildPropertyFormat => String.Format($"$({PropertyName})");
```
rainersigwald: Is the String.Format necessary because the compiler's not smart enough to know that an interpolated string is a function call? (just trying to calibrate my internal C# 6 compiler :))

AndyGerlicher: Yeah. Which is a little weird since $"" does string.format, but oh well. I didn't write that but I tried it without and it doesn't compile. Not sure what the recommended usage would be in a case like this.

src/XMakeBuildEngine/Definition/Toolset.cs
```diff
@@ -354,33 +347,34 @@ private Toolset(INodePacketTranslator translator)
...
         /// </summary>
-        public string ToolsVersion
-        {
-            get { return _toolsVersion; }
-        }
+        public string ToolsVersion => _toolsVersion;
 
         /// <summary>
-        /// Path to this toolset's tasks and targets. Corresponds to $(MSBuildToolsPath) in a project or targets file. 
+        /// Path to this toolsets tasks and targets. Corresponds to $(MSBuildToolsPath) in a project or targets file.
```
rainersigwald: This was right before

rainersigwald:
Looks like a lot of (good, wanted) modernization/cleanup changes. I wish they weren't mixed in the same commit with the substantial changes, though.

src/XMakeBuildEngine/Definition/Toolset.cs
```diff
@@ -354,33 +347,34 @@ private Toolset(INodePacketTranslator translator)
...
-            if (string.IsNullOrEmpty(refKind))
-                return null;
+            if (string.IsNullOrEmpty(expression) || ImportPropertySearchPathsTable == null)
+            {
+                return ProjectImportPathMatch.None;
+            }
 
-            List<string> paths;
-            if (MSBuildExtensionsPathSearchPathsTable != null && MSBuildExtensionsPathSearchPathsTable.TryGetValue(refKind, out paths))
+            foreach (var searchPath in _propertySearchPathsTable)
```
rainersigwald: Is there an ordering/subset question here? I guess not, since we're only really expanding initial properties?

rainersigwald: I guess it's not really initial properties that are relevant here, but that we're not _recursively_ expanding properties with substitutions.

src/XMakeBuildEngine/Definition/Toolset.cs
```diff
@@ -354,33 +347,34 @@ private Toolset(INodePacketTranslator translator)
...
             {
-                return new List<string>(paths);
+                if (expression.IndexOf($"$({searchPath.Key})", StringComparison.OrdinalIgnoreCase) >= 0)
```
rainersigwald: What if the import is of the form `Project="$(NonFallbackProperty)\something\$(FallbackProperty)\some.props"`? Do we still do the expected fallback in that case? Do we want to support it? Or should this be restricted to fallback for the initial property?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2086,10 +2079,11 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
             //
 
             // Adding the value of $(MSBuildExtensionsPath*) property to the list of search paths
-            var prop = _data.GetProperty(refKindInProject.StringRepresentation);
+            var prop = _data.GetProperty(fallbackSearchPathMatch.PropertyName);
+            var fallbackExtensionPaths = fallbackSearchPathMatch.SearchPaths.ToList();
```
rainersigwald: should this `ToList` be moved to ProjectImportPathMatch construction time so you don't have to do it for every import?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2086,10 +2079,11 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
             fallbackExtensionPaths.Insert(0, prop.EvaluatedValue);
```
rainersigwald: I'd like to see this block more explicit, and inserting at item 0 of a list reads wrong to me.

My instinct is something like
```C#
var pathsToSearch =
    // The actual value of the property, with no fallbacks
    new [] { prop.EvaluatedValue}
    // The list of fallbacks, in order
    .Concat(fallbackSearchPathMatch.SearchPaths).ToList();
```

Though I don't love that either.

src/XMakeBuildEngine/Definition/ProjectImportPathMatch.cs
rainersigwald: Makes sense, thanks for explaining. Should this be precomputed once at constructor time and accessed as a field?

---
In reply to: [60618769](https://github.com/Microsoft/msbuild/pull/586#discussion_r60618769) [](ancestors = 60618769)

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2514,35 +2508,33 @@ private string GetCurrentDirectoryForConditionEvaluation(ProjectElement element)
...
 
         /// <summary>
-        /// Throws InvalidProjectException because we failed to import a project from $(MSBuildExtensionsPath*)
-        /// <param name="refKindInProject">MSBuildExtensionsPath reference kind found in the Project attribute of the Import element</param>
+        /// Throws InvalidProjectException because we failed to import a project which contained a ProjectImportSearchPath fall-back.
+        /// <param name="searchPathMatch">MSBuildExtensionsPath reference kind found in the Project attribute of the Import element</param>
         /// <param name="importElement">The importing element for this import</param>
         /// </summary>
-        private void ThrowForImportedProjectFromExtensionsPathNotFound(string refKindInProject, ProjectImportElement importElement)
+        private void ThrowForImportedProjectFromWithSearchPathsNotFound(ProjectImportPathMatch searchPathMatch, ProjectImportElement importElement)
```
rainersigwald: Remove `From` from name

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2514,35 +2508,33 @@ private string GetCurrentDirectoryForConditionEvaluation(ProjectElement element)
...
+                                                                    importElement.Project.Replace(searchPathMatch.MsBuildPropertyFormat, extensionsPathPropValue),
                                                                     ExpanderOptions.ExpandProperties, importElement.ProjectLocation);
 
             string relativeProjectPath = FileUtilities.MakeRelative(extensionsPathPropValue, importExpandedWithDefaultPath);
 
-            var onlyFallbackSearchPaths =
-                _data.Toolset.GetMSBuildExtensionsPathSearchPathsFor(refKindInProject)
-                    .Select(s => _data.ExpandString(s))
-                    .ToList();
+            var onlyFallbackSearchPaths = searchPathMatch.SearchPaths.Select(s => _data.ExpandString(s)).ToList();
```
rainersigwald: Only enumerated once, skip `ToList()`?

src/XMakeBuildEngine/Definition/ToolsetConfigurationReader.cs
```diff
@@ -223,50 +192,50 @@ protected override IEnumerable<ToolsetPropertyDefinition> GetSubToolsetPropertyD
...
-                if (kind.Equals(MSBuildExtensionsPathReferenceKind.None))
+                if (string.IsNullOrEmpty(property.Value) || string.IsNullOrEmpty(property.Name))
                 {
                     continue;
                 }
 
                 var paths = new List<string>();
+
+                //FIXME: handle ; in path on Unix
                 foreach (var extnPath in property.Value.Split(new [] { _separatorForExtensionsPathSearchPaths }, StringSplitOptions.RemoveEmptyEntries))
```
rainersigwald: should have seen this before, but this could just be:
```C#
var paths =
    property.Value.Split(new[] {_separatorForExtensionsPathSearchPaths},
        StringSplitOptions.RemoveEmptyEntries).Distinct();
```
right?

Do we need/want to worry about case consolidation for Windows (and default OS X HFS+)?

rainersigwald:
What happened to the just-operate-on-fully-expanded-paths idea? Trying this more-restrictive policy first?

rainersigwald:
Many nits, several questions, no big problems. LGTM

cdmihai:
>                 }

If I get this right, you'd get different expansions when indirection is present:
```
$a = foo
$b = $(a)bar
$c = $(b)$(a)
```
If $a has fallback for "zoo", $c gets expanded to foobarzoo instead of zoobarzoo. This happens because only the direct reference of $a is referenced.

Would this then actually work with the ASP.NET's msbuild code `<Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />` where $VSToolsPath is similar to $C?




---
Refers to: src/XMakeBuildEngine/Evaluation/Evaluator.cs:2111 in 855f679. [](commit_id = 855f679cdd2c945277e6e53a1d95d962d7eaf1c7, deletion_comment = False)

src/XMakeBuildEngine/Evaluation/Evaluator.cs
AndyGerlicher: It's for StringifyList below which needs an indexer.

AndyGerlicher:
@rainersigwald I went down the route of expand based on values and it worked but it was super ugly. Especially with the corner cases. It worked ok if the replacements are simple and the replacements are rooted paths, but to handle generic cases it was really ugly and ended up evaluating and keeping track of a lot of extra things. And since this would have to execute for *every* import I don't really know what it would do to the perf. This implementation stays simple and somewhat limited and I think that ended up being better, at least for now.

AndyGerlicher:
@cdmihai In the cases you mentioned, $(c) would get expanded to `foobarfoo` the first time. And if that import failed then it would try `foobarzoo`. It would never try zoobarzoo, and that's really the reason I kept this the simpler method. When you start dealing with non-rooted paths it gets very ugly.

For the ASP.NET case, that was definitely the issue is it needed the indirect reference. In that case evaluation 1 it expands `VSToolsPath` which does include `MSBuildExtensionsPath` and it fails. But I also added VSToolsPath as a search path, so 2nd try it would try to import the fallback for that path. Whether or not the 2nd level contains a fallback path is irreverent, it won't get replaced.

src/XMakeBuildEngine/Definition/Toolset.cs
AndyGerlicher: This would work fine. It would first try as is and then try with $(FallbackProperty) replaced with something else. $(NonFallbackProperty) just behaves normally.

src/XMakeBuildEngine/Definition/Toolset.cs
AndyGerlicher: Order does matter, I put a note in the xmldoc that it's the first one found. Order is based on what's in the config file.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
AndyGerlicher: It was previously copied to a new list elsewhere so it could be modified here (at least that's what I assumed). I fixed that up a little bit in the latest iteration and added your suggestion, which I like much better.

src/XMakeBuildEngine/Definition/Toolset.cs
rainersigwald: Ah, but if it's  `Project="$(FallbackProperty1)\something\$(FallbackProperty2)\some.props"`, it won't try both/the combinatorial explosion? I'm pretty ok with that :ok_hand:

---
In reply to: [60695397](https://github.com/Microsoft/msbuild/pull/586#discussion_r60695397) [](ancestors = 60695397)

PR #588, opened 2016-04-22
Exec: Don't add `cd $working_dir` as that is set when the process is

radical:
started, so it isn't required and would be incorrect.


radical:
I'm not sure whether [this](https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeTasks/Exec.cs#L229-L231) should be changed too.

rainersigwald:
LGTM. The `pushd` command should **not** be changed; it's being used for a side effect: when you `pushd \\some\shared\folder` this has the side effect of "mounting" the share to a drive letter, which is required to set the working directory in CMD. I'll clarify the comment to make that more explicit.

PR #591, opened 2016-04-26
Update ToolsVersion to 15.0

AndyGerlicher:
We want the assembly version to (for now at least) not be 15.0 which would
conflict with Visual Studio "15" Preview which is loaded in the GAC.
However, we do want to be compatible with the ToolsVersion. This change
updates the ToolsVersion from 15.1 to 15.0 and leaves the assembly version
at 15.1.0.0.


rainersigwald:
As we've discussed offline: this is confusing, but I think it's the best compromise. This way, the toolset version matches the VS version.

LGTM

PR #593, opened 2016-04-27
Fix OSX CI for master

AndyGerlicher:
It should be disabled (same as Ubuntu).


rainersigwald:
This is one of those things where it only takes effect after going into master, huh?

LGTM

PR #595, opened 2016-04-28
Update fopx setup package name

AndyGerlicher:
It doesn't need the version. Only used internally for setup.


rainersigwald:
LGTM

PR #596, opened 2016-04-28
Fix signing issue with setup package

AndyGerlicher:
The created fopx had to start with the OutDir in order to sign. I added
another folder (layout) and that was no longer true.


rainersigwald:
LGTM

PR #599, opened 2016-05-02
FI (partial) master to VS2015 Update 3

AndyGerlicher:
Initial set of changes to be included in Update 3.


AndyGerlicher:
There were no conflicts, so just feedback on if the commits should or should not go with Update 3. Many of them are just purely build changes I made for perf and warnings.

rainersigwald:
List LGTM

PR #605, opened 2016-05-04
Override FileUtilities process info when in Visual Studio

AndyGerlicher:
The GetExecutionInfo method is not very clear on when it should override
the execution info (i.e. what process we're in). With this change,
devenv.exe is no longer considered as a test harness (not sure why it was
in the first place) and we now always override the execution info unless
we're in msbuild.exe. I believe this is the intent of the method in the
first place, but there have been so many issues in it I'm not entirely
sure.

This resolves an issue where in certain cases running in Visual Studio
configuration is read from devenv.exe.config for toolsets.

### DIFF
--- a/src/XMakeBuildEngine/Definition/ToolsetReader.cs
+++ b/src/XMakeBuildEngine/Definition/ToolsetReader.cs
@@ -8,6 +8,7 @@
 using System;
 using System.IO;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Collections;
 using Microsoft.Build.Execution;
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -122,28 +122,28 @@ ToolsetDefinitionLocations locations
 
                 if (configurationReader != null)
                 {
-                    configurationReaderToUse = configurationReader == null ? new ToolsetConfigurationReader(environmentProperties, globalProperties) : configurationReader;
```
rainersigwald: Wow.

rainersigwald:
:+1:

src/XMakeBuildEngine/Definition/ToolsetReader.cs
AndyGerlicher: It's what you do when you're really serious about not wanting it to be null!

PR #611, opened 2016-05-11
Fix for ToolLocationHelper issue finding tools path

AndyGerlicher:
GetPathToBuildToolsFile(..) would return null when in Visual Studio (wrong
config file read) and the bitness was ignored. This change will ensure
the correct .config file is read.

Also added support for 64-bit tools path when reading from the config file
toolset (added MSBuildToolsPath64).

Sample code:
```C#
string buildTools15 = ToolLocationHelper.GetPathToBuildToolsFile("msbuild.exe", "15.0", DotNetFrameworkArchitecture.Bitness32);
```

Unfortunately this scenario already had tests for it, but they were disabled :(


src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -627,10 +627,35 @@ public void TestGetPathToBuildToolsFile_64Bit()
...
-            Assert.Equal(tv12path, ToolLocationHelper.GetPathToBuildToolsFile("msbuild.exe", ToolLocationHelper.CurrentToolsVersion, UtilitiesDotNetFrameworkArchitecture.Bitness64));
+                Assert.Equal(tv12path, ToolLocationHelper.GetPathToBuildToolsFile("msbuild.exe", ObjectModelHelpers.MSBuildDefaultToolsVersion, UtilitiesDotNetFrameworkArchitecture.Bitness64));
+                Assert.Equal(tv12path, ToolLocationHelper.GetPathToBuildToolsFile("msbuild.exe", ToolLocationHelper.CurrentToolsVersion, UtilitiesDotNetFrameworkArchitecture.Bitness64));
+            }
+            catch (Exception)
+            {
+                if (created)
+                {
+                    FileUtilities.DeleteDirectoryNoThrow(toolsPath64, true);
+                }
```
rainersigwald: throw ?

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -592,14 +591,15 @@ public void TestGetPathToBuildToolsFile_32Bit()
...
+
+            var toolsPath32 = ProjectCollection.GlobalProjectCollection.GetToolset(ObjectModelHelpers.MSBuildDefaultToolsVersion).Properties["MSBuildToolsPath32"];
+            string tv12path = Path.Combine(Path.GetFullPath(toolsPath32.EvaluatedValue), "msbuild.exe");
 
             Assert.Equal(tv12path, ToolLocationHelper.GetPathToBuildToolsFile("msbuild.exe", ObjectModelHelpers.MSBuildDefaultToolsVersion, UtilitiesDotNetFrameworkArchitecture.Bitness32));
             Assert.Equal(tv12path, ToolLocationHelper.GetPathToBuildToolsFile("msbuild.exe", ToolLocationHelper.CurrentToolsVersion, UtilitiesDotNetFrameworkArchitecture.Bitness32));
         }
 
-        [Fact(Skip = "Ignored in MSTest")]
-        // Ignore: Test requires installed toolset.
```
rainersigwald: We should audit for this string since we no longer require installation.

rainersigwald:
:shipit:

rainersigwald:
[looks at CI build results]

Looks like tests are crashing pretty hard and not even leaving a results file. Not happening locally?

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
AndyGerlicher: uhg, good catch!

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
AndyGerlicher: I tried the other two skipped tests in this file but they failed because of `$(WindowsSDK80Path)` not defined. But I can't figure out where it gets defined even in the installed version :(. But yes in general we should audit all of the skipped tests and understand the reasons.

AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:
This review has taken a lot longer and was more involved than I'd hoped! My initial change uncovered a pretty significant gap in using what we're building vs. what is installed on the machine.

In the latest commits since initial review I had to enable MSBuildTaskHost.exe, which we really needed anyway for compat reasons. I wish we didn't, but I don't see any way around it. With that enabled I also updated the common .tasks and .overridetasks to match what is shipping in VS2015.

AndyGerlicher:
I tried to enable VerifyToolsetAndToolLocationHelperAgreeWhenVisualStudioVersionIs11, but it fails on CI and not locally and I can't figure out why :(.

rainersigwald:
Possibly related to having multiple versions of VS installed/not installed on a machine?

---
In reply to: [219268250](https://github.com/Microsoft/msbuild/pull/611#issuecomment-219268250) [](ancestors = 219268250)

rainersigwald:
I'm confused by the `.overridetasks` conditions. Are you just pulling them over from the VS repo, or do you understand them? Either way I'd appreciate comments in the commit message.

Also, does revitalizing MSBuildTaskHost mean that there are some disabled tests we can reenable? Seems like it should.

AndyGerlicher:
I updated the commit message. I copied `.overridetasks` and the conditions as is. The only thing I changed was 14.0 -> 15.1. I do wish I understood more of it :(.

PR #612, opened 2016-05-12
Adding build support for Platform target x86 or x64

AndyGerlicher:
Initial support. I still need to hook up the build.proj to enable these to build in parallel.

Also hitting some FileTracker test failures because the folders weren't cleaned up properly, so I fixed a few issues there.


src/Utilities/UnitTests/TrackedDependencies/FileTrackerTests.cs
```diff
@@ -1828,65 +1828,82 @@ public void MoveAlwaysRecordsWrites()
...
+                // Item1: tlog pattern
+                // Item2: # times it's expected to appear
+                IList<Tuple<string, int>> tlogPatterns = new List<Tuple<string, int>>();
+                tlogPatterns.Add(new Tuple<string, int>("ProcessLaunchTest-abc*tlog", 3));
+
+                LaunchDuplicateToolsAndVerifyTlogExistsForEach(testDir, contextSpecifications, tlogPatterns, createTestDirectory: false);
+            }
+            catch (Exception)
+            {
+                FileUtilities.DeleteDirectoryNoThrow(testDir, true);
```
rainersigwald: throw;

src/Utilities/UnitTests/TrackedDependencies/FileTrackerTests.cs
```diff
@@ -2101,34 +2118,44 @@ public void LaunchMultipleDifferentTools()
...
+                contextSpecifications.Add(new Tuple<string, IList<Tuple<string, string, int>>>("ProcessLaunchTest", toolsToLaunch));
+                contextSpecifications.Add(new Tuple<string, IList<Tuple<string, string, int>>>("ProcessLaunchTest2", toolsToLaunch));
+
+                // Item1: tlog pattern
+                // Item2: # times it's expected to appear
+                IList<Tuple<string, int>> tlogPatterns = new List<Tuple<string, int>>();
+                tlogPatterns.Add(new Tuple<string, int>("ProcessLaunchTest-abc*tlog", 3));
+                tlogPatterns.Add(new Tuple<string, int>("ProcessLaunchTest2-abc*tlog", 3));
+
+                LaunchDuplicateToolsAndVerifyTlogExistsForEach(testDir, contextSpecifications, tlogPatterns, false);
```
rainersigwald: I like leaving the optional name in for bools, personally.

rainersigwald:
Looking at the existing binaries, only MSBuild.exe is actually built for an architecture (internally this seems to be controlled by `PlatformSpecificBuild=true`:

```
o:\msbuild>corflags "C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe"
Microsoft (R) .NET Framework CorFlags Conversion Tool.  Version  4.6.1055.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0xb
ILONLY    : 1
32BITREQ  : 1
32BITPREF : 0
Signed    : 1
```

Everything else is AnyCpu:
```
o:\msbuild>corflags "C:\Program Files (x86)\MSBuild\14.0\Bin\Microsoft.Build.dll"
Microsoft (R) .NET Framework CorFlags Conversion Tool.  Version  4.6.1055.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0x9
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 1
```

We should probably emulate that here.

rainersigwald:
Can you include more background in the commit message?

MSBuild ships in explicit x86/x64 binaries as opposed to AnyCPU, for historical reasons and because a custom task assembly may wish to P/Invoke native code. Having explicit entry points makes it easy to ensure that MSBuild runs with the right architecture for your process.

As a rule, though, tasks should be written as managed and AnyCPU.

AndyGerlicher:
@dotnet-bot test this please

CreateSetupPackageLayout.proj
```diff
@@ -4,74 +4,43 @@
...
+      <X64OutputPath>$(BaseOutputPath)X64\$(OS)\$(Configuration)\</X64OutputPath>
       <SetupPackageRoot>$(BinDir)Setup\</SetupPackageRoot>
       <SetupPackageLayoutDir>$(SetupPackageRoot)layout\</SetupPackageLayoutDir>
       <SetupPackageOutputFile>$(SetupPackageRoot)MSBuild.fopx</SetupPackageOutputFile>
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- Commenting out files that aren't produced by the open source build
-    (When comparing against VS2015 installed to Program Files\MSBuild\14.0\Bin) -->
+    <BinX86Output Include="$(X86OutputPath)Microsoft.Build.dll" />
```
rainersigwald: The redundancy here is crying out for a "common" item + transforms. But it'll wait.

dir.traversal.targets
```diff
@@ -16,10 +16,9 @@
   </Target>
 
   <Target Name="TestAllProjects">
-    <!-- To Serialize we use msbuild's batching functionality '%' to force it to batch all similar projects with the same identity
-         however since the project names are unique it will essentially force each to run in its own batch -->
+    <!-- MSBuild tests must be sequential -->
     <MSBuild Targets="Test" Projects="@(Project)" Condition="'$(SerializeProjects)'=='true'" Properties="Dummy=%(Identity)"/>
-    <MSBuild Targets="Test" Projects="@(Project)" Condition="'$(SerializeProjects)'!='true'" BuildInParallel="true" />
+    <MSBuild Targets="Test" Projects="@(Project)" Condition="'$(SerializeProjects)'!='true'" BuildInParallel="false" />
```
rainersigwald: Get rid of the double specification and `$(SerializeProjects)` stuff here? It's redundant now.

CreateSetupPackageLayout.proj
AndyGerlicher: I have to redo all this soon anyway. I think there's a new VSIX format the fopx. I'll try to make it better then :).

rainersigwald: Works for me!

PR #619, opened 2016-05-13
Skip CI test result archiving if no test result files found

cdmihai:
We can't differentiate PR jobs by the branch they target. This is an issue because PRs targeting master should only have Windows CI jobs. 

For tests, this means the CI should only collect test results when test result files are there. The only risk here is that if someone turns off the tests entirely, the main build succeeds and Jenkins will not complain.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -9,6 +9,7 @@ def project = GithubProject
     ['Windows_NT', 'OSX', 'Ubuntu'].each {osName ->
         def isPR = false
         def newJobName = ''
+        def skipTestsWhenResultsNotFound = true
 
         if (branch == 'pr') {
             isPR = true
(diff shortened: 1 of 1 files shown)


rainersigwald:
Change seems reasonable. Is this fixing a problem we're currently seeing?

AndyGerlicher:
Only thing I don't like about this is if there are no tests reported I would consider that a failure. How about always report when on Windows and set `reported=true` then if `reported==false` after the switch optionally report?

cdmihai:
Added a flag so windows builds require test results to be there.

cdmihai:
No test run summaries for !Windows CI jobs in the PR CI results.

---
In reply to: [219143399](https://github.com/Microsoft/msbuild/pull/619#issuecomment-219143399) [](ancestors = 219143399)

AndyGerlicher:
LGTM

rainersigwald:
Got it.

LGTM

---
In reply to: [219188863](https://github.com/Microsoft/msbuild/pull/619#issuecomment-219188863) [](ancestors = 219188863,219143399)

PR #620, opened 2016-05-13
Adopt exception filters for IO exception filtering

rainersigwald:
This reduces a fair amount of boilerplate that happened all over.


AndyGerlicher:
+255 -555, very nice!

PR #627, opened 2016-05-17
Respect $(TargetFrameworkRootPath) in ToolLocationHelper APIs

radical:
`FrameworkPathOverride` set in `Microsoft.Common.CurrentVersion.targets`,
does not use the `$(TargetFrameworkRootPath)` while calling
`ToolLocationHelper.GetPathToStandardLibraries(...)` .

    <FrameworkPathOverride Condition="'$(FrameworkPathOverride)' == ''">$([Microsoft.Build.Utilities.ToolLocationHelper]::GetPathToStandardLibraries($(TargetFrameworkIdentifier), $(TargetFrameworkVersion), $(TargetFrameworkProfile), $(PlatformTarget)))</FrameworkPathOverride>

This means that if the project is targeting a custom framework,
which is installed in a directory different from the default root path,
then it won't be able to find the framework and eventually,
`FrameworkPathOverride` will be set to an incorrect path. This can be
seen with mono/osx which has an alternate directory for
Xamarin mobile frameworks, set on the command line like `/p:TargetFrameworkRootPath=/Library/Frameworks/Mono.framework/External/xbuild-frameworks` .

This manifests as an `_ExplicitReference` getting added to
`mscorlib.dll` but with the wrong path, causing compilation to fail.

### DIFF
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -1587,6 +1587,21 @@ namespace Microsoft.Build.Utilities
         /// <returns>Collection of reference assembly locations.</returns>
         public static string GetPathToStandardLibraries(string targetFrameworkIdentifier, string targetFrameworkVersion, string targetFrameworkProfile, string platformTarget)
         {
+            return GetPathToStandardLibraries(targetFrameworkIdentifier, targetFrameworkVersion, targetFrameworkProfile, platformTarget, null);
+        }
+
+        /// <summary>
+        /// Returns the path to mscorlib and system.dll
+        /// </summary>
+        /// <param name="targetFrameworkIdentifier">Identifier being targeted</param>
+        /// <param name="targetFrameworkVersion">Version being targeted</param>
+        /// <param name="targetFrameworkProfile">Profile being targeted</param>
+        /// <param name="platformTarget">What is the targeted platform, this is used to determine where we should look for the standard libraries. Note, this parameter is only used for .net frameworks less than 4.0</param>
+        /// <param name="targetFrameworkRootPath">Root directory where the target framework will be looked for. Uses default path if this is null</param>
+        /// <exception cref="ArgumentNullException">When the frameworkName is null</exception>
+        /// <returns>Collection of reference assembly locations.</returns>
+        public static string GetPathToStandardLibraries(string targetFrameworkIdentifier, string targetFrameworkVersion, string targetFrameworkProfile, string platformTarget, string targetFrameworkRootPath)
+        {
             ErrorUtilities.VerifyThrowArgumentLength(targetFrameworkIdentifier, "targetFrameworkIdentifier");
             ErrorUtilities.VerifyThrowArgumentLength(targetFrameworkVersion, "targetFrameworkVersion");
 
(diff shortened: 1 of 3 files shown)


src/Utilities/ToolLocationHelper.cs
```diff
@@ -1645,13 +1660,35 @@ public static string GetPathToStandardLibraries(string targetFrameworkIdentifier
...
+        {
             ErrorUtilities.VerifyThrowArgumentLength(targetFrameworkVersion, "targetFrameworkVersion");
             ErrorUtilities.VerifyThrowArgumentLength(targetFrameworkIdentifier, "targetFrameworkIdentifier");
             ErrorUtilities.VerifyThrowArgumentNull(targetFrameworkProfile, "targetFrameworkProfile");
 
             Version frameworkVersion = ConvertTargetFrameworkVersionToVersion(targetFrameworkVersion);
             FrameworkNameVersioning targetFrameworkName = new FrameworkNameVersioning(targetFrameworkIdentifier, frameworkVersion, targetFrameworkProfile);
-            return GetPathToReferenceAssemblies(targetFrameworkName);
+            if (String.IsNullOrEmpty(targetFrameworkRootPath))
+                targetFrameworkRootPath = FrameworkLocationHelper.programFilesReferenceAssemblyLocation;
```
cdmihai: You can replace this line with a call to the method bellow: GetPathToReferenceAssemblies(FrameworkNameVersioning frameworkName). It assumes a program files root path

radical: Good point, single place for the default.

src/Utilities/ToolLocationHelper.cs
```diff
@@ -1579,9 +1579,25 @@ public static string GetPathToStandardLibraries(string targetFrameworkIdentifier
...
+        /// <param name="platformTarget">What is the targeted platform, this is used to determine where we should look for the standard libraries. Note, this parameter is only used for .net frameworks less than 4.0</param>
+        /// <param name="targetFrameworkRootPath">Root directory where the target framework will be looked for. Uses default path if this is null</param>
+        /// <exception cref="ArgumentNullException">When the frameworkName is null</exception>
+        /// <returns>Collection of reference assembly locations.</returns>
+        public static string GetPathToStandardLibraries(string targetFrameworkIdentifier, string targetFrameworkVersion, string targetFrameworkProfile, string platformTarget, string targetFrameworkRootPath)
+        {
             ErrorUtilities.VerifyThrowArgumentLength(targetFrameworkIdentifier, "targetFrameworkIdentifier");
             ErrorUtilities.VerifyThrowArgumentLength(targetFrameworkVersion, "targetFrameworkVersion");
 
+            Console.WriteLine($"@@@@@@@@@@@@@@@@@@@@@@@ GetPathToStandardLibraries root: {targetFrameworkRootPath}");
```
AndyGerlicher: Can you remove this? Looks like debug only.

radical: I'll debug+update this PR, it was failing on CI but not locally. These statements were added earlier for debugging.

rainersigwald:
LGTM

AndyGerlicher:
LGTM (once the debug info is gone and tests pass)

mfilippov:
Any news about this PR?

Sarabeth-Jaffe-Microsoft:
@radical Are these changes ready for review and are there more to come?

radical:
Updated to fix the tests. The issue was that on Mono, the `FrameworkList.xml` for custom frameworks uses an attribute - `TargetFrameworkDirectory`, which points to an alternate location for the assemblies. This is required because of the way assemblies are laid out in a mono install. And `msbuild` supports [this](https://github.com/Microsoft/msbuild/blob/xplat/src/Utilities/ToolLocationHelper.cs#L2944) when running on Mono. And the tests here were using that attribute. And since the CI server on OSX is running tests for a CoreCLR build, they failed.

@Sarabeth-Jaffe-Microsoft Yes, now it is ready for review, thanks!

radical:
Any thoughts on this?

rainersigwald:
LGTM. @AndyGerlicher still look good to you?

AndyGerlicher:
@dotnet-bot test this please

PR #631, opened 2016-05-18
Adding Framework 4.6.2 for VS "15"

AndyGerlicher:
Also some code cleanup. There's a lot more that could be done (especially the constants at the top of the file that are poorly used), but hopefully it's improved some while I was there.


rainersigwald:
:shipit:

src/Shared/FrameworkLocationHelper.cs
```diff
@@ -269,6 +273,7 @@ internal static class FrameworkLocationHelper
                 dotNetFrameworkVersion451,
                 dotNetFrameworkVersion46,
                 dotNetFrameworkVersion461,
+                dotNetFrameworkVersion462
             }),
```
cdmihai: VS14 has .net 452 but VS15 doesn't. Is that how it should be?

PR #632, opened 2016-05-18
Bump up msbuild to newer corefx versions

cdmihai:
* bump up corefx, build tools, and CLI
* replaced TypeFilter usages with `Func<Type, object, bool> `
* Removed Environment.GetCommandLineArgs
* Replaced Assembly.Location with reflection access
* Conditionally add / remove whitelisted property functions based on .net core availability
* Delayed static property function receiver type resolution until the method name is known
* Replaced MSBuild's implementation of Path.GetFullPath with the one in the .Net framework
* Updated nuget framework monikers
* Replaced explicit runtimes with nuget --infer-runtimes to get around Nuget's stricter restore


cdmihai:

cdmihai:

rainersigwald:
>         /// Path.GetFullPath is slow and creates strings in its work: this version doesn't.

Can you update this comment with the new state of the world for 4.6.2 (?) and .NET Core?

---
Refers to: src/Shared/FileUtilities.cs:308 in b9a24b1. [](commit_id = b9a24b1157e7c9958a675b1eb95e36c269ea11a4, deletion_comment = False)

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -856,13 +856,13 @@ public void ItemIncludeContainsMultipleItemReferences()
...
 
+#if FEATURE_LEGACY_GETFULLPATH
         /// <summary>
         /// Bad path when getting metadata through ->Metadata function
         /// </summary>
         [Fact]
-        [Trait("Category", "netcore-osx-failing")]
-        [Trait("Category", "mono-osx-failing")]
-        public void InvalidPathAndMetadataItemFunction()
+        [PlatformSpecific(PlatformID.Windows)]
```
rainersigwald: :+1:

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -876,6 +876,27 @@ public void InvalidPathAndMetadataItemFunction()
...
             logger.AssertLogContains("MSB4023");
         }
+#endif
+
+        /// <summary>
+        /// Bad path with illegal windows chars when getting metadata through ->Metadata function
+        /// </summary>
+        [Fact]
+        [PlatformSpecific(PlatformID.Windows)]
+        public void InvalidPathAndMetadataItemFunctionInvalidWindowsPathChars()
```
rainersigwald: Should we add others for OS X and Linux? Worth filing an up-for-grabs issue.

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2473,24 +2537,71 @@ public void PropertyFunctionStaticMethodChained()
             Assert.Equal(DateTime.Parse(_dateToParse).ToString("yyyy/MM/dd HH:mm:ss"), result);
         }
 
+#if FEATURE_SPECIAL_FOLDERS
         /// <summary>
-        /// Expand property function calls a static method an enum argument
+        /// Expand property function calls a static method available only on net46 (Environment.GetFolderPath)
```
rainersigwald: *that* calls?

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2473,24 +2537,71 @@ public void PropertyFunctionStaticMethodChained()
...
+                currentPlatformString = OSPlatform.Linux.ToString();
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                currentPlatformString = OSPlatform.OSX.ToString();
+            }
+
+            var propertyFunction = "$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform(" +
+                                        $"$([System.Runtime.InteropServices.OSPlatform]::Create(\"{currentPlatformString}\")" +
+                                   ")))";
```
rainersigwald: We should still make a prettier way to expose this, right? As in #568?

targets/DeployDependencies.proj
```diff
@@ -78,10 +78,13 @@
   <UsingTask TaskName="PrereleaseResolveNuGetPackageAssets" AssemblyFile="$(ToolsDir)net45/Microsoft.DotNet.Build.Tasks.dll"/>
 
   <PropertyGroup>
+    <!-- netstandard is not a runnable framework moniker. If on .net core, target netcoreapp instead, which is runnable -->
```
rainersigwald: I might rephrase to

At build time, we target netstandard to be runnable on many runtimes. To get a set of runnable bits, we need to deploy a specific one, so we choose netcoreapp.

?

src/XMakeBuildEngine/Resources/Constants.cs
```diff
@@ -210,7 +210,7 @@ internal static long AssemblyTimestamp
     internal static class AvailableStaticMethods
     {
         /// <summary>
-        /// Static methods that are allowed in constants. Key = Type or Type::Method, Value = AssemblyQualifiedTypeName (where null = mscorlib)
+        /// Static methods that are allowed in constants. Key = Type or Type::Method, Value = Tuple of AssemblyQualifiedTypeName (where null = mscorlib) or the actual type object
```
rainersigwald: *and* the actual type object?

src/XMakeBuildEngine/Resources/Constants.cs
```diff
@@ -219,43 +219,94 @@ internal static class AvailableStaticMethods
...
+        /// 1st try: <see cref="typeFullName"/>;
+        /// 2nd try: <see cref="typeFullName"/>::<see cref="simpleMethodName"/>
+        /// 
+        /// </summary>
+        /// <param name="typeFullName">namespace qualified type name</param>
+        /// <param name="simpleMethodName"> name of the method</param>
+        /// <returns></returns>
+        internal static Tuple<string, Type> GetTypeInformationFromTypeCache(string typeFullName, string simpleMethodName)
+        {
+            var methodFullName = CreateQualifiedMethodName(typeFullName, simpleMethodName);
```
rainersigwald: Should this be inlined so we don't have to call `CreateQualifiedMethodName` if the `typeFullName` works?

src/XMakeBuildEngine/Resources/Constants.cs
```diff
@@ -219,43 +219,94 @@ internal static class AvailableStaticMethods
...
+        /// <summary>
+        /// Tries to retrieve the type information for a type name / method name combination. 
+        /// 
+        /// It does 2 lookups:
+        /// 1st try: <see cref="typeFullName"/>;
+        /// 2nd try: <see cref="typeFullName"/>::<see cref="simpleMethodName"/>
+        /// 
+        /// </summary>
+        /// <param name="typeFullName">namespace qualified type name</param>
+        /// <param name="simpleMethodName"> name of the method</param>
```
rainersigwald: leading space before name

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -2609,6 +2609,71 @@ internal enum InvokeType
         }
 #endif
 
+        private class FunctionBuilder<T>
```
rainersigwald: I might find the codeflow below clearer if this was a struct passed by ref--then the void methods that alter it in place would jump out to me more.

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -2769,67 +2820,76 @@ internal static Function<T> ExtractPropertyFunction(string expressionFunction, I
...
 
-                    objectType = typeof(string);
-                }
+                    var receiverType = typeof(string);
+                        
+                    // If we are recursively acting on a type that has been already produced
+                    // then pass that type inwards
+                    if (propertyValue != null)
+                    {
+                        receiverType = propertyValue.GetType();
```
rainersigwald: collapse this with the above to something like
```C#
// If we are recursively acting on a type that has been already produced
// then pass that type inwards, but default to string
var receiverType = propertyValue.?GetType() ?? typeof(string);
```
?

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -3055,94 +3115,99 @@ internal object Execute(object objectInstance, IPropertyProvider<T> properties,
             }
 
             /// <summary>
-            /// Return a Type object for the type we're trying to call static methods on
+            /// Given a type name and method name, try to resolve the type.
             /// </summary>
-            private static Type GetTypeForStaticMethod(string typeName)
+            /// <param name="typeName">May be full name or assembly qualified name</param>
+            /// <param name="simpleMethodName"></param>
```
rainersigwald: I think this deserves an explanation :angel:

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -3055,94 +3115,99 @@ internal object Execute(object objectInstance, IPropertyProvider<T> properties,
...
-                // If the env var MSBUILDENABLEALLPROPERTYFUNCTIONS=1 then we'll allow pretty much anything
-                Type objectType;
-                Tuple<string, Type> functionType;
+                Type receiverType;
+                Tuple<string, Type> cachedTypeInformation;
 
                 // If we don't have a type name, we already know that we won't be able to find a type.  
                 // Go ahead and return here -- otherwise the Type.GetType() calls below will throw.
-                if (String.IsNullOrEmpty(typeName))
+                if (string.IsNullOrEmpty(typeName.Trim()))
```
rainersigwald: `string.IsNullOrWhitespace` should do this in one shot, right?

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -3055,94 +3115,99 @@ internal object Execute(object objectInstance, IPropertyProvider<T> properties,
...
-                        typeName = functionType.Item1;
+                        var assemblyQualifiedTypeName = cachedTypeInformation.Item1;
 
                         // Get the type from the assembly qualified type name from AvailableStaticMethods
-                        objectType = Type.GetType(typeName, false /* do not throw TypeLoadException if not found */, true /* ignore case */);
+                        receiverType = Type.GetType(assemblyQualifiedTypeName, false /* do not throw TypeLoadException if not found */, true /* ignore case */);
+
+                        // If the type information from the cache is not loadable, it means the cache information got corrupted somehow
+                        // Throw here to prevent adding null types in the cache
+                        ErrorUtilities.VerifyThrowInternalNull(receiverType, $"Type information for {typeName} was present in the whitelist cache as {assemblyQualifiedTypeName} but the type could be loaded.");
```
rainersigwald: could *not* be loaded?

rainersigwald:
:shipit:

cdmihai:

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
cdmihai: Opened #636

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
cdmihai: That is a separate topic, and it should not update how this test looks because it tests a lot of stuff. I also just beefed up the test to exercise many property function cases :)

targets/DeployDependencies.proj
cdmihai: This will be fun to write down :)

For actual, runnable, msbuild deployments, we only target the netcoreapp runtime moniker with the .netstandard1.3 moniker. If we want to deploy msbuild for net46, we'll use the net46 moniker, not .netstandard1.3. If you deploy the .netstandard1.3 flavor of msbuild on net46, you get a full framework msbuild with the feature set of xplat msbuild (no gac, no registry, etc). We never tested that combination.

On the other hand, for API usage (the use case where users create tasks and need to extend ITask) the users won't deploy msbuild for a specific runtime moniker, they will deploy their own task for a specific runtime moniker. Their app just happens to have msbuild in its dependency closure.

This means that:
- we are testing the use case of `get a runnable msbuild deployment for netstandard1.3 on .net core`
- we are not testing the use case of `get a runnable msbuild deployment for netstandard1.3 on net46`, since this does not make much sense (as stated above, you'd get a full framework msbuild with the crossplatform feature set)
- we are not testing the use case of `msbuild is part of the dependency graph of another app, and that app publishes for multiple runtime framework monikers`. We should try and do this :)

src/XMakeBuildEngine/Resources/Constants.cs
cdmihai: It's a union type where at least one has to be non-null. If both are null, the code fallbacks to Type.GetType(`the type string that the user put in the build script`)

rainersigwald: Ah, I see. If we were using code contracts I'd whine and want that as its own class with postconditions but oh well!

cdmihai:
@dotnet-bot test this please

PR #629, opened 2016-05-17
Xplat script improvements

rainersigwald:
A few script fixes:

* Use the just-restored `corerun` for the second (post-bootstrap) build, for better test coverage and since that's required after the merge from master.
* Make `init-tools.sh` actually fail in more failure cases.

Fixes #628 and #590.

Thanks, @akoeplinger!

PR #634, opened 2016-05-19
Archive MSBuild logs on job failure

rainersigwald:
The build logs are often *most* interesting when the build fails, so we should be capture them regardless of job failure.

Noticed this when trying to investigate the Ubuntu failure on #629.


AndyGerlicher:
LGTM

PR #635, opened 2016-05-19
Exec: Convert dos line endings to unix ones when running on Unix.

radical:
The contents of `Command` are put into a script file and having `\r` an
empty line causes bash to complain, like:

    <Exec Command="^M
    echo hello
    ^M
    "/>

.. when built :

    /Users/ankit/test/msbuild/exec.proj(4,3): error MSB3073: The command "
    /Users/ankit/test/msbuild/exec.proj(4,3): error MSB3073: echo hello
    /Users/ankit/test/msbuild/exec.proj(4,3): error MSB3073:
    /Users/ankit/test/msbuild/exec.proj(4,3): error MSB3073: " exited with code 127.

### DIFF
--- a/src/XMakeTasks/Exec.cs
+++ b/src/XMakeTasks/Exec.cs
@@ -58,13 +58,25 @@ namespace Microsoft.Build.Tasks
         private readonly List<ITaskItem> _nonEmptyOutput = new List<ITaskItem>();
         private Encoding _standardErrorEncoding;
         private Encoding _standardOutputEncoding;
+        private string _command;
 
         #endregion
 
         #region Properties
 
         [Required]
-        public string Command { get; set; }
+        public string Command
+        {
+            get { return _command; }
+            set
+            {
+                _command = value;
+                if (Path.DirectorySeparatorChar == '/')
+                {
+                    _command = _command.Replace("\r\n", "\n");
+                }
+            }
+        }
 
         public string WorkingDirectory { get; set; }
 


rainersigwald:
This one should probably go to `xplat` since it's Unix only.

src/XMakeTasks/Exec.cs
```diff
@@ -58,13 +58,25 @@ public Exec()
...
 
         [Required]
-        public string Command { get; set; }
+        public string Command
+        {
+            get { return _command; }
+            set
+            {
+                _command = value;
+                if (Path.DirectorySeparatorChar == '/')
```
cdmihai: Isn't it better to use NativeMethodsShared.IsUnixLike?

radical: I thought I would go with the existing convention being used in the [file](https://github.com/microsoft/msbuild/blob/xplat/src/XMakeTasks/Exec.cs#L196), but I guess `IsUnixLike` would be better. And looks like it is being used in the [file](https://github.com/microsoft/msbuild/blob/xplat/src/XMakeTasks/Exec.cs#L553) already.

radical:
And I should add a test for this. I will have to open a new PR for targeting xplat, will do that!

src/XMakeTasks/Exec.cs
cdmihai: Could you also please replace all the occurences of `if (Path.DirectorySeparatorChar == '/')` with IsUnixLike? :)

radical:
Opened #640 instead and closing this.

PR #643, opened 2016-05-23
Build only MSBuild.exe platform specific

### LINKED ISSUE #618: Update x86 / x64 builds to match release binaries
Looking at the existing binaries, only MSBuild.exe is actually built for an architecture (internally this seems to be controlled by `PlatformSpecificBuild=true`:

```
o:\msbuild>corflags "C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe"
Microsoft (R) .NET Framework CorFlags Conversion Tool.  Version  4.6.1055.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0xb
ILONLY    : 1
32BITREQ  : 1
32BITPREF : 0
Signed    : 1
```

Everything else is AnyCpu:
```
o:\msbuild>corflags "C:\Program Files (x86)\MSBuild\14.0\Bin\Microsoft.Build.dll"
Microsoft (R) .NET Framework CorFlags Conversion Tool.  Version  4.6.1055.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0x9
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 1
```

We should probably emulate that here.

AndyGerlicher:
This change will build MSBuild.exe and MSBuildTaskHost.exe as x86/x64 and
all other assemblies AnyCPU. This matches the shipping configuration.

Also replaced AnyCPU default with x86 to match the the default in our
shipping code.

Closes #618

When building **x86**:

```
MSBuild.exe
Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0xb
ILONLY    : 1
32BITREQ  : 1
32BITPREF : 0
Signed    : 1
```
```
Microsoft.Build.dll
Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0x9
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 1
```

When building **x64**:
```
Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32+
CorFlags  : 0x9
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 1
```
```
Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0x9
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 1
```

### DIFF
--- a/CreateBootstrappedMSBuild.cmd
+++ b/CreateBootstrappedMSBuild.cmd
@@ -1,12 +1,11 @@
 :: Usage:
-:: BuildAndCopy
+:: Rebuild the repo and copy to bin\bootstrap. Includes required MSBuild
+:: companion files copied from your local machine.
 :: 
-:: Example: BuildAndCopy.cmd
+:: Example: CreateBootstrappedMSBuild.cmd <AdditionalBuildCommand>
 
 @echo off
 setlocal
-
-set DebugBuildOutputPath=%~dp0bin\Windows_NT\Debug
 
 :: Check prerequisites
 if not defined VS140COMNTOOLS (
@@ -16,15 +15,14 @@ if not defined VS140COMNTOOLS (
 )
 
 if not "%1"=="" (
-    set OutputPath=%1
+    set AdditionalBuildCommand=%1
 )
 
 echo ** Building with the installed MSBuild
-echo ** Output Path: %DebugBuildOutputPath%
-echo ** Additional Build Parameters:%AdditionalBuildCommand%
+echo ** Additional Build Parameters: %AdditionalBuildCommand%
 echo.
 :: Build MSBuild
-call "%~dp0build.cmd" /t:Rebuild %AdditionalBuildCommand%
+call "%~dp0build.cmd" /t:Rebuild /p:Platform=x86 %AdditionalBuildCommand%
 set BUILDERRORLEVEL=%ERRORLEVEL%
 
 :: Kill Roslyn, which may have handles open to files we want
@@ -38,7 +36,7 @@ if %BUILDERRORLEVEL% NEQ 0 (
 
 :: Make a copy of our build
 echo ** Copying bootstrapped MSBuild to the bootstrap folder
-msbuild /verbosity:minimal CreatePrivateMSBuildEnvironment.proj
+msbuild /verbosity:minimal CreatePrivateMSBuildEnvironment.proj /p:Platform=x86
 if %ERRORLEVEL% NEQ 0 (
     echo.
     echo Failed building CreatePrivateMSBuildEnvironment.proj with error %ERRORLEVEL%
--- a/RebuildWithLocalMSBuild.cmd
+++ b/RebuildWithLocalMSBuild.cmd
@@ -1,12 +1,10 @@
 :: This script will:
-::   1) Rebuild MSBuild source tree.
-::   2) Create a copy of the build output in bin\MSBuild
+::   1) Rebuild MSBuild source tree (x86 only).
+::   2) Create a "bootstrapped" copy of the build output in bin\MSBuild.
 ::   3) Build the source tree again with the MSBuild.exe in step 2.
 
 @echo off
 setlocal
-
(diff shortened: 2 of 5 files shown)


dir.props
```diff
@@ -77,8 +77,13 @@
   <!-- Set default Configuration and Platform -->
   <PropertyGroup>
     <Configuration Condition="'$(Configuration)' ==''">Debug</Configuration>
-    <Platform Condition="'$(Platform)'==''">AnyCPU</Platform>
-    <PlatformTarget Condition="'$(PlatformTarget)'==''">$(Platform)</PlatformTarget>
+    <Platform Condition="'$(Platform)'==''">x86</Platform>
```
cdmihai: Shouldn't this also be gated by PlatformSpecificBuild == true?

RebuildWithLocalMSBuild.cmd
```diff
@@ -1,26 +1,25 @@
 :: This script will:
-::   1) Rebuild MSBuild source tree.
-::   2) Create a copy of the build output in bin\MSBuild
+::   1) Build MSBuild source tree.
```
cdmihai: Rainer told me to keep this as rebuild when I also renamed it to build on the xplat branch, since it ends up calling the Rebuild target :)

CreateBootstrappedMSBuild.cmd
```diff
@@ -16,15 +15,14 @@ if not defined VS140COMNTOOLS (
...
 )
 
 echo ** Building with the installed MSBuild
-echo ** Output Path: %DebugBuildOutputPath%
-echo ** Additional Build Parameters:%AdditionalBuildCommand%
+echo ** Additional Build Parameters: %AdditionalBuildCommand%
 echo.
 :: Build MSBuild
-call "%~dp0build.cmd" /t:Rebuild %AdditionalBuildCommand%
+call "%~dp0build.cmd" /t:Rebuild /p:Platform=x86 %AdditionalBuildCommand%
```
cdmihai: Could you make this a parameter that gets pushed down here from RebuildWithLocalMSBuild? That way the script can be called for any architecture, though I don't know how relevant that requirement is.

dir.props
AndyGerlicher: That's just the default Platform. Before it was AnyCPU (for everything). Now it's x86, but only on projects that are `PlatformSpecificBuild==true`.

dir.props
```diff
@@ -77,8 +77,13 @@
...
+
+    <!-- Only build for specific platforms when PlatformSpecificBuild == true. Otherwise
+         build for AnyCPU. Only MSBuild.exe and MSBuildTaskHost.exe should need this -->
+    <PlatformTarget Condition="'$(PlatformSpecificBuild)'!='true'">AnyCPU</PlatformTarget>
+    <PlatformTarget Condition="'$(PlatformSpecificBuild)'=='true'">$(Platform)</PlatformTarget>
+	<ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
```
rainersigwald: The dreaded tab character!

rainersigwald: What throws the target-arch-mismatch warning? I would expect x86/x64 -> anycpu to always work, am I misremembering?

AndyGerlicher: I thought so as well, but there were issue I can't remember right now what they were. Perhaps unit test projects. I'll take a look at it again, for now it works though.

CreateBootstrappedMSBuild.cmd
AndyGerlicher: I didn't do this because it doesn't really make sense to bootstrap x64 like this. I want to convert the bootstrap to build both platforms and deploy x64 to the amd64 sub-folder, but maybe next time.

PR #645, opened 2016-05-23
Update setup package creation

AndyGerlicher:
- Convert fopx to swixproj format
 - Include version number in manifest. Not really ideal as it's calculated
   again, but it should always be the same version.
 - Update to latest version of GitVersioning package.


setup/MSBuild.swixproj
```diff
@@ -0,0 +1,49 @@
...
+    <FilesToSign Include="@(SigningTarget)">
+      <Authenticode>Vsix</Authenticode>
+    </FilesToSign>
+  </ItemGroup>  
+
+  <ItemGroup>
+    <Package Include="Files.swr" />
+  </ItemGroup>
+
+  <Target Name="GetVxisPackageVersion" BeforeTargets="Build" DependsOnTargets="GetBuildVersion">
```
rainersigwald: `Vsix` instead of `vxis`?

rainersigwald: Nit: can you be more specific than "build"? And maybe rename this to "SetVsixProperties"?

rainersigwald:
LGTM with some nits.

PR #652, opened 2016-05-25
Update setup

AndyGerlicher:
- Add vsmanproj to merge the final manifest.
 - Add tag to ngen our exe/dlls.

### DIFF
--- a/setup/MSBuild.swixproj
+++ b/setup/MSBuild.swixproj
@@ -45,5 +45,7 @@
     </PropertyGroup>
   </Target>
 
+  <!-- MicroBuild plugin fails if GetNativeManifest is not found for some reason -->
+  <Target Name="GetNativeManifest" />
   <Import Project="$(MicroBuildDir)MicroBuild.Core.targets" />
 </Project>
--- a/setup/MSBuild.vsmanproj
+++ b/setup/MSBuild.vsmanproj
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\dir.props" />
+  <Import Project="$(MicroBuildDir)MicroBuild.Core.props" />
+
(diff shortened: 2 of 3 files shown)


setup/MSBuild.swixproj
```diff
@@ -45,5 +45,7 @@
     </PropertyGroup>
   </Target>
 
+  <!-- MicroBuild plugin fails if GetNativeManifest is not found for some reason -->
+  <Target Name="GetNativeManifest" />
```
rainersigwald: This is pretty common behavior--I think it's just from using `ProjectReference` from common targets.

setup/MSBuild.vsmanproj
```diff
@@ -0,0 +1,26 @@
...
+    <OutDir>$(BaseOutputPath)setup\</OutDir>
+    <OutputPath>$(OutputPath)setup\</OutputPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="MSBuild.swixproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <MergeManifest Include="$(BaseOutputPath)setup\*.json" />
```
rainersigwald: Because these files are produced by `MSBuild.swixproj`, this wildcard is perilous. Since it's outside a target, the wildcard is eagerly expanded in pass 3 of this project's evaluation, before any targets (including the target that deals with `ProjectReference`s) are built. So it's not guaranteed to contain anything (on a clean build starting with this project).

I would recommend moving this inside a target with `DependsOn="ResolveProjectReferences"` and `BeforeTargets=` whatever consumes this item. Then it would always be up to date.

Since this is built as a separate step after the other project in the build definition, this is isn't a must-fix here.

rainersigwald:
LGTM

setup/MSBuild.vsmanproj
AndyGerlicher: Good catch! Eventually these two swix and vsman projects can be merged and I'll make sure that's taken into account. For now they're just called separately from a build definition and not meant to be consumed outside of that.

PR #654, opened 2016-05-26
Replace uses of #!/bin/bash with #!/usr/bin/env bash

jamesqo:
It's generally best practice to prefix a Bash script with `#!/usr/bin/env bash` rather than `#!/bin/bash`. Although 99% of the time the `bash` executable will be in `/bin` (or at least a symlink of it will be), in some cases it may not be-- see [this StackOverflow answer](http://stackoverflow.com/a/21613044/4077294) for an example.

Using `env` ensures we search through all the directories in $PATH for `bash`, rather than restricting it to just the one in `/bin`. Also, there shouldn't be a performance loss since `env` replaces its own process with the arguments it's given, rather than starting a new one (similar to the difference between `command` and [`exec command`](http://wiki.bash-hackers.org/commands/builtin/exec) in Bash).

### DIFF
--- a/cibuild.sh
+++ b/cibuild.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
--- a/src/Utilities/ToolTask.cs
+++ b/src/Utilities/ToolTask.cs
@@ -1506,7 +1506,8 @@ namespace Microsoft.Build.Utilities
 
                     if (!runningOnWindows)
                     {
-                        File.AppendAllText(_temporaryBatchFile, "#!/bin/bash\n"); // first line for UNIX is ANSI
+                        // Use sh rather than bash, as not all 'nix systems necessarily have Bash installed
+                        File.AppendAllText(_temporaryBatchFile, "#!/bin/sh\n"); // first line for UNIX is ANSI
                         // This is a hack..!
                         File.AppendAllText(_temporaryBatchFile, commandLineCommands.Replace('\\', Path.DirectorySeparatorChar), EncodingUtilities.CurrentSystemOemEncoding);
                     }
--- a/src/XMakeTasks/Exec.cs
+++ b/src/XMakeTasks/Exec.cs
@@ -244,7 +244,8 @@ namespace Microsoft.Build.Tasks
                 }
                 else
                 {
-                    sw.WriteLine("#!/bin/bash");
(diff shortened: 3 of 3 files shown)


rainersigwald:

rainersigwald:
Thank you! This looks like a good change to me. Thanks especially for the link to why this is a better way.

jonpryor:
> It's generally best practice to prefix a Bash script with #!/usr/bin/env bash 

I believe the best practice should *actually* be to use `sh`, e.g. `#!/usr/bin/env sh`, *not* `bash`.

Rationale: not all Linux distros or Unixy systems will have `bash` installed. For example, [Debian Squeeze and later install *dash*, which is /bin/sh`](https://wiki.debian.org/Shell).

I thus agree with using `/usr/bin/env`, but I don't see why `bash` should be required. `sh` should be used.

rainersigwald:
That thought crossed my mind too, @jonpryor. I think we actually use bash-isms in `cibuild.sh` (right @cdmihai?) but the wrapper scripts used in `Exec` and `ToolTask` should be `sh`.

@jamesqo Would you like to make that change as well? If you're not interested we can do it separately since this is better than the previous situation.

radical:
Could we please move the `cibuild.sh` change to a separate commit? And can we include @jonpryor's rationale for `sh` in the commit messages too?

cdmihai:
I had changed `cibuild.sh` from `sh` to `bash` because it ran on osx but failed on ubuntu with bad argument evaluation. Since it was blocking me on my other task, and since I did not quite understand the difference in interpretation, I went with bash.

As @rainersigwald said, for `Exec` and `ToolTask` we should switch to `sh`. But then we have to make sure the rest of the msbuild codebase does not use bash-isms.

jamesqo:
@jonpryor I agree too, `sh` should probably be used rather than `bash` (other than the cibuild script). If we go down that route, though, I think having `env` is redundant: bash isn't installed on some systems because it isn't part of POSIX (iirc), but all systems are required to have a `sh` executable in `/bin`. See: http://stackoverflow.com/a/5725402/4077294

rainersigwald:

This looks great to me.

cdmihai:
LGTM.

Let's hold off merging this until we enable tests on Linux CI

Sarabeth-Jaffe-Microsoft:
Thanks! We're going to merge this since the Linux tests are taking a while.

PR #659, opened 2016-05-26
Update nuspecs from targetting dotnet to targetting netstandard1.3

cdmihai:
The `dotnet` moniker is the deprecated moniker for `dnxcore50`, and `dnxcore50` is the deprecated moniker for `netcoreapp1.0`
Since we want msbuild packages to be referenceable from both `net46` and `netcoreapp1.0`, they should really target the intersection APIs, which is `netstandard1.3`

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -21,7 +21,7 @@
 
   <!-- Tool versions -->
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


rainersigwald:
:shipit:

cdmihai:

PR #660, opened 2016-05-26
Add tests for running portable tasks

cdmihai:
An attempt to write some automated tests to check that tasks should be runnable by all msbuild flavors.
The tests invoke our sample portable task with the desktop msbuild, and with the non-desktop msbuilds.

Unfortunately this is not ideal. The ideal case would also check that msbuild's nuget packaging allows consumers of msbuild packages to target both `net46` and `netcoreapp1.0` / `netstandard1.3` and assert that the msbuild packages get successfully restored.

Unfortunately this requires a big change to xplat's build logic so it executes in 3 steps:
1) build and package msbuild
2) restore the sample projects using the local, freshly built packages
3) run msbuild tests (some of them depend on the sample projects)

### DIFF
--- a/src/XMakeTasks/UnitTests/PortableTasks_Tests.cs
+++ b/src/XMakeTasks/UnitTests/PortableTasks_Tests.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using System.Text.RegularExpressions;
+using Microsoft.Build.Shared;
+using Microsoft.Build.SharedUtilities;
+using Xunit;
+
+namespace Microsoft.Build.UnitTests
+{
+    public sealed class PortableTasks_Tests
+    {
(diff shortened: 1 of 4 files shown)


src/XMakeTasks/UnitTests/PortableTasks_Tests.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.Build.Shared;
+using Microsoft.Build.SharedUtilities;
+using Xunit;
+
+namespace Microsoft.Build.UnitTests
+{
+    public sealed class PortableTasks_Tests
+    {
+        private static readonly string ProjectFilePath = Path.Combine(FileUtilities.CurrentExecutableDirectory, "portableTaskTest.proj");
```
rainersigwald: Maybe assert that this exists in `RunMSBuildOnProjectWithPortableTaskAndAssertOutput`? I don't think I'd have drawn the connection (I didn't even remember that it existed).

rainersigwald:
:shipit:

cdmihai:

PR #661, opened 2016-05-26
gitignore: Ignore VS Code-generated files

jamesqo:
Noticed that the `.vscode` directory was showing up in `git status` after making my changes; this should help it ignore such directories, for users of VS Code.


cdmihai:
LGTM

Issue #544, opened 2016-03-27
AssignLinkMetadata is missing in xplat Microsoft.Tasks.Core


agocke:

AndyGerlicher:
We haven't for whatever reason converted this to .NET Core yet. Looking quickly at the code I think it should be pretty straightforward. Is this blocking anything for you (to help with priority)?

agocke:
@AndyGerlicher We don't have anything depending on it, but it did force me to remove the reference in the Microsoft.Common.CurrentVersion.targets file. If you remove the reference yourself until it's implemented I don't think it require any fixups on our side.

PR #678, opened 2016-06-08
Remove AssignLinkMetadata from targets

agocke:
When the AssignLinkMetadata task is implemented this should be added
back, but until then it causes builds which don't depend on it
to fail unnecessarily. Works around #544


rainersigwald:
Change looks good to me (thanks!) but I'm worried about forgetting to restore this. Maybe we just need to move #544 into some sort of "xplat v1 release" milestone? @cdmihai @AndyGerlicher thoughts?

AndyGerlicher:
I'll close this in favor of the fix for this task. Currently the tests fail, once I fix that it should go in.

PR #680, opened 2016-06-09
Add MSBuildFrameworkToolsPath64 to app.config

### LINKED ISSUE #679: MSBuildFrameworkToolsPath64
One of our goal is to separate the bitness of MSBuild from the compiler/link/etc tools.  One issue came up when looking for TLBREF.dll used by midl.exe.  64bit MIDL couldn’t find the 64bit tlbref.dll because Framework64 wasn’t in the path because it isn't available on 32bit MSBuild.

MSBuildFrameworkToolsPath32 – always 32bit Framework
MSBuildFrameworkToolsPath – depends on msbuild bitness.

Proposal: To add $(MSBuildFrameworkToolsPath64) on 32bit MSBuild to point to Framework64=="C:\windows\Microsoft.NET\Framework64\v4.0.30319".

AndyGerlicher:
Also updated VCTargetsPaths to match current installed toolset.

@yuehuang010 can you also review the `VCTargetsPath` for legacy? The current `VCTargetsPath` points to the low-impact location.

Closes #679

### DIFF
--- a/src/XMakeCommandLine/app.config
+++ b/src/XMakeCommandLine/app.config
@@ -42,17 +42,22 @@
         <property name="MSBuildRuntimeVersion" value="4.0.30319" />
         <property name="MSBuildFrameworkToolsPath" value="$(SystemRoot)\Microsoft.NET\Framework\v$(MSBuildRuntimeVersion)\" />
         <property name="MSBuildFrameworkToolsPath32" value="$(SystemRoot)\Microsoft.NET\Framework\v$(MSBuildRuntimeVersion)\" />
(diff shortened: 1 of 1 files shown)


yuehuang010:
For completeness, is this for only 32bit MSBuild or both 64bit MSBuild?  Should the variable be available in both?

AndyGerlicher:
Implemented this way it would be for both, the path just wouldn't exist on 64. I'm not sure how it could be done for only 64-bit like this. That might be something that would need to be defined in targets?

rainersigwald:
LGTM

PR #681, opened 2016-06-09
Detecting Host Process to launch new instances of msbuild when /m is passed in

joperezr:
We are currently trying to move all the tools that are used to build corefx(which includes MSBuild) to run on the DotNet CLI and the shared framework that comes with it. The problem with MSBuild, is that when `\m` is passed in, it always tries to look for corerun.exe (or corerun in non-Windows) to lunch new processes like `corerun.exe MSBuild.exe ...`. With my change, msbuild will now detect the host process that was used originally, and use the same one for its children. For example:
```
>dotnet.exe msbuild.exe /m build.proj
//That will cause that the children processes will be like:
>dotnet.exe msbuild.exe ...
```

PTAL: @cdmihai @rainersigwald @dsplaisted 
FYI: @weshaggard @stephentoub

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -490,12 +490,13 @@ namespace Microsoft.Build.BackEnd
             CommunicationsUtilities.Trace("Launching node from {0}", msbuildLocation);
 
 #if RUNTIME_TYPE_NETCORE
-            string coreRunName = NativeMethodsShared.IsUnixLike ? "corerun" : "CoreRun.exe";
-            string exeName = Path.Combine(Path.GetDirectoryName(msbuildLocation), coreRunName);
+            // Run the child process with the same host as the currently-running process.
+            string pathToHost;
+            using (Process currentProcess = Process.GetCurrentProcess()) pathToHost = currentProcess.MainModule.FileName;
             commandLineArgs = "\"" + msbuildLocation + "\" " + commandLineArgs;
 
             ProcessStartInfo processStartInfo = new ProcessStartInfo();
-            processStartInfo.FileName = exeName;
+            processStartInfo.FileName = pathToHost;
             processStartInfo.Arguments = commandLineArgs;
             processStartInfo.CreateNoWindow = (creationFlags | BackendNativeMethods.CREATENOWINDOW) == BackendNativeMethods.CREATENOWINDOW;
             processStartInfo.UseShellExecute = false;


src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -490,8 +490,18 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
             CommunicationsUtilities.Trace("Launching node from {0}", msbuildLocation);
 
 #if RUNTIME_TYPE_NETCORE
-            string coreRunName = NativeMethodsShared.IsUnixLike ? "corerun" : "CoreRun.exe";
-            string exeName = Path.Combine(Path.GetDirectoryName(msbuildLocation), coreRunName);
+            // Try to get the process name in case of different host running.
+            string hostName = Process.GetCurrentProcess().MainModule.FileName;
```
stephentoub: Nit: it'd be good to dispose of the Process instance when you're done with it

joperezr: do you mean something like this?
```c#
var currentProcess = Process.GetCurrentProcess();
string hostName = currentProcess.MainModule.FileName;
currentProcess.Dispose()
```

stephentoub: Yes, or more specifically:
```C#
string hostname;
using (Process p = Process.GetCurrentProcess()) hostName = p.MainModule.FileName;
```

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -490,8 +490,18 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
...
 
 #if RUNTIME_TYPE_NETCORE
-            string coreRunName = NativeMethodsShared.IsUnixLike ? "corerun" : "CoreRun.exe";
-            string exeName = Path.Combine(Path.GetDirectoryName(msbuildLocation), coreRunName);
+            // Try to get the process name in case of different host running.
+            string hostName, exeName;
+            using (Process currentProcess = Process.GetCurrentProcess()) hostName = currentProcess.MainModule.FileName;
+            if (string.IsNullOrEmpty(hostName))
+            {
+                hostName = NativeMethodsShared.IsUnixLike ? "corerun" : "CoreRun.exe";
```
weshaggard: Do we even need this fallback? Are there any known cases where currentProcess.MainModule.FileName returns empty?

joperezr: mm actually no, I just added it as a fallback in the case where the process name couldn't be retrieved but without having a specific case in mind. I'll remove it unless somebody thinks that this might be empty. @stephentoub is that right?

weshaggard: I kind of prefer to not have corerun hard-coded at all and if we end up finding cases then we should come up with a better fix.

stephentoub: This shouldn't be necessary.  If it is, it's a product bug we should fix.

joperezr: Done.

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -490,12 +490,13 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
             CommunicationsUtilities.Trace("Launching node from {0}", msbuildLocation);
 
 #if RUNTIME_TYPE_NETCORE
-            string coreRunName = NativeMethodsShared.IsUnixLike ? "corerun" : "CoreRun.exe";
-            string exeName = Path.Combine(Path.GetDirectoryName(msbuildLocation), coreRunName);
+            // Try to get the process name in case of different host running.
```
rainersigwald: Maybe say "Run the child process with the same host as the currently-running process" to better capture intent?

rainersigwald:
LGTM with my comment update. Can you please squash to a single commit since I don't think the evolution here will be interesting in the long-term?

joperezr:
Done.

cdmihai:
LGTM

brthor:
For .net core could we expose the host currently being used to run msbuild as a well-known property? This would alleviate the need to pass through the host being used so csc.exe is invoked properly in the cli.

https://github.com/dotnet/cli/pull/3639

Does this seem like a reasonable approach?

rainersigwald:
@brthor Sounds reasonable to me. Want to file an issue so we can figure out a name, and if there are any lurking issues?

joperezr:
Whenever that happens, then we should remove the [current workaround we have in corefx](https://github.com/dotnet/corefx/blob/master/dir.props#L52)

brthor:
Will do @rainersigwald thanks!

PR #684, opened 2016-06-09
RI dev14 Update 3 RC

rainersigwald:
Note: much duplication here is from cherry-picks @AndyGerlicher did to get bugfixes into U3, so the changes should already exist in master.


src/XMakeTasks/GetFrameworkPath.cs
```diff
@@ -115,13 +115,13 @@ public override bool Execute()
         public string FrameworkVersion452Path => s_version452Path.Value;
 
         /// <summary>
-        /// Path to the v4.6 framework, if available
+        /// Path to the v4.5.2 framework, if available
```
rainersigwald: Fixing

PR #687, opened 2016-06-13
Update xunit version in dir.props

rainersigwald:
This was missed in the update to xunit 2.1.0 RTM in a86122bd0. Tests
passed on machines that had the old package installed but failed on clean
machines.


AndyGerlicher:
LGTM

PR #691, opened 2016-06-15
Update Setup / Shim Files

AndyGerlicher:
- Update setup vsix package (renamed package to Microsoft.Build).
- Add default MSBuild.rsp (we ship with this today).
- Add shim files to setup package and renamed for consistency.

These changes are made to match the setup package we produce internally with what we currently ship.


rainersigwald:
:shipit:

PR #693, opened 2016-06-15
ToolsetReader: Fixes and code improvements for mono build

mfilippov:
I fix build on Mono

### DIFF
--- a/src/XMakeBuildEngine/Definition/ToolsetReader.cs
+++ b/src/XMakeBuildEngine/Definition/ToolsetReader.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Build.Evaluation
             ToolsetDefinitionLocations locations
             )
         {
-            PropertyDictionary<ProjectPropertyInstance> initialProperties =
+            var initialProperties =
                 new PropertyDictionary<ProjectPropertyInstance>(environmentProperties);
 
             initialProperties.ImportProperties(globalProperties);
@@ -184,7 +184,7 @@ namespace Microsoft.Build.Evaluation
                         if (v4Dir != null && !toolsets.ContainsKey("4.0"))
                         {
                             // Create standard properties. On Mono they are well known
-                            PropertyDictionary<ProjectPropertyInstance> buildProperties =
+                            var buildProperties =
                                 CreateStandardProperties(globalProperties, "4.0", libraryPath, v4Dir);
 
                             toolsets.Add(
@@ -209,24 +209,25 @@ namespace Microsoft.Build.Evaluation
                             {
                                 var version = Path.GetFileName(d);
                                 var binPath = Path.Combine(d, "bin");
-                                if (!toolsets.ContainsKey(version))
+                                if (toolsets.ContainsKey(version))
                                 {
-                                    // Create standard properties. On Mono they are well known
-                                    PropertyDictionary<ProjectPropertyInstance> buildProperties =
-                                        CreateStandardProperties(globalProperties, version, xbuildToolsetsDir, binPath);
-
-                                    toolsets.Add(
-                                        version,
-                                        new Toolset(
-                                            version,
-                                            binPath,
-                                            buildProperties,
-                                            environmentProperties,
-                                            globalProperties,
-                                            null,
-                                            currentDir,
-                                            string.Empty));
+                                    continue;
                                 }
+                                // Create standard properties. On Mono they are well known
+                                var buildProperties =
+                                    CreateStandardProperties(globalProperties, version, xbuildToolsetsDir, binPath);
+
+                                toolsets.Add(
+                                    version,
+                                    new Toolset(
+                                        version,
+                                        binPath,
+                                        buildProperties,
+                                        environmentProperties,
+                                        globalProperties,
+                                        null,
+                                        currentDir,
+                                        string.Empty));
                             }
                         }
                     }
@@ -254,7 +255,7 @@ namespace Microsoft.Build.Evaluation
             if (((locations & ToolsetDefinitionLocations.Registry) != 0) && !toolsets.ContainsKey("2.0")
                 && FrameworkLocationHelper.PathToDotNetFrameworkV20 != null)
             {
-                Toolset synthetic20Toolset = new Toolset(
+                var synthetic20Toolset = new Toolset(
                     "2.0",
                     FrameworkLocationHelper.PathToDotNetFrameworkV20,
                     environmentProperties,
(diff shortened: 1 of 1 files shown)


mfilippov:
I see problem in unit tests I try fix it.

rainersigwald:
Thanks! The I'm filing bugs for all of the not-related-to-your-change test failures.

Can you amend the GitVersioning commit so that it mentions it's a workaround for #694 (that I just filed)? That way we can be sure to unify the logic later.

rainersigwald:
(Sorry, you may have beat me to it!)

rainersigwald:
#695 is the Windows failure

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -204,7 +204,8 @@ ToolsetDefinitionLocations locations
 
                         // Other toolsets are installed in the xbuild directory
                         var xbuildToolsetsDir = Path.Combine(libraryPath, "xbuild");
-                        var r = new Regex(xbuildToolsetsDir + Path.DirectorySeparatorChar + @"\d+\.\d+");
+                        var regexPath = xbuildToolsetsDir + Path.DirectorySeparatorChar;
+                        var r = new Regex(regexPath.Replace(@"\", @"\\") + @"\d+\.\d+");
```
rainersigwald: Can you use `Regex.Escape` instead?

rainersigwald:
Fantastic, thanks! Looks good to me, once you've changed the commit message for the GitVersioning commit.

mfilippov:
Commit message changed.

mfilippov:
I little change GitVersioning workaround for include targets only if it exists.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -153,23 +153,26 @@ ToolsetDefinitionLocations locations
...
-
-                // We do not accumulate properties when reading them from the registry, because the order
-                // in which values are returned to us is essentially random: so we disallow one property
-                // in the registry to refer to another also in the registry
-                defaultToolsVersionFromRegistry = registryReader.ReadToolsets(toolsets, globalProperties,
-                    initialProperties, false /* do not accumulate properties */, out overrideTasksPathFromRegistry,
-                    out defaultOverrideToolsVersionFromRegistry);
-            }
+
+                if (NativeMethodsShared.IsWindows || registryReader != null)
```
radical: Is this runtime `IsWindows` check required here, since `FEATURE_WIN32_REGISTRY` should be enabled only for windows/full framework builds anyway? And line 163 (`registryReader = registryReader ?? new ToolsetRegistryReader(...` means that we don't need to check for `registryReader != null` either.

And so, the new set of braces being added here are not required.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -240,8 +243,9 @@ ToolsetDefinitionLocations locations
                         }
                     }
                 }
+#if FEATURE_WIN32_REGISTRY
```
radical: Because of the comment above, this should not be required either, just drop the extra brace.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -205,42 +205,45 @@ ToolsetDefinitionLocations locations
                         // Other toolsets are installed in the xbuild directory
                         var xbuildToolsetsDir = Path.Combine(libraryPath, "xbuild");
```
radical: Nitpick - maybe we can add the `+ Path.DirectorySeparatorChar` here, instead of later. So, it's use with `Regex.Escape` later becomes easier to read.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -205,42 +205,45 @@ ToolsetDefinitionLocations locations
...
+                                            buildProperties,
+                                            environmentProperties,
+                                            globalProperties,
+                                            null,
+                                            currentDir,
+                                            string.Empty));
+                                }
+                            }
 
+                            if (!toolsets.ContainsKey(MSBuildConstants.CurrentToolsVersion))
```
radical: Does this `if` block need to be under the directory check? IIUC, this should be run irrespective of the earlier block.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -204,7 +204,7 @@ ToolsetDefinitionLocations locations
 
                         // Other toolsets are installed in the xbuild directory
                         var xbuildToolsetsDir = Path.Combine(libraryPath, "xbuild");
-                        var r = new Regex(xbuildToolsetsDir + Path.DirectorySeparatorChar + @"\d+\.\d+");
+                        var r = new Regex(Regex.Escape(xbuildToolsetsDir + Path.DirectorySeparatorChar) + @"\d+\.\d+");
```
radical: Could you please squash this with the related commit in the `ToolsetReader`?

radical: Though, maybe it is good as a separate commit. Just add some reasoning for the change, like add the stack trace from the issue in the commit message here.

radical: And you could move that `var r = new Regex...` inside the directory.exists block.

radical:
Also, could you please add more information/explanation to the commit messages like for `Check folder exists` and `fix path escaping...` .

And thanks for your PR :)

src/XMakeBuildEngine/Definition/ToolsetReader.cs
mfilippov: If NativeMethodsShared.IsWindows but registryReader == null conditions is true and registryReader = registryReader ?? new ToolsetRegistryReader.. required. I little rewrite method for remove else. I commit updated version soon.

mfilippov:
All code review comments fixed.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
radical: Ah, my comment was incorrect. `FEATURE_WIN32_REGISTRY` will be enabled even for mono builds, which can run on non-windows platforms too. So, the runtime check `IsWindows` is useful here.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -157,89 +157,86 @@ ToolsetDefinitionLocations locations
             {
 #if FEATURE_WIN32_REGISTRY
                 if (NativeMethodsShared.IsWindows || registryReader != null)
```
radical: Since, this check is indeed required (https://github.com/Microsoft/msbuild/pull/693#discussion_r68432809), the only thing required to be fixed here is the opening brace for this `if` block. And just indent the corresponding code. Something like - https://gist.github.com/radical/f085ffb25f57b6fc179f785e4a47b930 .

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -117,7 +117,7 @@ internal static string ReadAllToolsets
             ToolsetDefinitionLocations locations
             )
         {
-            PropertyDictionary<ProjectPropertyInstance> initialProperties =
+            var initialProperties =
```
radical: Please don't add unrelated changes to the commit.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -157,89 +157,86 @@ ToolsetDefinitionLocations locations
...
-                                    string.Empty));
-                        }
+                    // Other toolsets are installed in the xbuild directory
+                    var xbuildToolsetsDir = Path.Combine(libraryPath, "xbuild");
+                    var r = new Regex(xbuildToolsetsDir + Path.DirectorySeparatorChar + @"\d+\.\d+");
+                    foreach (var d in Directory.GetDirectories(xbuildToolsetsDir).Where(d => r.IsMatch(d)))
+                    {
+                        var version = Path.GetFileName(d);
+                        var binPath = Path.Combine(d, "bin");
+                        if (toolsets.ContainsKey(version))
```
radical: This change here was easy to miss, because of the big diff and most of it being just "indentation change". This commit is fixing the "merge" issue, so, it should not contain these unrelated changes. It can be fixed in the separate commit where you are fixing the Regex issue.

src/XMakeBuildEngine/Definition/ToolsetReader.cs
```diff
@@ -199,29 +199,32 @@ ToolsetDefinitionLocations locations
                     }
 
                     // Other toolsets are installed in the xbuild directory
-                    var xbuildToolsetsDir = Path.Combine(libraryPath, "xbuild");
-                    var r = new Regex(xbuildToolsetsDir + Path.DirectorySeparatorChar + @"\d+\.\d+");
-                    foreach (var d in Directory.GetDirectories(xbuildToolsetsDir).Where(d => r.IsMatch(d)))
+                    var xbuildToolsetsDir = Path.Combine(libraryPath, $"xbuild{Path.DirectorySeparatorChar}");
```
radical: Commit message should explain the reasoning for the changes. What is being changed can be seen in the commit itself, but the reason is important and can be useful in debugging or understanding it in future. So, something like this:

```
ToolsetReader: Escape paths before using them as part of a regex

We are trying to find toolset paths for mono, and using regex to match such
paths to a pattern. But this fails when running on windows and the path contains
`\` like `D:\j\workspace\_Windows_NT`, because `Regex` interprets it as a escape
sequence. It manifests as:

MSBUILD : error MSB1025: An internal failure occurred while running MSBuild.
11:59:04 System.ArgumentException: parsing 'D:\j\workspace\_Windows_NT_p---ab87db35\bin\xbuild\\d+\.\d+' - Unrecognized escape sequence \\j.
11:59:04    at System.Text.RegularExpressions.RegexParser.ScanCharEscape()
11:59:04    at System.Text.RegularExpressions.RegexParser.ScanBasicBackslash()
11:59:04    at System.Text.RegularExpressions.RegexParser.ScanRegex()
11:59:04    at System.Text.RegularExpressions.RegexParser.Parse(String re, RegexOptions op)
11:59:04    at System.Text.RegularExpressions.Regex..ctor(String pattern, RegexOptions options, TimeSpan matchTimeout, Boolean useCache)
11:59:04    at System.Text.RegularExpressions.Regex..ctor(String pattern)
11:59:04    at Microsoft.Build.Evaluation.ToolsetReader.ReadAllToolsets(Dictionary`2 toolsets, PropertyDictionary`1 environmentProperties, PropertyDictionary`1 globalProperties, ToolsetDefinitionLocations locations)
11:59:04    at Microsoft.Build.Evaluation.ProjectCollection.InitializeToolsetCollection()
11:59:04    at Microsoft.Build.Evaluation.ProjectCollection..ctor(IDictionary`2 globalProperties, IEnumerable`1 loggers, IEnumerable`1 remoteLoggers, ToolsetDefinitionLocations toolsetDefinitionLocations, Int32 maxNodeCount, Boolean onlyLogCriticalEvents)
11:59:04    at Microsoft.Build.CommandLine.MSBuildApp.BuildProject(String projectFile, String[] targets, String toolsVersion, Dictionary`2 globalProperties, ILogger[] loggers, LoggerVerbosity verbosity, DistributedLoggerRecord[] distributedLoggerRecords, Int32 cpuCount, Boolean enableNodeReuse, TextWriter preprocessWriter, Boolean debugger, Boolean detailedSummary)
11:59:04    at Microsoft.Build.CommandLine.MSBuildApp.Execute(String[] commandLine)

Use `Regex.Escape` for the path portion of the pattern.

Also, see https://github.com/Microsoft/msbuild/issues/695
```

AndyGerlicher:
LGTM, but can you drop your commit about the Git Versioning? #721 should solve that issue for now.

mfilippov:
@AndyGerlicher Git Versioning commit dropped, @radical I try rearrange PR base on you comments.

radical:
If the ToolsetReader commits are being retained in this PR itself, then do update the PR commit message to indicate that this is more than just build fix. Other than that, LGTM!

rainersigwald:
Great, thank you @mfilippov!

PR #702, opened 2016-06-16
Calculate a better MSBuild tools/extensions path.

AndyGerlicher:
Add a hook into the intrinsic functions to get the full path to the
MSBuild tools path and VS install root for the extensions path variables.

This is blocking some tests internally because the "." is getting read and evaluated to a different path. I also think it given that this logic is now somewhat more complicated that we should do something like this to calculate the paths rather than limit to a single line property evaluation as it was (I never liked the `..\..\..\` or `.` appearing in the variable).

Note: I'm looking for feedback and ideas on where to put this code, this is not complete. I may commit this to unblock the tests internally, but either way this needs to be updated.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
+++ b/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
@@ -349,6 +349,31 @@ namespace Microsoft.Build.Evaluation
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


AndyGerlicher:
Created #704 to address this.

rainersigwald:
LGTM (as temporary fix)

PR #706, opened 2016-06-17
Rename setup package ID in files.swr

AndyGerlicher:
Missed this in my rename of the setup package. This doesn't get caught until it tries to insert the package in the VS setup build.


rainersigwald:
LGTM

PR #710, opened 2016-06-20
Adding ManifestUtil string resources to MS.Build.Tasks

AndyGerlicher:
- This should have been embedded in Microsoft.Build.Tasks to match the
  shipping product.


rainersigwald:
:shipit:

PR #717, opened 2016-06-21
Initial xlf files

cdmihai:
We're changing the way we obtain localized builds of MSBuild. We will be storing localized resources in xlf files and keep them in sync with the resx files. This means that the community can contribute to translation.

I'm sending the xlf files ahead of the sync build code so the localization team can begin translating in advance


rainersigwald:
LGTM, assuming it's ok that everything's currently marked as "new"--I take it the loc team already has a plan to drag in the existing translations from their internal home?

cdmihai:

cdmihai:

Issue #704, opened 2016-06-16
Calculate a better MSBuild tools/extensions path


AndyGerlicher:
Need to fix the [FileUtilities](https://github.com/Microsoft/msbuild/blob/master/src/Shared/FileUtilities.cs) class and extract logic to a new class. These are the scenarios I want to consider.

## Tools / Extensions Paths
- When running in MSBuild.exe:
  - Set the Tools Path to current directory (expanded, not using `"."`).
  - Assume we're running in `<Root>\MSBuild\15.0\Bin` and choose `<Root>\MSBuild` for Extensions Path (`ToolsPath\..\..`).
- Running as a dependency and that folder is "complete":
  - Use the reference folder location as the Tools Path.
  - Assume we're running in `<Root>\MSBuild\15.0\Bin` and choose `<Root>\MSBuild` for Extension (`ToolsPath\..\..`).
- Running as a dependency that is not "complete":
  - Discard that folder and try to fallback to a well known location. 
  - For Dev15 (`VisualStudioVersion >= 15'`), this would be `%VSINSTALLDIR%\MSBuild\15.0\Bin` for Tools Path and  `%VSINSTALLDIR%\MSBuild` for Extensions Path.

_Note: By "complete" I mean contains enough of MSBuild to do a build. I will probably check for msbuild.exe and msbuild.exe.config as a minimum bar. This will cover the case that you create an app that references Microsoft.Build.*.dll and it gets copied to your output. Your output should not be considered a tools path unless it contains those files._

### If `%VSINSTALLDIR%` doesn't exist

- We could throw an Exception and tell the user to run from a Visual Studio command prompt.? We no longer have the ability to determine a valid location.
- Use global Visual Studio API to determine a location for Visual Studio (there may be multiple) and pick one of them.

### Running Open Source Build

- We don't want to pick up the Visual Studio Tools Path or Extensions Path in this scenario. Note that Extensions Path would be broken by the above logic because our build layout is like `bin\x86\Windows_NT\Debug\`. The Extensions Path would point to `x86` or something similarly wrong.
- This isn't really solvable without producing a build layout in the right format and pulling in all our run-time dependencies (what the bootstrap proj does now).

This is related to #702 which is a temporary workaround.

Sarabeth-Jaffe-Microsoft:
Related to: #114

PR #718, opened 2016-06-21
Workaround for issue causing ToolsPath character escaping

AndyGerlicher:
Related to #704
When referring to a function in a property definition MSBuild will escape
the value. This was causing the MSBuildToolsPath to contain escape
characters since the value is used directly. Referring to an already
defined property (VsInstallRoot) will workaround this issue.


AndyGerlicher:
This scenario needs to be addressed in #704. `MSBuildToolsPath` (and similar) should not be escaped when set.

rainersigwald:
LGTM (as an ugly workaround)

PR #721, opened 2016-06-22
Only include GitVersioning when it exists

rainersigwald:
Extracting this from #693, because it's causing us to fail to build NuGet packages in our (internal) build that produces them. On my machine, this can build after clean when invoked like the build definition does (with machine MSBuild).

Thanks for the fix, @mfilippov!

### DIFF
--- a/src/dir.targets
+++ b/src/dir.targets
@@ -50,8 +50,9 @@
     <StartAction>Program</StartAction>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


rainersigwald:
@AndyGerlicher After we talked, I now lean toward just clearly disabling this entirely, then coming back in to fix it.

mfilippov:
I fix all comments in my PR #693

AndyGerlicher:
LGTM

Issue #722, opened 2016-06-23
TestGetPathToBuildToolsFile_XXBit tests fail on netcore


rainersigwald:
This is in xplat. Looks like the problem is that the magic property `MSBuildToolsPath32` isn't getting set, so it throws on getting its `EvaluatedValue`.

```
   Microsoft.Build.UnitTests.ToolLocationHelper_Tests.TestGetPathToBuildToolsFile_64Bit [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object.
      Stack Trace:
            at Microsoft.Build.UnitTests.ToolLocationHelper_Tests.TestGetPathToBuildToolsFile_64Bit()
   Microsoft.Build.UnitTests.ToolLocationHelper_Tests.TestGetPathToBuildToolsFile_32Bit [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object.
      Stack Trace:
            at Microsoft.Build.UnitTests.ToolLocationHelper_Tests.TestGetPathToBuildToolsFile_32Bit()
```

PR #723, opened 2016-06-23
Merge #611 to xplat

rainersigwald:


src/Shared/FrameworkLocationHelper.cs
```diff
@@ -1188,6 +1180,26 @@ private static void RedirectVersionsIfNecessary(ref Version dotNetFrameworkVersi
             }
         }
 
+#if FEATURE_SYSTEM_CONFIGURATION
```
rainersigwald: Had to `#if` this out because of lack of System.Configuration--but all of its callsites were already out.

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -951,8 +974,8 @@ public void GetPathToWindowsSdk()
                                     }
   ";
 
-        [Fact(Skip = "Ignored in MSTest")]
-        // Ignore: Test requires installed toolset.
+#if FEATURE_CODETASKFACTORY
```
rainersigwald: This was just moved from the next test case, we didn't notice the need for it before because it was skipped.

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -593,9 +593,7 @@ public void TestGetPathToBuildToolsFile()
             Assert.Equal(tv12path, ToolLocationHelper.GetPathToBuildToolsFile("MSBuild.exe", ToolLocationHelper.CurrentToolsVersion));
         }
 
-        [Fact(Skip = "Ignored in MSTest")]
-
-        // Ignore: Test requires installed toolset.
+        [Fact(Skip = "https://github.com/Microsoft/msbuild/issues/722")]
```
rainersigwald: Added the ignore in a separate commit so it's easily revertable once #722 is understood.

src/dirs.proj
```diff
@@ -7,6 +7,8 @@
     <Project Include="Utilities\Microsoft.Build.Utilities.csproj" />
     <Project Include="XMakeBuildEngine\Microsoft.Build.csproj" />
     <Project Include="XMakeCommandLine\MSBuild.csproj" />
+    <Project Include="XMakeCommandLine\MSBuildTaskHost\MSBuildTaskHost.csproj"
+             Condition="'$(FullFrameworkBuild)' == 'true'" />
```
rainersigwald: Condition make sense, at least for now?

rainersigwald:
@AndyGerlicher CAn you take a look? The only substantial change here should be from you :)

src/dirs.proj
AndyGerlicher: Seems fine for now. The only thing I worry about is the `CLR2COMPATIBILITY` const to build some of the sources in that mode. I hope we didn't diverge too far and it becomes a huge pain to get things to build for 3.5. And it would be good if we refactor for that to break sooner than later.

AndyGerlicher:
LGTM

Issue #694, opened 2016-06-15
Ensure versioning works for all host runtimes


rainersigwald:
In master, we're using Nerdbank.GitVersioning, which works great on full framework. It should support running on .NET Core soon, and we should move to that version and make sure it works everywhere.

rainersigwald:
This one's interesting in xplat. Right now it's only working because of a confluence of bugs.

When I merged the adoption of GitVersioning in, I accidentally broke the restore-build-time-packages code, so that it never runs (because it's [still looking for the old packages.config](https://github.com/Microsoft/msbuild/blob/2001dcb8afab050d017acdf411bcc82e0a952c82/src/.nuget/packageLoad.targets#L41)).

That keeps the target file from ever existing and made the fix from 5fd4249 equivalent to just disabling it--which is fine for now, since this issue is to track making sure it works everywhere.

The reason this didn't break in CI is because xplat only builds for netcore in CI, and the "standard" build uses an old netcore MSBuild to bootstrap, so the old condition always evaluates to "don't bother with GitVersioning".

rainersigwald:
For ease of future checks, this is dependent on https://github.com/AArnott/Nerdbank.GitVersioning/issues/56

rainersigwald:
Removing from the great-merge milestone: it's sufficient for now to produce nicely-versioned packages on Windows. Our official build will run only on Windows for the foreseeable future (because we need Microsoft-internal signing steps), so we can wait indefinitely for the nice-to-have "also version correctly on other OSes".

AArnott:
Nerdbank.GitVersioning has a 'dotnet' compatible version now. [Details here](https://github.com/AArnott/Nerdbank.GitVersioning/issues/56#issuecomment-280549760)

PR #726, opened 2016-06-24
Update Framework hardcoded version to 15.1

rainersigwald:
Workaround for #694--the version shouldn't be hardcoded at all, but
stamped in by the build process.


AndyGerlicher:
:shipit:

PR #229, opened 2015-09-23
Remove unused usings

rainersigwald:
We had an awful lot of these.

@AndyGerlicher if you're more worried about churn we can table this.  It shouldn't really be hurting anything.

PR #730, opened 2016-06-27
fixes of mistakes and warnings

DamirAinullin:
Some fixes of mistakes and warnings


rainersigwald:
@DamirAinullin Thanks for taking the time to do this cleanup! It looks good and is definitely a direction we'd like to go in (those grayed-out usings bug me all the time!) but we're holding off on this kind of refactoring at the moment while we try to bring the .NET Core and cross-platform work back into the mainline of our codebase. After we've accomplished the [Great Merge](https://github.com/Microsoft/msbuild/milestones/The%20Great%20Merge) and have a single development branch, we'd love for you to propose this patch again!

Hopefully it won't be many months before we can do this, as it has been since #229!

PR #737, opened 2016-06-27
Update nuspecs

rainersigwald:


AndyGerlicher:
LGTM

src/.nuget/packageLoad.targets
```diff
@@ -38,31 +38,17 @@
   <Target Name="_RestoreBuildToolsPackagesConfig"
```
cdmihai: Should this target have a dual xplat / master behaviour? On xplat it uses CLI and project.json, on master it uses packages.config. Or should we just pick the CLI from buildtools for both?

cdmihai:
Don't know if this worth doing now but: As long as we still have that nuspec reference generator still working, it might be worth having it delete unused nuspec references, if there are any. To do that, you delete all references except the p2p msbuild ones, and run the build. The reference generator should add back only the stuff that's required.

src/.nuget/packageLoad.targets
rainersigwald: In the glorious future, we'll just do "dotnet restore". For now, I'm just letting this file diverge between xplat and master.

rainersigwald:
@cdmihai Ok, I'll do that instead of this second commit. Just a sec.

rainersigwald:
@cdmihai Totally worth doing now since it caught two stale references that still pointed to non-RTM packages. Good catch!

cdmihai:
:shipit::

PR #742, opened 2016-06-28
Copy task support for symlinks

### LINKED ISSUE #662: Copy task support for symlinks
This is a feature request from an internal Microsoft team. `Copy` currently supports hardlinks via `UseHardlinksIfPossible`. It could also support symlinks.

On Windows, this should probably have a nice error message when the build is run unelevated and the user doesn't have `SECreateSymbolicLinkPrivilege`.

maddin2016:
fixed #662 

Add support to use symbolic links in copy tasks


src/XMakeTasks/Strings.resx
```diff
@@ -2671,6 +2678,7 @@
             MSB3861 - MSB3870   Task: ErrorFromResources
             MSB3871 - MSB3880   Targets: Code sharing
             MSB3881 - MSB3890   Task: Compiler
+            MSB3891 - MSB3900   Task: Copy
```
rainersigwald: Say "overflow" here for consistency with others above.

rainersigwald:
@dotnet-bot test this please

src/XMakeTasks/UnitTests/Copy_Tests.cs
```diff
@@ -1845,6 +1846,46 @@ public CopyNotHardLink_Tests()
         }
     }
 
+    public class CopyHardAndSymbolicLink_Tests : Copy_Tests
```
rainersigwald: Does this need to inherit from `Copy_Tests`? It contains only a single test and doesn't need to run all of the copy tests in a different configuration.

rainersigwald:
@maddin2016 Looks like the test failure is related:

```
MESSAGE:
Assert.Equal() Failure\r\n ↓ (pos 9)\r\nExpected: This is a source temp file.\r\nActual: This is another source temp file.\r\n ↑ (pos 9)
+++++++++++++++++++
STACK TRACE:
at Microsoft.Build.UnitTests.CopySymbolicLink_Tests.CopyToDestinationFolderWithSymbolicLinkCheck() in D:\j\workspace\_Windows_NT_p---ab87db35\src\XMakeTasks\UnitTests\Copy_Tests.cs:line 2232
```

src/XMakeTasks/UnitTests/Copy_Tests.cs
```diff
@@ -2120,4 +2161,83 @@ public void CopyToDestinationFolderWithHardLinkFallbackTooManyLinks()
...
+                // we'll then check that the destination file automatically
+                // has the same content (i.e. it's been hard linked)
+                using (StreamWriter sw = new StreamWriter(sourceFile, false))    // HIGHCHAR: Test writes in UTF8 without preamble.
+                    sw.Write("This is another source temp file.");
+
+                // Read the destination file (it should have the same modified content as the source)
+                using (StreamReader sr = new StreamReader(destFile))
+                    destinationFileContents = sr.ReadToEnd();
+
+                Assert.Equal(destinationFileContents, "This is another source temp file."); //                     "Expected the destination hard linked file to contain the contents of source file. Even after modification of the source"
```
rainersigwald: While you're in here, can you add a commit that swaps the order of these arguments? It should be `expected, actual`.  Please do so above in the hardlink test as well.

src/XMakeTasks/UnitTests/Copy_Tests.cs
```diff
@@ -2120,4 +2161,83 @@ public void CopyToDestinationFolderWithHardLinkFallbackTooManyLinks()
...
+
+                ((MockEngine)t.BuildEngine).AssertLogDoesntContain("MSB3026"); // Didn't do retries
```
rainersigwald: This should check for the new MSB3891, right?

rainersigwald:
Pretty sure the test is failing because the CI user doesn't have admin/`SECreateSymbolicLinkPrivilege` rights. We need to handle that case gracefully. But I'm not sure what the right way is--we definitely want to be able to pass tests when running as a normal user, but we also don't want to let symlinks get silently broken just because we're all low-priv users and only test that way. Do you have any thoughts on that?

maddin2016:
Maybe we should check for error `80070522` and let this test fail with an error message that these test has to be run with higher privileges.

maddin2016:
it's a little bit tricky, but whatever :sweat_smile:
```
var needHigherPrivileges = me.Log.Contains("0x80070522");

if (needHigherPrivileges)
         Assert.False(needHigherPrivileges, "It seems that you don't have the permission to create symbolic links. Try to run this test again with higher privileges");
```

maddin2016:
Hi @rainersigwald. Are there any updates on this?

src/XMakeTasks/UnitTests/Copy_Tests.cs
rainersigwald: Looking at this again, was I wrong? We do want to assert that there were no retries . . .

---
In reply to: [69204320](https://github.com/Microsoft/msbuild/pull/742#discussion_r69204320) [](ancestors = 69204320)

rainersigwald:
Sorry for the delay. I'm ok with explicitly checking for that HRESULT--but the action taken if you see it should be that the test passes. I think that's the best we can do without `Assert.Inconclusive`. Please reference #763 in a comment around this code and the commit itself so we know we can remove this workaround once we have a better solution.

If we're just looking through the log for the HRESULT, the cross-platform implications should be minimal, which is nice.

Can you also please add a test case that checks for the MSB3891 error when both hard- and symlinks are requested?

This is looking good!

src/XMakeTasks/UnitTests/Copy_Tests.cs
```diff
@@ -2202,36 +2202,45 @@ public void CopyToDestinationFolderWithSymbolicLinkCheck()
 
                 bool success = t.Execute();
 
-                Assert.True(success); // "success"
-                Assert.True(File.Exists(destFile)); // "destination exists"
-                Microsoft.Build.UnitTests.MockEngine.GetStringDelegate resourceDelegate = new Microsoft.Build.UnitTests.MockEngine.GetStringDelegate(AssemblyResources.GetString);
+                var needHigherPrivileges = me.Log.Contains("0x80070522");
 
-                me.AssertLogContainsMessageFromResource(resourceDelegate, "Copy.SymbolicLinkComment", sourceFile, destFile);
+                if (needHigherPrivileges)
```
maddin2016: @rainersigwald, is this what you mean with `test passes`??

maddin2016:
Hi @rainersigwald, is [](url) what you mean with `test passes`.

maddin2016:
Oops :see_no_evil: wrong button

src/XMakeTasks/UnitTests/Copy_Tests.cs
```diff
@@ -1845,6 +1846,46 @@ public CopyNotHardLink_Tests()
         }
     }
 
+    public class CopyHardAndSymbolicLink_Tests
+    {
+        [Fact]
+        public void CopyWithHardAndSymbolicLinks()
```
maddin2016: Here is the test which simulate when create choose sym and hardlinks.

maddin2016:
Hi @rainersigwald. Any updates on this?

src/XMakeTasks/UnitTests/Copy_Tests.cs
rainersigwald: Yes, that'll work.

---
In reply to: [70453742](https://github.com/Microsoft/msbuild/pull/742#discussion_r70453742) [](ancestors = 70453742)

src/XMakeTasks/Strings.resx
```diff
@@ -281,6 +285,9 @@
     <value>MSB3030: Could not copy the file "{0}" because it was not found.</value>
     <comment>{StrBegin="MSB3030: "} LOCALIZATION: {0} is a number.</comment>
   </data>
+  <data name="Copy.ExactlyOneTypeOfLink">
+    <value>MSB3891: Both "{0}" and "{1}" were specified in the project file. Please choose one or the other.</value>
```
rainersigwald: Put this way down the file in order, for consistency.

src/XMakeTasks/UnitTests/Copy_Tests.cs
rainersigwald: Still worried about this. I guess really there should be _neither_ `MSB3891` nor `MSB3026`.

---
In reply to: [70086908](https://github.com/Microsoft/msbuild/pull/742#discussion_r70086908) [](ancestors = 70086908,69204320)

rainersigwald:
Sorry about the delay! I think a couple of minor tweaks are all that's left.

@Microsoft/msbuild-maintainers can someone else take a look?

maddin2016:
Sounds good. I will fix your comments ASAP

src/XMakeTasks/UnitTests/Copy_Tests.cs
maddin2016: Do you mean we should remove this line? Can you please explain me the `MockEngine` `BuildEngine` a little bit more. Is this just a wrapper to log in unit tests?

src/XMakeTasks/Copy.cs
```diff
@@ -323,20 +332,52 @@ private void LogDiagnostic(string message, params object[] messageArgs)
...
-                if (!hardLinkCreated)
+                if (!linkCreated)
                 {
                     int errorCode = Marshal.GetHRForLastWin32Error();
                     Exception hardLinkException = Marshal.GetExceptionForHR(errorCode);
                     // This is only a message since we don't want warnings when copying to network shares etc.
                     Log.LogMessageFromResources(MessageImportance.Low, "Copy.RetryingAsFileCopy", sourceFileState.Name, destinationFileState.Name, hardLinkException.Message);
                 }
             }
+            else if (UseSymboliclinksIfPossible)
```
cdmihai: The symlink branch looks identical to the hardlink branch. Can you please factor out the common stuff? Something like TryCopyViaLink which receives some strings and lambdas

maddin2016: Hi @cdmihai. Something like this?
```c#
 // If we want to create hard or symbolic links, then try that first
            if (UseHardlinksIfPossible || UseSymboliclinksIfPossible)
            {
                // Do not log a fake command line as well, as it's superfluous, and also potentially expensive
                Log.LogMessageFromResources(MessageImportance.Normal, true == UseHardlinksIfPossible ? "Copy.HardLinkComment" : "Copy.SymbolicLinkComment", sourceFileState.Name, destinationFileState.Name);

                if (!_overwriteReadOnlyFiles)
                {
                    destinationFileExists = destinationFileState.FileExists;
                }

                // CreateHardLink and CreateSymbolicLink cannot overwrite an existing file or link
                // so we need to delete the existing entry before we create the hard or symbolic link.
                // We need to do a best-effort check to see if the files are the same
                // if they are the same then we won't delete, just in case they refer to the same
                // physical file on disk.
                // Since we'll fall back to a copy (below) this will fail and issue a correct
                // message in the case that the source and destination are in fact the same file.
                if (destinationFileExists && !IsMatchingSizeAndTimeStamp(sourceFileState, destinationFileState))
                {
                    FileUtilities.DeleteNoThrow(destinationFileState.Name);
                }

                linkCreated = true == UseHardlinksIfPossible? NativeMethods.CreateHardLink(destinationFileState.Name, sourceFileState.Name, IntPtr.Zero /* reserved, must be NULL */) : NativeMethods.CreateSymbolicLink(destinationFileState.Name, sourceFileState.Name, SymbolicLink.File);

                if (!linkCreated)
                {
                    int errorCode = Marshal.GetHRForLastWin32Error();
                    Exception linkException = Marshal.GetExceptionForHR(errorCode);
                    // This is only a message since we don't want warnings when copying to network shares etc.
                    Log.LogMessageFromResources(true == UseHardlinksIfPossible ? MessageImportance.Low : MessageImportance.High, "Copy.RetryingAsFileCopy", sourceFileState.Name, destinationFileState.Name, linkException.Message);
                }
            }
```

cdmihai: I was actually thinking of an extract method with some params to disambiguate between hard and sym links

AndyGerlicher:
Thanks again for the contribution! LGTM pending some what @cdmihai wanted and you need to rebase/squash your commits before we merge please.

src/XMakeTasks/Copy.cs
```diff
@@ -298,45 +307,17 @@ private void LogDiagnostic(string message, params object[] messageArgs)
...
 
-            // If we want to create hard links, then try that first
+            // If we want to create hard or symbolic links, then try that first
             if (UseHardlinksIfPossible)
-            {
-                // Do not log a fake command line as well, as it's superfluous, and also potentially expensive
-                Log.LogMessageFromResources(MessageImportance.Normal, "Copy.HardLinkComment", sourceFileState.Name, destinationFileState.Name);
+                TryCopyViaLink("Copy.HardLinkComment", MessageImportance.Low, sourceFileState, destinationFileState, ref destinationFileExists, ref linkCreated, () => NativeMethods.CreateHardLink(destinationFileState.Name, sourceFileState.Name, IntPtr.Zero /* reserved, must be NULL */));
+            else if (UseSymboliclinksIfPossible)
+                TryCopyViaLink("Copy.SymLinkComment", MessageImportance.High, sourceFileState, destinationFileState, ref destinationFileExists, ref linkCreated, () => NativeMethods.CreateSymbolicLink(destinationFileState.Name, sourceFileState.Name, SymbolicLink.File));
```
cdmihai: >destinationFileState.Name, sourceFileState.Name [](start = 200, length = 47)

You can turn destinationFileState.name and sourceFileState.name from the lambda into parameters, since you are passing them both into TryCopyViaLink. The function will become Func(string, string, bool). Also, can you please invert the order of the lambda params so it's source and then destination. It's confusing if one method has them in one order, and another in a different order.

cdmihai: >MessageImportance.High [](start = 54, length = 22)

Looks like sym links use a high message importance and hard links a low one. Is there a specific reason for this difference? Can't it be the same for both?

maddin2016: I have set the priority to high that you see the error code if you have not the privileges to create symbolic links. I have wonder the first time why it have not create symbolic links because i have read on the error message. here is what i mean
`Low`
image

`Normal`
image

`High`
image

cdmihai: OK, High verbosity looks much better for that error. Let's make it High for both hard and sym links then :)

cdmihai:
Looks good to go, after:
- you set high verbosity when both hard / sym links could not be made
- rebase on top of master. Apparently there are some conflicts
- CI passes

maddin2016:
Again @rainersigwald and @cdmihai many thanks for your help!!

cdmihai:
Looks like the failing test might be related to your changes

maddin2016:
mmhh, on my local machine all tests passes :confused:
image

cdmihai:
@dotnet-bot test this please

cdmihai:
@maddin2016 The CI test runs most likely do not run elevated. Can you check that case locally too?

maddin2016:
Hi @cdmihai, the tests locally on my machine runs elevated. I will test this without elevated privileges. But since my last update from master branch there are many tests which are failing. :confused: So i have to look whats going wrong here.

cdmihai:
@dotnet-bot test this please

maddin2016:
image
All these failing tests have the same error message. But myaybe it's a problem locally on my machine.

rainersigwald:
@maddin2016 That looks like something may have mis-merged with #776, that introduced `BuildEnvironmentHelperSingleton`? Can you get a callstack of the type initialization failure?

maddin2016:
Hi @rainersigwald, i got two different failures on two different machines. At one machine which i usually use all the time i get the following error
```
Testname:	Microsoft.Build.UnitTests.BackEnd.TaskRegistry_Tests.RetrieveFromCacheMatchingTaskDoesNotExist_FuzzyMatch
Test FullName:	Microsoft.Build.UnitTests.BackEnd.TaskRegistry_Tests.RetrieveFromCacheMatchingTaskDoesNotExist_FuzzyMatch
Testquelle:	C:\Users\scholzmartin\Documents\GitHub\msbuild\src\XMakeBuildEngine\UnitTests\BackEnd\TaskRegistry_Tests.cs Zeile 510
Testergebnis:	Fehlgeschlagene
Testdauer:	0:00:00,001

Ergebnis StackTrace:	
----- Inner Stack Trace #1 (System.Exception) -----
   bei Microsoft.Build.UnitTests.BackEnd.TaskAssemblyFixture.GetTestTaskAssemblyLocation() in C:\Users\scholzmartin\Documents\GitHub\msbuild\src\XMakeBuildEngine\UnitTests\BackEnd\TaskRegistry_Tests.cs:Zeile 112.
   bei Microsoft.Build.UnitTests.BackEnd.TaskAssemblyFixture..ctor() in C:\Users\scholzmartin\Documents\GitHub\msbuild\src\XMakeBuildEngine\UnitTests\BackEnd\TaskRegistry_Tests.cs:Zeile 38.
----- Inner Stack Trace #2 (Xunit.Sdk.TestClassException) -----
Ergebnis Meldung:	
System.AggregateException : Mindestens ein Fehler ist aufgetreten.
---- System.Exception : CSC task failed creating custom task: 
---- The following constructor parameters did not have matching fixture data: TaskAssemblyFixture fixture

```
So all that backend tests fail. Plus some other escape tests
```
Testname:	Microsoft.Build.UnitTests.EscapingInProjects_Tests.FullProjectsUsingMicrosoftCommonTargets.DollarSignInAssemblyName_UsingTaskHost
Test FullName:	Microsoft.Build.UnitTests.EscapingInProjects_Tests.FullProjectsUsingMicrosoftCommonTargets.DollarSignInAssemblyName_UsingTaskHost
Testquelle:	C:\Users\scholzmartin\Documents\GitHub\msbuild\src\XMakeBuildEngine\UnitTests\EscapingInProjects_Tests.cs Zeile 1186
Testergebnis:	Fehlgeschlagene
Testdauer:	0:00:00,074

Ergebnis StackTrace:	
bei Xunit.Assert.True(Nullable`1 condition, String userMessage) in C:\BuildAgent\work\cb37e9acf085d108\src\xunit.assert\Asserts\BooleanAsserts.cs:Zeile 90.
   bei Xunit.Assert.True(Boolean condition) in C:\BuildAgent\work\cb37e9acf085d108\src\xunit.assert\Asserts\BooleanAsserts.cs:Zeile 57.
   bei Microsoft.Build.UnitTests.ObjectModelHelpers.BuildTempProjectFileWithTargetsExpectSuccess(String projectFileRelativePath, String[] targets, IDictionary`2 additionalProperties) in C:\Users\scholzmartin\Documents\GitHub\msbuild\src\Shared\UnitTests\ObjectModelHelpers.cs:Zeile 791.
   bei Microsoft.Build.UnitTests.ObjectModelHelpers.BuildTempProjectFileExpectSuccess(String projectFileRelativePath) in C:\Users\scholzmartin\Documents\GitHub\msbuild\src\Shared\UnitTests\ObjectModelHelpers.cs:Zeile 780.
   bei Microsoft.Build.UnitTests.EscapingInProjects_Tests.FullProjectsUsingMicrosoftCommonTargets.DollarSignInAssemblyName_UsingTaskHost() in C:\Users\scholzmartin\Documents\GitHub\msbuild\src\XMakeBuildEngine\UnitTests\EscapingInProjects_Tests.cs:Zeile 1228.
Ergebnis Meldung:	
Assert.True() Failure
Expected: True
Actual:   False

```
And on the other machine i got that failure below. I send the trace later. But to come back to the originally question from @cdmihai. 
>The CI test runs most likely do not run elevated. Can you please check that case locally too?

Yes, if i run the test not elevated, it fails for me too.

rainersigwald:
@maddin2016 Ah, I see that same failure when running tests in Visual Studio--but not on the command line. Can you try running `RebuildWithLocalMSBuild.cmd` from both elevated and unelevated command prompts to try to repro the test failure?

maddin2016:
yeah, `RebuildWithLocalMSBuild.cmd` works fine. From not elevated command prompt that unit test fail and with elevated command prompt it passes.

maddin2016:
One more thing, i have to install .net 3.5 to run successfully. Is that right?
image

rainersigwald:
@maddin2016 Yes, 3.5 is required for a full test pass. We have special functionality (`MSBuildTaskHost`) to support it that is tested.

It's interesting that the test fails and doesn't fall into the detect-error condition you put in. Maybe explicitly check for `SECreateSymbolicLinkPrivilege` and early exit if not present?

maddin2016:
Hi @rainersigwald, thats because it throws another error
```
{System.Runtime.InteropServices.COMException (0x800703F0): Es wurde versucht, auf ein Token zuzugreifen, das nicht vorhanden ist. (Ausnahme von HRESULT: 0x800703F0)
```
???

maddin2016:
Hi @rainersigwald, in my last commit i have add a method to check for `SECreateSymbolicLinkPrivilege`. 
Thanks to [this post](https://bytes.com/topic/c-sharp/answers/441340-gettokeninformation). Is there any easier way? If not, where we can place this. Under `NativeMethods`?

rainersigwald:
Yikes, that _is_ ugly. But it's the pedantically correct thing to do. It would be cleaner to just use `principal.IsInRole(WindowsBuiltInRole.Administrator)`--but less correct. If Windows ever joins the rest of the world in allowing symlink creation as a normal user, I think this code will just Do The Right Thing.

My instinct is to say to do it this way (since you already figured it out!) and leave the code in the test since it's only used there. @cdmihai @AndyGerlicher thoughts?

cdmihai:
LGTM. On minor nit, can you please extract out the permission checking code outside of the unit test method? That will allow other potential symlink tests from that file to reuse it, and it will make the intent of the actual test more clear.

Beyond that, looks ready to merge.

maddin2016:
>can you please extract out the permission checking code outside of the unit test method? 
Into `Copy_Tests.cs` or some where else?

rainersigwald:
@maddin2016 I think a separate method in this file is fine. If we start using it elsewhere we can move it to a common place.

rainersigwald:
LGTM

cdmihai:
Thank you @maddin2016 for your contribution!

rainersigwald:
Thank you @maddin2016! 🎉

maddin2016:
:tada: :+1:

maddin2016:
What i though want to ask, what is to do for the xplat branch? Should i create a new request? And how you test on mac, because i don't have one.

rainersigwald:
@maddin2016 It would be fantastic if you could merge f71174c into xplat and submit a PR--though you need to be aware of #922 which will need to go in first.

If you can get it to work on Linux, one of us can test on our Macs to help with the final stretch.

maddin2016:
That sounds good. To make it clearly to me. I merge [f71174c](https://github.com/Microsoft/msbuild/commit/f71174cc2a9d025ee3d78cad63dd4f9c8cef5116) into xplat. Wait for #922 to be merged, and then begin to work on this and submit a PR. Is this correct?

rainersigwald:
I'd wait for #922 first, then merge, work, submit. And please submit the PR early (even when it's not working/done), so we know you're working on it.

AndyGerlicher:
Also just so you're aware I filed #928. This change unfortunately breaks all Xamarin iOS (at least) targets. No idea if this would be more widespread or not (overriding our `Copy` task wouldn't really be a recommended thing to do!). There's a potential of having to back out this change in master since we're inserting into VS and don't want to break that scenario. Just a heads up.

maddin2016:
No good news :hushed:

PR #749, opened 2016-06-29
Split assembly search paths in two halves

simonthum:
The rationale is to enable reproducible builds by cancelling assembly resolution
sources that are not usually controlled by the project to build.

In particular, I find the OutDir to be a highly unexpected source. It's nice as a gauge of developer competency, but down on earth, it's not really helpful in a reproducible build setting, especially if MSBuild is not the only such tool.

This is actually more of a feature than a pull request since I did not  test it but I trust that you won't allow changes in that file which didn't pass a number of eyes and robots.


simonthum:
I guess I don't need to break it but this would be a good countermeasure for a number of bugs, e.g. #173 and related, at least for those who keep their house in order.

Sarabeth-Jaffe-Microsoft:
Team Triage: @simonthum this is an interesting pull request but it has widespread ramifications. Can you please open an issue explaining your rationale so we can discuss the change?

Thanks!
MSBuild Team

Issue #222, opened 2015-09-21
MSBuild should have a solution import in Microsoft.Common.Targets


Serivy:
My request is to be able to put enterprise level logic in a msbuild target at a solution level rather than modifying every single .*proj file. This way I can modify one file in the solution directory and all projects which implement Microsoft.Common.Targets can set things like $(OutDir), Reference Paths, Analyzers, Nuget restores/updates or even Build server logic (Strongname signing, digital signuates, assembly versions). It would be priceless to enforce logic on a solution without relying on developers to install the custom target file into the right directory. Currently I do this by modifying all the projects to import a solution wide target file.

There are more details on the uservoice page including other suggestions to change but it would essentially be adding `<Import Project="$(SolutionDir)Custom.$(MSBuildThisFile)" Condition="$(SolutionDir)Custom.$(MSBuildThisFile)')"/>` or some variation of this to Microsoft.Common.CurrentVersion.targets just under $(MSBuildProjectFullPath).user.

I created a user voice post previously but when I saw the news about MSBuild being open source I decided to raise it directly with you here. https://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/9674760-add-a-custom-solution-import-in-microsoft-common-t 
I apologise for my persistence with this request however it would make maintaining an enterprise build system much easier. 

I am happy to answer any queries you may have, especially if it strengthens the possibility of this becoming a reality! Thanks

rainersigwald:
This is an interesting idea, and there is definitely room for improvement in large-scale configurability.  However, I have a number of concerns about this approach.

The main issue is that, since a solution isn't a core MSBuild concept, using this extensibility point would be confusing for anyone who lives in a world where they don't only build through the solution.

That would be a problem for many CI build servers that are configured to build through a traversal project rather than a solution (which is quite common).

It would also be confusing in the fairly common case of having multiple "subsolutions" for different teams or roles within a codebase.  One would have to have the canonical configuration, but how would changes be coordinated between solutions?

And tying a project to a solution would also dramatically change single-project builds from the command line, since MSBuild would have no way to know what solution to find configuration in.

In general, I favor the explicit approach of adding an import to all projects, as you currently do.

Thoughts?

Serivy:
Thanks for taking the time to write a reply. Sorry if this is a long winded reply but you asked for my thoughts. :)

[...]

PR #751, opened 2016-06-30
Implementation of ImportDirectoryBuildProps and ImportDirectoryBuildTargets

jeffkl:
#222 MSBuild should have a solution import in Microsoft.Common.Targets

Here is my initial implementation of the "directory" import.  I optimized it as much as possible by not continuing if GetDirectoryNameOfFileAbove() returns String.Empty.

1. _DirectoryBuildPropsFile is sort of private with the underscore and is only there so that I didn't have global.build.props in a few places.  I can remove it if we want.
2. _DirectoryBuildPropsBasePath is used to store the value of GetDirectoryNameOfFileAbove() in the event that the file isn't found.
3. DirectoryBuildPropsPath is the property that stores the combination of the file and base path, unless the user overrides it.

Unit tests have also been implemented.


jeffkl:
@dotnet-bot test this

rainersigwald:
@jeffkl Pretty sure you have to be more polite than that.

(In other words, the "please" is mandatory.  I didn't configure the bot)

jeffkl:
Haha that's awesome.  

@dotnet-bot test this please?

src/XMakeTasks/Microsoft.Common.targets
```diff
@@ -121,7 +121,23 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        Determine the path to the global build targets file if the user did not disable $(ImportGlobalBuildTargets) and
+        they did not already specify an absolute path to use via $(GlobalBuildTargetsPath)
+    -->
+  <PropertyGroup Condition="'$(ImportGlobalBuildTargets)' == 'true' and '$(GlobalBuildTargetsPath)' == ''">
+    <_GlobalBuildTargetsFile Condition="'$(_GlobalBuildTargetsFile)' == ''">global.build.targets</_GlobalBuildTargetsFile>
+    <_GlobalBuildTargetsBasePath Condition="'$(_GlobalBuildTargetsBasePath)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), '$(_GlobalBuildTargetsFile)'))</_GlobalBuildTargetsBasePath>
+    <GlobalBuildTargetsPath Condition="'$(_GlobalBuildTargetsBasePath)' != '' and '$(_GlobalBuildTargetsFile)' != ''">$([System.IO.Path]::Combine('$(_GlobalBuildTargetsBasePath)', '$(_GlobalBuildTargetsFile)'))</GlobalBuildTargetsPath>
+  </PropertyGroup>
+
+  <Import Project="$(GlobalBuildTargetsPath)" Condition="'$(ImportGlobalBuildTargets)' == 'true' and exists('$(GlobalBuildTargetsPath)')"/>
```
rainersigwald: I think it makes sense to put this block before the `ImportAfter` imports. That seems to better go with the intent of "before/after".

jeffkl: I agree.  Do you think it should import before or after Microsoft.Common.CurrentVersion.targets?

rainersigwald:
Odd, the first failure indicated one failed test, `Microsoft.Build.UnitTests.OM.Instance.ProjectInstance_Internal_Tests.Build`. The second looks like xunit crashed without writing a results file.

@dotnet-bot test this please

src/XMakeTasks/Microsoft.Common.targets
jeffkl: Ping?  Do you like the new placement?  Is everyone okay with the unit tests?

rainersigwald: I think I'd go after `CommonTargetsPath`. That would make it easier to override targets if you needed to.

---
In reply to: [69346571](https://github.com/Microsoft/msbuild/pull/751#discussion_r69346571) [](ancestors = 69346571)

jeffkl: Done

rainersigwald:
Implementation and tests look good!

I'm not totally settled about the names. "Global" makes sense by analogy to `global.json` in .NET CLI-land, but that's not commonly used now. And these aren't global settings--the existing machine- or user-wide magic imports would be, but these are limited to child directories.

jeffkl:
Hmm I'm afraid because the build keeps failing because of something weird.

@dotnet-bot test this please

rainersigwald:
LGTM

AndyGerlicher:
LGTM

PR #758, opened 2016-07-03
Fix mono/osx build and tests

radical:


src/XMakeBuildEngine/UnitTests/Definition/ToolsetConfigurationReader_Tests.cs
```diff
@@ -581,12 +581,12 @@ public void ExtensionPathsTest_Basic1()
 #if XPLAT
```
rainersigwald: (discussion, doesn't need to be done now)

Should we remove this condition? It seems . . . imprecise, and I vaguely recall that it was mostly to reduce code churn caused by porting a feature that originated in xplat to master.

(I'm [working on it!](https://github.com/Microsoft/msbuild/milestones/The%20Great%20Merge))

rainersigwald: Oh, you had to add the `XPLAT` define in this review. Then I think it should be removed entirely. If I'm reading the code right that shouldn't require any code change--it was just there to wall off the platform-detection methods that don't exist in master yet.

rainersigwald:
Thanks! Looks good once we just get rid of the `XPLAT `define..

radical:
Removed `XPLAT` completely.

rainersigwald:
Awesome. Looks good to me. @Microsoft/msbuild-maintainers ?

cdmihai:
LGTM

PR #759, opened 2016-07-05
Partial master → xplat merge

rainersigwald:
As usual I'll go through and annotate the conflicts/interesting parts.


dir.props
```diff
@@ -22,9 +22,19 @@
...
+    <!-- Output directories -->
+    <BinDir>$(ProjectDir)bin\</BinDir>
+    <TestWorkingDir>$(BinDir)tests\</TestWorkingDir>
+    <BootstrapDestination>$(BinDir)Bootstrap\</BootstrapDestination>
+
+    <!-- Package version / directories -->
+    <BuildToolsVersion>1.0.25-prerelease-00199</BuildToolsVersion>
+    <CompilerToolsVersion>1.0.0</CompilerToolsVersion>
     <XunitVersion>2.1.0</XunitVersion>
-    <CompilerToolsVersion>1.2.0-beta1-20160108-01</CompilerToolsVersion>
```
rainersigwald: Hmm, this downgrade is probably not good/intentional.

src/Shared/TaskLoader.cs
```diff
@@ -209,4 +210,4 @@ internal static void RemoveAssemblyResolver()
         }
 #endif
```
rainersigwald: This conflicted and I just added this endif.

src/Shared/UnitTests/FileUtilities_Tests.cs
```diff
@@ -329,38 +330,20 @@ public void CheckDerivableItemSpecModifiers()
...
-                NativeMethodsShared.GetModuleFileName(NativeMethodsShared.NullHandleRef, sb, sb.Capacity);
-#else
-                NativeMethodsShared.GetModuleFileName(IntPtr.Zero, sb, sb.Capacity);
-#endif
-                path = sb.ToString();
-            }
+            var path = Path.Combine(
+                Path.GetDirectoryName(FileUtilities.ExecutingAssemblyPath)
+                    .TrimEnd(new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar }),
+                "MSBuild.exe");
```
rainersigwald: I don't understand what this block was doing before in xplat. Does anyone else?

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -497,34 +497,37 @@ static internal Project CreateInMemoryProject
             // Anonymous in-memory projects use the current directory for $(MSBuildProjectDirectory).
             // We need to set the directory to something reasonable.
             string originalDir = Directory.GetCurrentDirectory();
+            try
+            {
 #if FEATURE_SPECIAL_FOLDERS
```
rainersigwald: This conflicted between `try` and `#if`, shouldn't be big.

src/XMakeBuildEngine/UnitTests/BackEnd/TaskRegistry_Tests.cs
```diff
@@ -76,56 +72,64 @@ public override bool Execute()
...
 
-            File.Delete(outputFile);
-            bool succeeded = true;
+            var codeFile = FileUtilities.GetTemporaryFile();
+            var outputFile = FileUtilities.GetTemporaryFile(".dll");
+
+            // Setting the current directory to the MSBuild running location. It *should* be this
+            // already, but if it's not some other test changed it and didn't change it back. If
+            // the directory does not include the reference dlls the compilation will fail.
+            Directory.SetCurrentDirectory(FileUtilities.CurrentExecutableDirectory);
```
rainersigwald: I think the conflicts in this file are mostly from @andygerlicher

rainersigwald:
>                 return; // "TODO: under Unix this runs out of stack. Investigate"

Ugh :(

---
Refers to: src/XMakeBuildEngine/UnitTests/Definition/ToolsetRegistryReader_Tests.cs:109 in ea4dac8. [](commit_id = ea4dac840fb0056db9768158d15107a4eead9608, deletion_comment = False)

src/XMakeBuildEngine/UnitTests/Utilities_Tests.cs
```diff
@@ -101,7 +100,7 @@ public void CommentsInPreprocessing()
                 Assert.Equal(
                     MSBuildApp.ExitType.Success,
                     MSBuildApp.Execute(
-                        new[] { @"c:\bin\msbuild.exe", '"' + input + + '"',
+                        new[] { @"c:\bin\msbuild.exe", '"' + input + '"',
```
rainersigwald: conflict here on a double + that I don't understand how it ever worked.

src/XMakeCommandLine/UnitTests/XMake_Tests.cs
```diff
@@ -1843,22 +1822,33 @@ public void ProcessConsoleLoggerSwitches()
         }
         #endregion
 
-        private void CopyMSBuildExeToPath(string destPath)
+        private string CopyMSBuild()
```
rainersigwald: This method was interesting

src/XMakeBuildEngine/UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -563,7 +563,7 @@ public void OutOfProcNodeForwardCertainproperties()
         /// we launch a child node that we get only that value. Also, make sure that when a project is pulled from the results cache
         /// and we have a list of properties to serialize that we do not crash. This is to prevent a regression of 826594
         /// </summary>
-        [Fact(Skip = "Test fails in xunit when multiple tests are run")]
+        [Fact(Skip = "FEATURE: OOP NODES")]
```
rainersigwald: Made this like the others nearby.

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -3280,7 +3280,10 @@ public void Medley()
                 "$(e.Substring($([System.IO.Path]::Combine(`a`, `b`))))",
                 "$([]::())",
                 "$((((",
+#if FEATURE_WIN32_REGISTRY
+                // If no registry, this gets expanded to the empty string
                 "$(Registry:X)",
+#endif
```
rainersigwald: I don't love this but don't really have a better idea at the moment. Opinions?

src/Shared/UnitTests/FileUtilities_Tests.cs
rainersigwald: Probably worth looking at 864bde6c in its totality.

---
In reply to: [69628944](https://github.com/Microsoft/msbuild/pull/759#discussion_r69628944) [](ancestors = 69628944)

src/XMakeCommandLine/UnitTests/XMake_Tests.cs
```diff
@@ -564,18 +567,14 @@ public void SetConsoleUICulture()
         }
 #endif
 
+#if FEATURE_SYSTEM_CONFIGURATION
```
rainersigwald: @andygerlicher Do we know the right behavior here for netcore yet?

dir.props
rainersigwald: Updated it to 1.3.2 here and in the `project.json`, even though nothing seemed to care.

rainersigwald:

rainersigwald:
Well, I'm not happy about it, but since I couldn't repro the hang on two different Macs and now it worked in CI on retry . . .

@Microsoft/msbuild-maintainers I think this is ready for review. Interesting bits of the merge called out, plus it'd be good to look at the individual commits since eb9397a.

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
cdmihai: IsWindows runtime check?

src/XMakeTasks/UnitTests/Exec_Tests.cs
```diff
@@ -79,6 +80,7 @@ public void ExitCodeCausesFailure()
         }
 
         [Fact]
+        [PlatformSpecific(PlatformID.Windows)] // TODO: https://github.com/Microsoft/msbuild/issues/434
```
cdmihai: Is it worth pointing the issue to this test?

cdmihai:
LGTM. Though I'd feel safer if @AndyGerlicher also took a look :)

src/XMakeTasks/UnitTests/Exec_Tests.cs
rainersigwald: Done via commit message magic.

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
rainersigwald: Doesn't work well here because it's an array (so fixed length)--what should be put in its place? I thought about adding logic in the loop that consumes the list but that got complicated.

A better solution here would be to convert this to an xunit `Theory` data-driven test and avoid that entry via an IsWindows check. But I'm trying to _reduce_ merge conflicts here :)

cdmihai: You could add it inside IsWindows if below (https://github.com/Microsoft/msbuild/pull/759/files?diff=unified#diff-adf408463d779e2551ad431a97da518dR3291). It seems to already be including a windows only invalid string.

Or is the registry not available on xplat at all, regardless of OS? In that case you're right, Theory with ifdef would be the only way, but not worth the pain right now.

rainersigwald: As is, today, it's not available in Core flavors at all, so I think this is the only way to go :(

rainersigwald:
@dotnet-bot test this please

PR #760, opened 2016-07-05
Updating path to versionless registry for GenericBootstrapper

vladbarosan:
Since this is just for Willow, I removed the code which used versions. 

Also I removed the path BOOTSTRAPPER_WOW64_REGISTRY_PATH_BASE  as the content
under WOW6432 will be automatically be presented to 32 bit apps running on 64 bit OS.

### DIFF
--- a/src/XMakeTasks/BootstrapperUtil/Util.cs
+++ b/src/XMakeTasks/BootstrapperUtil/Util.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Build.Tasks.Deployment.Bootstrapper
         }
 
         // This is the 4.0 property and will always point to the Dev10 registry key so that we don't break backwards compatibility.
-        // Applications relying on 4.5 will need to use the new method that is introduced in 4.5
+        // Applications relying on 4.5 will need to use the new method that is introduced in 4.5.
         public static string DefaultPath
         {
             get
@@ -101,7 +101,7 @@ namespace Microsoft.Build.Tasks.Deployment.Bootstrapper
         // This method is not going to cache the path as it could be different depending on the Visual Studio version.
         public static string GetDefaultPath(string visualStudioVersion)
         {
-            // if the Visual Studio Version is not a valid string, we will fall back to using the v4.0 property
+            // if the Visual Studio Version is not a valid string, we will fall back to using the v4.0 property.
             if (String.IsNullOrEmpty(visualStudioVersion))
             {
                 return DefaultPath;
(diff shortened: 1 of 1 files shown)


rainersigwald:
Can you update your commit message to have more context about the change? I'm mostly curious about why the old behavior is now and what the expectation is moving forward.

src/XMakeTasks/BootstrapperUtil/Util.cs
```diff
@@ -14,9 +14,7 @@ namespace Microsoft.Build.Tasks.Deployment.Bootstrapper
     internal static class Util
     {
         private const string BOOTSTRAPPER_REGISTRY_PATH_BASE = "Software\\Microsoft\\GenericBootstrapper\\";
-        private const string BOOTSTRAPPER_WOW64_REGISTRY_PATH_BASE = "Software\\Wow6432Node\\Microsoft\\GenericBootstrapper\\";
```
rainersigwald: I suspect that this is here for the case when 64-bit command-line MSBuild is run even though VS is 32-bit installed. Does that scenario no longer make sense?

vladbarosan: Yes it actually makes sense to have it if running from 64 bit command line

vladbarosan:
I have switched to keeping the old checkings in place as from what you say backwards compatibility is still needed.

AndyGerlicher:
@vladbarosan I think the change looks good, but I agree with @rainersigwald we'll have no idea what this is in 6 months. Can you elaborate more in the commit? Even an example in code comments or commit would be great (what the previous keys looked like and what the new one is).

AndyGerlicher:

vladbarosan:
@AndyGerlicher in mentioned in a comment that in Dev15 we are going versionless on the Registry Key path. 
e.g.: We will use **Software\Wow6432Node\Microsoft\GenericBootstrapper\** and the value name **Path** and 
it will point outside of VS hive to **Program Files (x86)\Microsoft SDKs\ClickOnce Bootstrapper**

Until now we where using  **Software\Wow6432Node\Microsoft\GenericBootstrapper\$(vsversion)**  with value name **Path** and it is pointing inside VS hive to **C:\Program Files (x86)\Microsoft Visual Studio 14.0\SDK\Bootstrapper**

This is because we are installing this as a singleton MSI and we need it to be outside of VS.

rainersigwald:

PR #767, opened 2016-07-08
Fix project import search paths when property undefined.

AndyGerlicher:
Fix a null ref Exception when the evaluator tries to implement import
search paths for a property that does not (yet) exist. In this case the
search path(s) will still be checked but the current null value will be
ignored.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -2083,7 +2083,7 @@ namespace Microsoft.Build.Evaluation
 
             var pathsToSearch =
                 // The actual value of the property, with no fallbacks
-                new[] {prop.EvaluatedValue}
+                new[] { prop?.EvaluatedValue }
                 // The list of fallbacks, in order
                 .Concat(fallbackSearchPathMatch.SearchPaths).ToList();
 
@@ -2100,6 +2100,11 @@ namespace Microsoft.Build.Evaluation
             // 2. 1 or more project files *found* but ignored (like circular, self imports)
             foreach (var extensionPath in pathsToSearch)
             {
+                // In the rare case that the property we've enabled for search paths hasn't been defined
+                // we will skip it, but continue with other paths in the fallback order.
+                if (string.IsNullOrEmpty(extensionPath))
+                    continue;
+
                 string extensionPathExpanded = _data.ExpandString(extensionPath);
 
                 if (!Directory.Exists(extensionPathExpanded))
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
```diff
@@ -605,6 +605,71 @@ public void FallbackImportWithIndirectReference()
             }
         }
 
+        // Fall-back search path with custom variable
```
rainersigwald: comment not particularly meaningful

rainersigwald:
>                 string extensionPathExpanded = _data.ExpandString(extensionPath);

Instead of special-casing above, should we just
```C#
if (string.IsNullOrEmpty(extensionPath)) {
  continue;
}
```

?

---
Refers to: src/XMakeBuildEngine/Evaluation/Evaluator.cs:2115 in 15468b8. [](commit_id = 15468b8b3fbaac401f7e1397e1fdcc759e8dfe19, deletion_comment = False)

rainersigwald:
>                 string extensionPathExpanded = _data.ExpandString(extensionPath);

Instead of special-casing above, should we just
```C#
if (string.IsNullOrEmpty(extensionPath)) {
  continue;
}
```

?

---
Refers to: src/XMakeBuildEngine/Evaluation/Evaluator.cs:2115 in 15468b8. [](commit_id = 15468b8b3fbaac401f7e1397e1fdcc759e8dfe19, deletion_comment = False)

rainersigwald:
>                 string extensionPathExpanded = _data.ExpandString(extensionPath);

Instead of special-casing above, should we just
```C#
if (string.IsNullOrEmpty(extensionPath)) {
  continue;
}
```

?

---
Refers to: src/XMakeBuildEngine/Evaluation/Evaluator.cs:2115 in 15468b8. [](commit_id = 15468b8b3fbaac401f7e1397e1fdcc759e8dfe19, deletion_comment = False)

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2080,10 +2080,11 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             // Adding the value of $(MSBuildExtensionsPath*) property to the list of search paths
             var prop = _data.GetProperty(fallbackSearchPathMatch.PropertyName);
+            var propValue = prop?.EvaluatedValue;
```
rainersigwald: Does this provide anything now?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2100,6 +2101,11 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
             // 2. 1 or more project files *found* but ignored (like circular, self imports)
             foreach (var extensionPath in pathsToSearch)
             {
+                // In the rare case that the property we've enabled for search paths hasn't been defined
+                // we will still add the fall-back search paths to our path(s) to search.
```
rainersigwald: "skip it, but continue with other paths in the fallback order."

rainersigwald:
LGTM after my nits.

Issue #764, opened 2016-07-08
Failure to load ResolveComReference in OOP node


rainersigwald:
This simple project:
```XML
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Target Name="b">
    <Message MSBuildArchitecture="x64"
             Text="Hello!" />
    <ResolveComReference MSBuildArchitecture="x64" />
  </Target>
</Project>
```

Causes an MSBuild hang when using the GitHub version of MSBuild (with the Willow Preview 3).

There are two problems: a failure to load the task, and a hang as a result. This issue is for the former. #765 is the latter.

The failure to load the task is caused by [a check in `TaskLoader`](https://github.com/Microsoft/msbuild/blob/da283a7bea8a023c47cea73630fa3126c57b49a1/src/Shared/TaskLoader.cs#L121) that attempts to ensure that the AppDomain that will run the task loaded the same task that we have loaded. If I just step over that check in the debugger, everything seems fine.

I find it really interesting that the `Message` task works fine, but `ResolveComReference` does not.

<details>
<summary>Full log to the hang</summary>
Build started 7/8/2016 9:46:06 AM.
Project "C:\Users\raines\Desktop\test.proj" on node 1 (default targets).
Building with tools version "15.0".
Project file contains ToolsVersion="". This toolset may be unknown or missing, in which case you may be able to resolve this by installing the appropriate version of MSBuild, or the build may have been forced to a particular ToolsVersion for policy reasons. Treating the project as if it had ToolsVersion="15.0". For more information, please see http://go.microsoft.com/fwlink/?LinkId=293424.
Target "b" in project "C:\Users\raines\Desktop\test.proj" (entry point):
Using "Message" task from assembly "Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
Task "Message"
  Launching task "Message" from assembly "Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" in an external task host with a runtime of "CLR4" and a process architecture of "x64".
  Hello!
Done executing task "Message".
Using "ResolveComReference" task from assembly "Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
Task "ResolveComReference"
[...]

PR #768, opened 2016-07-08
Update setup to drop x86 AnyCPU binaries for AMD64

AndyGerlicher:
This will ensure that the AnyCPU binaries in the root and AMD64 folder are
bitwise identical. This was a suspected culprit in an issue where x64 OOP
MSBuild ended up loading multiple duplicate types.

Related to #764


rainersigwald:
LGTM

PR #769, opened 2016-07-08
Refactor ResolveAssemblyReference_Tests.

AndyGerlicher:
- Move to AssemblyDependency folder (to match Tasks code).
- Move classes to their own file.

Making this it's own PR before I make changes and add tests. I did not make any changes to the code itself.


rainersigwald:
`git blame` is making me very sad about this. Even with `-C -C -C10 -w` it's attributing most of the diffs to you :(

But I tried doing just a single class at a time in the hopes that doing that would provide a better hint for Git's heuristics and it didn't work. So might as well do it all at once.

`Rainer.DisdainForGitIntegrationRecordHandling++;`

LGTM

Issue #734, opened 2016-06-27
Enhance OM: Add Remove attribute for static item tags


cdmihai:
Parent issue: #700 

Goal: Ability to remove items during evaluation time.

Design so far:

```
<Compile Remove="bin\**;obj\**">
```

- Evaluation time (when update tag is not in target)
  - For all ProjectItems produced in tags above this xml location (including items from imports)
  - That have Type=Compile and EvaluatedInclude matches any item in the Remove attribute
  - Remove from project

Design todo:
 - [ ] how much tweaking does the current dynamic Remove require to make it work during evaluation?
 - [ ] add construction APIs

Issue #733, opened 2016-06-27
Enhance OM: Add Update attribute for static item tags


cdmihai:
Parent issue: #700

Goal: Support for modifying metadata of a file that was included via glob

Design so far:

```xml
<Project>
   <ItemGroup>
      <Compile Update="foo.cs">    // or Update="*.designer.cs"
        <MetadataKey>MetadataValue</MetadataKey>
      </Compile>
   </ItemGroup>
</Project>
```

- Static item during evaluation (when update tag is not in target)
  - For all ProjectItems produced in tags above this xml location (including items from imports)
  - That have Type=Compile and EvaluatedInclude="Foo.cs"
  - AddOrUpdate Metadakey=MetadataValue

PR #770, opened 2016-07-09
Evaluate item lists lazily

dsplaisted:
This is a first step towards efficient support for updating item metadata (#733) and removing items (#734) during evaluation.  So far, it doesn't support either of those.

What it does is instead of fully evaluating items as item elements are encountered during the evaluation, it builds an operation graph, with immutable, lazily evaluated item lists.  Take the case where you have the following in ItemGroups:

```xml
<Compile include="**\*.cs" />

<Compile remove="bin\**\*.*;obj\**\*.*;node_modules\**\*.*" />
```

The idea is that this would basically create a Glob Include operation followed by a Remove operation.  The Remove operation would depend on the value of the Glob Include operation, but when it requests the value of the item list, it could pass the list of globs that should be excluded, so when the file system code globs for `**\*.cs`, it can ignore traversing any directories where the results would subsequently be removed in the subsequent operation.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -816,11 +816,41 @@ namespace Microsoft.Build.Evaluation
             string endPass2 = String.Format(CultureInfo.CurrentCulture, "Evaluate Project {0} - End Pass 2 (Item Definitions)", projectFile);
             DataCollection.CommentMarkProfile(8818, endPass2);
 #endif
+            LazyItemEvaluator<P, I, M, D> lazyEvaluator = null;
+            lazyEvaluator = new LazyItemEvaluator<P, I, M, D>(_data, _itemFactory, _buildEventContext, _loggingService);
+
             // Pass3: evaluate project items
             foreach (ProjectItemGroupElement itemGroupElement in _itemGroupElements)
             {
-                EvaluateItemGroupElement(itemGroupElement);
+                EvaluateItemGroupElement(itemGroupElement, lazyEvaluator);
             }
+
+            if (lazyEvaluator != null)
+            {
+                IList<LazyItemEvaluator<P, I, M, D>.ItemData> items = lazyEvaluator.GetAllItems();
+                foreach (var itemData in items)
+                {
+                    if (itemData.ConditionResult)
+                    {
+                        _data.AddItem(itemData.Item);
+
+                        if (_data.ShouldEvaluateForDesignTime)
+                        {
+                            _data.AddToAllEvaluatedItemsList(itemData.Item);
+                        }
+                    }
+                }
+
+                if (_data.ShouldEvaluateForDesignTime)
+                {
+                    foreach (var itemData in items)
+                    {
+                        _data.AddItemIgnoringCondition(itemData.Item);
+                    }
+                }
+
+            }
+
 #if (!STANDALONEBUILD)
             CodeMarkers.Instance.CodeMarker(CodeMarkerEvent.perfMSBuildProjectEvaluatePass3End);
 #endif
@@ -1200,7 +1230,7 @@ namespace Microsoft.Build.Evaluation
         /// <summary>
         /// Evaluate the items in the itemgroup and add the applicable ones to the data passed in
         /// </summary>
-        private void EvaluateItemGroupElement(ProjectItemGroupElement itemGroupElement)
+        private void EvaluateItemGroupElement(ProjectItemGroupElement itemGroupElement, LazyItemEvaluator<P, I, M, D> lazyEvaluator)
         {
 #if FEATURE_MSBUILD_DEBUGGER
             if (DebuggerManager.DebuggingEnabled)
@@ -1209,13 +1239,21 @@ namespace Microsoft.Build.Evaluation
             }
 #endif
 
-            bool itemGroupConditionResult = EvaluateCondition(itemGroupElement, ExpanderOptions.ExpandPropertiesAndItems, ParserOptions.AllowPropertiesAndItemLists);
+            bool itemGroupConditionResult;
+            if (lazyEvaluator != null)
+            {
+                itemGroupConditionResult = lazyEvaluator.EvaluateConditionWithCurrentState(itemGroupElement, ExpanderOptions.ExpandPropertiesAndItems, ParserOptions.AllowPropertiesAndItemLists);
+            }
+            else
+            {
+                itemGroupConditionResult = EvaluateCondition(itemGroupElement, ExpanderOptions.ExpandPropertiesAndItems, ParserOptions.AllowPropertiesAndItemLists);
+            }
 
             if (itemGroupConditionResult || _data.ShouldEvaluateForDesignTime)
             {
                 foreach (ProjectItemElement itemElement in itemGroupElement.Items)
                 {
-                    EvaluateItemElement(itemGroupConditionResult, itemElement);
+                    EvaluateItemElement(itemGroupConditionResult, itemElement, lazyEvaluator);
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -1209,13 +1240,21 @@ private void EvaluateItemGroupElement(ProjectItemGroupElement itemGroupElement)
             }
 #endif
 
-            bool itemGroupConditionResult = EvaluateCondition(itemGroupElement, ExpanderOptions.ExpandPropertiesAndItems, ParserOptions.AllowPropertiesAndItemLists);
+            bool itemGroupConditionResult;
+            if (lazyEvaluator != null)
+            {
+                itemGroupConditionResult = lazyEvaluator.EvaluateConditionWithCurrentState(itemGroupElement, ExpanderOptions.ExpandPropertiesAndItems, ParserOptions.AllowPropertiesAndItemLists);
```
cdmihai: >EvaluateConditionWithCurrentState [](start = 57, length = 33)

I saw that item definition groups expand some metadata when evaluating their conditions. Didn't look farther into it, whether it's some actual item metadata or some special item definition metadata

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -0,0 +1,704 @@
...
+                        //  Code corresponds to EngineFileUtilities.GetFileList
+                        bool containsEscapedWildcards = EscapingUtilities.ContainsEscapedWildcards(includeSplitEscaped);
+                        bool containsRealWildcards = FileMatcher.HasWildcards(includeSplitEscaped);
+
+                        if (containsEscapedWildcards && containsRealWildcards)
+                        {
+                            // Umm, this makes no sense.  The item's Include has both escaped wildcards and 
+                            // real wildcards.  What does he want us to do?  Go to the file system and find
+                            // files that literally have '*' in their filename?  Well, that's not going to 
+                            // happen because '*' is an illegal character to have in a filename.
```
cdmihai: >/ happen because '*' is an illegal character to have in a filename. [](start = 29, length = 67)

:)

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -0,0 +1,704 @@
...
+                    }
+                }
+            }
+
+            //  Code corresponds to Evaluator.EvaluateItemElement
+
+            // STEP 4: Evaluate, split, expand and subtract any Exclude
+            if (itemElement.Exclude.Length > 0)
+            {
+                string evaluatedExclude = _expander.ExpandIntoStringLeaveEscaped(itemElement.Exclude, ExpanderOptions.ExpandProperties, itemElement.ExcludeLocation);
```
cdmihai: >itemElement.Exclude [](start = 81, length = 19)

For GetItemProvenance, the exclude string will probably have to go through the same treatment as the include, to detect values, globs, expressions, etc

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
dsplaisted: I don't get it.  Is it not illegal on non-Windows filesystems?

BTW these comments aren't mine, they're just copied from other code (in this case EngineFileUtilities.GetFileList)

cdmihai: I know it's an old comment, but it means we have to fix it for crossplat, at some point in the future.

Here's from osx:
```
osx-build-1:test micodoba$ touch a*a
osx-build-1:test micodoba$ touch \*a
osx-build-1:test micodoba$ touch a\*
osx-build-1:test micodoba$ ls
*a      a*      a*a
```

Nothing to do with lazy item evaluation.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -0,0 +1,704 @@
...
+                            }
+
+#if FEATURE_MSBUILD_DEBUGGER
+                        //if (DebuggerManager.DebuggingEnabled)
+                        //{
+                        //    DebuggerManager.PulseState(metadatumElement.Location, _itemPassLocals);
+                        //}
+#endif
+
+                            string evaluatedValue = _expander.ExpandIntoStringLeaveEscaped(metadatumElement.Value, ExpanderOptions.ExpandAll, metadatumElement.Location);
```
cdmihai: Can it happen for metadata strings to reference items that are not in ReferencedItems?

```
<Foo include="1;2;3">
<Bar Include="a">
   <m>@(Foo)<m>
<Bar>
```

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
dsplaisted: Can you file a separate issue for that then?

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.ItemFactoryWrapper.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.Build.BackEnd;
+using Microsoft.Build.Construction;
+using Microsoft.Build.Execution;
+using Microsoft.Build.Shared;
+
+namespace Microsoft.Build.Evaluation
+{
+    internal partial class LazyItemEvaluator<P, I, M, D>
+    {
+        class ItemFactoryWrapper : IItemFactory<I, I>
```
cdmihai: What is this wrapper used for? It seems to do exactly what the wrapped factory is doing, but storing the current item element twice.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
cdmihai: Opened #781

dsplaisted:
OK, I think this is ready for review.

AndyGerlicher:
LGTM. We just need to write a few more tests for this at some point.

cdmihai:
LGTM. Going to merge this is in tomorrow, if nobody objects.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -816,11 +816,41 @@ private static ProjectTargetInstance ReadNewTargetElement(ProjectTargetElement t
...
+            lazyEvaluator = new LazyItemEvaluator<P, I, M, D>(_data, _itemFactory, _buildEventContext, _loggingService);
+
             // Pass3: evaluate project items
             foreach (ProjectItemGroupElement itemGroupElement in _itemGroupElements)
             {
-                EvaluateItemGroupElement(itemGroupElement);
+                EvaluateItemGroupElement(itemGroupElement, lazyEvaluator);
+            }
+
+            if (lazyEvaluator != null)
```
rainersigwald: This is "unlazify", right? A comment for intention would be nice.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -816,11 +816,41 @@ private static ProjectTargetInstance ReadNewTargetElement(ProjectTargetElement t
...
+                            _data.AddToAllEvaluatedItemsList(itemData.Item);
+                        }
+                    }
+                }
+
+                if (_data.ShouldEvaluateForDesignTime)
+                {
+                    foreach (var itemData in items)
+                    {
+                        _data.AddItemIgnoringCondition(itemData.Item);
```
rainersigwald: Does this need to be a separate pass through the list?

rainersigwald:
Minor comments, not blocking. LGTM

cdmihai:
@dotnet-bot test this please

src/XMakeBuildEngine/Evaluation/Evaluator.cs
cdmihai: Yup. I'll add a comment for it

src/XMakeBuildEngine/Evaluation/Evaluator.cs
cdmihai: It was trying to emulate [this ](https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeBuildEngine/Evaluation/Evaluator.cs#L1788-L1807)code, but it can actually be removed in this case since the structure changed a bit.

cdmihai:
Closed via #868

PR #776, opened 2016-07-12
Calculate a better MSBuild tools/extensions path

### LINKED ISSUE #704: Calculate a better MSBuild tools/extensions path

AndyGerlicher:
Move all build path environment discovery to new class
 - With Dev15, finding MSBuild is a bit trickier. This change tries to
   address that by moving the logic to a single place and streamlining the
   usage.
 - Audited code that needed to find the MSBuild path to use this.
   Previously there were a variety of methods used, ideally this should be
   the only one (other than perhaps finding tools from another toolset).

Closes #704

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -1,0 +1,342 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+
+namespace Microsoft.Build.Shared
+{
+    internal class BuildEnvironmentHelper
+    {
+        private static readonly string[] s_testRunners =
+        {
+            "XUNIT", "NUNIT", "MSTEST", "VSTEST", "TASKRUNNER",
+            "VSTESTHOST", "QTAGENT32", "CONCURRENT", "RESHARPER", "MDHOST", "TE.PROCESSHOST"
+        };
+
+        /// <summary>
+        /// Name of the Visual Studio process(es)
+        /// </summary>
+        private static readonly string[] s_visualStudioProcess = {"DEVENV"};
+
+        /// <summary>
+        /// Name of the MSBuild process(es)
+        /// </summary>
+        private static readonly string[] s_msBuildProcess = {"MSBUILD"};
+
+        /// <summary>
+        /// Gets the cached Build Environment instance.
+        /// </summary>
+        public static BuildEnvironment Instance => BuildEnvironmentHelperSingleton.s_instance;
+        
+        /// <summary>
+        /// Find the location of MSBuild.exe based on the current environment.
+        /// </summary>
+        /// <returns>Build environment.</returns>
+        private static BuildEnvironment Initialize()
+        {
+            // Get the executable we are running
+            var processNameCommandLine = s_getProcessFromCommandLine();
+            var processNameCurrentProcess = s_getProcessFromRunningProcess();
+            var executingAssembly = s_getExecutingAssmblyPath();
+            var currentDirectory = s_getCurrentDirectory();
+
+
+            // Check if our current process name is in the list of own test runners
+            var runningTests = IsProcessInList(processNameCommandLine, s_testRunners) ||
+                               IsProcessInList(processNameCurrentProcess, s_testRunners);
+
+            // Check to see if we're running inside of Visual Studio
+            bool runningInVisualStudio;
+            var visualStudioPath = FindVisualStudio(
+                new[] {processNameCommandLine, processNameCurrentProcess},
+                out runningInVisualStudio);
+
+            string msbuildFromVisualStudioRoot = null;
+            if (!string.IsNullOrEmpty(visualStudioPath))
+            {
+                msbuildFromVisualStudioRoot = FileUtilities.CombinePaths(visualStudioPath, "MSBuild", "15.0", "Bin");
+            }
+
+            var possibleLocations = new Func<BuildEnvironment>[]
+            {
+                // See if we're running from MSBuild.exe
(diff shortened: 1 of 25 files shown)


AndyGerlicher:

AndyGerlicher:

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -0,0 +1,314 @@
...
+
+        private static string GetCurrentDirectory()
+        {
+            return Directory.GetCurrentDirectory();
+        }
+
+        /// <summary>
+        /// Resets the current singleton instance (for testing).
+        /// </summary>
+        internal static void ResetInstance(Func<string> getProcessFromCommandLine = null, Func<string> getProcessFromRunningProcess = null, Func<string> getExecutingAssmblyPath =  null, Func<string> getCurrentDirectory = null)
```
cdmihai: Is there a convention for marking test only methods? In some other class (I think the whitelisted property functions) I saw the test methods prepended with "_TestOnly"

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -0,0 +1,314 @@
...
+                }
+            }
+
+            // VSInstallDir is set from the Developer Command Prompt
+            var vsInstallDir = Environment.GetEnvironmentVariable("VSINSTALLDIR");
+            var vsVersion = Environment.GetEnvironmentVariable("VisualStudioVersion");
+
+            if (!string.IsNullOrEmpty(vsInstallDir) &&
+                !string.IsNullOrEmpty(vsVersion) &&
+                vsVersion == "15.0" &&
```
cdmihai: Is this something that's going to have to be updated with each new VS version? If so, can you replace it with something like (Float.Parse(vsVersion) >= 15)?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -0,0 +1,314 @@
...
+            bool runningInVisualStudio;
+            var visualStudioPath = FindVisualStudio(
+                new[] {processNameCommandLine, processNameCurrentProcess},
+                out runningInVisualStudio);
+
+            string msbuildFromVisualStudioRoot = null;
+            if (!string.IsNullOrEmpty(visualStudioPath))
+            {
+                msbuildFromVisualStudioRoot =
+                    Path.Combine(Path.Combine(Path.Combine(visualStudioPath, "MSBuild"), "15.0"), "Bin");
```
cdmihai: Same thing for hardcoded 15

src/Shared/Strings.shared.resx
```diff
@@ -247,10 +247,14 @@
     <value>MSB5021: "{0}" and its child processes are being terminated in order to cancel the build.</value>
     <comment>{StrBegin="MSB5021: "}</comment>
   </data>
+  <data name="Shared.CanNotFindValidMSBuildLocation">
+    <value>MSB5024: Could not determine a valid location to MSBuild. Try running this process from the Developer Command Prompt for Visual Studio.</value>
```
cdmihai: This will be a good test for the entire localization pipeline :)

src/XMakeBuildEngine/UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -0,0 +1,236 @@
...
+                ["VisualStudioVersion"] = Environment.GetEnvironmentVariable("VisualStudioVersion"),
+            };
+
+            public EmptyBuildEnviroment()
+            {
+                try
+                {
+                    var files = new[] { "msbuild.exe", "msbuild.exe.config" };
+                    TempFolderRoot = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N"));
+                    BuildDirectory = Path.Combine(TempFolderRoot, "MSBuild", "15.0", "Bin");
```
cdmihai: This mock will get interesting if the layout changes in future VS versions :)

cdmihai:
LGTM with minor comments for hardcoded "15.0". Really great to see needless duplication removed! :)

AndyGerlicher:

AndyGerlicher:

PR #779, opened 2016-07-14
Localized builds

cdmihai:
- localized builds are enabled by setting the property `LocalizedBuild` to true
- localization is done via Satellite Assemblies
- localized resx files are generated on the fly in the intermediary output and not under version control
- loc testing can be done by setting `LocalizedTestBuild` to true. This replaces all strings with `!{resource id}!{English resource}!{Translated resource}!`
   - setting `LocalizedTestString` to some value replaces all resource strings with that value
- CI does localized builds by default

Future work. I suggest doing these in future PRs:
- xlab machines have a weird codepage set for their console (437, which is very USish). Because of that, [this method](https://github.com/Microsoft/msbuild/blob/f90dcff1db3b10c3b136bed1a23ce710f7c579cd/src/XMakeCommandLine/XMake.cs#L1216-L1247) resets the UI culture to en-US, which causes the English satellites to be picked up. I do not know whether normal windows installations have the same weird codepage. Contacted the world ready folks to ask them about that method, and what we should with it, if anything. To get localized satellites picked on xlab machines the workaround is doing `chcp 65001`
- migrate xlf sync tasks to xliffparser repo
- update our dependency to xliffparser when a new package is available, so we can use the `state=updated` xlf note for updated strings (as opposed to what happens now, where both new and updated resources get stamped with a "new" tag).

Todo before merging:
- [x] Write wiki on localization process and interaction with the localization team


cdmihai:
@dotnet-bot test this please

netci.groovy
```diff
@@ -29,7 +29,7 @@ def project = GithubProject
             case 'Windows_NT':
                 newJob.with{
                     steps{
-                        batchFile("call \"C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\vcvarsall.bat\" && RebuildWithLocalMSBuild.cmd")
+                        batchFile("call \"C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\vcvarsall.bat\" && RebuildWithLocalMSBuild.cmd /p:LocalizedBuild=true")
```
rainersigwald: Changing the script to take arguments is good, but I think I'd rather see the default be localized (with an opt-out). Not urgent though.

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
...
+      Condition="%(Identity) != ''"
+      ResxPath="%(ChildResx)"
+      XlfPath="@(XlfFiles)"
+      ReplaceWithTestString="$(LocalizedTestBuild)"
+      TestString="$(LocalizedTestString)"
+    />
+
+  </Target>
+
+    <!-- Copy paste and modify OpenSourceSign from sign.targets because it only signes the main assembly
```
rainersigwald: typo: `signes`

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
...
+      ReplaceWithTestString="$(LocalizedTestBuild)"
+      TestString="$(LocalizedTestString)"
+    />
+
+  </Target>
+
+    <!-- Copy paste and modify OpenSourceSign from sign.targets because it only signes the main assembly 
+  and provides no extension mechanism to add more assemblies to it
+  
+  If satellite assemblies do not get their delay sign bit switched off, the main assemblies won't load them
```
rainersigwald: obligatory "have you sent a PR upstream?"

build/LocalizationTasks/Microsoft.Build.LocalizationTasks.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <Compile Include="*.cs"/>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System"/>
+    <Reference Include="System.Text.RegularExpressions"/>
+    <Reference Include="Microsoft.Build.Framework" />
+    <Reference Include="Microsoft.Build.Utilities.Core" />
+    <Reference Include="fmdev.XliffParser">
+      <HintPath>$(XliffParserDir)fmdev.XliffParser.dll</HintPath>
```
rainersigwald: Can you not get this reference via NuGet?

src/Utilities/Microsoft.Build.Utilities.csproj
```diff
@@ -209,15 +209,20 @@
     </Compile>
     <!-- Win32 RC Files -->
     <RCResourceFile Include="native.rc" />
+
```
rainersigwald: delete spaces :)

src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -1193,10 +1193,6 @@
     <value>MSB4207: Only an in-process affinity may be specified for projects which have registered host objects.</value>
     <comment>{StrBegin="MSB4207: "}</comment>
   </data>
-  <data name="InvalidAffinityForProjectWithHostObject" UESanitized="false" Visibility="Public">
-    <value>MSB4208: Only an in-process affinity may be specified for projects which have registered host objects.</value>
-    <comment>{StrBegin="MSB4208: "}</comment>
-  </data>
```
rainersigwald: can you make removing this (redundant?) string its own commit?

src/XMakeCommandLine/UnitTests/XMake_Tests.cs
```diff
@@ -1783,11 +1783,32 @@ private string CopyMSBuild()
...
                 }
 
+                foreach (var directory in directoriesToCopy)
+                {
+                    foreach (var sourceFile in Directory.EnumerateFiles(directory, "*"))
+                    {
+                        var destinationFile = sourceFile.Replace(source, dest);
+
+                        var directoryName = Path.GetDirectoryName(destinationFile);
+                        if (!Directory.Exists(directoryName))
```
rainersigwald: The usual wisdom is that this is a race condition (the directory could get deleted between this line and the next), so there's no point in checking first--just call `CreateDirectory`.

src/XMakeTasks/System.Design.cs
```diff
@@ -91,11 +91,24 @@ internal sealed class SR
         internal const string InvalidIdentifier = "InvalidIdentifier";
 
         private static SR s_loader = null;
-        private ResourceManager _resources;
+        private MainAssemblyFallbackResourceManager _resources;
+
+        /// <summary>
+        /// The containing assembly is set to lookup up resources for the neutral language in satellite assemblies, not in the main assembly.
```
rainersigwald: typo: `lookup up`

src/XMakeTasks/Strings.resx
```diff
@@ -1562,12 +1562,6 @@
         LOCALIZATION: Please don't localize "CopyLocal" this is an item meta-data name.
    </comment>
   </data>
-  <data name="ResolveAssemblyReference.NoBecauseParentReferencesFoundInGac">
-    <value>This reference is not "CopyLocal" because the CopyLocalDependenciesWhenParentReferenceInGac property is set to false and all the parent references for this reference are found in the GAC.</value>
-    <comment>
-        LOCALIZATION: Please don't localize "CopyLocal" this is an item meta-data name. "CopyLocalDependenciesWhenParentReferenceInGac" is a property name.
-   </comment>
-  </data>
```
rainersigwald: you can put this one with the other remove-redundant thing

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="UpdateXlf" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  
+  <UsingTask AssemblyFile="$(BuildBinDir)\Microsoft.Build.LocalizationTasks.dll" Condition="$(LocalizedBuild) == 'true'" TaskName="SaveXlfToResx"/>
+  <UsingTask AssemblyFile="$(BuildBinDir)\Microsoft.Build.LocalizationTasks.dll" Condition="$(LocalizedBuild) == 'true'" TaskName="UpdateXlfFromResx"/>
+  <UsingTask AssemblyFile="$(BuildBinDir)\Microsoft.Build.LocalizationTasks.dll" Condition="$(LocalizedBuild) == 'true'" TaskName="EmitLocalizedResources"/>
+
+  <!-- Emit the localized EmbeddedResource items before the common targets start using them -->
+  <PropertyGroup>
+    <PrepareResourceNamesDependsOn>
```
rainersigwald: I'm not a fan of this mechanism of injecting your process. I prefer using `BeforeTargets` on the targets you're injecting.

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
...
+      LocalizedResxRoot="$(IntermediateOutputPath)"
+      RelativePathToXlfRoot=".\xlf\">
+
+      <Output TaskParameter="ResolvedXlfResources" ItemName="Xlf_Temp" />
+      <Output TaskParameter="ResolvedLocalizedResxResources" ItemName="Resx_temp" />
+
+    </EmitLocalizedResources>
+
+    <ItemGroup>
+      <XlfFiles Include="@(Xlf_Temp)"/>
```
rainersigwald: Why these temp items?

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
...
+      <EmbeddedResource Include="@(Resx_temp)"/>
+    </ItemGroup>
+
+  </Target>
+
+    <!-- The default english resources do not have corresponding xlf files. Simply copy them and add en in the name -->
+    <!-- TODO: Maybe create a dummy xlf file for English so we can delete this target? -->
```
rainersigwald: This seems like a fine idea. Is there a downside?

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
...
+  <Target Name="UpdateLocalizedResxForEnglishResx">
+
+      <ItemGroup>
+        <EnglishResources Include="@(NeutralResources -> '$(IntermediateOutputPath)%(Filename).en%(Extension)')"/>
```
rainersigwald: Could you do this transform in the task directly to avoid a temporary Item?

src/UpdateLocalizedResources.targets
```diff
@@ -0,0 +1,108 @@
...
+  <!-- Emit the localized EmbeddedResource items before the common targets start using them -->
+  <PropertyGroup>
+    <PrepareResourceNamesDependsOn>
+      EmitLocalizedResources;
+      UpdateLocalizedResxForEnglishResx;
+      $(PrepareResourceNamesDependsOn)
+    </PrepareResourceNamesDependsOn>
+  </PropertyGroup>
+
+  <Target Name="EmitLocalizedResources">
```
rainersigwald: Targets written by an MSBuild core team member that don't have incremental behavior? :scream:

(this can wait, but there should at least be a bug!)

build/LocalizationTasks/EmitLocalizedResources.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            xlf.SetMetadata(ChildResxMetadata, resxPath);
+
+            return resxItem;
+        }
+
+        private IEnumerable<ITaskItem> ComputeXlfResources(ITaskItem neutralResource)
+        {
+            var neutralResxRootDirectory = Path.GetDirectoryName(neutralResource.GetMetadata("FullPath"));
+            var xlfRootPath = Path.Combine(neutralResxRootDirectory, RelativePathToXlfRoot);
```
rainersigwald: Why prefer this to just doing the computation in msbuild and using it directly here?

build/LocalizationTasks/EmitLocalizedResources.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            // todo: throw if xlf root does not exist; throw in other places too
```
rainersigwald: sounds good :)

build/LocalizationTasks/EmitLocalizedResources.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Log.LogMessage("No neutral resources found");
+                return true;
+            }
+
+            ResolvedXlfResources = NeutralResources.SelectMany(ComputeXlfResources).ToArray();
+
+            ResolvedLocalizedResxResources = ResolvedXlfResources.Select(ComputeLocalizedResource).ToArray();
+
+            return true;
```
rainersigwald: MSBuild idiom: always return `!Log.HasLoggedErrors`--this prevents situations where you fail the build (`return false`) but never logged an error (so there's nothing to go on when it happens and your build fails).

build/LocalizationTasks/EmitLocalizedResources.cs
```diff
@@ -0,0 +1,112 @@
...
+            var xlfRootPath = Path.Combine(neutralResxRootDirectory, RelativePathToXlfRoot);
+
+            // todo: throw if xlf root does not exist; throw in other places too
+
+            var resx = neutralResource.ItemSpec;
+
+            return (Directory.EnumerateFiles(xlfRootPath).Where(f => IsValidXlf(f, resx)).Select((f => CreateXLFTaskItemForNeutralResx(f, resx))));            
+        }
+
+        private bool IsValidXlf(string xlfPath, string resx)
```
rainersigwald: Why is this needed? It seems a bit scary.

Actually the whole enumerate-files approach scares me a bit; I prefer an explicit-inputs approach.

build/LocalizationTasks/SaveXlfToResx.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (!string.IsNullOrEmpty(ReplaceWithTestString) && ReplaceWithTestString.Equals("true"))
+                {
+                    ReplaceResourceValuesWithTestString(xlfDocument);
+                }
+
+                xlfDocument.SaveAsResX(ResxPath);
+
+                return true;
+            }
+            catch (Exception e)
```
rainersigwald: can you be more specific than just `Exception`?

build/LocalizationTasks/SaveXlfToResx.cs
```diff
@@ -0,0 +1,64 @@
...
+        public string XlfPath { get; set; }
+
+        /// <summary>
+        /// When set to "true" it makes the task replace the resource values with another value.
+        /// Useful for testing localized builds.
+        /// 
+        /// If <see cref="TestString"/> is non empty, resource values get replaced with it.
+        /// If it is empty, resource values get replaced with !resource_id!english_resource!localized_resource!
+        /// </summary>
+        public string ReplaceWithTestString { get; set; }
```
rainersigwald: This seems pretty cool, but is it so strongly coupled with this functionality that it belongs here?

netci.groovy
cdmihai: Thought about that, but it means whenever you run RebuildWithLocalMSBuild on your local machine, you might get dirty xlf files if any resource changed. This might lead to people committing xlf file changes while the loc team is working on them as well.

src/UpdateLocalizedResources.targets
cdmihai: Shamefully, no. I'll add it to my todo list, along with moving the tasks to xliff parser.

netci.groovy
rainersigwald: Good point; I find that convincing.

build/LocalizationTasks/Microsoft.Build.LocalizationTasks.csproj
cdmihai: I had forgotten csproj and project.json actually work together :)

src/XMakeCommandLine/UnitTests/XMake_Tests.cs
cdmihai: True, that is indeed wise. :)

src/UpdateLocalizedResources.targets
cdmihai: I also had a philosophical debate with myself about this one. Can I do `BeforeTargets="$(PrepareResourceNamesDependsOn)"`? I really want these two targets (that inject and create resource items and files) to be the very first, since most likely all the other targets in that property will depend on them.

Also, I want to reference the property and not peek inside it to see what's the first target there, since that feels a bit like breaking encapsulation, and other projects could preppend other targets in that string, making my first choice not "the first one" anymore.

src/UpdateLocalizedResources.targets
cdmihai: Potential money wise issues with translation. I'll go ahead and ask the loc team if it's alright.

src/UpdateLocalizedResources.targets
cdmihai: I didn't know whether task outputs are additive, or whether they overwrite the item type with new contents. Seemed safer and more explicit this way. Do they overwrite, or do they add? :)

build/LocalizationTasks/EmitLocalizedResources.cs
cdmihai: I tried writing the msbuild code for it but after 1-2 hours of failed attempts and very dubious looking hacked up property functions, I succumbed to something more elegant, C# :)

build/LocalizationTasks/EmitLocalizedResources.cs
cdmihai: For explicit inputs, I would have to pass in another item of allowed language codes (`en;cs;de;ja; etc`) and pre-compute the xlf names from those. But then `en` is special, since there are no xlfs for it, and I'd have to have special code for it. Because of that, I went with automatic discover via IsValidXlf. One alternative is to not include `en` in the item of language codes. But then `en` resources still popup via another target, making things more confusing.

build/LocalizationTasks/SaveXlfToResx.cs
cdmihai: Good question. My reasoning was on the lines of: "I have no idea what exceptions might get transitively thrown from xliff parser, so then instead of letting a potential exception break out of the task, I'd rather log it as an error and have the task return false".

build/LocalizationTasks/SaveXlfToResx.cs
cdmihai: Kind of, because I have to edit the in-memory xlf and then ask it to serialize to resx. If I take it out, it will actually be the entire thing and this task would remain a copy but without the testing code in it. 

The alternative is to implement the localization testing functionality in a separate task and have it parse the generated resx and edit that. But that's more work. Do you think the testing functionality should be done at the resx level, via resx resource readers / writers?

src/Utilities/Microsoft.Build.Utilities.csproj
cdmihai: I specifically added them in to make the csproj easier to read. I'll take them out if I'm breaking some styling convention :)

src/UpdateLocalizedResources.targets
cdmihai: I had tried but bailed out of the change because:
- it required splitting a target in multiple pieces (some parts needed to output items indifferent of incremental check), making things more ugly
- a task from a target also depends on a property (doing localized test builds), besides files. If I make the target incremental, you have to do a clean build if you want to transform the satellites into test satellites.

src/Utilities/Microsoft.Build.Utilities.csproj
rainersigwald: I care (a bit) about the trailing space characters. Bare newlines for paragraph spacing I wholeheartedly approve of.

(I'm cursed by my ongoing "need" to have visible whitespace on, so I see this sort of things.)

build/LocalizationTasks/SaveXlfToResx.cs
rainersigwald: Yeah, that's reasonable. I was thinking more along the lines of factoring most of this out and having two tasks call into the base code, one with editing. But that's the architecture :rocket: in me. This seems fine.

build/LocalizationTasks/SaveXlfToResx.cs
rainersigwald: The standard advice, which I'm mostly just parroting here, is that catching `Exception` is too general, because it catches 
* Interesting runtime failures that might be actionable (what you want)
* Runtime failures due to programmer error that are bugs but not actionable for the end user (_maybe_ ok, might be better represented elsewhere)
* Systemic failures due to process or machine state (when you mostly just want to crash)

MSBuild itself already [provides an all-purpose task-exception handler](https://github.com/Microsoft/msbuild/blob/4b30fac51ab58dadfffedb2a57e83b8398e3a5e1/src/XMakeBuildEngine/BackEnd/Components/RequestBuilder/TaskBuilder.cs#L773-L902), so I'd say to just depend on that if there's no `XliffParserException` that's specific to the code that's being called. Plus, it doesn't look like there's [much in the way of exceptions](https://github.com/fmuecke/XliffParser/search?utf8=%E2%9C%93&q=throw) in there, currently.

Short further reading: https://msdn.microsoft.com/en-us/library/ms182137.aspx
LONG but interesting further reading: [Midori error model](http://joeduffyblog.com/2016/02/07/the-error-model/)

build/LocalizationTasks/EmitLocalizedResources.cs
rainersigwald: Is this correct?

* The list of languages is canonically determined by files in the repo.
* There's also an implicit `en` resource that needs to be treated differently.
* If you had a list of languages, you'd have to reconstruct file paths from the language (it's not a guaranteed map)?

That might be enough to convince me. But could you just do `<XlfFiles Include="*.xlf" />` or something?

src/UpdateLocalizedResources.targets
rainersigwald: > some parts needed to output items indifferent of incremental check

I'm not sure I understand. Can you elaborate?

> test satellites

That's definitely interesting. Maybe `Touch` an additional output if there you're doing test satellites? That would require a clean build to switch back, still.

src/UpdateLocalizedResources.targets
rainersigwald: I'm convinced. Can you put that reasoning in the comment or commit message?

src/UpdateLocalizedResources.targets
rainersigwald: It's additive, but I'll be darned if I can find that documented anywhere! #805

src/UpdateLocalizedResources.targets
rainersigwald: Yes, you can depend on the property, and I think that makes sense. You might do `BeforeTargets="$(PrepareResourceNamesDependsOn);PrepareResourceNames"` in case it's empty/conditioned out.

src/Utilities/Microsoft.Build.Utilities.csproj
cdmihai: okay, gonna scan for spurious white-spaces. I should enable white-space visibility as well :)

src/UpdateLocalizedResources.targets
cdmihai: will do

src/UpdateLocalizedResources.targets
cdmihai: > > some parts needed to output items indifferent of incremental check
> I'm not sure I understand. Can you elaborate?

The following target:
```xml
  <Target Name="UpdateLocalizedResxForEnglishResx">

      <ItemGroup>
        <EnglishResources Include="@(NeutralResources -> '$(IntermediateOutputPath)%(Filename).en%(Extension)')"/>
      </ItemGroup>

      <Copy
        SourceFiles="@(NeutralResources)"
        DestinationFiles="@(EnglishResources)"
      />

      <ItemGroup>
        <EmbeddedResource Include="@(EnglishResources)">
          <LogicalName>$(AssemblyName).%(FileName).resources</LogicalName>
        </EmbeddedResource>
      </ItemGroup>      

  </Target>
```

Both produces the intermediary english resx, and adds them to EmbeddedResource. To make copy incremental, I have to extract the item emitting part to another target that always runs after this one. Or, I can move item emitting code in the previous target that emits other items, but then I have to make sure the incremental english resx target runs before that one, so the file specific transforms to compute the logical name work. OK, now that I talked through it it looks feasible :)

> That's definitely interesting. Maybe Touch an additional output if there you're doing test satellites? That would require a clean build to switch back, still.
Aha, make a file lock as an input that gets touched in a previous target if the non-file state is the right one? I see, clever trick :)

Ok then, nailed both problems. Incremental targets it is :)

build/LocalizationTasks/EmitLocalizedResources.cs
cdmihai: That's correct. In my task, the main inputs are the neutral resources. From their localization I compute the root of xlf and then enumerate them. And then from the xlf file names I generate the intermediary localized resx paths from.

The problem with my approach is that if you delete one xlf file, nothing tells the build is invalid (it just enumerates files that match a pattern). So for better correctness it would be nice to also have an explicit list of language codes and throw if the xlf of any is missing.

> But could you just do `<XlfFiles Include="*.xlf" />` or something?
I guess I could do `<XlfFiles Include="@(NeutralResource->'%Root%Directory\xlf\*xlf')" />`. And then compute the intermediary resx off of that.

cdmihai: Actually that would be a one to many transformation. I'll see if msbuild supports that

build/LocalizationTasks/SaveXlfToResx.cs
cdmihai: New bed-time reading :)

I guess it makes sense overall. I guess we can also beef up documentation for task writers on what's the contract around exceptions and error handling. I remember the msbuild book saying something on the lines of "tasks should avoid throwing exceptions at all costs". Hence I caught everything and transformed it into a log error :)

src/XMakeTasks/Strings.resx
cdmihai: I tried, but bailed after ~15min. Some of the edits are in files that I also moved, making everything a pain. I vote for not splitting that monster commit :)

rainersigwald: :ok: Thanks for trying :)

cdmihai:

cdmihai:
Wrote localization documentation: https://github.com/Microsoft/msbuild/wiki/Dealing-With-Localization

rainersigwald:
Documentation 👍 

LGTM.

PR #780, opened 2016-07-14
Fix mono/OSX build - missing `Microsoft.DotNet.BuildTools.TestSuite`

radical:
.. nuget . This can be seen if you clean the global/user and local nuget
cache and then build:

  Microsoft.Build.Framework -> /Users/ankit/dev/msbuild/bin/x86/OSX/Debug-MONO/Output/Microsoft.Build.Framework.dll
MSBUILD : error : The package Microsoft.DotNet.BuildTools.TestSuite with version 1.0.0-prerelease-00508-01 could not be found in /Users/ankit/.nuget/packages. Run a NuGet package restore to download the package. [/Users/ankit/dev/msbuild/targets/DeployDependencies.proj]

### DIFF
--- a/targets/runtimeDependencies/project.json
+++ b/targets/runtimeDependencies/project.json
@@ -14,7 +14,8 @@
       "dependencies": {
         "xunit.runner.visualstudio": "2.1.0",
         "System.Runtime.InteropServices.RuntimeInformation": "4.0.0",
-        "System.Threading.Tasks.Dataflow": "4.6.0"
+        "System.Threading.Tasks.Dataflow": "4.6.0",
+        "Microsoft.DotNet.BuildTools.TestSuite": "1.0.0-prerelease-00508-01"
       }
     },
     "netcoreapp1.0": {


cdmihai:
LGTM.

I see that the TestSuite package is fairly new (from June). Most likely they split up their package in two.

rainersigwald:
Oddly, I don't see a break on current xplat without this change (I deleted `~/.nuget` first--maybe that's not enough?). Can you be more explicit about your repro steps?

(the change looks fine, but I hate to add a direct reference to something that should be referenced transitively if we don't have to)

rainersigwald:
@radical Any idea what might be up with my lack of failure?

radical:
No :/ After you earlier comment, I tried to reproduce it again and I couldn't. Then after a few days I hit that again! I am trying it now and I am able to hit it, on repeated tries too.

`msbuild $ rm -Rf ~/.nuget bin Tools packages`

.. and then I build. I don't know why you are not seeing this.

rainersigwald:
How very odd! I tried again yesterday and it worked fine from clean. Then I tried again this morning and it failed in the exact same way you're seeing.

It looks like this won't conflict with #835, so let's do it.

PR #783, opened 2016-07-18
LOC CHECKIN | MSBuild 20160718 | XLFs localized

v-zbsail:


cdmihai:
Thank you for the contribution!

I tested out the localized xlfs, and they work.

Some notes:
- I see that the xlfs under https://github.com/Microsoft/msbuild/tree/master/src/XMakeTasks/ManifestUtil/Resources/xlf are not in this PR. Is there a specific reason for this?
- In the meantime, we introduced one new string (in the shared xlf). Do you want us to merge the xlfs you provided now, and then you make a separate, new PR just for the new string? Or do you want to modify the current PR to include the new string (after we commit the updated xlf). The first options lets us test the pipeline of what happens when just one new string gets updated.

PR #785, opened 2016-07-18
Fix temp path issue

### LINKED ISSUE #606: If TEMP is set to a bare drive, FileTracker may throw
Ran across http://stackoverflow.com/questions/3221565/microsoft-build-utilities-filetracker-threw-an-exception-error-happens-with-dif while looking for something. It looks like that code still exists.

akoken:
When TEMP/TMP environment variable points to a drive root folder (D:\) without any further folder nesting, TEMP path results in null. This fix allows TEMP path to point to a drive root and also fixes #606.

### DIFF
--- a/src/Utilities/TrackedDependencies/FileTracker.cs
+++ b/src/Utilities/TrackedDependencies/FileTracker.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Build.Utilities
     {
         #region Static Member Data
         // The default path to temp, used to create explicitly short and long paths
-        private static string s_tempPath = Path.GetDirectoryName(Path.GetTempPath());
+        private static string s_tempPath = Path.GetTempPath();
 
         // The short path to temp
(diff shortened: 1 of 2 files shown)


rainersigwald:
@akoken Thanks!

The contents of the change looks good to me, but I'd like to see some improved metadata. Can you please:
* Make sure the committer address is associated with your GitHub account for proper credit (I currently see this:
image)?
* Make the commit description a bit more meaningful so we can understand it in 6 months. Maybe something like `Allow TEMP path to point to a drive root`?
* Use one of the [magic bug-fixing phrases](https://help.github.com/articles/closing-issues-via-commit-messages/) to associate this commit with the bug?

AndyGerlicher:
LGTM with the changes. Thanks for the contribution!

akoken:
Sorry for inconvenience @rainersigwald . It should be fine now. Thanks.

rainersigwald:
No problem, @akoken.  Thanks for the fix!

LGTM.

PR #789, opened 2016-07-18
Import design time targets for Roslyn Project System

natidea:
Conditionally import design time targets that will be installed with the next Preview of Visual Studio 15. These targets override CPS project properties schema to enable functionality introduced by the [Roslyn Project System](https://github.com/dotnet/roslyn-project-system).

/cc @AndyGerlicher @srivatsn @davkean 

For reference:
- Roslyn Project System [Design Time targets](https://github.com/dotnet/roslyn-project-system/tree/SigningTesting/src/Targets)
- Work Item: https://github.com/dotnet/roslyn-project-system/issues/128

### DIFF
--- a/src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
@@ -361,6 +361,12 @@ using System.Reflection%3b
         <_ExplicitReference Include="$(FrameworkPathOverride)\mscorlib.dll" />
     </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


srivatsn:
👍

rainersigwald:
Should the conditional import also include a condition for whether the targets will be run (something like `$(BuildingInsideVisualStudio)`?

If that doesn't seem like a good idea, LGTM.

srivatsn:
In the future we may want other hosts to get the properites\items in the design time targets as well. So BuildingInsideVisualStudio may not be strictly correct.

PR #792, opened 2016-07-19
Avoid XmlLicenseTransform in AuthenticodeSignLicenseDom

### LINKED ISSUE #791: SignFile task failure: Issuer node is required
[Mirrored from an internal bug]

We’re getting the following error when trying to use MSBuild’s SignFile task: An error occurred while signing: Issuer node is required. This is happening in the MSBuild that is being shipped with VS in project Willow, which is how I first ran into the issue.  I tried debugging it last week with MSBuild’s master and I managed to reproduce it simply by running the SignFile task by itself (proj and manifest files attached). The repro steps would be:

•Place both files in the same directory
•Replace the thumbprint with one from a local certificate
◦I created a test certificate from within VS by going into any project Properties à Signing à Create Test Certificate (figure attached)
◦From there I got the thumbprint from More Details
•Run msbuild on test.proj

```XML
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
<Target Name="Build">
<SignFile
CertificateThumbprint="2197530A7E887D6E86CB25D7B42D4D8AA2EF1D5F"
SigningTarget="$(MSBuildThisFileDirectory)\ExcelAddIn2.dll.manifest"
TargetFrameworkVersion="v4.5.2"
/>
</Target>
</Project>
```

I was debugging MSBuild and the exception was being thrown by this line: mansign2.cs:1425. The call stack points to XmlLicenseTransform.LoadInput and for .NET 4.6.1 it looks like it was coming from this line: transform.cs:1421. If I understand correctly, it seems like the license MSBuild is generating (SignedCmiManifest.CreateLicenseDom) has a different format from expected (at least for that transform). Not sure if it helps, but I also have a dump file.

rainersigwald:
Resolves #791

This line was #ifed out in the VS repo for quite a long time, but the
initial GitHub code drop was missing the condition. Restoring it to
prevent a regression in signing scenarios.

### DIFF
--- a/src/XMakeTasks/ManifestUtil/mansign2.cs
+++ b/src/XMakeTasks/ManifestUtil/mansign2.cs
@@ -1415,7 +1415,9 @@ namespace System.Deployment.Internal.CodeSigning
 
             // Add an enveloped and an Exc-C14N transform.
             reference.AddTransform(new XmlDsigEnvelopedSignatureTransform());
-            reference.AddTransform(new XmlLicenseTransform());
+#if (false) // BUGBUG: LTA transform complaining about issuer node not found.
+            reference.AddTransform(new XmlLicenseTransform()); 
+#endif
             reference.AddTransform(new XmlDsigExcC14NTransform());
 
             // Add the reference.


AndyGerlicher:
LGTM

PR #793, opened 2016-07-19
Update xlf files with new string

cdmihai:
New string appeared between xlf creation and when localization team translated them. This commit syncs the xlf against the latest resource state on master.

### DIFF
--- a/src/Shared/Resources/xlf/Strings.shared.cs.xlf
+++ b/src/Shared/Resources/xlf/Strings.shared.cs.xlf
@@ -269,6 +269,11 @@
         <target state="translated">Tato kolekce je jen pro čtení.</target>
         <note />
       </trans-unit>
+      <trans-unit id="Shared.CanNotFindValidMSBuildLocation">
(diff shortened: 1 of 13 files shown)


rainersigwald:
LGTM, but I have some process questions.

What keeps these up to date going forward? Do we need to do anything special when we add a new string? What about changing an existing one?

cdmihai:
Here is an overview of the localization process so far. I'll write a more detailed wiki on it.

When sufficient changes have accumulated, and\or before a VS release, we update the loc team to perform localization. While the loc team is translating, we should not update our xlf since that might lead to races and missed updates (they translate a string that updates while translating).

So here's the process on our end:
1. CI does localized builds all the time
2. We only do a localized build and commit the xlf file changes **before** notifying the loc team.
3. We do not do localized builds and commit the xlf while the loc team did not return from our last notification. Synchronous process :)

You can view this commit as step 2.

cdmihai:

PR #797, opened 2016-07-20
Modify identical unit test. This also closes #729.

akoken:
ExpressionShredder_Tests.ExtractItemVectorExpressionsSingleExpression7
and
ExpressionShredder_Tests.ExtractItemVectorExpressionsSingleExpression8
were identical. I modified
ExpressionShredder_Tests.ExtractItemVectorExpressionsSingleExpression8
and added seperator to expression. This also closes #729.

PR #798, opened 2016-07-20
Update MyGet feed URLs

rainersigwald:
Pretty sure the stale feed URLs were causing CI failures like in #797.

Not 100% sure why we'd need the corefx feed in master, but I suspect it's for transitive dependencies from buildtools.

### DIFF
--- a/src/.nuget/NuGet.Config
+++ b/src/.nuget/NuGet.Config
@@ -5,8 +5,8 @@
   </packageRestore>
   <packageSources>
     <clear />
-    <add key="myget.org buildtools" value="https://www.myget.org/F/dotnet-buildtools/" />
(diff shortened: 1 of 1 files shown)


cdmihai:
Good assumption. On the xplat branch the packages.config was replaced with its project.json equivalent, so its lock file should elucidate on who needs stuff from the core feed.

According to [this](https://github.com/Microsoft/msbuild/commit/41f95849cf3cdb2432174210754025ee70bd905f) commit, it came from Roselyn :)

cdmihai:
LGTM. Though I'd be curious if the build still succeeds without the corefx feed. Though it is irrelevant once master becomes xplat, since that one will surely have the feed.

PR #802, opened 2016-07-20
Add MSBuildTaskHost to MSBuild.sln

rainersigwald:
MSBuildTaskHost.csproj has been building since 2e2298a, but it wasn't
added to the solution, so it didn't show up in Visual Studio.

I also removed the AnyCPU configuration for MSBuild.exe and
MSBuildTaskHost.exe, since they should both always be built for an
explicit architecture.

@AndyGerlicher objection?


rainersigwald:

cdmihai:
LGTM

PR #803, opened 2016-07-20
Add BuildEnvironmentHelper support for Blend.exe

AndyGerlicher:
- When running in Blend.exe, treat as if we're in Visual Studio.
   Previously we could not find Visual Studio and ended up setting the
   value to null.

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -18,9 +18,9 @@ namespace Microsoft.Build.Shared
         };
 
         /// <summary>
-        /// Name of the Visual Studio process(es)
+        /// Name of the Visual Studio (and Blend) process.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -18,9 +18,9 @@ internal class BuildEnvironmentHelper
         };
 
         /// <summary>
-        /// Name of the Visual Studio process(es)
+        /// Name of the Visual Studio (and Blend) process.
         /// </summary>
-        private static readonly string[] s_visualStudioProcess = {"DEVENV"};
+        private static readonly string[] s_visualStudioProcess = {"DEVENV", "BLEND"};
```
rainersigwald: Well, at least it was already multi-valued!

rainersigwald:
LGTM

PR #808, opened 2016-07-21
Remove workaround fallback for _CppCommonExtensionTargets

AndyGerlicher:
This was added as a temporary workaround. Removing fallback from that
import property.


rainersigwald:
lgtm

Issue #732, opened 2016-06-27
Enhance OM: Project.GetItemProvenance


cdmihai:
Parent issue: #700 

Goal: Track down the history of what happened to an item (did it come from a glob, where was it updated, where was it excluded, etc)

Design so far: 
- Inputs:
 - string itemValue
   - String that represents an item
    - [?] Should there be any normalization done on the name (e.g. relative vs full path)
  - string itemType
    - Scope down the search by only looking at item tags of a certain type
  - item
    - The ProjectItem to get the value and type from. Also, start the search from this item's tag upwards
- Return
  - Project evaluation order sorted list of item tags that did something to that item (included, excluded, updated, removed)
- Remarks
  - GetItemProvenance(string) and GetItemProvenance(string, string) are flow insensitive, they scan the entire item list of a Project. GetItemProvenance(ProjecItem) is flow sensitive, it starts the search from the underlying ProjectItemElement upwards
  - Globs are not exploded, but rather pattern matching is used to see if the item fits a glob. This enables calling this method for a newly added file without having to re-evaluate the project
  - Indirect item references are ignored
<Compile Include="@(Sources);$(ItemLiteralsInsideAProperty)">. If this contains the item "Program.cs", then calling GetItemProvenance("Program.cs") won't return anything for this tag
     - [?] how to deal with a ProjectItem that may come from multiple include fragments? The OM does not link ProjectItem to what portion of the include string generated it.
  - Since an item can come from multiple sources inside an include (e.g., from glob AND literals) the Provenance enums can be aggregated (Provenance.FromGlob | Provenance.FromLiteral)

```
List<ProvenanceResult> Project.GetItemProvenance(string itemValue)
List<ProvenanceResult> Project.GetItemProvenance(ProjectItem item)
List<ProvenanceResult> Project.GetItemProvenance(string itemValue, string itemType)

struct ProvenanceResult
{
	ProjectItemElement
	Operation
	Provenance
}

enum Operation
{
	Include
	Exclude
	Update
	Remove
}

[Flags]
Provenance
{
	FromGlob //"*.cs"
	FromLiteral //"Program.cs"
	FromParametirizedLiteral //"Program$(Extension)"
	//FromItem
	//FromProperty
	//Inconclusive
}
```

The indirect reference problem:

```
1.   <PropertyGroup>
2.     <ItemsInString>*.a</ItemsInString>
3.     <ItemsInString>1.a;2.a;3.a;$(ItemsInString)</ItemsInString>
[...]

PR #809, opened 2016-07-21
Get item provenance

cdmihai:
Addresses #732 

This implementation:
- is performed after evaluation => lacks flow accuracy of inter-item references
- appends `Provenance.Inconclusive` to the result when the match is inside an indirect property / item reference
- does not expand any globs

Since the expander / evaluator do not keep a trace of what happened, this implementation expands the string twice and diffs the results to understand what happened and to give accurate occurrence results.

Also, this implementation is a temporary one. It will get refactored when it has to consider Update and Remove. And there's a chance it will get removed and re-implemented on top of lazy evaluation to gain flow accuracy. And then re-implemented again if / when we have scanners, parsers, and interpreters over the itemspec context-sensitive mini-language.

### DIFF
--- a/src/XMakeBuildEngine/Definition/Project.cs
+++ b/src/XMakeBuildEngine/Definition/Project.cs
@@ -2529,6 +2529,12 @@ namespace Microsoft.Build.Evaluation
                 get { return _items; }
             }
 
+            public List<ProjectItemElement> EvaluatedItemElements
+            {
+                get;
+                private set;
+            }
+
             /// <summary>
             /// List of items that link the XML items and evaluated items,
             /// evaluated as if their conditions were true.
@@ -2758,6 +2764,7 @@ namespace Microsoft.Build.Evaluation
                 this.AllEvaluatedProperties = new List<ProjectProperty>();
                 this.AllEvaluatedItemDefinitionMetadata = new List<ProjectMetadata>();
                 this.AllEvaluatedItems = new List<ProjectItem>();
+                this.EvaluatedItemElements = new List<ProjectItemElement>();
 
                 if (_globalPropertiesToTreatAsLocal != null)
                 {
--- a/src/XMakeBuildEngine/Definition/Project.cs
+++ b/src/XMakeBuildEngine/Definition/Project.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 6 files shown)


src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2415,6 +2416,251 @@ public void ImportPropertyEvaluatingToInvalidPath()
...
+        private static void AssertProvenanceResult(List<Tuple<string, Operation, Provenance, int>> expected, string project, string itemValue, int position)
+        {
+            var p = ObjectModelHelpers.CreateInMemoryProject(project);
+            var item = p.Items.Where(i => i.EvaluatedInclude.Equals(itemValue)).ElementAt(position);
+
+            var provenanceResult = p.GetItemProvenance(item);
+            AssertProvenanceResult(expected, provenanceResult);
+        }
+
+        private static void AssertProvenanceResult(List<Tuple<string, Operation, Provenance, int>> expected, List<ProvenanceResult> actual)
```
rainersigwald: Why tuple instead of a ProvenanceResult? If they were the same type, there's an `Assert.Equal` overload that takes two `IEnumerable<T>`.

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2415,6 +2416,251 @@ public void ImportPropertyEvaluatingToInvalidPath()
...
+
+        [Fact]
+        public void GetItemProvenanceOnlyStringLiteral()
+        {
+            var project =
+                @"<Project ToolsVersion='msbuilddefaulttoolsversion' DefaultTargets='Build' xmlns='msbuildnamespace'>
+                  <ItemGroup>
+                    <A Include=`1;2;3`/>
+                    <B Include=`1;2;3` Exclude=`1`/>
+                    <C Include=`2;3` Exclude=`2`/>
```
rainersigwald: What's the expected behavior if this had been Exclude=1? In other words, does the API return "any item element in which this path was mentioned" or "any item element in which this path could have been excluded"?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1032,6 +1033,170 @@ public static string GetEvaluatedItemIncludeEscaped(ProjectItemDefinition item)
             return ((IItem)item).EvaluatedIncludeEscaped;
         }
 
+        public List<ProvenanceResult> GetItemProvenance(string itemValue)
```
rainersigwald: Doc comments for all of these please. I may also want `remarks` sections explaining use cases.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1032,6 +1033,170 @@ public static string GetEvaluatedItemIncludeEscaped(ProjectItemDefinition item)
...
+                return 0;
+            }
+
+            // look into the itemspec as if it were expanded by the Expander
+            Provenance provenanceFromExpandedPropertiesAndItems;
+            var expandedMatches = ItemMatchesInSpec(itemToMatch, expand(itemSpec, ExpanderOptions.ExpandPropertiesAndItems), out provenanceFromExpandedPropertiesAndItems);
+
+            // look into the itemspec when properties are expanded, but not items
+            Provenance provenanceFromExpandedProperties;
+            ItemMatchesInSpec(itemToMatch, expand(itemSpec, ExpanderOptions.ExpandProperties), out provenanceFromExpandedProperties);
```
rainersigwald: Maybe only do this computation inside the if below?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1032,6 +1033,170 @@ public static string GetEvaluatedItemIncludeEscaped(ProjectItemDefinition item)
...
+            // look into the raw itemspec
+            Provenance provenanceFromNonExpandedString;
+            var nonExpandedMatches = ItemMatchesInSpec(itemToMatch, itemSpec, out provenanceFromNonExpandedString);
+
+            if (expandedMatches > nonExpandedMatches)
+            {
+                // return the number of occurences when properties AND items are expanded to get the correct occurence count
+
+                // return the provenance WITHOUT item expansion. Otherwise the items coming from a referenced item get interpreted as StringLiteral
+                // include="*.cs;@(Compile)" needs to return Inconclusive|Glob and not Inconclusive|Glob|StringLiteral
```
rainersigwald: I don't think I understand this.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1032,6 +1033,170 @@ public static string GetEvaluatedItemIncludeEscaped(ProjectItemDefinition item)
...
+
+            return occurrences;
+        }
+
+        private bool ItemMatchesStringLiteral(string itemToMatch, string itemFragment)
+        {
+            var thisProjectPath = _data.Directory;
+
+            // It is either a direct string match or the two strings refer to the same file, relative to the project directory
+            return itemToMatch.Equals(itemFragment) || string.Equals(FileUtilities.GetFullPath(itemToMatch, thisProjectPath), FileUtilities.GetFullPath(itemFragment, thisProjectPath));
```
rainersigwald: What about case sensitivity?

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
cdmihai: ProvenanceResult points to a ProjectItemElement which is hard to mock :(
I'll look into putting it to null.

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
cdmihai: Right now it's the former, without any real cause other than it felt "good at the time". I'll run it with CPS.

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: If I match `itemToMatch` against the expanded string with `ExpanderOptions.ExpandPropertiesAndItems`, then I will get bad data from items coming inside an item reference.

For example, computing GetItemProvenance("Program.cs"):
```
a include="*.cs"
b include="@(a)"
```
For `a`, I want to get `Provenance.FromGlob` and Occurrences=1
For `b` I want to get `Provenance.Inconclusive` and Occurrences=1

However, for `b`, if I expand items, I will get `Provenance.Inconclusive | Provenance.FromLiteral`. This is because the expander loads the items for `a` from its already evaluated items, so all the globs will have been expanded and I only get string literals.

Ideally, for `b`, I'd want GetItemProvenance to return actual nodes from some AST / interpretation dag that matched the string. But until we have such a structure, I am going to not flow information from indirect item references and add  `Inconclusive` to the bit flag. This is accomplished by matching against a [Property only expansion](https://github.com/Microsoft/msbuild/pull/809/files#diff-fc487db8bc5bcfe97c6c0996f8b14fefR1114).

For properties though, the expander gives them as is (does not expand globs, or follow item reference). So for properties, I am actually providing 'correct' provenance for literals and globs, but I still add `Inconclusive`, to signify that a reference has been traversed to get that information.

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: I'll ask CPS :)

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
rainersigwald: Definitely not a big deal as long as there's a reason. Maybe worth a comment though.

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
cdmihai: Got an answer. Excludes should not match if Includes did not match.

cdmihai:
Implemented feedback and fixed one bug. Ready for next round :)

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2415,6 +2418,425 @@ public void ImportPropertyEvaluatingToInvalidPath()
...
+            var expected = new ProvenanceResultTupleList
+            {
+                Tuple.Create("A", Operation.Exclude, Provenance.Glob, 1)
+            };
+
+            AssertProvenanceResult(expected, project, @"bin\1.cs");
+        }
+
+        [Fact]
+        public void GetItemProvenanceShouldHandleComplexGlobMissmatch()
```
rainersigwald: typo: mismatch

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2415,6 +2418,425 @@ public void ImportPropertyEvaluatingToInvalidPath()
...
+            {
+                Tuple.Create("B", Operation.Include, Provenance.StringLiteral, 1),
+                Tuple.Create("A", Operation.Include, Provenance.Inconclusive | Provenance.Glob, 3)
+            };
+
+            AssertProvenanceResult(expected, project, "1");
+        }
+
+        [Fact]
+        public void GetItemProvenanceShouldRespectItemConditions()
```
rainersigwald: This is interesting--is CPS on board with this behavior? It's nice and consistent, but I think I can imagine reasons to want to know provenance for a file that's not in the current configuration.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1033,6 +1034,232 @@ public static string GetEvaluatedItemIncludeEscaped(ProjectItemDefinition item)
...
+            Func<IElementLocation, Func<string, ExpanderOptions, string>> expandForXmlLocation = (l) => (s, o) => expander.ExpandIntoStringLeaveEscaped(s, o, l);
+
+            var includeResult = ComputeProvenanceResult(itemToMatch, itemElement.Include, expandForXmlLocation(itemElement.IncludeLocation));
+
+            if (includeResult == null)
+            {
+                return null;
+            }
+
+            var excludeResult = ComputeProvenanceResult(itemToMatch, itemElement.Exclude, expandForXmlLocation(itemElement.ExcludeLocation));
```
rainersigwald: Looks like both uses of `expandForXmlLocation` use the same arg. Can you bake that in and avoid the currying?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1033,6 +1034,232 @@ public static string GetEvaluatedItemIncludeEscaped(ProjectItemDefinition item)
...
+        ///     - we have no proper AST and interpreter for itemspecs that we can do analysis on
+        ///     - GetItemProvenance needs to have correct counts for exclude strings (as correct as it can get while doing it after evaluation)
+        /// 
+        /// The temporary hack is to use the expander to expand the strings, and if any property or item references were encountered, return Provenance.Inconclusive
+        /// </summary>
+        /// <param name="itemToMatch"></param>
+        /// <param name="itemSpec"></param>
+        /// <param name="expand"></param>
+        /// <param name="provenance"></param>
+        /// <returns></returns>
```
rainersigwald: No need for empty doc comments for params.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -3142,4 +3377,46 @@ internal string GetPropertyValue(string name)
             }
         }
     }
+
+    /// <summary>
+    /// Bit flag enum that specifies how a string representing an item matched against an itemspec.
+    /// Undefined is the bottom element and should not appear in actual results
+    /// Inconclusive means that the match is indirect, coming from either property or item references.
+    /// </summary>
```
rainersigwald: Might be nicer to put `<summary>` tags on the values themselves.

rainersigwald:
:shipit:

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
cdmihai: Good point. I'll ask. At the time I wrote it made sense to not look into false condition, but I do see your point.

AndyGerlicher:
:shipit:

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
cdmihai: Talked with CPS. The current behaviour is OK for now. At some point (post rtm?) they will want the capability to choose whether conditions are respected or not.

cdmihai:

PR #815, opened 2016-07-25
Remove invalid check in CommandLineBuilder

cameron314:
Take two of my previous pull request #419 (closed due to inactivity on my part, sorry)! This version includes a unit test.

When escaping quotes, a check was done to ensure that the number of quotes that will be escaped is even. This doesn't make sense, since there's plenty of command line parsers (including Windows's own CreateProcess) that correctly process any number of embedded quotes (as long as they're escaped properly).

One way that the effects of this check can be seen at a user-interface level is by attempting to define a macro in the project properties for a C++ file which has in its body an odd number of quotes (e.g. a string with an embedded quote). For example, add FOO="\"" to the C++ Preprocessor Definitions in the property pages and observe the MSB6001 error that results when you attempt to build.

### DIFF
--- a/src/Utilities/Resources/xlf/Strings.cs.xlf
+++ b/src/Utilities/Resources/xlf/Strings.cs.xlf
@@ -18,11 +18,6 @@
         <target state="translated">Řetězec v uvozovkách v hodnotě příkazového řádku [{0}] je neplatný.</target>
         <note />
       </trans-unit>
-      <trans-unit id="General.StringsCannotContainOddNumberOfDoubleQuotes">
-        <source>The value [{0}] contains an odd number of double-quote characters. Only even numbers of literal double-quote characters are acceptable to command line tools.</source>
-        <target state="translated">Hodnota [{0}] obsahuje lichý počet znaků uvozovek. Nástroje příkazového řádku připouštějí pouze sudý počet znaků uvozovek použitých jako literálů.</target>
-        <note />
-      </trans-unit>
       <trans-unit id="General.ToolCommandFailedNoErrorCode">
         <source>The command exited with code {0}.</source>
         <target state="translated">Příkaz byl ukončen s kódem {0}.</target>
--- a/src/Utilities/Resources/xlf/Strings.cs.xlf
+++ b/src/Utilities/Resources/xlf/Strings.cs.xlf
@@ -18,6 +18,11 @@
         <target state="translated">Řetězec v uvozovkách v hodnotě příkazového řádku [{0}] je neplatný.</target>
         <note />
       </trans-unit>
+      <trans-unit id="General.StringsCannotContainOddNumberOfDoubleQuotes">
+        <source>The value [{0}] contains an odd number of double-quote characters. Only even numbers of literal double-quote characters are acceptable to command line tools.</source>
+        <target state="translated">Hodnota [{0}] obsahuje lichý počet znaků uvozovek. Nástroje příkazového řádku připouštějí pouze sudý počet znaků uvozovek použitých jako literálů.</target>
+        <note />
+      </trans-unit>
       <trans-unit id="General.ToolCommandFailedNoErrorCode">
         <source>The command exited with code {0}.</source>
         <target state="translated">Příkaz byl ukončen s kódem {0}.</target>
--- a/src/Utilities/UnitTests/CommandLineBuilder_Tests.cs
+++ b/src/Utilities/UnitTests/CommandLineBuilder_Tests.cs
@@ -414,8 +414,8 @@ namespace Microsoft.Build.UnitTests
         public void AppendSwitchWithOddNumberOfLiteralQuotesInParameter()
         {
             CommandLineBuilder c = new CommandLineBuilder();
-            c.AppendSwitchIfNotNull("/D", @"A='""'");
-            Assert.Equal(@"/D""A='\""'""", c.ToString());
+            c.AppendSwitchIfNotNull("/D", @"A='""'");     //   /DA='"'
+            Assert.Equal(@"/D""A='\""'""", c.ToString()); //   /D"A='\"'"
         }
 
         internal class TestCommandLineBuilder : CommandLineBuilder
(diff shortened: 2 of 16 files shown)


src/Utilities/Resources/xlf/Strings.cs.xlf
```diff
@@ -18,11 +18,6 @@
         <target state="translated">Řetězec v uvozovkách v hodnotě příkazového řádku [{0}] je neplatný.</target>
         <note />
       </trans-unit>
-      <trans-unit id="General.StringsCannotContainOddNumberOfDoubleQuotes">
-        <source>The value [{0}] contains an odd number of double-quote characters. Only even numbers of literal double-quote characters are acceptable to command line tools.</source>
-        <target state="translated">Hodnota [{0}] obsahuje lichý počet znaků uvozovek. Nástroje příkazového řádku připouštějí pouze sudý počet znaků uvozovek použitých jako literálů.</target>
-        <note />
-      </trans-unit>
```
rainersigwald: @cdmihai My understanding is that these files shouldn't be updated by a change like this that removes a string from the `.resx`. Is that correct?

cameron314: Ah? How are they updated then? I just assumed that since they were under version control they were not automatically managed by any tools.

src/Utilities/UnitTests/CommandLineBuilder_Tests.cs
```diff
@@ -405,19 +405,19 @@ public void AppendSwitchUnQuotedTaskItem()
...
         {
-            Assert.Throws<ArgumentException>(() =>
-            {
-                CommandLineBuilder c = new CommandLineBuilder();
-                c.AppendSwitchIfNotNull("/D", @"ASSEMBLY_KEY_FILE=""c:\\foo\\FinalKeyFile.snk");
-            }
-           );
+            CommandLineBuilder c = new CommandLineBuilder();
+            c.AppendSwitchIfNotNull("/D", @"A='""'");
+            Assert.Equal(@"/D""A='\""'""", c.ToString());
```
rainersigwald: Quote escaping rules here are hurting my brain! For my own edification:

`@"A='""'" == "A='\"'"` is `A='"'`

`@"/D""A='\""'""" == "/D\"A='\\\"'\""` is `/D"A='\"'"`

Which makes sense, since `"` is in the [`definitelyNeedQuotesRegexNoHyphen`](https://github.com/Microsoft/msbuild/blob/70db3602ec75e0e4c50b613eb6588e158d5dfd9f/src/Utilities/CommandLineBuilder.cs#L135) list.

src/Utilities/Resources/xlf/Strings.cs.xlf
rainersigwald: @cameron314 It's an odd mix of tool-augmented version-controlled files. @cdmihai is working on this stuff so soon we'll at least have a how-to wiki entry.

src/Utilities/UnitTests/CommandLineBuilder_Tests.cs
cameron314: Oops, yeah, sorry about that. I adopted the style of the other tests that all seem to use verbatim string literals :P

rainersigwald: I started typing this comment to say "use regular string literals" but halfway through translating them I decided they weren't any clearer.

Maybe it's worth adding the no-escape strings as a comment, though? It's definitely character soup in there.

cameron314: Fair point, coming right up.

src/Utilities/Resources/xlf/Strings.cs.xlf
cameron314: Interesting. Let me know if you need me to make any changes.

src/Utilities/UnitTests/CommandLineBuilder_Tests.cs
rainersigwald: Oh yeah, I think that's definitely worth it! Thanks.

src/Utilities/Resources/xlf/Strings.cs.xlf
cdmihai: You can remove the changes to the xlf files and only keep the ones in the .resx.

The xlf files will automatically get updated from their corresponding neutral .resx files during a localized build, once #779 goes in.

cdmihai:
Instead of completely removing the check, would it be safer to change it so it only counts non-escaped quotes? i.e., don't count `quote ::= ' | "` when the character before it is `\`.

My reason for this is build safety. Since the build will do irreversible things to your file system, we should try and protect against malformed command lines which might do undefined stuff to the file system.

src/Utilities/Resources/xlf/Strings.cs.xlf
cameron314: OK, done!

cameron314:
@cdmihai: No, this would not be correct either. Take for instance the example I put in the unit test, which passes the string `'"'` to a tool -- it's completely valid, yet contains only a single double quote, which is not escaped. The `CommandLineBuilder` does not know what tool it is invoking, nor with what command line parameters (since they often come from user input), and hence cannot impose general limitations like this. All syntactically valid strings must be properly escaped and passed on to the tool.

Anyway, there's plenty of ways to mess up a build that do not involve quotes :-)

cameron314:
To clarify further, the job of the `AppendQuotedTextToBuffer` method is to take an opaque string and produce the command-line-escaped equivalent such that when that string makes its way into a tool, it comes back to the original string. If the escaping is done properly, there should not be any possibility of corrupting the command line regardless of the contents of the string. There may be certain limitations due to the nature of the command line, but quotes are not one of them since they can always be escaped.

rainersigwald:
I tend to agree with @cameron314's analysis, @cdmihai. I'd also rather not attempt to reimplement a combination of `sh` and `cmd.exe` and CRT escaping rules here--just not trying seems like a better approach.

cdmihai:
OK, sold :)

:shipit:

rainersigwald:
Thanks, @cameron314!

cameron314:
Great, thanks everyone!

PR #817, opened 2016-07-25
Directory item metadata should be relative

### LINKED ISSUE #813: %(Directory) should always be unrooted
When working with an ItemGroup on Linux one can get a fully qualified path to a resource via `@(IG->'%(Directory)')`. This results in a rooted path e.g. `/path/to/my/item`.

On Windows, the same expression would result in `path\to\my\item` without the leading drive qualifier, therefore providing a non-fully-qualified path to a resource. Obtaining a fully qualified path on Windows requires `@(IG->'%(RootDir)%(Directory)')`.

I'm not familiar with the intention of `%(Directory)` omitting the drive specifier, but the discrepancy was a bit jarring when building a build script on OS X and expecting it to work. I'm fairly familiar with MSBuild so I knew to look at the MSDN documentation for well-known metadata and to identify a fix, but folks less familiar are likely to spend considerable time understanding what went awry. 

It seems that parity might mean that Windows `%(Directory)` values should lead with a `\` to specify that the path is relative to the drive root, with `%(RootDir)` specifying the drive `X:` only. This should provide an equivalent experience between Windows and non-Windows platforms. If that is not possible because it is a breaking change then we may want to consider also requiring `%(RootDir)` on non-Windows platforms as well so that folks are faced with this challenge consistently across all platforms.

rainersigwald:
Fixes #813

On Windows, `%(Directory)` strips the drive letter-colon-backslash
prefix from full paths. For unixy paths, the same meaning isn't exactly
clear, but there's a known root (`/`) that could be stripped
equivalently.

After this change, the correct metadata reference to get full directory
is `%(RootDir)%(Directory)` on Windows (as it always has been) and on
other OSes (as Mono's xbuild has been doing, and as seems most
consistent).

### DIFF
--- a/src/Shared/Modifiers.cs
+++ b/src/Shared/Modifiers.cs
@@ -466,19 +466,33 @@ namespace Microsoft.Build.Shared
                         GetItemSpecModifier(currentDirectory, itemSpec, definingProjectEscaped, ItemSpecModifiers.FullPath, ref fullPath);
 
                         modifiedItemSpec = GetDirectory(fullPath);
-                        Match root = FileUtilitiesRegex.DrivePattern.Match(modifiedItemSpec);
 
-                        if (!root.Success)
+                        if (NativeMethodsShared.IsWindows)
                         {
-                            root = FileUtilitiesRegex.UNCPattern.Match(modifiedItemSpec);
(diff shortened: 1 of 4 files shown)


AndyGerlicher:
:shipit:

PR #824, opened 2016-07-26
Fix SignFile task for Dev15 ClickOnce signed publishing

vladbarosan:
- Update search path to include the new location for the SignTool installed by a singleton MSI. This is necessary as otherwise SignTool comes from Windows SDK which is a heavyweight component which is not justified to be included as a dependency just for the SignTool.
- Update condition for continuing the search to include the actual existence of the file in case the path is not null as there are scenarios in Dev15 in which the "toolPath" to be found but the signtool is not present.

### DIFF
--- a/src/XMakeTasks/ManifestUtil/SecurityUtil.cs
+++ b/src/XMakeTasks/ManifestUtil/SecurityUtil.cs
@@ -762,9 +762,12 @@ namespace Microsoft.Build.Tasks.Deployment.ManifestUtilities
         {
 #pragma warning disable 618 // Disabling warning on using internal ToolLocationHelper API. At some point we should migrate this.
             string toolPath = ToolLocationHelper.GetPathToWindowsSdkFile(ToolName, TargetDotNetFrameworkVersion.VersionLatest, VisualStudioVersion.VersionLatest);
-            if (toolPath == null)
-                toolPath = ToolLocationHelper.GetPathToWindowsSdkFile(ToolName, TargetDotNetFrameworkVersion.Version45, VisualStudioVersion.Version110);
-            if (toolPath == null)
+            if (toolPath == null || !File.Exists(toolPath))
+            {
+                toolPath = ToolLocationHelper.GetPathToWindowsSdkFile(ToolName, TargetDotNetFrameworkVersion.Version45,
+                    VisualStudioVersion.Version110);
+            }
+            if (toolPath == null || !File.Exists(toolPath))
             {
                 var pathToDotNetFrameworkSdk = ToolLocationHelper.GetPathToDotNetFrameworkSdk(TargetDotNetFrameworkVersion.Version40, VisualStudioVersion.Version100);
                 if (pathToDotNetFrameworkSdk != null)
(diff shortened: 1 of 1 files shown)


src/XMakeTasks/ManifestUtil/SecurityUtil.cs
```diff
@@ -815,5 +828,23 @@ private static bool IsCertInStore(X509Certificate2 cert)
             }
             return false;
         }
+
+        private static string GetVersionIndependentToolPath(string toolName)
+        {
+            RegistryKey localMachineKey = Registry.LocalMachine;
+            string versionIndependentToolKeyName = @"Software\Microsoft\ClickOnce\SignTool";
```
rainersigwald: const?

src/XMakeTasks/ManifestUtil/SecurityUtil.cs
```diff
@@ -762,20 +762,33 @@ internal static string GetPathToTool(System.Resources.ResourceManager resources)
...
                 var pathToDotNetFrameworkSdk = ToolLocationHelper.GetPathToDotNetFrameworkSdk(TargetDotNetFrameworkVersion.Version40, VisualStudioVersion.Version100);
                 if (pathToDotNetFrameworkSdk != null)
                 {
                     toolPath = Path.Combine(pathToDotNetFrameworkSdk, "bin", ToolName);
                 }
             }
-            if (toolPath == null)
+            if (toolPath == null || !File.Exists(toolPath))
+            {
+                toolPath = GetVersionIndependentToolPath(ToolName);
```
rainersigwald: Should this be the *last* fallback, or the first? Right now you prefer the Windows SDK, which seems to make sense--but I want to be explicit about it.

vladbarosan: I would say it makes sense to have it last I would say, not first as we want this to be only of last resort when there is no SDK installed. I left the check for the users current directory after this one as the error message will point to the file missing from his project directory and it would make more sense to go online, download it and put it there instead of some "random" folder from his perspective

src/XMakeTasks/ManifestUtil/SecurityUtil.cs
```diff
@@ -762,20 +762,33 @@ internal static string GetPathToTool(System.Resources.ResourceManager resources)
         {
 #pragma warning disable 618 // Disabling warning on using internal ToolLocationHelper API. At some point we should migrate this.
```
vladbarosan: @rainersigwald  not sure if this pragma warning is still needed. The methods from **ToolLocationHelper** are public (not internal ). It is still showing a warning as the methods used from **ToolLocationHelper** are marked with Obsolete but I couldn't find the recommended method **GetPlatformSDKLocationFile**, can you share some light on this ?

vladbarosan:

src/XMakeTasks/ManifestUtil/SecurityUtil.cs
rainersigwald: I don't know the story on either. Can you file a bug for us to figure it out? Since you're not making things worse I think this is fine.

src/XMakeTasks/ManifestUtil/SecurityUtil.cs
rainersigwald: Sounds good, just wanted to check.

rainersigwald:
LGTM. @Microsoft/msbuild-maintainers ?

src/XMakeTasks/ManifestUtil/SecurityUtil.cs
vladbarosan: Sure. you guys prefer an issue here or a bug in VSTS  ?

rainersigwald: Here, please.

vladbarosan: Done

AndyGerlicher:
LGTM

PR #825, opened 2016-07-26
Fix mono/osx build: The "real" targets file corresponding to the shim

radical:
.. targets, might not be available like on OSX, so make the imports
conditional on their existence.


rainersigwald:
LGTM

AndyGerlicher:
LGTM

PR #827, opened 2016-07-26
Fix mono/osx tests

radical:


src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -3294,20 +3290,33 @@ public void Medley()
...
+#if FEATURE_WIN32_REGISTRY
+            if (NativeMethodsShared.IsWindows)
+            {
+                errorTests.Add("$(Registry:X)");
+            }
+            else
+            {
+                // If no registry, and not running on windows, this gets expanded to the empty string
+                // example: xplat build running on OSX
+                validTests.Add(new string[]{"$(Registry:X)", ""});
```
rainersigwald: Should this be inside the `#if` guard? Or is it true for netcore and mono both?

src/XMakeTasks/UnitTests/GenerateResourceOutOfProc_Tests.cs
```diff
@@ -2313,6 +2313,8 @@ public void Regress25163_OutputResourcesContainsInvalidPathCharacters()
     public class References
     {
         [Fact]
+        //FIXME: mono: looks for csc.exe
```
rainersigwald: Can you file an MSBuild bug and put the URL here?

src/XMakeBuildEngine/UnitTests/Evaluation/Expander_Tests.cs
radical: Hm the existing code tested this case only for !netcore. But this should probably be tested for netcore too. It seems to resolve to `""` even for netcore, so, I will update this.

radical:
Updated. Thanks for the review!

rainersigwald:
LGTM

AndyGerlicher:
:shipit:

PR #831, opened 2016-07-27
Update app.config for C++ target default properties

AndyGerlicher:
- Add trailing '\' on VCTargetsPath
 - Add AdroidTargetsPath and iOSTargetsPath per C++ team. This matches
   values that were previously set by optional installed components.


rainersigwald:
:shipit:

PR #832, opened 2016-07-27
Initialize .NET 3.5 key name before use

rainersigwald:
In the initial xplat work for MSBuild, these were changed from being
compile-time constants to static fields, because they now have conditional
behavior dependent on whether the code is running on Windows.

That introduced a subtle bug: `s_dotNetFrameworkRegistryKeyV35` needs to
know `dotNetFrameworkVersionFolderPrefixV35`, but was textually before it.
Since static field initialization is
[executed in the textual order in which they appear in the class
declaration](https://msdn.microsoft.com/en-us/library/aa645758.aspx), the
registry key got the bogus value (with a null string appended to it
instead of the .NET 3.5 version string).


jeffkl:
LGTM

AndyGerlicher:
:shipit:

Issue #834, opened 2016-07-28
UnGACed MSBuild can fail type checks when a task loads MSBuild assemblies from the amd64 folder


rainersigwald:
The same type, when loaded in a new appdomain in an amd64 MSBuild process, will be loaded from the copy of the DLL in the amd64 subfolder, not the one from bin. This means that checks for type equality will fail, because the types being compared may be from different physical assemblies on disk (even if they're bitwise identical).

This is new to the Willow experience, because MSBuild was always in the GAC before--so any attempt to load an MSBuild assembly would load it from the GAC, rather than the path that was requested. That ensured that there was no mismatch.

One possible symptoms of this:

```
The "GenerateResource" task failed unexpectedly.
Microsoft.Build.Shared.InternalErrorException: MSB0001: Internal MSBuild Error: Somehow got a wrong and possibly incompatible type for ProcessResourceFiles.
   at Microsoft.Build.Shared.ErrorUtilities.ThrowInternalError(String message, Exception innerException, Object[] args)
   at Microsoft.Build.Tasks.GenerateResource.Execute()
   at Microsoft.Build.CommandLine.OutOfProcTaskAppDomainWrapperBase.InstantiateAndExecuteTask(IBuildEngine oopTaskHostNode, LoadedType taskType, String taskName, String taskLocation, String taskFile, Int32 taskLine, Int32 taskColumn, AppDomainSetup appDomainSetup, IDictionary`2 taskParams)
```

rainersigwald:
Long fixed.

PR #836, opened 2016-07-28
Set codeBase for MSBuild assemblies in x64 MSBuild

rainersigwald:
Initial fix for #834. Setting `codeBase` to ensure that all MSBuild
assemblies get loaded from the same location (in the MSBuild bin folder)
even if they're loaded from the amd64 MSBuild.exe.

This isn't a great solution--the app.config files should be generated from a common
source, because they should be identical except for the codeBase. Also, we
probably shouldn't even ship copies of the assemblies into the amd64
folder if they'll never be used. But this is an urgent bug so I'm
providing a quick workaround to be followed up with a long-term fix.

### DIFF
--- a/src/XMakeCommandLine/app.amd64.config
+++ b/src/XMakeCommandLine/app.amd64.config
@@ -1,0 +1,79 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+  <configuration>
+    <configSections>
+      <section name="msbuildToolsets" type="Microsoft.Build.Evaluation.ToolsetConfigurationSection, Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+    </configSections>
+    <startup useLegacyV2RuntimeActivationPolicy="true">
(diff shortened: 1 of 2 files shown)


src/XMakeCommandLine/app.amd64.config
```diff
@@ -0,0 +1,79 @@
...
+      <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6" />
+    </startup>
+    <runtime>
+      <DisableFXClosureWalk enabled="true" />
+      <generatePublisherEvidence enabled="false" />
+      <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+        <dependentAssembly>
+          <assemblyIdentity name="Microsoft.Build.Framework" culture="neutral" publicKeyToken="b03f5f7f11d50a3a" />
+          <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.1.0.0" />
+          <codeBase version="15.1.0.0" href="..\Microsoft.Build.Framework.dll"/>
```
rainersigwald: Hard to see in this view but this is a copy of app.config with these `codeBase` lines added.

cdmihai:
LGTM

Issue #342, opened 2015-11-09
xplat CI builds should build full-framework and netcore MSBuild


rainersigwald:
Currently, the `RebuildWithLocalMSBuild.cmd` script that is the CI entry point only builds the `Debug-netcore` configuration of MSBuild in the xplat branch.  In preparation for branch unification, we need to make sure  MSBuild is working correctly in full-framework mode in that branch as well.

AndyGerlicher:
Lets rename that to something like cibuild.cmd though :)

rainersigwald:
This has been done for a while. We should add Mono soon.

PR #842, opened 2016-07-29
Introduce cibuild.cmd

rainersigwald:
📝 My plan is to get this in, set up a failing CI build, and then make it pass. I have that mostly working on my machine, so now it's time to see if Jenkins agrees with me.

This is a port of cibuild.sh, because RebuildWithLocalMSBuild didn't
support arguments and cibuild.sh is a better name. The code was initially
copied from RebuildWithLocalMSBuild.cmd.

Needed for #342.

### DIFF
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -140,6 +140,6 @@ exit /b 1
 :error
 echo.
 echo ---------------------------------------
-echo - RebuildWithLocalMSBuild.cmd FAILED. -
+echo - cibuild.cmd FAILED. -
 echo ---------------------------------------
 exit /b 1
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -6,7 +6,7 @@ if "%1"=="" goto doneParsingArguments
 if /i "%1"=="--scope" set SCOPE=%2&& shift && shift && goto parseArguments
 if /i "%1"=="--target" set TARGET=%2&& shift && shift && goto parseArguments
 if /i "%1"=="--host" set HOST=%2&& shift && shift && goto parseArguments
-if /i "%1"=="--skip-bootstrap" set BOOTSTRAP_ONLY=true&& shift && goto parseArguments
+if /i "%1"=="--bootstrap-only" set BOOTSTRAP_ONLY=true&& shift && goto parseArguments
 
 :: Unknown parameters
 goto :usage
@@ -134,7 +134,7 @@ echo Options
 echo   --scope ^<scope^>                Scope of the build ^(Compile / Test^)
 echo   --target ^<target^>              CoreCLR or Desktop ^(default: Desktop^)
 echo   --host ^<host^>                  CoreCLR or Desktop ^(default: Desktop^)
-echo   --skip-bootstrap               Do not rebuild msbuild with local binaries
+echo   --bootstrap-only               Do not rebuild msbuild with local binaries
 exit /b 1
 
 :error
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -48,14 +48,11 @@ if not defined HOST (
 )
 
 set RUNTIME_HOST=
-set HOST_SPECIFIED=
 if /i "%HOST%"=="CoreCLR" (
     set RUNTIME_HOST=%~dp0Tools\CoreRun.exe
     set MSBUILD_CUSTOM_PATH=%~dp0Tools\MSBuild.exe
-    set HOST_SPECIFIED=true
 ) else if /i "%HOST%"=="Desktop" (
     set RUNTIME_HOST=
-    set HOST_SPECIFIED=true
 ) else (
     echo Unsupported host detected: %HOST%. Aborting.
     goto :error
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -11,8 +11,6 @@ if /i "%1"=="--skip-bootstrap" set BOOTSTRAP_ONLY=true&& shift && goto parseArgu
 :: Unknown parameters
 goto :usage
 
-set CONFIGURATION=Debug-NetCore
-
 :doneParsingArguments
 
 if "%SCOPE%"=="Compile" (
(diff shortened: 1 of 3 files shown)


cibuild.cmd
```diff
@@ -0,0 +1,140 @@
...
+if /i "%1"=="--skip-bootstrap" set BOOTSTRAP_ONLY=true&& shift && goto parseArguments
+
+:: Unknown parameters
+goto :usage
+
+set CONFIGURATION=Debug-NetCore
+
+:doneParsingArguments
+
+if "%SCOPE%"=="Compile" (
```
jeffkl: Case insensitive comparison with /i ?

rainersigwald: I went this way because `cibuild.sh` doesn't do insensitive, but `/i` would definitely be the more Windows-y way. Should we prefer that?

cibuild.cmd
```diff
@@ -0,0 +1,140 @@
...
+set CONFIGURATION=
+if "%TARGET%"=="CoreCLR" (
+    set CONFIGURATION=Debug-NetCore
+)
+if "%TARGET%"=="Desktop" (
+    set CONFIGURATION=Debug
+)
+
+if not defined CONFIGURATION (
+    echo Unsupported target detected: %TARGET%. Configuring as if for Desktop
```
jeffkl: In my opinion, this should stop executing in this case...

cibuild.cmd
jeffkl: It'd be pretty frustrating if the script won't run because I forgot to proper case an argument value when you're doing case-insensitive comparison of the argument names above

cibuild.cmd
rainersigwald: Yeah, I think the ideal behavior is to special case "not specified" and abort on "specified but not understood".

josemesona: Okay well then technically you should use IF ... ELSE IF ... ELSE against %TARGET% and then bail out right?

---
In reply to: [72823020](https://github.com/Microsoft/msbuild/pull/842#discussion_r72823020) [](ancestors = 72823020)

cibuild.cmd
```diff
@@ -1,5 +1,148 @@
...
+@if not defined _echo echo off
+setlocal
 
-echo cibuild.cmd invoked from the xplat branch--calling RebuildWithLocalMSBuild.cmd
+:parseArguments
+if "%1"=="" goto doneParsingArguments
+if /i "%1"=="--scope" set SCOPE=%2&& shift && shift && goto parseArguments
+if /i "%1"=="--target" set TARGET=%2&& shift && shift && goto parseArguments
+if /i "%1"=="--host" set HOST=%2&& shift && shift && goto parseArguments
+if /i "%1"=="--skip-bootstrap" set BOOTSTRAP_ONLY=true&& shift && goto parseArguments
```
cdmihai: Renaming --skip-bootstrap to --bootstrap-only might make more sense (in both scripts). Bootstrap being the phase where a downloaded msbuild is used to build the first binaries.

cibuild.cmd
```diff
@@ -1,5 +1,148 @@
...
+    set TARGET=Desktop
+)
+
+set BUILD_CONFIGURATION=
+if /i "%TARGET%"=="CoreCLR" (
+    set BUILD_CONFIGURATION=Debug-NetCore
+) else if /i "%TARGET%"=="Desktop" (
+    set BUILD_CONFIGURATION=Debug
+) else (
+    echo Unsupported target detected: %TARGET%. Aborting.
```
cdmihai: No mono targetting on windows?

cibuild.cmd
rainersigwald: Agreed. I was thinking about doing that in a future commit. Right now the external interface matches between this and cibuild.sh.

cibuild.cmd
```diff
@@ -1,5 +1,148 @@
...
+
+set RUNTIME_HOST=
+set HOST_SPECIFIED=
+if /i "%HOST%"=="CoreCLR" (
+    set RUNTIME_HOST=%~dp0Tools\CoreRun.exe
+    set MSBUILD_CUSTOM_PATH=%~dp0Tools\MSBuild.exe
+    set HOST_SPECIFIED=true
+) else if /i "%HOST%"=="Desktop" (
+    set RUNTIME_HOST=
+    set HOST_SPECIFIED=true
```
cdmihai: Should this be false? Is HOST_SPECIFIED used anywhere?

cibuild.cmd
rainersigwald: Could and maybe should be added. I'm pretty sure our CI machines don't have Mono installed so the second phase of the build wouldn't work as is. Not sure if it's worth the effort. @radical -- thoughts?

cibuild.cmd
rainersigwald: It's a remnant of an older approach. I should remove it.

cibuild.cmd
```diff
@@ -1,5 +1,148 @@
...
+if /i "%TARGET%"=="CoreCLR" (
+    set MSBUILD_CUSTOM_PATH="%~dp0bin\Bootstrap\MSBuild.exe"
+) else (
+    set MSBUILD_CUSTOM_PATH="%~dp0bin\Bootstrap\15.0\Bin\MSBuild.exe"
+)
+
+echo.
+echo ** Rebuilding MSBuild with locally built binaries
+
+call "%~dp0build.cmd" /t:RebuildAndTest /p:Configuration=%BUILD_CONFIGURATION%
```
cdmihai: >/p:Configuration=%BUILD_CONFIGURATION% [](start = 40, length = 38)

wonder if the msbuild configuration computation can be pushed into build.cmd, since it looks like all calls to build.cmd form it the same way

cdmihai:
Ideas for having localized builds on CI:
- pass through arguments, like the Rebuild script had
- put the msbuild args into a `--extra-msbuild-arguments` parameter: `cibuild.cmd --extra-msbuild-arguments "/p:LocalizedBuild=true /fl"`

cibuild.cmd
radical: We can build with .NET, targeting Mono (`/p:Configuration=Debug-MONO`) and test it with .net again.

rainersigwald:
@cdmihai I think I prefer the `--extra-MSBuild-arguments` approach. Can we wait on that, or do we need it in the initial checkin here?

cibuild.cmd
rainersigwald: Are you thinking of having `build.cmd` read `BUILD_CONFIGURATION` from the environment, or teaching it about an explicit arg?

cdmihai:
Microbuild does not use the scripts, so it's fine. The only issue is having a CI not doing localized builds for a while, but I guess that's OK. Or could you temporarily hardcode the `LocalizedBuild` property in the scripts?

rainersigwald:
Let's leave this as is for now, and when I've gotten the desktop build green we can figure out what to do about loc in xplat.

cdmihai:
Sounds good.

:shipit:

PR #846, opened 2016-07-29
Enable multi-runtime builds in CI infrastructure

rainersigwald:
Needed for #342.

This adds another dimension to the CI build definitions, so we can start
building both the .NET Core and Desktop framework versions of MSBuild out
of the xplat branch.

Since `netci.groovy` only creates jobs, it can't know what branch is being
targeted for a PR-validation job. So the validation script must exist
everywhere. Since master doesn't build for .NET Core, I'm adding a stub
cibuild.cmd that's just a shim for the existing CI build script. Builds in
master will trigger two full-framework builds, with one labeled "CoreCLR".
This should be a temporary situation since the branches should be unified
soon.

### DIFF
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -1,0 +1,5 @@
+@echo off
+
+echo cibuild.cmd invoked from the master branch--calling RebuildWithLocalMSBuild.cmd
+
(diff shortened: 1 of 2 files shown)


cdmihai:
:shipit:

PR #847, opened 2016-07-29
Stub cibuild.cmd to enable multi-runtime CI builds

rainersigwald:
Checking this into xplat so that cf10c8a9 has a cibuild.cmd script to
target in the xplat branch. As-is, it always builds for .NET Core, even if
asked to build for desktop. A follow-up, full-featured cibuild.cmd will be
capable of building both desktop and Core.


cdmihai:
:ship:

PR #849, opened 2016-08-01
Support for excluding patterns from file matching, and support for removing items outside of targets

dsplaisted:
This PR builds on #770.

It adds support to the file matching (globbing) code to allow a set of patterns to be specified that should be ignored when searching for files.  This makes it possible to efficiently ignore subdirectories which may have a large number of files (for example `node_modules`).

This PR also adds support for removing items during evaluation (previously it was only allowed inside of targets).  This uses the lazy evaluation and the newly added support for ignoring a set of patterns in the file matcher.  This should allow the following pattern, and be efficient no matter how many files are in the ignored subfolders:

```xml
<!-- Possibly inside the project file -->
<Compile Include="**\*.cs" />

<!-- Probably inside a targets file the project file imports -->
<Compile Remove="bin\**\*.*;obj\**\*.*;node_modules\**\*.*" />
```

More tests for this should be added.


src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -436,6 +436,10 @@
     <comment>{StrBegin="MSB4020: "}UE: This is a generic message that is displayed when we find a project element with an incorrect value for one of its
     attributes e.g. &lt;Import Project=""&gt; -- the value of Project should not be an empty string.</comment>
   </data>
+  <data name="InvalidAttributeExclusive" UESanitized="false" Visibility="Public">
+    <!-- TODO: Replace ???? with a real error code -->
+    <value>MSB????: At most one of the include, remove, and update attributes may be specified for an item element.</value>
```
AndyGerlicher: >????: [](start = 14, length = 5)

Can we pick one before we check this in? Or at least file an issue and put it in the RC milestone.

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
```diff
@@ -1441,6 +1441,45 @@ public void RemoveMetadataImported()
             }
            );
         }
+
+        //  TODO: Should remove tests go in project item tests, project item instance tests, or both?
```
cdmihai: Ideally both. Practically, the common Evaluator bahaviour can go into ProjectItem_Tests and the differences introduced by ProjectInstance in ProjectItemInstance_Tests. Like whether to consider conditions or not.

Sarabeth-Jaffe-Microsoft:
Closing this one in favor of #870.

PR #850, opened 2016-08-01
Build and test for Desktop framework in xplat

rainersigwald:
This builds on #842--comments on `cibuild.cmd` should go there.

On my machine, I get a mysterious failure loading `System.Runtime` _only_ in the Engine tests. I haven't figured out why yet. Will be interesting to see if the CI builds produce the same thing.


rainersigwald:

cibuild.cmd
```diff
@@ -1,5 +1,148 @@
...
+echo   --scope ^<scope^>                Scope of the build ^(Compile / Test^)
+echo   --target ^<target^>              CoreCLR or Desktop ^(default: Desktop^)
+echo   --host ^<host^>                  CoreCLR or Desktop ^(default: Desktop^)
+echo   --skip-bootstrap               Do not rebuild msbuild with local binaries
+exit /b 1
+
+:error
+echo.
+echo ---------------------------------------
+echo - RebuildWithLocalMSBuild.cmd FAILED. -
```
AndyGerlicher: >RebuildWithLocalMSBuild [](start = 7, length = 23)

cibuild.cmd FAILED?

AndyGerlicher:
:shipit:

cibuild.cmd
rainersigwald: Good catch. Fixed in #842.

rainersigwald:
I'd like to go in with this even though I haven't fixed the `System.Runtime` thing. It doesn't make the red CI any worse and raises the bar . . .

@dotnet-bot test this please

rainersigwald:

src/Shared/NativeMethodsShared.cs
```diff
@@ -613,17 +613,42 @@ internal static string FrameworkBasePath
...
+                            _systemInformationInitialized = true;
+                        }
+                    }
+                }
+                return _systemInformation;
+            }
+        }
+
+        private static SystemInformationData _systemInformation;
+        private static bool _systemInformationInitialized;
```
cdmihai: Make this volatile so threads don't work on cached values?

rainersigwald: [Allegedly](http://stackoverflow.com/a/1964832) unnecessary--if two threads get stale cached values, they'll both attempt to acquire the lock, one will succeed and pass the inner check, the other will block, then get the lock and fail the inner check.

cdmihai: OK, sold. Good enough for CLR :)

cdmihai:
:shipit:

PR #854, opened 2016-08-02
Allow MSBUILDDEBUGONSTART=2 for .NET Core

rainersigwald:
This was disabled when removing calls to `Debugger.Launch`, but in mode 2
`MSBUILDDEBUGONSTART` doesn't actually need that API and is .NET Core
friendly.


cdmihai:
:shipit:

PR #856, opened 2016-08-03
Updates to support XamlBuildTask

dmetzgar:
This PR is to fix an issue with VS "15" being unable to build workflow projects. Workflow projects use a custom MSBuild task called XamlBuildTask. XamlBuildTask creates a new AppDomain that attempts and fails to load the GAC version of MSBuild assemblies rather than the applocal ones in VS "15".

Microsoft.Xaml.targets in this version of MSBuild is a shim to load the .NET Framework version. The XamlBuildTaskPath MSBuild property is used as the ApplicationBase for the new AppDomains by XamlBuildTask. This PR sets this property to the applocal folder. In concert, we are placing XamlBuildTask.dll in the applocal folder of VS "15".

The other change is to apply an assembly binding redirect from the framework version (4.0.0.0) of XamlBuildTask to the VS "15" version (15.1.0.0). This is to ensure we are not loading the framework version.

### DIFF
--- a/src/XMakeTasks/Microsoft.Xaml.targets
+++ b/src/XMakeTasks/Microsoft.Xaml.targets
@@ -13,7 +13,7 @@
         the .NET Framework directory and thus will no longer be the right answer. Override it to point
         to the correct .NET Framework location. --> 
    <PropertyGroup>
-      <XamlBuildTaskPath Condition="'$(XamlBuildTaskPath)' == ''">$(MSBuildToolsPath)</XamlBuildTaskPath>
+      <XamlBuildTaskPath Condition="'$(XamlBuildTaskPath)' == ''">$(MSBuildThisFileDirectory)</XamlBuildTaskPath>
    </PropertyGroup>
 
    <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.Xaml.targets" />
--- a/src/XMakeTasks/Microsoft.Xaml.targets
+++ b/src/XMakeTasks/Microsoft.Xaml.targets
@@ -13,7 +13,7 @@
         the .NET Framework directory and thus will no longer be the right answer. Override it to point
         to the correct .NET Framework location. --> 
    <PropertyGroup>
-      <XamlBuildTaskPath Condition="'$(XamlBuildTaskPath)' == ''">$(MSBuildFrameworkToolsPath)</XamlBuildTaskPath>
+      <XamlBuildTaskPath Condition="'$(XamlBuildTaskPath)' == ''">$(MSBuildToolsPath)</XamlBuildTaskPath>
    </PropertyGroup>
 
    <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.Xaml.targets" />
--- a/src/XMakeCommandLine/app.amd64.config
+++ b/src/XMakeCommandLine/app.amd64.config
@@ -35,8 +35,12 @@
           <codeBase version="15.1.0.0" href="..\Microsoft.Build.Utilities.Core.dll"/>
         </dependentAssembly>
         <dependentAssembly>
+          <assemblyIdentity name="Microsoft.Activities.Build" culture="neutral" publicKeyToken="31bf3856ad364e35" />
+          <bindingRedirect oldVersion="4.0.0.0" newVersion="15.0.0.0" />
+        </dependentAssembly>
+        <dependentAssembly>
           <assemblyIdentity name="XamlBuildTask" culture="neutral" publicKeyToken="31bf3856ad364e35" />
-          <bindingRedirect oldVersion="4.0.0.0" newVersion="15.1.0.0" />
+          <bindingRedirect oldVersion="4.0.0.0" newVersion="15.0.0.0" />
         </dependentAssembly>
       </assemblyBinding>
     </runtime>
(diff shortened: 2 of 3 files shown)


src/XMakeTasks/Microsoft.Xaml.targets
```diff
@@ -13,7 +13,7 @@
         the .NET Framework directory and thus will no longer be the right answer. Override it to point
         to the correct .NET Framework location. --> 
    <PropertyGroup>
-      <XamlBuildTaskPath Condition="'$(XamlBuildTaskPath)' == ''">$(MSBuildFrameworkToolsPath)</XamlBuildTaskPath>
+      <XamlBuildTaskPath Condition="'$(XamlBuildTaskPath)' == ''">$(MSBuildToolsPath)</XamlBuildTaskPath>
```
AndyGerlicher: This doesn't quite make sense. This property would be `$(MSBuildToolsPath)\Microsoft.Xaml.targets` which when deployed and executing is _this_ file. If this file + that DLL is going to come from the VS installer, I don't think there's any change needed to this file. Ideally this could come from a NuGet package so we could enable this scenario outside of VS. But I'm fine if that's not the case for now.

AndyGerlicher:
Overall I think it looks ok to unblock, but I think it's worth doing something else with the "shim" target. Since it's pretty much guaranteed not to work, perhaps even removing the file entirely might be appropriate? That will cause builds to fail in a more obvious way when in this scenario. At minimum it should not try to import itself though, even if it's just left alone.

Also, can you please update the commit message? This is very odd and not at all obvious what's going on without more context, I don't want someone else to be completely confused in 6 months. It would be worth noting that this is in place so that VS can place the target/DLL in the correct location at setup time and the scenario is currently only supported from VS. And come to think of it I do think it's worth touching the "shim" file so it gets included in the commit for the message, even if you just add a comment that the file is now replaced.

Thanks!

rainersigwald:
Why is the binding redirect necessary? Doesn't the version in the `UsingTask` cause MSBuild to load the right version?

src/XMakeTasks/Microsoft.Xaml.targets
dmetzgar: The inclusion of the framework version of Microsoft.Xaml.targets is done on the import below this property group. The XamlBuildTaskPath is used as the ApplicationBase for AppDomains. Setting XamlBuildTaskPath to the framework directory causes the new XamlBuildTask AppDomains to attempt to load assemblies from the framework directory.

rainersigwald: Can you set this as a default inside `Microsoft.Xaml.targets`, using `$(MSBuildThisFileDirectory)` or something? That seems more robust (for example, it'd work if you packaged the task assembly in a NuGet package, which is desirable at some point).

AndyGerlicher: Sorry! I thought you were replacing the `Microsoft.Xaml.targets` file. When I read this I figured you were importing the one you were replacing. Sorry for the misunderstanding. Looks like you're only dropping the task dll and changing where it looks for it. Got it. This is definitely needed then.

dmetzgar:
Is the suggestion to replace the shim targets file with a copy of the original? We could do that but it's not necessary. It boils down to the same change to the XamlBuildTaskPath property. I like the shim because it's clearer what the delta is between the framework version and the VS 15 version when we're doing maintenance in the future.

rainersigwald:
@dmetzgar I still don't understand the need for the binding redirects. Can you explain? It's a high burden for people that ship apps that do programmatic builds to add your version redirects, so it'd be nice to avoid it. And as long as the task is getting loaded from a `.targets` file you control, you shouldn't need to redirect a task assembly.

dmetzgar:
@rainersigwald What we're seeing is that without the binding redirect both the local version and the GAC version get loaded into memory. @erikcai8, can you add more detail?

erikcai8:
@rainersigwald We did not take control of Microsoft.Xaml.targets in this change yet. It still redirects to the on released by NetFx. That's probably next step since we don't have enough time to fully test it in this VS Preview 4 release.
The reason why MSBuild would load 2 XamlBuildTask.dll in MSBuild without binary redirections is because "Microsoft.Xaml.targets" specifies to load the first XamlBuildTask.dll in GAC, and in the loaded XamlBuildTask.dll, it creates new app domains which would load the 2nd XamlBuildTask.dll in app-local.

rainersigwald:
@erikcai8 So the redirect is a temporary measure until you can update the targets file to load the new version of the DLL? That would be fine with me.

erikcai8:
@rainersigwald Yes, that's the plan if the binary redirection is not preferred.

src/XMakeTasks/Microsoft.Xaml.targets
dmetzgar: Ya, MSBuildThisFileDirectory is a better suggestion. Just pushed that change.

dmetzgar:
@AndyGerlicher @rainersigwald please have a look. We're hoping to get this in through shiproom today so there's one day to test before the end of escrow.

rainersigwald:
LGTM.  @AndyGerlicher ?

PR #857, opened 2016-08-03
Localized vsix packages

cdmihai:
Getting ready to turn on localized builds in microbuild and VS insertion.

This PR adds the satellite assemblies into the vsix and fixes up the build to eliminate a race


AndyGerlicher:
LGTM

PR #859, opened 2016-08-03
Pass dotnet host options to worker nodes

rainersigwald:
Multiprocess builds were hanging when called through `dotnet build3`,
because the child process spawned from `LaunchNode` failed immediately on
start with

```
A fatal error was encountered. The library 'hostpolicy.dll' required to
execute the application was not found
```

That was because MSBuild preserved the path to `dotnet`, but not the
additional arguments required for it to assemble a runtime environment.

This commit uses the property bag that the host provides to find the right
deps file. At the moment, there is no programmatic way to access the
runtimeconfig, so I'm relying on a heuristic--that it'll match the deps
file.

The property-bag access code is based on
https://github.com/dotnet/cli/blob/b7075b857b48b08908bd340ffd3f41f756318339/src/Microsoft.DotNet.Cli.Utils/Muxer.cs#L36-L40

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -12,6 +12,7 @@ using System.Globalization;
 using System.IO;
 using System.IO.Pipes;
 using System.Diagnostics;
+using System.Reflection;
 using System.Threading;
 using System.Runtime.InteropServices;
 using System.Security.Principal;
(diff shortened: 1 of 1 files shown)


rainersigwald:
@piotrpMSFT, @eerhardt--I have some questions!

Is this temporary?

Is there a bug or PR I can crosslink for the "programmatically get runtimeconfig" part?

piotrpMSFT:
@rainersigwald I'm working on a PR now. Once it's ready you won't need to get runtimeconfig anymore. Feel free to create an issue on /cli if you want a reference.

AndyGerlicher:
:shipit:

Issue #858, opened 2016-08-03
Nail down issues related to assembly loading


tmeschter:
This is related to issue #658. The fix for that issue is minimal, as we do a very simple examination of the files next to task assemblies when looking for dependencies.

There are a couple potential issues with this approach that we may want to address.

1. Assembly identity issues.
 Say we have the following four files on disk:

 - Alpha\AlphaTasks.dll
 - Alpha\Utility.dll
 - Beta\BetaTasks.dll
 - Beta\Utility.dll

 Assume the two copies of Utility.dll have the same simple name, but different versions (or public key tokens) and different contents. Since we only consider the simple name when trying to find and load a dependency, both AlphaTasks and BetaTasks will both end up using the same Utility--either the one from Alpha or the one from Beta--even though that is not what they expect to happen. Which one is selected will depend entirely on whether AlphaTasks or BetaTasks is loaded first.

 We could decide to take the full assembly name into account when satisfying dependencies, and thus ensure both AlphaTasks and BetaTasks get the copies of Utility.dll they expect.

2. Support for satellite assemblies.
 Right now if a task assembly or dependency needs to load a satellite assembly (to retrieve a localized resource, for example) we don't know to look for them in culture-specific subdirectories like en-US and de-DE.

 I'm not actually sure if this is an issue in .Net Core or if localized resources are handled in some other way.

eerhardt:
Another scenario that would be great to enable is when a Build Task assembly's dependencies come from other NuGet packages.

For example, in the .NET Core SDK Build Tasks we are dependent on a bunch of NuGet assemblies, and on Microsoft.Extensions.DependencyModel.  With the current support, we need to package all our dependencies into a single .nupkg.  But it would be great if our .nupkg could just list Microsoft.Extensions.DependencyModel in its "dependencies", and MSBuild knew to look in that NuGet package location as well.

cdmihai:
@eerhardt Are you saying that tasks should have a new dependency fallback search from a deps.json (or deps.runtime.json, whichever is the one with runtime dependencies), like .net core apps do?

eerhardt:
I'm not trying to design it.  I'm just listing the scenario/issue. Packaging all my dependencies into my .nupkg doesn't feel like the correct thing to do.

But yes, basically you can imagine a .deps.json and runtimeconfig.json approach that works like .net core apps do.

[...]

PR #861, opened 2016-08-03
Add support and tests for loading assembly dependencies

tmeschter:


tmeschter:
@anurse @rainersigwald Could you take a look at this, please?

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
```diff
@@ -0,0 +1,138 @@
...
+    public class TypeLoader_Dependencies_Tests
+    {
+        private static readonly string ProjectFilePath = Path.Combine(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory, "TaskWithDependencyTest.proj");
+        private static readonly string TaskDllFileName = "TaskWithDependency.dll";
+        private static readonly string OriginalTaskDllPath = Path.Combine(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory, TaskDllFileName);
+        private static readonly string DependencyDllFileName = "Dependency.dll";
+        private static readonly string OriginalDependencyDllPath = Path.Combine(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory, DependencyDllFileName);
+
+        [Fact]
+        public void LoadAssemblyAndDependency_Inside()
```
rainersigwald: Can you be a bit more explicit than "inside"/"outside"? I'm not sure a failure with this test name would be clear to me.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                UndoDLLOperations(tempDir, moveBack: true);
+            }
+        }
+
+        /// <summary>
+        /// </summary>
+        /// <param name="copy"></param>
+        /// <returns>Path to temporary directory</returns>
+        private string MoveOrCopyDllsToTempDir(bool copy)
```
rainersigwald: Is the copy option needed? Looks like you're only calling it one way.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
```diff
@@ -0,0 +1,138 @@
...
+                Assert.True(File.Exists(newTaskDllPath));
+                Assert.True(File.Exists(newDependencyDllPath));
+            }
+            else
+            {
+                File.Move(OriginalTaskDllPath, newTaskDllPath);
+                File.Move(OriginalDependencyDllPath, newDependencyDllPath);
+
+
+                Assert.True(File.Exists(newTaskDllPath));
```
rainersigwald: Nit: two blank lines

rainersigwald:
Can you amend the initial commit so that it doesn't have `testProject.proj`?

rainersigwald:
There'll be some ugly merges with #850 but I'll deal with them.

I kinda wish we didn't have to make new projects for the task with reference, but this is probably better than adding a spurious dependency to the existing sample.

LGTM.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
tmeschter: I was following a pattern in TypeLoader_Tests.cs, but I can clarify it.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
tmeschter: No, that's a result of copying functionality from TypeLoader_Tests.cs.

rainersigwald: Ahh. Not knowing those tests bites again! 😇

tmeschter:
Yeah, I'll take out testProject.proj.

I didn't really want to touch the samples, no.

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -97,6 +111,21 @@ protected override Assembly Load(AssemblyName assemblyName)
                     return assembly;
                 }
 
+                foreach (var dependencyPath in _dependencyPaths)
+                {
+                    foreach (var extension in _extensions)
+                    {
+                        var candidatePath = Path.Combine(dependencyPath, $"{assemblyName.Name}.{extension}");
+                        if (IsAssemblyAlreadyLoaded(candidatePath) ||
+                            !FileMatchesAssemblyName(candidatePath, assemblyName.Name))
```
anurse: We don't actually need `FileMatchesAssemblyName` since we're constructing the candidate path from the assembly name in the first place. We do need to test if the file exists however so we only load the one that matches (if we find one).

It only worked in my initial spike because of a fluke :) (the `.dll` was picked because it was the first extension that passed the `FileMatchesAssemblyName` test)

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -114,5 +143,15 @@ private Assembly LoadAndCache(string fullPath)
 
             return assembly;
         }
+
+        private bool IsAssemblyAlreadyLoaded(string path)
+        {
+            return _pathsToAssemblies.ContainsKey(path);
+        }
+
+        private bool FileMatchesAssemblyName(string path, string assemblySimpleName)
```
anurse: Won't be needed after the above comment is resolved. It was only called in one place

anurse:
One small comment (that was equally broken in my spike ;P), otherwise good to go 👍 !

tmeschter:
@anurse @rainersigwald I've updated the commits.

What's the correct etiquette once the tests pass and I have reviewer sign off? Can I go ahead and merge the request at that point?

rainersigwald:
@tmeschter We have a two-maintainer signoff policy. The second maintainer to sign off generally hits merge, unless the submitter was also a maintainer. @cdmihai can you be another set of eyes here?

rainersigwald:
Let's see how this does on the desktop build I just enabled. It shouldn't pass (until #850), but the new stuff should work.

@dotnet-bot test this please

rainersigwald:
^ ok, looks no worse to me. Hooray!

src/dirs.proj
```diff
@@ -17,6 +17,9 @@
     <Project Include="XMakeTasks\UnitTests\Microsoft.Build.Tasks.UnitTests.csproj" />
     <Project Include="XMakeBuildEngine\UnitTests\Microsoft.Build.Engine.UnitTests.csproj" />
     <Project Include="XMakeBuildEngine\UnitTestsPublicOM\Microsoft.Build.Engine.OM.UnitTests.csproj" />
+
+    <Project Include="Test\Projects\Dependency\Dependency.csproj" />
+    <Project Include="Test\Projects\TaskWithDependency\TaskWithDependency.csproj" />
```
cdmihai: The sample projects should go into this other dirs.proj: [https://github.com/Microsoft/msbuild/blob/xplat/Samples/dirs.proj](https://github.com/Microsoft/msbuild/blob/xplat/Samples/dirs.proj)

src/Shared/TypeLoader.cs
```diff
@@ -189,7 +189,8 @@ private static Assembly LoadAssembly(AssemblyLoadInfo assemblyLoadInfo)
                     // - otherwise, the assembly from the user specified path is loaded, if it exists.
                      
                     var assemblyNameInExecutableDirectory = Path.Combine(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory,
-                        Path.GetFileName(assemblyLoadInfo.AssemblyFile));
+
+                    Path.GetFileName(assemblyLoadInfo.AssemblyFile));
```
cdmihai: Funky formatting

cdmihai:
Can you please move the test projects into Samples? That's where the test samples went so far.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
```diff
@@ -0,0 +1,138 @@
...
+
+using System;
+using System.IO;
+using Microsoft.Build.Shared;
+using Microsoft.Build.SharedUtilities;
+using Xunit;
+
+namespace Microsoft.Build.UnitTests
+{
+    public class TypeLoader_Dependencies_Tests
```
cdmihai: Rather than adding this new test file and new test projects, could you instead modify PortableTask to add a dependency to it? That should test task dependencies without the need to add extra projects or tests.

src/Shared/TypeLoader.cs
tmeschter: @cdmihai As in, too many blank lines?

src/dirs.proj
tmeschter: @cdmihai So the projects under Samples aren't really samples per se, but projects for use in test runs?

src/Shared/TypeLoader.cs
cdmihai: Yes, and the second argument to Combine is not indented, making it look like a separate statement, which makes the Path.Combine look like invalid syntax.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
tmeschter: @cdmihai I could, but these tests cover a somewhat different scenario than TypeLoader_Tests--loading dependencies, rather than "top level" task assemblies. I generally prefer a larger number of smaller tests, rather than fewer tests that exercise multiple concepts.

src/dirs.proj
cdmihai: Both, unfortunately.

From what I can tell about msbuild's "convention" regarding test projects:
- Samples are like big, end 2 end smoke tests. But they are also examples that show how to do certain things. They are like big stakes in the ground. They really are not exhaustive :).
- Each test class defines its own convention on test samples. Sadly. Some just have a csproj as a literal string. Some construct entire test projects on the fly in some temp directory. When I first added tests to `Typeloader_Tests` to test task dll loading under xplat, I chose to reuse the PortableTask sample. In your case, it feels that the test project that you added might be of use as a reference for other people wanting to create tasks with dependencies, so putting it in Samples makes some sense.

src/Shared/UnitTests/TypeLoader_Dependencies_Tests.cs
cdmihai: When you put it like that, it makes sense :)

I guess one point of improvement would be to make some test utilities around task loading to reduce code duplication. But I'm fine without it as well.

cdmihai:
@rainersigwald @AndyGerlicher 

As far as you know, besides the ability of a task to load dependencies sitting next to it, does full framework MSBuild support other ways for tasks to load their dependencies (other than GAC :) )?

rainersigwald:
@cdmihai I think they could do whatever they wanted using the usual `AssemblyResolve` event handler, but the default find-assembly behavior is just .NET's defaults: https://msdn.microsoft.com/en-us/library/yx7xezcf(v=vs.110).aspx

cdmihai:
Mkay, so tasks should have the same dependency loading behaviour as a normal app, including the ability to provide their own app.configs and so on? Is it worth creating a new "up-for-grabs" issue to have tests for all those fallbacks? :)

Does this mean that, for .net core, tasks can have `deps.json` and `deps.runtime.json files`?

rainersigwald:
I don't think a DLL can drag along its own app.config. But it can install code in the AppDomain before loading, so it can alter loading behavior. I don't know if there's a way to stop that, but we can just not test and not recommend it :)

I'm still unclear on how tasks get integrated into MSBuild's deps.json loading environment. It's come up a few times but I don't think it's resolved.

tmeschter:
@cdmihai @rainersigwald I've opened #858 to track other issues related to assembly loading. It sounds like there are a couple more that should be added to that issue.

tmeschter:
@cdmihai I've updated the commits to fix that formatting issue and to move the test projects into the Samples directory.

Also, I see that the "Windows_NT Build for Desktop" is failing. Since this is going back into the xplat branch, is that a concern or can it be ignored? The failure doesn't seem to have anything to do with my change.

rainersigwald:
@tmeschter The Desktop build is ok. I'm fixing it in another PR and you don't seem to have made it worse.

cdmihai:
:shipit:

PR #863, opened 2016-08-03
Add support for assembly resolution from config file

AndyGerlicher:
The code is mostly copied from AssemblyFoldersEx and follows most of the
same logic. I didn't implement all the features and options from that
class as they don't seem to be used/needed.

In Dev15, Visual Studio setup will drop this file into MSBuild folder in
order for Add References dialog and Build to find reference assemblies.
This will allow for projects to reference, say Microsoft.Build.dll, without
providing a hint path. Previously this was done via the registry but Dev15
will no longer write those entries in all cases. When they are not in the
registry, MSBuild will use this file.


src/XMakeTasks/UnitTests/AssemblyDependency/AssemblyFoldersFromConfig_Tests.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                t.BuildEngine = new MockEngine();
+
+                t.Assemblies = new ITaskItem[] { new TaskItem("assemblyfromconfig_common.dll") };
+                t.SearchPaths = new[] { moniker };
+                t.TargetProcessorArchitecture = "x86";
+                var success = Execute(t);
+                Assert.True(success);
+
+                Assert.Equal(0, t.ResolvedFiles.Length);
```
rainersigwald: I assume this is mapped from the other tests--test for success and 0 length returns, no test for warnings?

Are you not failing on explicitly-passed-in-nonexistent-file because you've changed common targets to pass in the path unconditionally? That seems ok, but worth being explicit.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        /// <summary>
+        /// Construct.
+        /// </summary>
+        /// <param name="targetRuntimeVersion">The runtime version property from the project file.</param>
+        internal AssemblyFoldersFromConfig(string configFile, string targetRuntimeVersion, ProcessorArchitecture targetArchitecture)
+        {
+            if (!FileUtilities.FileExistsNoThrow(configFile))
+            {
+                return;
```
rainersigwald: Comment here why this is a no-failure early return.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+
+            var collection = AssemblyFolderCollection.Load(configFile);
```
rainersigwald: Since `Load` just uses deserialization, what's the malformed-input experience like? Should we catch and give a nicer error, or is an uncaught exception understandable?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+            if (targets.Any())
+            {
+                _directoryNames.AddRange(targets);
+            }
+        }
+
+        internal static List<AssemblyFolder> GatherVersionStrings(string targetRuntimeVersion, AssemblyFolderCollection collection)
+        {
+            var validTargets = new List<AssemblyFolder>();
+            Version targetVersion = VersionUtilities.ConvertToVersion(targetRuntimeVersion);
```
rainersigwald: Error cases here interesting?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            if (!FileUtilities.FileExistsNoThrow(configFile))
+            {
+                return;
+            }
+
+            var collection = AssemblyFolderCollection.Load(configFile);
+            var assemblyTargets = GatherVersionStrings(targetRuntimeVersion, collection);
+
+            bool is64BitOs = Environment.Is64BitOperatingSystem;
```
rainersigwald: Only used once. Inline?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+            bool targeting64Bit = targetArchitecture == ProcessorArchitecture.Amd64 ||
+                                  targetArchitecture == ProcessorArchitecture.IA64;
```
rainersigwald: Ah, this is pretty confusing, but this shouldn't be needed here. It's in `AssemblyFoldersEx` because you have to look in the WOW registry hive explicitly depending on your own bitness, but that won't be true of the config file.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+                        target => !string.IsNullOrEmpty(target.Platform) && target.Platform.Equals("x86", StringComparison.OrdinalIgnoreCase));
+                    FindDirectories(assemblyTargets,
+                        target => !string.IsNullOrEmpty(target.Platform) && target.Platform.Equals("x64", StringComparison.OrdinalIgnoreCase));
+                }
+            }
+            else
+            {
+                FindDirectories(assemblyTargets,
+                    target => !string.IsNullOrEmpty(target.Platform) && target.Platform.Equals("x86", StringComparison.OrdinalIgnoreCase));
+            }
```
rainersigwald: Can this whole flow be simplified to 

```C#
FindDirectories(assemblyTargets,
  target => string.IsNullOrEmpty(target.Platform) || target.Platform.Equals(ProcessorArchitecture.FromString(targetArchitecture), StringComparison.OrdinalIgnoreCase));
```

?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/Serialization/AssemblyFolder.cs
```diff
@@ -0,0 +1,33 @@
...
+        [DataMember(IsRequired = true, Order = 2)]
+        internal string FrameworkVersion { get; set; }
+
+        [DataMember(IsRequired = true, Order = 3)]
+        internal string Path { get; set; }
+
+        [DataMember(IsRequired = false, Order = 4)]
+        internal string Platform { get; set; }
+
+        public Version AdjustedFrameworkVersion
```
rainersigwald: Adjusted how? Just by type?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBUILDMYDEBUG")))
+            {
+                while (!System.Diagnostics.Debugger.IsAttached)
+                {
+                    System.Threading.Thread.Sleep(500);
+
+                }
+                System.Diagnostics.Debugger.Break();
+            }
```
rainersigwald: If this will continue to be useful, give it a more descriptive name than `MY`. But probably just remove it?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+            // Crack the search path just one time.
+            Match match = s_crackAssemblyFoldersFromConfigSentinel.Match(this.searchPathElement);
+            _wasMatch = false;
+
+            if (match.Success)
+            {
+                _targetRuntimeVersion = match.Groups["TARGETRUNTIMEVERSION"].Value.Trim();
+                _assemblyFolderConfigFile = match.Groups["ASSEMBLYFOLDERCONFIGFILE"].Value.Trim();
+
+                // Disregard if there are any empty values in the {Registry} tag.
```
rainersigwald: Stale comment

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    // Tolerate version keys that don't begin with "v" as these could come from user input
+                    if (!_targetRuntimeVersion.StartsWith("v", StringComparison.OrdinalIgnoreCase))
+                    {
+                        _targetRuntimeVersion = _targetRuntimeVersion.Insert(0, "v");
+                    }
+
+                    _wasMatch = true;
+
+                    bool useCache = Environment.GetEnvironmentVariable("MSBUILDDISABLEASSEMBLYFOLDERSEXCACHE") == null;
```
rainersigwald: Should this continue to be the same flag? I can see a case for either way.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+                    string key = "da22615d-aa83-444b-80b9-b32f3d5db098" + searchPathElement;
```
rainersigwald: Should this get a different GUID?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+                                if (targetProcessorArchitecture != ProcessorArchitecture.MSIL && targetProcessorArchitecture != ProcessorArchitecture.None)
+                                {
+                                    foundPath = candidatePath;
+                                    return true;
+                                }
+                                
+                                // Lets see if the processor architecture matches, note this this method will cache the result when it was first called.
+                                AssemblyNameExtension foundAssembly = getAssemblyName(candidatePath);
+
+                                // If the processor architecture does not match the we should continue to see if there is a better match.
```
rainersigwald: typo: *then* we should

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+            }
+
+            return false;
+        }
+    }
+
+    /// <summary>
+    /// Contains information about entries in the AssemblyFoldersEx registry keys.
+    /// </summary>
+    internal class AssemblyFoldersFromConfigCache
```
rainersigwald: split into its own file?

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfigResolver.cs
```diff
@@ -0,0 +1,297 @@
...
+                    if (!FileUtilities.DirectoryExistsNoThrow(assemblyFolder.DirectoryPath)) return;
+
+                    string[] files = Directory.GetFiles(assemblyFolder.DirectoryPath, "*.*", SearchOption.TopDirectoryOnly);
+
+                    lock (lockobject)
+                    {
+                        foreach (var file in files)
+                        {
+                            _filesInDirectories.Add(file);
+                        }
```
rainersigwald: `_filesInDirectories.UnionWith(files);`

rainersigwald: Actually this whole mess could be dramatically simplified:

```C#
_filesInDirectories = assemblyFoldersEx.AsParallel()
    .Where(assemblyFolder => FileUtilities.DirectoryExistsNoThrow(assemblyFolder.DirectoryPath))
    .SelectMany(
        assemblyFolder =>
            Directory.GetFiles(assemblyFolder.DirectoryPath, "*.*", SearchOption.TopDirectoryOnly))
    .ToImmutableHashSet(StringComparer.OrdinalIgnoreCase);
```

---
In reply to: [73542581](https://github.com/Microsoft/msbuild/pull/863#discussion_r73542581) [](ancestors = 73542581)

rainersigwald:
LGTM with minor edits.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
AndyGerlicher: It is needed to behave the same. When on 64-bit if targeting 64-bit it is supposed to add the 64-bit assemblies first and then the 32 bit.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
AndyGerlicher: I'll add a comment. But no I don't think it can be simplified that much because there are really 3 cases:
64bit and targeting 64 bit
64bit and targeting 32 bit
32bit

Those scenarios add folders in a different order.

src/XMakeTasks/AssemblyDependency/AssemblyFoldersFromConfig/AssemblyFoldersFromConfig.cs
rainersigwald: Ah, you're right.

rainersigwald:
:shipit:

PR #864, opened 2016-08-04
Merge Xaml task changes back to master

rainersigwald:
#856 was needed for Visual Studio "15" Preview 4, but we've taken changes into `master` that we don't want to ship quite yet. I cherry-picked the changes to a newly created stabilization branch. This just merges those back into master so future history spelunking is easier.


rainersigwald:
Note that, like the 🍰, the diff is a lie. The changes are already present in `master`.

PR #866, opened 2016-08-04
Make MS.Build InternalsVisibleTo MS.Build.Conversion.Core

AndyGerlicher:
Previous versions have shipped with InternalsVisibleTo set for the
Conversion.Core assembly. OSS version has never had that set which
causes some issues. For example, crashes trying to use the
ErrorUtilities VerifyThrow functionality which is all internal.


rainersigwald:
🚢

PR #871, opened 2016-08-08
Update CI badges

rainersigwald:
The URLs for CI badges were stale after the badges were not updated to account for new URLs after 3384e573b73b3d162ed612d47e6e8c5f32e9ee5e. Fixing that, and also trying to make it clearer what's master and what's xplat.


AndyGerlicher:
LGTM

PR #873, opened 2016-08-08
Update .nuspec for CoreXT package

tmeschter:
Update the .nuspec file for the CoreXT package. It turns out we don't need the
architecture (x86, x64, etc.) burned into the name, so drop that. However, we
do want the branch name burned into the description. There may be times when
two different branches are producing packages with very similar version numbers,
and we'll want a reasonably easy way to figure out where they came from.


tmeschter:
@AndyGerlicher @rainersigwald This is a small change; could you take a look?

rainersigwald:
LGTM

Issue #874, opened 2016-08-08
Worker process invocation on .NET Core is broken


rainersigwald:
In what is likely a regression from #859, `dotnet MSBuild.exe /m` fails when it tries to spawn a worker node, emitting:

```
The specified deps.json [O:\msbuild\bin\Bootstrap\MSBuild.deps.json] does not exist
```

And hanging forever.

rainersigwald:
Should be fixed by #881.

PR #875, opened 2016-08-08
Disable multiprocess build (work around #874)

rainersigwald:
It's not entirely clear why yet, but when MSBuild tries to spawn another
node, it emits

```
The specified deps.json [O:\msbuild\bin\Bootstrap\MSBuild.deps.json] does
not exist
```

And hangs. Disabling to get clean CI builds while debugging.


jeffkl:
LGTM, can't wait to find out what caused the issue...

PR #881, opened 2016-08-09
Only override deps and runtimeconfig if they exist

rainersigwald:
cd53996d introduced some tricks to ensure that, when we're invoked by
`dotnet build3`, a worker node can get spawned correctly. Unfortunately,
the technique that worked in that situation gave bad information in the
MSBuild selfhost build post-bootstrap, after 51f6d76 switched us to use
dotnet.exe instead of corerun.

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -12,7 +12,6 @@ using System.Globalization;
 using System.IO;
 using System.IO.Pipes;
 using System.Diagnostics;
-using System.Reflection;
 using System.Threading;
 using System.Runtime.InteropServices;
 using System.Security.Principal;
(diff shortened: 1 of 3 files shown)


cdmihai:
:shipit:

rainersigwald:
@piotrpMSFT @eerhardt Does this seem reasonable? In our selfhost build we invoke MSBuild with `dotnet.exe MSBuild.exe` rather than the `exec {options}` command line that gets used from `dotnet` CLI. The reflection seems to have lied, indicating that the current `.deps.json` file was a file that didn't exist on disk. This gets our selfhost build back to passing, but I haven't tested it in CLI yet (as soon as there's a working version with our last drop I'll take this in and make sure it works there too).

cdmihai:

rainersigwald:

src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -516,7 +516,10 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
                 // from `dotnet build3`, using `dotnet.deps.json` and `dotnet.runtimeconfig.json`
                 var runtimeConfigFile = depsFile.Replace(".deps.json", ".runtimeconfig.json");
 
-                dotnetConfigArguments = $"exec --depsfile {depsFile} --runtimeconfig {runtimeConfigFile}";
+                if (File.Exists(depsFile) && File.Exists(runtimeConfigFile))
```
eerhardt: Do want want each of these checks separate?  What if the .deps.json file exists, but the runtimeConfig doesn't?

rainersigwald: @eerhardt Are you thinking something like

```C#
string depsArgument = null;
string runtimeConfigArgument = null;

if (File.Exists(depsFile))
{
    depsArgument = $"--depsfile {depsFile}";
}

if (File.Exists(runtimeConfigFile))
{
    runtimeConfigArgument = $"--runtimeconfig {runtimeConfigFile}";
}

if (!string.IsNullOrEmpty(depsArgument) || !string.IsNullOrEmpty(runtimeConfigArgument))
{
    dotnetConfigArguments = $"exec {depsArgument} {runtimeConfigArgument}";
}
```

?

rainersigwald: I can do that. I don't understand the implications, though.

eerhardt: Just looking at the code, it seemed odd to not add the deps file to the args because the runtimeconfig didn't exist.

rainersigwald:
Discussed this offline with @piotrpMSFT. The initial change was needed to get MSBuild /m working in the `dotnet` CLI environment while it was explicitly passing in its own `dotnet.deps.json`. With dotnet/cli#4024, that's no longer the case and MSBuild now has the "more normal" situation of having its own `MSBuild.deps.json`. That means that no special arguments are required to `dotnet.exe`, so I'm entirely removing the logic that used deep knowledge of the host to derive those arguments.

This leaves open the possibility that someone _other_ than dotnet CLI will want to run MSBuild with a custom deps/runtimeconfig, but that seems like enough of an edge case that for now we'll do nothing about it.

AndyGerlicher:
:shipit:

PR #676, opened 2016-06-05
Use static semaphore instead of named semaphore

jasonwilliams200OK:
Named semaphores are not supported on Unix in .NET Core
Ref: https://github.com/dotnet/coreclr/blob/5e9a125/src/mscorlib/src/System/Threading/Semaphore.cs#L96

We use static Semaphore which is instantiated per process when the first
node is created, and release it only when "shutdown packet" is encountered.

This fixes some of the failing tests on Ubuntu.

PR #883, opened 2016-08-09
Do not require msbuild.exe.config on .NET Core

rainersigwald:
All of the code that could read configuration data out of
msbuild.exe.config is dependent on System.Configuration, which isn't
available in .NET Core. The reading code was all #if'ed out, but we still
required that the file be physically present, or MSBuild would fail on
initialization.

I made `IsValidMSBuildPath` dependent on the same compile-time check as
src/Shared/ToolsetElement.cs.

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -134,10 +134,19 @@ namespace Microsoft.Build.Shared
         /// <returns>True when the path to MSBuild is valid.</returns>
         private static bool IsValidMSBuildPath(string path)
         {
-            return !string.IsNullOrEmpty(path) &&
-                   Path.GetFileName(path).Equals("MSBuild.exe", StringComparison.OrdinalIgnoreCase) &&
-                   File.Exists(path) &&
+            bool msbuildExeExists = !string.IsNullOrEmpty(path) &&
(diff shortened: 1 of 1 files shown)


rainersigwald:
@AndyGerlicher does this look reasonable to you?

I thought we were still reading toolsets on Core but we clearly weren't because all the toolset-reading code is not built. We shouldn't require an auxiliary file that we don't need . . .

AndyGerlicher:
:shipit:

AndyGerlicher:

jasonwilliams200OK:
@rainersigwald, @AndyGerlicher, that OSX build break http://dotnet-ci.cloudapp.net/job/Microsoft_msbuild/job/_OSX_CoreCLR_prtest/86/ was same as #676 (http://dotnet-ci.cloudapp.net/job/Microsoft_msbuild/job/_OSX_CoreCLR_prtest/89/). This probably is intermittent issue, but is there a tracking issue for this or should we have one? It might be the issue with CoreCLR etc. but it probably needs more investigation. :)

piotrpMSFT:
is there a nupkg we can reference?

rainersigwald:
@piotrpMSFT I kicked off a build, will let you know the version when it's done.

rainersigwald:
@piotrpMSFT `0.1.0-preview-00030-160810` should contain this change (and up to 5dc5fd5ffdf125f5300cbe7af30308f57d894b87).

PR #888, opened 2016-08-10
Add StrongName metadata to FilesToSign item

cdmihai:
Something is causing the satellite assemblies to not be signed.
Trying this out as a potential solution


AndyGerlicher:
:shipit:

cdmihai:
CI failure is due to flaky test. Mering in.

Issue #889, opened 2016-08-10
Enhance OM: Add Update attribute for dynamic item tags


cdmihai:
Parent issue: #700 

Tracks the work needed to implement Update in item groups inside targets (dynamic items). Update in itemgroups not inside targets is tracked in #733.

Design so far:

```xml
<Target>
   <ItemGroup>
      <Compile Update="foo.cs">    // or Update="*.designer.cs"
        <MetadataKey>MetadataValue</MetadataKey>
      </Compile>
   </ItemGroup>
</Target>
```

- Behaviour:
  - For all ProjectItems
  - That have Type=Compile and EvaluatedInclude="Foo.cs"
  - AddOrUpdate Metadakey=MetadataValue

PR #870, opened 2016-08-06
Globbing exclusion and Remove

cdmihai:
This is a remake of #849 but manually rebased on master so the diff is easier to read (does not include the changes in #770)

PR #891, opened 2016-08-10
Update on lazy

cdmihai:
Based on PR #870 

This PR adds Update over the lazy evaluator. Update design discussion is here: #733 

Update for dynamic items is TBD and tracked by a separate issue: #889


src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -59,9 +59,24 @@ IList<I> GetReferencedItems(string itemType, ImmutableHashSet<string> globsToIgn
                 }
             }
 
-            public abstract void Apply(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore);
+            public virtual void Apply(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
+            {
+                var items = SelectItems(listBuilder, globsToIgnore).ToList();
+                MutateItems(items);
```
dsplaisted: This seems like it may modify the metadata on items from previous versions of the list, which are supposed to be immutable.  To avoid this, each item might need to be cloned before modifying the metadata.

cdmihai: I wrote a unit test and it seems to work:

```xml
                            <i2 Include='a;b;c'>
                                <m1>m1_contents</m1>
                                <m2>m2_contents</m2>
                            </i2>

                            <i Include='@(i2)'/>

                            <i2 Update='a;b;c'>
                                <m1>m1_updated</m1>
                                <m2>m2_updated</m2>
                            </i2>
```

Do you have another case in mind?

AndyGerlicher:
LGTM but seems to have a lot of commits that are already in?

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
dsplaisted: @cdmihai OK, so we aren't currently caching the items in a `LazyItemList`.  So every time `GetItems` is called, everything is reevaluated from scratch, and a new set of items is created.

We probably want to cache the list of items we return from `LazyItemList.GetItems`.  If we do, then with the current code the metadata update of `i2` would probably inadvertently apply to the items from the previous version of `i2` that were "copied" into `i`.

cdmihai: Hmm, even then, the cache would be per LazyItemList instance. For example, the lists representing `[include]` and `[include;update]` would have 2 different GetItems caches because they are 2 objects. So another item pointing to the first list would not get affected by subsequent operations, since those would be in different lazylist objects.

We do have a problem if we decide to do structural sharing between GetItems results on the same item type.

cdmihai: @dsplaisted I opened #906 where to document future optimizations

PR #894, opened 2016-08-11
Add legacy projects

tmeschter:
Add the projects for a couple of assemblies that are no longer in active development but we still need to ship for the time being.

These changes are mostly about basic cleanup to get the sources added to the repo and integrated with the build. Future changes will get the associated unit tests, if any, up and running, and introduce NuGet packages for these assemblies.


tmeschter:
@AndyGerlicher @rainersigwald @cdmihai Could you review, please?

src/XMakeConversion/XMakeConversion.csproj
```diff
@@ -1,48 +1,31 @@
...
+    <AssemblyName>Microsoft.Build.Conversion.Core</AssemblyName>
+    <RootNamespace>Microsoft.Build.Conversion.Core</RootNamespace>
   </PropertyGroup>
+  <!-- Default configurations to help VS understand the configurations -->
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x64'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x64'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x86'" />
```
rainersigwald: These two project are just DLLs so we should be able to get away with just an AnyCPU configuration--though that'll require some configuration manager gymnastics to get the solution working.

rainersigwald:
LGTM, but I think @cdmihai might be the best person to get the localized build going.

src/MSBuild.sln
```diff
@@ -50,6 +50,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MSBuildTaskHost", "XMakeCom
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Build.LocalizationTasks", "..\build\LocalizationTasks\Microsoft.Build.LocalizationTasks.csproj", "{06347A6C-32A4-4218-A930-0A22699C726C}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrcasEngine", "OrcasEngine\OrcasEngine.csproj", "{D21C4BF9-E131-4ACB-8960-794797F19A39}"
```
AndyGerlicher: Do we want these in the sln? We want them to build with build.cmd, but seems like it will make the sln slower and we don't want to give the impression people should use/change it. I would have left them out, but I'm fine either way.

AndyGerlicher:
Thanks for doing this!

LGTM except for the loc part failing CI. Ideally we could turn it off for now? @cdmihai is there an easy way? I don't want to block the package creation on work for loc.

cdmihai:
You need to add a bail out mechanism for projects that don't want to be localized. Have their csprojes define a property like '$(DoNotLocalizeMe)` and then update the localization import in here: https://github.com/Microsoft/msbuild/blob/master/src/dir.targets#L12

tmeschter:
@cdmihai I've updated the commits to disable localization for these two projects.

cdmihai:
Loc changes look good.

cdmihai:
Also, please run poli-check before merging :)

tmeschter:
@cdmihai Ran Policheck and cleaned up a few things. I'll merge this once the pull request builds complete successfully.

Issue #692, opened 2016-06-15
Create NuGet packages for all MSBuild assemblies


niklaslundberg:
I want to reference MSBuild from my projects to be able to programmatically parse files like .csproj. There seems to be no way of doing this for .NET Core projects.

daveaglick:
I was actually looking for this exact same thing this morning and was a little surprised to find that there isn't already an official package. I'd love to know if one is planned, even if it won't be available for a while.

I did find a Roslyn.Microsoft.Build package that contains the MSBuild assemblies, but I don't know how "official" it is or if it'll get updates:
https://www.nuget.org/packages/Roslyn.Microsoft.Build/

rainersigwald:
We're currently publishing packages to a private feed: https://dotnet.myget.org/gallery/dotnet-buildtools. But they may not be what you want, since they're (currently) only consumed by dotnet-buildtools to build ourselves and CoreFX.

Having a usable set of packages is on our radar, but not solved yet. The main thing is an interesting complication: in addition to our DLLs, you must have the right set of targets files in order to evaluate an MSBuild project. It'd be frustrating to set up MSBuild package references and then just get "invalid project: can't import Microsoft.CSharp.targets"!

Since this same problem is encountered by OmniSharp, NuGet, and dotnet CLI, I expect we'll have something together in the dev15/dotnet tooling RTM timeframe (but I'm not promising anything here :innocent:).

daveaglick:
@rainersigwald Thanks a bunch for the feedback. I totally understand that this is a moving target, it's helpful to know where you're at. Personally, my main interest is in evaluation of MSBuild files and not compilation (I.e., the `Microsoft.Build.Evaluation` namespace in the `Microsoft.Build.dll` assembly). Looking through the packages in the MyGet feed, I don't actually see that assembly in any of them so I'll continue to sit tight until the packaging story irons out.

daveaglick:
> There should be a Microsoft.Build package in there that should do the trick

Thanks - totally missed that when it didn't show up on the first page in the NuGet Package Explorer. I'll check it out.

wjk:
[...]

PR #895, opened 2016-08-11
Fix NuGet package versions

### LINKED ISSUE #528: Versions of MSBuild NuGet packages are incorrect
The authoring of numerous MSBuild packages, including Microsoft.Build.Utilities.Core, includes dependencies to other MSBuild packages without version specifiers, like so:

```
<group targetFramework=".NETPlatform5.0">
    <dependency id="Microsoft.Build.Framework" />
</group>
```

This causes the implicit version reference to be 0.0.0, which is a release version. However, the versions for the packages are all prerelease, which causes NuGet to produce warnings about downgrading since we're referencing a prerelease version instead of a release version, e.g.,

```
 ToolsetPackages (>= 1.0.0) -> Roslyn.Build.Util (>= 0.9.2-portable) -> Microsoft.Build.Utilities.Core (>= 0.1.0-preview-00005) -> Microsoft.Build.Framework
 ToolsetPackages (>= 1.0.0) -> Roslyn.Build.Util (>= 0.9.2-portable) -> Microsoft.Build.Framework (>= 0.1.0-preview-00005)
```

Roslyn solves this by including a `version="[$version$]"` when building our packages, where `$version$` gets replaced during the build by the NuGet package version. This causes all our packages to have explicit references on each other. MSBuild should probably do the same.

jeffkl:
* Update NUSPEC so that the version is coming from build properties
* Update NUSPEC so that dependency version is coming from build properties
* Don't build packages during the first build

Fixes #528, related to #692

### DIFF
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -65,7 +65,7 @@ echo.
 echo ** Rebuilding MSBuild with downloaded binaries
 
 set MSBUILDLOGPATH=%~dp0msbuild_bootstrap_build.log
-call "%~dp0build.cmd" /t:Rebuild /p:Configuration=%BUILD_CONFIGURATION%
+call "%~dp0build.cmd" /t:Rebuild /p:Configuration=%BUILD_CONFIGURATION% /p:"SkipBuildPackages=true"
 
 if %ERRORLEVEL% NEQ 0 (
     echo.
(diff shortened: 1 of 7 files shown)


rainersigwald:
:shipit:

PR #896, opened 2016-08-11
Avoid node reuse during desktop-hosted builds

rainersigwald:
If msbuild is invoked with the (default) `/nodeReuse:true` argument,
worker nodes hang around for a period of time after the build is complete
to see if more work will be incoming. This is normally a nice
startup-time-reducing perf optimization. But in our own build, it means
that many newly-created MSBuild assemblies are opened and uneditable,
failing builds and causing cleans to fail.

Changing the Windows build script to just always pass `/nodeReuse:false`.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -19,7 +19,7 @@ set MSBUILD_ARGS=%MSBUILD_ARGS% /fileloggerparameters:Verbosity=diag;LogFile="%M
 
 :: Check for a runtime host. If not defined, do not use a host
 if not defined RUNTIME_HOST (
-	set BUILD_COMMAND="%MSBUILD_CUSTOM_PATH%" %MSBUILD_ARGS%
+	set BUILD_COMMAND="%MSBUILD_CUSTOM_PATH%" /nodeReuse:false %MSBUILD_ARGS%
 
     :: Check prerequisites for full framework build
  	if not "%VisualStudioVersion%" == "14.0" (


jeffkl:
LGTM

rainersigwald:
D'oh, forgot that we dropped that switch entirely on core, not just made it a no-op.

jeffkl:
Is this because there is no node reuse on core?  Or you can't turn it off on core?

rainersigwald:
@jeffkl You can't turn it on. The way it's implemented on Windows uses a named pipe named after the PID, so you can enumerate all MSBuild processes and ask if they're available for your use. But that's a potential elevation-of-privilege attack if you can talk to an elevated process (or another user) while yourself running in low-priv mode. Windows provides APIs to check that both sides of the pipe are the same user + elevation level, but other OSes don't, so we just cut the feature for Core. See 86f213b and #506.

jeffkl:
Interesting, well then I'm sure you know what to do, conditionally add the argument...

cdmihai:
Can you branch whether to use the switch or not in the caller of `build.cmd`? I think that one knows what the targeting configuration is.

rainersigwald:
Turned out to be easier to branch inside `build.cmd` but after checking for a host.

AndyGerlicher:
:shipit:

PR #901, opened 2016-08-13
Setup correctly MSBuildExtensionsPath on macOS

mfilippov:
Setup correctly MSBuildExtensionsPath on macOS  when read toolsets from xbuild directory.

### DIFF
--- a/src/XMakeBuildEngine/Definition/ToolsetReader.cs
+++ b/src/XMakeBuildEngine/Definition/ToolsetReader.cs
@@ -216,18 +216,49 @@ namespace Microsoft.Build.Evaluation
                                 // Create standard properties. On Mono they are well known
                                 var buildProperties =
                                     CreateStandardProperties(globalProperties, version, xbuildToolsetsDir, binPath);
-
-                                toolsets.Add(
-                                    version,
-                                    new Toolset(
+                                if (NativeMethodsShared.IsOSX)
+                                {
+                                    toolsets.Add(
                                         version,
-                                        binPath,
-                                        buildProperties,
-                                        environmentProperties,
-                                        globalProperties,
-                                        null,
-                                        currentDir,
-                                        string.Empty));
+                                        new Toolset(
+                                            version,
+                                            binPath,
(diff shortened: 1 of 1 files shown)


radical:
This is incorrect. `/Library/Frameworks/Mono.framework/External/xbuild/` is meant as a fallback path and not the default path. And as a fallback path, it should be set in app.config .

We seem to be missing the fallback paths in the app.config. I'll fix that.

mfilippov:
@radical If I use MSBuild engine as library(without MSBuild.exe) I want to use this fallback for evaluate Xamarin apps. Wihtout this fix it is impossible on macOS.

radical:
You will need to set this your app.config then. 

If you are using msbuild from mono then add the following:

```
                        <msbuildExtensionsPathSearchPaths>
                                <searchPaths os="osx">
                                        <property name="MSBuildExtensionsPath" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                                        <property name="MSBuildExtensionsPath32" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                                        <property name="MSBuildExtensionsPath64" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                                </searchPaths>
                        </msbuildExtensionsPathSearchPaths>
```

.. under a toolset. If you are using msbuild from the xplat branch, then use:

```
                        <projectImportSearchPaths>
                                <searchPaths os="osx">
                                        <property name="MSBuildExtensionsPath" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                                        <property name="MSBuildExtensionsPath32" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                                        <property name="MSBuildExtensionsPath64" value="/Library/Frameworks/Mono.framework/External/xbuild/"/>
                                </searchPaths>
                        </projectImportSearchPaths>
```

This will become consistent with the next mono release, but for now this difference exists.

mfilippov:
@radical Ok I'll try it way.

PR #903, opened 2016-08-15
Bump up System.Collections.Immutable

cdmihai:
We had added a dependency to System.Collections.Immutable which got overwritten by the VS build, thus breaking MSBuild. This PR bumps up collections.immutable and csc.


AndyGerlicher:
LGTM

PR #904, opened 2016-08-15
Add NuGet packages for desktop projects

tmeschter:
Add NuGet package definitions for the Microsoft.Build.Engine and
Microsoft.Build.Conversion.Core assemblies.


tmeschter:
@cdmihai @AndyGerlicher @rainersigwald 

This just adds the .nuspec files, which are based on those in the xplat branch. I am by no means an expert in NuGet packages, so there is probably something wrong with what I've got here. :-)

Also, the master branch doesn't have the logic present in the xplat branch for actually creating the NuGet packages as part of the build. Would it make sense for me to simply copy that logic from the xplat branch?

AndyGerlicher:
LGTM for the nuspec. And yeah I think copying the xplat logic would be the way to go. And trying to keep as much identical as possible to make the merges easier.

cdmihai:
@jeffkl about the nuspects. They lgtm, except for their dependencies. As far as I know, you have to put something in [even for standard GACed .net stuff](https://docs.nuget.org/Create/NuSpec-Reference#specifying-framework-assembly-references-(gac)). But the documentation is all gooey and I'm not too sure :)

Also, might it be better to just do this over in xplat land? Rather than copy paste modify hairy build logic that will then have to get merged back into xplat?

src/nuget/Microsoft.Build.Conversion.Core.nuspec
```diff
@@ -0,0 +1,27 @@
...
+    <title>MSBuild APIs</title>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=624683</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <summary>MSBuild APIs</summary>
+    <description>
+      This package includes APIs that support writing MSBuild tasks.
```
rainersigwald: Maybe mention something about deprecation?

rainersigwald:
LGTM, and I agree with Mihai: just do the package production in xplat.

josemesona:
They look correct, the dependency version is "$version$" which will always align the 2 assemblies.

---
In reply to: [240147643](https://github.com/Microsoft/msbuild/pull/904#issuecomment-240147643) [](ancestors = 240147643)

src/nuget/Microsoft.Build.Conversion.Core.nuspec
tmeschter: Yes, I was thinking the same thing after I posted this.

jeffkl:
They look correct, the dependency version is "$version$" which will always align the 2 assemblies.

tmeschter:
The issue with xplat is that we need to package the desktop versions of these assemblies. Or at least I think we do... @AndyGerlicher, am I right about that?

rainersigwald:
@tmeschter Yeah, these assemblies should not be ported to netcore. But we build both flavors out of the xplat branch.

tmeschter:
I've updated the review.

The Microsoft.Build.UnitTests.Evaluation.Expander_Tests.Medley test is failing, but I don't know why. If it is due to my change, it isn't obvious to me what went wrong.

dir.props
```diff
@@ -33,14 +33,15 @@
     <BootstrapDestination>$(BinDir)Bootstrap\</BootstrapDestination>
 
     <!-- Package version / directories -->
-    <BuildToolsVersion>1.0.25-prerelease-00199</BuildToolsVersion>
+    <BuildToolsVersion>1.0.26-prerelease-00627-02</BuildToolsVersion>
```
tmeschter: This is the version used in the xplat branch.

src/nuget/MSBuild.nuspec
```diff
@@ -0,0 +1,31 @@
...
+      <group targetFramework="net46">
+        <dependency id="Microsoft.Build.Framework" version="$version$" />
+        <dependency id="Microsoft.Build" version="$version$" />
+        <dependency id="Microsoft.Build.Utilities.Core" version="$version$" />
+        <dependency id="Microsoft.Build.Tasks.Core" version="$version$" />
+      </group>
+    </dependencies>
+  </metadata>
+  <files>
+    <!--<file src="_._" target="lib\net46" />-->
```
tmeschter: This file exists in the xplat branch, but not in the master branch.

src/nuget/Microsoft.Build.Engine.nuspec
```diff
@@ -0,0 +1,22 @@
...
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=624683</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <summary>MSBuild APIs</summary>
+    <description>
+      This package includes APIs that support writing MSBuild tasks.
+    </description>
+    <copyright>Copyright © Microsoft Corporation</copyright>
+  </metadata>
```
tmeschter: I think this is missing some dependencies.

src/nuget/Microsoft.Build.Targets.nuspec
```diff
@@ -0,0 +1,60 @@
...
+        <dependency id="Microsoft.Build.Tasks.Core" version="$version$" />
+        <dependency id="MSBuild" />
+      </group>
+    </dependencies>
+    <contentFiles>
+      <files include="any/any/**/*.*" buildAction="None" copyToOutput="true" />
+    </contentFiles>
+  </metadata>
+  <files>
+    <!--<file src="_._" target="lib\net46" />-->
```
tmeschter: Another file that exists in xplat but not in master.

src/nuget/Microsoft.Build.Targets.nuspec
```diff
@@ -0,0 +1,60 @@
...
+    <file src="Microsoft.Common.CurrentVersion.targets" target="runtimes\any\native" />
+    <file src="Microsoft.Common.targets" target="runtimes\any\native" />
+    <file src="Microsoft.CSharp.CurrentVersion.targets" target="runtimes\any\native" />
+    <file src="Microsoft.CSharp.targets" target="runtimes\any\native" />
+    <file src="Microsoft.NETFramework.CurrentVersion.targets" target="runtimes\any\native" />
+    <file src="Microsoft.NETFramework.targets" target="runtimes\any\native" />
+    <file src="Microsoft.VisualBasic.CurrentVersion.targets" target="runtimes\any\native" />
+    <file src="Microsoft.VisualBasic.targets" target="runtimes\any\native" />
+    <file src="Microsoft.Common.overridetasks" target="runtimes\any\native" />
+    <!--<file src="Extensions\14.1\Microsoft.Common.props" target="contentFiles\any\any\14.1" />-->
```
tmeschter: Ditto.

src/nuget/Microsoft.Build.Conversion.Core.nuspec
```diff
@@ -0,0 +1,27 @@
...
+    </description>
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <dependencies>
+      <group targetFramework="net46">
+        <dependency id="Microsoft.Build.Engine" version="$version$" />
+      </group>
```
tmeschter: I think I need to add a few more dependencies.

src/dirs.proj
```diff
@@ -19,11 +19,44 @@
...
+         NerdBack.GitVersioning. -->
+    <GetNuGetPackageVersionsDependsOn>SetPackageVersion</GetNuGetPackageVersionsDependsOn>
+  </PropertyGroup>
+
+  <PropertyGroup>
     <!-- Explicitly set the OutDir as it is used by the packaging targets -->
     <OutDir>$(BaseOutputPathWithConfig)</OutDir>
   </PropertyGroup>
 
+  <Target Name="SetPackageVersion" DependsOnTargets="GetNuGetPackageVersion" Returns="$(PackageVersion)">
```
rainersigwald: Do you need the `Returns`? That should be the case only if you call this from another project.

src/nuget/MSBuild.nuspec
rainersigwald: IIRC we had to add an empty file with this name to force NuGet to create an empty folder, which we needed to convince . . . something . . . that this package targeted 4.6. So this might be needed here too.

src/dirs.proj
tmeschter: I... have no idea. :-) My knowledge of MSBuild is not that deep.

rainersigwald: (tl;dr: remove it here)

Well, this is one of our dark corners.

Mostly, you don't need a return value from a target: you're running it for its side effects, either on disk (from running tasks that produce files or change the world's state), or in the project (by changing properties and items in memory). But sometimes you do need a concept of a return value: when you're asking another project "what's your output file?", for example.

MSBuild's behavior is that the return value of a target defaults to its `Outputs`, but that can be overridden by adding a `Returns` attribute. That's useful in two main situations:
1. You have multiple incidental outputs, but one "main" output (maybe you produce debug symbols and want to rerun if they're stale, but your real output is only a DLL).
1. You want to communicate something that's not a file-on-disk output to another project--like a computed property.

I don't think you're in either case here, so there shouldn't be a need to have the `Returns`.

src/dirs.proj
```diff
@@ -19,11 +19,44 @@
...
+
+  <PropertyGroup Condition="Exists('$(ToolsDir)packages.targets') and '$(ImportGetNuGetPackageVersions)' != 'false'">
+    <TraversalBuildDependsOn>
+      $(TraversalBuildDependsOn);
+      GetNuspecProperties;
+      BuildPackages;
+    </TraversalBuildDependsOn>
+    <!-- We use Microsoft.DotNet.Build tools to create our packages, but we want the version comes from 
+         NerdBack.GitVersioning. -->
+    <GetNuGetPackageVersionsDependsOn>SetPackageVersion</GetNuGetPackageVersionsDependsOn>
```
rainersigwald: Actually, looking closer, this seems a bit funky. `SetPackageVersion` below has `DependsOnTargets="GetNuGetPackageVersion"`, but by the usual convention, this would set the opposite `DependsOnTargets` on the `GetNuGetPackageVersions` target.

What is the dependency situation? My understanding from looking at this change is that you need to define a couple of properties after GitVersioning runs and before `BuildPackages` runs--is that correct? If so, you should be able to do that a bit more simply with something like

```XML
<Target Name="SetPackageVersion
        DependsOnTargets="GetNuGetPackageVersion"
        BeforeTargets="BuildPackages">
  <PropertyGroup>
    <PackageVersion>$(NuGetPackageVersion)</PackageVersion>
  </PropertyGroup>
  <ItemGroup>
    <!-- Used by packages.targets:BuildPackages to inject properties into the nuspec-->
    <NuspecProperties Include="version=$(NuGetPackageVersion)" />
  </ItemGroup>
</Target>
```

You'll still have to wire `BuildPackages` into the traversal build somehow (via `TraversalBuildDependsOn` as you've done here is fine).

tmeschter: @rainersigwald So `BeforeTargets="BuildPackages"` is enough to ensure `SetPackageVersion` is run, even if nothing depends on it directly?

There are two things I'm trying to achieve: get the package version from GitVersioning and pass that along to `BuildPackages` (by setting `PackageVersion` before `BuildPackages` runs), and create the `NuspecProperties` items, which also need the version from GitVersioning and also need to be created before `BuildPackages` runs.

I'll try your suggestion.

rainersigwald: `BeforeTargets="BuildPackages"` means "if you're going to run `BuildPackages`, ensure that you've already run this target first.

Both of the things you're trying to do are affecting only in-memory state of the current project, so I think my way will do the trick. Let me know if not :)

tmeschter:
@rainersigwald I've updated the review.

tmeschter:
@cdmihai @jeffkl Could you take another look?

Two open questions:

1. Do I need to add the empty `_._` file?
2. Do the packages without dependencies still need an empty `<group targetFramework="net46" />` element?

jeffkl:
I think the empty folder helps NuGet with resolving package applicability but I'm not 100%.  Can you send me the NuGet packages via email so I can crack them open and take a look?

PR #908, opened 2016-08-17
Update stale references to Roslyn targets

rainersigwald:
These were updated in the runtime package list (so our final output was
ok), but not in these consumers.


AndyGerlicher:
:shipit:

PR #912, opened 2016-08-17
Merge master -> xplat

AndyGerlicher:
Gets up up to date with lazy, AssemblyFoldersFromConfig, and OrcasEngine/ConversionCore projects.


src/Shared/EnvironmentUtilities.cs
```diff
@@ -10,52 +10,12 @@ internal static partial class EnvironmentUtilities
     {
         public static bool Is64BitProcess => Marshal.SizeOf<IntPtr>() == 8;
 
-        //  Copied from .NET Framework source for Environment.Is64BitOperatingSystem (along with Win32Native APIs)
-        public static bool Is64BitOperatingSystem
```
AndyGerlicher: This was interesting because it failed my unit test (returned false). No idea why, but I think they've since added `RuntimeInformation.OSArchitecture` so I don't think there's any reason for this.

AndyGerlicher:

rainersigwald:
Microsoft.Build.Tasks.csproj--can you rearrange this so the comment doesn't come unstuck from what it's describing?

```
 +    <!-- Needed by GenerateResource's ResXResourceReader: UNDONE: When CLR has moved this type to improve layering, remove this reference -->
++	<Reference Include="System.Runtime.Serialization" />
 +    <Reference Include="System.Windows.Forms" />
```

src/MSBuild.sln
```diff
@@ -64,6 +64,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TaskWithDependency", "..\Sa
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Build.LocalizationTasks", "..\build\LocalizationTasks\Microsoft.Build.LocalizationTasks.csproj", "{06347A6C-32A4-4218-A930-0A22699C726C}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrcasEngine", "OrcasEngine\OrcasEngine.csproj", "{D21C4BF9-E131-4ACB-8960-794797F19A39}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "XMakeConversion", "XMakeConversion\XMakeConversion.csproj", "{5274C277-F122-4A44-B7A0-00A1B3F39803}"
+EndProject
```
rainersigwald: Nit: these new project shouldn't have mono and netcore configurations

src/OrcasEngine/Resources/Constants.cs
```diff
@@ -87,10 +87,14 @@ internal static string AssemblyVersion
         {
             get
             {
-                Version version = new Version(ThisAssembly.AssemblyVersion);
+#if STANDALONEBUILD
```
rainersigwald: Is this right? `STANDALONEBUILD` seems wrong since it's just pulling from the current version number.

AndyGerlicher: I copied what was in Build.Engine's [Resources/Constants.cs](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Resources/Constants.cs#L159). The error was in ThisAssembly. ThisAssembly is defined in the DevDiv build so this is wrong for our OSS build. I think GitVersioning also defines ThisAssembly but if that's turned off in netcore we don't want to use it anyway. I'm not positive this is right, but it is what we're doing in Microsoft.Build.dll.

rainersigwald:
:shipit:

PR #922, opened 2016-08-19
Merge master nuget commits -> xplat

AndyGerlicher:
I did this real quick so we can get it on the radar. @tmeschter, thoughts? I basically did take target on the nuspec. Do we want to do something different? Or do you want to do the merge and get these full framework packages producing in xplat now (might be easier now than later)? 

Pretty much every file from each commit had a conflict though so it would be good to get on this one.


tmeschter:
@AndyGerlicher Does a build of xplat already produce both Desktop and .Net Standard binaries?

rainersigwald:
@tmeschter Yes, xplat builds both since #850.

AndyGerlicher:
@dotnet-bot test this please

src/dirs.proj
```diff
@@ -58,7 +58,11 @@
     </TraversalBuildDependsOn>
   </PropertyGroup>
 
-
-
+  <!-- Workaround to remove items from PackagesNuSpecFiles (defined in packages.targets) -->
```
AndyGerlicher: Added this to stop trying to create packages those projects on the NetCore build.

rainersigwald: Maybe also comment that they're added via wildcard there, so there's not a great way to remove them otherwise?

rainersigwald:
:shipit:

brthor:
When the nuget package for this build becomes available, will the item remove syntax be supported outside targets in xplat msbuild?

PR #924, opened 2016-08-20
Fix FileMatcher bug around trailing slashes

### LINKED ISSUE #917: Exclude for glob pattern with a directory doesn't work
I have a folder structure like 

```
Test 
      | Test.csproj
      | obj
             | Debug
                          | TempFile.cs
```

In Test.csproj if I specify `<Compile Include="**\*.cs" Exclude = "obj\**" />`, TempFile.cs gets included as well.

cdmihai:
Fixes #917
There was a trailing slash missmatch between two parts of code that have never been put together until now

### DIFF
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -614,7 +614,7 @@ namespace Microsoft.Build.Shared
             public bool NeedsRecursion { get; }
         }
 
-        struct GetFilesRecursionState
+        struct RecursionState
         {
             /// <summary>
             /// The directory to search in
@@ -643,12 +643,12 @@ namespace Microsoft.Build.Shared
         private static void GetFilesRecursive
         (
             System.Collections.IList listOfFiles,
-            GetFilesRecursionState recursionState,
+            RecursionState recursionState,
             string projectDirectory,
             bool stripProjectDirectory,
             GetFileSystemEntries getFileSystemEntries,
-            IList<GetFilesRecursionState> searchesToExclude,
-            Dictionary<string, List<GetFilesRecursionState>> searchesToExcludeInSubdirs
+            IList<RecursionState> searchesToExclude,
+            Dictionary<string, List<RecursionState>> searchesToExcludeInSubdirs
         )
         {
             ErrorUtilities.VerifyThrow((recursionState.SearchData.Filespec== null) || (recursionState.SearchData.RegexFileMatch == null),
@@ -740,18 +740,18 @@ namespace Microsoft.Build.Shared
                     newRecursionState.BaseDirectory = subdir;
                     newRecursionState.RemainingWildcardDirectory = nextStep.RemainingWildcardDirectory;
 
-                    List<GetFilesRecursionState> newSearchesToExclude = null;
+                    List<RecursionState> newSearchesToExclude = null;
 
                     if (excludeNextSteps != null)
                     {
-                        newSearchesToExclude = new List<GetFilesRecursionState>();
+                        newSearchesToExclude = new List<RecursionState>();
 
                         for (int i = 0; i < excludeNextSteps.Length; i++)
                         {
                             if (excludeNextSteps[i].Subdirs != null &&
                                 excludeNextSteps[i].Subdirs.Any(excludedDir => excludedDir.Equals(subdir, StringComparison.Ordinal)))
                             {
-                                GetFilesRecursionState thisExcludeStep = searchesToExclude[i];
+                                RecursionState thisExcludeStep = searchesToExclude[i];
                                 thisExcludeStep.BaseDirectory = subdir;
                                 thisExcludeStep.RemainingWildcardDirectory = excludeNextSteps[i].RemainingWildcardDirectory;
                                 newSearchesToExclude.Add(thisExcludeStep);
@@ -761,13 +761,14 @@ namespace Microsoft.Build.Shared
 
                     if (searchesToExcludeInSubdirs != null)
                     {
-                        List<GetFilesRecursionState> searchesForSubdir;
+                        List<RecursionState> searchesForSubdir;
+
                         if (searchesToExcludeInSubdirs.TryGetValue(subdir, out searchesForSubdir))
                         {
                             //  We've found the base directory that these exclusions apply to.  So now add them as normal searches
                             if (newSearchesToExclude == null)
                             {
-                                newSearchesToExclude = new List<GetFilesRecursionState>();
+                                newSearchesToExclude = new List<RecursionState>();
                             }
                             newSearchesToExclude.AddRange(searchesForSubdir);
                         }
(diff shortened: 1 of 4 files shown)


src/Shared/FileMatcher.cs
```diff
@@ -1438,7 +1440,7 @@ enum GetSearchDataResult
                 needsRecursion);
 
             result.SearchData = searchData;
-            result.BaseDirectory = fixedDirectoryPart;
+            result.BaseDirectory = fixedDirectoryPart.TrimEnd(Path.AltDirectorySeparatorChar).TrimEnd(Path.DirectorySeparatorChar);
```
cdmihai: This is the actual fix

src/Shared/UnitTests/FileMatcher_Tests.cs
```diff
@@ -915,7 +915,7 @@ private int GetMatchingFiles(string[] candidates, string path, string pattern, H
                         string candidateDirectoryName = "";
                         if (normalizedCandidate.IndexOfAny(FileMatcher.directorySeparatorCharacters) != -1)
                         {
-                            candidateDirectoryName = Path.GetDirectoryName(normalizedCandidate) + @"\";
+                            candidateDirectoryName = Path.GetDirectoryName(normalizedCandidate);
```
cdmihai: FileMatcher_Tests were not picking up the bug because the mock file system was also appending spurious slashes :)

src/Shared/UnitTests/FileMatcher_Tests.cs
```diff
@@ -995,8 +995,14 @@ private void GetMatchingDirectories(string[] candidates, string path, string pat
                             int nextSlash = normalizedCandidate.IndexOfAny(FileMatcher.directorySeparatorCharacters, path.Length + 1);
                             if (nextSlash != -1)
                             {
-                                string match = normalizedCandidate.Substring(0, nextSlash + 1);
+                                string match;
+
+                                match = normalizedCandidate.StartsWith(@"\\") && string.IsNullOrEmpty(path)
+                                    ? normalizedCandidate.Substring(0, nextSlash + 1)
+                                    : normalizedCandidate.Substring(0, nextSlash);
```
rainersigwald: Not sure I understand this. Can you comment?

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
```diff
@@ -442,42 +441,98 @@ public void ExcludeVectorWithIncludeVector()
...
+            ExcludeWithWildCards,
+            "a.*",
+            "*.1",
+            new[] { "a.1", "a.2", "a.1" },
+            new[] { "a.2" })]
+        [InlineData(
+            ExcludeWithWildCards,
+            @"**\*.cs",
+            @"a\**",
+            new[] { "1.cs", @"a\2.cs", @"a\b\3.cs", @"a\b\c\4.cs" },
```
rainersigwald: This with all of its embedded slashes will be a merge nightmare into xplat. I think you'll have to go all the way to the enumerable-driven theory. See [https://github.com/Microsoft/msbuild/blob/63c0ab3ef9ca42d4c6d99cccc7dbbd6e5ed26df3/src/XMakeBuildEngine/UnitTests/ExpressionTreeExpression_Tests.cs](https://github.com/Microsoft/msbuild/blob/63c0ab3ef9ca42d4c6d99cccc7dbbd6e5ed26df3/src/XMakeBuildEngine/UnitTests/ExpressionTreeExpression_Tests.cs) for an example.

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
```diff
@@ -442,42 +441,98 @@ public void ExcludeVectorWithIncludeVector()
...
+            new string[]
+            {
+                @"a\a\foo",
+                @"a\b\foo",
+                @"a\foo",
+                "build.proj"
+            })]
+        public void ExcludeVectorWithWildCards(string projectContents, string includeString, string excludeString, string[] inputFiles, string[] expectedInclude)
+        {
+            //todo run the test twice with slashes and back-slashes :)
```
rainersigwald: Oh--that's actually interesting, since there are both filesystem and MSBuild-include-syntax impacts to this. I guess you can ignore my comment as long as creating files is doable with these strings.

src/Shared/FileMatcher.cs
rainersigwald: Comment, please :)

---
In reply to: [75569935](https://github.com/Microsoft/msbuild/pull/924#discussion_r75569935) [](ancestors = 75569935)

rainersigwald:
>                 fixedDirectoryPart = filespec.Substring(0, indexOfLastDirectorySeparator + 1);

Is this a better place for the fix? Or are there other parts that depend on this having the trailing slash?

---
Refers to: src/Shared/FileMatcher.cs:482 in 0abc39d. [](commit_id = 0abc39d8408ecc34207d9afd3faa3d6df95cc733, deletion_comment = False)

rainersigwald:
I would prefer this be split into multiple commits:
1. Fix the tests to match observed behavior (with a commit description that this was why the bug wasn't caught by tests).
2. Better tests.
3. The actual fix
4. The `GetFilesRecursionState` -> `RecursionState` rename.

AndyGerlicher:
:shipit:

src/Shared/UnitTests/FileMatcher_Tests.cs
cdmihai: After I made the test's mock file system not append slashes (to get it consistent with stuff in `System.IO` which does not do that when enumerating directories) a bunch of UNC tests started crashing.  UNC path need that +1 so the proper `\\` fragment is retained. I'll add a comment about it.

rainersigwald: UNC paths return with trailing slashes?

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
cdmihai: System.IO _should_ be able to work with both. We'll see when porting over to xplat. My plan was to leave that comment there until we want to put time into adding `/` support throughout msbuild.

src/Shared/UnitTests/FileMatcher_Tests.cs
cdmihai: No, but this code is trying to parse paths and the initial +1 enabled:
- correct parsing of paths starting with `\\`
- incorrect handling of paths, causing a `\` to remain there

By doing the if I kept the correct case and removed the incorrect one.

cdmihai:
>                 fixedDirectoryPart = filespec.Substring(0, indexOfLastDirectorySeparator + 1);

I fixed it downstream since I didn't quite understand the impact of removing it here (hard to tell where all of this flows into :) )

---
In reply to: [241442240](https://github.com/Microsoft/msbuild/pull/924#issuecomment-241442240) [](ancestors = 241442240)

---
Refers to: src/Shared/FileMatcher.cs:482 in 0abc39d. [](commit_id = 0abc39d8408ecc34207d9afd3faa3d6df95cc733, deletion_comment = False)

cdmihai:
Almost all the file matcher tests start failing if I fix it upstream so I am going to keep it as is.

PR #925, opened 2016-08-20
Xplat

honza1313:


rainersigwald:
We can't do this until we've completed the work tracked in https://github.com/Microsoft/msbuild/milestone/6. It will be a glorious day.

PR #930, opened 2016-08-23
Fix TargetFrameworkIdentifier for CoreCLR builds

jeffkl:
Microsoft.Portable.Core.props blindly sets <TargetFrameworkIdentifier /> so our assemblies show that as the framework instead of NETStandard1.3

I had to use a Before.Microsoft.Common.targets because the assembly info file contents are generated before our dir.targets is imported.


rainersigwald:
Gross, but well-commented. :shipit:

PR #934, opened 2016-08-23
Common Item spec representation

cdmihai:
I've started doing this so addressing bugs like [this one](https://github.com/Microsoft/msbuild/issues/931) does not require me to change 3 different code clones (or more).

I extracted out the ItemSpec class out of the lazy evaluator code and gave it the capability of matching items against itself.

Todos:
- [x] change item provenance and getallglobs to use it
- [ ] change execution time item operations to use it (**future PR**)
- [x] introduce non-canonic path matching (so `.\program.cs` matches with `program.cs` which should match with `.\.\.\program.cs`)

### DIFF
--- a/src/XMakeBuildEngine/Definition/Project.cs
+++ b/src/XMakeBuildEngine/Definition/Project.cs
@@ -6,8 +6,8 @@
 //-----------------------------------------------------------------------
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
@@ -1097,28 +1097,21 @@ namespace Microsoft.Build.Evaluation
 
         private IEnumerable<GlobResult> GetAllGlobs(ProjectItemElement itemElement)
         {
-            Func<string, IElementLocation, HashSet<string>> expandItemSpecIntoFragments = (s, l) =>
-                {
-                    var expandedItemSpec = _data.Expander.ExpandIntoStringListLeaveEscaped(s, ExpanderOptions.ExpandProperties, l);
+            var includeItemspec = new ItemSpec<ProjectProperty, ProjectItem>(itemElement.Include, _data.Expander, itemElement.IncludeLocation);
+            var excludeItemspec = new ItemSpec<ProjectProperty, ProjectItem>(itemElement.Exclude, _data.Expander, itemElement.ExcludeLocation);
 
-                    // don't care about duplicates
-                    var set = new HashSet<string>(expandedItemSpec);
+            var excludeSet =
+                new Lazy<IEnumerable<string>>(
+                    () =>
+                        excludeItemspec.Fragments.Where(
+                            // take out item references, we can't reason about them
+                            f => !(f is ItemExpressionFragment<ProjectProperty, ProjectItem>))
+                            .Select(f => f.ItemSpecFragment)
+                            .ToImmutableHashSet());
 
-                    // take out item references, we can't reason about them
-                    set.RemoveWhere(IsItemReferenceFragment);
-
-                    return set;
-                };
-
-            var includeFragments = expandItemSpecIntoFragments(itemElement.Include, itemElement.IncludeLocation);
-            var excludeFragments = expandItemSpecIntoFragments(itemElement.Exclude, itemElement.ExcludeLocation);
-
-            foreach (var itemFragment in includeFragments)
+            foreach (var globFragment in includeItemspec.Fragments.Where(f => f is GlobFragment))
             {
-                if (IsGlobFragment(itemFragment))
-                {
-                    yield return new GlobResult(itemElement, itemFragment, excludeFragments);
-                }
+                yield return new GlobResult(itemElement, globFragment.ItemSpecFragment, excludeSet.Value);
             }
         }
 
@@ -1191,6 +1184,11 @@ namespace Microsoft.Build.Evaluation
         /// </param>
         public List<ProvenanceResult> GetItemProvenance(ProjectItem item)
         {
+            if (item == null)
+            {
+                return new List<ProvenanceResult>();
+            }
+
             var itemElementsAbove = GetItemElementsAboveItem(_data.EvaluatedItemElements, item);
 
             return GetItemProvenance(item.EvaluatedInclude, itemElementsAbove);
@@ -1222,8 +1220,6 @@ namespace Microsoft.Build.Evaluation
 
         private ProvenanceResult ComputeProvenanceResult(string itemToMatch, ProjectItemElement itemElement)
         {
-            Func<IElementLocation, Func<string, ExpanderOptions, string>> expandForXmlLocation = (l) => (s, o) => _data.Expander.ExpandIntoStringLeaveEscaped(s, o, l);
-
             Func<string, IElementLocation, Operation, ProvenanceResult> singleItemSpecProvenance = (itemSpec, elementLocation, operation) =>
             {
                 if (elementLocation == null)
@@ -1231,7 +1227,9 @@ namespace Microsoft.Build.Evaluation
                     return null;
                 }
 
-                var result = ComputeProvenanceResult(itemToMatch, itemSpec, expandForXmlLocation(elementLocation));
+                Provenance provenance;
+                var matchOccurrences = ItemMatchesInSpecCompareViaExpander(itemToMatch, itemSpec, elementLocation, _data.Expander, out provenance);
+                var result = matchOccurrences > 0 ? Tuple.Create(provenance, matchOccurrences) : null;
 
                 return result?.Item2 > 0
                     ? new ProvenanceResult(itemElement, operation, result.Item1, result.Item2)
@@ -1265,14 +1263,6 @@ namespace Microsoft.Build.Evaluation
             return provenanceProviders.Select(provider => provider()).FirstOrDefault(provenanceResult => provenanceResult != null);
         }
 
-        private Tuple<Provenance, int> ComputeProvenanceResult(string itemToMatch, string itemSpecToLookIn, Func<string, ExpanderOptions, string> expand)
-        {
-            Provenance provenance;
-            var matchOccurrences = ItemMatchesInSpecCompareViaExpander(itemToMatch, itemSpecToLookIn, expand, out provenance);
-
-            return matchOccurrences > 0 ? Tuple.Create(provenance, matchOccurrences) : null;
-        }
-
         /// <summary>
         /// Since:
         ///     - we have no proper AST and interpreter for itemspecs that we can do analysis on
@@ -1280,7 +1270,7 @@ namespace Microsoft.Build.Evaluation
         /// 
         /// The temporary hack is to use the expander to expand the strings, and if any property or item references were encountered, return Provenance.Inconclusive
         /// </summary>
-        private int ItemMatchesInSpecCompareViaExpander(string itemToMatch, string itemSpec, Func<string, ExpanderOptions, string> expand, out Provenance provenance)
+        private int ItemMatchesInSpecCompareViaExpander(string itemToMatch, string itemSpec, IElementLocation elementLocation, Expander<ProjectProperty, ProjectItem> expander, out Provenance provenance)
         {
             if (string.IsNullOrEmpty(itemSpec))
             {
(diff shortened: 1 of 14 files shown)


src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
```diff
@@ -61,14 +58,14 @@ protected override ICollection<I> SelectItems(ImmutableList<ItemData>.Builder li
                     }
                 }
 
-                foreach (var operation in _operations)
+                foreach (var fragment in _itemSpec.Fragments)
                 {
-                    if (operation.Item1 == ItemOperationType.Expression)
+                    if (fragment is ItemExpressionFragment<P, I>)
```
cdmihai: One funky idea is to also let ItemSpec be responsible for producing string values for each fragment type. This would eliminate this type switching in the include operation. If I find another code clone doing the same thing I'll look into it.

cdmihai: One funky idea is to also let ItemSpec be responsible for producing string values for each fragment type. This would eliminate this type switching in the include operation. If I find another code clone doing the same thing I'll look into doing it.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -232,48 +225,8 @@ protected void DecorateItemsWithMetadata(ICollection<I> items, ImmutableList<Pro
...
-                        throw new InvalidOperationException(operation.Item1.ToString());
-                    }
-                }
-
-                var globbingMatcher = EngineFileUtilities.GetMatchTester(globsToMatch);
-
-                return listBuilder
-                    .Select(itemData => itemData.Item)
-                    .Where(item => itemValuesToMatch.Contains(item.EvaluatedInclude) || globbingMatcher(item.EvaluatedInclude))
+                return _itemSpec
```
cdmihai: This is the actual part which now delegates to ItemSpec

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -271,82 +246,27 @@ public void ProcessItemElement(string rootDirectory, ProjectItemElement itemElem
...
             // Proces Update attribute
-            ProcessItemSpec(operationBuilder, itemElement.Update, itemElement.UpdateLocation);
+            ProcessItemSpec(itemElement.Update, itemElement.UpdateLocation, operationBuilder);
 
             ProcessMetadataElements(itemElement, operationBuilder);
 
             return new UpdateOperation(operationBuilder, this);
         }
 
-        void ProcessItemSpec(OperationBuilder operationBuilder, string itemSpec, ElementLocation itemSpecLocation)
```
cdmihai: fragment construction moved to ItemSpec so other code parts can redo this.

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -170,6 +175,12 @@ internal Expander(IPropertyProvider<P> properties, IItemProvider<I> items, IMeta
             _metadata = metadata;
         }
 
+        public Expander(IPropertyProvider<P> properties, IItemProvider<I> items, IItemFactory<I, I> defaultItemFactory) : this(properties, items)
```
rainersigwald: Why `public` rather than `internal` like the others?

Will we need an overload with the `defaultItemFactory` _and_ `metadata` arguments?

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -412,6 +423,12 @@ internal IList<T> ExpandSingleItemVectorExpressionIntoItems<T>(string expression
                 includeNullEntries, out isTransformExpression, elementLocation);
         }
 
+        internal IList<I> ExpandExpressionCaptureIntoItems(ExpressionShredder.ItemExpressionCapture expressionCapture, bool includeNullEntries, IElementLocation elementLocation)
+        {
+            bool throwway;
```
rainersigwald: typo: `throwaway`

src/XMakeBuildEngine/Evaluation/ItemSpec.cs
```diff
@@ -0,0 +1,217 @@
...
+            return items.Where(i => Fragments.Any(f => f.MatchesItem(i.EvaluatedInclude)));
+        }
+
+        public IEnumerable<string> FilterItems(IEnumerable<string> items)
+        {
+            return items.Where(s => Fragments.Any(f => f.MatchesItem(s)));
+        }
+    }
+
+    internal interface ItemFragment
```
rainersigwald: Usual convention is to name this `IItemFragment`, right?

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
rainersigwald: Sounds like a good plan.

---
In reply to: [75950700](https://github.com/Microsoft/msbuild/pull/934#discussion_r75950700) [](ancestors = 75950700)

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -232,48 +225,8 @@ protected void DecorateItemsWithMetadata(ICollection<I> items, ImmutableList<Pro
             /// </summary>
             protected ICollection<I> SelectItemsMatchingItemSpec(ImmutableList<ItemData>.Builder listBuilder, IElementLocation elementLocation)
```
rainersigwald: Not directly related to this PR, but why does this take a list builder instead of just a list?

rainersigwald:
:shipit:

src/XMakeBuildEngine/Evaluation/Expander.cs
cdmihai: I usually make the type internal and leave the members public.

I am still fleshing out the interaction with the expander. I just found out that I will actually have to mutate the default item factory from each item spec, which is a bit of a terrible thing, so most likely I'll end up undoing the expander changes and have the item spec receive an IValuatorData instead of an expander. Using that, an itemspec could create it's own personal expander that would not risk destroying the state of other users of the same expander object.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
cdmihai: All the operations against one item type are collected into an immutable list. Then, when evaluation is done (or other scenarios) somebody calls [GetItems](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs#L149-L180) on the list, which produces an immutable list of actual items. Each operation's apply thus gets a listbuilder to add / mutate items in. So far the common patterns for all the operations has been: produce some items (creation or selection), potentially do something to them, potentially save them back to the list builder.

cdmihai:
I added to this PR. Everything after @rainersigwald's :shipit: comment is new.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1097,28 +1106,14 @@ private List<GlobResult> GetAllGlobs(List<ProjectItemElement> projectItemElement
...
-                    var expandedItemSpec = _data.Expander.ExpandIntoStringListLeaveEscaped(s, ExpanderOptions.ExpandProperties, l);
-
-                    // don't care about duplicates
-                    var set = new HashSet<string>(expandedItemSpec);
+            var includeItemspec = new ItemSpec<ProjectProperty, ProjectItem>(itemElement.Include, _data.Expander, itemElement.IncludeLocation);
+            var excludeItemspec = new ItemSpec<ProjectProperty, ProjectItem>(itemElement.Exclude, _data.Expander, itemElement.ExcludeLocation);
 
-                    // take out item references, we can't reason about them
-                    set.RemoveWhere(IsItemReferenceFragment);
+            var excludeSet = new Lazy<IEnumerable<string>>(() => excludeItemspec.Fragments.Where(f => !(f is ItemExpressionFragment<ProjectProperty, ProjectItem>)).Select(f => f.ItemSpecFragment).ToImmutableHashSet());
```
rainersigwald: I'm beating my head against this `Where` clause. What are you trying to do?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1191,6 +1186,11 @@ public List<ProvenanceResult> GetItemProvenance(string itemToMatch, string itemT
         /// </param>
         public List<ProvenanceResult> GetItemProvenance(ProjectItem item)
         {
+            if (item == null)
+            {
+                return Enumerable.Empty<ProvenanceResult>().ToList();
```
rainersigwald: Is there a reason to prefer `Enumerable.Empty<>().ToList()` over `new List<>()`?

src/XMakeBuildEngine/Evaluation/ItemSpec.cs
```diff
@@ -0,0 +1,261 @@
...
+
+            return new ItemExpressionFragment<P, I>(capture, expression, this);
+        }
+
+        public IEnumerable<I> FilterItems(IEnumerable<I> items)
+        {
+            return items.Where(i => Fragments.Any(f => f.ItemMatches(i.EvaluatedInclude) > 0));
+        }
+
+        public IEnumerable<ItemFragment> FragmentsMatchingItem(string itemToMatch, out int matches)
```
rainersigwald: The sense of this seems a bit off to me. I'd expect the `int` as the return value and the list in the out param.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1289,98 +1286,59 @@ private int ItemMatchesInSpecCompareViaExpander(string itemToMatch, string itemS
...
-                if (ItemMatchesStringLiteral(itemToMatch, itemFragment))
+                else if (fragment is ItemExpressionFragment<ProjectProperty, ProjectItem>)
                 {
-                    provenance |= Provenance.StringLiteral;
-                    occurrences++;
+                    provenance |= Provenance.Inconclusive;
+                }
+                else
+                {
+                    throw new NotImplementedException();
```
rainersigwald: This is scary. Are there current scenarios where this could get hit, or is it just for possible future expansion?

(side note: this will all be so much clearer with C# 7.0!)

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -1640,71 +1651,134 @@ private static class ItemExpander
...
-                    isTransformExpression = true;
+                    return null;
+                }
+
+                result = new List<T>(itemsFromCapture.Count);
+
+                foreach (var itemTuple in itemsFromCapture)
+                {
+                    var itemSpec = itemTuple.Item1;
+                    var originalItem = itemTuple.Item2;
```
rainersigwald: Destructuring assignment . . . soon!

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -1640,71 +1651,134 @@ private static class ItemExpander
...
+            /// <param name="itemsFromCapture">
+            /// List of items.
+            /// 
+            /// Item1 represents the item string, escaped
+            /// Item2 represents the original item.
+            /// 
+            /// Item1 differs from Item2's string when it is coming from a transform
+            /// 
+            /// </param>
+            internal static bool ExpandExpressionCapture<S>(
```
rainersigwald: I don't think I understand what this does. Is there a longer-but-more-descriptive name?

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.RemoveOperation.cs
```diff
@@ -24,16 +28,7 @@ protected override void SaveItems(ICollection<I> items, ImmutableList<ItemData>.
...
-                foreach (var operation in _operations)
-                {
-                    if (operation.Item1 == ItemOperationType.Glob)
-                    {
-                        ret.Add((string)operation.Item2);
-                    }
-                }
-
-                return ret;
+                return _itemSpec.Fragments.OfType<GlobFragment>().Select(g => g.ItemSpecFragment).ToImmutableHashSet().ToBuilder();
```
rainersigwald: Seems inefficient to seal this into an immutable set, then get a builder out. Can you `return new ImmutableHashSet<string>.Builder(_itemSpec.Fragments.OfType<GlobFragment>().Select(g => g.ItemSpecFragment))`  instead?

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2917,7 +3017,7 @@ public void GetAllGlobsShouldFindIndirectlyReferencedGlobsFromProperties()
 
             var expected = new GlobResultList
             {
-                Tuple.Create("A", "*", new HashSet<string> {"*"}),
+                Tuple.Create("A", "*", new HashSet<string> {"*"}.ToImmutableHashSet()),
```
rainersigwald: Should be able to go from array to immutable hash set directly without the mutable `HashSet` in between, right?

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: Retrieve all the itemspec fragments of the exclude that are not item expressions.

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: Nope. I should have deleted that one. I was exploring how I can get a singleton empty collection without instantiating a new object. Exploration was futile :)

src/XMakeBuildEngine/Evaluation/ItemSpec.cs
cdmihai: I returned the more complex object (enumerable vs int) because:
- it decreases clutter on the caller's side (they can `var` the enumerable but are forced to explicitly declare the int)
- the actual fragments are more interesting than the number of matches

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: If that line hits, there's either a bug in the evaluator, or somebody added a new language feature to the itemspec mini-language and forgot to update this if-then-else thing to record provenance from it.

(which c# 7.0 feature are you referring to?)

src/XMakeBuildEngine/Evaluation/Expander.cs
cdmihai: Given an item expression capture (sort of AST for item expressions), it expands it to strings. If it's a transform, then it also provides the original items the new strings came from.

```
<a Include="1;2">
<b include="@(a->'%(Identity)a')">
```
`ExpandExpressionCapture(ItemExpressionCapture("@(a)"))` will return `[("1a", ProjectItem("1")), ("2a", ProjectItem("2"))]`

Alternative names:
`ExpandItemExpressionCaptureIntoStringsWithOriginatingItems`
`ExpandItemExpressionCaptureIntoTupleList`
`ExpandItemExpressionCaptureIntoStringAndOriginalItemList`

cdmihai:
@dotnet-bot build Windows_NT Build for CoreCLR please

cdmihai:

src/XMakeBuildEngine/Evaluation/ItemSpec.cs
rainersigwald: I'd say the trend goes the other way but I don't feel deeply about it. Next year we'll make it return a tuple anyway :)

src/XMakeBuildEngine/Definition/Project.cs
rainersigwald: Got it. Can you use `ErrorUtilities.ThrowInternalErrorUnreachable()` instead for better self-documentation?

Yeah, I was talking about [switch statements with patterns](https://blogs.msdn.microsoft.com/dotnet/2016/08/24/whats-new-in-csharp-7-0/#user-content-switch-statements-with-patterns) to simplify this whole block.

src/XMakeBuildEngine/Evaluation/Expander.cs
rainersigwald: Thanks, that helps! It might be clearer to name the `Tuple<string, S>` type. But that can wait. Looking around, I understood what you were doing at the point of usage, which is more important.

---
In reply to: [77750310](https://github.com/Microsoft/msbuild/pull/934#discussion_r77750310) [](ancestors = 77750310)

rainersigwald:
LGTM

cdmihai:

PR #935, opened 2016-08-23
Remove incorrect trailing semicolon from WriteCodeFragment output

### LINKED ISSUE #936: xplat WriteCodeFragment adds incorrect trailing semicolons to C# output
The non-CodeDom, xplat implementation of WriteCodeFragment is putting  semicolons after attribute declarations. These are incorrect and cause build failures.

nguerrera:
The .NET Core (without CodeDom) implementation of WriteCodeFragment was putting incorrect semicolons after the attribute declarations. 

It is blocking some work that I'm doing so I'm hoping that this can be merged relatively quickly. 

Separately, I also spotted an eyebrow-raising (but ultimately benign) call to DateTime.Now and removed it so nobody else gets confused. This is a separate commit -- b13a816:

>The non-xplat code path (using CodeDom) was updated to not write out a timestamp in order to get determinism. The resource string no longer takes any arguments so the result of DateTime.Now does not surface anywhere. This is just clean up that will prevent a reader from wondering if there is still a determinism issue.

@rainersigwald @AndyGerlicher 

FYI, @dotnet/project-system

Fix #936


nguerrera:
Two more things to follow up on but hopefully we can do them separately so as to unblock things as quickly as possible.

1. Tests should have caught this, but they just do a Contains for the correct output and an extra semicolon doesn't trip that.
2. I also see that the values coming in are not being escaped.

I propose that I file issues to track those.

eerhardt:
:shipit: 

When do you think the tests would be written, if not now?

rainersigwald:

(We should disable that test like we did its sibling.)

nguerrera:
> When do you think the tests would be written, if not now?

I started to fix it and discovered I can't get a green run on my machine even without my changes. :( I'm imagining some latency to get set up correctly, etc.  and I'd prefer if that latency didn't flow to the follow-up work.

src/XMakeTasks/WriteCodeFragment.cs
```diff
@@ -295,7 +295,7 @@ private string GenerateCodeCoreClr(out string extension)
                     if (AssemblyAttributes == null) return string.Empty; 
 
                     extension = "cs";
-                    code.AppendLine("// " + ResourceUtilities.FormatResourceString("WriteCodeFragment.Comment", DateTime.Now));
+                    code.AppendLine("// " + ResourceUtilities.FormatResourceString("WriteCodeFragment.Comment"));
```
rainersigwald: 👍

rainersigwald:
LGTM, thanks!

rainersigwald:
Oh and yes--please at least file a bug for the tests.

nguerrera:
I've filed #936 for tests and #937 for escaping.

rainersigwald:
@nguerrera Also please let us know what tests are failing on your machine--we should be clone + build.

nguerrera:
I have a test change nearly done, so you can wait for that before merging.

nguerrera:
`RebuildWithLocalMSBuild.cmd` passes locally with 5f0adb1 but I think this is going to expose differences in output with desktop. Trying to run desktop tests again locally.

nguerrera:
I think I've got the tests working on desktop too with a212e52. Fingers crossed that CI agrees.

Also, since testing is now part of the PR, I've morphed #936 to represent the whole issue fixed here and not just the tests.

nguerrera:
> Also please let us know what tests are failing on your machine--we should be clone + build.
#939

nguerrera:
@rainersigwald, @eerhardt Please review again. I've fixed the tests.

nguerrera:

eerhardt:
:shipit: when green.

rainersigwald:

rainersigwald:
Thanks for the test improvements! :shipit:

nguerrera:
Thanks for the reviews. Anything blocking getting this merged?

cdmihai:
:shipit:

rainersigwald:
@nguerrera I built packages at version `0.1.0-preview-00032-160824` with this change.

PR #944, opened 2016-08-24
Import all search paths when importing a wildcard

AndyGerlicher:
Changed the import search path logic to import matches from ALL folders
when a wildcard is specified. For example:
  - <Project Import="$(propsdir)\*.props">
    - (Where $(propsdir) has fallback to c:\1, c:\2)
  - MSBuild will import c:\1\*.props AND c:\2\*.props.

The motivation behind this change is when packages install to common
extension points in VS "15". Some packages may still install to the old
(global) MSBuildExtensionsPath location and others may install to the
app-local path. This change will allow for MSBuild to import the union of
those extension points.

Note: We discussed as a team having MSBuild "de-dupe" imports (i.e. if it
imported foo.props in the original location it would not import a fallback
foo.props). This method becomes very tricky with all the import options
(**\*.props, for example).

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -2197,6 +2197,9 @@ namespace Microsoft.Build.Evaluation
 
             bool atleastOneExactFilePathWasLookedAtAndNotFound = false;
 
+            var allProjects = new List<ProjectRootElement>();
+            bool containsWildcards = FileMatcher.HasWildcards(importElement.Project);
+
             // Try every extension search path, till we get a Hit:
             // 1. 1 or more project files loaded
             // 2. 1 or more project files *found* but ignored (like circular, self imports)
@@ -2221,18 +2224,21 @@ namespace Microsoft.Build.Evaluation
                     continue;
                 }
 
-                var newExpandedImportPath = importElement.Project;
-                newExpandedImportPath = newExpandedImportPath.Replace(extensionPropertyRefAsString, extensionPathExpanded);
-
+                var newExpandedImportPath = importElement.Project.Replace(extensionPropertyRefAsString, extensionPathExpanded);
                 _loggingService.LogComment(_buildEventContext, MessageImportance.Low, "TryingExtensionsPath", newExpandedImportPath, extensionPathExpanded);
 
                 List<ProjectRootElement> projects;
-                var result = ExpandAndLoadImportsFromUnescapedImportExpression(directoryOfImportingFile, importElement, newExpandedImportPath,
-                                        false, out projects);
+                var result = ExpandAndLoadImportsFromUnescapedImportExpression(directoryOfImportingFile, importElement, newExpandedImportPath, false, out projects);
 
                 if (result == LoadImportsResult.ProjectsImported)
                 {
-                    return projects;
+                    // If we don't have a wildcard and we had a match, we're done.
+                    if (!containsWildcards)
+                    {
+                        return projects;
+                    }
+
+                    allProjects.AddRange(projects);
                 }
 
                 if (result == LoadImportsResult.FoundFilesToImportButIgnored)
@@ -2240,7 +2246,14 @@ namespace Microsoft.Build.Evaluation
                     // Circular, Self import cases are usually ignored
                     // Since we have a semi-success here, we stop looking at
                     // other paths
-                    return projects;
+
+                    // If we don't have a wildcard and we had a match, we're done.
+                    if (!containsWildcards)
+                    {
+                        return projects;
+                    }
+
+                    allProjects.AddRange(projects);
                 }
 
                 if (result == LoadImportsResult.TriedToImportButFileNotFound)
@@ -2253,12 +2266,14 @@ namespace Microsoft.Build.Evaluation
             // Found at least one project file for the Import, but no projects were loaded
             // atleastOneExactFilePathWasLookedAtAndNotFound would be false, eg, if the expression
             // was a wildcard and it resolved to zero files!
-            if (atleastOneExactFilePathWasLookedAtAndNotFound && (_loadSettings & ProjectLoadSettings.IgnoreMissingImports) == 0)
+            if (allProjects.Count == 0 &&
+                atleastOneExactFilePathWasLookedAtAndNotFound &&
+                (_loadSettings & ProjectLoadSettings.IgnoreMissingImports) == 0)
             {
                 ThrowForImportedProjectWithSearchPathsNotFound(fallbackSearchPathMatch, importElement);
             }
 
-            return new List<ProjectRootElement>();
+            return allProjects;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


AndyGerlicher:
@radical do you see any problem with this for Mono? We could potentially put it behind a flag in the `<projectImportSearchPaths>` definition or something if this would completely break your scenarios?

rainersigwald:
:shipit:

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2197,6 +2197,10 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             bool atleastOneExactFilePathWasLookedAtAndNotFound = false;
 
+            var allProjects = new List<ProjectRootElement>();
+            bool containsWildcards = EscapingUtilities.ContainsEscapedWildcards(importElement.Project) ||
+                                     FileMatcher.HasWildcards(importElement.Project);
```
rainersigwald: What's the motivation for checking for escaped wildcards? Shouldn't only real wildcards matter?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2221,26 +2224,36 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                    allProjects.AddRange(projects);
                 }
 
                 if (result == LoadImportsResult.FoundFilesToImportButIgnored)
                 {
                     // Circular, Self import cases are usually ignored
                     // Since we have a semi-success here, we stop looking at
                     // other paths
-                    return projects;
+
```
radical: Can we add or expand the existing test for this case too? And for the `  ThrowForImportedProjectWithSearchPathsNotFound(fallbackSearchPathMatch, importElement)` code path below. Like maybe a file from a 3rd search path that tries to import itself?

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2197,6 +2197,9 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             bool atleastOneExactFilePathWasLookedAtAndNotFound = false;
 
+            var allProjects = new List<ProjectRootElement>();
+            bool containsWildcards = FileMatcher.HasWildcards(importElement.Project);
+
             // Try every extension search path, till we get a Hit:
             // 1. 1 or more project files loaded
             // 2. 1 or more project files *found* but ignored (like circular, self imports)
```
radical: And a comment, maybe here, about the new wildcard behaviour.

src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
```diff
@@ -167,6 +168,47 @@ public void ImportFromExtensionsPathWithWildCard()
         }
 
         [Fact]
+        public void ImportFromExtensionsPathWithWildCardAndSelfImport()
```
radical: You need to add a 3rd proj file which gets imported after the `main.proj` attempt. So, maybe from a `extnDir3`. That way we would test the new behavior of the search+import not stopping when a circular import is hit.

src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
```diff
@@ -167,6 +168,47 @@ public void ImportFromExtensionsPathWithWildCard()
...
+        {
+            string mainTargetsFileContent = @"
+                <Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003' >
+                    <Target Name='Main'>
+                        <Message Text='Running Main'/>
+                    </Target>
+
+                    <Import Project='$(MSBuildExtensionsPath)\cirularwildcardtest\*.proj'/>
```
radical: s,cirular,circular,

src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
AndyGerlicher: I'm not completely sure I follow. But I did update it, is this what you mean? One of the projects that is imported has an import in it? This doesn't cover any additional code/scenario from my last one so I'm not completely sure this is what you mean. Let me know if it's not. It does make the coverage a little more obvious though, so if that's all your suggesting then how's this?

radical:
```
                 if (result == LoadImportsResult.FoundFilesToImportButIgnored)
                  {
                      // Circular, Self import cases are usually ignored
                      // Since we have a semi-success here, we stop looking at
                      // other paths
-                      return projects;
```

The old code would try the search path `extensions2`, hit the circular import and the above code path would cause it to *stop* the search and return at that point. But the new code ...

```
 +                    // If we don't have a wildcard and we had a match, we're done.
 +                    if (!containsWildcards)
 +                    {
 +                        return projects;
 +                    }
 +
 +                    allProjects.AddRange(projects);
``` 

... does not stop, in case of a wildcard. For that, the test here has an file available in `extensions3`, which gets imported only for the new code and not with the old one. Using `extnTargetsFileContent` without the Import should be fine for `extn3.proj`. Maybe we can add a comment about this in the test, to make it explicit.

AndyGerlicher: Added a comment to the test:
```C#
// Note: This project will try to be imported again and cause a warning (MSB4210). This test should ensure that the
// code does not stop looking in the fallback locations when this happens (extn3.proj should still be imported).
```

src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
```diff
@@ -143,25 +144,80 @@ public void ImportFromExtensionsPathWithWildCard()
...
+                    </Target>
+                </Project>";
+
+            // Importing a wildcard will union all matching results from all fallback locations.
+            string extnDir1 = GetNewExtensionsPathAndCreateFile("extensions1", Path.Combine("circularwildcardtest", "extn.proj"),
+                string.Format(extnTargetsFileContent, "FromExtn1"));
+            string extnDir2 = GetNewExtensionsPathAndCreateFile("extensions2", Path.Combine("circularwildcardtest", "extn.proj"),
+                string.Format(extnTargetsFileContent, "FromExtn2"));
+            string extnDir3 = GetNewExtensionsPathAndCreateFile("extensions3", Path.Combine("circularwildcardtest", "extn3.proj"),
+                string.Format(extnTargetsFileContent2, "FromExtn3"));
```
radical: nit: using `extnTargetsFileContent`, same as for the other project files should be fine here too.

radical:
LGTM! 👍

rainersigwald:
LGTM

PR #946, opened 2016-08-25
Null check was with main object instead of casting object. 

nikson:
In `AssemblyRemapping.cs`, there is invalid null object checking, there `name` variable should be checked for null instead of  main `obj`.


AndyGerlicher:
Looks good, thank you for the contribution!

cdmihai:
LGTM

PR #948, opened 2016-08-25
Remove unnecessary string formatting

BBosman:
As the format strings don't contain any format items the `Uri` object is never inserted into the string, making the `Uri` instantiation (and therefore the `String.Format` calls) unnecessary.


AndyGerlicher:
Odd. This one seems very intentional with something that doesn't seem to have any obvious use.

LGTM. Thank you for the fixes, they are all appreciated!

cdmihai:
LGTM. Most likely the tests were copy pasted and then modified :(

cdmihai:
Thank you for your contribution!

PR #950, opened 2016-08-25
Drop unnecessary if statements

BBosman:
On line [826](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/BootstrapperUtil/BootstrapperBuilder.cs#L826) the local boolean variable `validate` is checked to be `true`.

As there is no code between there and these `if` statements that can modify `validate`, these `if` statements will always resolve to `true`, making them unnecessary.


AndyGerlicher:
LGTM

cdmihai:
LGTM

Issue #750, opened 2016-06-29
Native image cannot be loaded multiple times


brthor:
This error based on offline discussions:
```
C:\Users\brthor\code\cli\build\package\Microsoft.DotNet.Cli.Layout.targets(64,5): error MSB4062: The "Microsoft.Build.Tasks.Copy" task could not be loaded from the assembly Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a. Native image cannot be loaded multiple times Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [C:\Users\brthor\code\cli\build.proj]
```

Creating a standalone repro of this is proving challenging but I have shared a repro offline with 
@rainersigwald and @cdmihai 

Using this issue for tracking workarounds we are making for this issue.

tmeschter:
@rainersigwald @brthor @cdmihai I've been asked to help out with these MSBuild issues, and I've already picked up #658. This issue looks to be related (at least to the extent that it has to do with assembly loading).

Would it be possible for one of you to share the repro with me?

eerhardt:
To repro:
* clone https://github.com/dotnet/cli
* checkout the `feature/msbuild` branch
* Remove these lines https://github.com/dotnet/cli/blob/feature/msbuild/build.proj#L47-L52
* `build` from the root of the repo

tmeschter:
@eerhardt Thanks!

tmeschter:
@eerhardt I'm not able to reproduce the issue using the provided steps. With or without the `<Message>` element in build.proj, the build succeeds.

brthor:
@tmeschter If the entire `MSBuildWorkaroundTarget` and references are removed do you see the behavior?

eerhardt:
Also comment out the Exec call on line 79:

https://github.com/dotnet/cli/blob/feature/msbuild/build.proj#L79

eerhardt:
The issue seems to occur when the root project doesn't use any built-in tasks (like Message or Exec), and it imports a separate .targets file that uses a built-in task (like Message or Exec).

tmeschter:
@brthor @eerhardt Still can't repro this, even after removing the `MSBuildWorkaroundTarget` and all references to it, and removing the `Exec` call on line 79. Of course, without the `Exec` call the build fails with an unrelated problem...

[...]

PR #951, opened 2016-08-25
Update assembly loader

tmeschter:
Related to issue #750.

We need the ability to load task-containing assemblies from arbitrary
locations on disk determined at run time, along with thei dependencies.
It is assumed that their dependencies will be located next to them in
the file system, or in some well-known location.

Ideally we would handle the dependencies by hooking the `Resolving`
event on the default `AssemblyLoadContext`. However, due to
https://github.com/dotnet/coreclr/issues/5837 we can't currently do
this. Instead, we need to implement our own `AssemblyLoadContext`.

This is what we currently do. However, we run into a problem when a task
assembly depends on, say, Microsoft.Build.dll (which is pretty much
guaranteed). Microsoft.Build will already be loaded into the default
`AssemblyLoadContext`, but since our task assembly was loaded into *our*
context, we are likely to be asked to load it again when the task needs
it. We currently handle this by looking for it in the app's base
directory, which seems like a good bet.

So we may end up loading this assembly twice, in two different
contexts--not great, but everything still works. Issue #750 comes into
play when Microsoft.Build.dll has been CrossGen'd to a native assembly.
Both contexts will find and try to load the native assembly--and this is
currently not allowed.

So to work around these other issues in assembly loading, we need to
maintain our own `AssemblyLoadContext` for the time being. However, we
now ask the default `AssemblyLoadContext` to first try satisfying the
load. If it succeeds, the load will occur in that context. Only if it
can't find the assembly in question do we attempt to find and load the
assembly ourselves (putting it in our context).

### DIFF
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -7,20 +7,27 @@ using System.IO;
 using System.Reflection;
 using System.Runtime.Loader;
 
-
-//using Roslyn.Utilities;
-
-//using static Microsoft.CodeAnalysis.AnalyzerAssemblyLoadUtils;
-
 namespace Microsoft.Build.Shared
 {
-    /// Core CLR compatible wrapper for loading analyzers.
+    /// <summary>
+    /// CoreCLR-compatible wrapper for loading task assemblies.
+    /// </summary>
+    /// <remarks>
+    /// This type defines a new <see cref="AssemblyLoadContext"/>. This isn't really what we want to
+    /// do, as we aren't trying to isolate our assembly loads from the default context. All we want
+    /// to do is add support for loading task assemblies and their dependencies, which we assume are
+    /// located immediately next to them in the file system. Ideally we wouldn't derive from
+    /// <see cref="AssemblyLoadContext"/> at all, but simply hook the <see cref="AssemblyLoadContext.Resolving"/>
+    /// event to handle finding the dependencies. For the moment, however, that approach is blocked
+    /// by https://github.com/dotnet/coreclr/issues/5837.
+    /// </remarks>
     internal sealed class CoreClrAssemblyLoader : AssemblyLoadContext
     {
         private readonly Dictionary<string, Assembly> _pathsToAssemblies = new Dictionary<string, Assembly>(StringComparer.OrdinalIgnoreCase);
         private readonly Dictionary<string, Assembly> _namesToAssemblies = new Dictionary<string, Assembly>();
         private readonly List<string> _dependencyPaths = new List<string>();
         private readonly object _guard = new object();
+
         private static readonly string[] _extensions = new[] { "ni.dll", "ni.exe", "dll", "exe" };
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


tmeschter:
@AndyGerlicher @cdmihai @rainersigwald @jeffkl Could you please take a look?

I haven't yet verified that this actually fixes #750--I'm still figuring out how to generate the proper native images. However, if this doesn't work then we're out of options for working around the bugs in the runtime. :-)

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -97,8 +111,26 @@ protected override Assembly Load(AssemblyName assemblyName)
...
+                    {
+                        return assembly;
+                    }
+                }
+                catch
+                {
+                    // LoadFromAssemblyName indicates failure by throwing exceptions; we need to
+                    // catch them, ignore them, and try to load the assembly ourselves.
+                    // This may mask underlying issues where _defaultContext.LoadFromAssemblyName
+                    // should have succeeded, but there isn't anything we can do about that.
```
rainersigwald: So there are a bunch of different exceptions that might be thrown, all indicating the same situation that this assembly isn't already loaded?

rainersigwald:
:shipit:

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -7,22 +7,36 @@
...
+    /// by https://github.com/dotnet/coreclr/issues/5837.
+    /// </remarks>
     internal sealed class CoreClrAssemblyLoader : AssemblyLoadContext
     {
         private readonly Dictionary<string, Assembly> _pathsToAssemblies = new Dictionary<string, Assembly>(StringComparer.OrdinalIgnoreCase);
         private readonly Dictionary<string, Assembly> _namesToAssemblies = new Dictionary<string, Assembly>();
         private readonly List<string> _dependencyPaths = new List<string>();
         private readonly object _guard = new object();
+
+        private readonly AssemblyLoadContext _defaultContext;
```
jeffkl: >_defaultContext [](start = 45, length = 15)

Was there a reason to introduce this instead of just using the static property AssemblyLoadContext.Default?

src/Shared/CoreCLRAssemblyLoader.cs
tmeschter: If the assembly is already loaded, there are two possibilities: it isn't loaded at all, or it is loaded in another load context but not this one.

The exceptions indicate that the default context couldn't locate an assembly with the given name, or located what *should* have been the assembly but wasn't actually (e.g., a native file with the same name, or a corrupted file) or maybe couldn't read the file due to permission issues.

We could think about letting some of these through. For example, if the default context located the assembly but couldn't read it (BadImageFormatException or an exception representing a permission issue) then it would probably be better to let that exception propagate rather than catch or ignore it. It's not clear where we would draw the line, though, and since this whole approach is a workaround for other runtime issues I'd rather just keep it as simple as possible.

rainersigwald: Works for me. Can you add that to the commit description?

src/Shared/CoreCLRAssemblyLoader.cs
tmeschter: I think that was a holdover from when I was trying to hook the `Resolving` event.

It may also be possible that the default `AssemblyLoadContext` could change, and you could argue that we're better off using the same one every time rather than (possibly) a different one, so we should capture it up front. I'm not sure how likely that scenario is.

jeffkl: I would just remove it then

AndyGerlicher:

AndyGerlicher:
LGTM

src/Shared/CoreCLRAssemblyLoader.cs
tmeschter: @rainersigwald Will do.

src/Shared/CoreCLRAssemblyLoader.cs
tmeschter: @jeffkl Will do.

PR #953, opened 2016-08-25
Enable AL task on Net Core

cdmihai:
Addresses https://github.com/dotnet/sdk/issues/95

### DIFF
--- a/src/XMakeTasks/Microsoft.Build.Tasks.csproj
+++ b/src/XMakeTasks/Microsoft.Build.Tasks.csproj
@@ -90,6 +90,9 @@
       <Link>Constants.cs</Link>
       <ExcludeFromStyleCop>True</ExcludeFromStyleCop>
     </Compile>
+    <Compile Include="Al.cs">
+      <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
+    </Compile>
     <Compile Include="ConvertToAbsolutePath.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
@@ -493,6 +496,7 @@
     <Compile Include="ResolveProjectBase.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
+    <Compile Include="SdkToolsPathUtility.cs" />
     <Compile Include="TaskExtension.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
@@ -514,9 +518,6 @@
     <Compile Include="WriteCodeFragment.cs" />
   </ItemGroup>
   <ItemGroup Condition="$(TargetFrameworkIdentifier) == '.NETFramework'">
-    <Compile Include="Al.cs">
-      <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
-    </Compile>
     <Compile Include="AppDomainIsolatedTaskExtension.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
@@ -634,7 +635,6 @@
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
     <Compile Include="ResolveSDKReference.cs" />
-    <Compile Include="SdkToolsPathUtility.cs" />
     <Compile Include="RequiresFramework35SP1Assembly.cs" Condition="'$(MonoBuild)' != 'true'">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -3264,7 +3264,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Target Name="GenerateSatelliteAssemblies"
           Inputs="$(MSBuildAllProjects);@(_SatelliteAssemblyResourceInputs);$(IntermediateOutputPath)$(TargetName)$(TargetExt)"
           Outputs="$(IntermediateOutputPath)%(Culture)\$(TargetName).resources.dll"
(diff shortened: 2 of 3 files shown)


cdmihai:
@eerhardt I enabled AL on net core. It turned out to not require any porting at all. Does this look good to you?

jeffkl:
Should I enable this task in the v14.0 netstandard1.3 ref assembly in the NuGet package?

cdmihai:
I guess it won't hurt :)

eerhardt:
Does this work on non-Windows machines?  Isn't `al.exe` a native executable that's shipped with the Windows SDK?

We still need resource assemblies generated when building on OSX/Linux.

rainersigwald:
That was my understanding: this wasn't ported because the executable won't work cross-platform. @jaredpar owned `al.exe` the last I heard.

jaredpar:
@rainersigwald is correct.  The `al.exe` tool is actually a C++ Win32 binary and has no hope of running cross platform at the moment.

cdmihai:
So, what's the story of generating satellite assemblies crossplatform? What did the CLI use before msbuild?

eerhardt:
@cdmihai - see https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-resgen/ResgenCommand.cs

jaredpar:
@cdmihai I've lost track of the discussion around cross plat `al.exe`.  CC'ing @weshaggard, @davkean  who may better know where this landed.  I agree though it seems likely that we'regoing to have to deliver this.  Too many tool chains depend on it.

davkean:
We're looking at an approach that doesn't use AL for satellite assemblies.

davkean:
Are we aware of other usage?

jaredpar:
Other usages I've seen in the past: CoreFX, other internal .NET repos and Windows.  

If there is a better way to generate sattelite assemblies I'd definitely be interested. In particular if it can produce deterministic binaries.  That's been a blocker for us in the past.

eerhardt:
The current plan is to bring the CLI code in https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-resgen/ResgenCommand.cs forward to work in MSBuild.

jaredpar:
@eerhardt is that code going to be moved over here or will we reference it?  Curious because we need to be able to support deterministic satellite assembly generation in order to have full deterministic builds in repos that depend on them.  I want to file bugs in the appropriate places to make sure the building of that assembly is deterministic.  Today it is not due to the lack of proper API usage.

eerhardt:
@Sridhar-MS is working on this task.  He has a question to the @Microsoft/msbuild-maintainers team on whether that code should go in this repo, or elsewhere.

jaredpar:
@Sridhar-MS do we have an issue tracking this?

borgdylan:
If I am not mistaken, mono has an al.exe implementation in terms of plain MSIL.

jaredpar:
@borgdylan they do.  About ~6 months ago I took a look at that implementation to see if we could just use it.  There appeared to be some missing features so it's not a 100% replacement.  But I do think it could fill a number of scenarios for us for a small amount of work. 

My main concern at that time was working out the licensing issues.  I never had the time to follow up on that though but feel ilke it was workable.

borgdylan:
@jaredpar Remember that Xamarin is part of MSFT now, so you could work it out with them. The license should be MIT on all of Mono.

jaredpar:
@borgdylan 

>  Remember that Xamarin is part of MSFT now, so you could work it out with them. 

Yep.  That's the case today but wasn't when I was doing my initial investigation.  

> The license should be MIT on all of Mono.

Yep.  At least before we still had to clear some small legal hurdles to include this work.  Maybe less now that we are one company.  

Again want to emphasize this is all very doable.  Just six months ago ended up getting distracted by other things and the underlying motivation for this being included ended up not being important hence I didn't finish up the task.

akoeplinger:
@jaredpar we recently got a very nice PR which added a bunch of missing stuff to our AL.exe implementation: https://github.com/mono/mono/pull/3463

It also switched from RefEmit to IKVM.Reflection (which got ported to netstandard as [Managed.Reflection](https://www.nuget.org/packages/Managed.Reflection/)) so it should be considerably easier to adopt this here now.

rainersigwald:
@jaredpar I think the most important question here is: what is the .NET Core/cross-platform satellite-assembly-generation story?

From the MSBuild perspective, I don't really care: we'll take a PR to use whatever the answer is.

From the broader perspective: it seems like it should be an executable delivered through a NuGet package that runs under the `dotnet` CLI host. Is that true of the Mono `al.exe`? Requiring Mono seems like it might be a nonstarter for the scenarios that @eerhardt is focusing on.

eerhardt:
/cc @Sridhar-MS, since he is actively working on porting the CLI code that generates satellite assemblies to MSBuild.

> Requiring Mono seems like it might be a nonstarter for the scenarios that @eerhardt is focusing on.

Correct, we will not require Mono to be installed on the machine in order to do .NET Core development.  So if the Mono al.exe is what we decide to go with, we will have to port that code to .NET Core.

jeffkl:
@cdmihai and @rainersigwald should we close or merge this?

rainersigwald:
Yeah, this isn't the path forward. But I'm not entirely sure what happened to the scenario. @Sridhar-MS did you get it working in a different way?

eerhardt:
We got this working in the dotnet/sdk by generating the satellite assemblies ourselves on .net core:

https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.targets#L156-L198

https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/GenerateSatelliteAssemblies.cs

kasper3:
@eerhardt, @jaredpar, it is not super clear reading this thread that what is the fate of ResGen and AL in .NET Core in future.

PowerShell folks still hope that oneday they will remove this code once it is available upstream https://github.com/PowerShell/PowerShell/tree/982dbfe/src/ResGen.

Now this thread accounts for the implementation from mono, sdk, cli and powershell.. can these be stitched together and port to DotNetTools / MSBuild?

PR #956, opened 2016-08-26
Remove unnecessary if statements

BBosman:
Similar to #950.

All these `if`-checks have nearly identical checks just before, making these unnecessary as the result won't be different.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Expander.cs
+++ b/src/XMakeBuildEngine/Evaluation/Expander.cs
@@ -3042,10 +3042,7 @@ namespace Microsoft.Build.Evaluation
                 if (Environment.GetEnvironmentVariable("MSBUILDENABLEALLPROPERTYFUNCTIONS") == "1")
                 {
                     // We didn't find the type, so go probing. First in System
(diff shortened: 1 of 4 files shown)


src/OrcasEngine/Engine/Expander.cs
```diff
@@ -1394,10 +1394,7 @@ private static Type GetTypeForStaticMethod(string typeName)
                 if (Environment.GetEnvironmentVariable("MSBUILDENABLEALLPROPERTYFUNCTIONS") == "1")
                 {
                     // We didn't find the type, so go probing. First in System
-                    if (objectType == null)
-                    {
-                        objectType = GetTypeFromAssembly(typeName, "System");
-                    }
+                    objectType = GetTypeFromAssembly(typeName, "System");
```
cdmihai: see https://github.com/Microsoft/msbuild/pull/963/files#r76450977

cdmihai:
Looks good, except for the change in `src/OrcasEngine`

BBosman:

rainersigwald:
LGTM, thanks.

PR #957, opened 2016-08-26
Remove unnecessary terniary checks

BBosman:
Both `configurationReader` and `_directMetadata` have `if`-checks just before the ternary checks, ensuring that they're not `null`, making the ternary expression unnecessary. 

I also dropped the `configurationReaderToUse` variable, as it's a copy of `configurationReader` and the only place it's used is in the next line where we still have access to the same `configurationReader`.

### DIFF
--- a/src/XMakeBuildEngine/Definition/ProjectItem.cs
+++ b/src/XMakeBuildEngine/Definition/ProjectItem.cs
@@ -752,7 +752,7 @@ namespace Microsoft.Build.Evaluation
                 temporary.Add(metadatum);
             }
 
-            _directMetadata = (_directMetadata == null ? null : new PropertyDictionary<ProjectMetadata>(_directMetadata.Count));
+            _directMetadata = new PropertyDictionary<ProjectMetadata>(_directMetadata.Count);
 
             foreach (ProjectMetadata metadatum in temporary)
             {


src/OrcasEngine/Engine/ToolsetReader.cs
```diff
@@ -159,10 +158,9 @@ internal static string ReadAllToolsets(ToolsetCollection toolsets, BuildProperty
 
                 if (configurationReader != null)
                 {
-                    configurationReaderToUse = configurationReader == null ? new ToolsetConfigurationReader() : configurationReader;
                     // Accumulation of properties is okay in the config file because it's deterministically ordered
                     defaultToolsVersionFromConfiguration =
-                        configurationReaderToUse.ReadToolsets(toolsets, globalProperties, initialProperties, true /* accumulate properties */);
+                        configurationReader.ReadToolsets(toolsets, globalProperties, initialProperties, true /* accumulate properties */);
```
cdmihai: see https://github.com/Microsoft/msbuild/pull/963/files#r76450977

cdmihai:
:shipit: for the change in ProjectItem.cs

BBosman:

rainersigwald:
LGTM, thanks.

PR #958, opened 2016-08-26
More unnecessary if drops

BBosman:
Both `proj` and `lazyEvaluator` are local variables which get instantiated earlier in the functions guaranteeing them to be non-`null`, making the checks unnecessary.

In the `proj` case there are a couple of references to `proj.ProjectName` before the check, proving the point. :smile:

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -825,27 +825,23 @@ namespace Microsoft.Build.Evaluation
                 EvaluateItemGroupElement(itemGroupElement, lazyEvaluator);
             }
 
-            if (lazyEvaluator != null)
+            // Tell the lazy evaluator to compute the items and add them to _data
+            IList<LazyItemEvaluator<P, I, M, D>.ItemData> items = lazyEvaluator.GetAllItems();
+            foreach (var itemData in items)
             {
-
-                // Tell the lazy evaluator to compute the items and add them to _data
-                IList<LazyItemEvaluator<P, I, M, D>.ItemData> items = lazyEvaluator.GetAllItems();
-                foreach (var itemData in items)
+                if (itemData.ConditionResult)
                 {
-                    if (itemData.ConditionResult)
-                    {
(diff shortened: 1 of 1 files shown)


src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -825,28 +825,24 @@ private static ProjectTargetInstance ReadNewTargetElement(ProjectTargetElement t
                 EvaluateItemGroupElement(itemGroupElement, lazyEvaluator);
             }
 
-            if (lazyEvaluator != null)
+            // Tell the lazy evaluator to compute the items and add them to _data
```
cdmihai: The lazy evaluator was previously introduced and the null / local instantiation are a sort of a poor man's escape hatch to revert back to the old behaviour if need be. At some point in the future (after the next VS release maybe?) we should remove the escape hatch and delete all the downstream code.

src/OrcasEngine/Shared/SolutionParser.cs
```diff
@@ -594,16 +594,13 @@ private void ParseProject(string firstLine)
             ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile(line != null, "SubCategoryForSolutionParsingErrors",
                 new BuildEventFileInfo(SolutionFile), "SolutionParseProjectEofError", proj.ProjectName);
 
-            if (proj != null)
+            // Add the project to the collection
```
cdmihai: see https://github.com/Microsoft/msbuild/pull/963/files#r76450977

cdmihai:
I'm fine with the change in `src/XMakeBuildEngine/Construction/Solution/SolutionFile.cs`. Can you please extract it out in a separate PR so it can be merged? Or we can wait until the rest of the diffs are discussed.

BBosman:
I've updated the PR to remove the `src/OrcasEngine` change from it and I've moved the `src/XMakeBuildEngine/Construction/Solution/SolutionFile.cs` change to its own PR (#971).

cdmihai:
Closing as this change will become obsolete once we decide to do #975

PR #959, opened 2016-08-26
Drop PropertyGroupsAll else if

BBosman:
At line [104](https://github.com/BBosman/msbuild/blob/master/src/OrcasEngine/Choose/GroupEnumeratorHelper.cs#L104) the `if` statement scopes `this.type` to being `ListType.ItemGroupsAll` or `ListType.PropertyGroupsAll`, therefore it will always be `ListType.PropertyGroupsAll` within these `else` clauses, making the extra check unnecessary.

### DIFF
--- a/src/OrcasEngine/Choose/GroupEnumeratorHelper.cs
+++ b/src/OrcasEngine/Choose/GroupEnumeratorHelper.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Build.BuildEngine
                                     yield return nestedGroup;
                                 }
                             }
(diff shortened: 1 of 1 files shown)


cdmihai:
`src/OrcasEngine` is a deprecated project that got put there for historical reasons. I think it's supposed to be read-only.
@AndyGerlicher

AndyGerlicher:
Agreed. Thanks for submitting but we won't take a change unless absolutely necessary. We should put a `readme.md` file noting that in the src.

PR #960, opened 2016-08-26
Drop another unnecessary if

BBosman:
Another unnecessary `if`.

The `if (isQuotedTransform)` on line 255, combined with the `continue` on line 265 ensures `!isQuotedTransform` will always evaluate to `true`.

The `if (functionCapture != null)` on line 270, combined with  the `continue` on line 278 ensures `functionCapture == null` will always evaluate to `true`.

Therefore this `if`-statement will always evaluate to `true`, making it unnecessary.

The same applies to the second instance.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/ExpressionShredder.cs
+++ b/src/XMakeBuildEngine/Evaluation/ExpressionShredder.cs
@@ -278,11 +278,8 @@ namespace Microsoft.Build.Evaluation
                             continue;
                         }
 
-                        if (!isQuotedTransform && functionCapture == null)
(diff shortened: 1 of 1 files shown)


cdmihai:
I'd keep this in for documentation purposes. It is superfluous, but better documents the intent of that murky code.

I'd be ok with removing the condition if you can replace it with a meaningful comment on what the intention was.

BBosman:
I took a stab at a meaningful comment, but couldn't come up with anything useful.

I'll close this out in favor of this `if`-statement documenting the intent of that murky code

PR #961, opened 2016-08-26
Simplify logic

BBosman:
In `Evaluator.cs` the `if (itemsAndMetadataFound.Metadata.Values.Count > 0)` check is already done at line 1806, making the `if` at 1825 unnecessary.

After dropping it, the logic in the `if` at line 1814 is identical to the `else` at line 1818, making the use of the `if`-statement unnecessary as well.

I didn't really know how to handle the comment, so I left that in place. 

The same applies to `LazyItemEvaluator.LazyItemOperation.cs`.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Evaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/Evaluator.cs
@@ -1811,22 +1811,12 @@ namespace Microsoft.Build.Evaluation
                     include.Add(itemElement.Include);
                     ItemsAndMetadataPair itemsAndMetadataFromInclude = ExpressionShredder.GetReferencedItemNamesAndMetadata(include);
 
-                    if (itemsAndMetadataFromInclude.Items != null && itemsAndMetadataFromInclude.Items.Count > 0)
-                    {
-                        needToProcessItemsIndividually = true;
-                    }
-                    else
-                    {
-                        // If there is bare built-in metadata, we must always run items individually, as that almost
-                        // always differs between items.
-
-                        // UNDONE: When batching is implemented for real, we need to make sure that
-                        // item definition metadata is included in all metadata operations during evaluation
-                        if (itemsAndMetadataFound.Metadata.Values.Count > 0)
-                        {
-                            needToProcessItemsIndividually = true;
-                        }
-                    }
+                    // If there is bare built-in metadata, we must always run items individually, as that almost
+                    // always differs between items.
+                    // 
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -1811,22 +1811,12 @@ private void DecorateItemsWithMetadataFromProjectItemElement(ProjectItemElement
                     include.Add(itemElement.Include);
```
radical: These 3 lines can be removed, since `itemsAndMetadataFromInclude` is not being used anymore.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -1811,22 +1811,12 @@ private void DecorateItemsWithMetadataFromProjectItemElement(ProjectItemElement
...
-                        // always differs between items.
-
-                        // UNDONE: When batching is implemented for real, we need to make sure that
-                        // item definition metadata is included in all metadata operations during evaluation
-                        if (itemsAndMetadataFound.Metadata.Values.Count > 0)
-                        {
-                            needToProcessItemsIndividually = true;
-                        }
-                    }
+                    // If there is bare built-in metadata, we must always run items individually, as that almost
```
radical: IIUC, the earlier comment on lines 1808-1809 and this can be "merged" to be something like:

```
// If there is bare metadata of any kind, we must always run items individually. The two cases are:
//
// 1. If the Include involved an item list, we need to run items individually, as even non-built-in
//     metadata might differ between items
// 2. For bare *built-in* metadata, we need to run items individually, as that almost always
//     differs between items
//
// UNDONE: When batching is implemented for real, we need to make sure that
// item definition metadata is included in all metadata operations during evaluation
```

And the same can be done for the changes in `LazyItemOperation.cs` too.

Somebody familiar with this code will need to ensure that the above is correct though!

rainersigwald: That sounds right to me, but my head definitely hurts a little now, and I'm not completely confident. However, I don't think there's anyone actually familiar with this code around . . .

cdmihai: I'd vote to just leave this code as is, until we have to fix a bug in it. Then, we can refactor the whole thing.

rainersigwald:
Let's go with the "wait and then remove this" approach.

PR #963, opened 2016-08-26
Fix asserts

BBosman:
The surrounding `if`-clauses prevent the checks from actually guarding against anything.

### DIFF
--- a/src/XMakeCommandLine/XMake.cs
+++ b/src/XMakeCommandLine/XMake.cs
@@ -2835,10 +2835,7 @@ namespace Microsoft.Build.CommandLine
             {
                 logger = loggerDescription.CreateLogger();
 
-                if (logger == null)
-                {
-                    InitializationException.VerifyThrow(logger != null, "LoggerNotFoundError", unquotedParameter);
-                }
+                InitializationException.VerifyThrow(logger != null, "LoggerNotFoundError", unquotedParameter);
             }
             catch (IOException e)
             {


src/OrcasEngine/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -1516,10 +1516,7 @@ internal void AddEventFinished(string projectTargetNames, BuildEventContext buil
                     entryPoint.AddEventFinished(null, buildEventContext, eventTimeStamp);
                 }
 
-                if (startedEvent == null)
-                {
-                    Debug.Assert(startedEvent != null, "Cannot have finished counter without started counter. ");
-                }
+                Debug.Assert(startedEvent != null, "Cannot have finished counter without started counter. ");
```
cdmihai: This is in deprecated code. Can you please remove this diff?
Also, `startedEvent` is a field which may be null when this method is called, right?

src/XMakeCommandLine/XMake.cs
```diff
@@ -2835,10 +2835,7 @@ string unquotedParameter
             {
                 logger = loggerDescription.CreateLogger();
 
-                if (logger == null)
-                {
-                    InitializationException.VerifyThrow(logger != null, "LoggerNotFoundError", unquotedParameter);
-                }
+                InitializationException.VerifyThrow(logger != null, "LoggerNotFoundError", unquotedParameter);
```
rainersigwald: But regardless of how `CreateLogger` might change, the null check is duplicated in the `if` and the `VerifyThrow` argument, right?

rainersigwald:
My theory on how this happened is that guards like this can be useful if they avoid a `string.Format` for the message. But that's not the case for these instances.

LGTM

src/OrcasEngine/Logging/ParallelLogger/ParallelConsoleLogger.cs
rainersigwald: I prefer leaving the legacy projects as-is.

src/XMakeCommandLine/XMake.cs
cdmihai: Oh, yeah, true, I should also look what's inside the if, not just at the condition :))

BBosman:

cdmihai:
:shipit:

PR #965, opened 2016-08-26
More if-tweaking

BBosman:
The first `if` is identical to the one at line 250 and `_log.HasLoggedErrors` can't have changed in the meantime, making the check obsolete.

And `pfxSuccess` is set to `false` in line 165. It is only modified in the `if` at line 190 and therefore will always be `true` in the else, making the `!pfxSuccess` check obsolete.


cdmihai:
:shipit:

PR #966, opened 2016-08-26
Drop foundInvalidNode check

BBosman:
Due to the usage of `continue` the `if (foundInvalidNode)` check will only ever be reached when `foundInvalidNode` is `true`, making the check (and the `foundInvalidNode` variable) obsolete.

### DIFF
--- a/src/XMakeTasks/CodeTaskFactory.cs
+++ b/src/XMakeTasks/CodeTaskFactory.cs
@@ -529,7 +529,6 @@ namespace Microsoft.Build.Tasks
         /// </summary>
         private bool ValidateTaskNode()
         {
-            bool foundInvalidNode = false;
             if (_taskNode.HasChildNodes)
             {
                 foreach (XmlNode childNode in _taskNode.ChildNodes)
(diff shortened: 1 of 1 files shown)


cdmihai:
:shipit:

Extra credit: that method as it is, does a simple thing but isn't very readable. Can you extract out an IsValidNode XmlNode -> bool method?

jasonwilliams200OK:
Probably `static bool IsValidNode()` instead? Also inverting the condition would reduce the nesting; early return.

This whole method can be transformed into something like:

```c#
private bool ValidateTaskNode()
{
    if(!_taskNode.HasChildNodes || _taskNode.ChildNodes.Any(HasInvalidNode))
    {
        return true;
    }

    _log.LogErrorWithCodeFromResources("CodeTaskFactory.InvalidElementLocation", childNode.Name, _taskNode.Name);
    return false;
}

private static bool IsInvalidNode(XmlNode node)
{
    switch (childNode.NodeType)
    {
        case XmlNodeType.Comment:
        case XmlNodeType.Whitespace:
        case XmlNodeType.Text:
            // These are legal, and ignored
            continue;
        case XmlNodeType.Element:
            if (childNode.Name.Equals("Code", StringComparison.OrdinalIgnoreCase) || childNode.Name.Equals("Reference", StringComparison.OrdinalIgnoreCase) || childNode.Name.Equals("Using", StringComparison.OrdinalIgnoreCase))
            {
                continue;
            }

            return true;
        default:
            return true;
            break;
    }

    return false;
}
```

AndyGerlicher:
@BBosman Closing since we haven't seen any update since the original request. We are willing to accept with changes such as what ghost suggested. Please do reopen if you'd like to submit again.

PR #967, opened 2016-08-26
Drop superfluous casts

BBosman:
As the return types for those methods/properties are already of the type indicated, the casts are superfluous.


cdmihai:
:shipit:

rainersigwald:
LGTM

PR #968, opened 2016-08-26
Enable desktop tests that were skipped in xplat

rainersigwald:
Some I conditionally disabled on netcore, some just worked.


AndyGerlicher:
LGTM

PR #971, opened 2016-08-27
More if drops

BBosman:
This PR contains just the `src/XMakeBuildEngine/Construction/Solution/SolutionFile.cs` change from #958, as requested by @cdmihai.


rainersigwald:
Can you update the commit message to give more context? Maybe something like

```
Drop null check in SolutionFile.ParseProject

This null check happens after several dereferences, so it is useless.
```

(based on your comments in the other PR)

Change looks good to me.

cdmihai:
LGTM. Thanks!

I'll update the commit message when doing the merge.

PR #976, opened 2016-08-29
Update version number

tmeschter:
Update version.json to specify "15.1-preview5" rather than just "15.1".
This makes it clear that the binaries and NuGet packages are pre-release
verions.

This impacts the naming of our NuGet packages, which will now be called
something like "MSBuild.15.1.208-preview5-g524a1042af.nupkg".

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,1 +1,6 @@
-{ "version": "15.1" }
+{
+  "version": "15.1-preview5",
+  "publicReleaseRefSpec": [
+    "^refs/heads/master$"
+  ]
+}


tmeschter:
@jeffkl @rainersigwald @AndyGerlicher @cdmihai Could you take a look, please?

tmeschter:
The motivation for this change is to have pre-release packages properly marked as such prior to uploading anything to MyGet/NuGet.

jeffkl:
looks good to me

tmeschter:
Would it be preferable for NuGet packages built out of master (and xplat?) to drop the `-g524a1042af` part of the package name? It looks like there are a couple of ways of doing this.

rainersigwald:
I do think it'd be nice to drop the SHA from the package versions, if that's not too difficult.

AndyGerlicher:
LGTM

PR #978, opened 2016-08-30
Port #917 to xplat

cdmihai:
Port #917 to xplat

### DIFF
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -642,6 +642,7 @@ namespace Microsoft.Build.Shared
         /// <param name="stripProjectDirectory"></param>
         /// <param name="getFileSystemEntries">Delegate.</param>
         /// <param name="searchesToExclude">Patterns to exclude from the results</param>
+        /// <param name="searchesToExcludeInSubdirs">exclude patterns that might activate farther down the directory tree. Assumes no trailing slashes</param>
         private static void GetFilesRecursive
         (
             System.Collections.IList listOfFiles,
@@ -1442,7 +1443,11 @@ namespace Microsoft.Build.Shared
                 needsRecursion);
 
             result.SearchData = searchData;
-            result.BaseDirectory = fixedDirectoryPart;
+
+            // The double trim fixes https://github.com/Microsoft/msbuild/issues/917
+            // System.IO does not leave trailing slashes in directory enumerations.
+            // Since this string will be matched against System.IO directory enumerations, both need to agree on trailing slashes
+            result.BaseDirectory = fixedDirectoryPart.TrimEnd(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
             result.RemainingWildcardDirectory = wildcardDirectoryPart;
 
             return GetSearchDataResult.RunSearch;
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -617,7 +617,7 @@ namespace Microsoft.Build.Shared
             public bool NeedsRecursion { get; }
         }
 
-        struct GetFilesRecursionState
+        struct RecursionState
         {
             /// <summary>
             /// The directory to search in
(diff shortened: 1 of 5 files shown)


cdmihai:
Interesting failures on OSX. I would have expected all the item exclude with globs tests to fail due to the slashes and backslashes fiasco, but only a particular select ones fail.

cdmihai:

src/Shared/FileMatcher.cs
```diff
@@ -1441,7 +1443,11 @@ enum GetSearchDataResult
                 needsRecursion);
 
             result.SearchData = searchData;
-            result.BaseDirectory = fixedDirectoryPart;
+
+            // The double trim fixes https://github.com/Microsoft/msbuild/issues/917
+            // System.IO does not leave trailing slashes in directory enumerations.
+            // Since this string will be matched against System.IO directory enumerations, both need to agree on trailing slashes
+            result.BaseDirectory = fixedDirectoryPart.TrimEnd(Path.AltDirectorySeparatorChar).TrimEnd(Path.DirectorySeparatorChar);
```
rainersigwald: You could do this with a single call: `.TrimEnd(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)`

src/Shared/UnitTests/FileMatcher_Tests.cs
```diff
@@ -456,7 +456,7 @@ public void ParentWithoutSlash()
         [Fact]
         public void Unc()
         {
-            // Check UNC functionality
+            //Check UNC functionality
```
rainersigwald: Doesn't our style include the space?

src/Shared/UnitTests/FileMatcher_Tests.cs
```diff
@@ -1040,8 +1040,15 @@ private void GetMatchingDirectories(string[] candidates, string path, string pat
                             int nextSlash = normalizedCandidate.IndexOfAny(FileMatcher.directorySeparatorCharacters, path.Length + 1);
                             if (nextSlash != -1)
                             {
-                                string match = normalizedCandidate.Substring(0, nextSlash + 1);
+                                string match;
+
+                                //UNC paths start with a \\ fragment. Match against \\ when path is empty (i.e., inside the current working directory)
+                                match = normalizedCandidate.StartsWith(@"\\") && string.IsNullOrEmpty(path)
+                                    ? @"\\"
+                                    : normalizedCandidate.Substring(0, nextSlash);
```
rainersigwald: I don't understand why this belongs here. And since UNC paths are windows-only should this be conditioned?

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1193,37 +1206,89 @@ internal static void VerifyAssertLineByLine(string expected, string actual)
...
         {
             Assert.True(Directory.Exists(rootDirectory), $"Directory {rootDirectory} does not exist");
 
             var result = new string[files.Length];
 
             for (var i = 0; i < files.Length; i++)
             {
-                var fullPath = Path.Combine(rootDirectory, files[i]);
+                // On Unix there is the risk of creating one file with '\' in its name instead of directories.
+                // Therefore split the arguments into path fragments and recompose the path.
```
rainersigwald: Are there a lot of these lingering in our code? We could also just assert that it doesn't contain a backslash and then fix the tests . . .

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
```diff
@@ -450,44 +450,100 @@ public void ExcludeVectorWithIncludeVector()
...
+            new[] { "1.cs", @"a\2.cs", @"a\b\3.cs", @"a\b\c\4.cs", @"a\b\c\5.txt" },
+            new[] { "1.cs", @"a\2.cs", @"a\b\c\5.txt", "build.proj" })]
+        [InlineData(
+            ExcludeWithWildCards,
+            @"src\**\proj\**\*.cs",
+            @"src\**\proj\**\none\**\*",
+            new[]
+            {
+                "1.cs",
+                @"src\2.cs",
```
rainersigwald: Ah, or maybe this is the reason you want to keep backslashes as valid. :)

rainersigwald:
:shipit:

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
cdmihai: Yes. Next step is to change that test so it runs the same data with slashes and backslashes (and maybe a random combination of both in the same path?)

src/Shared/UnitTests/FileMatcher_Tests.cs
cdmihai: It was transplanted from master. I'd keep it as is and not make the conditions more complicated.

cdmihai:

PR #980, opened 2016-08-30
Sign product binaries

tmeschter:
I discovered that our fully-signed builds were no longer actually
signing our product binaries. We use MicroBuild for strong-name and
Authenticode signing, and it looks for FilesToSign items in projects.
Searching through our code, I could only find these in a couple of
places related to resources assembly and VSIX production. These
correspond to the binaries that are, in fact, still being signed.

I'm not 100% sure what happened, but I think this is due to moving to a
new version of Microsoft.DotNet.BuildTools in commit db26c2d3. I assume
that it used to generate these items, and no longer does so. The fix
here is to explicitly add a FilesToSign item to each project producing
a product binary.

### DIFF
--- a/src/Framework/Microsoft.Build.Framework.csproj
+++ b/src/Framework/Microsoft.Build.Framework.csproj
@@ -170,5 +170,11 @@
   <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
+  <ItemGroup>
+    <FilesToSign Include="$(OutDir)\$(AssemblyName)$(TargetExt)">
+      <Authenticode>Microsoft</Authenticode>
+      <StrongName>StrongName</StrongName>
+    </FilesToSign>
+  </ItemGroup>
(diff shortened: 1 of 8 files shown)


tmeschter:
@AndyGerlicher @rainersigwald @cdmihai @jeffkl Could you take a look? I want to get the signed build fixed as quickly as possible.

src/Framework/Microsoft.Build.Framework.csproj
```diff
@@ -170,5 +170,11 @@
   <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
+  <ItemGroup>
+    <FilesToSign Include="$(OutDir)\$(AssemblyName).dll">
```
jeffkl: For the file name, can you try `$(AssemblyName)$(TargetExt)` ?  For all of these

jeffkl:
I think we should put something generic in dir.targets to sign everything unless a project opts out.  But this is good for now.

AndyGerlicher:
LGTM

PR #983, opened 2016-08-31
Remove *.nuget.targets and *.nuget.props

jeffkl:
These files are generated by NuGet restore and shouldn't be checked in (they are already in the `.gitignore`)

I also set the property `IncludeNuGetImports` to `false` so these files won't be imported because they just generate a warning anyway:

```
warning : Packages containing MSBuild targets and props files cannot be fully installed in
projects targeting multiple frameworks. The MSBuild targets and props files have been ignored.
```


AndyGerlicher:
LGTM

Issue #979, opened 2016-08-30
Willow VSIX doesn't have Microsoft.Common.props


rainersigwald:
Our [VSIX manifest](https://github.com/Microsoft/msbuild/blob/master/setup/files.swr) doesn't list `Microsoft.Common.props`, and Visual Studio "15" Preview 4 shipped with an old copy (listed in a different VSIX). That means that Preview 4 doesn't have the `Directory.Build.props` behavior, even though it does work for `.targets`.

PR #984, opened 2016-08-31
Add Microsoft.Common.props to setup package.

AndyGerlicher:
Somehow this was missed and placed by another component in VS. Adding it
to the package layout.


jeffkl:
Looks good to me

rainersigwald:
:shipit: Don't forget that when we take the drop into VS we'll need to unlist from the other VSIX.

AndyGerlicher:
Fixes #979

PR #986, opened 2016-08-31
Fix --scope argument in cibuild.cmd

jeffkl:
The variable value wasn't being passed to build.cmd
Also add possible value of "build" and make string comparison case insensitive


rainersigwald:
[belated 👍]

jeffkl:
Sorry the longer I waited the more I realized I probably didn't need to get approval for it...

PR #990, opened 2016-09-02
Restore "xunit ->" message when running tests

rainersigwald:
This got lost on the xplat branch, but it's nice to have.

cc @AndyGerlicher--pretty sure we talked about this bugging us.


AndyGerlicher:
Nice! That had been bugging me.

PR #992, opened 2016-09-02
Update NerdBank.GitVersioning

tmeschter:
Update NerdBank.GitVersioning from 1.4.30 to 1.5.46. We have a problem
where VSTS builds don't properly recognize that they're producing
builds for public release and continue to add the commit ID (for example
"-g12345678") to NuGet package names, which is undesirable. So far we
haven't been able to track down what's going wrong, and I'm hoping a
newer version of GitVersioning will solve the problem.

Even if it doesn't, upgrading to a newer version doesn't hurt.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -37,7 +37,7 @@
     <CompilerToolsVersion>2.0.0-beta3</CompilerToolsVersion>
     <XunitVersion>2.1.0</XunitVersion>
     <MicroBuildVersion>0.2.0</MicroBuildVersion>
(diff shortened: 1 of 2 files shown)


tmeschter:
@rainersigwald @cdmihai @jeffkl @AndyGerlicher Would you be able to take a look at this, please?

jeffkl:
I approve if the tests pass

rainersigwald:
:shipit:

PR #994, opened 2016-09-06
Implement imports for project extensions

### LINKED ISSUE #784: Mechanism for "magic" pickup of nuget restore logic during build
On full framework, desktop installs, the nuget msi dropped its targets into `$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.targets\ImportAfter\Microsoft.NuGet.ImportAfter.targets`. This ensured that all the builds on the machine got Nuget functionality.

Problems with this approach:
- no machine wide program files directory for .net core apps. The most you can do with this approach is drop targets into the applocal extension path (xplat msbuild resolves the extension path relative to msbuild.exe)
- not friendly with other consumers. Where would each project put its special logic coming in from nuget packages?. Would they all write their files inside the CLI SDK, into MSBuild's applocal extension path? This does not scale well since a project can now potentially influence other projects

Solution:
- create a new convention for loading project specific build logic that comes from packages
- common.props imports a special file (something like `<project_name>.generated.targets`), if it is there
- the file sits in the intermediate output directory
- nuget can generate its contents inside of it.
- there should be a switch to disable this convention

jeffkl:
Implementation according to this: https://github.com/Microsoft/msbuild/issues/784#issuecomment-241880891

Fixes #784

- Should this import come before or after the directory.build.props?  I had the import come **after** but this might give users less control because their local imports will happen before the NuGet imports.

FYI @eerhardt

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.props
+++ b/src/XMakeTasks/Microsoft.Common.props
@@ -32,6 +32,33 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Import Project="$(DirectoryBuildPropsPath)" Condition="'$(ImportDirectoryBuildProps)' == 'true' and exists('$(DirectoryBuildPropsPath)')"/>
 
   <!-- 
+        Prepare to import project extensions which usually come from packages.  Package management systems will create a file at:
+          $(MSBuildProjectExtensionsPath)\$(MSBuildProjectFile).<SomethingUnique>.props
+          
+        Each package management system should use a unique moniker to avoid collisions.  It is a wild-card import so the package
+        management system can write out multiple files but the order of the import is alphabetic because MSBuild sorts the list.
+    -->
+  <PropertyGroup>
+    <!--
+            The declaration of $(BaseIntermediateOutputPath) had to be moved up from Microsoft.Common.CurrentVersion.targets
+            in order for the $(MSBuildProjectExtensionsPath) to use it as a default.
+        -->
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">obj\</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath Condition="!HasTrailingSlash('$(BaseIntermediateOutputPath)')">$(BaseIntermediateOutputPath)\</BaseIntermediateOutputPath>
+    <MSBuildProjectExtensionsPath Condition="'$(MSBuildProjectExtensionsPath)' == '' ">$(BaseIntermediateOutputPath)</MSBuildProjectExtensionsPath>
+    <!--
+        Import paths that are relative default to be relative to the importing file.  However, since MSBuildExtensionsPath
+        defaults to BaseIntermediateOutputPath we expect it to be relative to the project directory.  So if the path is relative
+        it needs to be made absolute based on the project directory.
+      -->
+    <MSBuildProjectExtensionsPath Condition="'$([System.IO.Path]::IsPathRooted($(MSBuildProjectExtensionsPath)))' == 'false'">$([System.IO.Path]::Combine('$(MSBuildProjectDirectory)', '$(MSBuildProjectExtensionsPath)'))</MSBuildProjectExtensionsPath>
+    <MSBuildProjectExtensionsPath Condition="!HasTrailingSlash('$(MSBuildProjectExtensionsPath)')">$(MSBuildProjectExtensionsPath)\</MSBuildProjectExtensionsPath>
+    <ImportProjectExtensionProps Condition="'$(ImportProjectExtensionProps)' == ''">true</ImportProjectExtensionProps>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.props" Condition="'$(ImportProjectExtensionProps)' == 'true' and exists('$(MSBuildProjectExtensionsPath)')" />
+
+  <!-- 
         Import wildcard "ImportBefore" props files if we're actually in a 12.0+ project (rather than a project being
         treated as 4.0)
     -->
--- a/src/XMakeTasks/Microsoft.Common.targets
+++ b/src/XMakeTasks/Microsoft.Common.targets
@@ -113,6 +113,19 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(CommonTargetsPath)" />
 
+  <!--
+      Prepare to import project extensions which usually come from packages.  Package management systems will create a file at:
+          $(MSBuildProjectExtensionsPath)\$(MSBuildProjectFile).<SomethingUnique>.targets
+          
+        Each package management system should use a unique moniker to avoid collisions.  It is a wild-card import so the package
+        management system can write out multiple files but the order of the import is alphabetic because MSBuild sorts the list.
+    -->
+  <PropertyGroup>
+    <ImportProjectExtensionTargets Condition="'$(ImportProjectExtensionTargets)' == ''">true</ImportProjectExtensionTargets>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.targets" Condition="'$(ImportProjectExtensionTargets)' == 'true' and exists('$(MSBuildProjectExtensionsPath)')" />
+
   <PropertyGroup>
     <ImportDirectoryBuildTargets Condition="'$(ImportDirectoryBuildTargets)' == ''">true</ImportDirectoryBuildTargets>
   </PropertyGroup>
--- a/src/XMakeBuildEngine/UnitTests/EscapingInProjects_Tests.cs
+++ b/src/XMakeBuildEngine/UnitTests/EscapingInProjects_Tests.cs
@@ -952,6 +952,7 @@ namespace Microsoft.Build.UnitTests.EscapingInProjects_Tests
             // ---------------------
             ObjectModelHelpers.CreateFileInTempProjectDirectory("foo.csproj", @"
                 <Project DefaultTargets=`Build` ToolsVersion=`msbuilddefaulttoolsversion` xmlns=`http://schemas.microsoft.com/developer/msbuild/2003`>
+                    <Import Project=`$(MSBuildBinPath)\Microsoft.Common.props` />
                     <PropertyGroup>
                         <Configuration Condition=` '$(Configuration)' == '' `>Debug</Configuration>
                         <Platform Condition=` '$(Platform)' == '' `>AnyCPU</Platform>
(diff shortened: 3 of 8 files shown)


src/XMakeTasks/Microsoft.Common.props
```diff
@@ -11,43 +11,70 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        defaults to BaseIntermediateOutputPath we expect it to be relative to the project directory.  So if the path is relative
+        it needs to be made absolute based on the project directory.
+      -->
+    <MSBuildProjectExtensionsPath Condition="'$([System.IO.Path]::IsPathRooted($(MSBuildProjectExtensionsPath)))' == 'false'">$([System.IO.Path]::Combine('$(MSBuildProjectDirectory)', '$(MSBuildProjectExtensionsPath)'))</MSBuildProjectExtensionsPath>
+    <MSBuildProjectExtensionsPath Condition="!HasTrailingSlash('$(MSBuildProjectExtensionsPath)')">$(MSBuildProjectExtensionsPath)\</MSBuildProjectExtensionsPath>
+    <ImportProjectExtensionProps Condition="'$(ImportProjectExtensionProps)' == ''">true</ImportProjectExtensionProps>
+  </PropertyGroup>
 
-    <Import Project="$(DirectoryBuildPropsPath)" Condition="'$(ImportDirectoryBuildProps)' == 'true' and exists('$(DirectoryBuildPropsPath)')"/>
+  <Import Project="$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.props" Condition="'$(ImportProjectExtensionProps)' == 'true' and exists('$(MSBuildProjectExtensionsPath)')" />
```
eerhardt: Have we decided on an ordering mechanism here?  Or is it up to each package manager to order inside their generated .props file?  (please don't block this PR on this question)

What is the ordering of a wildcard import?

jeffkl: I thought it was up to the OS but it turns out we sort them...

[https://github.com/Microsoft/msbuild/blob/31122efbd2cb79988ecacc7406a1c16ca34d4b34/src/OrcasEngine/Utilities/EngineFileUtilities.cs#L64](https://github.com/Microsoft/msbuild/blob/31122efbd2cb79988ecacc7406a1c16ca34d4b34/src/OrcasEngine/Utilities/EngineFileUtilities.cs#L64)


---
In reply to: [77725944](https://github.com/Microsoft/msbuild/pull/994#discussion_r77725944) [](ancestors = 77725944)

jeffkl:
Use this link to view the PR with whitespace changes ignored: https://github.com/Microsoft/msbuild/pull/994/files?w=1

jeffkl:
@rainersigwald, @cdmihai, please have a look today.

src/XMakeTasks/Microsoft.Common.props
```diff
@@ -11,43 +11,70 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+  </PropertyGroup>
+
+  <Import Project="$(DirectoryBuildPropsPath)" Condition="'$(ImportDirectoryBuildProps)' == 'true' and exists('$(DirectoryBuildPropsPath)')"/>
+
+  <!-- 
+        Prepare to import project extensions which usually come from packages.  Package management systems will create a file at:
+          $(MSBuildProjectExtensionsPath)\$(MSBuildProjectFile).<SomethingUnique>.props
+          
+        Each package management system should use a unique moniker to avoid collisions.  It is a wild-card import so the package
+        management system can write out multiple files but the order of the import is based on the file system's ordering.
```
rainersigwald: Can you update this since you found the sort?

src/XMakeTasks/Microsoft.Common.props
```diff
@@ -11,43 +11,70 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    -->
+  <PropertyGroup>
+    <!--
+            The declaration of $(BaseIntermediateOutputPath) had to be moved up from Microsoft.Common.CurrentVersion.targets
```
rainersigwald: This says "moved" but I don't see a corresponding delete. Intentional?

src/XMakeTasks/Microsoft.Common.props
```diff
@@ -11,43 +11,70 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+            in order for the $(MSBuildProjectExtensionsPath) to use it as a default.
+        -->
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">obj\</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath Condition="!HasTrailingSlash('$(BaseIntermediateOutputPath)')">$(BaseIntermediateOutputPath)\</BaseIntermediateOutputPath>
+    <MSBuildProjectExtensionsPath Condition="'$(MSBuildProjectExtensionsPath)' == '' ">$(BaseIntermediateOutputPath)</MSBuildProjectExtensionsPath>
+    <!--
+        Import paths that are relative default to be relative to the importing file.  However, since MSBuildExtensionsPath
+        defaults to BaseIntermediateOutputPath we expect it to be relative to the project directory.  So if the path is relative
+        it needs to be made absolute based on the project directory.
```
rainersigwald: :+1:

src/XMakeTasks/Microsoft.Common.props
jeffkl: Yeah I wasn't sure if it was safe to remove it from Microsoft.Common.CurrentVersion.targets.  Is there a way to confirm that it is safe?

---
In reply to: [77846900](https://github.com/Microsoft/msbuild/pull/994#discussion_r77846900) [](ancestors = 77846900)

src/XMakeTasks/Microsoft.Common.props
jeffkl: Absolutely

---
In reply to: [77846165](https://github.com/Microsoft/msbuild/pull/994#discussion_r77846165) [](ancestors = 77846165)

src/XMakeTasks/Microsoft.Common.props
rainersigwald: Remove it and listen for screams?

Seriously: I don't know of a good way to guarantee that. But since common.targets imports common.props if it's not already in the project it _seems_ safe, because the new codepath should always get hit.

jeffkl: Okay I'll keep my ears open :)

---
In reply to: [77850989](https://github.com/Microsoft/msbuild/pull/994#discussion_r77850989) [](ancestors = 77850989)

src/XMakeTasks/Microsoft.Common.targets
```diff
@@ -129,6 +129,19 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(DirectoryBuildTargetsPath)" Condition="'$(ImportDirectoryBuildTargets)' == 'true' and exists('$(DirectoryBuildTargetsPath)')"/>
 
+  <!--
+      Prepare to import project extensions which usually come from packages.  Package management systems will create a file at:
+          $(MSBuildProjectExtensionsPath)\$(MSBuildProjectFile).<SomethingUnique>.targets
+          
+        Each package management system should use a unique moniker to avoid collisions.  It is a wild-card import so the package
+        management system can write out multiple files but the order of the import is based on the file system's ordering.
```
rainersigwald: This one too please :)

jeffkl:
I still need to know if this import should come before or after directory.build.props.  I kind of think that stuff coming externally, like NuGet packages, should be imported first so that stuff locally, ie checked in, can override it.  I also think there would be cases where the external stuff would need to come first, what does everyone think?

rainersigwald:
:shipit: after the other comment change + appropriate history rewriting

rainersigwald:
Re: directory.build.props: that's a great question; I'm not sure it's obvious.

I think I lean toward packages, then directory.build.props--then you can override a package default for a whole directory that all reference the packages.

rainersigwald:
and whitespace change: I'm totally on board, but would [broken record] ideally like it in a separate commit.

jeffkl:
I'll do this in a separate PR and fix the merge conflict here

---
In reply to: [245330376](https://github.com/Microsoft/msbuild/pull/994#issuecomment-245330376) [](ancestors = 245330376)

jeffkl:
@rainersigwald I had to fix the tests instead of our build layout because our build layout needs a lot more work.  So the last question is import before or after directory.build.props.  I count 2 votes for importing project extensions first and directory props second.

rainersigwald:
Works for me.

cdmihai:
Why not do the package import after the directory import? :)
That way, people have control to influence how package importing works for their entire project tree.

jeffkl:
In my opinion, it's better to yield control to the local environment versus something coming from an external source such as NuGet packages.  This order lets packages load, then loads stuff that I have locally.  You should still be able to override stuff in packages right?

I guess the case of precedence comes into play here.  Say a package does something like:
``` xml
<PropertyA Condition=" '$(PropertyA)' == ''">Default value</PropertyA>
```

Then your directory imports would need to just set this property with no condition.  And of course on the other hand, you might also lose some control.

Maybe the import of project extension `.props` should come before directory `.props` but `.targets` should come after?

rainersigwald:
That makes me think of two scenarios where directory.build.props _before_ seems useful:

```XML
<ImportProjectExtensionProps>false</ImportProjectExtensionProps>
```

and
```XML
<MSBuildProjectExtensionsPath>some/custom/magic</MSBuildProjectExtensionsPath>
```

I think that latter one is enough to sway me over to that side.

cdmihai:
You're right. In both cases you loose some control. 

Props:
If project extension props imports come first, then you loose the ability to control `BaseIntermediateOutputPath`, `MSBuildProjectExtensionsPath`, and `ImportProjectExtensionProps` but you gain the ability to control the arbitrary logic they import.

Targets
With project extension imports first, you loose the ability to control `ImportProjectExtensionTargets` but gain the ability to override arbitrary stuff extensions import.

From this point of view, having directory imports last gives you the ability to control potentially more behaviour.

I also like your idea of having directory props import first, but directory targets import last. This way, you gain the ability to control both (`BaseIntermediateOutputPath`, `MSBuildProjectExtensionsPath`, and `ImportProjectExtensionProps`) AND the random project extension props (by putting those overrides in your directory targets import). You only loose the ability to override stuff from project extension targets import. However, those should only be targets, and they _should_ be fully configurable from their props.

jeffkl:
Yeah is it safe to say that **control** means you get to set a property first and declare a target last?  As long as everyone does the `Condition="'$(property)' == ''"` check before setting a property, I think it does.

jeffkl:
So then what's the decision here?

cdmihai:
I'd go for `extension props import | directory props import` and `extension targets import | directory targets import`. The only thing you can't control with this is `BaseIntermediateOutputPath`, `MSBuildProjectExtensionsPath`, and `ImportProjectExtensionProps`.

jeffkl:
I think not getting to override `BaseIntermediateOutputPath` is bad.  I now think the order should be:

1. directory imports
2. project extension imports
3. project extension targets
4. directory targets

jeffkl:

cdmihai:
I'm fine with that.

jeffkl:
Thanks!  I'll merge this once the builds have passed.  Seems like Windows_NT for Desktop is hanging...

PR #997, opened 2016-09-07
Ensure the MSBuild.exe.config is properly binplaced for unit tests

jeffkl:
The app.config in the test deployment directory is the same as the normal one.  The binplacing isn't working as part of the project because everything gets built to Windows_NT first and then copied to Windows_NT_Test second.  The first project containing an msbuild.exe.config binplaces the file but the rest, including the unit tests, don't bother because the file is up to date.


rainersigwald:
LGTM, but heads up to @AndyGerlicher--this might be yet another place we need to be careful to set up fallback search paths :(

PR #998, opened 2016-09-07
Improved test logging

rainersigwald:
This is mostly to help me diagnose a mysterious `ToolLocationHelper` test, but it also includes some infrastructure that should help many future tests (once they get opted into xunit logging).

### DIFF
--- a/src/Shared/UnitTests/MockLogger.cs
+++ b/src/Shared/UnitTests/MockLogger.cs
@@ -302,13 +302,13 @@ namespace Microsoft.Build.UnitTests
         {
             if (eventArgs is BuildWarningEventArgs)
             {
-                BuildWarningEventArgs w = (BuildWarningEventArgs)eventArgs;
+                BuildWarningEventArgs w = (BuildWarningEventArgs) eventArgs;
 
                 // hack: disregard the MTA warning.
                 // need the second condition to pass on ploc builds
                 if (w.Code != "MSB4056" && !w.Message.Contains("MSB4056"))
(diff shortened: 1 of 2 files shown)


src/Shared/UnitTests/MockLogger.cs
```diff
@@ -302,47 +302,53 @@ internal void LoggerEventHandler(object sender, BuildEventArgs eventArgs)
...
             {
-                BuildWarningEventArgs w = (BuildWarningEventArgs)eventArgs;
+                BuildWarningEventArgs w = (BuildWarningEventArgs) eventArgs;
 
                 // hack: disregard the MTA warning.
                 // need the second condition to pass on ploc builds
                 if (w.Code != "MSB4056" && !w.Message.Contains("MSB4056"))
                 {
-                    _fullLog.AppendFormat("{0}({1},{2}): {3} warning {4}: {5}\r\n",
+                    string logMessage = string.Format("{0}({1},{2}): {3} warning {4}: {5}",
```
jeffkl: >logMessage [](start = 27, length = 10)

Would it be better to switch this and the other below to use string interpolation?

jeffkl:
:shipit:

src/Shared/UnitTests/MockLogger.cs
rainersigwald: Sounds like an awesome cleanup after the grand unification. Should just be syntactic sugar so I prefer to leave this for now.

PR #1002, opened 2016-09-07
Change #if RETAIL to !DEBUG

rainersigwald:
When we moved to GitHub, we never set up the projects to define RETAIL for
release builds, as was done in the old repo. This led to a memory-use
regression because we were calling the debug method instead of the
optimized one.

Instead of defining RETAIL, I've switched to !DEBUG which reduces the
number of independent switches but seems to keep the required concept
afloat.

### DIFF
--- a/src/Shared/MSBuildNameIgnoreCaseComparer.cs
+++ b/src/Shared/MSBuildNameIgnoreCaseComparer.cs
@@ -129,7 +129,6 @@ namespace Microsoft.Build.Collections
                 return false;
             }
 
-#if RETAIL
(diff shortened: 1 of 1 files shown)


jeffkl:
LGTM

cdmihai:
:shipit:

I also searched for other usages of the RETAIL flag but found only this one.

rainersigwald:
We talked about this in standup and decided to just drop the condition altogether, and always run this optimized version of the code. I'll update the PR shortly.

rainersigwald:
@AndyGerlicher latest change moves to just-always-be-RETAIL as we discussed.

AndyGerlicher:
:shipit:

PR #1003, opened 2016-09-07
Copy the 64 bit satellite assemblies for 64 bit msbuild in the vsix package

cdmihai:
64 bit msbuild failed because our previous vsix setup copied the x86 satellites into the x64 bin directory.

To test this, I built the vsix package locally and ran msbuild from the command line from both the x86 and the x64 vsix bin directories.

### DIFF
--- a/setup/files.swr
+++ b/setup/files.swr
@@ -159,89 +159,89 @@ folder InstallDir:\MSBuild\15.0\Bin\amd64
   file source=$(X86BinPath)Microsoft.WorkflowBuildExtensions.targets
 
 folder InstallDir:\MSBuild\15.0\Bin\amd64\cs
-  file source=$(X86BinPath)cs\Microsoft.Build.resources.dll vs.file.ngen=yes
-  file source=$(X86BinPath)cs\Microsoft.Build.Tasks.Core.resources.dll vs.file.ngen=yes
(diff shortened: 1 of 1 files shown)


cdmihai:
When this one goes in, I'll bump up the msbuild version in VS as well.

AndyGerlicher:
Can we not force the satellite assemblies to always be AnyCPU?

cdmihai:
@AndyGerlicher 

I don’t see a way to force satellite assemblies to be AnyCPU, since the AL task is using the PlatformTarget property: https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L3291

We'd need to change the `GenerateSatelliteAssembly` target to accept an alternative value for the bitness and fallback to `PlatformTarget` if the new property is not set

tmeschter:
Alternatively, we could use corflags.exe to strip the 32BITREQ flag after the satellite assembly is produced (but before it is signed). However, it's probably simpler to just build both 32-bit and 64-bit versions.

tmeschter:
LGTM

cdmihai:
I merged this in and created #1011 and #1012 to improve our handling of satellites

PR #1004, opened 2016-09-08
Use streamreader instead of string to load xml

### LINKED ISSUE #985: Error building when project has percent '%' char in the path
We work with mercurial and hgflow
When a feature branch is created in hgflow, by default the branch will be named like 'feature/[FEATURENAME]'
When Jenkins wants to build that branch, creates a folder with the branch name. But it escapes the slash, resulting in something like 'feature%2F[FEATURENAME]'
Then when Jenkins launches MSBuild, the build will fail because it unescapes paths, and tries to find everything inside 'feature/[FEATURENAME]'. But that folder does not exist.

Why does MSBuild try to Unescape the path always?

maddin2016:
fixes #985

If you have URL characters in your project path like `C:\Project\Project%20\proj.csproj` `XMLTextReader.Read()` method try to unescape these characters. In result the build is failed because it doesn't find the path. Workaround is to load xml from `StreamReader` instead of `string`.

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -1939,7 +1939,9 @@ namespace Microsoft.Build.Construction
                     string beginProjectLoad = String.Format(CultureInfo.CurrentCulture, "Load Project {0} From File - Start", fullPath);
                     DataCollection.CommentMarkProfile(8806, beginProjectLoad);
 #endif
-                    using (XmlTextReader xtr = new XmlTextReader(fullPath))
+                    var stream = new StreamReader(fullPath);
+
+                    using (XmlTextReader xtr = new XmlTextReader(stream))
                     {
                         // Start the reader so it has an idea of what the encoding is.
                         xtr.DtdProcessing = DtdProcessing.Ignore;
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -1939,11 +1939,13 @@ namespace Microsoft.Build.Construction
                     string beginProjectLoad = String.Format(CultureInfo.CurrentCulture, "Load Project {0} From File - Start", fullPath);
                     DataCollection.CommentMarkProfile(8806, beginProjectLoad);
 #endif
+                    XmlReaderSettings dtdSettings = new XmlReaderSettings();
+                    dtdSettings.DtdProcessing = DtdProcessing.Ignore;
+
                     using (var stream = new StreamReader(fullPath))
-                    using (XmlReader xtr = XmlReader.Create(stream))
+                    using (XmlReader xtr = XmlReader.Create(stream, dtdSettings))
                     {
                         // Start the reader so it has an idea of what the encoding is.
-                        xtr.Settings.DtdProcessing = DtdProcessing.Ignore;
                         xtr.Read();
                         var encoding = xtr.GetAttribute("encoding");
                         _encoding = !string.IsNullOrEmpty(encoding) ? Encoding.GetEncoding(encoding) : Encoding.Default;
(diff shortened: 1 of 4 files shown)


cdmihai:
Hmm, I wonder if the tests fail because the stream is not disposed. Can you try adding the stream in the using statement?

rainersigwald:
So . . . was the old way doing encoding detection based on XML magic, and the new way now isn't? Or is there another explanation for those test failures?

maddin2016:
@rainersigwald, for now we use `XmlTextReader` to import the xml. `XmlTextReader` has the ability to read the encoding from string. But since we now read from a stream it can't read the encoding and use a default one. This is why these tests are failing.

maddin2016:
I have added a commit to change from `XmlTextReader` to `XmlReader`. I've tested with url characters like `%20` and all works fine 
image

cdmihai:
@dotnet-bot test this please

rainersigwald:
Ah, nice detective work @maddin2016.

Also

> Starting with the .NET Framework 2.0, we recommend that you use the System.Xml.XmlReader class instead.

😆 Well, that class was deprecated so recently!

maddin2016:
:thumbsup: :wink:

src/XMakeBuildEngine/UnitTests/Construction/ElementLocation_Tests.cs
```diff
@@ -112,7 +112,7 @@ public void TestLargeElementLocationUsedLargeColumn()
             }
             catch (InvalidProjectFileException ex)
             {
-                Assert.Equal(70012, ex.ColumnNumber);
+                Assert.Equal(1, ex.ColumnNumber);
```
rainersigwald: These results are *wrong*, though, aren't they? In `xplat` we have them disabled for netcore.

maddin2016: Do you mean `ex.ColumnNumber` should be `70012`?

rainersigwald: It sure seems like it--we're deliberately inserting 70k spaces and no newline, so I'd expect that element to start around column 70k.

maddin2016: Ahh...ok, it wasn't clear to me what `column` exactly means.

maddin2016: So it's a bug?

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1935,16 +1935,20 @@ private XmlDocumentWithLocation LoadDocument(string fullPath)
...
-#if MSBUILDENABLEVSPROFILING 
+#if MSBUILDENABLEVSPROFILING
                     string beginProjectLoad = String.Format(CultureInfo.CurrentCulture, "Load Project {0} From File - Start", fullPath);
                     DataCollection.CommentMarkProfile(8806, beginProjectLoad);
 #endif
-                    using (XmlTextReader xtr = new XmlTextReader(fullPath))
+                    XmlReaderSettings dtdSettings = new XmlReaderSettings();
+                    dtdSettings.DtdProcessing = DtdProcessing.Ignore;
+
+                    using (var stream = new StreamReader(fullPath, true))
```
maddin2016: I stumbled upon this [here](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectRootElement_Tests.cs#L670) where encoding is added with BOM and we should be able to read this encoding instead of using a default one.

src/XMakeBuildEngine/UnitTests/Construction/ElementLocation_Tests.cs
rainersigwald: Yeah, it seems like a bug to me. But I don't know where, exactly.

rainersigwald:
@dotnet-bot test this please

(the tests should fail after #1189 made the location stuff work again)

jeffkl:
Actually, I think the tests were disabled in 430a0515214a9d61fb447ecd53a8ce944866c4c5.  @maddin2016 can you undo your changes to the `ElementLocation_Tests.cs` file?  The tests should be fixed now via #1189.

rainersigwald:
@maddin2016 also revert decd35176d01f022a85cf8a57fc4d8b681116186 please--the current state should fail since the column-number problem should now be fixed.

maddin2016:
🙌 🎉

jeffkl:
Thanks a lot for this contribution!

PR #1009, opened 2016-09-09
Update NuGet packages to include satellite assemblies

tmeschter:
Update our NuGet packages to include the satellite assemblies for all
supported languages. This only covers assemblies that we are currently
included in NuGet packacges for which we also produce satellite
assemblies. Notably, we ship Microsoft.Build.Engine.dll and
Microsoft.Build.Conversion.Core.dll in NuGet packages but we do not yet
have localization setup for those.

### DIFF
--- a/src/nuget/MSBuild.nuspec
+++ b/src/nuget/MSBuild.nuspec
@@ -27,5 +27,21 @@
   <files>
     <!--<file src="_._" target="lib\net46" />-->
     <file src="MSBuild.exe" target="runtimes\any\native" />
+  
+    <!-- Satellite assemblies -->
+    <file src="cs\MSBuild.resources.dll" target="runtimes\any\native\cs" />
+    <file src="de\MSBuild.resources.dll" target="runtimes\any\native\de" />
+    <file src="en\MSBuild.resources.dll" target="runtimes\any\native\en" />
+    <file src="es\MSBuild.resources.dll" target="runtimes\any\native\es" />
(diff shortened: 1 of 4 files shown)


tmeschter:
@AndyGerlicher @rainersigwald @cdmihai @jeffkl Please review.

In particular, please note if any satellite assemblies are missing. I noticed that we do not seem to produce satellite assemblies for Microsoft.Build.Framework.dll; is that correct?

rainersigwald:
:shipit:

@cdmihai gets final say but I don't see a resx for Framework so it looks right.

Do files in this path get magically binplaced to consuming project outputs? If not this might not be necessary (bloats the compile-time package with useless bits and doesn't help at runtime).

tmeschter:
@rainersigwald Yeah, I'm not going to merge this until we have a better idea of how we expect our packages to be consumed at runtime. I don't know how binplacing will work.

jeffkl:
Sorry for the delay. #Approved

cdmihai:
Framework does not have any resources of its own (no resx), so it does not require satellites. Conversion and Orcas do have .resx files, so, theoretically, if users want to see non-English text coming out of them they need to be localized :)

I will contact the loc team and ask if we can get an exception for them

AndyGerlicher:
Closing this for now as we now work without satellite assemblies being present. We can re-open this if we have a requirement to ship localized satellite assemblies in the future.

Issue #942, opened 2016-08-24
Nupkg is missing dependency on .System.Runtime.Serialization.Xml v4.1.0


nguerrera:
See dotnet/cli#4083

@jeffkl

Issue #682, opened 2016-06-09
Latest Preview Package does not actually target NETStandard 1.3


tannergooding:
The latest preview package `Microsoft.Build v0.1.0-preview-00023-160527` (dotnet-buildtools MyGet feed) claims to target NETStandard 1.3.

However, referencing the assembly from a net46 project fails, because of the dependency on `System.Runtime.Loader` which is strictly a NETStandard1.5 (net462) library.

cdmihai:
I tried to replicate this by creating a portable task with the following project.json:

```
{
  "dependencies": {
    "Microsoft.Build.Framework": "0.1.0-preview-00023-160527",
    "Microsoft.Build.Utilities.Core": "0.1.0-preview-00023-160527"
  },
  "frameworks": {
    "net46": {},
    "netstandard1.3": {
      "dependencies": {
        "System.Net.NetworkInformation": "4.1.0-rc3-24109-00",
        "Microsoft.NETCore": "5.0.1-rc3-24109-00",
        "Microsoft.NETCore.Portable.Compatibility": "1.0.1-rc3-24109-00",
        "Microsoft.NETCore.Runtime": "1.0.2-rc3-24109-00"
      }
    }
  }
}
```

Restore finished successfully producing this object for net46 in the lock file:

```
".NETFramework,Version=v4.6": {
      "Microsoft.Build.Framework/0.1.0-preview-00023-160527": {
        "type": "package",
        "dependencies": {
          "System.Collections": "4.0.11-rc3-24109-00",
          "System.Diagnostics.Debug": "4.0.11-rc3-24109-00",
          "System.Globalization": "4.0.11-rc3-24109-00",
          "System.Runtime": "4.1.0-rc3-24109-00",
          "System.Runtime.InteropServices": "4.1.0-rc3-24109-00",
          "System.Threading": "4.0.11-rc3-24109-00",
          "System.Threading.Thread": "4.0.0-rc3-24109-00"
        },
        "compile": {
          "lib/netstandard1.3/Microsoft.Build.Framework.dll": {}
        },
        "runtime": {
          "lib/netstandard1.3/Microsoft.Build.Framework.dll": {}
        }
      },
      "Microsoft.Build.Utilities.Core/0.1.0-preview-00023-160527": {
        "type": "package",
        "dependencies": {
          "Microsoft.Build.Framework": "(, )",
          "Microsoft.Win32.Primitives": "4.0.1-rc3-24109-00",
[...]

Issue #556, opened 2016-04-07
Create MSBuild Nuget packages targetting multiple platforms


cdmihai:
The [current](https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.Build) MSBuild nuget packages only target .net core. We should come up with a packaging strategy that targets .net core, net46, and mono.

The .net core binaries are built on xplat. net46 binaries are built on master. Mono binaries could be built on either.

cdmihai:
https://github.com/dotnet/corefx/issues/8398 suggests this approach as well.

PR #1013, opened 2016-09-09
Create NuGet packages for all MSBuild class libraries

### LINKED ISSUE #692: Create NuGet packages for all MSBuild assemblies

jeffkl:
Checked in the generated ref assembly source files and added projects that build them.  Then I added a project to create the NuGet packages.  I tested this in our official build which now has to build Release then Release-NetCore, and finally create NuGet packages.  

Fixes #692, #942 #682, and #556.

### DIFF
--- a/build/NuGetPackages/CreateNuGetPackages.proj
+++ b/build/NuGetPackages/CreateNuGetPackages.proj
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+
+  <Import Project="$(SourceDir)BuildValues.props" />
+  
+  <PropertyGroup>
+    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
+    <NuGetPackageLicenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</NuGetPackageLicenseUrl>
+    <NuGetPackageIconUrl>https://go.microsoft.com/fwlink/?linkid=825694</NuGetPackageIconUrl>
+    <NuGetPackageProjectUrl>http://go.microsoft.com/fwlink/?LinkId=624683</NuGetPackageProjectUrl>
+    <PackagesBasePath>$([System.IO.Path]::GetFullPath('$(BaseOutputPathWithConfig)..'))</PackagesBasePath>
+    <PackagesOutDir>$([System.IO.Path]::Combine($(PackagesBasePath), 'Packages'))</PackagesOutDir>
+
+    <SkipVersionGeneration>true</SkipVersionGeneration>
+    <BaseVersion>15.1.0-preview</BaseVersion>
+    <BuildNumberMajor>$(RevisionNumber)</BuildNumberMajor>
+    <BuildNumberMinor>$([System.DateTime]::Now.ToString(`yMMdd`))</BuildNumberMinor>
+
+    <!-- The version for msbuild nuget packages. Used by packages.targets:AddNuGetPackageVersionMetadataToNuspecs -->
+    <PackageVersion>$(BaseVersion)-$(BuildNumberMajor)-$(BuildNumberMinor)</PackageVersion>
+
+    <!-- Turn off the automated package generation in packages.targets:AddNuGetPackageVersionMetadataToNuspecs and use the above PackageVersion-->
+    <DoNotGeneratePackageVersion>true</DoNotGeneratePackageVersion>
+  </PropertyGroup>
+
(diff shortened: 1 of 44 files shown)


build/NuGetPackages/Microsoft.Build.Conversion.Core.nuspec
```diff
@@ -0,0 +1,26 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
```
rainersigwald: These `.nuspec`s are going to conflict with the ones @tmeschter introduced in `master` with 6f17735d76a3c9fb11a93795fec98572c35a2b10, aren't they? Can you coordinate on location/content?

jeffkl: Yeah I think I'm going to just delete the one's under src/nuget.  These will replace them anyway and the only reason their under src/nuget is to use BuildTools which we can't really use anymore with this change. 

---
In reply to: [78386596](https://github.com/Microsoft/msbuild/pull/1013#discussion_r78386596) [](ancestors = 78386596)

build/NuGetPackages/CreateNuGetPackages.proj
```diff
@@ -0,0 +1,75 @@
...
+      $(TraversalBuildDependsOn);
+      BuildPackagesEx;
+    </TraversalBuildDependsOn>
+  </PropertyGroup>
+
+  <Target Name="BuildPackagesEx" Inputs="%(NuSpecs.Identity);$(MSBuildThisFileFullPath)" Outputs="$(PackagesOutDir)\%(NuSpecs.Filename).$(PackageVersion).nupkg">
+    
+    <ItemGroup>
+      <NuspecProperties Include="version=$(PackageVersion)"/>
+      <NuspecProperties Include="id=%(NuSpecs.Filename)"/>
```
rainersigwald: Style: since you're already batching at the target level, I'd prefer you use an item transform like `@(NuSpecs->'%(Filename)')` instead of using syntax inside the target that implies re-batching.

targets/GenApi.targets
```diff
@@ -0,0 +1,26 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
rainersigwald: Long-term, we should just be using the logic included in the GenAPI package's targets/props, right? So this file will be mostly unnecessary?

rainersigwald:
Can you amend the initial-source-checkin commits to include the generation instructions (even it it's just "ran a build after establishing the infrastructure")?

rainersigwald:
:shipit:

targets/GenApi.targets
jeffkl: This is here to allow the enabling and disabling of the functionality.  If GenerateReferenceAssemblySources is false, then the GenApi package targets are not imported.  I could have put all of this in dirs.targets but thought it looked cleaner it it's own file.  What's your opinion?

---
In reply to: [78388700](https://github.com/Microsoft/msbuild/pull/1013#discussion_r78388700) [](ancestors = 78388700)

rainersigwald: For now: 👍. Long-term, I would expect this to be accomplished by a conditional NuGet package reference.

tmeschter:
LGTM. Though merging the .nuspec files between master and xplat is going to be fun...

jeffkl:
With this functionality, the source code is created during build.  So there's no instructions, you just build.

---
In reply to: [246374883](https://github.com/Microsoft/msbuild/pull/1013#issuecomment-246374883) [](ancestors = 246374883)

build/NuGetPackages/CreateNuGetPackages.proj
jeffkl: I tried using the transform but it didn't like it:

error MSB4012: The expression "id=@(NuSpecs->'%(Filename)')" cannot be used in this context. Item lists cannot be concatenated with other strings where an item list is expected. Use a semicolon to separate multiple item lists.

I was able to do use it in other places within this target though

---
In reply to: [78387425](https://github.com/Microsoft/msbuild/pull/1013#discussion_r78387425) [](ancestors = 78387425)

rainersigwald:
@dotnet-bot test this please

In the hopes that this fixes:
image

rainersigwald:

I assume this isn't related:
```

MESSAGE:
The directory '/var/folders/46/45rqwh295fg741sg3jkf4s_m0000gn/T/TempDirForMSBuildUnitTests/src/foo/obj' should exist but doesn't.\nExpected: True\nActual: False
+++++++++++++++++++
STACK TRACE:
at Microsoft.Build.UnitTests.ProjectExtensionsImportTestBase.DoesNotImportProjectWhenDisabled() in /Users/dotnet-bot/j/workspace/Microsoft_msbuild/_OSX_CoreCLR_prtest/src/XMakeTasks/UnitTests/ProjectExtensionsImportTestBase.cs:line 97
```

jeffkl:

No, it _shouldn't_ be related :confused:

PR #1015, opened 2016-09-12
XML doc files for our assemblies

jeffkl:
I'm going to include these in the NuGet packages.  But we're missing some comments on stuff so I started this pull request to get through them all.

Please verify my doc comment updates are accurate.

### DIFF
--- a/src/dir.props
+++ b/src/dir.props
@@ -7,8 +7,4 @@
     <SolutionFile Condition="'$(OsEnvironment)'=='Windows_NT'">$([System.IO.Path]::Combine($(SolutionRoot),"MSBuild.sln"))</SolutionFile>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(EnableDocumentationFile)' == 'true'">
-    <DocumentationFile>$(OutputPath)$(TargetName).xml</DocumentationFile>
-  </PropertyGroup>
-
 </Project>
--- a/src/dir.props
+++ b/src/dir.props
@@ -7,4 +7,8 @@
     <SolutionFile Condition="'$(OsEnvironment)'=='Windows_NT'">$([System.IO.Path]::Combine($(SolutionRoot),"MSBuild.sln"))</SolutionFile>
   </PropertyGroup>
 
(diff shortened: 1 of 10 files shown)


jeffkl:

src/dir.props
```diff
@@ -7,4 +7,8 @@
     <SolutionFile Condition="'$(OsEnvironment)'=='Windows_NT'">$([System.IO.Path]::Combine($(SolutionRoot),"MSBuild.sln"))</SolutionFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(EnableDocumentationFile)' == 'true'">
+    <DocumentationFile>$(OutputPath)$(TargetName).xml</DocumentationFile>
```
rainersigwald: Does this need to be in props? It'd be nice to avoid the props-before-the-props-import pattern.

jeffkl: The DocumentationFile property must be set before importing the targets

[https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets#L155](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets#L155)

I guess this could be inside the dirs.targets before it imports Microsoft.CSharp.targets.  Is that preferable?

---
In reply to: [78440641](https://github.com/Microsoft/msbuild/pull/1015#discussion_r78440641) [](ancestors = 78440641)

rainersigwald: Yeah, I'd prefer it in the targets to keep project files streamlined.

jeffkl: Is it okay to introduce this $(EnableDocumentationFile) property?

---
In reply to: [78442785](https://github.com/Microsoft/msbuild/pull/1015#discussion_r78442785) [](ancestors = 78442785)

rainersigwald: Yeah, the property works for me.

---
In reply to: [78443213](https://github.com/Microsoft/msbuild/pull/1015#discussion_r78443213) [](ancestors = 78443213,78442785)

PR #1017, opened 2016-09-12
Add Microsoft.VisualStudio.v15.Common.props

AndyGerlicher:
This file is currently being dropped into VS internally.


rainersigwald:
:shipit:

PR #1018, opened 2016-09-12
Traversal targets for cross-targeting

nguerrera:
This implements the portion of cross-targeting that we agreed must happen first in msbuild proper and as soon as possible to uncover any unforseen breaking impact on existing projects.

1. Redirect Microsoft.VisualBasic/CSharp.targets to new Microsoft.Common.Traversal.targets if $(TargetFrameworks) is set, but $(TargetFramework) is empty.
2. Implement the protocol where every project will pass its TFM down to project references and get the properties that configure the destination to the best compatible target.
3. For now, the common traversal targets just import based on a variable that we will set in a props hook, but ultimately some general logic should probably be live there, and there should be a way for arbitrary nuget packages to participate. This needs design we have not yet done so going with a minimal, unblocking start.

### DIFF
--- a/src/XMakeTasks/Microsoft.CSharp.targets
+++ b/src/XMakeTasks/Microsoft.CSharp.targets
@@ -17,12 +17,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <!-- 
-        We are doing a cross-targeting traversal build if there is no list of target frameworks specified
+        We are doing a cross-targeting build if there is no list of target frameworks specified
         nor is there a current target framework being built individually. In that case, this import is
-        redirected to Microsoft.Common.Traversal.targets.
+        redirected to Microsoft.Common.CrossTargeting.targets.
    -->
    <PropertyGroup Condition="'$(TargetFrameworks)' != '' and '$(TargetFramework)' == ''">
-      <IsTraversalBuild>true</IsTraversalBuild>
+      <IsCrossTargetingBuild>true</IsCrossTargetingBuild>
    </PropertyGroup>
 
    <!--
@@ -49,9 +49,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <CscToolPath Condition="'$(CscToolPath)' == '' and '$(BuildingInsideVisualStudio)' != 'true'">$(MsBuildFrameworkToolsPath)</CscToolPath>
          </PropertyGroup>
       </When>
-      <When Condition="'$(IsTraversalBuild)' == 'true'">
+      <When Condition="'$(IsCrossTargetingBuild)' == 'true'">
          <PropertyGroup>
-            <CSharpTargetsPath>$(MSBuildToolsPath)\Microsoft.Common.Traversal.targets</CSharpTargetsPath>
+            <CSharpTargetsPath>$(MSBuildToolsPath)\Microsoft.Common.CrossTargeting.targets</CSharpTargetsPath>
           </PropertyGroup>
       </When>
       <Otherwise>
--- a/src/XMakeTasks/Microsoft.CSharp.targets
+++ b/src/XMakeTasks/Microsoft.CSharp.targets
@@ -17,6 +17,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <!-- 
+        We are doing a cross-targeting traversal build if there is no list of target frameworks specified
+        nor is there a current target framework being built individually. In that case, this import is
+        redirected to Microsoft.Common.Traversal.targets.
+   -->
+   <PropertyGroup Condition="'$(TargetFrameworks)' != '' and '$(TargetFramework)' == ''">
+      <IsTraversalBuild>true</IsTraversalBuild>
+   </PropertyGroup>
+
+   <!--
         In VS 2010 SP1 and VS 2012, both supported for asset compatibility, the MSBuild installed 
         as part of them did not enforce using the local ToolsVersion (4.0) in all cases, but instead 
         just used whatever ToolsVersion was in the project file if it existed on the machine, and 
@@ -39,6 +48,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
                  MSBuildToolsPath, which would be incorrect in this case -->
             <CscToolPath Condition="'$(CscToolPath)' == '' and '$(BuildingInsideVisualStudio)' != 'true'">$(MsBuildFrameworkToolsPath)</CscToolPath>
          </PropertyGroup>
+      </When>
+      <When Condition="'$(IsTraversalBuild)' == 'true'">
+         <PropertyGroup>
+            <CSharpTargetsPath>$(MSBuildToolsPath)\Microsoft.Common.Traversal.targets</CSharpTargetsPath>
+          </PropertyGroup>
       </When>
       <Otherwise>
          <PropertyGroup>
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1508,8 +1508,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         Targets="GetTargetFrameworkProperties"
         BuildInParallel="$(BuildInParallel)"
         Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); ReferringTargetFramework=$(TargetFrameworkMoniker)"
-        Condition="'%(_MSBuildProjectReferenceExistent.BuildReference)' == 'true' and '$(BuildingProject)' == 'true'"
-        ContinueOnError="$(ContinueOnError)"
+        ContinueOnError="!$(BuildingProject)"
         RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
 
       <Output TaskParameter="TargetOutputs" PropertyName="_ProjectReferenceTargetFrameworkConfiguration" />
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1511,17 +1511,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         ContinueOnError="!$(BuildingProject)"
         RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
 
-      <Output TaskParameter="TargetOutputs" PropertyName="_ProjectReferenceTargetFrameworkConfiguration" />
+      <Output TaskParameter="TargetOutputs" PropertyName="_ProjectReferenceTargetFrameworkProperties" />
     </MSBuild>
 
     <ItemGroup>
       <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.Identity)' == '%(Identity)'">
-        <SetTargetFramework>$(_ProjectReferenceTargetFrameworkConfiguration)</SetTargetFramework>
+        <SetTargetFramework>$(_ProjectReferenceTargetFrameworkProperties)</SetTargetFramework>
       </_MSBuildProjectReferenceExistent>
     </ItemGroup>
 
     <PropertyGroup>
-      <_ProjectReferenceTargetFrameworkConfiguration />
+      <_ProjectReferenceTargetFrameworkProperties />
     </PropertyGroup>
   </Target>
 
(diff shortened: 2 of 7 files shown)


src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1490,6 +1490,82 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    target framework.
+
+    The referring project's $(TargetFrameworkMoniker) is passed 
+    in as $(ReferringTargetFramework)
+  -->
+  <Target Name="GetTargetFrameworkProperties" />
+
+  <!--
+    ============================================================
+                                        PrepareProjectReferences
```
rainersigwald: This is a nice improvement 👍

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1490,6 +1490,82 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    Builds the GetTargetFrameworkProperties target of all existent project references,
+    passing $(TargetFrameworkMoniker) as $(ReferringTargetFramework) and sets the
+    SetTargetFramework metadata of the project reference to the value that is returned.
+
+    This allows a cross-targeting project to select how it should be configured to
+    build against the most appropriate target for the referring target framework.
+
+    ======================================================================================
+  -->
+  <Target Name="_GetProjectReferenceTargetFrameworkProperties" Outputs="%(_MSBuildProjectReferenceExistent.Identity)">
```
rainersigwald: Is the `Outputs` here necessary? I don't see it used in this commit's diff.

rainersigwald: Or is it here to batch the target so the loose `%(Identity)` metadata below works?

---
In reply to: [78624878](https://github.com/Microsoft/msbuild/pull/1018#discussion_r78624878) [](ancestors = 78624878)

nguerrera: > Or is it here to batch the target so the loose %(Identity) metadata below works?
Yes. This whole thing feels messy to me, but I couldn't find a better way to express it. I imagined there would be a nice way to say "invoke this task for each item and put an output value from the task in to metadata for the item", but I couldn't find one. Am I just missing the idiomatic way to do this?

src/XMakeTasks/Microsoft.CSharp.targets
```diff
@@ -17,6 +17,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <!-- 
+        We are doing a cross-targeting traversal build if there is no list of target frameworks specified
+        nor is there a current target framework being built individually. In that case, this import is
+        redirected to Microsoft.Common.Traversal.targets.
+   -->
```
rainersigwald: Am I right that this needs to be defined in whatever is the top-level import, because the goal is to short-circuit most of the build logic for the outer build?

nguerrera: Correct.

src/XMakeTasks/Microsoft.Common.Traversal.targets
```diff
@@ -0,0 +1,16 @@
...
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(CoreTraversalTargetsPath)"
```
rainersigwald: What defines this?

nguerrera: I will define it in our after-props hook. We are only redirecting targets, not props so it will work. We need to discuss the outer build extensibility and how nuget packages participate, etc. This is just enough to (1) unblock my follow up work (2) smoke test that the common targets hijacking doesn't interfere with existing projects.

nguerrera:
Ready for review, @AndyGerlicher @cdmihai @srivatsn @davkean.

nguerrera:
FYI, @dotnet/project-system

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: I'm not seeing a great option, either. Still poking at a thing or two . . .

---
In reply to: [78629179](https://github.com/Microsoft/msbuild/pull/1018#discussion_r78629179) [](ancestors = 78629179)

rainersigwald: Ok, I have an approach that allows building the refs in parallel. But I'm not sure it's actually better--it requires adding redundant metadata to `_MSBuildProjectReferenceExistent` (to allow for [shared-metadata batching](https://msdn.microsoft.com/en-us/library/ms171474.aspx#Anchor_1) on the path of the project):

```XML
  <Target Name="b">
    <ItemGroup>
      <_MSBuildProjectReferenceExistent>
        <MSBuildSourceProjectFile>%(FullPath)</MSBuildSourceProjectFile>
    </ItemGroup>

    <MSBuild
        Projects="@(_MSBuildProjectReferenceExistent)"
        Targets="GetTargetFrameworkProperties"
        BuildInParallel="$(BuildInParallel)">
      <Output TaskParameter="TargetOutputs" ItemName="_ProjectReferenceTargetFrameworkConfiguration" />
    </MSBuild>

    <ItemGroup>
      <_MSBuildProjectReferenceExistent Condition="'%(MSBuildSourceProjectFile)' != '' and '@(_MSBuildProjectReferenceExistent->'%(FullPath)')' == '@(_ProjectReferenceTargetFrameworkConfiguration->'%(MSBuildSourceProjectFile)')'">
        <Setting>@(_ProjectReferenceTargetFrameworkConfiguration->'%(Identity)')</Setting>

      <Test Include="@(_MSbuildprojectreferenceexistent)" />
    </ItemGroup>
  </Target>
```

Salient features:
* Add `MSBuildSourceProjectFile` metadata to `_MSBuildProjectReferenceExistent`
* Build in parallel, using an **Item** to catch the output (it'll automatically have `MSBuildSourceProjectFile`)
* Shared-metadata-batch on that to get refs one at a time
* set the desired metadata from the (also now one-at-a-time) output of the query

I'm not sure which use of obscure hard-to-reason-about batching features is better. This way is probably slightly more performant (because of parallelism in the remote builds). @cdmihai @AndyGerlicher @jeffkl--thoughts?

nguerrera: This looks fine. I'm not finding the documentation that causes "it'll automatically have MSBuildSourceProjectFile" to hold. Can you point me to a specific passage?

rainersigwald: I can't deep-link to the right table item, but it's under `TargetOutputs` [here](https://msdn.microsoft.com/en-us/library/z7f65y0d.aspx):

>The TargetOutputs parameter also contains the following metadata:
> * MSBuildSourceProjectFile: The MSBuild project file that contains the target that set the outputs.
> * MSBuildSourceTargetName: The target that set the outputs.

nguerrera: Got it. I was imagining it was some magic flow, but it's just what's set on TargetOutputs. :)

rainersigwald:
I'm not a fan of calling this "traversal". Inside Microsoft, there's a long tradition of "traversal projects" (traditionally `dirs.proj`) that are lightweight directory-hierarchy traversers. MSBuild has this concept as well, for instance in [BuildRequestConfiguration](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs#L75).

That said, I'm not sure what the right name for this concept is. "Inner" and "outer" as we've been casually calling it isn't very descriptive. Maybe something like `CrossTargetingCoordinating`? That's certainly not elegant.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
nguerrera: I'm having trouble getting things to work with this change. :( Given the time constraint (we want this in today so that the new (hopefully-not-but-potentially-breaking) targets are in Preview 5. Would you be ok iterating on the final pattern here as an immediate follow-up change?

nguerrera:
I'm happy to rename but we need a quick decision if this is really going in tonight. Traversal was what we had said in Friday's design meeting and I reiterated in my notes sent immediately after that. `CrossTargetCoordinating` is too long. Maybe we just go with what we've been saying and use `Outer`. I know @srivatsn doesn't like `Outer` though.

srivatsn:
I'll defer to the MSBuild folks wrt naming here.

src/XMakeTasks/Microsoft.CSharp.targets
```diff
@@ -17,6 +17,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <!-- 
+        We are doing a cross-targeting traversal build if there is no list of target frameworks specified
```
cdmihai: If there's a design doc on some github SDK page about what a traversal build is, this comment would be a good spot to reference it in case users want to understand the bigger picture. :)

src/XMakeTasks/Microsoft.CSharp.targets
```diff
@@ -17,6 +17,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        nor is there a current target framework being built individually. In that case, this import is
+        redirected to Microsoft.Common.Traversal.targets.
+   -->
+   <PropertyGroup Condition="'$(TargetFrameworks)' != '' and '$(TargetFramework)' == ''">
```
cdmihai: If there's only a single target framework specified, then the thing still loops, right? I remember some talk about single target frameworks not doing a loop. Did that change?

nguerrera: IF the single target framework is specified in TargetFrameworks, then yes it loops. You can also configure yourself as an inner build and we won't loop, but then there should be no TargetFrameworks with an s and this condition works too.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
AndyGerlicher: Let's go with the implementation that works we can adjust later.

rainersigwald: Yeah, that's fine.

AndyGerlicher:
How about `Microsoft.Common.CrossTargeting.targets`?

nguerrera:
> How about Microsoft.Common.CrossTargeting.targets?
OK.

nguerrera:
And IsCrossTargetingBuild to mean that we're in the outer build. Inner build is non cross-targeting since it is one TFM. I think this is a fine way to go.

AndyGerlicher:
LGTM once `IsTraversalBuild` is renamed.

nguerrera:
@AndyGerlicher Thanks and done.

nguerrera:
@AndyGerlicher @cdmihai @rainersigwald Thanks again! Can you also help me get this ported to xplat and published to a package that the cli can consume?

cdmihai:
@nguerrera 

I think this is how you get into xplat:
- open another PR from your branch targetting the xplat branch
- add the targets to the following nuspec: https://github.com/Microsoft/msbuild/blob/xplat/src/nuget/Microsoft.Build.Targets.nuspec

@jeffkl, does that look alright?

kzu:
Where is the cross targetting landing on the C#/VB/NETCore side? Need to peek to ensure NuGetizer is forward-compatible (ideally?)

nguerrera:
@cdmihai @jeffkl I've done this as #1025.

rainersigwald:
For posterity: this change caused an error in an internal project that had a custom target defined as:

```xml
<Target Name="ValidateReferences"
        DependsOnTargets="
            $(ResolveReferencesDependsOn);
            ValidateExportsFilesExist; 
            ValidatePackageReferences; 
            ValidateBinplace"
        AfterTargets="$(ResolveReferencesDependsOn)">
```

Note that it hooks after (any of) `$(ResolveReferencesDependsOn)` but then requires (all of) `$(ResolveReferencesDependsOn)` to run. Adding this target created a cycle on Clean:

Clean depends on CleanReferencedProjects depends on PrepareProjectReferences depends on AssignProjectConfiguration.

ValidateReferences has an AfterTargets for AssignProjectConfiguration, so ValidateReferences attempts to run, but first we have to run its DependsOnTargets, so we try to add them to the target stack, but when we consider PrepareProjectReferences, it’s already on the stack, so trying to add it would create a cycle, and we fail.

The error was

```
error MSB4006: There is a circular dependency in the target dependency graph involving target "PrepareProjectReferences".
```

This can be fixed by changing the custom target's hook:

```diff
             ValidateExportsFilesExist; 
             ValidatePackageReferences; 
             ValidateBinplace"
-          AfterTargets="$(ResolveReferencesDependsOn)">
+          BeforeTargets="ResolveReferences">
 
     <Error Text="@(ValidationError->'&#xa;%(Identity): %(Text)','')"
            Condition="@(ValidationError) != '' and $(ValidateProjectSettings) == 'true'"/>
```

PR #1019, opened 2016-09-12
Set AutoUnifyAssemblyReferences only if it hasn't already been set

srivatsn:
For .NET Core projects, we want to pass AutoUnify=true to the RAR task. That gets set through the AutoUnifyAssemblyReferences property which derives it value from the AutoGenerateBindingRedirects property. It’s odd to have AutoGenerateBindingRedirects=true in the NETCore.props given binding redirects don’t mean anything in .NETCore. This change makes it so that we can set AutoUnifyAssemblyReferences in the props file.

/cc @rainersigwald @cdmihai @AndyGerlicher


rainersigwald:
This seems reasonable to me, and I can't think of a way it would break things.

:shipit:

AndyGerlicher:
:shipit:

PR #1021, opened 2016-09-14
Isolate Exec-doesn't-pollute-TEMP test files

rainersigwald:
I've seen this test fail because there are different numbers of files in
TEMP before and after the test--but it's always because something else was
running at the same time and using temp files.


jeffkl:
:ship: it

rainersigwald:
@dotnet-bot test this please

(Something racily horked the state in the middle of `LogBuildStartedCriticalOnly` on two different flavors⁉️)

AndyGerlicher:
Nice! I did a really careful search for all code that changed it explicitly a while back, but I didn't think to search for a change to the env variable directly. I will be happy when these never happen again!

PR #1035, opened 2016-09-16
Make FileMatcher more friendly with forward and backward slashes

### LINKED ISSUE #724: Item Exclude syntax doesn't support '/' as a path separator
1. Create a new Console application
2. Create a new folder called "Exclude", copy Program.cs into it
3. Unload -> Edit
4. Replace the `<Compile Include` items with `<Compile Include="**/*.cs" Exclude="Excluded/**" />`
5. Reload project

Expected: For Excluded\Program.cs to not be included in the project
Actual: Excluded\Program.cs is included in the project.

It appears that while '/' is understood in Include, it's not in Exclude. Does this prevent us from writing platform-agnostic paths?

cdmihai:
Fixes #724 

While excludes with slashes currently work crossplatform, they mostly work out of accident. This PR makes FileMatcher work with both slashes and backslashes. Also wrote more tests.

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -56,6 +56,9 @@ namespace Microsoft.Build.Shared
             cacheDirectory = null;
         }
 
+        // TODO: assumption on file system case sensitivity: https://github.com/Microsoft/msbuild/issues/781
+        internal static readonly StringComparison PathComparison = StringComparison.OrdinalIgnoreCase;
+
         /// <summary>
         /// Retrieves the MSBuild runtime cache directory
         /// </summary>
@@ -327,7 +330,7 @@ namespace Microsoft.Build.Shared
                        // is available here - we don't want managed apps mucking 
                        // with this for security reasons.
                     */
-                    if (startIndex == finalFullPath.Length || finalFullPath.IndexOf(@"\\?\globalroot", StringComparison.OrdinalIgnoreCase) != -1)
+                    if (startIndex == finalFullPath.Length || finalFullPath.IndexOf(@"\\?\globalroot", PathComparison) != -1)
                     {
                         finalFullPath = Path.GetFullPath(finalFullPath);
                     }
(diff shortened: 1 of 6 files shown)


src/Shared/UnitTests/FileMatcher_Tests.cs
```diff
@@ -902,9 +984,10 @@ internal int FileHits3
             /// <param name="pattern">The pattern to search for.</param>
             /// <param name="files">Hashtable receives the files.</param>
             /// <returns></returns>
-            private int GetMatchingFiles(string[] candidates, string path, string pattern, Hashtable files)
+            private int GetMatchingFiles(string[] candidates, string path, string pattern, ISet<string> files)
```
rainersigwald: :+1:

src/Shared/FileUtilities.cs
```diff
@@ -836,5 +837,36 @@ internal static string CombinePaths(string root, params string[] paths)
...
+        /// Replace all backward slashes to forward slashes
+        /// </summary>
+        internal static string ToSlash(this string s)
+        {
+            return s.Replace('\\', '/');
+        }
+
+        internal static string NormalizeForPathComparison(this string s) => s.ToSlash().TrimTrailingSlashes();
+
+        internal static bool PathsEqual(string path1, string path2)
```
rainersigwald: In an ideal world this wouldn't be so allocatey. We should keep an eye out--if we see increased GC pressure, this could be a culprit (depends how often it's called).

cdmihai: It potentially gets called a lot. Should I hold off the PR until I do some measurements ~next week?

One alternative would be to convert the string to byte arrays and work on those, in-place. But you'd still have 2 new arrays for each call. Another alternative is to rewrite the file matcher from scratch and try to work around the need to do this equality

rainersigwald: It may not be too bad, though I do wish we had a better way to know. I'm certainly not suggesting rewriting it from scratch just for this hunch! I don't think we need to hold this PR for this concern, but we should definitely have this in the list of things to blame if there's a slowdown.

PR #1036, opened 2016-09-16
Support preserving formatting when opening a project

### LINKED ISSUE #707: Round trip formatting / comments
**This issue is in the open design phase and is a part of the [project.json -> csproj ](https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/) conversion effort.**

We should preserve user authored content/whitespace in project files when saving.
E.g.
```XML
<PropertyGroup>
  <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion> <!-- 4.5.2 for back-compat -->
</PropertyGroup>
```
Today will be re-saved as:
```XML
<PropertyGroup>
  <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion> 
  <!-- 4.5.2 for back-compat -->
</PropertyGroup>
```

We should investigate to see if there's an easy win here, but we should not be creating our own XML writer for this release.

dsplaisted:
Fixes #707

Adds the following overloads to `ProjectRootElement`:

```C#
public static ProjectRootElement Create(XmlReader xmlReader, ProjectCollection projectCollection, bool preserveFormatting)
public static ProjectRootElement Open(string path, ProjectCollection projectCollection, bool preserveFormatting)
```


rainersigwald:
We need to make sure this doesn't conflict with #1004, where @maddin2016 is playing in similar areas.

src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs
```diff
@@ -227,6 +227,8 @@ internal ProjectRootElement Get(string projectFile, OpenProjectRootElement openP
                             // use: it checks the file content as well as the timestamp. That's better than completely disabling
                             // the cache as we get test coverage of the rest of the cache code.
                             XmlDocument document = new XmlDocument();
+                            //  TODO: does PreserveWhitespace need to be an option here?
+                            //document.PreserveWhitespace = true;
```
rainersigwald: Seems like it'd need to match what gets inserted into the cache, which would depend on how it was loaded, right?

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -170,6 +172,18 @@ public void InsertAfterChild(ProjectElement child, ProjectElement reference)
...
+            {
+                //  If we are trying to preserve formatting of the file, then the new node won't automatically be indented.
+                //  So try to match the surrounding formatting by checking the whitespace that precedes the node we inserted
+                //  after, and inserting the same whitespace between the previous node and the one we added
+                if (reference.XmlElement.PreviousSibling != null &&
+                    reference.XmlElement.PreviousSibling.NodeType == XmlNodeType.Whitespace)
+                {
+                    var newWhitespaceNode = XmlDocument.CreateWhitespace(reference.XmlElement.PreviousSibling.Value);
+                    XmlElement.InsertAfter(newWhitespaceNode, reference.XmlElement);
+                }
```
rainersigwald: It's unfortunate that we have to do this manually. But this seems like a reasonable strategy (and good comment).

rainersigwald: But can it be abstracted and reused?

rainersigwald:
>         /// DO NOT BREAK THIS VERY USEFUL SAFETY CONTRACT.

Does this still hold?

---
Refers to: src/XMakeBuildEngine/Construction/ProjectElementContainer.cs:305 in 1a2e1da. [](commit_id = 1a2e1dab5dc928252b1b90219dc98a78a0e01c4a, deletion_comment = False)

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -437,6 +476,36 @@ private void AddInitialChild(ProjectElement child)
...
+                    //  a new line with the same indentation.
+                    string whitespace = XmlElement.FirstChild.Value;
+                    XmlElement.FirstChild.Value = whitespace + DEFAULT_INDENT;
+                    var newWhitespaceNode = XmlDocument.CreateWhitespace(whitespace);
+                    XmlElement.InsertAfter(newWhitespaceNode, child.XmlElement);
+                }
+                else if (XmlElement.PreviousSibling != null &&
+                         XmlElement.PreviousSibling.NodeType == XmlNodeType.Whitespace)
+                {
+                    //  This container didn't have any whitespace in in.  This probably means it didn't have separate open
```
rainersigwald: in **it**

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -437,6 +476,36 @@ private void AddInitialChild(ProjectElement child)
...
+                    //  and close tags.  So add a whitespace node before the new child with additional indentation over the
+                    //  container's indentation, and add a whitespace node with the same level of indentation as the container
+                    //  after the new child so the closing tag will be indented properly.
+                    string whitespace = XmlElement.PreviousSibling.Value;
```
rainersigwald: rename to `parentWhitespace`?

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectFormatting_Tests.cs
```diff
@@ -0,0 +1,462 @@
...
+    </ProjectReference>
+  </ItemGroup>
+</Project>");
+
+            string reformattedContent = ObjectModelHelpers.CleanupFileContents(@"
+<Project DefaultTargets=`Build` ToolsVersion=`msbuilddefaulttoolsversion` xmlns=`msbuildnamespace`>
+  <ItemGroup>
+    <ProjectReference Include=`..\CLREXE\CLREXE.vcxproj`>
+      <!-- Comment -->
+      <Project>{3699f81b-2d03-46c5-abd7-e88a4c946f28}</Project>
```
rainersigwald: I'm not sure I prefer this behavior. It can be useful to have both
```C
// Long comments explaining
code;
```

and
```C
code; // clarified
```

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectFormatting_Tests.cs
```diff
@@ -0,0 +1,462 @@
...
+            ProjectRootElement xml = ProjectRootElement.Create(XmlReader.Create(new StringReader(content)), ProjectCollection.GlobalProjectCollection,
+                preserveFormatting: true);
+            Project project = new Project(xml);
+
+            StringWriter writer = new StringWriter();
+
+            project.SaveLogicalProject(writer);
+
+            string actual = writer.ToString();
+            string expected = @"<?xml version=""1.0"" encoding=""utf-16""?>" +
```
rainersigwald: attn: @andygerlicher

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectFormatting_Tests.cs
```diff
@@ -0,0 +1,462 @@
...
+    <Compile Include=""Program.cs"" />
+  </ItemGroup>
+</Project>");
+
+            string actual = writer.ToString();
+
+            VerifyAssertLineByLine(expected, actual);
+        }
+
+        [Fact(Skip = "https://github.com/Microsoft/msbuild/issues/362")]
```
rainersigwald: So the preserve-whitespace flag doesn't fix this? Boo :(

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
dsplaisted: How do you mean abstracted and reused?  ProjectElementContainer seems like it is already the generic place to put this logic.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -437,6 +476,36 @@ private void AddInitialChild(ProjectElement child)
 
             XmlElement.AppendChild(child.XmlElement);
 
+            if (XmlDocument.PreserveWhitespace)
+            {
+                //  If we are trying to preserve formatting of the file, then the new node won't automatically be indented.
+                //  So try to match the surrounding formatting and add one indentation level
+                if (XmlElement.FirstChild.NodeType == XmlNodeType.Whitespace)
+                {
+                    //  This container had a whitespace node, which should generally be a newline and the indent
```
dsplaisted: I thought about this and it seemed OK to leave it as is.  If there's spacing between the elements but not a newline, then we'll still be matching the surrounding formatting, even though the formatting seems weird.

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectFormatting_Tests.cs
dsplaisted: The behavior today is that comments are put on separate lines.  If you specify the option to preserve formatting, then the location of the comment will be preserved.  This test covers both behaviors, which is why you see the reformatted version.

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectFormatting_Tests.cs
dsplaisted: Apparently not... :disappointed:

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
rainersigwald: This and below have very similar logic to find, clone, and insert whitespace. Though I guess it's only twice, so the generalization may not be that general.

dsplaisted:
@rainersigwald 

>>         /// DO NOT BREAK THIS VERY USEFUL SAFETY CONTRACT.
>
>Does this still hold?
>
>---
>Refers to: src/XMakeBuildEngine/Construction/ProjectElementContainer.cs:305 in 1a2e1da. [](commit_id = 1a2e1dab5dc928252b1b90219dc98a78a0e01c4a, deletion_comment = False)

Yes.  The comment is referring to the NextSibling and PreviousSibling properties of `ProjectElement`, which don't include the whitespace nodes.

src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs
dsplaisted: I've updated this so that it won't use the cached version if it doesn't match the preserveFormatting setting requested.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
dsplaisted: Yeah, I didn't see an elegant way to unify the different cases that exist within `ProjectElementContainer`.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
dsplaisted: Fixed

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
dsplaisted: Done.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
rainersigwald: Ok, I buy it. Can you put that in the comment?

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectFormatting_Tests.cs
rainersigwald: Ohhh. Gotcha.

rainersigwald:
Ah, now I see that I evidently didn't hit "comment" on my _most important comment_. Sorry.

My instinct is to say that this shouldn't be an option--that is, we should always preserve whitespace. But you've plumbed through an option. Do you think that should be exposed to the user?

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
dsplaisted: Done.

dsplaisted:
I went with an opt-in option to avoid a compat risk (as [suggested by Dave](https://github.com/Microsoft/msbuild/issues/707#issuecomment-228195384)). How much of a compat risk would there be to always preserving formatting?

Also note that there's some risk that there are still cases where preserving the formatting will result in ugly formatting when the project is modified, as would be the case without the changes on `ProjectElementContainer` to add or remove whitespace when adding or removing child elements.

davkean:
@dsplaisted Where did we land with this? Did we sync with CPS to call this overload?

dsplaisted:
@davkean No, AFAIK CPS isn't consuming this yet.  I'm not sure if this has made it into a drop of MSBuild that they are using.  @AndyGerlicher, can you comment?

srivatsn:
@lifengl have we reacted to this in CPS?

PR #1037, opened 2016-09-16
Error for copy task if copying multiple sources to the same destination path

### LINKED ISSUE #989: Multiple Items with CopyToOutput and conflicting TargetPath copies last file
A project with multiple items with CopyToOutputDirectory and conflicting assigned TargetPaths will cause the last file to be copied to the TargetPath. 

I'm not certain this is a bug, but the behavior seems non-intuitive. It seems like multiple files trying to output to the same TargetPath should cause an error. 

An example:
```
<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworkIdentifier>.NETCoreApp</TargetFrameworkIdentifier>
    <TargetFrameworkVersion>v1.0</TargetFrameworkVersion>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="**\*.cs" Exclude="$(GlobalExclude)" />
    <EmbeddedResource Include="**\*.resx" Exclude="$(GlobalExclude)" />
    <Content Include="Program.cs;project.json">
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
      <Link>otherfilename.json</Link>
    </Content>
  </ItemGroup>

  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
</Project>
```

only`project.json` will be copied to `otherfilename.json` and no error is thrown.

cc @rainersigwald @cdmihai @eerhardt

maddin2016:
Fixes #989

### DIFF
--- a/src/Utilities/TaskItem.cs
+++ b/src/Utilities/TaskItem.cs
@@ -489,6 +489,7 @@ namespace Microsoft.Build.Utilities
 
             return (IDictionary)_metadata.Clone();
         }
+
         #endregion
     }
 }
--- a/src/Utilities/TaskItem.cs
+++ b/src/Utilities/TaskItem.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Build.Utilities
     /// <comment>
     /// Surprisingly few of these Utilities TaskItems are created: typically several orders of magnitude fewer than the number of engine TaskItems.
     /// </comment>
-    public sealed class TaskItem : MarshalByRefObject, ITaskItem, ITaskItem2
+    public sealed class TaskItem : MarshalByRefObject, ITaskItem, ITaskItem2, IEquatable<TaskItem>
     {
         #region Member Data
 
@@ -488,6 +488,21 @@ namespace Microsoft.Build.Utilities
             }
 
             return (IDictionary)_metadata.Clone();
+        }
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as TaskItem);
+        }
+
+        public override int GetHashCode()
+        {
+            return ItemSpec == null ? 0 : ItemSpec.GetHashCode();
+        }
+
+        public bool Equals(TaskItem other)
+        {
+            return other.ItemSpec == ItemSpec;
         }
 
         #endregion
--- a/src/Utilities/TaskItem.cs
+++ b/src/Utilities/TaskItem.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Build.Utilities
     /// <comment>
     /// Surprisingly few of these Utilities TaskItems are created: typically several orders of magnitude fewer than the number of engine TaskItems.
     /// </comment>
-    public sealed class TaskItem : MarshalByRefObject, ITaskItem, ITaskItem2, IEquatable<TaskItem>
+    public sealed class TaskItem : MarshalByRefObject, ITaskItem, ITaskItem2
     {
         #region Member Data
 
@@ -489,22 +489,6 @@ namespace Microsoft.Build.Utilities
 
             return (IDictionary)_metadata.Clone();
         }
-
-        public override bool Equals(object obj)
-        {
-            return Equals(obj as TaskItem);
-        }
-
-        public override int GetHashCode()
-        {
-            return ItemSpec == null ? 0 : ItemSpec.GetHashCode();
-        }
-
-        public bool Equals(TaskItem other)
-        {
-            return other.ItemSpec == ItemSpec;
-        }
-
         #endregion
     }
 }
--- a/src/XMakeTasks/Copy.cs
+++ b/src/XMakeTasks/Copy.cs
@@ -434,7 +434,9 @@ namespace Microsoft.Build.Tasks
 
                         foreach (var subitem in item)
                         {
-                            logSource += '"' + subitem.GetMetadata("OriginalItemSpec");
+                            var originalPath = subitem.GetMetadata("OriginalItemSpec");
+                            subitem.SetMetadata("CopiedFrom", originalPath);
+                            logSource += '"' + originalPath;
 
                             logSource += !subitem.Equals(lastItem) ? "\"," : "\"";
                         }
--- a/src/XMakeTasks/Copy.cs
+++ b/src/XMakeTasks/Copy.cs
@@ -424,14 +424,14 @@ namespace Microsoft.Build.Tasks
                     foreach (var item in sameDestinationFiles)
                     {
                         string logSource = "";
-                        var lastItem = item.Last();
+                        var sameItemsList = item.ToList();
 
-                        foreach (var subitem in item)
+                        for (int i = 0; i < sameItemsList.Count(); i++)
                         {
-                            var originalPath = subitem.GetMetadata("CopiedFrom");
+                            var originalPath = sameItemsList[i].GetMetadata("CopiedFrom");
                             logSource += '"' + originalPath;
 
-                            logSource += !subitem.Equals(lastItem) ? "\"," : "\"";
+                            logSource += i != sameItemsList.Count - 1 ? "\"," : "\"";
                         }
 
                         string logDestination = '"' + item.Key + '"';
--- a/src/XMakeTasks/Copy.cs
+++ b/src/XMakeTasks/Copy.cs
@@ -410,10 +410,18 @@ namespace Microsoft.Build.Tasks
                 return true;
             }
             
-            if (_destinationFiles != null && _destinationFiles.Distinct().Count() != _destinationFiles.Length)
+            if (_destinationFiles != null)
             {
-                Log.LogErrorFromResources("Copy.SameDestinationPath");
-                return false;
+                var sameDestinationFiles = _destinationFiles.GroupBy(x => x).Where(g => g.Count() > 1).Select(y => y.First());
+
+                if ((sameDestinationFiles.Count() + _destinationFiles.Length) != _destinationFiles.Length)
+                {
+                    foreach(var item in sameDestinationFiles)
+                    {
+                        Log.LogErrorFromResources("Copy.SameDestinationPath", item.ItemSpec);
+                    }
+                    return false;
+                }
             }
 
             if (!(ValidateInputs() && InitializeDestinationFiles()))
--- a/src/XMakeTasks/Copy.cs
+++ b/src/XMakeTasks/Copy.cs
@@ -414,16 +414,33 @@ namespace Microsoft.Build.Tasks
             {
                 return false;
             }
-            
+
             if (_destinationFiles != null)
             {
-                var sameDestinationFiles = _destinationFiles.GroupBy(x => x.ItemSpec).Where(g => g.Count() > 1).Select(y => y.First());
+                var sameDestinationFiles = _destinationFiles.GroupBy(x => x.ItemSpec).Where(g => g.Count() > 1);
 
                 if (sameDestinationFiles.Any())
                 {
+                    for (int i = 0; i < _sourceFiles.Length; ++i)
+                    {
+                        // Copy meta-data from source to destinationFiles.
+                        _sourceFiles[i].CopyMetadataTo(_destinationFiles[i]);
+                    }
+
                     foreach (var item in sameDestinationFiles)
                     {
-                        Log.LogErrorFromResources("Copy.SameDestinationPath", item.ItemSpec);
+                        string logSource = "";
+                        var lastItem = item.Last();
+
+                        foreach (var subitem in item)
+                        {
+                            logSource += '"' + subitem.GetMetadata("OriginalItemSpec");
+
+                            logSource += !subitem.Equals(lastItem) ? "\"," : "\"";
+                        }
+
+                        string logDestination = '"' + item.Key + '"';
+                        Log.LogErrorFromResources("Copy.SameDestinationPath", logSource, logDestination);
                     }
                     return false;
                 }
--- a/src/XMakeTasks/Copy.cs
+++ b/src/XMakeTasks/Copy.cs
@@ -14,6 +14,7 @@ using System.Runtime.InteropServices;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using Microsoft.Build.Shared;
+using System.Linq;
 
 namespace Microsoft.Build.Tasks
 {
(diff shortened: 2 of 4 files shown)


rainersigwald:
We do this in **our own build**? 😂

However, I do think the failure indicates that the error message needs to explicitly call out the source item's identities and destinations to be useful, which will make the implementation more complicated.

cdmihai:
Maybe use GroupBy to print the collisions?

brthor:
Nice 👍

AndyGerlicher:
Thanks for working on this! Looks good, but I do agree with @rainersigwald it really does need to specify which files or this could be very annoying to debug. Also I would like to see a test or two covering the scenario please.

maddin2016:
Monday i can push a commit for tests and specify the sources in error messsage.

maddin2016:
@rainersigwald, can you please look at my last commit. I stumbled upon this while writing the unit test and wonder why `Distinct` method is not working when there are two different objects. See [here](https://github.com/Microsoft/msbuild/pull/1037/files#diff-f55b015e3d3239e1dd087b0fee47ecaeR1774). So i decided to compare the `ItemSpec` value. What do you say?

maddin2016:
Another approach would be
```c#
if (_destinationFiles != null)
            {
                var duplicatedCount = _destinationFiles.Length;
                var duplicatedHash = new HashSet<string>();
                
                for(int i = 0; i < _destinationFiles.Length; i++)
                {
                    if (duplicatedHash.Contains(_destinationFiles[i].ItemSpec))
                    {
                        ++duplicatedCount;
                    }
                    else
                    {
                        duplicatedHash.Add(_destinationFiles[i].ItemSpec);
                    }
                }

                if(duplicatedCount != _destinationFiles.Length)
                {
                    Log.LogErrorFromResources("Copy.SameDestinationPath");
                    return false;
                }
            }
```

src/XMakeTasks/UnitTests/Copy_Tests.cs
```diff
@@ -1766,6 +1766,41 @@ public void TooFewRetriesThrows()
...
+                MockEngine engine = new MockEngine();
+                t.BuildEngine = engine;
+                t.SourceFiles = new ITaskItem[] { new TaskItem(sourceFile), new TaskItem(sourceFile2) };
+                t.DestinationFiles = new ITaskItem[] { fileOut, fileOut2 };
+                t.SkipUnchangedFiles = false;
+                bool success = t.Execute();
+
+                Assert.False(success);
+
+                ((MockEngine)t.BuildEngine).AssertLogContains("MSB3892"); // Didn't do retries, no op then invalid
```
rainersigwald: Comment is wrong here.

src/XMakeTasks/Copy.cs
```diff
@@ -408,6 +409,20 @@ CopyFileWithState copyFile
                 _copiedFiles = new TaskItem[0];
                 return true;
             }
+            
+            if (_destinationFiles != null)
```
rainersigwald: I think we really need to have both conflict-by-name-collision-in-destination-folder and conflict-by-explicit-destination-path errors for this to be fully realized.

Fortunately, it looks like that's pretty easy. Move this block to after the call to `InitializeDestinationFiles`, and you can get the full list, including paths generated by unpacking `DestinationFolder`.

Please also add a test for that situation (copy two files with the same name from different folders into a destination folder and validate that the error arises).

src/XMakeTasks/Copy.cs
```diff
@@ -408,6 +409,20 @@ CopyFileWithState copyFile
...
+            {
+                var sameDestinationFiles = _destinationFiles.GroupBy(x => x).Where(g => g.Count() > 1).Select(y => y.First());
```
rainersigwald: Should the `GroupBy` operate on the `FullPath` metadata instead? That seems more correct--but maybe I'm missing where the item's include is guaranteed to be a full path?

rainersigwald: I'm not so sure selecting the first of every set of duplicates is the best way to go. I think I'd prefer an error message more like "Cannot copy multiple files to the same destination. SourceFiles 'X', 'Y', and 'Z' all have the destination 'Q'."

src/XMakeTasks/Copy.cs
```diff
@@ -408,6 +409,20 @@ CopyFileWithState copyFile
...
+
+                if ((sameDestinationFiles.Count() + _destinationFiles.Length) != _destinationFiles.Length)
```
rainersigwald: Why not just `sameDestinationFiles.Any()`?

src/XMakeTasks/Copy.cs
rainersigwald: Hmm, although localization might make the multiple-inputs-in-one-message thing tricky. Maybe a "related" message for the second + subsequent sources?

src/XMakeTasks/Resources/Strings.resx
```diff
@@ -2606,6 +2606,9 @@
   <data name="Copy.ExactlyOneTypeOfLink">
     <value>MSB3891: Both "{0}" and "{1}" were specified in the project file. Please choose one or the other.</value>    
   </data>
+  <data name="Copy.SameDestinationPath">
+    <value>MSB3892: There are multiple source files that try to write to the same destination file: "{0}</value>
```
rainersigwald: I think we'll be changing this for other reasons but: missing quote and period at the end of the message.

src/Utilities/TaskItem.cs
```diff
@@ -489,7 +489,6 @@ IDictionary ITaskItem2.CloneCustomMetadataEscaped()
 
             return (IDictionary)_metadata.Clone();
         }
-
```
rainersigwald: Can you revert this change too since it's just whitespace?

src/XMakeTasks/Copy.cs
maddin2016: 👍 :no_mouth:

maddin2016:
Here is the error message
```    C:\Users\scholzmartin\Documents\GitHub\msbuild\bin\x86\Windows_NT\Debug\Microsoft.Common.CurrentVersion.targets(4232,
5): error : MSB3892: There are multiple source files that try to write to the same destination file: "bin\Debug\otherfi
lename.js [C:\Users\scholzmartin\Documents\Visual Studio 2015\Projects\ConsoleApplication2\ConsoleApplication2\ConsoleA
pplication2.csproj]
  ```

File: src/XMakeTasks/Copy.cs (hunk 3 above)
Comment (maddin2016) [inline src/XMakeTasks/Copy.cs]: I'm currently trying this
```
var sameDestinationFiles = _destinationFiles.GroupBy(x => x.ItemSpec).Where(g => g.Count() > 1);

                foreach(var item in sameDestinationFiles)
                {
                    var log = "SourceFiles ";

                    foreach(var subitem in item)
                    {
                        log += subitem.GetMetadata("OriginalItemSpec");
                    }

                    log += "all have the destination " + item.Key + ".";
                }
```
Problem here is that i don't have access to `OriginalItemSpec` because it is added [later](https://github.com/maddin2016/msbuild/blob/e0d90d241e267e223fc60254cb45713c2fdb85fd/src/XMakeTasks/Copy.cs#L572) 😞

Comment (maddin2016) [inline src/XMakeTasks/Copy.cs]: There is one point that is not clear to me. Why are `_destinationFiles[]` are present at this time?? Where is the point where `Copy` constructor is called and `_destinationFiles` is set.

Comment (maddin2016) [inline src/XMakeTasks/Copy.cs]: Ahh, i think it is [this](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L4236) point, right?

Comment (maddin2016) [inline src/XMakeTasks/Copy.cs]: @rainersigwald, what do you think about that.
```C#
 if (_destinationFiles != null)
            {
                var sameDestinationFiles = _destinationFiles.GroupBy(x => x.ItemSpec).Where(g => g.Count() > 1);
                
                if (sameDestinationFiles.Any())
                {
                    for(int i = 0; i < _sourceFiles.Length; ++i)
                    {
                        // Copy meta-data from source to destinationFiles.
                        _sourceFiles[i].CopyMetadataTo(_destinationFiles[i]);
                    }

                    foreach (var item in sameDestinationFiles)
                    {
                        string logSource = "";
                        foreach(var subitem in item)
                        {
                            logSource += '"' + subitem.GetMetadata("OriginalItemSpec") + "\",";
                        }

                        string logDestination = '"' + item.Key + '"';
                        Log.LogErrorFromResources("Copy.SameDestinationPath", logSource, logDestination);
                    }
                    return false;
                }
            }
```
then error looks like this
image

rainersigwald: A task's constructor is called by the engine before setting the task inputs and calling `Execute()`.

rainersigwald: That output looks pretty good, modulo grammar nitpicks. But I hope we can avoid doing quite so much work to get it. Could you set metadata while populating `_destinationFiles`--maybe something like `Source` or `CopiedFrom`?

rainersigwald:
@maddin2016 That wasn't quite what I meant by setting the `CopiedFrom` metadata. I was thinking: after populating the `_destinationFiles[i]` element [here](https://github.com/maddin2016/msbuild/blob/cb27e0a9278f00ffaf7a727f440d5a918d14dac6/src/XMakeTasks/Copy.cs#L591), set `CopiedFrom` based on `_sourceFiles[i].ItemSpec`, and then consume it in your new check-for-duplicates code.

maddin2016:
Thats the problem. I can't access that point because destinationFolder is not null!!!

maddin2016:
Thats what I mean with
>Why are _destinationFiles[] are present at this time?

rainersigwald:
@maddin2016 Still not understanding, sorry. At what time?

maddin2016:
[here](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Copy.cs#L530)

rainersigwald:
Ohhhh! 🤦

Got it. It's not null there because it's being explicitly specified via the `DestinationFiles` argument, instead of computed from `DestinationFolder`.

I think you could just loop over `_sourceFiles` and `_destinationFiles` inside `InitializeDestinationFiles` but **outside** that `if`, and set the metadata there. They're [guaranteed](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Copy.cs#L508) to be the same length.

maddin2016:
Great 👍

src/Utilities/TaskItem.cs
```diff
@@ -27,7 +27,7 @@ namespace Microsoft.Build.Utilities
     /// <comment>
     /// Surprisingly few of these Utilities TaskItems are created: typically several orders of magnitude fewer than the number of engine TaskItems.
     /// </comment>
-    public sealed class TaskItem : MarshalByRefObject, ITaskItem, ITaskItem2
+    public sealed class TaskItem : MarshalByRefObject, ITaskItem, ITaskItem2, IEquatable<TaskItem>
```
radical: It would be nice to move this change to `TaskItem` to a separate commit and add explicit `TaskItem` tests for this, not related to the `Copy` task.

maddin2016:
Hi @rainersigwald, dou you have any idea why these tests are failing??

jeffkl:
@maddin2016 The Windows_NT builds are failing because _our_ codebase has a `<Copy />` task with duplicate sources to a single destination.
```
error : MSB3892: Cannot copy multiple files to the same destination. SourceFiles "D:\j\workspace\_Windows_NT_C---c4a97696\src\XMakeTasks\Microsoft.Common.overridetasks""D:\j\workspace\_Windows_NT_C---c4a97696\src\XMakeTasks\Microsoft.Common.overridetasks" all have the same destination "D:\j\workspace\_Windows_NT_C---c4a97696\bin\x86\Windows_NT\Debug\\Microsoft.Common.overridetasks" [D:\j\workspace\_Windows_NT_C---c4a97696\src\XMakeBuildEngine\UnitTests\Microsoft.Build.Engine.UnitTests.csproj]
```

Our build runs twice where the first one uses a shipped MSBuild and then it builds again using the MSBuild binaries that were just built.  This second build is failing because your modifications to the Copy task identify an error in our build logic.  We'll either need to fix our projects to eliminate the duplicate files or add a flag to your implementation to disable the error.

You should be able to repro this error locally by running our CIBuild script from a Windows machine: `cibuild.cmd --scope Build`

maddin2016:
@jeffkl, at which point you define such attributes?

maddin2016:
I would create a class in `Microsoft.Build.Task` something like `DisableSameDestinationCheckAttribute` and then check for that attribute. Would that be a solution?

maddin2016:
Or `ExcludeFromSameDestinationCheckAttribute`

maddin2016:
I see that's not a good solution . What about if you set a new Metadata for what i can check?

maddin2016:
Hi @rainersigwald, are there any Updates how we can fix the failing Tests?

jeffkl:
@maddin2016 I would recommend you add a new property to the Copy task like "AllowDuplicates" that defaults to true.  Then condition your check on this property.  Then users can opt in by setting that in their targets.  You could also add an MSBuild property `$(CopyAllowDuplicates)` and in our code base we could set it to `true`.

maddin2016:
@jeffkl. sounds good 👍

maddin2016:
@jeffkl, i have think about your solution. Is it not better to set default to false? If we set true many people forgot that there is a switch to disable and we have the same issue. I think its better to throw an error and show a note for example `If you want to copy into same destination, set "AllowDuplicates" to true`

jeffkl:
@maddin2016 my only concern is that people will upgrade to MSBuild 15 either willingly or by some chance and not understand why their build starts to fail.

@rainersigwald what do you think if we have this new AllowDuplicates property to have 3 possible values:

1. (Default) `warn` - Copy task emits a warning if a duplicate is detected
2. `false` - Copy task emits an error if a duplicate is detected
3. `true` - Copy task emits a low importance message if a duplicate is detected

Do we have another task/property that controls if a condition is a warning, an error, or a message that we can base this off of?

maddin2016:
This is a good objection. But my opinion is better to show what someone has done false instead to risk that they overread the warning and spend time to search what's wrong. It's really hard to decide

piotrpMSFT:
This change should affect CopyToOutput, CopyToPublish, etc. I think the Copy method is used in all of these places but want to be sure folks are thinking about that in this context :)

maddin2016:
Hi @rainersigwald, @jeffkl , long time not hearing from this PR. Is there any update on this topic?

AndyGerlicher:
Cleaning up old PRs. Closing this since there hasn't been movement. We would still consider taking the PR with the requested changes.

PR #1044, opened 2016-09-19
Change base path for CompillerToolsDir to NuGet packages folder.

mfilippov:
By default NuGet restore packages from .nuget/project.json to
HomeDirectory/.nuget/packages but CompilerToolsDir expect compiler package in
solution packages folder. I change it behaviour to find package in right
place. It's fix unresolved reference to Microsoft.Build.Tasks.CodeAnalysis
in projects.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -327,7 +327,9 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(NetCoreBuild)' != 'true'">
-    <CompilerToolsDir>$([System.IO.Path]::Combine($(ToolPackagesDir)Microsoft.Net.Compilers, $(CompilerToolsVersion), "tools"))$([System.IO.Path]::DirectorySeparatorChar)</CompilerToolsDir>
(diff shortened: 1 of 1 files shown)


rainersigwald:
Under what circumstances do you see that unresolved reference? That package should be [explicitly restored to the repo-local location early in the build](https://github.com/Microsoft/msbuild/blob/xplat/dir.props#L89).

mfilippov:
@rainersigwald I run build.cmd and open solution in VS after that.

rainersigwald:
@mfilippov Does it still happen if you do `cibuild.cmd --bootstrap-only`?

mfilippov:
I write Demo target in Microsoft.Build.csproj

mfilippov:
If I use my patch solution opened correctly on Windows in VS and on Mac OS X in Rider.

mfilippov:
If I use reference is correct ```cibuild.cmd --bootstrap-only```

PR #1058, opened 2016-09-21
Update resource fallback behavior

tmeschter:
Currently localized builds depend on resources for all cultures, including en.
This means that if you don't have satellite assemblies for en, you won't be
able to load resources at all.

We want to keep this behavior for localized *debug* builds since it will help
us flush out problems related to localization and how we layout our satellite
assemblies on disk. For *release* builds, however, we want en to fall back to
resources in the main assembly. This allows the MSBuild binaries to be used
even without satellite assemblies installed.


tmeschter:
@cdmihai @AndyGerlicher @rainersigwald @jeffkl @Sarabeth-Jaffe-Microsoft Please review.

PR #1059, opened 2016-09-21
Port wildcard import of project extensions from xplat

### LINKED ISSUE #784: Mechanism for "magic" pickup of nuget restore logic during build

rainersigwald:
These extensions will generally come from package management systems by writing to a file named:

  $(MSBuildProjectExtensionsPath)\$(MSBuildProjectFile).<SomethingUnique>.props

Each package management system should use a unique moniker to avoid collisions.  It is a wild-card import so the order of the import is based on the file system's ordering.

Fixes #784

- Included unit tests
- Fix tests that broke when BaseIntermedateOutputPath was moved.

Current import order is:

1. directory build props
2. project extension props
3. project extension targets
4. directory build targets

PORT TO master:

 Conflicts:
	src/XMakeTasks/Microsoft.Common.props
	src/XMakeTasks/UnitTests/Microsoft.Build.Tasks.UnitTests.csproj

### DIFF
--- a/src/XMakeBuildEngine/UnitTests/EscapingInProjects_Tests.cs
+++ b/src/XMakeBuildEngine/UnitTests/EscapingInProjects_Tests.cs
@@ -899,6 +899,7 @@ namespace Microsoft.Build.UnitTests.EscapingInProjects_Tests
             // ---------------------
             ObjectModelHelpers.CreateFileInTempProjectDirectory("foo.csproj", @"
                 <Project DefaultTargets=`Build` ToolsVersion=`msbuilddefaulttoolsversion` xmlns=`http://schemas.microsoft.com/developer/msbuild/2003`>
+                    <Import Project=`$(MSBuildBinPath)\Microsoft.Common.props` />
                     <PropertyGroup>
                         <Configuration Condition=` '$(Configuration)' == '' `>Debug</Configuration>
                         <Platform Condition=` '$(Platform)' == '' `>AnyCPU</Platform>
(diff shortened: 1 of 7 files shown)


nguerrera:
👍

rainersigwald:
@dotnet-bot Windows_NT Build for Desktop please

(race in XSLT test that we've seen acouple of times before)

rainersigwald:
Pushing since we want this fix ASAP, and the one failure on Desktop was a race and didn't repro on the (on this branch lying) "CoreCLR" build which is identical.

PR #1060, opened 2016-09-21
Only record extra Evaluator data when needed

cdmihai:
This saves memory. Otherwise, the extra list tips over the memory sensitive GC aware code from the ProjectCollection cache (and potentially other special MSBuild caches) leading to a lot of GC calls and thus performance hits.

### DIFF
--- a/src/XMakeBuildEngine/Definition/Project.cs
+++ b/src/XMakeBuildEngine/Definition/Project.cs
@@ -61,7 +61,12 @@ namespace Microsoft.Build.Evaluation
         /// <summary>
         /// Throw an exception and stop the evaluation of a project if any circular imports are detected
         /// </summary>
-        RejectCircularImports = 4
+        RejectCircularImports = 4,
+
+        /// <summary>
+        /// Record the item elements that got evaluated
+        /// </summary>
+        RecordEvaluatedItemElements = 8
     }
 
     /// <summary>
@@ -1070,6 +1075,7 @@ namespace Microsoft.Build.Evaluation
         /// </returns>
         public List<GlobResult> GetAllGlobs()
         {
+            ReevaluateForPostMortemAnalysisIfNecessary();
             return GetAllGlobs(_data.EvaluatedItemElements);
         }
 
@@ -1083,6 +1089,8 @@ namespace Microsoft.Build.Evaluation
             {
                 return new List<GlobResult>();
             }
+
+            ReevaluateForPostMortemAnalysisIfNecessary();
             return GetAllGlobs(GetItemElementsByType(_data.EvaluatedItemElements, itemType));
         }
 
@@ -1157,6 +1165,7 @@ namespace Microsoft.Build.Evaluation
         /// </returns>
         public List<ProvenanceResult> GetItemProvenance(string itemToMatch)
         {
+            ReevaluateForPostMortemAnalysisIfNecessary();
             return GetItemProvenance(itemToMatch, _data.EvaluatedItemElements);
         }
 
@@ -1167,6 +1176,7 @@ namespace Microsoft.Build.Evaluation
         /// <param name="itemType">The item type to constrain the search in</param>
         public List<ProvenanceResult> GetItemProvenance(string itemToMatch, string itemType)
         {
+            ReevaluateForPostMortemAnalysisIfNecessary();
             return GetItemProvenance(itemToMatch, GetItemElementsByType(_data.EvaluatedItemElements, itemType));
         }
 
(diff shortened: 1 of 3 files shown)


cdmihai:
@dsplaisted @AndyGerlicher

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -61,7 +61,12 @@ public enum ProjectLoadSettings
         /// <summary>
         /// Throw an exception and stop the evaluation of a project if any circular imports are detected
         /// </summary>
-        RejectCircularImports = 4
+        RejectCircularImports = 4,
+
+        /// <summary>
+        /// Throw an exception and stop the evaluation of a project if any circular imports are detected
```
rainersigwald: Copy pasta from above?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1185,11 +1195,30 @@ public List<ProvenanceResult> GetItemProvenance(ProjectItem item)
...
 
+            ReevaluateForPostMortemAnalysisIfNecessary();
             var itemElementsAbove = GetItemElementsThatMightAffectItem(_data.EvaluatedItemElements, item);
 
             return GetItemProvenance(item.EvaluatedInclude, itemElementsAbove);
         }
 
+        /// <summary>
+        /// Some project APIs need to do analysis that requires the Evaluator to record more data than usual as it evaluates.
+        /// This method checks if the Evaluator was ran with the extra required settings and if not, does a re-evaluation.
```
rainersigwald: nit: `s/ran/run/`

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2536,6 +2536,37 @@ public void GetItemProvenanceOnlyGlob()
...
+
+            var expected = new ProvenanceResultTupleList
+            {
+                Tuple.Create("A", Operation.Include, Provenance.Glob, 1),
+                Tuple.Create("B", Operation.Exclude, Provenance.Glob, 1)
+            };
+
+            var project = ObjectModelHelpers.CreateInMemoryProject(projectContents);
+
+            var provenaceResult = project.GetItemProvenance("2.foo");
```
rainersigwald: nit: `s/provenace/provenance/`

dsplaisted:
If possible, I'd prefer it if the getter for `EvaluatedItemElements` would call `ReevaluateForPostMortemAnalysisIfNecessary` rather than everyone that accesses it knowing they need to call this method.  That would make it less error prone.

Otherwise, :shipit:

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -2536,6 +2536,37 @@ public void GetItemProvenanceOnlyGlob()
...
+                </Project>
+            ";
+
+            var expected = new ProvenanceResultTupleList
+            {
+                Tuple.Create("A", Operation.Include, Provenance.Glob, 1),
+                Tuple.Create("B", Operation.Exclude, Provenance.Glob, 1)
+            };
+
+            var project = ObjectModelHelpers.CreateInMemoryProject(projectContents);
```
rainersigwald: Does this implicitly evaluate without the new flag? If so, please comment it :)

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -1167,6 +1176,7 @@ public List<ProvenanceResult> GetItemProvenance(string itemToMatch)
         /// <param name="itemType">The item type to constrain the search in</param>
         public List<ProvenanceResult> GetItemProvenance(string itemToMatch, string itemType)
         {
+            ReevaluateForPostMortemAnalysisIfNecessary();
```
rainersigwald: I don't love having to do this manually in all the right places, but I can't think of a better way at the moment.

cdmihai:
@dsplaisted I looked into doing that but it hits ugly re-entrancy issues because the Evaluator is also calling that getter while evaluating. Making it work nicely complicates that part of the code quite a bit, so I don't think it's worth doing.

dsplaisted:
If possible, I'd prefer it if the getter for `EvaluatedItemElements` would call `ReevaluateForPostMortemAnalysisIfNecessary` rather than everyone that accesses it knowing they need to call this method. That would make it less error prone.

Otherwise, :shipit:

Issue #1039, opened 2016-09-16
Create a NuGet package containing the MSBuild runtime


jeffkl:
We need a NuGet package that customers can reference to load projects and evaluate properties/items or do in-proc builds.

## Proposal
Create a new package containing the `.props`, `.targets`, and `.tasks` files needed for evaluating projects.  This will supersede the `Microsoft.Build.Targets' and `MSBuild` packages which we will deprecate.

The new package, `Microsoft.Build.Runtime` will have the following layout:
```
contentFiles
└─ any
   ├─ any
   |  ├─ 15.0
   |  |  └ Microsoft.Common.props
   |  │  *.props
   |  │  *.targets
   |  └  *.tasks
   ├─ net46
   |  │  MSBuild.exe
   |  └  MSBuild.exe.config
   └─ netcoreapp1.0
      └  MSBuild.exe
```

This will place what we consider to be the core MSBuild runtime into the end user's output directory.  

### .NET Core via `dotnet run`
In this scenario, the CLR loads assemblies from their NuGet package location.  This means that our BuildEnvironmentHelper will need to be updated to look in `AppContext.BaseDirectory` instead of the directory containing `Microsoft.Build.dll`.

### .NET Core via `dotnet publish`
This command places all dependencies in a folder named 'publish' and when running an application from this folder, the assemblies are loaded from that directory.  The `AppContext.BaseDirectory` is still the correct place to use as MSBuild directory.

### .NET Framework via building against net46
NuGet copies all of the referenced assemblies and content files to the output directory.  The `BuildEnvironmentHelper` can remain the same for this scenario and use the directory containing `Microsoft.Build.dll` as the MSBuild directory.

### Visual Studio
The `BuildEnvironmentHelper` already looks for the MSBuild directory based on the process name being `devenv.exe` or the `VSINSTALLDIR` environment variable.  This means that extensions within Visual Studio or the Visual Studio Developer Command Prompt will use the MSBuild that ships with Visual Studio and won't need to redistribute the MSBuild runtime.

### In-proc builds and project evaluation
Not everything will work when building or evaluating projects.  This is because some build logic comes from various teams.  This includes building Azure cloud projects, C++ projects, FSharp, etc.  These teams will need to provide a package that users can reference along with the `Microsoft.Build.Runtime` package that will give them all of the required functionality to evaluate or build their projects.

## Tasks
[...]

PR #1069, opened 2016-09-22
Update BuildEnvironmentHelper to locate .NET Core app directory

jeffkl:
In .NET Core, the AppContext.BaseDirectory can be used to locate the application's dependencies.

Related to #1039

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -43,6 +43,7 @@ namespace Microsoft.Build.Shared
             var processNameCurrentProcess = s_getProcessFromRunningProcess();
             var executingAssembly = s_getExecutingAssmblyPath();
             var currentDirectory = s_getCurrentDirectory();
+            var appContextBaseDirectory = s_getAppContextBaseDirectory();
 
 
             // Check if our current process name is in the list of own test runners
@@ -83,7 +84,7 @@ namespace Microsoft.Build.Shared
                 // Try the base directory that the assembly resolver uses to probe for assemblies.
                 // Under certain scenarios the assemblies are loaded from spurious locations like the NuGet package cache
                 // but the toolset files are copied to the app's directory via "contentFiles".
-                () => TryFromFolder(AppContext.BaseDirectory, runningTests, runningInVisualStudio, visualStudioPath)
+                () => TryFromFolder(appContextBaseDirectory, runningTests, runningInVisualStudio, visualStudioPath)
 #endif
             };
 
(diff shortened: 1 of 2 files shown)


rainersigwald:
Can you elaborate on why this helps? I don't think I fully understand the scenario.

jeffkl:
When you execute `dotnet run`, the .NET Core runtime loads references from NuGet package locations.  This means `Microsoft.Build.Framework.dll` is loaded from one place, while `Microsoft.Build.dll` is loaded from another.

However, when you do `dotnet publish`, all of the assemblies are placed in the output folder and loaded form it.  In these two cases, `AppContext.BaseDirectory` is the location of the executing app which would reference `Microsoft.Build.Runtime` package and get the `.props` and `.targets` in their output directory.  So this change makes it work that your app's directory is used as the location of MSBuild.  Does that make sense?

rainersigwald:
Ahh, yes it does make sense! So with the new package, the toolset files will be in `AppContext.BaseDirectory` _even though the DLLs are loaded from elsewhere_.

Can you update the comment for this case to be explicit about that scenario and why loading from that directory is right?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -77,7 +77,12 @@ private static BuildEnvironment Initialize()
                 ()=> TryFromFolder(msbuildFromVisualStudioRoot, runningTests, runningInVisualStudio, visualStudioPath),
 
                 // Try from the current directory
-                () => TryFromFolder(currentDirectory, runningTests, runningInVisualStudio, visualStudioPath)
+                () => TryFromFolder(currentDirectory, runningTests, runningInVisualStudio, visualStudioPath),
+
+#if !CLR2COMPATIBILITY // Assemblies compiled against anything older than .NET 4.0 won't have a System.AppContext
+                // Try the base directory that the assembly resolver uses to probe for assemblies.
+                () => TryFromFolder(AppContext.BaseDirectory, runningTests, runningInVisualStudio, visualStudioPath)
```
rainersigwald: Now that I see what's up, I think this should be above the current-working-directory path. If someone's redistributing an MSBuild runtime, that should be preferred over something that happens to be in the CWD, just as it would be if we found it next to our MSBuild.exe.

jeffkl:
Yeah but ***only*** if you can help me understand why the test fails... Is it because my change needs to be mocked to not return a valid location?

src/Shared/BuildEnvironmentHelper.cs
jeffkl: I went back and forth on this too.  Technically if I see MSBuild.exe (and the runtime) in the current directory, I sort of expect it to be used.  That would make sense for an end user but not necessarily for the app developer.  It seems rare that someone would change the current directory to some place that has MSBuild and then run a completely different app...

jeffkl:
I have updated the comment, let me know about the test fix when you have a chance.

rainersigwald:
I think I get the test failure. Inside the test run I have `AppContext.BaseDirectory` set to `C:\src\msbuild\bin\Debug-NetCore\Windows_NT_Deployment_Test\`, which is accurate--that's where it's running from. But that folder also contains a full toolset--which your new fallback will find. Then a toolset will be found and the test for "don't find toolsets in condition X" will fail.

You should be able to fix this by adding a new delegate like the ones [that are already there](https://github.com/rainersigwald/msbuild/blob/670ad6579b549697994696d4c84ab584fa43b0d3/src/Shared/BuildEnvironmentHelper.cs#L273-L274) to return `AppContext.BaseDirectory` . . . unless you've overridden it to return `null` in the test, which you should also do.

src/XMakeBuildEngine/UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -105,7 +105,7 @@ public void BuildEnvironmentDetectsVisualStudioByEnvironment()
             {
                 Environment.SetEnvironmentVariable("VSINSTALLDIR", env.TempFolderRoot);
                 Environment.SetEnvironmentVariable("VisualStudioVersion", "15.0");
-                BuildEnvironmentHelper.ResetInstance_ForUnitTestsOnly();
+                BuildEnvironmentHelper.ResetInstance_ForUnitTestsOnly(getAppContextBaseDirectory: ReturnNull);
```
rainersigwald: Do you need to specify this? Seems like it should be redundant (since it's after the env-var fallback).

jeffkl: I was worried if the order ever changed then the test could start to fail so I put this in here to remove the AppContext.BaseDirectory from the equation for all cases.  I can remove it if you'd like

rainersigwald: I think I'd rather the tests fail in that case.

jeffkl:

src/XMakeBuildEngine/UnitTests/BuildEnvironmentHelper_Tests.cs
jeffkl: Okay I'll update the test

jeffkl:
@AndyGerlicher @rainersigwald do you guys approve?

PR #1073, opened 2016-09-23
Updating folder xaml to enable folder properties in CPS based projects

adrianvmsft:
Updating folder xaml to enable folder properties in CPS based projects by using a specialized data source that can handle implicit folder includes besides explicit folder includes.

This change should not be integrated in Preview 5 as it requires the new data source implementation to be present (otherwise CPS based projects will fail to load).

### DIFF
--- a/src/XMakeTasks/XamlRules/Folder.xaml
+++ b/src/XMakeTasks/XamlRules/Folder.xaml
@@ -6,11 +6,11 @@
   PageTemplate="generic"
   Description="Folder Properties"
   xmlns="http://schemas.microsoft.com/build/2009/properties">
-    <Rule.DataSource>
-        <DataSource Persistence="ProjectFileFolderItems" HasConfigurationCondition="False" ItemType="Folder" />
-    </Rule.DataSource>
+  <Rule.DataSource>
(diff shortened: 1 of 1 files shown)


rainersigwald:
Where are the other changes in VS to support this, and how does it break if they're not there? Unfortunately we're currently being pulled into a few different branches so it seems likely we'll see a mismatch.

adrianvmsft:
The rest of the changes are in CPS. This only affects CPS based projects.
- if CPS changes are in, but not the MSBuild changes - properties on folders will not be available in the Solution Explorer
- if CPS changes are not in, but the MSBuild changes are - cps based projects consuming the updated rule won't load

rainersigwald:
@adrianvmsft I just noticed that this file isn't in [our VSIX definition](https://github.com/Microsoft/msbuild/blob/master/setup/files.swr). It may be getting into VS from our legacy location in the VS repo, delivered by some other VSIX. Should it be part of ours?

adrianvmsft:
I really don't know how it gets installed as I am not familiar with the MSBuild installation infrastructure.

adrianvmsft:
This change will be made in the internal Visual Studio repo.

PR #1075, opened 2016-09-23
Update projects to target NETStandard versions correctly

### LINKED ISSUE #682: Latest Preview Package does not actually target NETStandard 1.3

jeffkl:
* Redo all `project.json` so that they correctly target NETStandard 1.3 or 1.5.  Remove any unneeded stuff and move any full framework references into the corresponding `.csproj`
* Default NETStandard version when building .NET Core is 1.3
* Upgrade `Microsoft.Build.dll` and `MSBuild.exe` to NETStandard 1.5 so they can use the `AssemblyLoadContext` class
* Fix any code which referenced namespaces that aren't actually available
* Deleted class `ParameterModifier` because it is available in NETStandard1.5
* Remove ref assemblies from NuGet packages since they aren't needed.  I'll address the removal of the ref assembly source code in another change.
* Fix logic around NuSpec.ReferenceGenerator so that `.nuspec` files contain the full set of dependencies

I still need to get feedback from some teams as to whether or not this will break them.

Fixes #682 Latest Preview Package does not actually target NETStandard 1.3


dir.props
```diff
@@ -229,7 +220,17 @@
     <WindowsAppContainer>false</WindowsAppContainer>
 
     <DefineConstants>$(DefineConstants);RUNTIME_TYPE_NETCORE</DefineConstants>
+    
+    <!-- target framework for msbuild nuget packages. Used by Nuspec.ReferenceGenerator.targets to update corefx dependencies-->
+    <NuSpecTfm>netstandard$(TargetFrameworkVersion.Substring(1))</NuSpecTfm>
```
jeffkl: I don't love this but we store the target framework version as something like "v1.3" but I want to glue it onto a moniker like "netstandard1.3".  So I strip off the first character.  Maybe this should be TrimStart("v") ?

src/Framework/Microsoft.Build.Framework.csproj
```diff
@@ -8,6 +8,7 @@
     <RootNamespace>Microsoft.Build.Framework</RootNamespace>
     <AssemblyName>Microsoft.Build.Framework</AssemblyName>
     <EnableDocumentationFile>true</EnableDocumentationFile>
+    <UpdateNuGetPackageDependencies>true</UpdateNuGetPackageDependencies>
```
jeffkl: What does everyone think about this property?  It enables the NuSpec Reference Generator for a particular assembly, does the name make sense?  Should I not introduce this property and assume if a corresponding `.nuspec` exists then the functionality should be enabled?

src/dir.targets
```diff
@@ -43,9 +43,21 @@
     <!-- Targets from buildtools that restore packages before building. -->
   <!-- TODO: ideally replaced by a standard "dotnet restore" over the solution. -->
   <Import Condition="Exists('$(ToolsDir)\packageresolve.targets')" Project="$(ToolsDir)\packageresolve.targets" />
+
+  <ItemGroup Condition="'$(NetCoreBuild)' == 'true' and '$(UpdateNuGetPackageDependencies)' == 'true'">
+    <NuSpecFile Include="$(ProjectDir)build\NuGetPackages\$(MSBuildProjectName).nuspec" Condition="Exists('$(ProjectDir)build\NuGetPackages\$(MSBuildProjectName).nuspec')">
+      <Visible>false</Visible>
+    </NuSpecFile>
+    <NuSpecFile Include="$(ProjectDir)build\NuGetPackages\$(MSBuildProjectName).Core.nuspec" Condition="Exists('$(ProjectDir)build\NuGetPackages\$(MSBuildProjectName).Core.nuspec')">
```
jeffkl: This sucks but our `.csproj`s don't contain ".Core" like `Microsoft.Build.Tasks.Core` so I had to add a fallback here 😢 

Solutions:
1. Rename `Microsoft.Build.Tasks.csproj` to `Microsoft.Build.Tasks.Core.csproj`
2. Rename `Microsoft.Build.Tasks.Core.nuspec` to `Microsoft.Build.Tasks.nuspec`
3. Leave this hack...

jeffkl:

build/NuGetPackages/Microsoft.Build.Framework.nuspec
```diff
@@ -26,7 +30,5 @@
     <file src="$configuration$\Output\$id$.xml" target="lib\net46\$id$.xml" />
     <file src="$configuration$-NetCore\Output\$id$.dll" target="lib\netstandard1.3\$id$.dll" />
     <file src="$configuration$-NetCore\Output\$id$.xml" target="lib\netstandard1.3\$id$.xml" />
-    <file src="$configuration$\Output\ref\$id$.dll" target="ref\net46\$id$.dll" />
-    <file src="$configuration$-NetCore\Output\ref\$id$.dll" target="ref\netstandard1.3\$id$.dll" />
```
rainersigwald: Don't we still want a ref assembly for 1.3 so that a consumer can be confident that they're targeting the intersection of the API surfaces?

dir.props
rainersigwald: I'd say `TrimStart('v')` is a bit clearer. Agree that it's not optimal.

src/Framework/Microsoft.Build.Framework.csproj
rainersigwald: I'd say if it exists, update it.

src/dir.targets
rainersigwald: I say leave the hack, rename post-Great-Merge.

build/NuGetPackages/Microsoft.Build.Framework.nuspec
jeffkl: The lib\netstandard1.3 will remain which should give consumer's the same functionality.  We added ref's originally because our libs weren't actually .NETStandard1.3 compatible.  But with this change, they are.

rainersigwald: Ohh, my confusion. I was thinking of building targeting common functionality but for desktop. But if you're trying to build a portable library you should target netstandard yourself, so this is fine. You're right, thanks.

src/Framework/Microsoft.Build.Framework.csproj
jeffkl: I tried to do this but `<Import />` doesn't let me use an item group in a condition.  I'll leave this for now but will probably remove it in the near future.

PR #1077, opened 2016-09-23
Update dotnet CLI and build tools

jeffkl:
Dotnet CLI to 1.0.0-preview2-003131
Build tools to latest 1.0.26-prerelease-00822-03


cdmihai:
@dotnet-bot test this please

jeffkl:
For some reason, this is failing to restore on Ubuntu.  I haven't been able to figure out why on my VM.

jeffkl:
I ended up copying the dotnet CLI version from corefx and then updating init-tools from there as well.  Not sure how stable this version is though.

@eerhardt might know if this version is safe...

jeffkl:
It's ironic that it works on Linux now but fails on OS X.  @rainersigwald could you please try this out when you get a chance?  I can't see the init-tools.log in our CI build right?

eerhardt:
That CLI version is just as good as any pre-release version.  It is newer than the Preview2 we shipped in July.  But it isn't the absolute latest from our "master" branch.  Since CoreFX is using it, it should be fine for you.

jeffkl:
Thanks Eric.

PR #1086, opened 2016-09-24
Pull in DesignTime.targets in the crosstargeting build

srivatsn:
When VS is installed it lays down a DesignTime.targets which tells the project system about the project (capabilities, msbuild rules etc.). This is pulled in by the common targets and so is available in the inner build. However it is also needed in the outer build to open a cross-targeting project in VS. When a project is opened, the CPS project system needs information like the project's capabilities, rules etc to initialize itself and without the design time targets that blows up. 

This change simply adds C# and VB cross-targeting targets and imports the designtime targets (if they exist).

@nguerrera @rainersigwald @cdmihai @AndyGerlicher

### DIFF
--- a/setup/files.swr
+++ b/setup/files.swr
@@ -26,12 +26,14 @@ folder InstallDir:\MSBuild\15.0\Bin
   file source=$(X86BinPath)Microsoft.Common.targets
   file source=$(X86BinPath)Microsoft.Common.tasks
   file source=$(X86BinPath)Microsoft.CSharp.CurrentVersion.targets
+  file source=$(X86BinPath)Microsoft.CSharp.CrossTargeting.targets
   file source=$(X86BinPath)Microsoft.CSharp.targets
   file source=$(X86BinPath)Microsoft.NetFramework.CurrentVersion.props
   file source=$(X86BinPath)Microsoft.NetFramework.CurrentVersion.targets
   file source=$(X86BinPath)Microsoft.NetFramework.props
   file source=$(X86BinPath)Microsoft.NetFramework.targets
   file source=$(X86BinPath)Microsoft.VisualBasic.CurrentVersion.targets
+  file source=$(X86BinPath)Microsoft.VisualBasic.CrossTargeting.targets
   file source=$(X86BinPath)Microsoft.VisualBasic.targets
   file source=$(X86BinPath)MSBuild.rsp
   file source=$(X86BinPath)Workflow.targets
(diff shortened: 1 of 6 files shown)


nguerrera:
So the thing we've been saying all along, that VS will deal with the configured project's targets only, is not true?

Are the design time targets going to do sensible things in the outer build or should they be s subbed appropriately?

Are we fixing things at the right layer? The old project system can load these projects ok. Certain things don't work, but it's perfectly graceful. Could that be considered a regression? Any project with a typo in its import would exhibit what you're fixing and this won't help.

nguerrera:
I should have asked if *all* of the design time targets are going to do sensible things in the outer context. IOW, do we need to split them? I'm sure targets that just retrieve items are ok, though they'll only see items that are not TFM specific. 

Does VS pick configuration and platform when invoking these targets?

nguerrera:
I had a look at the actual design time targets files. Most are indeed fine, except the thing that gets a CSC command line. That's not a reasonable question if no TFM is provided. I assume that wit it won't actually get invoked on unconfigured project, correct? 

I would like to split that out in to a file that only gets imported if '$(IsCrossTargetingBuild)' is false. And anything that is TFM specific in design-time targets should go to the split file. We can keep this single variable and just do the dispatch in the design-time targets themselves. 

So, there'd be nothing to change here, so LGTM apart from the missing swr update. 😀

srivatsn:
Good catch on the setupauthoring. Added that.

Yes the design time targets are fairly simple and has just the one target. I do agree that we need to start a conversation with CPS about the state of the project when there are msbuild errors.

setup/files.swr
```diff
@@ -26,12 +26,14 @@ folder InstallDir:\MSBuild\15.0\Bin
...
   file source=$(X86BinPath)Microsoft.Common.tasks
   file source=$(X86BinPath)Microsoft.CSharp.CurrentVersion.targets
+  file source=$(X86BinPath)Microsoft.CSharp.CrossTargeting.targets
   file source=$(X86BinPath)Microsoft.CSharp.targets
   file source=$(X86BinPath)Microsoft.NetFramework.CurrentVersion.props
   file source=$(X86BinPath)Microsoft.NetFramework.CurrentVersion.targets
   file source=$(X86BinPath)Microsoft.NetFramework.props
   file source=$(X86BinPath)Microsoft.NetFramework.targets
   file source=$(X86BinPath)Microsoft.VisualBasic.CurrentVersion.targets
+  file source=$(X86BinPath)Microsoft.VisualBasic.CrossTargeting.targets
```
nguerrera: IIRC every targets file should appear twice in this file. One for root and one for amd64 subfolder.

AndyGerlicher:
LGTM once you update the `files.swr`. The CI should probably pass on retry.

PR #1088, opened 2016-09-25
Change the MSBuildProjectExtension mechanism to look for projectname …

srivatsn:
…instead of projectfile

Using the projectfile means the file it imports is obj\foo.csproj.nuget.g.targets. It seems better to import obj\foo.nuget.g.targets (which is currently what nuget generates)


srivatsn:
@jeffkl @AndyGerlicher @rainersigwald @cdmihai

srivatsn:
We've decided to go with a NuGet fix instead.

PR #1089, opened 2016-09-26
Support to remove unnecessary XML in project file

### LINKED ISSUE #699: Remove boilerplate XML
**This issue is in the open design phase and is a part of the [project.json -> csproj](https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/) conversion effort.**

### Terser MSBuild syntax - Make the syntax less verbose (more compressed XML). 

**Can we remove:**
- [ ] XML tag (assume UTF8 unless specified)
- [ ] XML Namespace 

**Open Issues**
- There may be impact to XML parsing/editing tools

Note: there is much improvement to be done in the default template for MSBuild projects that does not involve Core MSBuild changes.

See: [roslyn-project-system feature](https://github.com/dotnet/roslyn-project-system/issues?q=is%3Aissue+is%3Aopen+label%3A%22Feature+-+Project+File+Cleanup%22)

AndyGerlicher:
This change ended up being somewhat more involved than I had originally hoped, and I wish I would have made the changes atomically in separate commits but I didn't have time to go back and do that after the fact. Most of the complexity revolves around mismatch between what the intent of the code appeared to be and what it actually did. The biggest example being `_writeXmlDeclaration` which was calculated but not used. To get the behavior I wanted (not writing the XML declaration when `UTF8`), this had to be fixed. But of course unit tests were expecting the old behavior so many helper methods had to be changed. We also have a fair bit of complexity because we derive from XmlDocument base classes so we can't use many of the static Create methods that were likely added later that add functionality to make a lot of this much easier (i.e. `XmlWriterSettings.OmitXmlDeclaration` or `XmlWriterSettings.Encoding`).

* Added support for NewProjectFileOptions to specify which features should
  appear in a new in-memory project file. Default will include all options
  as it did in previous versions.

* In ProjectWriter.cs, there seemed to be intent to only write an XML
  declaration in certain cases, but it was not implemented. Added support
  to omit the XML declaration when the encoding is not the default (UTF8)
  and loaded project did not declare it.

* Removed requirements that the namespace be present. Changed calls to
  XmlDocument to use the namespace loaded from the document, which can now
  be the default MSBuild namespace or empty.

* Changed the RawXml method in ProjectRootElement.cs to use UTF8 by
  default. This appeared to be the intent in both the code
  (s_defaultEncoding == UTF8) and in the default XML declaration tag.
  However, StringWriter does not support an encoding other than UTF16.
  This caused the XML declaration to always be written as UTF16. This
  had a somewhat cascading effect on tests that use this method and
  no longer get an XML declaration tag. However, I believe that this is
  closer to the original intent as the input did not specify an xml
  declaration nor any indication that UTF16 should be used.

Resolves #699

### DIFF
--- a/src/Shared/ProjectWriter.cs
+++ b/src/Shared/ProjectWriter.cs
@@ -112,9 +112,8 @@ namespace Microsoft.Build.Shared
             }
 
             // don't write an XML declaration unless the project already has one or has non-default encoding
-            _writeXmlDeclaration =
-                ((projectRootElementDeclaration != null) ||
-                ((_documentEncoding != Encoding.UTF8) && (_documentEncoding != null)));
+            _writeXmlDeclaration = (projectRootElementDeclaration != null) ||
+                                   (_documentEncoding != null && !Equals(_documentEncoding, Encoding.UTF8));
         }
 
         /// <summary>
@@ -166,12 +165,25 @@ namespace Microsoft.Build.Shared
             }
         }
 
+        /// <summary>
+        /// Override method in order to omit the xml declaration tag in certain cases. The tag will be written if:
+        ///  - The tag was present in the file/stream loaded.
+        ///  - The Encoding is specified and not default (UTF8)
+        /// </summary>
+        public override void WriteStartDocument()
+        {
+            if (_writeXmlDeclaration)
+            {
+                base.WriteStartDocument();
+            }
+        }
+
         #endregion
 
         // indicates whether an XML declaration e.g. <?xml version="1.0"?> will be written at the start of the project
         private bool _writeXmlDeclaration;
 
         // encoding of the document, if specified when constructing
-        private Encoding _documentEncoding = null;
+        private readonly Encoding _documentEncoding;
     }
 }
--- a/src/XMakeBuildEngine/Construction/Solution/ProjectInSolution.cs
+++ b/src/XMakeBuildEngine/Construction/Solution/ProjectInSolution.cs
@@ -305,11 +305,8 @@ namespace Microsoft.Build.Construction
 
                 if (mainProjectElement != null && mainProjectElement.LocalName == "Project")
                 {
-                    if (String.Compare(mainProjectElement.NamespaceURI, XMakeAttributes.defaultXmlNamespace, StringComparison.OrdinalIgnoreCase) == 0)
-                    {
-                        _canBeMSBuildProjectFile = true;
-                        return _canBeMSBuildProjectFile;
-                    }
+                    _canBeMSBuildProjectFile = true;
+                    return _canBeMSBuildProjectFile;
                 }
             }
             // catch all sorts of exceptions - if we encounter any problems here, we just assume the project file is not
--- a/src/XMakeBuildEngine/Evaluation/ProjectParser.cs
+++ b/src/XMakeBuildEngine/Evaluation/ProjectParser.cs
@@ -162,9 +162,7 @@ namespace Microsoft.Build.Construction
             ProjectErrorUtilities.VerifyThrowInvalidProject(element.LocalName == XMakeElements.project, element.Location, "UnrecognizedElement", element.Name);
 
             // If a namespace was specified it must be the default MSBuild namespace.
-            if (element.Prefix.Length > 0 ||
-                (!string.Equals(element.NamespaceURI, XMakeAttributes.defaultXmlNamespace,
-                     StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(element.NamespaceURI)))
+            if (!ProjectXmlUtilities.VerifyValidProjectNamespace(element))
             {
                 ProjectErrorUtilities.ThrowInvalidProject(element.Location, "ProjectMustBeInMSBuildXmlNamespace",
                     XMakeAttributes.defaultXmlNamespace);
--- a/src/XMakeBuildEngine/Evaluation/ProjectParser.cs
+++ b/src/XMakeBuildEngine/Evaluation/ProjectParser.cs
@@ -161,9 +161,17 @@ namespace Microsoft.Build.Construction
             ProjectErrorUtilities.VerifyThrowInvalidProject(element.Name != XMakeElements.visualStudioProject, element.Location, "ProjectUpgradeNeeded", _project.FullPath);
             ProjectErrorUtilities.VerifyThrowInvalidProject(element.LocalName == XMakeElements.project, element.Location, "UnrecognizedElement", element.Name);
 
-            if (element.Prefix.Length > 0 || !String.Equals(element.NamespaceURI, XMakeAttributes.defaultXmlNamespace, StringComparison.OrdinalIgnoreCase))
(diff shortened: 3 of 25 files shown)


src/XMakeBuildEngine/Construction/Solution/ProjectInSolution.cs
```diff
@@ -305,11 +305,8 @@ internal bool CanBeMSBuildProjectFile(out string errorMessage)
 
                 if (mainProjectElement != null && mainProjectElement.LocalName == "Project")
                 {
-                    if (String.Compare(mainProjectElement.NamespaceURI, XMakeAttributes.defaultXmlNamespace, StringComparison.OrdinalIgnoreCase) == 0)
-                    {
-                        _canBeMSBuildProjectFile = true;
-                        return _canBeMSBuildProjectFile;
-                    }
+                    _canBeMSBuildProjectFile = true;
```
rainersigwald: Should we remove this concept?

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -161,9 +161,17 @@ private void Parse()
...
             ProjectErrorUtilities.VerifyThrowInvalidProject(element.LocalName == XMakeElements.project, element.Location, "UnrecognizedElement", element.Name);
 
-            if (element.Prefix.Length > 0 || !String.Equals(element.NamespaceURI, XMakeAttributes.defaultXmlNamespace, StringComparison.OrdinalIgnoreCase))
+            // If a namespace was specified it must be the default MSBuild namespace.
+            if (element.Prefix.Length > 0 ||
+                (!string.Equals(element.NamespaceURI, XMakeAttributes.defaultXmlNamespace,
+                     StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(element.NamespaceURI)))
             {
-                ProjectErrorUtilities.ThrowInvalidProject(element.Location, "ProjectMustBeInMSBuildXmlNamespace", XMakeAttributes.defaultXmlNamespace);
+                ProjectErrorUtilities.ThrowInvalidProject(element.Location, "ProjectMustBeInMSBuildXmlNamespace",
```
rainersigwald: Update the string as well, to say "must be in this namespace or no namespace"?

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -161,9 +161,17 @@ private void Parse()
             ProjectErrorUtilities.VerifyThrowInvalidProject(element.Name != XMakeElements.visualStudioProject, element.Location, "ProjectUpgradeNeeded", _project.FullPath);
             ProjectErrorUtilities.VerifyThrowInvalidProject(element.LocalName == XMakeElements.project, element.Location, "UnrecognizedElement", element.Name);
 
-            if (element.Prefix.Length > 0 || !String.Equals(element.NamespaceURI, XMakeAttributes.defaultXmlNamespace, StringComparison.OrdinalIgnoreCase))
+            // If a namespace was specified it must be the default MSBuild namespace.
+            if (element.Prefix.Length > 0 ||
+                (!string.Equals(element.NamespaceURI, XMakeAttributes.defaultXmlNamespace,
+                     StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(element.NamespaceURI)))
```
rainersigwald: I think I'd prefer `!( X || Y )` here--logically equivalent but easier for me to understand.

src/XMakeBuildEngine/UnitTests/BackEnd/MSBuild_Tests.cs
```diff
@@ -1609,5 +1605,41 @@ public void TargetNameIsCaseInsensitive()
...
+                    </Target>	
+                </Project>
+                ");
+
+            try
+            {
+                Project project = new Project(projectFile2);
+                MockLogger logger = new MockLogger();
+
+                project.Build(logger);
```
rainersigwald: No assert-successful-build?

src/Shared/ProjectWriter.cs
```diff
@@ -166,12 +165,25 @@ public override void WriteString(string text)
...
+        /// <summary>
+        /// Override method in order to omit the xml declaration tag in certain cases. The tag will be written if:
+        ///  - The tag was present in the file/stream loaded.
+        ///  - The Encoding is specified and not default (UTF8)
+        /// </summary>
+        public override void WriteStartDocument()
+        {
+            if (_writeXmlDeclaration)
+            {
+                base.WriteStartDocument();
```
AndyGerlicher: It does not get stripped. In testing just now with VS, the BOM will be added when the encoding is UTF8. But it no longer (even without project system changes) adds the <xml> tag.

---
In reply to: [80529908](https://github.com/Microsoft/msbuild/pull/1089#discussion_r80529908) [](ancestors = 80529908)

src/XMakeBuildEngine/UnitTests/BackEnd/MSBuild_Tests.cs
AndyGerlicher: The next assert would catch an issue here, but good call. I added that on the test I copied this from as well :).

---
In reply to: [80769718](https://github.com/Microsoft/msbuild/pull/1089#discussion_r80769718) [](ancestors = 80769718)

src/XMakeBuildEngine/Construction/Solution/ProjectInSolution.cs
AndyGerlicher: I don't think so. This seems to be a way to "roughly" check if a file is an msbuild file. It's not as accurate as it was before, but at least checks to see if it has a <project> element at the root. I would say keep it. It's also used in quite a few places.

---
In reply to: [80768595](https://github.com/Microsoft/msbuild/pull/1089#discussion_r80768595) [](ancestors = 80768595)

rainersigwald: Ah, that works. I was thinking that you removed the only check--but it's still inside the if-<Project> one.

src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
```diff
@@ -94,14 +94,31 @@ internal static void ThrowProjectInvalidChildElement(string name, string parentN
...
+                    "CustomNamespaceNotAllowedOnThisChildElement", element.Name, element.ParentNode?.Name);
             }
         }
 
         /// <summary>
+        /// Verify that if a namespace is specified it matches the default MSBuild namespace.
+        /// </summary>
+        /// <param name="element">Element to check namespace.</param>
+        /// <returns>True when the namespace is in the MSBuild namespace or no namespace.</returns>
+        internal static bool VerifyValidProjectNamespace(XmlElementWithLocation element)
```
AndyGerlicher: I extracted this out to another method. Agreed it was not straightforward as it was, and it was also duplicated. This look better?

PR #1090, opened 2016-09-26
Project can throw when item elements are split

cdmihai:
Exception based implementation of https://github.com/Microsoft/msbuild/issues/736

### DIFF
--- a/src/XMakeBuildEngine/Definition/Project.cs
+++ b/src/XMakeBuildEngine/Definition/Project.cs
@@ -163,6 +163,23 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
+        /// Certain item operations split the item element in multiple elements if the include
+        /// contains globs, references to items or properties, or multiple item values.
+        ///
+        /// The items operations that may expand item elements are:
+        /// - <see cref="RemoveItem"/>
+        /// - <see cref="RemoveItems"/>
+        /// - <see cref="ProjectItem.ChangeItemType"/>
+        /// - <see cref="ProjectItem.Rename"/>
+        /// - <see cref="ProjectItem.RemoveMetadata"/>
+        /// - <see cref="ProjectItem.SetMetadataValue"/>
+        /// 
+        /// When this property is set to true, the previous item operations throw an <exception cref="InvalidOperationException"></exception>
+        /// instead of expanding the item element. 
+        /// </summary>
+        public bool ThrowInsteadOfSplittingItemElement { get; set; }
+
+        /// <summary>
         /// Construct an empty project, evaluating with the specified project collection's
         /// global properties and default tools version.
         /// Project will be added to the specified project collection when it is named.
@@ -2126,10 +2143,12 @@ namespace Microsoft.Build.Evaluation
         /// </remarks>
         internal bool SplitItemElementIfNecessary(ProjectItemElement itemElement)
         {
-            if (!FileMatcher.HasWildcardsSemicolonItemOrPropertyReferences(itemElement.Include))
+            if (!ItemElementRequiresSplitting(itemElement))
             {
                 return false;
             }
+
+            ErrorUtilities.VerifyThrowInvalidOperation(!ThrowInsteadOfSplittingItemElement, "OM_CannotSplitItemElementWhenSplittingIsDisabled", itemElement.Location, $"{nameof(Project)}.{nameof(ThrowInsteadOfSplittingItemElement)}");
 
             List<ProjectItem> relevantItems = new List<ProjectItem>();
 
@@ -2141,11 +2160,6 @@ namespace Microsoft.Build.Evaluation
                 }
             }
 
-            if (relevantItems.Count <= 1)
-            {
-                return false;
-            }
-
             foreach (ProjectItem item in relevantItems)
             {
                 item.SplitOwnItemElement();
@@ -2154,6 +2168,14 @@ namespace Microsoft.Build.Evaluation
             itemElement.Parent.RemoveChild(itemElement);
 
             return true;
+        }
+
+        internal bool ItemElementRequiresSplitting(ProjectItemElement itemElement)
+        {
+            var hasCharactersThatRequireSplitting = FileMatcher.HasWildcardsSemicolonItemOrPropertyReferences(itemElement.Include);
+            var hasMoreThanOneItem = Items.Count(i => i.Xml == itemElement) > 1;
+
+            return hasCharactersThatRequireSplitting && hasMoreThanOneItem;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


cdmihai:
Should I also change the documentation of each exploding method to say it can now throw InvalidOperationException?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -2162,6 +2176,14 @@ internal bool SplitItemElementIfNecessary(ProjectItemElement itemElement)
             return true;
         }
 
+        internal bool ItemElementRequiresSplitting(ProjectItemElement itemElement)
+        {
+            var hasCharactersThatRequireSplitting = FileMatcher.HasWildcardsSemicolonItemOrPropertyReferences(itemElement.Include);
+            var hasMoreThanOneItem = Items.Count(i => i.Xml == itemElement) > 1;
```
rainersigwald: I think this deserves a comment.

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -2132,11 +2149,13 @@ internal bool UsesProjectRootElement(ProjectRootElement xmlRootElement)
         /// </remarks>
         internal bool SplitItemElementIfNecessary(ProjectItemElement itemElement)
         {
-            if (!FileMatcher.HasWildcardsSemicolonItemOrPropertyReferences(itemElement.Include))
+            if (!ItemElementRequiresSplitting(itemElement))
             {
                 return false;
             }
 
+            ErrorUtilities.VerifyThrowInvalidOperation(!ThrowOnSplitItemElement, "OM_CannotSplitItemElementWhenSplittingIsDisabled", itemElement.Location, $"Project.{nameof(ThrowOnSplitItemElement)}");
```
rainersigwald: `{nameof(Project)}`?

src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -1426,6 +1426,9 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
   <data name="OM_CannotRemoveMetadataOriginatingFromItemDefinition">
     <value>Cannot remove the metadata named "{0}" as it originates from an item definition, rather than being directly defined on this item.</value>
   </data>
+  <data name="OM_CannotSplitItemElementWhenSplittingIsDisabled">
+    <value>A project item operation tried to split the item element at location {0} but item element splitting is disabled with {1}.</value>
```
rainersigwald: The name `ThrowOnSplitItemElement` doesn't read to me like it's "disabling". Maybe this could say

> The requested operation cannot be completed without splitting the item element at location {0} into individual components, but item element splitting is disabled with {1}.

?

src/XMakeBuildEngine/Definition/Project.cs
```diff
@@ -180,6 +180,23 @@ public Project()
...
+        /// - <see cref="RemoveItems"/>
+        /// - <see cref="ProjectItem.ChangeItemType"/>
+        /// - <see cref="ProjectItem.Rename"/>
+        /// - <see cref="ProjectItem.RemoveMetadata"/>
+        /// - <see cref="ProjectItem.SetMetadataValue"/>
+        /// 
+        /// When this property is set to true, the previous item operations throw an <exception cref="InvalidOperationException"></exception>
+        /// instead of expanding the item element. 
+        /// </summary>
+        public bool ThrowOnSplitItemElement { get; set; }
```
rainersigwald: Maybe `ThrowInsteadOfSplittingItemElement`?

cdmihai: sounds good

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: That one will never change :P

src/XMakeBuildEngine/Definition/Project.cs
cdmihai: I wouldn't know what to write other than "this is how the code did it" ... I can scour the history and see who added the splitting code.

rainersigwald: Ok, if it's not super clear why this works, at least the variable name declares a clear intention.

src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -1426,6 +1426,9 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
   <data name="OM_CannotRemoveMetadataOriginatingFromItemDefinition">
     <value>Cannot remove the metadata named "{0}" as it originates from an item definition, rather than being directly defined on this item.</value>
   </data>
+  <data name="OM_CannotSplitItemElementWhenSplittingIsDisabled">
+    <value>The requested operation needs to split the item element at location {0} into individual components but item element splitting is disabled with {1}.</value>
```
rainersigwald: I know this is the phrasing I suggested but I have a better idea. 😇 

Instead of "into individual components", "into individual elements"--then it's more precisely describing what would happen to the XML.

AndyGerlicher:
Are there any tests around the weird behavior of not expanding when there was only 1 match? Would be curious to know if anything fails if that were changed. Or if we should add one?

cdmihai:
@AndyGerlicher 

Looking at the operations that actually use the splitting method, if the element did not require splitting, then the operations apply on the actual xml of that item. So the item number condition seems more of a perf optimization: if there is only one item, don't bother splitting it out because the calling operations will operate directly on the xml,

Also, none of the item operations are glob aware, except Rename. Rename item actually checks if the new name is covered by the glob that produced the item. If it is covered, then it only updates the object and does nothing to the xml.

I will add a comment to the condition. :)

cdmihai:
@jviau does this look OK?

jviau:
@cdmihai - so this will not throw if you have `<Compile Include="**\*.cs" />`, only 1 .cs file covered by it  in the current evaluation and you try to perform an operation on that single item?

cdmihai:
@jviau That's correct. Apparently the splitting code does not take globbing into account. So if it is only a glob that produced one item, then the item operations will silently operate on the glob (rename, remove, change type, etc). I'll look into adding globbing characters in splitting method's blacklist.

jviau:
@cdmihai that would be preferable, we don't want to ever modify a glob from an evaluated ProjectItem, regardless of it is the only item or not produced by the glob.

cdmihai:
@jviau My bad, I misread the code. Globs are also considered when splitting, as expected :)
I'll write more tests to make sure exceptions are thrown for them too

cdmihai:
I made the following changes:
- splitting is not conditioned on the number of items anymore
- added tests to assert that the methods throw when globbing is present with either one or multiple items

jviau:
Looks great to me.

cdmihai:

PR #1091, opened 2016-09-27
Update versioning in new packages

jeffkl:
New packages use CreatePackages.proj instead of the dirs.proj.  This just updates the versioning logic to match what was previously done in master so that these packages will have the same versioning.

* Update version of Nerdbank.GitVersioning in our dir.props
* Add release branch of xplat for now

Related to #1039 

FYI @OmarTawfik


jeffkl:
I have launched a build and if the versions are correct I'll publish to MyGet and NuGet.

PR #1093, opened 2016-09-27
Allow item metadata to be expressed as attributes on the item element instead of child Xml elements

### LINKED ISSUE #708: Support Item Metadata as Attributes
**This issue is in the open design phase and is a part of the [project.json -> csproj ](https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/) conversion effort.**

Support (optional) terser syntax for specifying metadata on items. This could be useful in a scenario such as specifying a list of NuGet packages (purely as a hypothetical example):
```XML
<ItemGroup>
  <NuGetPackage Include="Newtonsoft.Json">
    <Version>9.0.1-beta1</Version>
  </NuGetPackage>
</ItemGroup>
```
Would become:
```XML
<ItemGroup>
  <NuGetPackage Include="Newtonsoft.Json" Version="9.0.1-beta1" />
</ItemGroup>
```

For compat reasons this could not be added to an existing file and only written if you opt-in to a new "schema version" (or whatever).

dsplaisted:
Resolves #708

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Text;
 using System.Collections.Generic;
+using System.Linq;
 using System.Xml;
 using System.Diagnostics;
 using Microsoft.Build.Framework;
@@ -429,6 +430,24 @@ namespace Microsoft.Build.Construction
                 //  TODO: What if there are multiple elements with the same name?  That shouldn't be allowed for elements
                 //      expressed as attributes.
                 ProjectXmlUtilities.SetOrRemoveAttribute(XmlElement, child.XmlElement.Name, value);
+
+                if (XmlElement.HasChildNodes)
+                {
+                    bool allWhitespace = true;
+                    foreach (XmlNode childXml in XmlElement.ChildNodes)
+                    {
+                        if (childXml.NodeType != XmlNodeType.Whitespace)
+                        {
+                            allWhitespace = false;
+                            break;
+                        }
+                    }
+
+                    if (allWhitespace)
+                    {
+                        XmlElement.IsEmpty = true;
+                    }
+                }
             }
             else
             {
--- a/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
@@ -420,34 +420,52 @@ namespace Microsoft.Build.Construction
             return clone;
         }
 
+        private void SetElementAsAttributeValue(ProjectElement child)
+        {
+            //  Assumes that child.ExpressedAsAttribute is true
+
+            string value = Microsoft.Build.Internal.Utilities.GetXmlNodeInnerContents(child.XmlElement);
+            ProjectXmlUtilities.SetOrRemoveAttribute(XmlElement, child.XmlElement.Name, value);
+
+            if (XmlElement.HasChildNodes)
+            {
+                bool allWhitespace = true;
+                foreach (XmlNode childXml in XmlElement.ChildNodes)
+                {
+                    if (childXml.NodeType != XmlNodeType.Whitespace)
+                    {
+                        allWhitespace = false;
+                        break;
+                    }
+                }
+
+                if (allWhitespace)
+                {
+                    XmlElement.IsEmpty = true;
+                }
+            }
+        }
+
+        //  If child "element" is actually represented as an attribute, update the value there
+        internal void UpdateElementValue(ProjectElement child)
+        {
+            if (child.ExpressedAsAttribute)
+            {
+                SetElementAsAttributeValue(child);
+            }
+        }
+
         internal void AddToXml(ProjectElement child)
         {
             if (child.ExpressedAsAttribute)
             {
-                string value = Microsoft.Build.Internal.Utilities.GetXmlNodeInnerContents(child.XmlElement);
+                //  Assume that the name of the child has already been validated to conform with rules in XmlUtilities.VerifyThrowArgumentValidElementName
 
-                //  TODO: Do we need special handling for namespaces for the name, or for names with colons in them?
-                //  TODO: What if there are multiple elements with the same name?  That shouldn't be allowed for elements
-                //      expressed as attributes.
-                ProjectXmlUtilities.SetOrRemoveAttribute(XmlElement, child.XmlElement.Name, value);
+                //  Make sure we're not trying to add multiple attributes with the same name
+                ProjectErrorUtilities.VerifyThrowInvalidProject(!XmlElement.HasAttribute(child.XmlElement.Name),
+                    XmlElement.Location, "InvalidChildElementDueToDuplication", child.XmlElement.Name, ElementName);
 
-                if (XmlElement.HasChildNodes)
-                {
-                    bool allWhitespace = true;
-                    foreach (XmlNode childXml in XmlElement.ChildNodes)
-                    {
-                        if (childXml.NodeType != XmlNodeType.Whitespace)
-                        {
-                            allWhitespace = false;
-                            break;
-                        }
-                    }
-
-                    if (allWhitespace)
-                    {
-                        XmlElement.IsEmpty = true;
-                    }
-                }
+                SetElementAsAttributeValue(child);
             }
             else
             {
--- a/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
@@ -430,7 +430,6 @@ namespace Microsoft.Build.Construction
             ProjectXmlUtilities.SetOrRemoveAttribute(XmlElement, child.XmlElement.Name, value);
         }
 
-        //  
         /// <summary>
         /// If child "element" is actually represented as an attribute, update the value in the corresponding Xml attribute
         /// </summary>
--- a/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
@@ -15,6 +15,7 @@ using Microsoft.Build.Shared;
 using System.Collections.ObjectModel;
 using Microsoft.Build.Construction;
 using Microsoft.Build.Collections;
+using Microsoft.Build.Internal;
 
 namespace Microsoft.Build.Construction
 {
@@ -171,19 +172,7 @@ namespace Microsoft.Build.Construction
                 child.NextSibling.PreviousSibling = child;
             }
 
-            XmlElement.InsertAfter(child.XmlElement, reference.XmlElement);
-            if (XmlDocument.PreserveWhitespace)
-            {
-                //  If we are trying to preserve formatting of the file, then the new node won't automatically be indented.
-                //  So try to match the surrounding formatting by checking the whitespace that precedes the node we inserted
-                //  after, and inserting the same whitespace between the previous node and the one we added
-                if (reference.XmlElement.PreviousSibling != null &&
-                    reference.XmlElement.PreviousSibling.NodeType == XmlNodeType.Whitespace)
-                {
-                    var newWhitespaceNode = XmlDocument.CreateWhitespace(reference.XmlElement.PreviousSibling.Value);
-                    XmlElement.InsertAfter(newWhitespaceNode, reference.XmlElement);
-                }
-            }
+            AddToXml(child);
 
             _count++;
             MarkDirty("Insert element {0}", child.ElementName);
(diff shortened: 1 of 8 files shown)


AndyGerlicher:
Looks good so far, I like the approach! I'd like to see some more tests before we commit. Especially around adding elements that can't really be attributes (need to be escaped, etc.).

dsplaisted:
I've added some more tests and fixed some minor things.  I still need to add more tests around error cases, and we may need to add APIs to the project model in addition to the construction APIs.

dsplaisted:
I've added more tests and fixed some issues.  I think this is ready for review now.

jeffkl:
:shipit:

dsplaisted:
Anyone else want to review this one, or shall I go ahead and merge it?

rainersigwald:
I'll look this morning

jviau:
I think this PR is fine without the error rainersigwald brought up. MSBuild already supports 2 children of the same tag, taking the last one as the evaluated result. I think in this scenario the attribute can be considered before the child, so the a child metadata will win over an attribute.

For example, MSBuild supports this in the project file with `anything` being the result of `m1`:
```xml
<i1 Include="I">
  <m1>v1</m1>
  <m1>anything</m1>
</i1>
```

On a side note, this is a bit interesting. With the duplicate metadata you can retrieve the `ProjectItem` for that element, remove the `v1` metadata, and the project item will no longer have that metadata, but the `ProjectItemElement` does! So when you re-evaluted, the 1st metadata comes back to the project item. This is inconsistent, and either the remove should remove both metadata, or the project item should report the 1st metadata after the 2nd is removed without a re-evaluation. But this is a separate issue I feel.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -455,6 +420,154 @@ protected internal virtual ProjectElementContainer DeepClone(ProjectRootElement
             return clone;
         }
 
+        private void SetElementAsAttributeValue(ProjectElement child)
+        {
+            //  Assumes that child.ExpressedAsAttribute is true
```
dsplaisted: Done

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -455,6 +420,154 @@ protected internal virtual ProjectElementContainer DeepClone(ProjectRootElement
...
+
+            string value = Microsoft.Build.Internal.Utilities.GetXmlNodeInnerContents(child.XmlElement);
+            ProjectXmlUtilities.SetOrRemoveAttribute(XmlElement, child.XmlElement.Name, value);
+
+            if (XmlElement.HasChildNodes)
```
dsplaisted: Moved to the RemoveFromXml method and added comments

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -455,6 +420,154 @@ protected internal virtual ProjectElementContainer DeepClone(ProjectRootElement
...
+                }
+
+                if (allWhitespace)
+                {
+                    XmlElement.IsEmpty = true;
+                }
+            }
+        }
+
+        //  If child "element" is actually represented as an attribute, update the value there
```
dsplaisted: Done

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -455,6 +420,154 @@ protected internal virtual ProjectElementContainer DeepClone(ProjectRootElement
...
+
+                //  Make sure we're not trying to add multiple attributes with the same name
+                ProjectErrorUtilities.VerifyThrowInvalidProject(!XmlElement.HasAttribute(child.XmlElement.Name),
+                    XmlElement.Location, "InvalidChildElementDueToDuplication", child.XmlElement.Name, ElementName);
+
+                SetElementAsAttributeValue(child);
+            }
+            else
+            {
+                if (child.PreviousSibling != null)
```
dsplaisted: I've added comments explaining this

src/XMakeBuildEngine/Construction/ProjectItemElement.cs
```diff
@@ -345,11 +345,17 @@ internal ProjectItemElement NextItem
         /// </summary>
         public ProjectMetadataElement AddMetadata(string name, string unevaluatedValue)
         {
+            return AddMetadata(name, unevaluatedValue, false);
+        }
+
+        public ProjectMetadataElement AddMetadata(string name, string unevaluatedValue, bool expressAsAttribute)
```
dsplaisted: Done

src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
```diff
@@ -48,6 +48,14 @@ public string Name
             set { ChangeName(value); }
         }
 
+        //  Add a new property with the same name here because this attribute should be public for ProjectMetadataElement,
+        //  but internal for ProjectElement, because we don't want it to be settable for arbintrary elements.
```
dsplaisted: Done

dsplaisted: Fixed

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -368,6 +389,53 @@ private ProjectItemElement ParseProjectItemElement(XmlElementWithLocation elemen
...
+                }
+                else if (name.Equals(s_knownAttributesOnItem[i], StringComparison.OrdinalIgnoreCase))
+                {
+                    isKnownAttribute = false;
+                    isValidMetadataNameInAttribute = false;
+                    return;
+                }
+            }
+
+            //  Resorve attributes starting with underscores in case we need to add more built-in attributes later
```
dsplaisted: Fixed

dsplaisted: Pretty much, unless we're OK with potentially breaking some people.  Using an underscore was the best thing I could think of... if you have a better idea I'd love to hear it.

src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -1161,6 +1161,10 @@
     <value>MSB4173: The element &lt;{0}&gt; beneath element &lt;{1}&gt; is invalid because a child element with that name already exists</value>
     <comment>{StrBegin="MSB4173: "}</comment>
   </data>
+  <data name="InvalidAttributeDueToDuplication" UESanitized="false" Visibility="Public">
+    <value>MSB4227: The attribute &lt;{0}&gt; on element &lt;{1}&gt; is invalid because an attribute with that name already exists</value>
```
dsplaisted: I'm matching the formatting here from the previous error message for a similar situation

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectMetadataElement_Tests.cs
```diff
@@ -77,6 +87,25 @@ public void ReadInvalidName()
...
+        {
+            Assert.Throws<InvalidProjectFileException>(() =>
+            {
+                string content = @"
+                    <Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003' >
+                        <ItemGroup>
+                            <i Include='i1' "+ "\u03A3" + @"='v1' />
+                        </ItemGroup>
+                    </Project>
+                ";
```
dsplaisted: I've fixed this (though the pattern was just copied from the previous test).

dsplaisted:
As far as having a metadata attribute with the same name as a metadata element, I'd like to disallow it... but we'd have to add the check in several places (adding an attribute, changing its name, changing whether it's expressed as an attribute, parsing the XML).  It could potentially affect the perf.

So since this is a rare scenario where we already have some funky behavior, I'm leaning towards saying it's not worth adding the checks.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -469,14 +467,19 @@ internal void AddToXml(ProjectElement child)
             }
             else
             {
+                //  We want to add the XmlElement to the same position in the child list as the corresponding ProjectElement.
+                //  Depending on whether the child ProjectElement has a PreviousSibling or a NextSibling, we may need to
+                //  use the InsertAfter, InsertBefore, or AppendChild methods to add it in the right place.
+                //
+                //  Also, if PreserveWhitespace is true, then the elements we add won't automatically get indented, so
+                //  we try to match the surrounding formatting.
```
rainersigwald: 👍

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -423,30 +423,18 @@ protected internal virtual ProjectElementContainer DeepClone(ProjectRootElement
...
-
-                if (allWhitespace)
-                {
-                    XmlElement.IsEmpty = true;
-                }
-            }
         }
 
-        //  If child "element" is actually represented as an attribute, update the value there
+        //
```
rainersigwald: ultra nit: extra blank comment line

src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
```diff
@@ -49,7 +49,14 @@ public string Name
         }
 
         //  Add a new property with the same name here because this attribute should be public for ProjectMetadataElement,
-        //  but internal for ProjectElement, because we don't want it to be settable for arbintrary elements.
+        //  but internal for ProjectElement, because we don't want it to be settable for arbitrary elements.
```
rainersigwald: nit: is this a comment before the doc comments? should it just be in `<remarks>`?

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
rainersigwald: One concern: there's a common pattern of naming _properties_ with leading underscores as a clue to the user that they're internal implementation details. I don't think this is common for item metadata, though, so I don't think we'll be stepping on too many toes here.

The only other idea I have to make this even rarer is to take a page from the C++ book and use double underscores. Which is very ugly.

Thoughts, @AndyGerlicher @cdmihai @Sarabeth-Jaffe-Microsoft @jeffkl ?

src/XMakeBuildEngine/Resources/Strings.resx
rainersigwald: All of the messages that I see referring to attributes use quotes (messages referring to elements use angle brackets). That makes sense to me.

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -358,6 +356,27 @@ private ProjectItemElement ParseProjectItemElement(XmlElementWithLocation elemen
 
             ProjectItemElement item = new ProjectItemElement(element, parent, _project);
 
+            foreach (XmlAttributeWithLocation attribute in element.Attributes)
+            {
+                bool isKnownAttribute;
+                bool isValidMetadataNameInAttribute;
+
+                CheckMetadataAsAttributeName(attribute.Name, out isKnownAttribute, out isValidMetadataNameInAttribute);
```
cdmihai: Should this invalid attribute checking logic also be callable from the construction APIs? Otherwise API callers might set an invalid metadata name and only know about it later when the xml does not parse.

cdmihai:
Should the following construction APIs be update as well?
- [ProjectItem.SetMetadata](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Definition/ProjectItem.cs#L515)
- [ProjectRootElement.CreateMetadataElement](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Construction/ProjectRootElement.cs#L1525)

src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
dsplaisted: It's talking about internal APIs, so I don't think it goes in the public API documentation that comes from the Xml comments

dsplaisted:
@cdmihai 

> Should the following construction APIs be updated as well?
> - [ProjectItem.SetMetadata](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Definition/ProjectItem.cs#L515)
> - [ProjectRootElement.CreateMetadataElement](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Construction/ProjectRootElement.cs#L1525)

I don't think they need to be.  `ProjectRootElement.CreateMetadataElement` returns a `ProjectMetadataElement`, which you can set the `ExpressedAsAttribute` property on.  Likewise, `ProjectItem.SetMetadata` returns a `ProjectMetadata`, which has an `Xml` property of type `ProjectMetadataElement`, which you can set the `ExpressedAsAttribute` property on.

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
dsplaisted: Ultra fixed :-)

src/XMakeBuildEngine/Resources/Strings.resx
dsplaisted: Oh, I see what you mean.  I've updated it accordingly.

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
dsplaisted: Good point, I've added checks to the construction APIs for this.

jviau:
@rainersigwald yes this is fine for CPS.

PR #1094, opened 2016-09-27
Allow extending the AllProjectOutputGroups target

kzu:
Custom output groups that want to plug into the AllProjectOutputGroups target
need a way to add their own targets to the list. One case is the upcoming
NuGetizer PackageProjectOutputGroup, which we'd like to nicely tie into the
AllProjectOutputGroups like all other built-in groups.

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -5127,6 +5127,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     of additional attributes may be placed on an output/dependency item.
     ============================================================
     -->
+  <PropertyGroup>
+    <AllProjectOutputGroupsDependsOn></AllProjectOutputGroupsDependsOn>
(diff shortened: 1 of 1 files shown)


kzu:
Used from [nugetizer](https://github.com/NuGet/NuGet.Build.Packaging/blob/dev/src/Build/NuGet.Build.Packaging.Tasks/NuGet.Build.Packaging.targets#L141) already whenever this lands :)

AndyGerlicher:
@kzu we're not sure this is necessary. The preferred pattern here would be to use `BeforeTargets="AllProjectOutputGroups"` on the Target you're trying to run before this one. The pattern you're using for extension is an older one before the `BeforeTargets` feature was added. We'd rather not see that grow if possible. Would that work?

kzu:
I wasn't aware the BeforeTargets/AfterTargets was preferred over the *DependsOn pattern :). I find that the latter, when used properly, makes it easier to read through targets and see what happens when, whereas the former, if sprinkled all over the place, makes things a tad harder.

But it makes sense to close this PR, no prob.

PR #1101, opened 2016-09-27
Tests for absolute / relative includes and excludes

cdmihai:
Dev15 MSBuild introduced a change of behaviour.

Dev14 MSBuild did not match excludes and includes when they referred to
the same file, but one was relative and the other absolute

Dev15 MSBuild knows how to match excludes and includes when one is relative and
the other is absolute

### DIFF
--- a/src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
+++ b/src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
@@ -663,26 +663,138 @@ namespace Microsoft.Build.UnitTests.OM.Definition
         }
 
         [Theory]
+        // exclude matches include; file is next to project file
         [InlineData(ItemWithIncludeAndExclude,
-            @"**\*",
-            "foo",
-            new[]
+            @"a", // include item
+            @"", //path relative from projectFile. Empty string if current directory
+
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs
```diff
@@ -663,26 +663,138 @@ private static void TestIncludeExclude(string projectContents, string[] inputFil
...
+                finally
+                {
+                    ObjectModelHelpers.DeleteDirectory(testRoot);
+                    Directory.Delete(testRoot);
+                }
+            };
+
+            runTest(true, false);
+            runTest(false, true);
+            runTest(true, true);
```
rainersigwald: If you adjust one of these to fail, is the error understandable? It might be more diagnosable to split each combination here into different test cases.

cdmihai: The assertion will be that the expected items do not match the actual items. Then by looking at the stack trace you see which call failed. Is that okay? :)

rainersigwald: It's not great, but I can live with it.

PR #1100, opened 2016-09-27
Update DotnetCli version to 1.0.0-preview3-003686

jeffkl:
This is needed in order to use some new stuff in the CLI

PR #1102, opened 2016-09-27
Archive `init-tools.log` in CI builds

rainersigwald:
This will make it easier to diagnose CI build failures that happen during
init--currently there's a message printed to stderr that says to look at a
file that's already deleted.


rainersigwald:
@jeffkl This should help diagnose what's wrong with #1100.

jeffkl:
You complete me

PR #1104, opened 2016-09-27
Prevent /p:TargetFramework=x from flowing to GetTargetFrameworkProperties

nguerrera:
TargetFramework was getting passed down from referencing project to referenced project, which would send the referenced project straight to wrong inner build and return no properties because it gets the default target for classic projects instead of the new one from the SDK.

@rainersigwald @eerhardt @srivatsn @AndyGerlicher

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1511,7 +1511,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         BuildInParallel="$(BuildInParallel)"
         Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); ReferringTargetFramework=$(TargetFrameworkMoniker)"
         ContinueOnError="!$(BuildingProject)"
-        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
+        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove);TargetFramework">
 
       <Output TaskParameter="TargetOutputs" PropertyName="_ProjectReferenceTargetFrameworkProperties" />
     </MSBuild>


src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1511,7 +1511,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         BuildInParallel="$(BuildInParallel)"
         Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); ReferringTargetFramework=$(TargetFrameworkMoniker)"
         ContinueOnError="!$(BuildingProject)"
-        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
+        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove);TargetFramework">
```
eerhardt: what do you think about adding this value to GlobalPropertiesToRemove?

rainersigwald: It's not new, is it? If it had some meaning before, adding it here seems like a scary breaking change.

nguerrera: I wasn't sure if that would break the other invocations where we pass it back down. Does it remove from the set in Properties?

nguerrera: Also, wanted to get in the change that matches what got the SDK tests happy again and not code more stuff blind. I'd like to start adding tests and would be more comfortable making further changes if the tests were in.

nguerrera: GlobalPropertiesToRemove is not new, if that's the question.

nguerrera: It affects all P2P builds. I think this is correct to only remove it from this one as the others will pass it back in with the right value.

rainersigwald: No, I was thinking of `TargetFramework`.

eerhardt: Or is it ONLY needed when calling "outer build" targets?

nguerrera: TargetFramework is new.

nguerrera: Correct behavior is:

* When calling GetTargetFrameworkProperties (here), never set TargetFramework, it's job is to tell you what the TargetFramework to pass is.
* When building other, pass TargetFramework returned from here.

PR #1107, opened 2016-09-28
CI localized builds via cibuild.cmd

cdmihai:
Depends on #1108


cdmihai:
I am unaware of such a special case, Can you please detail?
On master loc is always turned on with https://github.com/Microsoft/msbuild/blob/master/cibuild.cmd

rainersigwald:
That’s probably what I was thinking of.

Issue #800, opened 2016-07-20
Separate build phase to test our packages


cdmihai:
It would be nice if, after the build, we insert another build step that tests the nuget packages (+ other packages, like vsix)

- change PortableTask so it depends on the freshly built nuget packages and not on the projects
- publish the local nuget packages into some local nuget feed
- test that they restore
- test that Portable task works with them

Stretch goals:
- deploy msbuild from the local feed and test that it builds some sample projects
- figure out how to test the vsix

cdmihai:
The sample projects could also use the freshly built packages.

jeffkl:
The packages aren't built in CI because they now contain both full framework and .net core assemblies.  They are built in the official build after building both so it can pack up what it needs. :(

We would need a new set of projects that don't get built except in the official build after the packages are created...

cdmihai:
We could also create a new set of CI jobs that don't run tests and just build and test the packages.

PR #1111, opened 2016-09-28
Fix broken link in NuGet packages

ChrisMaddock:


jeffkl:
Thanks for catching this. We're better about catching this with #800

PR #1113, opened 2016-09-28
Update public API

jeffkl:
New overload to ProjectRootElement.Create()

Update nuspec with casing issue.


jeffkl:
FYI @AndyGerlicher

PR #1121, opened 2016-09-28
Separate Xlf file syncing from localized builds

cdmihai:
This improves development experience: ideally, we'd want devs to always produce satellites. If satellite production also updates xlf files, then this produces lots of unwanted diffs (if the dev does not want to commit xlf changes).

### DIFF
--- a/src/UpdateLocalizedResources.targets
+++ b/src/UpdateLocalizedResources.targets
@@ -1,73 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="UpdateXlf" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   
-  <UsingTask AssemblyFile="$(BuildBinDir)\Microsoft.Build.LocalizationTasks.dll" Condition="$(LocalizedBuild) == 'true'" TaskName="SaveXlfToResx"/>
-  <UsingTask AssemblyFile="$(BuildBinDir)\Microsoft.Build.LocalizationTasks.dll" Condition="$(LocalizedBuild) == 'true'" TaskName="UpdateXlfFromResx"/>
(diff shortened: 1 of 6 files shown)


cdmihai:
@jeffkl @rainersigwald 

Once this goes in I will update the xplat microbuild to pass in the `/p:SyncXlf=true` property. Otherwise, newly added properties will cause non English msbuilds to crash with missing resource (non english satellites get their resources from generated resx files, which are generated from xlf files)

src/UpdateLocalizedResources.targets
```diff
@@ -1,27 +1,31 @@
...
+          BeforeTargets="$(PrepareResourcesDependsOn);PrepareResources"
+          DependsOnTargets="SyncXlf;GenerateLocalizedResxResourcesFromXlf;AddEnglishResxFilesToEmbeddedResource"
+          Condition="'$(LocalizedBuild)' == 'true'">
 
-  <!-- Emit the localized EmbeddedResource items before the common targets start using them -->
-  <Target Name="EmitLocalizedResources"
-          BeforeTargets="$(PrepareResourceNamesDependsOn);PrepareResourceNames">
     <ItemGroup>
-      <NeutralResources Include="@(EmbeddedResource)" Condition="'%(Extension)' == '.resx'"/> 
+      <EmbeddedResource Include="@(LocalizedResxResources)"/>
```
rainersigwald: So will this just be the English ones now?

cdmihai:
@dotnet-bot test this please

src/UpdateLocalizedResources.targets
cdmihai: LocalizedResxResources represents all the generated, non English localized resx files from the xlf files. The dependency of this target AddEnglishResxFilesToEmbeddedResource does the English workaround and adds itself to EmbeddedResource

cdmihai:

cdmihai:

src/UpdateLocalizedResources.targets
rainersigwald: explained offline: you want all of the resources here, so that other logic can take only the ones it wants to create satellite assemblies.

PR #1126, opened 2016-09-29
Fix package creation

jeffkl:
For some reason this broke but it shouldn't be importing the parent dir.props because it overrides $(BaseOutputPath)


rainersigwald:
If it shouldn't be importing the parent directory's `dir.props`, don't use `MSBuild]::GetDirectoryNameOfFileAbove`--just import the right one directly.

PR #1131, opened 2016-09-29
Introduce CrossTargeting extensibility imports

### LINKED ISSUE #1061: Generalize how targets can be provided to cross-targeting build
To get things off the ground, I'm having to put an Import of Common\ImportAfter\*.NuGet.targets in CrossTargeting.targets. This is undue coupling to NuGet. We need to generalize, but can't import everything in that folder as the rest is not cross-targeting aware.

rainersigwald:
An "outer" build that imports `CrossTargeting.targets` can't use
`Microsoft.Common.targets`'s extensibility hooks, because some targets in
those folders may expect a fully-specified ("inner") build environment.
But NuGet (at least) needs to be able to extend the outer build--so that
it can restore to get to the inner builds.

This introduces a new folder
`Microsoft.Common.CrossTargeting.targets\ImportAfter` and a new property
`CustomAfterMicrosoftCommonCrossTargetingTargets` that allow the
CrossTargeting (outer) build to be extended.

Closes #1061.

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
+++ b/src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
@@ -11,6 +11,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 -->
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets Condition="'$(ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets)' == ''">true</ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.CrossTargeting.targets\ImportBefore\*.targets"
+          Condition="'$(ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets)' == 'true' and exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.CrossTargeting.targets\ImportBefore')"/>
+
+  <Import Project="$(CustomBeforeMicrosoftCommonCrossTargetingTargets)" Condition="'$(CustomBeforeMicrosoftCommonCrossTargetingTargets)' != '' and Exists('$(CustomBeforeMicrosoftCommonCrossTargetingTargets)')"/>
+
   <!--
   ============================================================
                                        DispatchToInnerBuilds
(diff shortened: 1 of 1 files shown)


src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -95,19 +95,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
   </Target>
 
-  <!--
-    This will import NuGet restore targets, which is a special case separate from the package -> project extension 
-    mechanism below. For obvious reasons,  we need restore to work before any package assets are available.
+  <Import Project="$(CustomAfterMicrosoftCommonCrossTargetingTargets)" Condition="'$(CustomAfterMicrosoftCommonCrossTargetingTargets)' != '' and Exists('$(CustomAfterMicrosoftCommonCrossTargetingTargets)')"/>
```
rainersigwald: I followed the pattern from common targets and put this before the folder include, but I'm not sure that's right. @Emgarten you mentioned wanting to carry the restore target outside of the MSBuild folder--will it _always_ be outside? If it's in both places, with this ordering the one in `MSBuildExtensionsPath` will win which is the opposite of what we'd want. In that case, I can move this import below the folder one.

rainersigwald: This won't be a problem if the NuGet targets are injected by the `ImportBefore` hook instead of `After`.

rainersigwald:
@nguerrera @Emgarten Does this look like it'll do the trick?

src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -95,19 +103,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    Allow extensions like NuGet restore to work before any package assets are available.
   -->
   <PropertyGroup>
-    <ImportByWildcardAfterMicrosoftCommonTargets Condition="'$(ImportByWildcardAfterMicrosoftCommonTargets)' == ''">true</ImportByWildcardAfterMicrosoftCommonTargets>
+    <ImportByWildcardAfterMicrosoftCommonCrossTargetingTargets Condition="'$(ImportByWildcardAfterMicrosoftCommonCrossTargetingTargets)' == ''">true</ImportByWildcardAfterMicrosoftCommonCrossTargetingTargets>
   </PropertyGroup>
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.targets\ImportAfter\*.NuGet.*.targets" 
-          Condition="'$(ImportByWildcardAfterMicrosoftCommonTargets)' == 'true' and exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.targets\ImportAfter')"/>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.CrossTargeting.targets\ImportAfter\*.targets" 
+          Condition="'$(ImportByWildcardAfterMicrosoftCommonCrossTargetingTargets)' == 'true' and exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.CrossTargeting.targets\ImportAfter')"/>
```
nguerrera: Can we have a grace period where the *.NuGet.*.targets is still imported from common/importafter? We're almost certain to have code flow such that we end up with a broken integration.

rainersigwald: Yeah, that's a really good idea.

rainersigwald: Done.

src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
emgarten: Yes, this looks like it allow NuGet to get extra targets into the project to continue the walk.

emgarten:
Yes, this looks like it will handle what NuGet needs, thanks!

rainersigwald:
Awesome. It's available in the latest packages from the buildtools feed (`0.1.0-preview-00044-160929`).

PR #1138, opened 2016-10-01
Localize legacy projects

cdmihai:
Localizing these 2 projects exposed a race in the `UpdateXlfFromResx` task. All projects share the `Strings.shared.resx` resource file. Since `Strings.shared.resx` does not have its own project, and since localization is project based, all projects try to update the xlf files in `/Shared/Resources/xlf` and the build fails. Happens maybe 60% of the time :(

I'll try and fix the race and send it as a separate PR.


src/XMakeConversion/Resources/xlf/Strings.cs.xlf
```diff
@@ -0,0 +1,88 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
+  <file datatype="xml" source-language="en" target-language="cs" original="../Strings.resx">
+    <body>
+      <group id="XMAKECONVERSION/RESOURCES/STRINGS.RESX" datatype="resx" />
+      <trans-unit id="CouldNotFindCompleteSolutionFile">
+        <source>MSB2013: The project-to-project reference with GUID {0} could not be converted because a valid .SLN file containing all projects could not be found.</source>
+        <target state="new">MSB2013: The project-to-project reference with GUID {0} could not be converted because a valid .SLN file containing all projects could not be found.</target>
```
rainersigwald: Wasn't there a reason that we didn't want these to say `new` for the initial checkin?

cdmihai: In this case `new` is fine, since they are new resources. We should use `update` though when we change a resource's contents but not change the resource ID. That change depended on the xliffparser posting a new nuget package, which seems to have happened. :)

cdmihai:

cdmihai:

cdmihai:

PR #1142, opened 2016-10-03
Update vsix setup

AndyGerlicher:
Include `Microsoft.Build.Conversion.Core.dll`, `Microsoft.Build.Engine.dll`, and the schema files to match existing install structure.

### DIFF
--- a/setup/files.swr
+++ b/setup/files.swr
@@ -10,7 +10,9 @@ folder InstallDir:\MSBuild\15.0
   file source=$(X86BinPath)Microsoft.VisualStudioVersion.v15.Common.props
 
 folder InstallDir:\MSBuild\15.0\Bin
+  file source=$(X86BinPath)Microsoft.Build.Conversion.Core.dll vs.file.ngen=yes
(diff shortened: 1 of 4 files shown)


setup/files.swr
```diff
@@ -135,7 +141,9 @@ folder InstallDir:\MSBuild\15.0\Bin\amd64
   file source=$(X64BinPath)MSBuild.exe.config
   file source=$(X64BinPath)MSBuildTaskHost.exe.config
 
+  file source=$(X86BinPath)Microsoft.Build.Conversion.Core.dll vs.file.ngen=yes
   file source=$(X86BinPath)Microsoft.Build.dll vs.file.ngen=yes
+  file source=$(X86BinPath)Microsoft.Build.Engine.dll vs.file.ngen=yes
```
rainersigwald: These really were AnyCPU before, too, right?

AndyGerlicher: Yeah they were and are. Verified corflags output is the same.

PR #1145, opened 2016-10-03
Fix xlf syncing race

cdmihai:
This solves a race in the SyncXlf task. Because localization is per project, and because MSBuild uses a Shared directory with shared resources (which is not a project but gets linked in by all the other projects),
each project will try to sync the shared xlf files and crash with concurrent writes.

Short term fix is to disable parallel builds when xlfs are synced. Long term fix would be turning the stuff under Shared into a csproj project.

#1138 depends on this.

### DIFF
--- a/build/LocalizationTasks/EmitLocalizedResources.cs
+++ b/build/LocalizationTasks/EmitLocalizedResources.cs
@@ -114,8 +114,7 @@ namespace Microsoft.Build.LocalizationTasks
 
         private static string GetCultureCodeFromPath(string localizedResourcePath)
         {
-            return Path.GetExtension(Path.GetFileNameWithoutExtension(localizedResourcePath)).Substring(1);
-                // remove the . at the beginning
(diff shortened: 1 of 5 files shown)


cdmihai:
The perf loss isn't that much, since MSBuild does not have many projects. This will only happen on dev machines, when somebody wants to update the xlf files and commit the changes.

True, we can also anoint one project to own syncing the shared files. It would mean adding an item remove, conditioned on whether $(AssemblyName) is the special project. Which one do you prefer? With this PR you loose speed when syncing xlf files (_should_ happen only when we want to ping the loc team). With your suggestion, we keep the speed but hardcode an assembly name, so if the name changes at some point, it will add ~30 minutes of debugging to whomever has to investigate and fix :)

rainersigwald:
Doesn't every official build also sync the xlf files?

I don't have a strong preference; we can go with this way as its already done.

cdmihai:
Ah, you're right, the CI does sync the xlf files. Alright, going to do the "special project" approach :)

PR #1143, opened 2016-10-03
Move HOME folder on CI machines

jeffkl:
This moves the HOME folder to $WORKSPACE so that CI machines will delete the NuGet package cache and start fresh every time.

PR #1146, opened 2016-10-03
Eliminate assembly load context

### LINKED ISSUE #750: Native image cannot be loaded multiple times

tmeschter:
Fixes #750.

Eliminate our use of `AssemblyLoadContext` in favor of simply hooking `AssemblyLoadContext.Resolving` on the default context.

This also allows us to delete the code that looks in the base directory for dependencies, since the default context will already do that. Also, we need to update the version of the runtime that gets deployed for tests, as this fix depends on other runtime fixes.


tmeschter:
@rainersigwald @jeffkl @cdmihai @AndyGerlicher Please review.

tmeschter:
@dotnet-bot Retest OSX please.

tmeschter:

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -12,35 +12,16 @@ namespace Microsoft.Build.Shared
...
-        /// sets that instance to be the default <see cref="AssemblyLoadContext" />,
-        /// and returns that instance. Throws if the Default is already set or the
-        /// binding model is already locked.
-        /// </summary>
-        public static CoreClrAssemblyLoader CreateAndSetDefault()
-        {
-            var assemblyLoader = new CoreClrAssemblyLoader();
-            return assemblyLoader;
-        }
+        private static readonly string[] _extensions = new[] { "ni.dll", "ni.exe", "dll", "exe" };
```
jeffkl: Looks like you can remove this now

tmeschter: No, this one is still needed. There was a separate list called `s_extensions` with *almost* the same information, and that one is no longer needed and has been removed.

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -66,6 +47,12 @@ public Assembly LoadFromPath(string fullPath)
 
             lock (_guard)
             {
+                if (!_resolvingHandlerHookedUp)
+                {
+                    AssemblyLoadContext.Default.Resolving += TryResolveAssembly;
```
cdmihai: Would it ever be relevant to remove this handler at some point? Like in between builds when nodereuse is true. @rainersigwald I don't remember, does node reuse work on xplat?

tmeschter: @cdmihai If we could, it would be useful to be able to return an msbuild process to some sort of "clean" state, where task assemblies were unloaded and the event handler removed. As we cannot currently unload the assemblies, removing the event handler isn't that interesting.

tmeschter:

jeffkl:
I think I fixed this in #1143 

tmeschter:

jeffkl:
hopefully fixed by #1158

jeffkl:

PR #1152, opened 2016-10-04
Add app.config redirects for deprecated assemblies

rainersigwald:
After c3f85b added these to our VSIX and we stopped using the ones built
from the VS repo (as version 15.0), various things in VS broke because
Microsoft.Build.Engine 15.0 couldn't be loaded. Adding redirects for it
and Conversion.Core which was changed similarly.

Fixes the MSBuild.exe side of
https://devdiv.visualstudio.com/DevDiv/_workitems?id=273046.


rainersigwald:
@AndyGerlicher I think this is what was asked for over email this morning.

PR #1154, opened 2016-10-04
Log xunit test cases as they run

rainersigwald:


dir.targets
```diff
@@ -88,11 +88,11 @@
...
 
     <!-- Run tests on the full framework -->
-    <Exec Command="$(ToolPackagesDir)\xunit.runner.console\$(XunitVersion)\tools\xunit.console.x86.exe @(AssemblyUnderTest, ' ') $(XunitOptions) -noshadow -parallel none -xml %(AssemblyUnderTest.FullPath)_TestResults.xml -html %(AssemblyUnderTest.FullPath)_TestResults.html > %(AssemblyUnderTest.FullPath)_stdout.txt"
+    <Exec Command="$(ToolPackagesDir)\xunit.runner.console\$(XunitVersion)\tools\xunit.console.x86.exe @(AssemblyUnderTest, ' ') $(XunitOptions) -noshadow -parallel none -xml %(AssemblyUnderTest.FullPath)_TestResults.xml -html %(AssemblyUnderTest.FullPath)_TestResults.html -verbose > %(AssemblyUnderTest.FullPath)_stdout.txt"
           Condition="'$(NetCoreBuild)' != 'true'" />
 
     <!-- Run tests using coreclr.  Ignore warnings/errors generated by tests to stderr. -->
-    <Exec Command="$(DotnetToolCommand) $(TestDeploymentDir)xunit.console.netcore.exe @(AssemblyUnderTest, ' ') $(XunitOptions) -noshadow -parallel none -notrait &quot;CrashesOnNetCore=true&quot; -xml %(AssemblyUnderTest.FullPath)_core_TestResults.xml $(XunitOptions) > %(AssemblyUnderTest.FullPath)_core_stdout.txt"
```
cdmihai: Why stop recording the stdout, now with enriched xunit test progress?

rainersigwald: Current theory from @jeffkl is that the redirection is what causes the hang. See https://github.com/Microsoft/msbuild/pull/1159.

jeffkl: The stdout is still captured by Jenkins so we won't lose anything

Issue #1162, opened 2016-10-04
MSBuild's ToolLocationHelper should support UAP SDKs with versioned content.


mirkomuggli:
UAP Sdk's starting with RS2 will support side-by-side installation. ToolLocationHelper is used throughout the UAP pipeline to find SDK content and needs to be able to point at specific versions of said content.
This will be achieved via a flag in the platform.xml.

I've already completed this change, just creating the issue to track it :)

rainersigwald:
Pretty sure this is long done--please correct me if I'm wrong @mirkomuggli.

PR #1164, opened 2016-10-05
adding support for versioned SDK content in ToolLocationHelper

mirkomuggli:
Add support for the new Platform.xml value "VersionedContent" (boolean).
If present and true, this flag instructs ToolLocationHelper to retrieve References from a versioned folder. SDK team will make the appropriate changes to Platform.xml, UAP.props, and UAP.arm.props to ensure other content paths are updated appropriately.


mirkomuggli:
This work addresses Issue #1162

PR #1166, opened 2016-10-05
FlatTrackingData ctor with TreatRootMarkersAsEntries

rainersigwald:
This allows avoiding costly operations in the C++ build.

Submitting on behalf of @olgaark.

### DIFF
--- a/src/Utilities/TrackedDependencies/FlatTrackingData.cs
+++ b/src/Utilities/TrackedDependencies/FlatTrackingData.cs
@@ -247,6 +247,28 @@ namespace Microsoft.Build.Utilities
             InternalConstruct(null, tlogFiles, tlogFilesToIgnore, false, missingFileTimeUtc, excludedInputPaths);
         }
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="tlogFiles">The .tlog files to interpret</param>
+        /// <param name="tlogFilesToIgnore">The .tlog files to ignore</param>
+        /// <param name="missingFileTimeUtc">The DateTime that should be recorded for missing file.</param>
+        /// <param name="excludedInputPaths">The set of paths that contain files that are to be ignored during up to date check, including any subdirectories.</param>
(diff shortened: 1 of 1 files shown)


src/Utilities/TrackedDependencies/FlatTrackingData.cs
```diff
@@ -462,7 +484,7 @@ private void ConstructFileTable()
                                 tlogEntry = tlog.ReadLine();
                                 continue;
                             }
-                            else if (tlogEntry[0] == '^' && TreatRootMarkersAsEntries) // This is a rooting record, and we should keep it
+                            else if (tlogEntry[0] == '^' && TreatRootMarkersAsEntries && tlogEntry.IndexOf('|') < 0) // This is a rooting non composite record, and we should keep it
```
rainersigwald: @olgaark Can you remind me what it would mean to be a composite entry?

olgaark: root marker with many files:
^E:\VCPKG\TOOLSRC\VCPKG\X64\DEBUG\COMMANDS_CACHE.OBJ|E:\VCPKG\TOOLSRC\VCPKG\X64\DEBUG\COMMANDS_CREATE.OBJ|E:\VCPKG\TOOLSRC\VCPKG\X64\DEBUG\COMMANDS_EDIT.OBJ

rainersigwald: So is it right to drop these? They shouldn't be split and added individually?

olgaark: And we don't need to treat them as entries because for the tools which have them the sources are naturally listed as inputs.

rainersigwald: Cool, thanks.

rainersigwald:
I can't approve "my own" review, but :shipit:. @Microsoft/msbuild-maintainers ?

olgaark:
Hi Rainer,
I don’t see this checkin go through. What do we still need to be able to commit it? Can I help somehow?
Thanks,
Olga

PR #1173, opened 2016-10-07
Allow MSBuild to query VS setup for toolset

### LINKED ISSUE #1161: On a machine with VS "15" installed, we should find the VS Toolset
Historically, MSBuild was in the GAC and toolsets were located via the registry. This meant that an application that used MSBuild APIs a) didn't have to ship MSBuild assemblies and b) could load and build any project that worked in VS, because MSBuild found the installed toolset.

This is no longer the case with Visual Studio "15". Now MSBuild doesn't have a registry impact and isn't necessarily in a known path. That means that even if you ship `Microsoft.Build.dll`, you must _also_ ship a full copy of tasks and targets--and there's no way of knowing what "full" means, since VSIXes can deliver arbitrary targets.

Ideally, an MSBuild caller could get access to the VS toolset for free.

AndyGerlicher:
This change allows for MSBuild to query the Visual Studio setup COM API (via managed interop) for instances of Visual Studio. When running on a machine with Visual Studio "15" this allows for a user to invoke the Build API without additional context (i.e. generic command window).

Also changed the behavior of the static initialization to not throw a TypeInitializationException on failure to find MSBuild.

Fixes #1161

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -5,7 +5,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Threading;
+using System.Collections.Generic;
 
 namespace Microsoft.Build.Shared
 {
@@ -51,15 +51,7 @@ namespace Microsoft.Build.Shared
 
             // Check to see if we're running inside of Visual Studio
             bool runningInVisualStudio;
-            var visualStudioPath = FindVisualStudio(
-                new[] {processNameCommandLine, processNameCurrentProcess},
-                out runningInVisualStudio);
-
-            string msbuildFromVisualStudioRoot = null;
-            if (!string.IsNullOrEmpty(visualStudioPath))
-            {
-                msbuildFromVisualStudioRoot = FileUtilities.CombinePaths(visualStudioPath, "MSBuild", "15.0", "Bin");
-            }
+            var visualStudioPath = GetVisualStudioPath(out runningInVisualStudio);
 
             var possibleLocations = new Func<BuildEnvironment>[]
             {
@@ -74,7 +66,7 @@ namespace Microsoft.Build.Shared
                 () => TryFromFolder(Path.GetDirectoryName(executingAssembly), runningTests, runningInVisualStudio, visualStudioPath),
 
                 // Try based on the Visual Studio Root
-                ()=> TryFromFolder(msbuildFromVisualStudioRoot, runningTests, runningInVisualStudio, visualStudioPath),
+                ()=> TryFromVisualStudioRoot(visualStudioPath, runningTests, runningInVisualStudio),
 
                 // Try from the current directory
                 () => TryFromFolder(currentDirectory, runningTests, runningInVisualStudio, visualStudioPath)
@@ -89,6 +81,14 @@ namespace Microsoft.Build.Shared
 
             ErrorUtilities.ThrowInvalidOperation("Shared.CanNotFindValidMSBuildLocation");
             return null; // Not reachable
+        }
+
+        private static BuildEnvironment TryFromVisualStudioRoot(string visualStudioPath, bool runningTests, bool runningInVisualStudio)
+        {
+            if (string.IsNullOrEmpty(visualStudioPath)) return null;
+
+            var msbuildFromVisualStudioRoot = FileUtilities.CombinePaths(visualStudioPath, "MSBuild", "15.0", "Bin");
+            return TryFromFolder(msbuildFromVisualStudioRoot, runningTests, runningInVisualStudio, visualStudioPath);
         }
 
         private static BuildEnvironment TryFromCurrentProcess(string runningProcess, bool runningTests, bool runningInVisualStudio, string visualStudioPath)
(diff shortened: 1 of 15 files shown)


AndyGerlicher:
I definitely don't like how many new references were added (from NuGet). But I don't really see a way around that since this is "shared" code.

src/dir.targets
```diff
@@ -27,4 +27,13 @@
     <Touch Files="$(RestorePackagesSemaphore)" AlwaysCreate="true" />
   </Target>
 
+  <!-- Taken from Microsoft.VisualStudio.SDK.EmbedInteropTypes NuGet package which tries to fix this issue
+       (but doesn't include this assembly in their list).  -->
+  <Target Name="LinkVSEmbeddableAssemblies" DependsOnTargets="ResolveReferences" AfterTargets="ResolveReferences">
```
rainersigwald: Why `DependsOnTargets`?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -91,6 +103,14 @@ private static BuildEnvironment Initialize()
             return null; // Not reachable
         }
 
+        private static BuildEnvironment TryFromVisualStudioRoot(string visualStudioPath, bool runningTests, bool runningInVisualStudio)
+        {
+            if (string.IsNullOrEmpty(visualStudioPath)) return null;
+
+            var msbuildFromVisualStudioRoot = FileUtilities.CombinePaths(visualStudioPath, "MSBuild", "15.0", "Bin");
```
rainersigwald: >"15.0" [](start = 102, length = 6)

`CurrentVisualStudioVersion` instead of `15.0`?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -143,25 +163,43 @@ private static bool IsValidMSBuildPath(string path)
...
+                () => TryGetVsFromInstalled(),
+                () => TryGetVsFromMSBuildLocation(processNameCommandLine),
+                () => TryGetVsFromMSBuildLocation(processNameCurrentProcess)
+            };
 
-                    // This assumes running from VS\Common7\IDE\<process>.exe.
-                    return FileUtilities.GetFolderAbove(process, 3);
-                }
-            }
+            return possibleLocations.Select(location => location()).FirstOrDefault(path => path != null);
```
rainersigwald: `FirstOrDefault(string.IsNullOrEmpty)`?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -174,32 +212,46 @@ private static string FindVisualStudio(string[] processNames, out bool runningIn
                 return vsInstallDir;
             }
 
+            return null;
+        }
+
+        private static string TryGetVsFromInstalled()
+        {
+            var instances = s_getVisualStudioInstances();
+            Version v = new Version("15.0");
```
rainersigwald: named const again

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -174,32 +212,46 @@ private static string FindVisualStudio(string[] processNames, out bool runningIn
...
+
+            // Get the first instance of Visual Studio that matches our Major/Minor compatible version
```
rainersigwald: Message about resolving ambiguity by picking the first?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -253,9 +307,15 @@ private static string GetCurrentDirectory()
         private static Func<string> s_getProcessFromRunningProcess = GetProcessFromRunningProcess;
         private static Func<string> s_getExecutingAssmblyPath = GetExecutingAssmblyPath;
         private static Func<string> s_getCurrentDirectory = GetCurrentDirectory;
+        private static Func<IEnumerable<VisualStudioInstance>> s_getVisualStudioInstances = VisualStudioLocationHelper.GetInstances;
 
         private static class BuildEnvironmentHelperSingleton
         {
+            // Explicit static constructor to tell C# compiler
+            // not to mark type as beforefieldinit
```
rainersigwald: why is this helpful?

src/Shared/VisualStudioLocationHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            Name = name;
+            Path = path;
+            Version = version;
+        }
+    }
+
+    internal class VisualStudioLocationHelper
+    {
+        private const int REGDB_E_CLASSNOTREG = unchecked((int)0x80040154);
```
rainersigwald: Comment, please :)

src/Shared/VisualStudioLocationHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+                var instances = new ISetupInstance[1];
+                do
+                {
+                    e.Next(1, instances, out fetched);
+                    if (fetched > 0)
+                    {
+                        var instance = PrintInstance(instances[0], helper);
+                        if (instance != null)
+                        {
+                            validInstances.Add(instance);
```
rainersigwald: If you're going to return `IEnumerable`, should you `yield return` here?

src/Shared/VisualStudioLocationHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+                        }
+                    }
+                }
+                while (fetched > 0);
+
+                return validInstances;
+            }
+            catch (Exception)
```
rainersigwald: :-1:

src/Shared/VisualStudioLocationHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        private static ISetupConfiguration GetQuery()
+        {
+            try
+            {
+                // Try to CoCreate the class object.
+                return new SetupConfiguration();
+            }
+
+            catch (COMException ex) when (ex.ErrorCode == REGDB_E_CLASSNOTREG)
```
rainersigwald: nit: extra blank line

src/Shared/VisualStudioLocationHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+    internal class VisualStudioLocationHelper
+    {
+        private const int REGDB_E_CLASSNOTREG = unchecked((int)0x80040154);
+
+        /// <summary>
+        /// Query the Visual Studio setup API to get instances of Visual Studio installed
+        /// on the machine. Will not include anything before Visual Studio "15".
+        /// </summary>
+        /// <returns>Enumerable list of Visual Studio instances</returns>
+        internal static IEnumerable<VisualStudioInstance> GetInstances()
```
rainersigwald: This generally seems a bit more complicated than I'd hoped, and it isn't super well commented. Is this as easy as it gets?

src/dir.targets
AndyGerlicher: I copied/pasted from that NuGet package I mentioned and updated the file name to match. Ideally NuGet would support this feature and we could remove this :(

Issue #1010, opened 2016-09-09
.NET Core: tasks built against v14 contracts don't load on v15


natemcmaster:
**Problem**
Trying to build a task using the Microsoft.Build.Utilities.Core, 14.0.0.0 contracts that will run on MSBuild 14 or 15. This works on .NET Framework version of MSBuild. The .NET Core version throws when loading the task.

**Repro**
Create a task compiled to netstandard1.3 using https://www.nuget.org/packages/Microsoft.Build.Utilities.Core/14.3.0-prerelease02
Import task via `<UsingTask>`

**Error**

> Microsoft.AspNetCore.Server.IISIntegration.Tasks.targets(27,9): error MSB4062: The "Microsoft.AspNetCore.Server.IISIntegration.Tasks.TransformWebConfig" task could not be loaded from the assembly build\netstandard1.5\Microsoft.AspNetCore.Server.IISIntegration.Tasks.dll. Could not load file or assembly 'Microsoft.Build.Utilities.Core, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Could not find or load a specific file. (Exception from HRESULT: 0x80131621) Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.

**Setup Info**
Using version of MSBuild in .NET Core SDK 1.0.0-preview3-003585

jeffkl:
Are you running this under MSBuild.exe?  We rely on assembly redirects in our [app.config](https://github.com/Microsoft/msbuild/blob/master/src/XMakeCommandLine/app.config).  Or is this in your own app?

natemcmaster:
I'm using `dotnet.exe build3`.

jeffkl:
Can you manually create a dotnet.exe.config next to dotnet.exe that contains this:

``` xml
<configuration>
    <runtime>
      <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
        <dependentAssembly>
          <assemblyIdentity name="Microsoft.Build.Framework" culture="neutral" publicKeyToken="b03f5f7f11d50a3a" />
          <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.1.0.0" />
        </dependentAssembly>
        <dependentAssembly>
[...]

PR #1175, opened 2016-10-07
Update assembly loader to account for version numbers

tmeschter:
When the default `AssemblyLoadContext` cannot find a dependency of a task assembly (because it does not know where to look) it fires the `Resolving` event, and uses whatever `Assembly` object is returned. Our implementation of the handler simply looks next to each known task assembly and loads and returns the first assembly with a matching short name.

This generally works fine but can be a problem if there are two or more task assemblies that depend on assemblies with the same short name but different versions--the task may ask for version 2.0.0.0, but end up with 1.0.0.0 instead. If it then tries to access an API that does not exist in 1.0.0.0 you'll get an exception.

The fix for this issue is to check the version of a candidate assembly before loading it. If it doesn't match, we simply continue the search until we find one that does (or we run out of candidates).

Related to this is the problem of tasks that were built against an older version of MSBuild. Even if a task was built against 14.0, for example, we want it to use the current version of the MSBuild binaries that ship with msbuild.exe. Since .NET Core does not support binding redirects via config files, we need to handle this ourselves in our `Resolve` handler. If the short name matches one of a well-known set, we update the `AssemblyName` of the requested assembly to use the current version number, and then call back to the default `AssemblyLoadContext` to find it.

### DIFF
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -22,6 +22,15 @@ namespace Microsoft.Build.Shared
         private bool _resolvingHandlerHookedUp = false;
 
         private static readonly string[] _extensions = new[] { "ni.dll", "ni.exe", "dll", "exe" };
+        private static readonly Version _currentAssemblyVersion = new Version(Microsoft.Build.Shared.MSBuildConstants.CurrentAssemblyVersion);
+        private static readonly HashSet<string> _wellKnownAssemblyNames = new HashSet<string>(
+            new[]
+            {
+                "Microsoft.Build",
+                "Microsoft.Build.Framework",
+                "Microsoft.Build.Tasks.Core",
+                "Microsoft.Build.Utilities.Core"
+            });
 
         public void AddDependencyLocation(string fullPath)
         {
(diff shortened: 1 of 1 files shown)


tmeschter:
@rainersigwald @cdmihai @AndyGerlicher @jeffkl Could you take a look at this, please?

tmeschter:
Related to #1010.

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -63,11 +72,29 @@ public Assembly LoadFromPath(string fullPath)
             }
         }
 
+        private Assembly TryGetWellKnownAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
```
rainersigwald: I find a method named `Try...` that doesn't return a `bool` a bit surprising in C#.

rainersigwald: Hah, but I didn't complain about `TryResolveAssembly` below. Consistency is good.

rainersigwald:
also related to #750, right?

tmeschter:
@dotnet-bot retest OSX Build for CoreCLR please

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -63,11 +72,29 @@ public Assembly LoadFromPath(string fullPath)
...
+        {
+            if (!_wellKnownAssemblyNames.Contains(assemblyName.Name))
+            {
+                return null;
+            }
+
+            assemblyName.Version = _currentAssemblyVersion;
+
+            return AssemblyLoadContext.Default.LoadFromAssemblyName(assemblyName);
```
cdmihai: Is it possible for this call to ever fail? If yes, then mutating the above assemblyName instead of cloning it might cause weird issues, since the mutation flows back into the callers.

cdmihai: Or is AssemblyName a struct? :)

cdmihai:
Is a test for the version worth it? :)

tmeschter:
@dotnet-bot retest OSX Build for CoreCLR please

src/Shared/CoreCLRAssemblyLoader.cs
tmeschter: @cdmihai If that call fails then something has gone seriously wrong with your app--most likely you're missing core, required MSBuild binaries. I'm not particularly worried about the behavior in that case.

tmeschter:
@cdmihai If I can create such a test without too much work, yes, I think it will be worth doing. I want to get this through and hopefully unblock dotnet CLI while I do that in parallel.

PR #1176, opened 2016-10-07
Merge master ->xplat

AndyGerlicher:
The main conflict was in `dir.props` which Mihai helped me with. We think it's right.


AndyGerlicher:
FYI added NuGet reference to `System.Linq.Parallel` to fix the CI.

PR #1178, opened 2016-10-10
Fix-up xplat desktop build.

AndyGerlicher:
- Change setup binpath to end in 'Output'
 - Remove #if around STATIC_VERSION_NUMBER. Visual Studio version was not
   defined in that way.
 - Move System.Collections.Immutable to NuGet reference of the correct
   version.

### DIFF
--- a/setup/Microsoft.Build.swixproj
+++ b/setup/Microsoft.Build.swixproj
@@ -20,8 +20,8 @@
   <PropertyGroup>
     <OutDir>$(BaseOutputPath)setup\</OutDir>
     <OutputPath>$(BaseOutputPath)setup\</OutputPath>
-    <X86BinPath>$(BaseOutputPath)X86\$(OS)\$(Configuration)\</X86BinPath>
-    <X64BinPath>$(BaseOutputPath)X64\$(OS)\$(Configuration)\</X64BinPath>
+    <X86BinPath>$(BaseOutputPath)X86\$(OS)\$(Configuration)\Output\</X86BinPath>
+    <X64BinPath>$(BaseOutputPath)X64\$(OS)\$(Configuration)\Output\</X64BinPath>
   </PropertyGroup>
 
   <!-- Needed for the MicroBuild signing plugin -->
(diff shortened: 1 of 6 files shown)


AndyGerlicher:
@rainersigwald let me know in the morning if you have any feedback I will address. Merging so I can do an official build.

src/Shared/Constants.cs
```diff
@@ -25,9 +25,7 @@ internal static class MSBuildConstants
         /// The most current Visual Studio Version known to this version of MSBuild. 
         /// </summary>
 #if STANDALONEBUILD
-#if STATIC_VERSION_NUMBER
         internal const string CurrentVisualStudioVersion = "15.0";
-#endif
 #else
         internal const string CurrentVisualStudioVersion = Microsoft.VisualStudio.Internal.BrandNames.VSGeneralVersion;
```
rainersigwald: We should eventually get rid of this.

setup/Microsoft.Build.swixproj
```diff
@@ -20,8 +20,8 @@
   <PropertyGroup>
     <OutDir>$(BaseOutputPath)setup\</OutDir>
     <OutputPath>$(BaseOutputPath)setup\</OutputPath>
-    <X86BinPath>$(BaseOutputPath)X86\$(OS)\$(Configuration)\</X86BinPath>
-    <X64BinPath>$(BaseOutputPath)X64\$(OS)\$(Configuration)\</X64BinPath>
+    <X86BinPath>$(BaseOutputPath)X86\$(OS)\$(Configuration)\Output\</X86BinPath>
+    <X64BinPath>$(BaseOutputPath)X64\$(OS)\$(Configuration)\Output\</X64BinPath>
```
rainersigwald: What's the motivation here? It seems fine but I don't understand why it's necessary.

src/XMakeTasks/Microsoft.Build.Tasks.csproj
```diff
@@ -952,7 +952,6 @@
       <HintPath>$(CompilerToolsDir)\Microsoft.Build.Tasks.CodeAnalysis.dll</HintPath>
     </Reference>
     <Reference Include="System" />
-    <Reference Include="System.Collections.Immutable" />
```
rainersigwald: Ooooohhh!

PR #1180, opened 2016-10-11
Fix issues in xplat project

mfilippov:
Fix some issues in xplat branch.


src/.nuget/NuGet.Config
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rainersigwald: This file was intentionally deleted in 1ea71faaa6d71d88e37b3f98ff93f379b8707636. I think it'd be better to update the solution item to point to the new file in the root of the repo.

src/xunit.runner.json
```diff
@@ -0,0 +1,6 @@
+{
```
rainersigwald: This file moved to `targets/xunit.runner.json`--can you update the `.sln` to point to that location?

jeffkl:

src/xunit.runner.json
mfilippov: No problem I do it.

mfilippov:
@rainersigwald I fix sln file and force push.

rainersigwald:
Thanks!

PR #1181, opened 2016-10-11
Merge master -> xplat

AndyGerlicher:
First attempt, builds on my machine. Checking it with CI. This was fairly hard to merge. Whoever committed `68b9672` did not make it easy!

Most of the complication was small things in `project.json` and areas that changed both in xplat and in master (`BuildEnvironmentHelper.cs`).

### DIFF
--- a/src/XMakeCommandLine/MSBuild.csproj
+++ b/src/XMakeCommandLine/MSBuild.csproj
@@ -114,6 +114,9 @@
     <Compile Include="..\Shared\TempFileUtilities.cs">
       <Link>TempFileUtilities.cs</Link>
     </Compile>
+    <Compile Include="..\Shared\VisualStudioLocationHelper.cs">
+      <Link>VisualStudioLocationHelper.cs</Link>
+    </Compile>
     <Compile Include="..\Shared\XMakeAttributes.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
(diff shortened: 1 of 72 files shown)


src/XMakeCommandLine/MSBuild.csproj
```diff
@@ -246,6 +245,10 @@
       <SubType>Designer</SubType>
     </None>
     <None Include="MSBuild.exe.manifest" />
+    <FilesToSign Include="$(OutDir)\$(AssemblyName)$(TargetExt)">
+      <Authenticode>Microsoft</Authenticode>
+      <StrongName>StrongName</StrongName>
+    </FilesToSign>
```
jeffkl: This should be left out of the merge because in xplat we did this in a .targets file

src/XMakeTasks/project.json
```diff
@@ -1,9 +1,13 @@
 ﻿{
   "dependencies": {
-    "System.Collections.Immutable": "1.2.0"
+    "System.Collections.Immutable": "1.2.0",
```
jeffkl: >, [](start = 43, length = 1)

I think trailing commas mess up project.json?

jeffkl:
:shipit:

src/XMakeTasks/project.json
AndyGerlicher: I don't think it does, but it does make editors happier. Fixed it locally.

rainersigwald: It doesn't with Newtonsoft.Json but it's technically nonconformant so this should be fixed (see also #1180).

PR #1182, opened 2016-10-11
Rename pt to pt-BR

cdmihai:


nschonni:
If you're using CLDR, then shouldn't this stay pt, but adding in a pt-PT? https://github.com/jquery/globalize#locales

cdmihai:
@nschonni 

I asked around and using `pt-br` is the standard code we are using internally for the Brazilian Portuguese translation.

PR #1187, opened 2016-10-12
Ignore whitespace text nodes when parsing projects

### LINKED ISSUE #270: LargeElementLocation tests fail on CoreCLR
It looks like the CoreFX version of System.XML is a bit more aggressive about what constitutes a Text node in XML.  The very large strings of whitespace embedded in files in these tests in `src\XMakeBuildEngine\UnitTests\Construction\ElementLocation_Tests.cs` show up as unexpected Text nodes, causing the `InvalidProjectException` to be thrown while pointing to a different location.

This may be a CoreFX problem, but for now I'll just disable the tests.

jeffkl:
This just ignores purely whitespace text nodes.  The XML reader only looks at the first 4k of text to determine if it should be ignored.  Our tests are generating 70k of whitespace so the XML reader gives up after 4k and assumes its a text node.

Closes #270

Fixes test failures in #1004 when this change is ported to master.

### DIFF
--- a/src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
+++ b/src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
@@ -48,6 +48,13 @@ namespace Microsoft.Build.Internal
                         VerifyThrowProjectValidNamespace(childElement);
                         children.Add(childElement);
                         break;
+                    case XmlNodeType.Text:
+                        // Whitespace is read as a #text element so if the node is purely whitespace, just ignore it.
+                        if (!String.IsNullOrWhiteSpace(child.InnerText) && throwForInvalidNodeTypes)
+                        {
+                            ThrowProjectInvalidChildElement(child.Name, element.Name, element.Location);
+                        }
+                        break;
 
                     default:
                         if (throwForInvalidNodeTypes)
--- a/src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
+++ b/src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
@@ -48,15 +48,15 @@ namespace Microsoft.Build.Internal
                         VerifyThrowProjectValidNamespace(childElement);
                         children.Add(childElement);
                         break;
-                    case XmlNodeType.Text:
-                        // Whitespace is read as a #text element so if the node is purely whitespace, just ignore it.
-                        if (!String.IsNullOrWhiteSpace(child.InnerText) && throwForInvalidNodeTypes)
-                        {
-                            ThrowProjectInvalidChildElement(child.Name, element.Name, element.Location);
-                        }
-                        break;
 
                     default:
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
```diff
@@ -48,6 +48,13 @@ private static List<XmlElementWithLocation> GetChildElements(XmlElementWithLocat
                         VerifyThrowProjectValidNamespace(childElement);
                         children.Add(childElement);
                         break;
+                    case XmlNodeType.Text:
+                        // Whitespace is read as a #text element so if the node is purely whitespace, just ignore it.
```
rainersigwald: It's odd that we're getting a whitespace-only text node when it should come back as a Whitespace node. But the reason is that there's a limit in `XmlTextReaderImpl`'s determine-node-type code: if it's [greater than 4kB of whitespace](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Private.Xml/src/System/Xml/Core/XmlTextReaderImpl.cs#L5571-L5586), it'll return as `Text`.

But given that it works this way, this check sounds good.

jeffkl:
@rainersigwald figured out the problem.  The XML parser only looks at the first 4k of text to see if it's whitespace or not.  Since our whitespace is 70k, it gives up and assumes it's text.  
https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Private.Xml/src/System/Xml/Core/XmlTextReaderImpl.cs#L5571

src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
```diff
@@ -48,6 +48,13 @@ private static List<XmlElementWithLocation> GetChildElements(XmlElementWithLocat
...
+                        if (!String.IsNullOrWhiteSpace(child.InnerText) && throwForInvalidNodeTypes)
+                        {
+                            ThrowProjectInvalidChildElement(child.Name, element.Name, element.Location);
+                        }
```
rainersigwald: I'd prefer

```C#
if (string.IsNullOrWhitespace(child.InnerText)) {
    // long whitespace comes back as text, see [link]
    break;
}
goto default;
```

and rely on the `default` case, instead of having to check whether or not to throw and throwing in this case.
with a fall

src/XMakeBuildEngine/Xml/ProjectXmlUtilities.cs
```diff
@@ -48,8 +48,15 @@ private static List<XmlElementWithLocation> GetChildElements(XmlElementWithLocat
                         VerifyThrowProjectValidNamespace(childElement);
                         children.Add(childElement);
                         break;
-
+
```
jeffkl: AHH more whitespace, i'll remove this. #whitespacematters

PR #1192, opened 2016-10-12
Cache Item computation

cdmihai:
On the test repo (solution with 72 VCX projects) this reduces evaluation by ~120ms. The improvement should get significantly better the more cross item references there are.

Items can cross-reference each other during evaluation via:
- item references (`Include="@(a)"`)
- conditions on item groups, items, and metadata (`Condition="@(a->Count()) != 0"`)
- metadata item references (`<m>@(a)</m>`)

In terms of memory, the cache pressure isn't that large because many other allocations are prevented by not doing the repeated computations, so the more cross item references there are, the more the cache pays for itself

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
+++ b/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
@@ -58,6 +58,8 @@ namespace Microsoft.Build.Evaluation
                     }
                 }
 
+                ISet<string> excludePatternsForGlobs = null;
+
                 foreach (var fragment in _itemSpec.Fragments)
                 {
                     if (fragment is ItemExpressionFragment<P, I>)
@@ -91,11 +93,20 @@ namespace Microsoft.Build.Evaluation
                     else if (fragment is GlobFragment)
                     {
                         string glob = ((GlobFragment)fragment).ItemSpecFragment;
+
+                        if (excludePatternsForGlobs == null)
+                        {
+                            excludePatternsForGlobs = BuildExcludePatternsForGlobs(globsToIgnore, excludePatterns);
+                        }
+
                         string[] includeSplitFilesEscaped = EngineFileUtilities.GetFileListEscaped(
                             _rootDirectory,
                             glob,
-                            excludePatterns.Count > 0 ? (IEnumerable<string>) excludePatterns.Concat(globsToIgnore) : globsToIgnore
+                            excludePatternsForGlobs
                             );
+
+                        // itemsToAdd might grow 0 or more times during the following iteration. Proactively increase its capacity to ensure only one growth happens
+                        IncreaseListCapacityIfNecessary(itemsToAdd, includeSplitFilesEscaped.Length);
 
                         foreach (string includeSplitFileEscaped in includeSplitFilesEscaped)
                         {
@@ -111,6 +122,29 @@ namespace Microsoft.Build.Evaluation
                 return itemsToAdd;
             }
 
+            private static ISet<string> BuildExcludePatternsForGlobs(ImmutableHashSet<string> globsToIgnore, ImmutableList<string>.Builder excludePatterns)
+            {
+                var anyExcludes = excludePatterns.Count > 0;
+                var anyGlobstoIgnore = globsToIgnore.Count > 0;
+
+                if (anyGlobstoIgnore && anyExcludes)
+                {
+                    return excludePatterns.Concat(globsToIgnore).ToImmutableHashSet();
+                }
+
+                return anyExcludes ? excludePatterns.ToImmutableHashSet() : globsToIgnore;
+            }
+
+            private void IncreaseListCapacityIfNecessary(List<I> list, int itemsToAdd)
+            {
+                var newLength = list.Count + itemsToAdd;
+
+                if (list.Capacity < newLength)
+                {
+                    list.Capacity = newLength;
+                }
+            }
+
             protected override void MutateItems(ICollection<I> items)
             {
                 DecorateItemsWithMetadata(items, _metadata);
@@ -118,7 +152,10 @@ namespace Microsoft.Build.Evaluation
 
             protected override void SaveItems(ICollection<I> items, ImmutableList<ItemData>.Builder listBuilder)
             {
-                listBuilder.AddRange(items.Select(item => new ItemData(item, _elementOrder, _conditionResult)));
+                foreach (var item in items)
+                {
+                    listBuilder.Add(new ItemData(item, _elementOrder, _conditionResult));
+                }
             }
         }
 
--- a/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
+++ b/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
@@ -154,7 +154,7 @@ namespace Microsoft.Build.Evaluation
             {
                 foreach (var item in items)
                 {
-                    listBuilder.Add(new ItemData(item, _elementOrder, _conditionResult));
+                    listBuilder.Add(new ItemData(item, _itemElement, _elementOrder, _conditionResult));
                 }
             }
         }
--- a/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
+++ b/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
@@ -11,13 +11,14 @@ namespace Microsoft.Build.Evaluation
 {
     internal partial class LazyItemEvaluator<P, I, M, D>
     {
-        abstract class LazyItemOperation
+        private abstract class LazyItemOperation : IItemOperation
         {
+            private readonly string _itemType;
+            private readonly ImmutableDictionary<string, LazyItemList> _referencedItemLists;
+            private readonly LazyItemEvaluator<P, I, M, D> _lazyEvaluator;
+
             protected readonly ProjectItemElement _itemElement;
-            protected readonly string _itemType;
             protected readonly ItemSpec<P, I> _itemSpec;
-            protected readonly ImmutableDictionary<string, LazyItemList> _referencedItemLists;
-            protected readonly LazyItemEvaluator<P, I, M, D> _lazyEvaluator;
             protected readonly EvaluatorData _evaluatorData;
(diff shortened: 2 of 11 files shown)


rainersigwald:
I am looking at this but I have passed the part of the day where I can reason about the correctness of the lookahead remove operations. Tomorrow!

cdmihai:
@dsplaisted

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -76,6 +61,22 @@ protected virtual ICollection<I> SelectItems(ImmutableList<ItemData>.Builder lis
 
             protected virtual void SaveItems(ICollection<I> items, ImmutableList<ItemData>.Builder listBuilder) { }
 
+            private IList<I> GetReferencedItems(string itemType, ImmutableHashSet<string> globsToIgnore)
```
cdmihai: Must have been some manual refactoring left-over. I'll undo the move.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -133,34 +133,97 @@ public ItemData(I item, int elementOrder, bool conditionResult)
             public bool ConditionResult { get { return _conditionResult; } }
         }
 
+        private class OperationWithCache : IItemOperation
```
cdmihai: I like MemoizedOperation. It's just that it's not self-memoizing and needs an external caller to tell it when to cache.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -133,34 +133,97 @@ public ItemData(I item, int elementOrder, bool conditionResult)
...
 
             public ImmutableList<ItemData>.Builder GetItems(ImmutableHashSet<string> globsToIgnore)
             {
-                //  TODO: Check for cached results
+                // Cache results only on the operation whose results are required by an external caller (via GetItems)
+                // If the head of this LazyItemList is uncached, then the tail either:
+                //   - has all operations uncached (nobody needed their results until now). In this case we have to compute the head plus the entire tail,
+                // but we don't need to cache each node of the tail because no external caller wanted their results
+                //   - has some operations cached (somebody needed their result at some point). In this case we have to compute the head plus the part of the 
+                // tail up to the first cached operation.
```
cdmihai: Yes, that would be the worst case. It happens if you have this:

```xml
1: <C Include="a"/>
2: <ForcesCacheOnC Include="@(C)"/>
3: <C Include="b"/>
4: <ForcesCacheOnC Include="@(C)"/>
5: <C Include="c"/>
6: <ForcesCacheOnC Include="@(C)"/>
```

However, the caches will share structural state. So the cache at operation `3` will share common state with the previously made cache at operation `1` (shared state is the collection with item 'a'). This is done via the immutable list. Also, when `GetItems` is called on operation `3`, it won't redo operation `1` but use its cache.

So no computation gets ever redone, but we will have a lot of `OperationWithCache` decorators if the above pattern is very common.

We could do other fancy things like reference counting (assuming an external caller calls `GetItems` only once, then we could decrement some CalleeNumber on each call per operation, and then `null` the cache for that operation when it has satisfied all of it's external callers). But I don't think it's worth making it more complex at this point until we get an outside ask about it :)

rainersigwald: > I don't think it's worth making it more complex at this point

💯

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
```diff
@@ -111,14 +127,27 @@ protected override ICollection<I> SelectItems(ImmutableList<ItemData>.Builder li
...
                 DecorateItemsWithMetadata(items, _metadata);
             }
 
             protected override void SaveItems(ICollection<I> items, ImmutableList<ItemData>.Builder listBuilder)
             {
-                listBuilder.AddRange(items.Select(item => new ItemData(item, _elementOrder, _conditionResult)));
+                foreach (var item in items)
+                {
+                    listBuilder.Add(new ItemData(item, _elementOrder, _conditionResult));
+                }
```
rainersigwald: Ahh, because this method is called so often! That totally makes sense, thanks.

cdmihai:
@dsplaisted That is indeed a real issue :(

It is curious though that the test I had written exactly for this scenario did not fail: https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeBuildEngine/UnitTests/Evaluation/ItemEvaluation_Tests.cs#L91

I think it does not fail because the [Dictionary<string, LazyItemList>](https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs#L51) the lazy evaluator is using gets some happy key ordering which hides this issue. I'll play around with the test and try to get it failing.

The fix for the actual issue would be for Update to delete the old items and create new ones.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
```diff
@@ -58,6 +58,19 @@ protected override ICollection<I> SelectItems(ImmutableList<ItemData>.Builder li
                     }
                 }
 
+                var excludePatternsForGlobs = new Lazy<ISet<string>>(() =>
```
cdmihai: Why would that be an improvement? The lazy setting before the first access vs using the Lazy seem to be doing the same thing (except creating that extra object for Lazy)

cdmihai:
- updated the caching code to ensure that each operation is applied at most once. Before, if a list had two item references to it (from other itemspecs or from conditions), and if the newer reference executed first, then the older reference would have re-computed the shared state between the two operations.
- redid the Update operation so it clones items before it mutates them

cdmihai:
@dotnet-bot test this please

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.IncludeOperation.cs
dsplaisted: Yeah it avoids extra object creation for the lazy object itself and possibly the closure used for the creation delegate.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -136,28 +137,68 @@ public ItemData(I item, int elementOrder, bool conditionResult)
...
             {
                 _operation = operation;
-                _cache = new Lazy<Dictionary<ISet<string>, ImmutableList<ItemData>>>(
-                    () => new Dictionary<ISet<string>, ImmutableList<ItemData>>());
             }
 
             public void Apply(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
             {
+#if DEBUG
+                Debug.Assert(!_applyCalled, "Apply should only be called once. Otherwise caching is not working");
```
dsplaisted: It should be valid for `Apply` to be called multiple times if the `globsToIgnore` is different each time.

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -136,28 +137,68 @@ public ItemData(I item, int elementOrder, bool conditionResult)
...
+                }
+
+                items = null;
+                return false;
+            }
+
+            /// <summary>
+            /// Somebody is referencing this operation
+            /// </summary>
+            public void AddReference()
```
dsplaisted: It might be better to find a different name.  `AddReference` sounds to me like this is doing ref counting.

cdmihai:
@dotnet-bot test this please

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -174,12 +215,19 @@ public LazyItemList(LazyItemList previous, LazyItemOperation operation)
...
+                //   - Callers of GetItems who have announced ahead of time that they would reference an operation (via AddReference())
+                // This includes: item references (Include="@(foo)") and metadata conditions (Condition="@(foo->Count()) == 0")
+                // Without ahead of time notifications more computation is done than needed when the results of a future operation are requested 
+                // The future operation is part of another item list referencing this one (making this operation part of the tail).
+                // The future operation will compute this list but since no ahead of time notifications have been made by callers, it won't cache the
+                // intermediary operations that would be requested by those callers.
+                //   - Callers of GetItems that cannot announce ahead of time. This includes item referencing conditions on
+                // Item Groups and Item Elements. However, those conditions are performed eagerly outside of the LazyItemEvaluator, so they will run before
+                // any item referencing operations from inside the LazyItemEvaluator. This
```
dsplaisted: Dangling "This"

src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
```diff
@@ -117,50 +118,171 @@ private static string GetCurrentDirectoryForConditionEvaluation(ProjectElement e
...
+
+            public MemoizedOperation(IItemOperation operation)
+            {
+                _operation = operation;
+            }
+
+            public void Apply(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
+            {
+#if DEBUG
+                Debug.Assert(!_applyCalled, "Apply should only be called once. Otherwise caching is not working");
```
dsplaisted: It should be valid for this method to be called multiple times if `globsToIgnore` is different each time.

(I left this comment on an outdated version of the file, so repeating it here in case that doesn't show up).

cdmihai: Ah, forgot about that. The assert condition should be if the passed in `globsToIgnore ` is already cached. Also have to write a unit test to make this faulty assert fail :)

cdmihai:
@dsplaisted 

Fixed the assert to: number of calls should be equal to cache size.
Also fixed a hidden bug that I introduced: the memoized operation no longer cast to RemoveOperation, so no globs flowed to previous operations. Hard to add an assert for this. One way to add an assert is to change Remove so that it flows all of its values (literals, item references, and globs) not just globs, and Remove is skipped altogether. With this change, we can just throw in Remove's apply :)

PR #1194, opened 2016-10-14
Support for building on ubuntu 16.04

jeffkl:
* Detect the runtime in cibuild.sh
* Pass that along to our runtime dependencies instead of it being hard coded to ubuntu 14

Closes #833

PR #1195, opened 2016-10-14
Enable CI builds on Ubuntu 16.04

rainersigwald:
While doing so, I changed the name of Ubuntu14.04 to its more explicit
version.

@jeffkl (this will only pass after #1194)


jeffkl:
There's no way to test this right, once you check it in the build defs will be rebuilt.  If it fails, we can just roll it back?

rainersigwald:
Exactly. Let's give it a shot!

PR #1199, opened 2016-10-15
Make static regex compiled

cdmihai:
To make them run faster (.net blog says 30% faster). The increased Regex creation time should be shadowed by how many regex calls are made during evaluation (e.g. for each metadata reference) and execution.

Some regexes I could not make lazy because they are referenced from the MSBuildTaskHost which compiles against .net 3.5

### DIFF
--- a/src/Shared/FileUtilitiesRegex.cs
+++ b/src/Shared/FileUtilitiesRegex.cs
@@ -20,10 +20,12 @@ namespace Microsoft.Build.Shared
     internal static class FileUtilitiesRegex
     {
         // regular expression used to match file-specs beginning with "<drive letter>:" 
-        internal static readonly Regex DrivePattern = new Regex(@"^[A-Za-z]:");
+        internal static readonly Regex DrivePattern = new Regex(@"^[A-Za-z]:", RegexOptions.Compiled);
 
         // regular expression used to match UNC paths beginning with "\\<server>\<share>"
-        internal static readonly Regex UNCPattern = new Regex(String.Format(CultureInfo.InvariantCulture,
-            @"^[\{0}\{1}][\{0}\{1}][^\{0}\{1}]+[\{0}\{1}][^\{0}\{1}]+", Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar));
+        internal static readonly Regex UNCPattern =
+            new Regex(
+                string.Format(CultureInfo.InvariantCulture, @"^[\{0}\{1}][\{0}\{1}][^\{0}\{1}]+[\{0}\{1}][^\{0}\{1}]+",
+                    Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), RegexOptions.Compiled);
     }
 }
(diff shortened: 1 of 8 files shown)


src/Shared/FileUtilitiesRegex.cs
```diff
@@ -20,10 +20,12 @@ namespace Microsoft.Build.Shared
...
-        internal static readonly Regex DrivePattern = new Regex(@"^[A-Za-z]:");
+        internal static readonly Regex DrivePattern = new Regex(@"^[A-Za-z]:", RegexOptions.Compiled);
 
         // regular expression used to match UNC paths beginning with "\\<server>\<share>"
-        internal static readonly Regex UNCPattern = new Regex(String.Format(CultureInfo.InvariantCulture,
-            @"^[\{0}\{1}][\{0}\{1}][^\{0}\{1}]+[\{0}\{1}][^\{0}\{1}]+", Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar));
+        internal static readonly Regex UNCPattern =
+            new Regex(
+                string.Format(CultureInfo.InvariantCulture, @"^[\{0}\{1}][\{0}\{1}][^\{0}\{1}]+[\{0}\{1}][^\{0}\{1}]+",
+                    Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar), RegexOptions.Compiled);
```
rainersigwald: No need for laziness here?

src/XMakeBuildEngine/Evaluation/ConditionEvaluator.cs
```diff
@@ -64,52 +65,55 @@ internal static void UpdateConditionedPropertiesTable
                     bool lastPiece = pieceSeparator < 0;
                     int pieceEnd = lastPiece ? leftValue.Length : pieceSeparator;
 
-                    Match singlePropertyMatch = s_singlePropertyRegex.Match(leftValue, pieceStart, pieceEnd - pieceStart);
-
-                    if (singlePropertyMatch.Success)
+                    if (leftValue.StartsWith("$("))
```
rainersigwald: Does this have a noticeable impact? I want to believe that the regex engine can do it just as fast.

If it does, please comment that it's an optimization with some justification.

src/Shared/FileUtilitiesRegex.cs
cdmihai: That class is referenced from `MSBuildTaskHost` which compiles against a framework version that does not know about Lazy. Alternatives would be to duplicate the code (task host will never change™), do some ifdefs, or re-implement the caller code via String operations. None of those seems worth it.

rainersigwald: Ah! I agree.

src/XMakeBuildEngine/Evaluation/ConditionEvaluator.cs
cdmihai: You're right, The regex should hopefully fail quickly in this case since it does not hit any fancy matchers for the starting characters. I'll revert it.

cdmihai:
@rainersigwald for my test solution with 72 projects I didn't see a spectacular improvement (maybe ~20-30 ms shaved off). But it looks like a good idea either way. Most of those regexes hit many times during evaluation (per condition AST node interpretation, per metadata instance, etc)

PR #1200, opened 2016-10-17
Ubuntu16 CI

rainersigwald:


rainersigwald:
I'm sneaking this in since there's no way to test it.

PR #1201, opened 2016-10-17
Update CI badges in README

rainersigwald:
A couple were pointed to bogus locations, and all were pointed to the old domain rather than the new `ci.dot.net`. New markdown was copy/pasted from Jenkins' "Markdown (without view)" from each job's "Embeddable Build Status" link.


jeffkl:
Should we update the table for Ubuntu 14 and 16?

rainersigwald:
Yeah, since it's green that seems like a good idea :)

PR #1202, opened 2016-10-18
Make GR incremental without binary serialization

danmosemsft:
The cache file (which needs BS) is to store the list of files linked from the resx. For most resx not created with the Winforms designer, there are just strings in it and this cache is not needed. Even without the cache, the only effect is that build will skip if the main resx did not change but a linked file did.

Totally untested.


danmosemsft:
https://github.com/Microsoft/msbuild/issues/1197

rainersigwald:
Closing this in favor of #1251. Thanks for the initial idea, @danmosemsft.

PR #1209, opened 2016-10-18
Update NuGet package versioning to support wildcards.

### LINKED ISSUE #557: Versioning scheme for package, assembly, and file versions
On xplat, we use different version numbers for packages (e.g. 0.1.0-preview-00022-160326), and assemblies and file versions (e.g. 14.1.0.0).

We should come up with a consistent story where we use one common set of version number seeds to generate package versions, assembly versions, and file versions. We should also define what is the version number relationship between multiple targeted platforms (#556)

### LINKED ISSUE #1156: Cannot consume nightly builds using NuGet wildcard syntax
Problem: the versioning schema of MSBuild nightly builds breaks support for wildcards. The nightlies -- currently versioned as `15.1.$(BuildNumber)-preview5` -- cannot be referenced using `"15.1.*"` because the packages contain the "-preview5" preversion suffix. 

> Unable to resolve 'Microsoft.Build.Runtime (>= 15.1.0)'

This means we have to manually edit dependency versions for every pre-release build.

Suggestion:
Use the versioning format used by .NET Core. `15.1.0-preview5-$(BuildNumber)`. This would allow consumers to reference prerelease versions using wildcards `"15.1.0-*"`.

jeffkl:
Set build number to 0
Use git commit height and build revision as a suffix to the prerelease tag

New version will be something like: 15.1.0-preview-000333-01

Related to #557 as it address the NuGet versions.
Closes #1156


build/NuGetPackages/CreateNuGetPackages.proj
```diff
@@ -58,5 +58,32 @@
 
     <Message Importance="High" Text="@(NuSpecs->'%(Filename)') NuGet Package -> $(PackagesOutDir)\@(NuSpecs->'%(Filename)').$(NuGetPackageVersion).nupkg" />
   </Target>
+
+<Target Name="GetNuGetPackageVersionEx" DependsOnTargets="GetNuGetPackageVersion">
+  
+  <PropertyGroup>
+    <_BuildRevisionRegEx>\d{8}\.(?&lt;Revision&gt;\d{1,2})</_BuildRevisionRegEx>
```
rainersigwald: lowercase the `E` #capitalizationmatters

build/NuGetPackages/CreateNuGetPackages.proj
```diff
@@ -58,5 +58,32 @@
...
+    This target is used to customize the versions generated by NerdBank.GitVersioning.  The $(PrereleaseVersion) is read from version.json and
+    then the date and build number is appended.
+  -->
+  <PropertyGroup>
+    <BuildNumber>0</BuildNumber>
+    <!-- Local builds always have a revision of 00 -->
+    <BuildRevision Condition="'$(BUILD_NUMBER)' == ''">00</BuildRevision>
+    
+    <!-- Official builds have a revision starting with 01 every day -->
+    <BuildRevision Condition="'$(BUILD_NUMBER)' != ''">$([System.Int32]::Parse($([System.Text.RegularExpressions.Regex]::Match($(BUILD_NUMBER), $(_BuildRevisionRegEx)).Groups["Revision"].Value)).ToString('D2'))</BuildRevision>
```
rainersigwald: Should we just change the build number format to be `$(rev:.rr)` and avoid this regex extraction (at the cost of requiring build configuration)?

version.json
```diff
@@ -1,7 +1,7 @@
 {
-  "version": "15.1-preview5",
+  "version": "15.1-preview",
```
rainersigwald: won't this asciibetically sort after `preview5`? Is that a problem?

build/NuGetPackages/CreateNuGetPackages.proj
jeffkl: We could, this just means it doesn't matter one way or the other.

jeffkl: Strike that, the build number is used for the folder name in the build drop

rainersigwald: Ah, that's a bit unfortunate. It'd be nice to have the folders match the nuget version but I don't see an easy way to compute it during the build and send it back. so this works.

version.json
rainersigwald: `rc`? That seems like a fairly accurate description of the contents.

jeffkl: I just checked and actually this will be fine.  People will specify `15.1.0-preview-*` and we've only published 1 package as version `15.1.0` and it's number is low enough that it will sort as last.

So we can keep `preview` as the tag unless we want it to be something else

rainersigwald: That's fine by me too.

jeffkl:

PR #1214, opened 2016-10-19
Allow updates to imports during the build

### LINKED ISSUE #1185: Cannot update an imported file within a build
This was reported offline by @kzu. Thanks!

In this situation:
1. A multi-project build with a "coordinating" project (like a `.sln` build).
2. The coordinating project calls into leaf projects with one set of global properties to extract some information.
3. The coordinating project updates files imported by leaf projects (think "NuGet restore").
4. The coordinating project then builds leaf projects again with a different set of global properties (so they get fully re-evaluated/rebuilt).

The builds in step 4 do not pick up the new versions of files updated by step 3--they continue to use the file contents from their original form read in step 2.

This breaks multi-step restore + build in a single invocation, even though the evaluation and builds are separated by being in different global property spaces.

This is happening because of caching of imported file contents done by `ProjectRootElementCache`.  The cache can be set to invalidate its entries based on file timestamp checks by [constructor argument](https://github.com/Microsoft/msbuild/blob/68b96728bc2b7c554a37d12723e68e5d37768743/src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs#L131), but this is not done by the [relevant call](https://github.com/Microsoft/msbuild/blob/cf7b9d2226e8466710972e0b3da53cd67ebba7f7/src/XMakeBuildEngine/Definition/ProjectCollection.cs#L256) in a command-line build, from this callstack:

```
    Microsoft.Build.dll!Microsoft.Build.Evaluation.ProjectRootElementCache.ProjectRootElementCache(bool autoReloadFromDisk) Line 137
    Microsoft.Build.dll!Microsoft.Build.Evaluation.ProjectCollection.ProjectCollection(System.Collections.Generic.IDictionary<string, string> globalProperties, System.Collections.Generic.IEnumerable<Microsoft.Build.Framework.ILogger> loggers, System.Collections.Generic.IEnumerable<Microsoft.Build.Logging.ForwardingLoggerRecord> remoteLoggers, Microsoft.Build.Evaluation.ToolsetDefinitionLocations toolsetDefinitionLocations, int maxNodeCount, bool onlyLogCriticalEvents) Line 256
    MSBuild.exe!Microsoft.Build.CommandLine.MSBuildApp.BuildProject(string projectFile, string[] targets, string toolsVersion, System.Collections.Generic.Dictionary<string, string> globalProperties, Microsoft.Build.Framework.ILogger[] loggers, Microsoft.Build.Framework.LoggerVerbosity verbosity, Microsoft.Build.CommandLine.DistributedLoggerRecord[] distributedLoggerRecords, bool needToValidateProject, string schemaFile, int cpuCount, bool enableNodeReuse, System.IO.TextWriter preprocessWriter, bool debugger, bool detailedSummary) Line 920
    MSBuild.exe!Microsoft.Build.CommandLine.MSBuildApp.Execute(string commandLine) Line 545
    MSBuild.exe!Microsoft.Build.CommandLine.MSBuildApp.Main() Line 202
```

There's no way to specify that the cache _should_ check for updated targets within a build. Flipping that bit with a debugger produces the expected result (from the `MSBUILDDEBUGXMLCACHE=1` debug log):

```
P 8232 | Satisfied from XML cache: C:\Program Files (x86)\MSBuild\Microsoft\NuGet\Microsoft.NuGet.targets
P 8232 | Forgetting: S:\NuGetizer\src\Build\Sample\Sample.nuget.targets
P 8232 | Out of date dropped from XML cache: S:\NuGetizer\src\Build\Sample\Sample.nuget.targets
P 8232 | Adding: S:\NuGetizer\src\Build\Sample\Sample.nuget.targets
P 8232 | Satisfied from XML cache: C:\Users\raines\.nuget\packages\GitInfo\1.1.32\build\GitInfo.targets
```

rainersigwald:
Closes #1185

Allowing a once-read imported file to be updated during the build enables a carefully-crafted build process to do a NuGet restore and a build in the same MSBuild invocation.

### DIFF
--- a/src/XMakeBuildEngine/Definition/ProjectCollection.cs
+++ b/src/XMakeBuildEngine/Definition/ProjectCollection.cs
@@ -175,11 +175,6 @@ namespace Microsoft.Build.Evaluation
 
         /// <summary>
         /// The cache of project root elements associated with this project collection.
-        /// Each is associated with a specific project collection for two reasons:
-        /// - To help protect one project collection from any XML edits through another one:
-        /// until a reload from disk - when it's ready to accept changes - it won't see the edits;
-        /// - So that the owner of this project collection can force the XML to be loaded again
-        /// from disk, by doing <see cref="UnloadAllProjects"/>.
         /// </summary>
         private ProjectRootElementCache _projectRootElementCache;
 
@@ -253,7 +248,7 @@ namespace Microsoft.Build.Evaluation
             _loadedProjects = new LoadedProjectCollection();
             _toolsetDefinitionLocations = toolsetDefinitionLocations;
             this.MaxNodeCount = maxNodeCount;
-            this.ProjectRootElementCache = new ProjectRootElementCache(false /* do not automatically reload changed files from disk */);
+            this.ProjectRootElementCache = new ProjectRootElementCache(autoReloadFromDisk: true);
             this.OnlyLogCriticalEvents = onlyLogCriticalEvents;
 
             try
@@ -817,11 +812,9 @@ namespace Microsoft.Build.Evaluation
 
         /// <summary>
         /// The cache of project root elements associated with this project collection.
-        /// Each is associated with a specific project collection for two reasons:
-        /// - To help protect one project collection from any XML edits through another one:
-        /// until a reload from disk - when it's ready to accept changes - it won't see the edits;
-        /// - So that the owner of this project collection can force the XML to be loaded again
-        /// from disk, by doing <see cref="UnloadAllProjects"/>.
+        /// Each is associated with a specific project collection so that the owner of
+        /// this project collection can force the XML to be loaded again from disk,
+        /// by doing <see cref="UnloadAllProjects"/>.
         /// </summary>
         internal ProjectRootElementCache ProjectRootElementCache
         {


src/XMakeBuildEngine/Definition/ProjectCollection.cs
```diff
@@ -817,11 +812,9 @@ internal int MaxNodeCount
 
         /// <summary>
         /// The cache of project root elements associated with this project collection.
-        /// Each is associated with a specific project collection for two reasons:
-        /// - To help protect one project collection from any XML edits through another one:
-        /// until a reload from disk - when it's ready to accept changes - it won't see the edits;
```
rainersigwald: I am a little bit concerned about losing this property, but the benefit seems worth it. Cases where something is sneakily updating imports during a build causing torn state seem like they'd be rare.

rainersigwald:
Discovered additional considerations here after talking with some folks on the [Live Unit Testing](https://blogs.msdn.microsoft.com/visualstudio/2016/11/18/live-unit-testing-visual-studio-2017-rc/) team. They are deliberately taking a snapshot of a set of projects by loading them into the cache, and relying on that to ensure that project-to-project references see a consistent view of the world, even while the user is typing.

Unfortunately, we now have conflicting desires: this issue, and the ability to take a snapshot. Since the legacy behavior allowed the LUT folks to get snapshots, I'm going to defer to that, meaning we can't solve this the "easy way" I proposed in this PR and need a more sophisticated design.

PR #1219, opened 2016-10-20
Add precision for detecting when running tests.

### LINKED ISSUE #1218: Build fails when calling Build API from Unit Test
Partner team reported this issue. They're running a build as part of a unit test and the build behaves different. Our logic to detect if we're running as part of our own test suite is too broad.

AndyGerlicher:
Previously we assumed that if the entry point for our code was a test
runner that we were running as part of our unit test suite and had a few
code path differences (specifically around loading toolsets from config
file). The test was too broad and would detect when we were in a unit
test not associated with MSBuild and cause the build to fail.

This change first detects if we're in a known test runner (as it did
before) and then also checks to ensure that one of our known test
assemblies is loaded.

Fixes #1218

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -25,6 +25,13 @@ namespace Microsoft.Build.Shared
             "VSTESTHOST", "QTAGENT32", "CONCURRENT", "RESHARPER", "MDHOST", "TE.PROCESSHOST"
         };
 
+        private static readonly string[] s_testAssemblies =
+        {
+            "Microsoft.Build.Tasks.UnitTests", "Microsoft.Build.Engine.UnitTests", "Microsoft.Build.Utilities.UnitTests",
+            "Microsoft.Build.CommandLine.UnitTests", "Microsoft.Build.Engine.OM.UnitTests",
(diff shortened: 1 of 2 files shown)


src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -168,6 +174,23 @@ private static bool IsValidMSBuildPath(string path)
                    File.Exists($"{path}.config");
         }
 
+        private static bool CheckIfRunningTests(string processNameCommandLine, string processNameCurrentProcess)
+        {
+            // First check if we're running in a known test runner.
+            if (IsProcessInList(processNameCommandLine, s_testRunners) ||
+                IsProcessInList(processNameCurrentProcess, s_testRunners))
+            {
+                // If we are, then ensure that one of the known test assemblies is loaded.
```
rainersigwald: Change to

```
If we are, then ensure we're running MSBuild's tests by seeing if any of our assemblies are loaded.
```

or similar?

AndyGerlicher:
I thought about perf, but it will be the same as it was before except when running under a test _slightly_ slower. I think that's fine, and this only runs once (per assembly it's linked in).

PR #1220, opened 2016-10-20
Enable VisualStudioSetup Feature

AndyGerlicher:
- This was missed when the feature was merged to xplat. The feature flag
  was added to the code but not defined in the build.

### DIFF
--- a/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
+++ b/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
@@ -2503,6 +2503,71 @@ namespace Microsoft.Build.Tasks.Xaml
         protected override bool ValidateParameters() { throw null; }
     }
 }
+namespace Microsoft.VisualStudio.Setup.Configuration
+{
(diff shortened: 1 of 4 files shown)


ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
```diff
@@ -2503,6 +2503,71 @@ public abstract partial class XamlDataDrivenToolTask : Microsoft.Build.Utilities
         protected override bool ValidateParameters() { throw null; }
     }
 }
+namespace Microsoft.VisualStudio.Setup.Configuration
```
rainersigwald: These shouldn't be exposed from our assemblies, should they?

AndyGerlicher: We don't ship the assembly, they're just embedded interop types for the COM API. There's a feature request on Roslyn to `/linkinternal` rather than `/link`, but we have no choice right now they have to be public.

rainersigwald: 😕 ok

PR #1227, opened 2016-10-21
Add stream buffering for named pipe reads

dsplaisted:
This change adds a read buffer for reads from the named pipe streams which are used to communicate between MSBuild processes.  As there are lots of short reads in the serialization code, this should improve performance of the communication.

### DIFF
--- a/src/Shared/BufferedReadStream.cs
+++ b/src/Shared/BufferedReadStream.cs
@@ -1,0 +1,152 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+// </copyright>
+// <summary>Base class for the implementation of a node endpoint for out-of-proc nodes.</summary>
+//-----------------------------------------------------------------------
+
+using System;
+using System.IO;
+
+namespace Microsoft.Build.BackEnd
+{
+    internal class BufferedReadStream : Stream
+    {
+        const int BUFFER_SIZE = 1024;
+
+        Stream _innerStream;
+        byte[] _buffer;
+
+        int _remainingBufferSize;
+        int _currentIndexInBuffer;
+
+        public BufferedReadStream(Stream innerStream)
+        {
+            _innerStream = innerStream;
+            _buffer = new byte[BUFFER_SIZE];
+
+            _remainingBufferSize = 0;
+        }
+
+        public override bool CanRead { get { return _innerStream.CanRead; } }
+
+        public override bool CanSeek { get { return false; } }
+
+        public override bool CanWrite { get { return _innerStream.CanWrite; } }
+
+        public override long Length { get { return _innerStream.Length; } }
+
+        public override long Position
+        {
+            get { return _innerStream.Position; }
+            set { _innerStream.Position = value; }
+        }
+
+        public override void Flush()
+        {
+            _innerStream.Flush();
+        }
+
+        public override int ReadByte()
+        {
+            if (_remainingBufferSize > 0)
+            {
+                int ret = _buffer[_currentIndexInBuffer];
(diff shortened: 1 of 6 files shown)


AndyGerlicher:
@dsplaisted did you change something that fixed the unit test? Looked like a legit failure (null ref so the underlying stream was null). Approach looks good though in general. Only thing I wish is we didn't have to have our own copy of that class. Did you copy this from somewhere? I saw there was one in coreclr, but it didn't seem to be public :(.

Were you able to verify this was any faster?

Thanks a ton for looking into this!

dsplaisted:
Yes, I fixed the unit test failure.

I didn't copy the code from anywhere, because the various buffered stream implementations I saw in .NET didn't exactly meet the requirements here.  They assumed that reads and writes to the stream applied to the same set of data, so writing to the stream would invalidate the read buffer, and vice versa.  The pipe streams we use are for duplex communication, so there's no relation between the reads and writes to the stream.  So the buffered stream I wrote buffers reads, but sends writes immediately to the underlying stream.

This sped up the scenario I was testing from around 70+ seconds to between 25 and 40 seconds.  There may be more perf to be gained.  This change only buffers the reads in one direction, and it would seem that also buffering in the other direction could help more but in my testing it slowed it down significantly (or maybe just hung, I stopped waiting after several minutes).

src/Shared/BufferedReadStream.cs
```diff
@@ -0,0 +1,152 @@
...
+                {
+                    Array.Copy(_buffer, _currentIndexInBuffer, buffer, offset, _remainingBufferSize);
+                    alreadyCopied = _remainingBufferSize;
+                    _currentIndexInBuffer = 0;
+                    _remainingBufferSize = 0;
+                }
+
+                int innerReadCount = _innerStream.Read(_buffer, 0, BUFFER_SIZE);
+                _currentIndexInBuffer = 0;
+                _remainingBufferSize = innerReadCount;
```
rainersigwald: Shouldn't this be `BUFFER_SIZE - innerReadCount`?

src/Shared/BufferedReadStream.cs
```diff
@@ -0,0 +1,152 @@
...
+            _remainingBufferSize = 0;
+        }
+
+        public override bool CanRead { get { return _innerStream.CanRead; } }
+
+        public override bool CanSeek { get { return false; } }
+
+        public override bool CanWrite { get { return _innerStream.CanWrite; } }
+
+        public override long Length { get { return _innerStream.Length; } }
```
cdmihai: Isn't this _innerStream.Length + _currentIndexInBuffer?

src/Shared/BufferedReadStream.cs
dsplaisted: This is the number of bytes from the underlying stream that have been read into the buffer but not by consumers of the buffered stream.  I've renamed the field to `_currentlyBufferedByteCount` to make this more clear.

src/Shared/BufferedReadStream.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        public override long Position
+        {
+            get { return _innerStream.Position; }
+            set { _innerStream.Position = value; }
```
dsplaisted: True.  This property shouldn't really be used for this scenario anyway, so I've changed it to throw a `NotSupportedException`.

dsplaisted:
> Overall looks good but I'd really like to see some tests for this.

This does get tested with tests that do multiprocess builds, so I'm assuming you mean unit-level tests.  Yeah... that would be nice... :-)

jeffkl:
In my perf testing, I didn't see any issues with the change but it didn't seem to be any faster either.

rainersigwald:
Ok, let's get this in for some mileage in RC. Tests should come ASAP #1272.

PR #1239, opened 2016-10-24
Add new method GetSDKContentFolderPath()

XianghanWang:
Add new method GetSDKContentFolderPath() to enable us retrieve content folder path from versioned path if needed. In RS2 SDK, we added version information in SDK path. This is the method to make sure we get correct path.

### DIFF
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -1075,7 +1075,7 @@ namespace Microsoft.Build.Utilities
             TargetPlatformSDK matchingSdk = GetMatchingPlatformSDK(targetPlatformIdentifier, targetPlatformVersion, diskRoots, null, registryRoot);
             string platformKey = TargetPlatformSDK.GetSdkKey(targetPlatformIdentifier, targetPlatformVersion);
             PlatformManifest manifest;
-            if(TryGetPlatformManifest(matchingSdk, platformKey, out manifest))
+            if (TryGetPlatformManifest(matchingSdk, platformKey, out manifest))
             {
                 if (manifest.VersionedContent)
                 {
@@ -1085,7 +1085,7 @@ namespace Microsoft.Build.Utilities
                 {
                     contractWinMDs = GetApiContractReferences(manifest.ApiContracts, matchingSdk.Path);
                 }
-            };
+            }
 
             return contractWinMDs;
         }
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -1167,6 +1167,88 @@ namespace Microsoft.Build.Utilities
         }
 
         /// <summary>
+        /// Return the versioned/unversioned SDK content folder path
+        /// </summary>
+        /// <param name="sdkIdentifier">The identifier of the SDK</param>
+        /// <param name="sdkVersion">The verision of the SDK</param>
+        /// <param name="targetPlatformIdentifier">The identifier of the targeted platform</param>
+        /// <param name="targetPlatformMinVersion">The min version of the targeted platform</param>
+        /// <param name="targetPlatformVersion">The version of the targeted platform</param> 
+        /// <param name="folderName">The content folder name under SDK path</param>
+        /// <returns>The SDK content folder path</returns>
+        public static string GetSDKContentFolderPath(
+              string sdkIdentifier,
+              string sdkVersion,
+              string targetPlatformIdentifier,
+              string targetPlatformMinVersion,
+              string targetPlatformVersion,
+              string folderName)
+        {
+            ErrorUtilities.VerifyThrowArgumentLength(sdkIdentifier, "sdkIdentifier");
+            ErrorUtilities.VerifyThrowArgumentLength(sdkVersion, "sdkVersion");
+
+            // Avoid exception in Path.Combine
+            if(folderName == null)
+            {
+                folderName = string.Empty;
+            }
+
+            // If no folder name is input or it isn't UWP SDK, return the root SDK path.
+            if (string.IsNullOrWhiteSpace(folderName) || sdkVersion != "10.0" || !string.Equals(sdkIdentifier, "Windows", StringComparison.OrdinalIgnoreCase))
+            {
+                string sdkLocation = GetPlatformSDKLocation(sdkIdentifier, sdkVersion);
+                return Path.Combine(sdkLocation, folderName);
+            }
+
+            ErrorUtilities.VerifyThrowArgumentLength(targetPlatformIdentifier, "targetPlatformIdentifier");
+            ErrorUtilities.VerifyThrowArgumentLength(targetPlatformVersion, "targetPlatformVersion");
+
+            string sdkContentFolderPath = null;
+            try
+            {
+                TargetPlatformSDK matchingSdk = GetMatchingPlatformSDK(targetPlatformIdentifier, targetPlatformVersion, null, null, null);
+                string platformManifestLocation = null;
+
+                if (matchingSdk != null)
+                {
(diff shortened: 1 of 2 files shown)


src/Utilities/ToolLocationHelper.cs
```diff
@@ -1167,6 +1167,88 @@ internal static string[] GetApiContractReferences(IEnumerable<ApiContract> apiCo
...
+                TargetPlatformSDK matchingSdk = GetMatchingPlatformSDK(targetPlatformIdentifier, targetPlatformVersion, null, null, null);
+                string platformManifestLocation = null;
+
+                if (matchingSdk != null)
+                {
+                    string platformKey = TargetPlatformSDK.GetSdkKey(targetPlatformIdentifier, targetPlatformVersion);
+
+                    if (!matchingSdk.Platforms.TryGetValue(platformKey, out platformManifestLocation))
+                    {
+                        ErrorUtilities.DebugTraceMessage("GetTargetPlatformReferencesFromManifest", "Target platform location '{0}' did not exist or did not contain Platform.xml", platformManifestLocation);
```
rainersigwald: Should this "category" (and the one below) be updated? Other places in this file seem to match the method name.

src/Utilities/ToolLocationHelper.cs
```diff
@@ -1167,6 +1167,88 @@ internal static string[] GetApiContractReferences(IEnumerable<ApiContract> apiCo
...
+            {
+                string sdkLocation = GetPlatformSDKLocation(sdkIdentifier, sdkVersion);
+                return Path.Combine(sdkLocation, folderName);
+            }
+
+            ErrorUtilities.VerifyThrowArgumentLength(targetPlatformIdentifier, "targetPlatformIdentifier");
+            ErrorUtilities.VerifyThrowArgumentLength(targetPlatformVersion, "targetPlatformVersion");
+
+            string sdkContentFolderPath = null;
+            try
```
rainersigwald: This block looks very similar to what's in `GetTargetPlatformReferencesFromManifest`. Should it be a reusable method?

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -103,6 +103,55 @@ public void GetApiContractReferencesFindsVersionedWinMDs()
         }
 
         [Fact]
+        public void GetWinBlueSDKLocation()
+        {
+            string sdkRootPath = ToolLocationHelper.GetPlatformSDKLocation("Windows", "8.1");
+
+            string returnValue = ToolLocationHelper.GetSDKContentFolderPath("Windows", "8.1", null, null, null, null);
+            Assert.True(string.Equals(returnValue, sdkRootPath));
```
rainersigwald: Can you use `Assert.Equal` for new tests here?

XianghanWang: updated.

src/Utilities/ToolLocationHelper.cs
XianghanWang: Updated.

src/Utilities/ToolLocationHelper.cs
XianghanWang: updated.

XianghanWang:
Updated based on Rainer's comments. I have verified the change works for our product code.

src/Utilities/ToolLocationHelper.cs
```diff
@@ -1166,6 +1138,96 @@ internal static string[] GetApiContractReferences(IEnumerable<ApiContract> apiCo
             return contractWinMDs.ToArray();
         }
 
+        private static void GetValueUsingMatchingSDKManifest(TargetPlatformSDK matchingSdk, string platformKey, Action<PlatformManifest> action)
```
jeffkl: It's rather unusual for this method to take a Action rather than just return the value.  Why not have it return the manifest and have the caller then process it?  We don't use this kind of pattern anywhere else in this file.

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -103,6 +103,55 @@ public void GetApiContractReferencesFindsVersionedWinMDs()
         }
 
         [Fact]
+        public void GetWinBlueSDKLocation()
+        {
+            string sdkRootPath = ToolLocationHelper.GetPlatformSDKLocation("Windows", "8.1");
+
+            string returnValue = ToolLocationHelper.GetSDKContentFolderPath("Windows", "8.1", null, null, null, null);
+            Assert.Equal(returnValue, sdkRootPath);
```
jeffkl: You have all of your Asserts backwards, it's supposed to be expected, then actual.
``` C#
void Equal(string expected, string actual)
```

src/Utilities/ToolLocationHelper.cs
XianghanWang: fixed.

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
jeffkl: The Assert at this line still looks backwards.  The returnValue is _actual_ right?

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -103,6 +103,55 @@ public void GetApiContractReferencesFindsVersionedWinMDs()
...
+            Assert.Equal(Path.Combine(sdkRootPath, @"DesignTime\CommonConfiguration\Neutral"), returnValue);
+        }
+
+        [Fact]
+        public void GetSDKRootLocation()
+        {
+            string expectedValue = ToolLocationHelper.GetPlatformSDKLocation("UAP", "10.0.10586.0");
+
+            string versionedSDKValue = ToolLocationHelper.GetSDKContentFolderPath("Windows", "10.0", "UAP", "10.0.14944.0", "10.0.14944.0", null);
+            Assert.Equal(versionedSDKValue, expectedValue);
```
jeffkl: Still backwards

src/Utilities/UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -145,63 +145,59 @@ public void GetUnversionedSDKUnionMetadataLocation()
         [Fact]
         public void GetVersionedSDKUnionMetadataLocation()
         {
-            string sdkRootPath = ToolLocationHelper.GetPlatformSDKLocation("Windows", "10.0");
-
             // Create manifest file
-            string platformFolder = Path.Combine(sdkRootPath, @"Platforms\UAP\10.0.14944.0");
+            string platformRootFolder = Path.Combine(Path.GetTempPath(), @"MockSDK");
+            string sdkRooFolder = Path.Combine(platformRootFolder, @"Windows Kits\10");
```
jeffkl: spelling

src/Utilities/ToolLocationHelper.cs
```diff
@@ -1072,47 +1072,20 @@ string registryRoot
...
-                            contractWinMDs = GetApiContractReferences(manifest.ApiContracts, matchingSdk.Path);
-                        }
-                    }
-                }
-            }
-            catch (Exception e) when (ExceptionHandling.IsIoRelatedException(e))
-            {
-                ErrorUtilities.DebugTraceMessage("GetTargetPlatformReferences", "Encountered exception trying to gather the platform references: {0}", e.Message);
-            }
+            };
```
rainersigwald: Can you fix this stray semicolon?

rainersigwald:
@XianghanWang I'll follow up with another PR to make this cross-platform compatible.

Issue #1247, opened 2016-10-25
GenerateResources does not support linked resources on .NET Core


rainersigwald:
See discussion in https://github.com/Microsoft/msbuild/issues/1197#issuecomment-254564243 and dotnet/corefx#8200.

We don't have `ResXResourceReader` on .NET Core, so we can't crack a `.resx` file to find its linked resources. But that's ok because a .NET Core application couldn't consume them even if they were embedded.

At the moment, we will ignore linked resources when running in .NET Core.

rainersigwald:
#4090 will improve this greatly for linked text files, but I'm leaving this open until it behaves correctly for incremental builds -- #1197 is long closed by #1251, but that doesn't consider linked resources.

kirakosyan:
Having strange behavior, when running locally with with .net core 2.2 on local machine, it works fine. while hosted in azure I get 
RegistrationEmail.html;System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089;utf-8

rainersigwald:
@kirakosyan You're hitting the difference between .NET Core MSBuild (`dotnet build`, evidently used by your Azure Pipelines CI) and full framework MSBuild (`msbuild.exe` and Visual Studio driven builds). Can you change your Azure DevOps build definition to use the "MSBuild" or "Visual Studio Build" tasks instead of "dotnet build"?

PR #1251, opened 2016-10-26
GenerateResources incrementality for netcore

### LINKED ISSUE #1197: GenerateResources task is not incremental on .NET Core
The `CoreResGen` target is [not incremental](https://github.com/Microsoft/msbuild/blob/672169fed3a16218f8301391f956c8279fc7dae7/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L2768-L2770), because it assumes that the `GenerateResources` task handles incrementality internally. It does . . . on full framework, when `FEATURE_BINARY_SERIALIZATION` is enabled and [`NeedToRebuildSourceFile`](https://github.com/Microsoft/msbuild/blob/575424a207f8b46b8299b48fdb930a4fb99ac402/src/XMakeTasks/GenerateResource.cs#L1410) gets called.

On .NET Core, though, that codepath falls back to the conservative-but-slow "always just rebuild".

(noticed by @eerhardt in https://github.com/dotnet/sdk/issues/284)

rainersigwald:
Closes #1197.

This implementation is sufficient for current .NET Core--see #1247 for
details. If we start supporting linked resources, this will produce false
up-to-date results and must be changed.

cc @danmosemsft @eerhardt

### DIFF
--- a/src/XMakeTasks/GenerateResource.cs
+++ b/src/XMakeTasks/GenerateResource.cs
@@ -1438,7 +1438,16 @@ namespace Microsoft.Build.Tasks
 
             return false;
 #else
-            return true;
+#if FEATURE_RESX_RESOURCE_READER
+            Compile error: if we get Resx reading before binary serialization
+            it might be interesting to get caching working some other way.
+#endif
+            // On .NET Core, we don't have binary serialization to maintain
+            // the cache, but conveniently .NET Core assemblies (the only
+            // supported target of .NET Core MSBuild) cannot use linked
+            // resources. So the only relevant comparison is input/output.
+            // See https://github.com/Microsoft/msbuild/issues/1197
+            return NeedToRebuildSourceFile(sourceTime, outputTime);
 #endif
         }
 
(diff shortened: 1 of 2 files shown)


danmosemsft:
LGTM thanks @rainersigwald . How do we get the dotnet/buildtools updated (which I guess updates the msbuild for dotnet/coreclr and dotnet/corefx)

danmosemsft:
:shipit:

rainersigwald:
Confirmed that this fixes the scenario from https://github.com/dotnet/sdk/issues/284 (on Windows). I'll look at the OS X failure shortly.

@danmosemsft To get this through to corefx/coreclr, we'll need to a) update buildtools and b) update the version reference in those repos. Unfortunately the current MSBuild version is pretty stale so part a) may be nontrivial. @joperezr has done those steps in the past.

joperezr:
Once you have this change in a package, I'm happy to help getting it into buildtools and then to corefx/coreclr.

rainersigwald:

(checking for stability since there's a race condition in a lot of the incrementality tests on HFS+)

danmosemsft:
@rainersigwald perhaps consider using #ifdef or XUnit attributes if possible so that the increased timeout is only for OSX. Otherwise all runs slow down unnecessarily.

rainersigwald:
@danmosemsft Good point. Made it conditional on OS.

rainersigwald:
@joperezr MSBuild packages versioned `15.1.0-preview-000358-00` should have this change. Ping me if there's trouble and we'll work through it.

PR #1252, opened 2016-10-26
CPS and Roslyn project system GUIDs

### LINKED ISSUE #1237: Linux builds fail if CPS project factory GUID is used in solution file
Moved from https://github.com/dotnet/sdk/issues/297

Using a solution with the CPS project factory GUIDs ( {13B669BE-BB05-4DDF-9536-439F39A36129}  instead of  {FAE04EC0-301F-11D3-BF4B-00C04F79EFBC} ) causes x-plat builds to fail.

The new CPS guid for C#\VB are:
9A19103F-16F7-4668-BE54-9A1E7A4F7556 and 778DAE3C-4631-46EA-AA77-85C1314464D9. This is apart from the CPS one mentioned above which works for both projects. 

These should work in parallel with the existing ones.

rainersigwald:
... are known to be MSBuild format, so treat them as such.

Fixes #1237.

### DIFF
--- a/src/XMakeBuildEngine/Construction/Solution/SolutionFile.cs
+++ b/src/XMakeBuildEngine/Construction/Solution/SolutionFile.cs
@@ -70,6 +70,9 @@ namespace Microsoft.Build.Construction
 
         private const string vbProjectGuid = "{F184B08F-C81C-45F6-A57F-5ABD9991F28F}";
         private const string csProjectGuid = "{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}";
+        private const string cpsProjectGuid = "{13B669BE-BB05-4DDF-9536-439F39A36129}";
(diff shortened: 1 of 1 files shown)


rainersigwald:
@dsplaisted I think this'll resolve the mysterious failures of https://github.com/dotnet/sdk/issues/297--is there a branch of the SDK repo with the changes that I can test it out on?

rainersigwald:
cc @srivatsn

dsplaisted:
@rainersigwald Here is a branch with the CPS Guids: https://github.com/dsplaisted/sdk/tree/CPSGuids

rainersigwald:
@dsplaisted Thanks! This does fix the problem then (cloned + built that branch, saw the problem, overwrote with binaries from this change, built SDK and it got way past this error).

rainersigwald:

rainersigwald:
Arg, I should not have squashed this. I'll cherry-pick it into master shortly.

PR #1256, opened 2016-10-26
Add init-tools missing step

danmosemsft:


danmosemsft:
@jeffkl good to go?

rainersigwald:
Unfortunately this only applies on xplat. I'll add a note.

PR #1261, opened 2016-10-26
Introduce IEventSource2 to not break usages of IEventSource

jeffkl:
We can still improve on this by having an ILogger2 and update Logger but this will keep the telemetry stuff in without breaking people who implement IEventSource

### DIFF
--- a/src/Utilities/MuxLogger.cs
+++ b/src/Utilities/MuxLogger.cs
@@ -325,7 +325,7 @@ namespace Microsoft.Build.Utilities
 #if FEATURE_APPDOMAIN
             MarshalByRefObject,
 #endif
-            IEventSource
+            IEventSource2
         {
             #region Fields
             /// <summary>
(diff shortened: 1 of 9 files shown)


src/Utilities/MuxLogger.cs
```diff
@@ -1341,7 +1346,11 @@ private void UnregisterAllEventHandlers()
                 _eventSourceForBuild.TaskFinished -= _taskFinishedEventHandler;
                 _eventSourceForBuild.TaskStarted -= _taskStartedEventHandler;
                 _eventSourceForBuild.WarningRaised -= _buildWarningEventHandler;
-                _eventSourceForBuild.TelemetryLogged -= _telemetryEventHandler;
+
+                if (_eventSourceForBuild is IEventSource2)
+                {
+                    ((IEventSource2)_eventSourceForBuild).TelemetryLogged -= _telemetryEventHandler;
```
rainersigwald: I think the usual advice is to do `as` and check for `null`, instead of checking for `is` followed by a cast.

jeffkl: you got it

jeffkl:
@dotnet-bot test Windows_NT Build for CoreCLR  please

rainersigwald:

PR #1262, opened 2016-10-26
Move design time targets import before Common targets

basoundr:
Imports for Design time targets is moved before common targets so that
targets that gets imported through common targets, like Nuget, will be
able to override the target and property values.

@Microsoft/build @srivatsn for review

### DIFF
--- a/src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
@@ -319,6 +319,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <Import Project="$(CSharpCoreTargetsPath)" />
+    
+    <!-- Import design time targets for Roslyn Project System. These are only available if Visual Studio is installed. -->
+    <!-- Import design time targets before the common targets, which import targets from Nuget. -->
+    <PropertyGroup>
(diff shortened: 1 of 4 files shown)


src/XMakeTasks/Microsoft.CSharp.CurrentVersion.targets
```diff
@@ -361,12 +368,6 @@ using System.Reflection%3b
         <_ExplicitReference Include="$(FrameworkPathOverride)\mscorlib.dll" />
     </ItemGroup>
 
-    <!-- Import design time targets for Roslyn Project System. These are only available if Visual Studio is installed. -->
-    <PropertyGroup>
-       <CSharpDesignTimeTargetsPath Condition="'$(CSharpDesignTimeTargetsPath)'==''">$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed\Microsoft.CSharp.DesignTime.targets</CSharpDesignTimeTargetsPath>
-    </PropertyGroup>
-    <Import Project="$(CSharpDesignTimeTargetsPath)" Condition="'$(CSharpDesignTimeTargetsPath)' != '' and Exists('$(CSharpDesignTimeTargetsPath)')" />
```
cdmihai: This means that the design time targets won't be able to override stuff in Microsoft.Common.targets. Is that alright?

srivatsn: Yes it defines additional stuff needed for VS.

basoundr: We are fine with that

PR #1268, opened 2016-10-27
Add --target All to cibuild.cmd

jeffkl:
I find myself doing this a lot:

```
D:\MSBuild>cibuild.cmd --scope Build && cibuild.cmd --target CoreCLR --scope build
```

This just builds both if you specify `--target both` and it also preserves the original command-line arguments.

Also fixed a missing echo from a recent change I made.

### DIFF
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -1,5 +1,6 @@
 @if not defined _echo echo off
-setlocal
+setlocal ENABLEDELAYEDEXPANSION ENABLEEXTENSIONS
+SET _originalScript=%~f0
 
 :parseArguments
 if "%1"=="" goto doneParsingArguments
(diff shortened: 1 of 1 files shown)


cibuild.cmd
```diff
@@ -34,6 +35,13 @@ if /i "%TARGET%"=="CoreCLR" (
     set BUILD_CONFIGURATION=Debug-NetCore
 ) else if /i "%TARGET%"=="Desktop" (
     set BUILD_CONFIGURATION=Debug
+) else if /i "%TARGET%"=="All" (
+    SET _originalArguments=%*
+    CALL "!_originalScript!" !_originalArguments:All=Desktop!
+    IF ERRORLEVEL 1 GOTO :error
+    CALL "!_originalScript!" !_originalArguments:All=CoreCLR!
```
rainersigwald: Sneaky.

rainersigwald:
We should just disable the cancelled-build test. It's waaaay too flaky.

rainersigwald:

cibuild.cmd
jeffkl: It was the cleanest way I could think to preserve the original args

PR #1274, opened 2016-10-27
Fix unhandled exception when killing processes on non-Windows

### LINKED ISSUE #1269: Cancelling builds on non-Windows can cause an unhandled exception
We are currently killing processes as part of a build cancellation but on non-Windows we are attempting to read the ExitCode before the process exists which can cause an unhandled exception:

`System.InvalidOperationException: Process must exit before requested information can be determined.`

This was found in the `CancelledBuildWithDelay40` test.

Also, the test `CancelledBuildInTaskHostWithDelay40` should be disabled on non-Windows

jeffkl:
Also change ObjectModelHelpers.SleepCommand() to use a TimeSpan and different commands.

Disable CancelledBuildInTaskHostWithDelay20 and CancelledBuildInTaskHostWithDelay40 for scenarios where there is no task host support

Closes #1269

### DIFF
--- a/src/Shared/UnitTests/ObjectModelHelpers.cs
+++ b/src/Shared/UnitTests/ObjectModelHelpers.cs
@@ -1445,32 +1445,15 @@ namespace Microsoft.Build.UnitTests
         }
 
         /// <summary>
-        /// Command you can pass to Exec to sleep for rough number of milliseconds.
-        /// @for /l %i in (1,1,X) do "@dir %windir% > nul"
-        /// sleeps for X/100 seconds, roughly
-        /// This works around not having sleep.exe on the path.
+        /// Gets a command that can be used by an Exec task to sleep for the specified amount of time.
         /// </summary>
-        internal static string SleepCommandInMilliseconds(int milliseconds)
+        /// <param name="timeSpan">A <see cref="TimeSpan"/> representing the amount of time to sleep.</param>
+        internal static string GetSleepCommand(TimeSpan timeSpan)
         {
             return
-                string.Format(
-                    NativeMethodsShared.IsWindows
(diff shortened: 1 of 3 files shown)


src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1445,32 +1445,15 @@ internal static void ClearDirtyFlag(ProjectRootElement project)
...
-        /// sleeps for X/100 seconds, roughly
-        /// This works around not having sleep.exe on the path.
-        /// </summary>
-        /// <param name="seconds"></param>
-        /// <returns></returns>
-        internal static string SleepCommand(int seconds)
-        {
-            return SleepCommandInMilliseconds(seconds * 1000);
+                NativeMethodsShared.IsWindows
+                ? $"@powershell -command &quot;Start-Sleep -Milliseconds {(int)timeSpan.TotalMilliseconds}&quot; &gt;nul"
```
rainersigwald: What happened to `ping`?

(this should be fine, just different from what we chatted about)

jeffkl: It fell apart when one of the tests specified the sleep time in milliseconds and the resulting command was:
`ping 127.0.0.1 -n 0`

Which resulted in ping returning a non-zero exit code and failing the test.  Then I realized that the time had to be compatible with fractions of a second so I found this PowerShell command for Windows and realized that `sleep` on Linux and OS X accepts a fraction.

I also tried out making the tests only pass seconds but I was afraid I was going to cause more damage than good.

rainersigwald: Makes sense, and works for me. If we start running tests on nano server or whatever we can deal with it then.

PR #1278, opened 2016-10-27
Avoid x-targeting perf impact on vcxproj

rainersigwald:
Targeted mitigation for #1276.

VC++ noticed a dramatic regression in solution-load performance related to
the time cost of ResolveProjectReferences. That affects all projects, but
we can safely assume that a .vcxproj is NOT using cross-targeting, while
it's impossible to detect from the outside whether a .csproj is.

This commit avoids the regression for C++ projects by just not calling the
MSBuild task again for .vcxproj references.

/cc @nguerrera @olgaark @srivatsn

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1504,14 +1504,19 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     ======================================================================================
   -->
-  <Target Name="_GetProjectReferenceTargetFrameworkProperties" Outputs="%(_MSBuildProjectReferenceExistent.Identity)">
+  <Target Name="_GetProjectReferenceTargetFrameworkProperties"
+          Outputs="%(_MSBuildProjectReferenceExistent.Identity)">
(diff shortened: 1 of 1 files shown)


nguerrera:
Was about to say looks good then saw the error is related to your change:
```
error MSB4116: The condition "'%(_MSBuildProjectReferenceExistent.Extension)' != '.vcxproj'" on the "_GetProjectReferenceTargetFrameworkProperties" target has a reference to item metadata. References to item metadata are not allowed in target conditions unless they are part of an item transform. 
```

rainersigwald:
Someday I will either internalize that restriction or remove it! Looking.

olgaark:
Looks good to me too. Thanks!

rainersigwald:
I am confirming that a) this doesn't break cross-targeting for .csproj, and b) it does reduce time for .vcxproj.

rainersigwald:
Validated:

* Significant reduction in `_PerfIntellisenseInfo` execution time on the customer project (`.vcxproj`)
* Cross-targeting solution from @nguerrera returns a `netstandard1.3` assembly when called from a .NET 4.6 project, and `netstandard1.1` when called from a .NET 4.5 project.

PR #1280, opened 2016-10-28
Only have _CheckForInvalidConfigurationAndPlatform target run before …

jeffkl:
…any build related target

This target is running before targets like Restore which don't need the check.  This change removes `_CheckForInvalidConfigurationAndPlatform` from `InitialTargets` and instead injects it before any of the `BuildDependsOn`, `RebuildDependsOn`, or `CleanDependsOn` targets.  This will work except when a user specifies a `BeforeTargets` of the first target in those lists.  To get around that, the user must specify a `DependsOn` of `_CheckForInvalidConfigurationAndPlatform` if they want the check to run before their target.

Related to https://github.com/dotnet/sdk/issues/203

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -14,7 +14,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 ***********************************************************************************************
 -->
 
-<Project DefaultTargets="Build" InitialTargets="_CheckForInvalidConfigurationAndPlatform" TreatAsLocalProperty="OutDir" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" TreatAsLocalProperty="OutDir" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="'$(MicrosoftCommonPropsHasBeenImported)' != 'true' and Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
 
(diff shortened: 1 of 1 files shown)


jeffkl:
FYI @srivatsn @eerhardt @nguerrera @emgarten @piotrpMSFT

piotrpMSFT:
Wonderful! once this merges we'll be able to remove a bunch of workarounds in CLI!

rainersigwald:
@jeffkl can you please rebase this onto master? We're still trying to maintain it as an emergency fallback, so it should be the primary checkin branch for non-xplat-specific things.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -713,12 +713,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    to ensure that the target runs before any build related targets.  
+    If your target requires this check and is running as a BeforeTargets of one of the first targets
+    of $(BuildDependsOn), $(RebuildDependsOn), or $(CleanDependsOn) you will need to set your DependsOn
+    to this target.
     ============================================================
     -->
   <Target
-      Name="_CheckForInvalidConfigurationAndPlatform">
+      Name="_CheckForInvalidConfigurationAndPlatform"
+      BeforeTargets="$(BuildDependsOn);$(RebuildDependsOn);$(CleanDependsOn)">
```
rainersigwald: @cdmihai had suggested also adding `;Build;Rebuild;Clean` to the list just in case someone cleaned out the lists. Seems like a reasonable defense mechanism.

jeffkl:
Rebased onto master

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
jeffkl: Should it be `BeforeBuild;Build;BeforeClean;Clean;BeforeRebuild;Rebuild` ?  Or just `Build;Clean;Rebuild`?

rainersigwald:

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Let's do the long list.

cdmihai: I'd say `$(BuildDependsOn);Build;$(RebuildDependsOn);Rebuild;$(CleanDependsOn);Clean`

The `Before*` targets are incorporated in the `*DependsOn` properties, and users might prepend to the `*DependsOn` property, thus making the `Before*` targets not be the first ones anymore.

rainersigwald: Works for me.

piotrpMSFT:
@rainersigwald can you tell us what version of msbuild we can pick up with this change?

rainersigwald:
@piotrpMSFT `15.1.0-preview-000370-00` should do the trick. I'm prepping insertion to VS now.

PR #1287, opened 2016-10-31
Keep path in invalid project exceptions

### LINKED ISSUE #1286: Malformed-XML error messages no longer have filename + location
The error message for bad XML is missing the filename, line, and column location that it used to have.

I suspect this is due to the changes in #1004 (the `XmlTextReader` we're using no longer knows the path of the file, because we're giving it a stream instead.

<details>
<summary>Expected and actual output</summary>

```XML
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003"
      DefaultTargets="t1">
  <Target Name="t1">
    <Message Text= [p2.t1]/>
  </Target>
  <Target Name="t2">
    <Message Text="[p2.t2]"/>
  </Target>
</Project>
```

Current:

```
S:\work\1244940>\msbuild\bin\Bootstrap\15.0\Bin\MSBuild.exe pE.proj
Microsoft (R) Build Engine version 15.1.0.0

Build started 10/31/2016 11:14:10 AM.
Project "S:\work\1244940\pE.proj" on node 1 (default targets).
MSBUILD : error MSB4025: The project file could not be loaded. '[' is an unexpected token. The expected token is '"' or '''. Line 4, position 26. [S:\work\1244940\pE.proj]
Done Building Project "S:\work\1244940\pE.proj" (default targets) -- FAILED.


Build FAILED.

"S:\work\1244940\pE.proj" (default target) (1) ->
  MSBUILD : error MSB4025: The project file could not be loaded. '[' is an unexpected token. The expected token is '"' or '''. Line 4, position 26. [S:\work\1244940\pE.proj]

    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.23
```

MSBuild 14:

```
S:\work\1244940>MSBuild.exe pE.proj
Microsoft (R) Build Engine version 14.0.25420.1

Build started 10/31/2016 11:14:39 AM.
Project "S:\work\1244940\pE.proj" on node 1 (default targets).
S:\work\1244940\pE.proj(4,26): error MSB4025: The project file could not be loaded. '[' is an unexpected token. The expected token is '"' or '''. Line 4, position 26.
Done Building Project "S:\work\1244940\pE.proj" (default targets) -- FAILED.


Build FAILED.

"S:\work\1244940\pE.proj" (default target) (1) ->
  S:\work\1244940\pE.proj(4,26): error MSB4025: The project file could not be loaded. '[' is an unexpected token. The expected token is '"' or '''. Line 4, position 26.

    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.02
```
</details>

rainersigwald:
When #1004 moved the standard XML reading approach to be stream-based
rather than file-based, the exceptions thrown on malformed XML
changed--System.Xml no longer knows the path, so it isn't included in the
XmlException. That caused MSBuild to fail to report the location of the
XML error in a nice way as it had done before.

Almost every case where we constructed a BuildEventFileInfo object already
had access to the full path, so I added a constructor that accepted that
as an argument and overrides the possibly-empty path returned from
XmlException.SourceUri.

Fixes #1286.

### DIFF
--- a/src/Shared/BuildEventFileInfo.cs
+++ b/src/Shared/BuildEventFileInfo.cs
@@ -97,6 +97,16 @@ namespace Microsoft.Build.Shared
             _endColumn = 0;
         }
 
+        /// <summary>
+        /// Creates an instance of this class using the information in the given XmlException and file location.
+        /// </summary>
+        internal BuildEventFileInfo(string file, XmlException e) : this(e)
+        {
+            ErrorUtilities.VerifyThrowArgumentNull(file, nameof(file));
+
+            _file = file;
+        }
+
         #endregion
 
         #region Properties
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -2059,7 +2059,7 @@ namespace Microsoft.Build.Construction
 
                     if (xmlException != null)
                     {
-                        fileInfo = new BuildEventFileInfo(xmlException);
+                        fileInfo = new BuildEventFileInfo(fullPath, xmlException);
                     }
                     else
                     {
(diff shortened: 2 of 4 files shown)


rainersigwald:
I did test it manually but I found a good place to add a test so I did (verified that it fails before the new commit).

jeffkl:
Well then now I **really** approve these changes!

PR #1292, opened 2016-11-01
Fix some build warnings, update ref assemblies

jeffkl:


jeffkl:
I always squash and merge so I never think to split up the commits anyway.  Should I be just doing merge commits?

rainersigwald:
My philosophy is: separate ideas in separate commits. If a PR has multiple related-but-independent changes, I like individual commits + an explicit merge. If it's one interrelated change, squash it down.

Definitions are of course super fuzzy! But the conjunction in your PR title made me expect multiple commits.

jeffkl:
Yeah this was sort of general cleanup, I was going to send multiple PRs but it seemed like a waste.  I'll try to keep things separate.

PR #1303, opened 2016-11-02
OrcasEngine should not use ProductVersion

cdmihai:
Since we changed ProductVersion to include the git commit, the previous code was trying to call `new Version("15.1.318-preview5+gfd25bc9c7d")` 

Discovered while looking into https://devdiv.visualstudio.com/DevDiv/MSBuild/_workitems?id=286389&triage=true&_a=edit

### DIFF
--- a/src/OrcasEngine/Engine/Engine.cs
+++ b/src/OrcasEngine/Engine/Engine.cs
@@ -584,20 +584,25 @@ namespace Microsoft.Build.BuildEngine
             {
                 if (engineVersion == null)
                 {
-                    // Get the file version from the currently executing assembly.
-                    // Use .CodeBase instead of .Location, because .Location doesn't
-                    // work when Microsoft.Build.Engine.dll has been shadow-copied, for example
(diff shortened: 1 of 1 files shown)


cdmihai:
@rainersigwald @jeffkl 
On xplat this just works, because there's no git hash there. Should I abandon this PR?

cdmihai:
Actually it's still relevant, xplat FF builds use git versioning which adds the commit it to the product verison

src/OrcasEngine/Engine/Engine.cs
```diff
@@ -584,20 +584,9 @@ public static Version Version
             {
                 if (engineVersion == null)
                 {
-                    // Get the file version from the currently executing assembly.
-                    // Use .CodeBase instead of .Location, because .Location doesn't
-                    // work when Microsoft.Build.Engine.dll has been shadow-copied, for example
```
rainersigwald: Are the concerns about shadow copying no longer relevant?

cdmihai: @rainersigwald I recreated the original logic.

PR #1311, opened 2016-11-03
Legacy assembly version

cdmihai:
- set the static assembly version for the legacy projects, just like the other projects have it set. Otherwise they are compiled with an assembly version of 0.0.0.0, which breaks legacy tasks calling into them.
- rename "Desktop" to "Full" so things are consistent between C# land, xml land, and script land.


cdmihai:
@dotnet-bot test this please

cdmihai:
@jeffkl I changed the ProjectCollection to use the assembly file version. Everything passes now :)

cibuild.cmd
```diff
@@ -79,14 +80,20 @@ if "%SYNC_XLF%"=="true" (
     set SYNC_XLF_ARGUMENT="/p:SyncXlf=true"
 )
 
+:: Full Framework MSBuild does not have the new built-in property MSBuildRuntimeType
+set RUNTIMETYPE_ARGUMENT=
+if "%TARGET%"=="Full" (
```
jeffkl: I think we need to use MSBuild 14.0 to build the first phase which is what we do in our official builds.  Otherwise when I build locally the versions will be wrong for CoreCLR.  But at the moment we use the "downloaded" MSBuild for the first build.

src/Framework/AssemblyInfo.cs
```diff
@@ -43,6 +43,7 @@
 
 #if STATIC_VERSION_NUMBER
 [assembly: AssemblyVersion("15.1.0.0")]
```
jeffkl: We should probably change this to use the constant like the rest

jeffkl:
I can't wait until GitVersioning supports .NET Core.  When it does, I think we should have a unit test that verifies our assembly attributes.

cibuild.cmd
rainersigwald: We use installed MSBuild for the first build on Desktop ("Full" now, I guess 🙂). It's not unreasonable to do so for CoreCLR as well, though it is nice to get a) coverage on .NET Core MSBuild b) uniformity between first- and second-phase builds and c) a no-install-required experience.

src/XMakeBuildEngine/Definition/ProjectCollection.cs
```diff
@@ -403,7 +403,7 @@ public static Version Version
                     // work when Microsoft.Build.dll has been shadow-copied, for example
                     // in scenarios where NUnit is loading Microsoft.Build.
                     var versionInfo = FileVersionInfo.GetVersionInfo(FileUtilities.ExecutingAssemblyPath);
-                    s_engineVersion = new Version(versionInfo.ProductMajorPart, versionInfo.ProductMinorPart, versionInfo.ProductBuildPart, versionInfo.ProductPrivatePart);
+                    s_engineVersion = new Version(versionInfo.FileMajorPart, versionInfo.FileMinorPart, versionInfo.FileBuildPart, versionInfo.FilePrivatePart);
```
rainersigwald: I was very confused by this and then I read the comment which now matches the implementation. 👏

src/Framework/AssemblyInfo.cs
cdmihai: Tried that and reverted the change because most of the unit test dlls stopped compiling with conflicting types. This is because MSBuild is using shared sources instead of project dependencies. Once we replace all the shared sources with p2p dependencies, we can also replace this string with a constant :)

PR #1318, opened 2016-11-07
Fix tests that weren't working on Linux

### LINKED ISSUE #1297: Some tests fail on Linux
In PR #1296 I had to disable some tests that were failing on Linux but pass on OS X:

1. ResX2ResourcesWithReferences
2. ForceSomeOutOfDate
3. NothingOutOfDateExceptReference

We need to investigate why this is happening so we can enable these tests. 

Once this is completed, we should also update netci.groovy in master so that the tests run in the CI environment.

jeffkl:
Casing issue and a timing issue.

Closes #1297

### DIFF
--- a/src/XMakeTasks/UnitTests/GenerateResource_Tests.cs
+++ b/src/XMakeTasks/UnitTests/GenerateResource_Tests.cs
@@ -419,7 +419,8 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
                 System.Threading.Thread.Sleep(200);
                 if (!NativeMethodsShared.IsWindows)
                 {
-                    // Must be > 1 sec on HFS+ timestamp granularity
+                    // Must be > 1 sec on some file systems for proper timestamp granularity
+                    // TODO: Implement an interface for fetching deterministic timestamps rather than relying on the file
                     System.Threading.Thread.Sleep(1000);
                 }
 
@@ -613,7 +614,8 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
 
                 if (!NativeMethodsShared.IsWindows)
                 {
-                    // Must be > 1 sec for HFS+ timestamp granularity
+                    // Must be > 1 sec on some file systems for proper timestamp granularity
+                    // TODO: Implement an interface for fetching deterministic timestamps rather than relying on the file
                     System.Threading.Thread.Sleep(1100);
                 }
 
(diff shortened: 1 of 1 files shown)


jeffkl:
I manually updated the CI build to run tests on linux just to test this PR

jeffkl:
Tests passed!

Ubuntu 14.04: https://ci.dot.net/job/Microsoft_msbuild/job/master/job/_Ubuntu14.04_CoreCLR_prtest/163/testReport/
Ubuntu 16.04: https://ci.dot.net/job/Microsoft_msbuild/job/master/job/_Ubuntu16.04_CoreCLR_prtest/165/testReport/

src/XMakeTasks/UnitTests/GenerateResource_Tests.cs
```diff
@@ -419,7 +417,7 @@ public void ForceSomeOutOfDate()
                 DateTime time2 = File.GetLastWriteTime(t.OutputResources[1].ItemSpec);
 
                 System.Threading.Thread.Sleep(200);
-                if (NativeMethodsShared.IsOSX)
+                if (!NativeMethodsShared.IsWindows)
```
rainersigwald: I'm very surprised to see this. It really should be only HFS+ that's in the dark ages of one-second granularity. Are you seeing otherwise?

radical: Are you seeing this on Linux? I had a similar fix earlier https://github.com/Microsoft/msbuild/commit/0698cc9cffb03edebc4f7d5cd5453f804c97582e . Are you seeing failure in `TrackingDependenciesTests` too?

jeffkl: > I'm very surprised to see this. It really should be only HFS+ that's in the dark ages of one-second granularity. Are you seeing otherwise?

The test was failing with expected true, actual false.  It doesn't print out the exact timing but I noticed the condition was OS X only and updating it made it pass on Linux.  

In my opinion, no test should ever depends on a a `Sleep`.  I can update the comment but I think my change makes sense.

> Are you seeing this on Linux? I had a similar fix earlier 0698cc9 . Are you seeing failures in TrackingDependenciesTests too?

The TrackingDependenciesTests are [disabled on .NET Core](https://github.com/Microsoft/msbuild/blob/xplat/src/Utilities/UnitTests/TrackedDependencies/TrackedDependenciesTests.cs#L20) and therefore don't run on Linux.

jeffkl: Technically we shouldn't be relying on the file system to give us timestamps , the code should be mock-able so that we're testing its logic in a more deterministic way.  But until we have our tests set up like that, I'll just update the comment 😢

PR #1323, opened 2016-11-08
Project Element cloning asks a new hook if it is OK to clone attributes

### LINKED ISSUE #1319: DeepCopyFrom fails with Metadata as Attribute
DeepCopyFrom fails when it attempts to copy an element with metadata as attribute. This leads to an exception thrown and a partially copied project.

```
"The element <A> beneath element <Compile> is invalid because a child element with that name already exists"
```

``` Csharp
string testProject =
  @"<?xml version=""1.0"" encoding=""utf-8""?>
  <Project xmlns = ""http://schemas.microsoft.com/developer/msbuild/2003"">
    <ItemGroup>
      <Compile Include=""Class1.cs"" A=""a"" />
      <Compile Include=""Class2.cs"" />
    </ItemGroup>
  </Project>";

  var attributeProject = ProjectRootElement.Create(XmlReader.Create(new StringReader(testProject)));
  var elementProject = ProjectRootElement.Create();
  try
  {
    elementProject.RemoveAllChildren();
    elementProject.DeepCopyFrom(attributeProject);
  }
  catch (Exception e) // InvalidProjectException thrown
  {
  }

  Assert.AreEqual(elementProject.Items.Count, attributeProject.Items.Count); // Fails! "Class2.cs" is missing.
```

cdmihai:
Fixes #1319

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
@@ -131,13 +131,13 @@ namespace Microsoft.Build.Construction
 
         internal static void ValidateValidMetadataAsAttributeName(string name, string parentName, IElementLocation parentLocation)
         {
-            if (!AttributeNameCanRepresentMetadata(name))
+            if (!AttributeNameIsValidMetadataName(name))
             {
                 ProjectErrorUtilities.ThrowInvalidProject(parentLocation, "InvalidMetadataAsAttribute", name, parentName);
             }
         }
 
-        internal static bool AttributeNameCanRepresentMetadata(string name)
+        internal static bool AttributeNameIsValidMetadataName(string name)
         {
             bool isReservedAttributeName;
             bool isValidMetadataNameInAttribute;
--- a/src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
@@ -139,12 +139,12 @@ namespace Microsoft.Build.Construction
 
         internal static bool AttributeNameCanRepresentMetadata(string name)
         {
-            bool isReserverdAttributeName;
+            bool isReservedAttributeName;
             bool isValidMetadataNameInAttribute;
 
-            ProjectParser.CheckMetadataAsAttributeName(name, out isReserverdAttributeName, out isValidMetadataNameInAttribute);
+            ProjectParser.CheckMetadataAsAttributeName(name, out isReservedAttributeName, out isValidMetadataNameInAttribute);
 
-            return !isReserverdAttributeName && isValidMetadataNameInAttribute;
+            return !isReservedAttributeName && isValidMetadataNameInAttribute;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


basoundr:
@cdmihai Could you have this checked-in soon and have an insertion in WPT? We can get this change to RC2. Without this change, persisting the Version Metadata as an Attribute crashes VS.

src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
```diff
@@ -131,17 +131,22 @@ internal void ChangeName(string newName)
...
-            if (isKnownAttribute || !isValidMetadataNameInAttribute)
+            if (!AttributeNameCanRepresentMetadata(name))
             {
                 ProjectErrorUtilities.ThrowInvalidProject(parentLocation, "InvalidMetadataAsAttribute", name, parentName);
             }
         }
 
+        internal static bool AttributeNameCanRepresentMetadata(string name)
+        {
+            bool isReserverdAttributeName;
```
rainersigwald: typo: `Reserverd`

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectRootElement_Tests.cs
```diff
@@ -1204,6 +1205,42 @@ public void DeepCloneWithProjectExtensionsElementOfXml()
...
+        </B>
+        <C Include=`c`/>
+      </ItemGroup>
+    </Target>
+
+  </Project>";
+
+            var pre = ProjectRootElement.Create(XmlReader.Create(new StringReader(ObjectModelHelpers.CleanupFileContents(project))));
+
+            ValidateDeepCloneAndCopyFrom(pre);
```
rainersigwald: And this test did repro the crash, right?

src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
```diff
@@ -131,17 +131,22 @@ internal void ChangeName(string newName)
...
-            ProjectParser.CheckMetadataAsAttributeName(name, out isKnownAttribute, out isValidMetadataNameInAttribute);
-
-            if (isKnownAttribute || !isValidMetadataNameInAttribute)
+            if (!AttributeNameCanRepresentMetadata(name))
             {
                 ProjectErrorUtilities.ThrowInvalidProject(parentLocation, "InvalidMetadataAsAttribute", name, parentName);
             }
         }
 
+        internal static bool AttributeNameCanRepresentMetadata(string name)
```
rainersigwald: I find this name confusing, but I can't think of one that's unambiguously better. `AttributeIsMetadata`? `AttributeIsNotReserved`?

Fine to go with current if we can't think of something better.

cdmihai: How about `AttributeNameIsValidMetadataName`?

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectRootElement_Tests.cs
cdmihai: Yup, it had failed with the same `"The element <A> beneath element <Compile> is invalid because a child element with that name already exists"`. The `RemoveAllChilddren` from @jviau's repro did not seem necessary to cause the crash.

src/XMakeBuildEngine/Construction/ProjectMetadataElement.cs
rainersigwald: I think that's better, yeah.

PR #1324, opened 2016-11-08
Keep BuildEventContext.SubmissionId in serialization

### LINKED ISSUE #651: Hang on build failure
I've seen this a couple of times now. I made an error in C# and rebuilt (here in the MSBuild repo), and MSBuild appeared to hang after emitting the error. Hitting ctrl-c spits out the expected `Attempting to cancel the build...` message, but doesn't seem to do anything--it's still hung.

rainersigwald:
Fixes #651, which was occurring because the serialized SubmissionId
returned from the remote node didn't match the saved SubmissionId, so it
wasn't marked as complete in BuildManager.OnProjectFinished. Since there
was an incomplete submission, the build never finished.

### DIFF
--- a/src/Shared/NodePacketTranslator.cs
+++ b/src/Shared/NodePacketTranslator.cs
@@ -1318,7 +1318,7 @@ namespace Microsoft.Build.BackEnd
                         projectContextId = context.ProjectContextId;
                         taskId = context.TaskId;
                         projectInstanceId = context.ProjectInstanceId;
-                        submissionId = context.ProjectInstanceId;
+                        submissionId = context.SubmissionId;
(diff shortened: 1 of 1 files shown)


rainersigwald:
I looked around and sadly, there's no good place for a unit test for this. `BuildRequest_Tests` [does test translation](https://github.com/Microsoft/msbuild/blob/a6069f840ef13ffb7a8038f53c6740ff95751ba4/src/XMakeBuildEngine/UnitTests/BackEnd/BuildRequest_Tests.cs#L127), but it's testing a different mechanism--not the `NodePacketTranslator` used here.

Issue #1299, opened 2016-11-01
Make MSBuildAllProjects magically append on import


nguerrera:
See https://github.com/dotnet/sdk/issues/277#issuecomment-257720464

Today, every file has to add itself manually and many don't.

@rainersigwald

danmosemsft:
It should have done this from the start.

davkean:
@rainersigwald How would you design this if this was automatic? Should it remain as a property for specific targets (such as CoreCompile)? Or just a implicit input into every target?

rainersigwald:
This issue is about just augmenting the property to avoid having to append to it manually.

I don't think it should be treated as an implicit input to every target--that would be a full rebuild for every change to a property, which is correct (conservative) but inefficient. I'd rather see a more sophisticated up-to-date check that includes the consumed values like #701.

rainersigwald:
Possible perf issue observed by @lifengl (https://github.com/Microsoft/msbuild/issues/3004#issuecomment-366406033):

> In the trace, this property uses about half of the memory used by evaluation model, about 150M memory, which is quite excessive, and almost has no use in the design time.
> 
> ```
> Name                                                                                                  	Inc %	          Inc	   Inc Ct	Exc %	        Exc	 Exc Ct	     Fold	Fold Ct
> Microsoft.Build!Microsoft.Build.Evaluation.Project                                                   	28.1	  339,988,704	3,931,804.750	 0.0	 37,928.055	862.005	        0	     0
> + Microsoft.Build!Microsoft.Build.Evaluation.Project+Data                                             	28.0	  339,090,400	3,905,625.750	 0.0	106,888.563	861.992	        0	     0
> |+ mscorlib!List<Microsoft.Build.Evaluation.ProjectProperty>                                          	15.7	  190,223,296	896,655.500	 0.3	3,801,356.250	1,661.694	3,780,692.250	800.686
> ||+ Microsoft.Build!Microsoft.Build.Evaluation.ProjectProperty+ProjectPropertyXmlBackedWithPredecessor	13.8	  166,770,416	158,654.063	 0.2	2,584,567.750	107,690.102	        0	     0
> |||+ mscorlib!String (Bytes > 1K)                                                                     	12.3	  149,074,112	31,419.730	12.3	149,074,112	31,419.730	        0	     0
> |||+ mscorlib!String (Bytes > 10K)                                                                    	 1.2	   14,496,685	1,368.634	 1.2	 14,496,685	1,368.634	        0	     0
[...]

PR #1328, opened 2016-11-09
Improve incremental build in presence of globs

AndyGerlicher:
First iteration, looking for feedback. Filename, where I ended up putting it (Common targets vs Roslyn Core targets, etc.).

* Fix issue with incremental build with globs.
Add a target that runs before CoreCompile that will record and store the state of the Compile ItemGroup. This file is added to the CustomAdditionalCompileInputs which is an input to CoreCompile. When this file changes, CoreCompile will run again. This fixes the issue of deleting a file or adding a file with an old timestamp to a glob (e.g. *.cs).

* Add WriteOnlyWhenDifferent to WriteLinesToFile
Add option to not write the file when the file would not have changed (preserves the timestamp).

The original prototype wrote the file after compile and then subsequent runs could read that before compile and compare the file to the Compile ItemGroup. I went with this approach as it was far simpler and hopefully more elegant, but required a change to WriteLinesToFile to preserve the time stamp when the file doesn't change.


src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -3166,6 +3167,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     ============================================================
+                                        _GenerateCompilerDependencyFile
+
+    Generate file with a list of compile source files. This is used to track
+    incremental build between build execution specifically around addition and
+    deletions of file globs (e.g. <Compile Include="**\*.cs" />).
+    ============================================================
+    -->
+  <Target Name="_GenerateCompilerDependencyFile" BeforeTargets="CoreCompile">
+    <WriteLinesToFile Lines="@(Compile->'%(FullPath)')" File="$(IntermediateOutputPath)Compile.deps" Overwrite="True" WriteOnlyWhenDifferent="True" />
```
rainersigwald: Definitely also references (via `ReferencePath` and `DependsOnTargets="ResolveAssemblyReferences"`, probably?). Other properties and items _should_ be caught by the `Inputs="$(MSBuildAllProjects)"` [in `Microsoft.CSharp.Core.targets`](https://github.com/dotnet/roslyn/blob/98fb1f2a1a784e7724a3f043af342484f323703c/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets#L4). That would be more reliable with #1299 (as is, it might not catch some user-authored files) but is . . . ok now.

rainersigwald: I don't like `.deps`, especially since there's already the `.deps.json` files for .NET Core. Maybe `.CompilerInputs`?

agocke:
Right now we don't have any logic around incremental build in the compiler and right now that feels appropriate.

Basically, the compiler itself just executes build commands from MSBuild -- so it feels like whether or not to issue the command is the responsibility of the MSBuild core tasks + targets, while figuring out *how* to execute the build is the responsibility of the Roslyn tasks + targets.

rainersigwald:
@agocke Agreed, that totally meshes with the MSBuild model. But do you have any objection to this approach of writing references + source files to a file and using that as an additional input, so that we can catch the "compile `*.cs`" + "delete `something.cs`" + build (and actually re-invoke the compiler)?

agocke:
@rainersigwald I think the general strategy is OK, but you'll need much more than just the references + source files to guarantee deterministic recompilation. @jaredpar Has the list of inputs for the compiler that need to be considered.

Regarding specifically adding files, have you considered looking at the last-modified timestamp of the directories in the glob? That may be simpler if you only care about adding/removing source files.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
cdmihai: `corecompile.cache`?

rainersigwald:
Yeah, this is not a complete did-anything-change implementation, but it addresses a problem that will be more prevalent with the new dotnet SDK templates' use of wildcarding (because you can delete a file without changing the project file if it's not individually listed). The existing did-a-project-file-change check is the main "miscellaneous changes" catch.

ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
```diff
@@ -1177,6 +1177,7 @@ public partial class WriteLinesToFile : Microsoft.Build.Tasks.TaskExtension
         public Microsoft.Build.Framework.ITaskItem File { get { throw null; } set { } }
         public Microsoft.Build.Framework.ITaskItem[] Lines { get { throw null; } set { } }
         public bool Overwrite { get { throw null; } set { } }
+        public bool WriteOnlyWhenDifferent { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
```
jaredpar: Why have the `CompilerGeneratedAttribute` here?  The use doesn't seem to fit.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
```diff
@@ -23,6 +17,7 @@ public class WriteLinesToFile : TaskExtension
         private ITaskItem[] _lines = null;
         private bool _overwrite = false;
         private string _encoding = null;
+        private static Encoding s_defaultEncoding = new UTF8Encoding(false, true);
```
jaredpar: Recommend using named arguments here.  On a casual read no idea what `false` and `true` mean here.

jaredpar: Recommend using `readonly here as well.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
```diff
@@ -83,12 +85,12 @@ public override bool Execute()
                     }
                 }
 
-                Encoding encode = null;
+                Encoding encoding = s_defaultEncoding;
                 if (_encoding != null)
```
jaredpar: Making the `s_defaultEncoding` field with `readonly` allows you to avoid this logic.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
```diff
@@ -110,29 +111,33 @@ public override bool Execute()
...
-                            // is not the same as calling the overload that does not take encoding!
-                            // Encoding.Default is based on the current codepage, the overload without encoding is UTF8-without-BOM.
-                            if (encode == null)
+                            string contentsAsString = null;
+
+                            // When WriteOnlyWhenDifferent is set, read the file and if they're the same return.
+                            if (WriteOnlyWhenDifferent && System.IO.File.Exists(File.ItemSpec))
                             {
-                                System.IO.File.WriteAllText(File.ItemSpec, buffer.ToString());
+                                var existingContents = System.IO.File.ReadAllText(File.ItemSpec);
```
jaredpar: What happens when `ReadAllText` throws?  The use of `File.Exists` above doesn't give an guarantees this will succeed.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
```diff
@@ -110,29 +111,33 @@ public override bool Execute()
...
                         else
                         {
-                            // Passing a null encoding, or Encoding.Default, to WriteAllText or AppendAllText
-                            // is not the same as calling the overload that does not take encoding!
-                            // Encoding.Default is based on the current codepage, the overload without encoding is UTF8-without-BOM.
-                            if (encode == null)
+                            string contentsAsString = null;
+
+                            // When WriteOnlyWhenDifferent is set, read the file and if they're the same return.
+                            if (WriteOnlyWhenDifferent && System.IO.File.Exists(File.ItemSpec))
```
jaredpar: Recommend avoiding `Exists` here.  Just try and read the file and deal with errors if they happen.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
```diff
@@ -110,29 +111,33 @@ public override bool Execute()
...
                             {
-                                System.IO.File.WriteAllText(File.ItemSpec, buffer.ToString());
+                                var existingContents = System.IO.File.ReadAllText(File.ItemSpec);
+                                if (existingContents.Length == buffer.Length)
+                                {
+                                    contentsAsString = buffer.ToString();
```
jaredpar: Isn't this going to be causing a lot more allocations to occur in MSBuild as a result?  You're basically bringing all these files into memory as a contiguous string object.

ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
AndyGerlicher: This is our ref assembly, auto-generated.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
AndyGerlicher: This is *if* the user specified encoding (`_encoding` string set). I set it to the default so we could always safely use it rather than checking the parsed value for null because it failed to convert the string to an `Encoding` object.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
AndyGerlicher: ReadAllText throws when the file doesn't exist. Since this is a pretty normal case (in the context I'm using it every clean build the file won't be there) I'd rather check first to avoid the allocation/read entirely.

ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
jaredpar: If it's auto-generated then why do only half have the attribute?

src/XMakeTasks/FileIO/WriteLinesToFile.cs
AndyGerlicher: This is true and it could be optimized. We ended up going with just the hash so the effect here is very minimal and not worth optimizing. I will file an issue to improve it.

src/XMakeTasks/FileIO/WriteLinesToFile.cs
jaredpar: Hmm ... then my follow up suggestion would be to rename `_encoding` or `encoding`.  Makes it easy to mis-read this code 😄

ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
AndyGerlicher: The property I added was an auto-property, and that crazy compiler we use added the `CompilerGeneratedAttribute` to the IL. The ref assembly code is just a dump of that, and the other properties all have backing fields so they don't have it.

src/XMakeTasks/UnitTests/Hash_Tests.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        [Fact]
+        public void HashTaskEmptyInputTest()
+        {
+            System.Diagnostics.Debugger.Launch();
+            System.Diagnostics.Debugger.Break();
+
+            // Hash should be valid for empty item
+            var emptyItemHash = ExecuteHashTask(new ITaskItem[] {new TaskItem("")});
```
cdmihai: There could be another test for an empty array: ExecuteHashTask(new ITaskItem[0]);

AndyGerlicher:

AndyGerlicher:

src/XMakeTasks/Hash.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System.Security.Cryptography;
+using System.Text;
+using Microsoft.Build.Framework;
+
+namespace Microsoft.Build.Tasks
+{
+    /// <summary>
+    /// Generates a hash of a given ItemGroup items. Metadata is not considered in the hash.
+    /// <remarks>Currently uses SHA1. Implementation subject to change between MSBuild versions.</remarks>
```
rainersigwald: Maybe some sort of "Not intended as a cryptographic security measure" type caveat?

src/XMakeTasks/Hash.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// Hash of the ItemsToHash ItemSpec.
+        /// </summary>
+        [Output]
+        public string HashResult { get; set; }
+
+        /// <summary>
+        /// Execute the task.
+        /// </summary>
+        /// <returns></returns>
```
rainersigwald: nit: remove empty `<returns>`

src/XMakeTasks/Hash.cs
```diff
@@ -0,0 +1,60 @@
...
+
+                foreach (var item in ItemsToHash)
+                {
+                    hashInput.Append(item.ItemSpec);
+                    hashInput.Append(ItemSeparatorCharacter);
+                }
+
+                using (var sha1 = SHA1.Create())
+                {
+                    var hash = sha1.ComputeHash(Encoding.UTF8.GetBytes(hashInput.ToString()));
```
rainersigwald: I wish we didn't have to materialize the full string but I don't see a better way.

src/XMakeTasks/Microsoft.Build.Tasks.csproj
```diff
@@ -400,6 +400,7 @@
     <Compile Include="FindAppConfigFile.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
+    <Compile Include="Hash.cs" />
```
rainersigwald: big nit: alphabetize?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -3166,6 +3167,31 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <!--
     ============================================================
+                                        _GenerateCompilerDependencyCache
```
rainersigwald: I'd name this `GenerateCompileDependencyCache` (no `r`), because it's around the `Compile` target. But I could go a different way.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -3010,6 +3010,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       _GenerateCompileInputs;
       BeforeCompile;
       _TimeStampBeforeCompile;
+      _GenerateCompilerDependencyCache;
```
rainersigwald: Why add this here instead of doing a `BeforeTargets="CoreCompile"`? Just for relative ordering?

src/XMakeTasks/Resources/Strings.resx
```diff
@@ -2085,6 +2085,10 @@
     <value>MSB3491: Could not write lines to file "{0}". {1}</value>
     <comment>{StrBegin="MSB3491: "}</comment>
   </data>
+  <data name="WriteLinesToFile.ErrorReadingFile">
+    <value>MSB3492: Could not read existing file "{0}" to compare contents against. Will attempt to overwrite only.</value>
```
rainersigwald: `Could not read existing file "{0}" to determine whether its contents are up to date. Overwriting it.` ?

PR #1332, opened 2016-11-10
GetItemProvenance should not unescape input

### LINKED ISSUE #1106: GetItemProvenance should not unescape arguments
GetItemProvenance should not unescape the strings it gets. The following examples show what should match and what should not match. With the current implementation all the cases match:

Include=`”%61b%63”`
Should match: GetItemProvenance(“abc”)
Should not Match: GetItemProvenance(“ab%63”)

Include=`”a?c”`
Should match: GetItemProvenance(“abc”)
Should not match: GetItemProvenance(“a%62c”)

Include=`”a?%63”`
Should match: GetItemProvenance(“abc”)
Should not match: GetItemProvenance(“a%62c”)

Include=”a*c”
Should match: GetItemProvenance(“abcdec”)
Should match due to `*` glob: GetItemProvenance(“a%62c”)

Weird cases:

Include=`”%62”` // %62 is b
Should match: GetItemProvenance(“b”)
Should match because the string is the same: GetItemProvenance(“%62”)

cdmihai:
Resolves #1106.

Also fixes various issues with slash normalization which got exposed by the fix for #1106:
- slashes are normalized only when the string is a path
- checking whether a string is a valid path uses a union of Windows and !Windows characters, based on: https://github.com/Microsoft/msbuild/issues/781#issuecomment-243942514
- Re-implement Path.GetFileName so it works with potentially malformed paths.
- Normalize the slashes before normalizing the path. Otherwise .Net's implementation of GetFullPath does not work.

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -56,14 +56,33 @@ namespace Microsoft.Build.Shared
         internal static readonly StringComparison PathComparison = StringComparison.OrdinalIgnoreCase;
 
         /// <summary>
-        /// Path.GetInvalidPathChars() clones the array, so caching the result here to avoid needless memory churn
+        /// Copied from https://github.com/dotnet/corefx/blob/056715ff70e14712419d82d51c8c50c54b9ea795/src/Common/src/System/IO/PathInternal.Windows.cs#L61
+        /// MSBuild should support the union of invalid path chars across the supported OSes, so builds can have the same behaviour crossplatform: https://github.com/Microsoft/msbuild/issues/781#issuecomment-243942514
         /// </summary>
-        internal static char[] InvalidPathChars = Path.GetInvalidPathChars();
+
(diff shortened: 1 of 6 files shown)


cdmihai:
@dotnet-bot test this please

cdmihai:

cdmihai:
@Microsoft/msbuild-maintainers this is ready for review, all tests pass now :)

cdmihai:
@jviau 
Does this work as expected? Here's the unit tests asserting how GetItemProvenance works with escaped strings: https://github.com/cdmihai/msbuild/blob/193050e06840a58cbe0cb5aa316f13a7bedef42b/src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs#L3103

They mainly express the behaviour stated in this issue: https://github.com/Microsoft/msbuild/issues/1106#issue-179641634

jviau:
@cdmihai looks good to me!

PR #1336, opened 2016-11-11
Remove "Extensions" folder from MSBuildExtensionsPath

### LINKED ISSUE #1335: Applocal MSBuildExtensionsPath should not include the "Extensions" folder
We generate the MSBuildExtensionsPath [here](https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeBuildEngine/Utilities/Utilities.cs#L462) but out NuGet package does not place Microsoft.Common.props under an Extensions folder, just a `15.0` folder.

The dotnet CLI is [overriding](https://github.com/dotnet/cli/blob/b918b2a6b6485d0a82d81b4ffc217d911055602a/src/Microsoft.DotNet.Cli.Utils/CommandResolution/MSBuildProject.cs#L85) this setting MSBuildExtensionsPath but when you reference our package and try to open projects, the import of Microsoft.Common.props fails.

We need to get rid of the subfolder in our code so that the NuGet package will work as-is.  We could also add the folder in the NuGet package but that means we'll have this folder around forever.

jeffkl:
This only applies for "app local" MSBuild.  Our NuGet package does not have the Extensions folder and the dotnet CLI is already setting this so property.

Closes #1335

### DIFF
--- a/src/XMakeBuildEngine/Utilities/Utilities.cs
+++ b/src/XMakeBuildEngine/Utilities/Utilities.cs
@@ -459,7 +459,7 @@ namespace Microsoft.Build.Internal
             // environment or as a global property.  
 
 #if !FEATURE_INSTALLED_MSBUILD
-            string extensionsPath = Path.Combine(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory, ReservedPropertyNames.appLocalExtensionsPathSuffix);
+            string extensionsPath = BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory;
             string extensionsPath32 = extensionsPath;
 #else
             // "MSBuildExtensionsPath32". This points to whatever the value of "Program Files (x86)" environment variable is;
--- a/src/XMakeBuildEngine/Resources/Constants.cs
+++ b/src/XMakeBuildEngine/Resources/Constants.cs
@@ -44,7 +44,6 @@ namespace Microsoft.Build.Internal
         internal const string buildNodeCount = "MSBuildNodeCount";
         internal const string lastTaskResult = "MSBuildLastTaskResult";
         internal const string extensionsPathSuffix = "MSBuild";
-        internal const string appLocalExtensionsPathSuffix = "Extensions";
         internal const string userExtensionsPathSuffix = "Microsoft\\MSBuild";
         internal const string programFiles32 = "MSBuildProgramFiles32";
         internal const string localAppData = "LocalAppData";


jeffkl:
The current logic for full framework is in the app.config:

``` xml
<property name="MSBuildExtensionsPath" value="$(VsInstallRoot)\MSBuild" />
...
<projectImportSearchPaths>
  <searchPaths os="windows">
    <property name="MSBuildExtensionsPath" value="$(MSBuildProgramFiles32)\MSBuild"/>
    <property name="MSBuildExtensionsPath32" value="$(MSBuildProgramFiles32)\MSBuild"/>
    <property name="MSBuildExtensionsPath64" value="$(MSBuildProgramFiles32)\MSBuild"/>
  </searchPaths>
</projectImportSearchPaths>
```

I think the assumption here is that full framework MSBuild is always installed.  We might need to modify the app.config in the NuGet package so if you're redistributing MSBuild, the extensions path is your app's output directory.  But we can tackle that in another change.

src/XMakeBuildEngine/Utilities/Utilities.cs
```diff
@@ -459,7 +459,7 @@ internal static PropertyDictionary<ProjectPropertyInstance> GetEnvironmentProper
             // environment or as a global property.  
 
 #if !FEATURE_INSTALLED_MSBUILD
```
jeffkl: I'd like to get rid of all of the feature stuff and only have NETCORE, MONO or FULL (if possible).

rainersigwald: Strong disagree on that. The advantage of flagging for individual features is that it provides information about why things are changed in different flavors. Consider this example: it now doesn't really represent its name, because what it really means is `FEATURE_PROGRAMFILES_EXTENSION_ROOT` or something. But since we now know that `INSTALLED_MSBUILD` is a bad name, we can look at all uses of it and figure out what to do now. If it was just behind `!FULL`, there would be many more cases to look at.

Or think about `FEATURE_REGISTRY`--we can decide to enable that for windows, which is just finding the ifdevs and replacing them with runtime checks--but only the ones with a given label.

jeffkl: When you're in town next week, I'd love to discuss this.  I see all these "feature flags" and I've heard of the concept but am not entirely sold.  But I'm assuming their useful and I just don't get it?

jeffkl:
Is this safe to merge?

rainersigwald:
I'm :shipit:

jaredpar:
This change appears to have caused breaking behavior for Roslyn consuming MSBuild.  The full issue details are here: https://github.com/Microsoft/msbuild/issues/1533

In short though after this change it appears the only way to consume MSBuild in a cross platform environment is to have a setup like the following:

``` xml
   <When Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable' AND '$(OS)' != 'Windows_NT'">
      <PropertyGroup>
        <MSBuildTargetsFilePath>$(MSBuildExtensionsPath32)\Extensions\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.$(MSBuildTargetsLanguageName).targets</MSBuildTargetsFilePath>
      </PropertyGroup>
    </When>
    <When Condition="'$(TargetFrameworkIdentifier)' == '.NETPortable'">
      <PropertyGroup>
        <MSBuildTargetsFilePath>$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.$(MSBuildTargetsLanguageName).targets</MSBuildTargetsFilePath>
      </PropertyGroup>
    </When>
    <Otherwise>
      <PropertyGroup>
        <MSBuildTargetsFilePath>$(MSBuildToolsPath)\Microsoft.$(MSBuildTargetsLanguageName).targets</MSBuildTargetsFilePath>
      </PropertyGroup>
    </Otherwise>
  </Choose>
```

Essentially special casing the Extensions directory.  Is there a better way to do this that we are missing?

rainersigwald:
@jeffkl looks like the problem is that the [`Microsoft.Portable.Targets` package](https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.Portable.Targets/0.1.1-dev) puts stuff under `Extensions/`--so we were ok for the `Runtime` package since it doesn't deliver any extensions, but not when combined with `Portable`. We should either back this out or push a new `Microsoft.Portable.Targets` without that layer of folder structure.

I don't remember who published the original portable-targets package, unfortunately.

jeffkl:
I can patch up a new version of this package with the Extensions folder gone.  Would that be enough?

jaredpar:
I believe that should be enough.  CC @agocke to verify.

jeffkl:
Okay the package is available `Microsoft.Portable.Targets` version `0.1.2-dev`.  The only difference is that the Extensions folder is gone.  Let me know if that solves the issue and lets you get rid of that special case for non-Windows.  Sorry for the issue!

PR #1337, opened 2016-11-11
Make SerializedException robust to load errors

### LINKED ISSUE #1246: Malformed XML leads to premature node exit in some cases
Unfortunately, I was not able to isolate this repro any further so here's how to get to the same state in dotnet/sdk where I discovered this:
- mkdir repro
- cd repro
- git clone https://github.com/nguerrera/sdk
- cd sdk
- git checkout msbuild-repro
- build

**Expected result**
 An error indicating that the targets file is malformed. 

See https://github.com/nguerrera/sdk/commit/f752fe35e0c18f96632d597a4a1e3983d79b0b8f#commitcomment-19567654

**Actual result**

```
MSBUILD : error MSB4166: Child node "5" exited prematurely. Shutting down. Diagnostic 
information may be found in files in the temporary files directory named MSBuild_*.failure.txt.

UNHANDLED EXCEPTIONS FROM PROCESS 11084:
=====================
10/25/2016 1:10:55 PM
System.IO.IOException: Pipe is broken.
   at System.IO.Pipes.PipeStream.WriteCore(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Pipes.PipeStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at Microsoft.Build.BackEnd.NodeEndpointOutOfProcBase.PacketPumpProc()
===================
```

@raines @jeffkl

rainersigwald:
`GetTypeInfo().Assembly` seems to produce bogus results at least some of
the time (noticed especially for XmlException). Since exceptions are
serialized by including the exception's type and assembly, and then
deserialized by reflecting over that assembly to find the type's
constructor, this could cause an unhandled exception if the assembly
couldn't be loaded.

This change wraps the attempt to find the constructor robust to
file-not-found errors--if the assembly can't be loaded, simply use the
preexisting FormattedError fallback case.

Fixes #1246.

### DIFF
--- a/src/Shared/NodePacketTranslator.cs
+++ b/src/Shared/NodePacketTranslator.cs
@@ -18,6 +18,7 @@ using System.Runtime.Serialization.Formatters.Binary;
 using Microsoft.Build.Collections;
 using Microsoft.Build.Execution;
 using Microsoft.Build.Framework;
+using Microsoft.Build.Internal;
 using Microsoft.Build.Shared;
 using System.Globalization;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


rainersigwald:

rainersigwald:
@Microsoft/msbuild-maintainers Ping?

PR #1338, opened 2016-11-11
Remove xml stuff from nuspecs

jeffkl:
After our NuGet packages are pushed, the XMLNS is different in the downloaded one as well as the encoding/BOM.

I'm hoping just removing this stuff is enough to fix this problem.

I also updated the solution to include the new nuspecs and removed the old ones.


jeffkl:
I wasn't going to test it but I'll do it anyway and find out...

jeffkl:
Well well, NuSpecReferenceGenerator is also touching these files by loading them and saving them.  I'm trying more things now to determine what we can do...

Issue #1317, opened 2016-11-07
Add TargetFramework property and PackageReference item to XSD


rainersigwald:
Some of the new elements that are included by default in new .NET Core project templates show blue squigglies on viewing the project file. While the XSD is not canonical, it would be nice if it didn't show problems in default templates.

PR #1347, opened 2016-11-15
Add TargetFramework and PackageReference to the schema

RaulPerez1:
Update schema to add TargetFramework to PropertyGroup, PackageReference to ItemGroup and the valid sub properties of PackageReference.

Related to #1317

### DIFF
--- a/src/XMakeCommandLine/Microsoft.Build.CommonTypes.xsd
+++ b/src/XMakeCommandLine/Microsoft.Build.CommonTypes.xsd
@@ -168,6 +168,51 @@ elementFormDefault="qualified">
       </xs:complexContent>
     </xs:complexType>
   </xs:element>
+  <xs:element name="PackageReference" substitutionGroup="msb:Item">
+    <xs:annotation>
+      <xs:documentation>
+        <!-- _locID_text="PackageReference" _locComment="" -->Reference to a package
+      </xs:documentation>
+    </xs:annotation>
+    <xs:complexType>
+      <xs:complexContent>
+        <xs:extension base="msb:SimpleItemType">
+          <xs:sequence minOccurs="0" maxOccurs="unbounded">
+            <xs:choice>
+              <xs:element name="Version">
+                <xs:annotation>
+                  <xs:documentation>
+                    <!-- _locID_text="PackageReference_Version" _locComment="" -->Version of dependency
+                  </xs:documentation>
+                </xs:annotation>
+              </xs:element>
+              <xs:element name="IncludeAssets">
+                <xs:annotation>
+                  <xs:documentation>
+                    <!-- _locID_text="PackageReference_IncludeAssets" _locComment="" -->Assets to include from this reference
+                  </xs:documentation>
+                </xs:annotation>
+              </xs:element>
+              <xs:element name="ExcludeAssets">
+                <xs:annotation>
+                  <xs:documentation>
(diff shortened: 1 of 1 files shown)


src/XMakeCommandLine/Microsoft.Build.CommonTypes.xsd
```diff
@@ -168,6 +168,51 @@ elementFormDefault="qualified">
...
+      <xs:documentation>
+        <!-- _locID_text="PackageReference" _locComment="" -->Reference to a package
+      </xs:documentation>
+    </xs:annotation>
+    <xs:complexType>
+      <xs:complexContent>
+        <xs:extension base="msb:SimpleItemType">
+          <xs:sequence minOccurs="0" maxOccurs="unbounded">
+            <xs:choice>
+              <xs:element name="Version">
```
srivatsn: If I typed `<PackageReference Include="" ` will intellisense show Version as a possible attribute? We should make it so that it doesn't guide people towards using Version as an element instead of an attribute.

RaulPerez1: All the examples on the public docs were using it as a tag. If the goal is to deprecate the tag and move people to the attribute I can make the change to have version as an attribute and remove the tag.

srivatsn: Yes we're moving Version to be attribute

jeffkl:
@Microsoft/msbuild-maintainers I wonder if we should update the xsd for all of the metadata as attributes?

Ideally, the schema would enforce that you can specify one or the other but not both.

src/XMakeCommandLine/Microsoft.Build.CommonTypes.xsd
```diff
@@ -168,6 +168,51 @@ elementFormDefault="qualified">
       </xs:complexContent>
     </xs:complexType>
   </xs:element>
+  <xs:element name="PackageReference" substitutionGroup="msb:Item">
+    <xs:annotation>
+      <xs:documentation>
+        <!-- _locID_text="PackageReference" _locComment="" -->Reference to a package
```
rainersigwald: Interesting. How do these things get localized? @cdmihai do you know?

src/XMakeCommandLine/Microsoft.Build.CommonTypes.xsd
rainersigwald: I agree that it's reasonable to add these `PackageReference` metadata names only as attributes, and let people get the default squiggles if they add them as a sub-tag.

rainersigwald: Oh, ok. Works for me.

src/XMakeCommandLine/Microsoft.Build.CommonTypes.xsd
cdmihai: Apparently they get localized similar to the xaml files, though it's not quite clear to me what the final destination of the xsd files is in the VS build layout. Added a note for next sprint

RaulPerez1: I don't what the process for localization/insertion is but from the setup side it's picked up on the Community resource package from under the "Localized" sub folders in the binary. Layout wise it's dropped under <VsInsallDir>\Xml\Schemas\MsBuild

PR #1349, opened 2016-11-16
Guard against null ref when a Glob exclude fails to match any files

hippiehunter:
[test.txt](https://github.com/Microsoft/msbuild/files/593579/test.txt)

I've included a sample project file that would cause a null ref in src/Shared/FileMatcher.cs. I looked around inside GetFilesRecursiveStep to see if Files should have just been set to an empty array, but it didn't appear to match stylistically with the surrounding code so I just put the guard in GetFilesRecursive.

### DIFF
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -722,9 +722,12 @@ namespace Microsoft.Build.Shared
                     filesToExclude = new HashSet<string>();
                     foreach (var excludeStep in excludeNextSteps)
                     {
-                        foreach (var file in excludeStep.Files)
+                        if (excludeStep.Files != null)
                         {
-                            filesToExclude.Add(file);
+                            foreach (var file in excludeStep.Files)
+                            {
+                                filesToExclude.Add(file);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


AndyGerlicher:

cdmihai:
Can you please:
- find out why that value gets to be null (probably the true source of the bug), and then see if you can fix the bug at object creation rather than usage, since that null value might flow to other usages under different execution paths
- write a unit test exposing the crash. You can use this test as a reference: https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeBuildEngine/UnitTestsPublicOM/Definition/ProjectItem_Tests.cs#L624

hippiehunter:
I've added the unit test to the pull request. Fixing the root of the null is very simple, but every other use of this property was already guarded by a null check, do you want me to prevent null coming out and remove all of the (now redundant) call site null checking?

cdmihai:
@hippiehunter 
Took a closer look. Let's leave it as is, since it looks like the safest option. This is the meaning of not null: https://github.com/Microsoft/msbuild/blob/xplat/src/Shared/FileMatcher.cs#L817-L826

Otherwise, looks good to me. @Microsoft/msbuild-maintainers ?

rainersigwald:
LGTM

rainersigwald:
Thanks, @hippiehunter!

Issue #68, opened 2015-04-10
Add a "treat warnings as errors" option


Porges:
I would like an option to be able to treat all warnings as errors, when running from the command line.

AndyGerlicher:
Thanks for the suggestion, sounds like a potentially useful feature. This doesn't meet the bar for our roadmap at this time however.

jbjoshi:
Is there a Roadmap defined somewhere?

eatdrinksleepcode:
If it is a potentially useful feature, why is the issue closed? Someone from the community may want to pick it up.

chandramouleswaran:
As a work around, you should be (I think) able to add TreatWarningsAsErrors property in a targets file - set it to true and invoke that target as a part of the msbuild command line argument along with your default targets. Wouldn't that work?

Porges:
@chandramouleswaran: that's for the compiler warnings, not for warnings from MSBuild itself

chandramouleswaran:
@Porges - Good point :+1:

rainersigwald:
Reopening and marking as up-for-grabs.  This isn't high on our priority list but it's a good feature to have.

iouri-s:
We are getting a bunch of these warnings as well. The code accumulated dozens of warnings such as importing a target file several times, even though we have <TreatWarningsAsErrors>true</TreatWarningsAsErrors> defined.

D3-LucaPiombino:
We are in the same situation.

## Use case

**Misconfigured Platform across a Solution**

Having some dependency that require a specific platform (e.g. x64) but that when build from msbuild using the solution configuration matrix produce:

```
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets (1820, 5)
There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "Microsoft.ServiceFabric.Internal", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.
```

Since this is plugged in a CI/CD workflow, it would be nice to be able to block the release process in this case.

dotMorten:
We want the appveyor builds to fail if a PR introduces new warnings. However we don't want the dev to deal with this permanently as they are working towards a PR. Essentially we want gated checkins that also prevents users from introducing new warnings.

[...]

PR #1355, opened 2016-11-17
Implement /WarnAsError to treat specified warnings as errors

jeffkl:
Specify just /WarnAsError to have all warnings logged as errors as well as have the build fail.
Specify a comma delimited list of warning codes to have just that set of warnings treated as errors as well as have the build fail.

Targets will still show as succeeded and the tasks will continue to execute but the overall build result will be a failure.

Related to #68 and will close in my next change to add /NoWarn.

Help text should be reviewed as well:
```
  /warnaserror[:code[;code2]]
                     List of warning codes to treats as errors.  Use a semicolon
                     or a comma to separate multiple warning codes. To treat all
                     warnings as errors use the switch with no values.
                     (Short form: /err[:c;[c2]])
                     Example:
                       /warnaserror:MSB4130

                     When a warning is treated as an error the target will
                     continue to execute as if it was a warning but the overall
                     build result will fail.
```

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/BuildManager/BuildManager.cs
+++ b/src/XMakeBuildEngine/BackEnd/BuildManager/BuildManager.cs
@@ -424,7 +424,7 @@ namespace Microsoft.Build.Execution
                 }
 
                 // Set up the logging service.
-                ILoggingService loggingService = CreateLoggingService(_buildParameters.Loggers, _buildParameters.ForwardingLoggers);
+                ILoggingService loggingService = CreateLoggingService(_buildParameters.Loggers, _buildParameters.ForwardingLoggers, _buildParameters.WarningsAsErrors);
 
                 _nodeManager.RegisterPacketHandler(NodePacketType.LogMessage, LogMessagePacket.FactoryForDeserialization, loggingService as INodePacketHandler);
                 try
@@ -575,7 +575,11 @@ namespace Microsoft.Build.Execution
         public BuildResult BuildRequest(BuildRequestData requestData)
         {
             BuildSubmission submission = PendBuildRequest(requestData);
-            return submission.Execute();
+            BuildResult result = submission.Execute();
+
+            SetOverallResultIfWarningsAsErrors(result);
+
+            return result;
         }
 
         /// <summary>
@@ -621,6 +625,12 @@ namespace Microsoft.Build.Execution
 
                 if (loggingService != null)
                 {
+                    // Override the build success if the user specified /warnaserror and any errors were logged outside of a build submission.
+                    if (_overallBuildSuccess && loggingService.HasBuildSubmissionLoggedErrors(BuildEventContext.InvalidSubmissionId))
+                    {
+                        _overallBuildSuccess = false;
+                    }
+
                     loggingService.LogBuildFinished(_overallBuildSuccess);
                 }
 
@@ -1578,6 +1588,9 @@ namespace Microsoft.Build.Execution
                 if (_buildSubmissions.ContainsKey(result.SubmissionId))
                 {
                     BuildSubmission submission = _buildSubmissions[result.SubmissionId];
+
+                    SetOverallResultIfWarningsAsErrors(result);
+
                     submission.CompleteResults(result);
 
                     // If the request failed because we caught an exception from the loggers, we can assume we will receive no more logging messages for
@@ -1723,7 +1736,7 @@ namespace Microsoft.Build.Execution
         /// <summary>
         /// Creates a logging service around the specified set of loggers.
         /// </summary>
-        private ILoggingService CreateLoggingService(IEnumerable<ILogger> loggers, IEnumerable<ForwardingLoggerRecord> forwardingLoggers)
+        private ILoggingService CreateLoggingService(IEnumerable<ILogger> loggers, IEnumerable<ForwardingLoggerRecord> forwardingLoggers, ISet<string> warningsAsErrors)
         {
             int cpuCount = _buildParameters.MaxNodeCount;
 
@@ -1750,6 +1763,7 @@ namespace Microsoft.Build.Execution
             loggingService.OnLoggingThreadException += _loggingThreadExceptionEventHandler;
             loggingService.OnProjectStarted += _projectStartedEventHandler;
             loggingService.OnProjectFinished += _projectFinishedEventHandler;
+            loggingService.WarningsAsErrors = warningsAsErrors;
 
             try
             {
@@ -1818,6 +1832,23 @@ namespace Microsoft.Build.Execution
             }
 
             return castPacket;
+        }
+
+        /// <summary>
+        /// Sets the overall result of a build only if the user had specified /warnaserror and there were any errors.
+        /// This ensures the old behavior stays intact where builds could succeed even if a failure was logged.
+        /// </summary>
+        private void SetOverallResultIfWarningsAsErrors(BuildResult result)
+        {
+            if (result.OverallResult == BuildResultCode.Success)
+            {
+                ILoggingService loggingService = ((IBuildComponentHost)this).LoggingService;
+
+                if (loggingService.HasBuildSubmissionLoggedErrors(result.SubmissionId))
+                {
+                    result.SetOverallResult(overallResult: false);
+                }
+            }
         }
 
         /// <summary>
--- a/src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
@@ -201,6 +201,11 @@ namespace Microsoft.Build.Execution
         private bool _onlyLogCriticalEvents = false;
 
         /// <summary>
+        /// A list of warnings to treat as errors.
+        /// </summary>
+        private ISet<string> _warningsAsErrors = null;
+
+        /// <summary>
         /// The location of the toolset definitions.
         /// </summary>
         private ToolsetDefinitionLocations _toolsetDefinitionLocations = ToolsetDefinitionLocations.Default;
@@ -320,6 +325,7 @@ namespace Microsoft.Build.Execution
             _disableInProcNode = other._disableInProcNode;
             _logTaskInputs = other._logTaskInputs;
             _logInitialPropertiesAndItems = other._logInitialPropertiesAndItems;
+            _warningsAsErrors = other._warningsAsErrors == null ? null : new HashSet<string>(other._warningsAsErrors, StringComparer.OrdinalIgnoreCase);
         }
 
 #if FEATURE_THREAD_PRIORITY
@@ -581,6 +587,15 @@ namespace Microsoft.Build.Execution
         {
             get { return _onlyLogCriticalEvents; }
             set { _onlyLogCriticalEvents = value; }
+        }
+
+        /// <summary>
+        /// A list of warnings to treat as errors.  To treat all warnings as errors, set this to an empty <see cref="HashSet{String}"/>.  
+        /// </summary>
+        public ISet<string> WarningsAsErrors
+        {
+            get { return _warningsAsErrors; }
+            set { _warningsAsErrors = value; }
         }
 
         /// <summary>
--- a/src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
+++ b/src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
@@ -5,6 +5,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 
@@ -129,6 +130,24 @@ namespace Microsoft.Build.BackEnd.Logging
             get;
             set;
         }
+
+        /// <summary>
+        /// A list of warnings to treat as errors.  If null, nothing is treated as an error.  If an empty set, all warnings are treated as errors.
+        /// </summary>
+        public ISet<string> WarningsAsErrors
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// A list of build submission IDs that have logged errors.  If an error is logged outside of a submission, the submission ID is <see cref="BuildEventContext.InvalidSubmissionId"/>.
+        /// </summary>
+        public ISet<int> BuildSubmissionIdsThatHaveLoggedErrors
+        {
+            get;
+        } = new HashSet<int>();
+        
         #endregion
 
         #region Methods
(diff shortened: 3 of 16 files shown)


jeffkl:
@dotnet-bot test this please

jeffkl:
@dotnet-bot test this please

src/XMakeBuildEngine/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1578,6 +1594,18 @@ private void ReportResultsToSubmission(BuildResult result)
...
+
+                    if (result.OverallResult == BuildResultCode.Success)
+                    {
+                        ILoggingService loggingService = ((IBuildComponentHost)this).LoggingService;
+                        // Check if the user had specified /warnaserror and then override the result if there were any errors.
+                        // This ensures the old behavior stays in tact where builds could succeed even if a failure was logged.
+                        if (loggingService.WarningsAsErrors != null && loggingService.HasBuildSubmissionLoggedErrors(submission.SubmissionId))
+                        {
+                            result.SetOverallResult(overallResult: false);
+                        }
```
rainersigwald: Is there a good way to abstract this logic? It's pretty repetitive in here. Maybe `EnsureFailureIfWarningsAsErrors(BuildResult, ILoggingService)`?

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -320,6 +325,7 @@ private BuildParameters(BuildParameters other)
             _disableInProcNode = other._disableInProcNode;
             _logTaskInputs = other._logTaskInputs;
             _logInitialPropertiesAndItems = other._logInitialPropertiesAndItems;
+            _warningsAsErrors = other._warningsAsErrors == null ? null : new HashSet<string>(other._warningsAsErrors, StringComparer.OrdinalIgnoreCase);
```
rainersigwald: It's probably not worth changing this, but I would hazard a guess that for this use case (small sets of inputs), a `SortedSet` might actually be faster.

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -584,6 +590,15 @@ public bool OnlyLogCriticalEvents
         }
 
         /// <summary>
+        /// A list of warnings to treat as errors.  To treat all warnings as errors, set this to an empty <see cref="HashSet{String}"/>.
```
rainersigwald: Hmm. The double-use here seems a bit mysterious, in a way it doesn't on the command line. But it's not *wrong* . . .

src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
```diff
@@ -129,6 +130,24 @@ public bool HaveLoggedBuildFinishedEvent
             get;
             set;
         }
+
+        /// <summary>
+        /// A list of warnings to treat as errors.  If null, nothing is treated as an error.  If an empty set, all warnings are treated as errors.
```
rainersigwald: I'm confused, is `0` a different magic number than `-1` here?

src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
```diff
@@ -129,6 +130,24 @@ public bool HaveLoggedBuildFinishedEvent
...
+        /// </summary>
+        public ISet<string> WarningsAsErrors
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// A list of build submission IDs that have logged errors.  If an error is logged outside of a submission, the submission ID is "0".
```
rainersigwald: Some of these comment lines are awful long, Mr. Ultra-Wide-Screen :)

src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
```diff
@@ -308,6 +354,9 @@ private void RaiseMessageEvent(object sender, BuildMessageEventArgs buildEvent)
         /// <exception cref="Exception">ExceptionHandling.IsCriticalException exceptions will not be wrapped</exception>
         private void RaiseErrorEvent(object sender, BuildErrorEventArgs buildEvent)
         {
+            // Keep track of build submissions that have logged errors.  If there is no build context, add a submission ID of "-1"
+            BuildSubmissionIdsThatHaveLoggedErrors.Add(buildEvent != null && buildEvent.BuildEventContext != null ? buildEvent.BuildEventContext.SubmissionId : -1);
```
rainersigwald:
```C#
buildEvent?.BuildEventContext?.SubmissionId ?? -1
```
?

src/XMakeBuildEngine/BackEnd/Components/Logging/ILoggingService.cs
```diff
@@ -461,7 +491,7 @@ bool HaveLoggedBuildFinishedEvent
         void Consume(BuildEventArgs buildEvent);
 
         /// <summary>
-        /// Shutsdown the sink and any resources it may be holding
+        /// Shuts down the sink and any resources it may be holding
```
rainersigwald: ❤️

src/XMakeBuildEngine/UnitTests/BackEnd/EventSourceSink_Tests.cs
```diff
@@ -83,6 +83,93 @@ public void ConsumeEventsGoodEventsNoHandlers()
...
+            Assert.Equal(expectedBuildEvent.HelpKeyword, actualBuildEvent.HelpKeyword);
+            Assert.Equal(expectedBuildEvent.Message, actualBuildEvent.Message);
+            Assert.Equal(expectedBuildEvent.SenderName, actualBuildEvent.SenderName);
+            Assert.Equal(expectedBuildEvent.ColumnNumber, actualBuildEvent.ColumnNumber);
+            Assert.Equal(expectedBuildEvent.EndColumnNumber, actualBuildEvent.EndColumnNumber);
+            Assert.Equal(expectedBuildEvent.EndLineNumber, actualBuildEvent.EndLineNumber);
+            Assert.Equal(expectedBuildEvent.LineNumber, actualBuildEvent.LineNumber);
+            Assert.Equal(expectedBuildEvent.BuildEventContext, actualBuildEvent.BuildEventContext);
+            Assert.Equal(expectedBuildEvent.ThreadId, actualBuildEvent.ThreadId);
+            Assert.Equal(expectedBuildEvent.Timestamp, actualBuildEvent.Timestamp);
```
rainersigwald: Yikes! But I guess it's not worth implementing `IEquatable` just for this test.

src/XMakeCommandLine/Resources/Strings.resx
```diff
@@ -574,6 +574,25 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+                     or a comma to separate multiple warning codes. To treat all
+                     warnings as errors use the switch with no values.
+                     (Short form: /err[:c;[c2]])
+
+                     Example:
+                       /warnaserror:MSB4130
+
+                     When a warning is treated as an error the target will
+                     continue to execute as if it was a warning but the overall
+                     build result will show as failed.
```
rainersigwald: "but the overall build will fail." seems a bit clearer.

src/XMakeCommandLine/UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -1153,6 +1156,129 @@ public void ExtractAnyLoggerParameterPickLast()
...
+                "/warnaserror:a",             // A different source with a single duplicate
+                "/warnaserror:a,b",           // A different source with  multiple duplicates
+                "/warnaserror",               // Nothing specified
+            }), commandLineSwitches);
+
+            ISet<string> actualWarningsAsErrors = MSBuildApp.ProcessWarnAsErrorSwitch(commandLineSwitches);
+
+            Assert.NotNull(actualWarningsAsErrors);
+
+            Assert.Equal(expectedWarningsAsErors, actualWarningsAsErrors, StringComparer.OrdinalIgnoreCase);
```
rainersigwald: Is this test implying that the switches are _additive_? I find that a bit surprising. And definitely I find it surprising that if the last one is a no-code-specified-turn-everything-on setting, that doesn't win.

src/XMakeCommandLine/UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -1153,6 +1156,129 @@ public void ExtractAnyLoggerParameterPickLast()
...
+            MSBuildApp.GatherCommandLineSwitches(new ArrayList(new [] { "/warnaserror" }), commandLineSwitches);
+            
+            ISet<string> actualWarningsAsErrors = MSBuildApp.ProcessWarnAsErrorSwitch(commandLineSwitches);
+
+            Assert.NotNull(actualWarningsAsErrors);
+
+            Assert.Equal(0, actualWarningsAsErrors.Count);
+        }
+
+#if !RUNTIME_TYPE_NETCORE
```
rainersigwald: Why?

src/XMakeCommandLine/XMake.cs
```diff
@@ -1966,7 +1975,7 @@ bool recursing
             }
 
             ErrorUtilities.VerifyThrow(!invokeBuild || !String.IsNullOrEmpty(projectFile), "We should have a project file if we're going to build.");
-
+
```
rainersigwald: 👻 trailing whitespace

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
jeffkl: I did some limited research on what was faster and from what I could tell there was no difference.  Do you know of any guidance on which is faster?  I wanted the fastest possible lookup for sure.

src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
jeffkl: This set is the warning codes to treat as errors.  When the `EventSourceSink` is created, this property is set to null, an empty collection, or a collection with items.  There is no magic number for this scenario.

src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
jeffkl: I use a Chrome extension to make GitHub wider too so I never notice :/

jeffkl: I knew I left a comment with 0, this should say -1.  I thought the build submission IDs started at 1 when I coded this up but later figured out they start at zero.

src/XMakeBuildEngine/BackEnd/Components/Logging/EventSourceSink.cs
jeffkl: Great suggestion, I looked at it and ReSharper didn't suggest this so I just moved on.  I will update.

src/XMakeBuildEngine/UnitTests/BackEnd/EventSourceSink_Tests.cs
jeffkl: Yeah I thought about making a `Clone()` method as well but I didn't want to go there...  This unit test made sure I was properly copying every single property when I mutate the warning into an error (and I did miss one :smile:)

src/XMakeCommandLine/Resources/Strings.resx
jeffkl: Haha that's what I had originally and then I second guessed myself :sad:

src/XMakeCommandLine/UnitTests/CommandLineSwitches_Tests.cs
jeffkl: I was imaging a scenario where the list of warning codes was specified in one or more response files, an environment variable and the command-line.  I figured I would just conglomerate them into one set.  If this is the case, then specifying the switch by itself doesn't add any more codes and some codes already exist so the functionality is to only treat the specified codes as errors.

This is sort of how `/p:foo=bar` works minus the ability to clear them all by specifying just `/p`.  Let's discuss this tomorrow.

src/XMakeCommandLine/UnitTests/CommandLineSwitches_Tests.cs
jeffkl: `ResourceManager.GetResourceSet()` doesn't exist on .NET core.  This test is just making sure I typed up the new command-line argument usage correctly and so it technically doesn't matter if it isn't running on .NET Core since it's testing all of the switches.  The RESX resources are not `#ifdef`'d out, only the display of them.

Should I put a comment?

src/XMakeCommandLine/XMake.cs
jeffkl: grr

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
rainersigwald: AFAIK it's highly dependent on the individual use case, so we probably can't tell for this case without extensive benchmarking.

My understanding is based on [Pike's third rule](http://users.ece.utexas.edu/~adnan/pike.html): Fancy algorithms are slow when n is small, and n is usually small. In this case, a hashtable is awesomely super fast, linear lookup . . . but requires hashing and indirection/pointer-following. A sorted set that is small, as I expect this thing to be, could easily use fewer instructions per lookup.

But we'd have to measure to know (Rule 2!) and I don't think this will be worth worrying about, so  I'd say to leave it as-is.

src/XMakeCommandLine/UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -36,8 +38,9 @@ public void BogusSwitchIdentificationTests()
             bool multipleParametersAllowed;
             string missingParametersErrorMessage;
             bool unquoteParameters;
+            bool allowEmptyParmaters;
```
rainersigwald: typo: `Parmaters`

src/XMakeCommandLine/CommandLineSwitches.cs
```diff
@@ -481,12 +491,22 @@ bool unquoteParameters
             // check if the switch has multiple parameters
             if (multipleParametersAllowed)
             {
-                // store all the switch parameters
-                int emptyParameters;
-                _parameterizedSwitches[(int)parameterizedSwitch].parameters.AddRange(QuotingUtilities.SplitUnquoted(switchParameters, int.MaxValue, false /* discard empty parameters */, unquoteParameters, out emptyParameters, s_parameterSeparators));
+                if (String.Empty.Equals(switchParameters) && emptyParametersAllowed)
+                {
+                    // Store an null parameter if its allowed
```
rainersigwald: typo: an -> a

src/XMakeCommandLine/UnitTests/CommandLineSwitches_Tests.cs
rainersigwald: For posterity: we discussed, and decided that the additive model makes sense, but a new no-code-specified setting should override previous specific ones. If you need to clear the setting you can do `/warnaserror /warnaserror:X` to get back to a "just X" state.

src/XMakeCommandLine/XMake.cs
```diff
@@ -2029,6 +2040,37 @@ internal static TextWriter ProcessPreprocessSwitch(string[] parameters)
             return writer;
         }
 
+        internal static ISet<string> ProcessWarnAsErrorSwitch(CommandLineSwitches commandLineSwitches)
+        {
+            // TODO: Parse an environment variable as well?
```
rainersigwald: I don't feel strongly about this. Doesn't seem worth it.

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
jeffkl: I just did some benchmarks.  If you have 10 warnings to set as errors and your build logs 10,000 warnings, the fastest is to use `HashSet<T>(StringComparer.OrdinalIgnoreCase)`.  I tried doing a case sensitive search but things placed in the set are upper case and the item being checked for is uppercased before the check but that was slower.  The `SortedSet<T>` was just a little bit slower.

In my scenario, it took 1.04 milliseconds to check 10,000 times for the 10 items.  I think the perf here is good enough but at least know we know. :emoji-for-the-'the-more-you-know'-public-service-annoncement:

jeffkl:
I'm going to squash this and re-push.  When we're ready I'd like to do a merge commit so I can start on the /NoWarn switch and not have any merge conflicts.

src/XMakeBuildEngine/BackEnd/BuildManager/BuildParameters.cs
rainersigwald: Nice!

cdmihai: Alternatively, declare this as an ImmutableHashset and just copy the reference rather than cloning a new one. Theoretically they are slower, with logn lookup. Can you benchmark this one as well? :)

---
In reply to: [88770185](https://github.com/Microsoft/msbuild/pull/1355#discussion_r88770185) [](ancestors = 88770185)

PR #1366, opened 2016-11-21
Implement /NoWarn functionality

### LINKED ISSUE #47: Support disabling XML comment warnings (1591,1573) when run against a .sln
MSBuild does not currently offer any way to pass-through properties to the projects within a .sln file.

[Despite the documentation](https://msdn.microsoft.com/en-us/library/ms164311.aspx), properties provided at the command line with a solution file are ignored. 

This makes it very difficult to disable XML warnings on a CI server in order to reduce output noise. Many developers feel that 30% of their lines of code should not be  `/// <summary>`, and that code noise (like output noise) must be justified .

### LINKED ISSUE #68: Add a "treat warnings as errors" option

### LINKED ISSUE #910: Ignoring multiple imports
Importing a target/props file multiple times causes a MSB4011 warning. The only mechanism to avoid it is to condition the import to happen only when some property is empty. This places the burden on the importer, and cannot be done for transitive imports.

I propose that a mechanism is added for project files to suppress this warning:

```
<Project DefaultTargets="Build" ToolsVersion="15.0" IgnoreMultipleImport="true">
```

jeffkl:
You cannot suppress all warnings, if you specify empty `/warnasmessage` or `/nowarn`, you get an error message.
Warnings are mutated into just a message.
Ignore the first commit which is the implementation of /WarnAsError

Closes #68
Closes #47
Closes #910 

So this:
```
MyProject.proj(3,18): warning MSB4130: The condition "true or true and true" may have been evaluated incorrectly 
in an earlier version of MSBuild. Please verify that the order of the AND and OR clauses is written as intended. To 
avoid this warning, add parentheses to make the evaluation order explicit.
```
Becomes this in with Verbosity=Detailed 
```
MyProject.proj(3,18): message MSB4130: The condition "true or true and true" may have been evaluated incorrectly
in an earlier version of MSBuild. Please verify that the order of the AND and OR clauses is written as intended. To
avoid this warning, add parentheses to make the evaluation order explicit.
```

Here is the help message for review:
```
  /warnasmessage[:code[;code2]]
                     List of warning codes to treats as low importance
                     messages.  Use a semicolon or a comma to separate
                     multiple warning codes.
                     (Short form: /nowarn[:c;[c2]])

                     Example:
                       /warnasmessage:MSB3026
```

### DIFF
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -941,6 +941,7 @@ namespace Microsoft.Build.Execution
         public System.Globalization.CultureInfo UICulture { get { throw null; } set { } }
         public bool UseSynchronousLogging { get { throw null; } set { } }
         public System.Collections.Generic.ISet<string> WarningsAsErrors { get { throw null; } set { } }
+        public System.Collections.Generic.ISet<string> WarningsAsMessages { get { throw null; } set { } }
         public Microsoft.Build.Execution.BuildParameters Clone() { throw null; }
         public Microsoft.Build.Evaluation.Toolset GetToolset(string toolsVersion) { throw null; }
     }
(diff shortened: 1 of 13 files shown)


jeffkl:
@dotnet-bot test this please

jeffkl:
@dotnet-bot test this please

PR #1367, opened 2016-11-21
Remove escape-hatch for NoReferenceAssemblyDirectoryFound

rainersigwald:
When this message was converted to a warning (for #173), we added an escape-hatch environment variable to revert to the previous warning-only behavior. A comment indicated that the escape hatch should be removed before the next major release, so I'm removing it.

@AndyGerlicher


AndyGerlicher:

rainersigwald:
@AndyGerlicher No, it's a real problem: you added a test to make sure the escape hatch works and I didn't remove that now-obsolete test. Will get back to this eventually.

PR #1373, opened 2016-11-21
Alter TEMP folder when running in CI

jeffkl:
Hoping this will insulate us a little bit from problems we're seeing in CI where things in TEMP are messed up.

### DIFF
--- a/cibuild.sh
+++ b/cibuild.sh
@@ -14,8 +14,6 @@ usage()
 }
 
 restoreBuildTools(){
-    # Work around issue where NuGet can't restore  because it can't get a lock
-
     eval "$THIS_SCRIPT_PATH/init-tools.sh"
 }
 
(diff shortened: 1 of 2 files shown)


cibuild.sh
```diff
@@ -14,6 +14,8 @@ usage()
 }
 
 restoreBuildTools(){
+    # Work around issue where NuGet can't restore  because it can't get a lock
+
```
rainersigwald: How does this work around?

jeffkl: By placing TEMP under WORKSPACE, then we know that Jenkins will clean it up.  This at least gives us a clean TEMP folder for each build.

rainersigwald: But this is just a comment. What does it apply to?

jeffkl: Whooops I typed this comment, added code, and then moved the code without the comment :sad:

jeffkl:
It turns out that the the workspace path has `---` in it which breaks some of our tests which try to write the path as an XML comment.  So I'm going to just do this for non-Windows for now.

What we should do in the long run is have a known TEMP path that gets deleted before and after the CI build.

PR #1374, opened 2016-11-22
Compare strong names only when both assemblies are strong named

### LINKED ISSUE #1314: XmakeTasks nativemethods did not port clr implementation correcly.
The method RefToDef looks like it was supposed to be a copy of 
https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/vm/baseassemblyspec.cpp

However, in the implementation it is not returning false if the ref is strong named but dev is not strong named. 

This is causing the build of some of our project to fail due to a null reference exception happening in CompareRefToDef because def has no strong name and will be null when retrieved.

if (refStrongNamed) 
            { 
                 var defPkt = def.GetPublicKeyToken(); 
                bool defStrongNamed = defPkt != null && defPkt.Length != 0; 
               <Should return false here if defStrongNamed is false> 
 
                 return CompareRefToDef(@ref, def); 
          }

cdmihai:
Fixes #1314

### DIFF
--- a/src/XMakeTasks/NativeMethods.cs
+++ b/src/XMakeTasks/NativeMethods.cs
@@ -1171,19 +1171,20 @@ typedef enum _tagAssemblyComparisonResult
         //  Based on coreclr baseassemblyspec.cpp (https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/vm/baseassemblyspec.cpp#L330)
         private static bool RefMatchesDef(AssemblyName @ref, AssemblyName def)
         {
-            var refPkt = @ref.GetPublicKeyToken();
-            bool refStrongNamed = refPkt != null && refPkt.Length != 0;
-            if (refStrongNamed)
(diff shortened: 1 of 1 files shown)


cdmihai:
@crmann1

src/XMakeTasks/NativeMethods.cs
```diff
@@ -1171,21 +1171,22 @@ internal static bool AllDrivesMapped()
...
-                return CompareRefToDef(@ref, def);
+                return IsStrongNamed(def) && CompareRefToDef(@ref, def);
             }
             else
             {
                 return @ref.Name.Equals(def.Name, StringComparison.OrdinalIgnoreCase);
             }
         }
 
+        private static bool IsStrongNamed(AssemblyName assembly)
```
rainersigwald: Is this what coreclr does? It wasn't immediately obvious from looking at their code. Link in comment?

cdmihai:

PR #1378, opened 2016-11-22
Fix issue with xml file line ending normalization

### LINKED ISSUE #1340: MSBuild Converts block comment line endings to `\n`, from `\r\n`
As the title says, when I save a project from VS, the line endings of block comments go from `\r\n` to `\n`. A simple example:

1. Create a blank .NET Core project (doesn’t matter what type, I used a console app).
2. In a separate editor, with the project open, create a block comment. Observe that the line endings are \r\n.
3. Switch back to VS, left-click on the project node, and hit CTRL+S.
4. Switch back to the project file in an external editor. Line endings are now \n for the block comment.

This creates a bad experience in VS, as VS will helpfully pop up a line ending normalization dialog when the project file is opened directly. I've only tested this is .NET Core apps, but I suspect that it'll hold true for others as well. Moved from https://github.com/dotnet/roslyn-project-system/issues/742.

AndyGerlicher:
I've spent way too much time on this already. Pushing to get feedback or ideas. You can see I tried to set _normalize = false. This worked on Full Framework, but seems to do nothing on .NET Core. Whatever implementation I use seems to have some sort of bug and is not usable...

Background:
* Previous implementation on Full Framework used XmlTextReader(path). This contained an issue with certain characters (#985) and was fixed by using streams (#1004). #1004 also changed from XmlTextReader to XmlReader.
* XmlReader contains logic to normalize line endings. Internally, it sets the Normalize property to true and replaces (some? all?) \r\n with \n.

This change switches implementation to use XmlTextReader with a stream on Full Framework. This class sets the internal Normalize to false and does not replace \r\n with \n. This fixes #1340 and keeps the fix @maddin2016 made for #985.

However, .NET Core does not ship with XmlTextReader, only XmlReader. #1340 still exists for .NET Core.

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
@@ -530,8 +530,8 @@ namespace Microsoft.Build.Construction
 
                         var parentIndentation = GetElementIndentation(XmlElement);
 
-                        var leadingWhitespaceNode = XmlDocument.CreateWhitespace("\n" + parentIndentation + DEFAULT_INDENT);
-                        var trailingWhiteSpaceNode = XmlDocument.CreateWhitespace("\n" + parentIndentation);
+                        var leadingWhitespaceNode = XmlDocument.CreateWhitespace(Environment.NewLine + parentIndentation + DEFAULT_INDENT);
+                        var trailingWhiteSpaceNode = XmlDocument.CreateWhitespace(Environment.NewLine + parentIndentation);
 
                         XmlElement.InsertBefore(leadingWhitespaceNode, child.XmlElement);
                         XmlElement.InsertAfter(trailingWhiteSpaceNode, child.XmlElement);
--- a/src/XMakeBuildEngine/Xml/XmlReaderExtension.cs
+++ b/src/XMakeBuildEngine/Xml/XmlReaderExtension.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using System.Xml;
+
+namespace Microsoft.Build.Internal
+{
+    /// <summary>
+    ///     Disposable helper class to wrap XmlReader / XmlTextReader functionality.
+    /// </summary>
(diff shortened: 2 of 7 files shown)


src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -530,8 +530,8 @@ internal void AddToXml(ProjectElement child)
 
                         var parentIndentation = GetElementIndentation(XmlElement);
 
-                        var leadingWhitespaceNode = XmlDocument.CreateWhitespace("\n" + parentIndentation + DEFAULT_INDENT);
-                        var trailingWhiteSpaceNode = XmlDocument.CreateWhitespace("\n" + parentIndentation);
+                        var leadingWhitespaceNode = XmlDocument.CreateWhitespace(Environment.NewLine + parentIndentation + DEFAULT_INDENT);
+                        var trailingWhiteSpaceNode = XmlDocument.CreateWhitespace(Environment.NewLine + parentIndentation);
```
rainersigwald: This seems fine, but it's probably not exactly right. We should have a file-format newline that's distinct from the current OS newline, or folks will be confused when projects created/modified on Windows and Linux are different.

src/XMakeBuildEngine/Xml/XmlReaderExtension.cs
```diff
@@ -0,0 +1,104 @@
...
+            return reader;
+#else
+            var xr = XmlReader.Create(input, new XmlReaderSettings {DtdProcessing = DtdProcessing.Prohibit});
+
+            // Set Normalization = false if possible. Without this, certain line endings will be normalized
+            // with \n (specifically in XML comments). Does not throw if if type or property is not found.
+            // This issue does not apply to XmlTextReader (above) which is not shipped with .NET Core yet.
+            
+            // NOTE: This doesn't work in .NET Core.
+            //var xmlReaderType = typeof(XmlReader).GetTypeInfo().Assembly.GetType("System.Xml.XmlTextReaderImpl");
```
rainersigwald: How does it fail? Just can't find `XmlTextReaderImpl` "next to" `XmlReader`?

PR #1380, opened 2016-11-22
Add fallback search path inside $(VsInstallRoot) for Target Frameworks

kzu:
dev15 added support for VSIXes to install custom target frameworks in an app-local
way, allowing for much better SxS support for tools like Xamarin.

### DIFF
--- a/src/XMakeCommandLine/app.config
+++ b/src/XMakeCommandLine/app.config
@@ -67,6 +67,7 @@
         <property name="MSBuildToolsRoot" value="$(VsInstallRoot)\MSBuild" />
         <property name="MSBuildExtensionsPath" value="$(VsInstallRoot)\MSBuild" />
         <property name="MSBuildExtensionsPath32" value="$(VsInstallRoot)\MSBuild" />
+        <property name="TargetFrameworkRootPathSearchPathsWindows" value="$(VsInstallRoot)\Common7\IDE\ReferenceAssemblies\Microsoft\Framework" />
 
         <property name="RoslynTargetsPath" value="$([MSBuild]::GetToolsDirectory32())\Roslyn" />
 


src/XMakeCommandLine/app.config
```diff
@@ -67,6 +67,7 @@
         <property name="MSBuildToolsRoot" value="$(VsInstallRoot)\MSBuild" />
         <property name="MSBuildExtensionsPath" value="$(VsInstallRoot)\MSBuild" />
         <property name="MSBuildExtensionsPath32" value="$(VsInstallRoot)\MSBuild" />
+        <property name="TargetFrameworkRootPathSearchPathsWindows" value="$(VsInstallRoot)\Common7\IDE\ReferenceAssemblies\Microsoft\Framework" />
```
cdmihai: Why define these here with OS specific names and not reuse the current fallback infrastructure via the `projectImportSearchPaths` node? Using that one, you'd define the same `TargetFrameworkRootPathSearchPaths` under each OS

AndyGerlicher:
We discussed this during our triage and have some significant concern on how this is implemented. I don't think this needs to be a change that is on the engine side. And by engine side I mean special logic for "fallbacks". Trying to come up with a better solution, could this be plumed through target/tasks? It looks like this could be defined in a property and passed into `GetReferenceAssemblyPaths` as a `;` delimited list. I think it would be very appropriate to extend that Task to offer a list of inputs rather than just `RootPath` for example (add `RootPaths` and pass in the same property for example). With that change this could be entirely done in targets potentially w/o modification to the toolset. Although, I'm not opposed to defining the paths in the toolset if this is a normal thing going forward. 

The main code smells I have with it is the need to use reflection and the need for `ToolLocationHelper` to read the toolset where it never has before. I think the approach above would eliminate those concerns.

Thoughts? Am I missing a scenario? I know this is rushed and under time crunch, we could do a conference call tomorrow morning if needed.

radical:
The advantage of doing it on the engine side is that it becomes available *everywhere* and not just in one specific task. Any other task or any other code trying to use `ToolLocationHelper.GetPathToReferenceAssemblies` will immediately support these fallback paths. Other than that, there is the case of `FrameworkPathOverride` also using this to find a suitable framework location.

If the approach is acceptable, then we could discuss how to improve the implementation. This is why I had the `[RFC]` in the commit!

src/Shared/FrameworkLocationHelper.cs
```diff
@@ -990,6 +990,91 @@ private static string GetPathToBuildToolsFromEnvironment(DotNetFrameworkArchitec
...
+
+            return configurationSection?.Toolsets.GetElement(toolsVersion);
+        }
+#endif
+
+        /// <summary>
+        /// OS name that can be used as the suffix for `TargetFrameworkRootPathSearchPaths` property name
+        /// in app.config
+        /// </summary>
+        private static string GetOSNameForTargetFrameworkRoot()
```
radical: The reason for duplication is the difference in casing. I will move it to NativeMethodsShared.

radical:
I opened a new PR for this main commit - https://github.com/Microsoft/msbuild/pull/1389 , so that I can update that as required.

kzu:
I'll rebase and add on top of @radical PR just the console config, which is what we need for VS2017

kzu:
Removed the commit that is now part of https://github.com/Microsoft/msbuild/pull/1389. 

This commit only adds the fallback property for the VsInstallRoot.

AndyGerlicher:
Closing out this old PR. I don't think this is needed until we settle on the design in PR #1389. Let me know if you think otherwise and we can reopen.

Issue #297, opened 2015-10-19
Failing tests in GenerateResource_Tests due to missing state file


cdmihai:
Around 19 tests are failing because the netcore version of GenerateResource does not serialize its cache. This disables incremental building for the task.

cdmihai:
This feature is currently not supported in .net core. I'm wrapping the state asserts with a preprocessor flag to make them run.

BenVillalobos:
Closing because stale

PR #1381, opened 2016-11-22
Fix (almost) build with mono/OSX

radical:


radical:
Build failure on windows isn't related:

```
18:36:53 Notifying endpoint with id 'HTTP:helix-int-notification-url'
18:37:23 ERROR: Failed to notify endpoint with id 'HTTP:helix-int-notification-url'
18:37:23 java.net.SocketTimeoutException: Read timed out
```

radical:
Scartch that, the issue seems to be:

```
18:43:42 ** Rebuilding MSBuild with locally built binaries
18:43:42 
18:43:42 ** Using the MSBuild in path: "D:\j\workspace\_Windows_NT_D---eb124103\bin\Bootstrap\15.0\Bin\MSBuild.exe"
18:43:42 ** Using runtime host in path: 
18:43:42 ** ""D:\j\workspace\_Windows_NT_D---eb124103\bin\Bootstrap\15.0\Bin\MSBuild.exe"" /nodeReuse:false "D:\j\workspace\_Windows_NT_D---eb124103\build.proj" /m /verbosity:minimal /t:RebuildAndTest /p:Configuration=Debug "/p:LocalizedBuild=true" /fileloggerparameters:Verbosity=diag;LogFile="D:\j\workspace\_Windows_NT_D---eb124103\msbuild_local_build-Full.log"
18:43:42 '""D:\j\workspace\_Windows_NT_D---eb124103\bin\Bootstrap\15.0\Bin\MSBuild.exe""' is not recognized as an internal or external command,
18:43:42 operable program or batch file.
18:43:42 ```

File: src/XMakeBuildEngine/Evaluation/Evaluator.cs [hunk 1]
```diff
@@ -1392,6 +1392,9 @@ private ICollection<P> AddBuiltInProperties()
 
 #if RUNTIME_TYPE_NETCORE
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.msbuildRuntimeType, "Core"));
+#elif MONO
+            builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.msbuildRuntimeType,
+                                                        NativeMethodsShared.IsMono ? "Mono" : "Full"));
```
cdmihai: Under what circumstance is the `MONO` symbol defined, but `NativeMethodsShared.IsMono` returns false?

src/XMakeTasks/UnitTests/GenerateResourceOutOfProc_Tests.cs
```diff
@@ -1895,8 +1895,8 @@ public void SourceItemMissing()
         ///  Non-existent StateFile yields message
         /// </summary>
         [Fact]
-        [Trait("Category", "mono-osx-failing")]
         [Trait("Category", "mono-windows-failing")]
+        [PlatformSpecific(Xunit.PlatformID.Windows)]
```
cdmihai: Why make this windows specific? Previous traits suggests that it only affected mono on windows and osx.

src/XMakeTasks/UnitTests/GenerateResource_Tests.cs
```diff
@@ -2071,6 +2071,7 @@ public void SourceItemMissing()
 #else
         [Fact(Skip = "https://github.com/Microsoft/msbuild/issues/297")]
 #endif
+        [PlatformSpecific(Xunit.PlatformID.Windows)]
```
cdmihai: Same as above. @rainersigwald, what was the resolution for the state file? does it exist / work on .net core? I remember @danmosemsft did something about it.

danmosemsft: IIRC the state file's purpose was to record any links to files from the resx, so that if the resx timestamp has not changed we can use the state file to get the list of other files to timestamp check. If that's the case then it's not relevant to xplat as resx there can't link files and the code in GR for that is ifdeffed out.

Also it relies on binary serialization which first appears in a corefx version post 1.1 so I suspect it's not available to you.

cdmihai: VIa this conversation maybe we can also revisit #297 for some more investigation / resolution.

rainersigwald: Yeah, the state file exists only when we have the full-fledged resource reader, meaning we have links to remote files to follow. I think we're as good as we can be for now; in a glorious future where our .NET Core GenerateResource task can do all of the things its Full flavor can, we should enable these tests everywhere.

The Mono implications aren't clear to me, but dropping down to the .NET Core behavior seems reasonable.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
radical: Built to target Mono and running on .NET runtime.

src/XMakeTasks/UnitTests/GenerateResource_Tests.cs
radical: This test creates a read-only (removes the write permission) state file and then expects the task to fail while trying to write to such a file. But the [code](https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeTasks/StateFileBase.cs#L40) (`GenerateResource.WriteStateFile -> ResGenDependencies.SerializeCache -> StateFileBase.SerializeCache`) that does this, deletes the file before writing to it.
- On Unix, it is legal to delete such a file. So, subsequently trying to write to that path succeeds and the test fails
- IIUC, trying to delete or write to this read-only would fail on windows, thus satisfying the test.

I should add this to the commit message.

radical: Also, the commit this belongs to, enables binary serialization for the mono builds. https://github.com/Microsoft/msbuild/pull/1381/commits/ad50a5356f1be61a6ac2e740a6a1c10618112ec4

radical:
Can this be merged?

cdmihai:
@dotnet-bot test this please

dir.props
```diff
@@ -309,7 +309,7 @@
     <DefineConstants>$(DefineConstants);FEATURE_RESGEN</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_RESOURCE_EXPOSURE</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_RESX_RESOURCE_READER</DefineConstants>
-    <DefineConstants>$(DefineConstants);FEATURE_RTLMOVEMEMORY</DefineConstants>
+    <DefineConstants Condition="'$(MonoBuild)' != 'true'">$(DefineConstants);FEATURE_RTLMOVEMEMORY</DefineConstants>
```
rainersigwald: 😞 good fix

radical:
@rainersigwald yeah, I had opened an issue for the bootstrapper - https://github.com/Microsoft/msbuild/issues/1382 . Even after this, though, it won't completely build on Mono. There are a couple more fixes required for that. I can open a PR for those once the bootstrapper thing is also fixed.

PR #1383, opened 2016-11-22
Signing specific loc targets use DependsOnTargets

cdmihai:
Because AfterTargets does not guarantee that the target will be executed
after its AfterTargets, but DependsOnTargets does.

This lead to the `AddSatelliteAssembliesToFilesToSignItem` target being executed before the item it read got computed.


cdmihai:

PR #1387, opened 2016-11-23
Allow override of VCTargetsPath property.

### LINKED ISSUE #1384: Msbuild.exe.config does not define MSBuildProgramFiles32 and VC specific props properly
Currently Msbuild.exe.config contains

<msbuildToolsets default="15.0">
      <toolset toolsVersion="15.0">
       …
        
        <property name="VCTargetsPath" value="$(VsInstallRoot)\Common7\IDE\VC\VCTargets\" />
        <property name="VCTargetsPath14" value="$([MSBuild]::ValueOrDefault('$(VCTargetsPath14)','$(MSBuildProgramFiles32)\MSBuild\Microsoft.Cpp\v4.0\V140\'))" />
 
In Prev VS versions we used [MSBuild]::ValueOrDefault syntax in the registry to allow env vars overriding toolset properties, i.e. if VCTargetsPath14 is set as env var it would be used instead of $(MSBuildProgramFiles32)\MSBuild\Microsoft.Cpp\v4.0\V140.
 
We need the ensure we have the same to Dev15 VCTargetsPath, as well as AndroidTargetsPath and iOSTargetsPath.

Also, VCTargetsPath14 and a number of other props as well as are using MSBuildProgramFiles32, which is empty. It either needs to be defined in Msbuild.exe.config or some other prop is used instead of it.

AndyGerlicher:
@olgaark look right?

Fixes #1384

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
+++ b/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
@@ -395,6 +395,11 @@ namespace Microsoft.Build.Evaluation
             return BuildEnvironmentHelper.Instance.VisualStudioInstallRootDirectory;
         }
 
+        public static string GetProgramFiles32()
+        {
+            return FrameworkLocationHelper.programFiles32;
+        }
+
         #region Debug only intrinsics
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


olgaark:
Leave a comment

olgaark:
VCTargetsPath looks good, but #1384 is also about MSBuildProgramFiles32 not being set and thus VCTargetsPath14 and some other properties not defined correctly.

Is it possible to define MSBuildProgramFiles32 there as well somehow?

AndyGerlicher:
@olgaark You're right. It was used below but in the fallback search paths is evaluated when that property exists. I added an intrinsic function and use that rather than the property.

Now does it look ok?

olgaark:
Looks great. Thanks!

PR #1388, opened 2016-11-23
Add preserveFormatting overloads. Expose PreserveFormatting as read-only property

jviau:
1. Add extra overloads for `ProjectRootElement.TryOpen` with `preserveFormatting` options.
2. Expose `PreserveFormatting` as a read-only property on `ProjectRootElement`. CPS needs this for the live-reload scenario.

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -1125,7 +1125,7 @@ namespace Microsoft.Build.Construction
         /// Uses the specified project collection.
         /// </summary>
         /// <param name="path">The path of the ProjectRootElement, cannot be null.</param>
-        // <param name="projectCollection">The <see cref="ProjectCollection"/> to load the project into.</param>
+        /// <param name="projectCollection">The <see cref="ProjectCollection"/> to load the project into.</param>
         /// <param name="preserveFormatting">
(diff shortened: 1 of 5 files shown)


AndyGerlicher:
Looks good to me except I think you need to sign the CLA. Did you get a message about that and where to sign it?

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1106,6 +1117,25 @@ public static ProjectRootElement TryOpen(string path)
...
+        // <param name="projectCollection">The <see cref="ProjectCollection"/> to load the project into.</param>
+        /// <param name="preserveFormatting">
+        /// The formatting to open with. Must match the formatting in the collection to succeed.
+        /// </param>
+        /// <returns>The loaded ProjectRootElement, or null if it is not currently in memory.</returns>
+        /// <remarks>
+        /// It is possible for ProjectRootElements to be brought into memory and discarded due to memory pressure. Therefore
+        /// this method returning false does not indicate that it has never been loaded, only that it is not currently in memory.
+        /// </remarks>
+        public static ProjectRootElement TryOpen(string path, ProjectCollection projectCollection, bool preserveFormatting)
```
cdmihai: Shouldn't the `preserveFormatting` parameter be piped into the TryGet call?

PR #1389, opened 2016-11-23
Add support for fallback search paths for Target frameworks

radical:
When running on Mono, target frameworks are searched for in
`$mono_prefix/lib/mono/xbuild-frameworks`, which becomes
`/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks`.

a *new* `/Library/Frameworks/Mono.framework/Versions/@VERSION@` directory
"lost" (For example, installed by Xamarin.Android). The file is still there, but xbuild
will no longer find it.

To address this upgrade scenario, Mono's xbuild also checks a fallback path:

    `/Library/Frameworks/Mono.framework/External/xbuild-frameworks`

Mono would like to eventually migrate to MSBuild instead of xbuild, which
requires bringing xbuild's fallback support to MSBuild.

Implementation:

- We add a new `$(TargetFrameworkFallbackSearchPaths)` property which
can have the list of fallback search paths and gets used by the `GetReferenceAssemblyPaths`.

- For Mono/OSX case, we point that to `/Library/Frameworks/Mono.framework/External/xbuild-frameworks`

  So, effectively, target frameworks are looked up, in order:

  1. `$(TargetFrameworkRootPath)` or the default location, if
     `$(TargetFrameworkRootPath)` is ''

  2. If not found in (1), then try to find it in the fallback search
     paths specified in `$(TargetFrameworkFallbackSearchPaths)`

### DIFF
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -214,6 +214,12 @@ namespace Microsoft.Build.Utilities
         /// </summary>
         private static List<string> s_targetFrameworkMonikers = null;
 
+        /// <summary>
+        /// Character used to separate search paths specified for MSBuildExtensionsPath* in
+        /// the config file
+        /// </summary>
+        private static char _separatorForFallbackSearchPaths = ';';
+
         private const string retailConfigurationName = "Retail";
         private const string neutralArchitectureName = "Neutral";
         private const string commonConfigurationFolderName = "CommonConfiguration";
@@ -1712,7 +1718,7 @@ namespace Microsoft.Build.Utilities
         /// <returns>Collection of reference assembly locations.</returns>
         public static string GetPathToStandardLibraries(string targetFrameworkIdentifier, string targetFrameworkVersion, string targetFrameworkProfile, string platformTarget)
         {
-            return GetPathToStandardLibraries(targetFrameworkIdentifier, targetFrameworkVersion, targetFrameworkProfile, platformTarget, null);
+            return GetPathToStandardLibraries(targetFrameworkIdentifier, targetFrameworkVersion, targetFrameworkProfile, platformTarget, null, null);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


borgdylan:
I think that this PR does not support Linux. On Linux one must search ```/usr/local/lib/mono/xbuild-frameworks``` and then ```/usr/lib/mono/xbuild-frameworks``` . More path would need searching is installing mono into /opt or in a snap/flatpak.

radical:
@borgdylan This is adding support for *fallback* search paths. When looking for frameworks, msbuild will look in the default location first, which would be `$mono_prefix/lib/mono/xbuild-frameworks` and if it cannot find it there, *then* it will look in the fallback paths. And so, no extra search paths are required on Linux, by default. It would be incorrect to look for frameworks in other mono installations on the machine, as these frameworks are closely tied to the corresponding mono installations.

The case on OSX is slightly different.

```
Target frameworks are searched for in `$mono_prefix/lib/mono/xbuild-frameworks`, which becomes 
`/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks`.

a *new* /Library/Frameworks/Mono.framework/Versions/@VERSION@ directory
"lost". The file is still there, but xbuild will no longer find it.

To address this upgrade scenario, Mono's xbuild also checks a fallback path:

    `/Library/Frameworks/Mono.framework/External/xbuild-frameworks`

Mono would like to eventually migrate to MSBuild instead of xbuild, which
requires bringing xbuild's fallback support to MSBuild.
```

I adapted this from the commit message of an older [PR](https://github.com/Microsoft/msbuild/commit/970168) . It would be useful to add this to the commit message here too, I think. Though the feature proposal does stand on it's own merit too.

AndyGerlicher:
Have you given more thought to the comments I suggested on the other thread? The two main things I need changed before this goes is are the reflection and the magic properties. By magic I mean that if you append an OS onto the property it deep down "engine" code treats it differently. I think both of these can be solved by plumbing the values through targets and having `GetPathToReferenceAssemblies` take a list of strings in place of `targetFrameworkRootPath` (new overload, and update the task as well).

The only question then is where does the overrides that set multiple values go. Without more context it's hard for me to make the case that the common targets should be aware of this (as opposed to Xamarin tool specific targets). I hadn't heard of app-local reference assemblies before so I can't speak to whether anyone else would want this functionality. If you think it will be more common then I'm not opposed to defining some of the fallback in targets or even the app.config if needed.

Can you investigate this approach and get back to us?

kzu:
Heya @radical it would be great to reach an agreement on this with @AndyGerlicher so that this makes it to the next release and we don't have to workaround it by RTM ;)

kzu:
BTW, @AndyGerlicher the app-local reference assemblies scenario comes from VS2017 now supporting app-local low-impact installation, which means things will more and more go into the VS install dir, rather than progfiles\Reference Assemblies. This should be the case for all the frameworks we ship, not only Xamarin, since otherwise nobody will have a good SxS story out of the box. 

See the section on [New Extension Capabilities](https://blogs.msdn.microsoft.com/visualstudio/2016/11/10/extensibility-in-visual-studio-15-increasing-reliability-and-performance/) for VS2017. That's the `ReferenceAssemblies` value in the `VSIX Sub Path` property. It would be quite important (I think) that that scenario is supported OOB, without target framework authors having to read even more documentation on how to provide override paths for a built-in product feature.

radical:
I agree with @AndyGerlicher's approach. And I will get back with an updated implementation.

radical:
Updated the PR with a completely new implementation. Ready for a fresh round of reviews :)

radical:
Any thoughts on this?

AndyGerlicher:
Sorry for the delay getting back! Most of us were out for the holidays and catching up.

Looks better. I think this approach is a lot cleaner. Thanks for changing directions.

Few things that should be fairly minor:
1. I prefer not to have fallback. It seems un-natural here. You're passing in a string that says look exactly here, but if that fails go ahead and look in all these other places too. It seems cleaner to just drop the fallback and pass in a list of places to look. Ideally typed as a list and if you pass in a string (legacy purposes) it would just pass that along as `new List<string> { singlePath }` to the new method that does the search that only takes a list.
2. Since you're changing the public interface make sure you build for other configurations. If you do you'll notice the ref cs files will be updated with the added methods. These need to be checked in as part of this change.
3. The other thing to note that does become more apparent if you look at that ref cs diff is adding an additional default property to a method changes the interface (it adds a required parameter with a default value). This means that apps compiled against the old code won't be able to provide that parameter and will break at run-time. I know it's uglier, but for changes to public methods can you please add an overload and keep all the existing methods?

AndyGerlicher:
Closing this for now. I'm in favor of the change in a generic way as I mentioned. Please do re-open when you'd like us to take another look. I would consider it for the 15.3 update time.

kzu:
This would actually be unnecessary if https://github.com/Microsoft/msbuild/issues/1749 is implemented

PR #1397, opened 2016-11-24
Add support cibuild in Visual Studio 2017 only environment.

mfilippov:


AndyGerlicher:
Thanks for this! We've been so deep in the changes to support .NET Core projects for VS 15 we haven't taken the time to do this. Which reminds me, I created #1399 to track conversion to new project types. Ultimately that's the goal but this is good as the new SDK is not ready quite yet.

It does look like this is a legit failure though. Did this repro locally? Not obvious to me what the issue is.

mfilippov:
I fix LocalizationTasks. I not sure about this fix because the task has a reference conflict. I use a conditional ItemGroup for fix it. Maybe a better way put a precompiled task in bootstrap and use it for the first build.

Samples/XmlFileLogger/XmlFileLogger.csproj
```diff
@@ -18,9 +18,6 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x86'" />
   <ItemGroup>
-    <Reference Include="Microsoft.Build, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
-    <Reference Include="Microsoft.Build.Framework, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
-    <Reference Include="Microsoft.Build.Utilities.Core, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
```
rainersigwald: I'm torn on this change. Should we instead move this project to build against the NuGet ref assemblies for 14.3? Then it could continue to live as an example of how a third-party might write a logger like this, and stay semi-separated from the internals of our build. @AndyGerlicher @jeffkl ?

build/LocalizationTasks/Microsoft.Build.LocalizationTasks.csproj
```diff
@@ -16,10 +16,26 @@
...
+    <Reference Include="Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>$(MSBuildToolsPath)\Microsoft.Build.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>$(MSBuildToolsPath)\Microsoft.Build.Framework.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.Build.Utilities.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>$(MSBuildToolsPath)\Microsoft.Build.Utilities.Core.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
rainersigwald: This solution is pretty nice, but we could potentially do the same NuGet package dance here (and get some dogfooding of version-redirecting type things).

Samples/XmlFileLogger/XmlFileLogger.csproj
cdmihai: Ideally. It would be cool if we changed all samples to use the nuget packages from the current build. This would allow us to actually test the packages to some degree. #800 talks about this.

jeffkl: I agree, these should come from NuGet since the projects are samples of what users should do.  They won't necessarily have MSBuild 14 installed or have our source code on their machines.

build/LocalizationTasks/Microsoft.Build.LocalizationTasks.csproj
jeffkl: These references should probably come from our own `.csproj`s right?  The assembly is only used during our build and we'd expect it to work with our current code I'm assuming.

@cdmihai what do you think?

rainersigwald: But don't we need the loc-tasks DLL _during_ our build before we've built new versions of our assemblies?

mfilippov:
Any news about this PR?

rainersigwald:
I think I'm more convinced that references to LKG packages should use NuGet rather than installed-and-GACed-if-available or ProjectReferences.

@jeffkl this overlaps with SDK migration work, I assume--since that will only be able to be built with MSBuild 15. What have you been doing in  your branch?

mfilippov:
Ok I'll wait new SDK in xplat.

jeffkl:
At the moment I haven't updated `cibuild.cmd` because I've only been testing within Visual Studio 2017.  But yes I'll have to update it to work soon as I migrate projects.  I think I'll have everything reference the NuGet packages because they'll still get MSBuild 14.0 or 15.0 at run time anyway.

We should also start the conversation about getting MSBuild 15.0 installed on the build agents in the CI environment soon.

rainersigwald:
@jeffkl I _think_ it'd be just pointing to this environment: https://github.com/dotnet/dotnet-ci/blob/d06dcf442b18c840fbf90a6b65dce7fe626eb20b/jobs/generation/Utilities.groovy#L215

Issue #1392, opened 2016-11-23
Implicit top & bottom SDK imports


cdmihai:
SDK imports represent the extension mechanism for MSBuild's build framework model: a certain build framework (e.g. UWP, .net core, web, etc) specifies what a build means by using an entry-point `.props` file and an entry-point `.targets` file.

A particular project implements a specific framework by putting its logic between and around the framework's entry-point imports. Its logic usually particularizes framework values, like the project name, references, etc. A project can implement multiple build frameworks by inserting their specific entry-point SDK imports (e.g. an F# based .net core web project).

With the move to .net core's human editable csproj files, the legacy top and bottom imports are no longer desirable, since they feel like magic values that are hard to remember and type from scratch:

```xml
<Project>
  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
...
  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />

</Project>
```

# Goals
1. Remove SDK PackageRef
2. Remove entry-point SDK imports
3. Remove `nuget restore` from sdk acquisition

## Impact on Templates
With the change in place, the .NET Core Library csproj template would become:
```xml
<Project ToolsVersion="15.0" Sdk="Microsoft.NET.Sdk/1.0.0-rc">
  <PropertyGroup>
    <TargetFramework>netstandard1.4</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="**\*.cs" />
    <EmbeddedResource Include="**\*.resx" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="NETStandard.Library" Version="1.6" />
  </ItemGroup>
</Project>
```

# Risks
Implicit `<Imports>` should not:
- get persisted to disk
- copied during cloning (except if they have an internal flag to mark them as syntactic sugar)
- Potential cache misses in ProjectCollection and ProjectRootElement caches
- Potential weird interaction with ProjectRootElement.Version

# Design
A new attribute is added to `<Project>` for representing SDKs: `Sdk="Microsoft.NET.Sdk/1.0.0-RC"`.
[...]

PR #1403, opened 2016-11-28
Implicit Sdk imports

rainersigwald:
First iteration of #1392.

Known issues:

* Doesn't define a default location for the SDKs. **MUST FIX**
* Allows `Sdk` attribute on any `<Project>` element, even ones that are (recursively) imported. That's confusing and wrong.
* Preprocessed output is not buildable (will double-import SDK imports)
* Not enough tests (I'll start working on this next, wanted to get this out)

Example of use:
`test.proj`

```xml
<Project Sdk="TestSdk">
  <PropertyGroup>
    <PropertyDefinedInProjectFromImportedProperty>Local/$(PropertyFromSdkImport)</PropertyDefinedInProjectFromImportedProperty>
  </PropertyGroup>

  <Target Name="Build">
    <Message Importance="High"
             Text="Build target. Prop: $(PropertyFromSdkImport) LocalProp: $(PropertyDefinedInProjectFromImportedProperty)" />
  </Target>
</Project>
```
`TestSdk\Sdk\Sdk.props`

```xml
<Project>
  <PropertyGroup>
    <PropertyFromSdkImport>Imported Property!</PropertyFromSdkImport>
  </PropertyGroup>
</Project>
```

`TestSdk\Sdk\Sdk.targets`

```xml
<Project>
  <Target Name="SdkTarget"
          BeforeTargets="Build">
    <Message Importance="High"
             Text="Target from the SDK" />
  </Target>
</Project>
```

```
S:\work\autoimports>s:\msbuild\bin\Bootstrap\15.0\Bin\MSBuild.exe test.proj
Microsoft (R) Build Engine version 15.1.425.23802

Build started 11/28/2016 2:58:47 PM.
Project "S:\work\autoimports\test.proj" on node 1 (default targets).
SdkTarget:
  Target from the SDK
Build:
  Build target. Prop: Imported Property! LocalProp: Local/Imported Property!
Done Building Project "S:\work\autoimports\test.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.55
```

preprocessor output:
```xml
<!--
============================================================================================================================================
S:\work\autoimports\test.proj
============================================================================================================================================
-->
<Project Sdk="TestSdk">
  <!--
============================================================================================================================================
  Import of "S:\work\autoimports\TestSdk\Sdk\Sdk.props" from Sdk "TestSdk" was implied by the Project element's Sdk attribute.

S:\work\autoimports\TestSdk\Sdk\Sdk.props
============================================================================================================================================
-->
  <PropertyGroup>
    <PropertyFromSdkImport>Imported Property!</PropertyFromSdkImport>
  </PropertyGroup>
  <!--
============================================================================================================================================
  </Import>

S:\work\autoimports\test.proj
============================================================================================================================================
-->
  <PropertyGroup>
    <PropertyDefinedInProjectFromImportedProperty>Local/$(PropertyFromSdkImport)</PropertyDefinedInProjectFromImportedProperty>
  </PropertyGroup>
  <Target Name="Build">
    <Message Importance="High" Text="Build target. Prop: $(PropertyFromSdkImport) LocalProp: $(PropertyDefinedInProjectFromImportedProperty)" />
  </Target>
  <!--
============================================================================================================================================
  Import of "S:\work\autoimports\TestSdk\Sdk\Sdk.targets" from Sdk "TestSdk" was implied by the Project element's Sdk attribute.

S:\work\autoimports\TestSdk\Sdk\Sdk.targets
============================================================================================================================================
-->
  <Target Name="SdkTarget" BeforeTargets="Build">
    <Message Importance="High" Text="Target from the SDK" />
  </Target>
  <!--
============================================================================================================================================
  </Import>

S:\work\autoimports\test.proj
============================================================================================================================================
-->
</Project>
```
```xml

### CODE REVIEW

File: src/XMakeBuildEngine/Construction/ProjectElementContainer.cs [hunk 1]
```diff
@@ -540,6 +545,15 @@ internal void AddToXml(ProjectElement child)
             }
         }
 
+        private static bool HasXmlRepresentation(ProjectElement element)
```
jeffkl: I think the `IsImplicit` property should be enough, do you think we need a method to call instead of checking the property?

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1581,6 +1587,40 @@ private static TempPaths GetTempPaths()
             }
         }
 
+        internal class TemporaryEnvironment : IDisposable
+        {
+            private bool _disposed;
+            private readonly string _name;
+            private readonly string _originalValue;
+
+            public TemporaryEnvironment(string name, string value)
```
jeffkl: Would it make sense for this to take an `IDictionary<string, string>` so you could have one or more variables set?

ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -83,6 +83,7 @@ public partial class ProjectExtensionsElement : Microsoft.Build.Construction.Pro
     public partial class ProjectImportElement : Microsoft.Build.Construction.ProjectElement
     {
         internal ProjectImportElement() { }
+        public bool Implicit { get { throw null; } }
```
jeffkl: `IsImplicit` sounds better to me but I'll leave it up to you.

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -186,6 +187,62 @@ private void ParseProjectElement(XmlElementWithLocation element)
...
             _project.SetProjectRootElementFromParser(element, _project);
 
+            if (element.HasAttribute(XMakeAttributes.sdk))
+            {
+                // TODO: don't get root of SDKs from the environment, use a built-in or toolset prop
+                // TODO: Don't null ref if MSBuildSDKsPath isn't defined
+
+                var sdksString = element.GetAttribute(XMakeAttributes.sdk);
+
+                var sdks = sdksString.Contains(";") ? sdksString.Split(';') : new[] {sdksString};
```
jeffkl: I'd use `.Split(';', RemoveEmptyEntries).Select(i => i.Trim()).Where(i => !String.IsNullOrWhiteSpace())` to be as robust as possible.

src/XMakeBuildEngine/Construction/ProjectImportElement.cs
```diff
@@ -44,26 +44,36 @@ public string Project
...
                 MarkDirty("Set Import Project {0}", value);
             }
         }
 
         /// <summary>
-        /// Location of the project attribute
+        /// Gets the Implicit state of the element: true if the element was not in the read XML.
         /// </summary>
-        public ElementLocation ProjectLocation
+        // TODO: *should* this be public? if it's not, you can't determine if an import is implicit from the public OM.
```
jeffkl: I think it should be public.  With no public setter, it seems safe.

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1714,8 +1714,23 @@ public void Save(Encoding saveEncoding)
...
                     using (ProjectWriter projectWriter = new ProjectWriter(_projectFileLocation.File, saveEncoding))
                     {
-                        projectWriter.Initialize(XmlDocument);
-                        XmlDocument.Save(projectWriter);
+                        var xmlWithNoImplicits = (XmlDocument)XmlDocument.Clone();
+
+                        var implicitElements =
+                            xmlWithNoImplicits.SelectNodes($"//*[@{XMakeAttributes.@implicit}]");
+
+                        if (implicitElements == null)
```
jeffkl: Would this ever happen?  I think if the document is parsed and successfully cloned, this would always return a value?

src/XMakeBuildEngine/Evaluation/Preprocessor.cs
```diff
@@ -183,8 +188,18 @@ private void CloneChildrenResolvingImports(XmlNode source, XmlNode destination)
...
-                        string importTag = "  <Import Project=\"" + importProject + "\"" + ((importCondition.Length > 0) ? " Condition=\"" + importCondition + "\"" : String.Empty) + ">";
-                        destination.AppendChild(destinationDocument.CreateComment("\r\n" + new String('=', 140) + "\r\n" + importTag + "\r\n\r\n" + resolved.FullPath.Replace("--", "__") + "\r\n" + new String('=', 140) + "\r\n"));
+                        string sdk = importSdk.Length > 0 ? $" {XMakeAttributes.sdk}=\"{importSdk}\"" : String.Empty;
+                        string importTag =
+                            $"  <Import Project=\"{importProject}\"{sdk}{condition}>";
+
+                        if (((XmlElement)child).GetAttribute(XMakeAttributes.@implicit).Length > 0)
+                        {
+                            importTag =
+                                $"  Import of \"{importProject}\" from Sdk \"{importSdk}\" was implied by the {XMakeElements.project} element's {XMakeAttributes.sdk} attribute.";
```
jeffkl: 👍

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -619,7 +633,11 @@ private void AddInitialChild(ProjectElement child)
             AddToXml(child);
 
             _count++;
-            MarkDirty("Add child element named '{0}'", child.ElementName);
+
+            if (HasXmlRepresentation(child))
```
dsplaisted: Wouldn't you also want to suppress marking the tree as dirty if the child is implicit in `InsertBeforeChild` and `InsertAfterChild`?

src/XMakeBuildEngine/Construction/ProjectImportElement.cs
```diff
@@ -44,26 +44,36 @@ public string Project
...
-            get { return XmlElement.GetAttributeLocation(XMakeAttributes.project); }
+            get { return XmlElement.HasAttribute(XMakeAttributes.@implicit); }
         }
 
+
+        /// <summary>
+        /// Location of the project attribute
+        /// </summary>
+        /// <remarks>
+        /// For an implicit import, the location points to the Sdk attribute on the Project element.
```
dsplaisted: This line of the comment doesn't seem to match the behavior in the code.

src/XMakeBuildEngine/Construction/ProjectImportElement.cs
```diff
@@ -44,26 +44,36 @@ public string Project
...
         /// <summary>
-        /// Location of the project attribute
+        /// Gets the Implicit state of the element: true if the element was not in the read XML.
         /// </summary>
-        public ElementLocation ProjectLocation
+        // TODO: *should* this be public? if it's not, you can't determine if an import is implicit from the public OM.
+        public bool Implicit
         {
-            get { return XmlElement.GetAttributeLocation(XMakeAttributes.project); }
+            get { return XmlElement.HasAttribute(XMakeAttributes.@implicit); }
```
dsplaisted: Could we represent whether an import is implicit or not with a field in this class instead of a separate attribute?  What happens if someone actually uses the attribute?

```xml
<Import Project="foo.targets" _Implicit="Foo" />
```

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1714,8 +1714,23 @@ public void Save(Encoding saveEncoding)
                 {
                     using (ProjectWriter projectWriter = new ProjectWriter(_projectFileLocation.File, saveEncoding))
                     {
-                        projectWriter.Initialize(XmlDocument);
-                        XmlDocument.Save(projectWriter);
+                        var xmlWithNoImplicits = (XmlDocument)XmlDocument.Clone();
```
dsplaisted: At first glance the variable name led me to believe that it wouldn't have any implicit imports when it was cloned, not that you were going to remove them afterwards.  It might be possible to make this clearer with a better name.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -1081,7 +1081,6 @@ private void PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
                 }
 
                 ProjectImportElement import = element as ProjectImportElement;
-
```
dsplaisted: No-op whitespace change here.

src/XMakeBuildEngine/Evaluation/Preprocessor.cs
```diff
@@ -166,13 +168,16 @@ private void CloneChildrenResolvingImports(XmlNode source, XmlNode destination)
...
                     string importProject = ((XmlElement)child).GetAttribute(XMakeAttributes.project);
+                    string importSdk = ((XmlElement)child).GetAttribute(XMakeAttributes.sdk);
 
                     IList<ProjectRootElement> resolvedList;
                     if (!_importTable.TryGetValue(child as XmlElement, out resolvedList))
                     {
                         // Import didn't resolve to anything; just display as a comment and move on
-                        string closedImportTag = "<Import Project=\"" + importProject + "\"" + ((importCondition.Length > 0) ? " Condition=\"" + importCondition + "\"" : String.Empty) + " />";
+                        string closedImportTag =
+                            $"<Import Project=\"{importProject}\"{condition} />";
```
dsplaisted: Should this also cover the case where there is an `Sdk` attribute on the import?

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -186,6 +187,62 @@ private void ParseProjectElement(XmlElementWithLocation element)
...
+
+                foreach (var sdk in sdks)
+                {
+                    var slashIndex = sdk.LastIndexOf("/", StringComparison.InvariantCulture);
```
dsplaisted: Would Ordinal string comparison be preferred to InvariantCulture here?

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -186,6 +187,62 @@ private void ParseProjectElement(XmlElementWithLocation element)
...
+                            "InvalidSdkFormatTooManySlashes");
+                    }
+
+                    // TODO: paths should just be Sdk.props/targets; Sdk-aware imports should do the rest of the path.
+                    var initialImportPath = Path.Combine(Environment.GetEnvironmentVariable("MSBuildSDKsPath"),
+                        sdkName, "Sdk", "Sdk.props");
+                    var finalImportPath = Path.Combine(Environment.GetEnvironmentVariable("MSBuildSDKsPath"),
+                        sdkName, "Sdk", "Sdk.targets");
+
+                    if (File.Exists(initialImportPath))
```
dsplaisted: I think we need to generate an error if an Sdk is specified that is not found.  However, that should be an error at evaluation, not parse time.  To get that behavior for now, I think we should always add both implicit import elements whether the files exist or not, which will mean that all of our Sdks will need to have both `Sdk.props` and `Sdk.targets`.

When we add support for an `Sdk` attribute on an `Import` and implement the Project-level `Sdk` attribute in terms of Sdk imports, we may be able to change how this works.

src/XMakeBuildEngine/Microsoft.Build.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dsplaisted: No-op changes (EOL probably) here and at end of file.

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ConstructionEditing_Tests.cs
```diff
@@ -3199,6 +3199,62 @@ private static string ComposeExpectedProjectString(string expectedItem)
...
+        public void AddProperty_WithSdk_KeepsSdkAndImplicitImports()
+        {
+            var testSdkRoot = Path.Combine(Path.GetTempPath(), "MSBuildUnitTest");
+            var testSdkDirectory = Path.Combine(testSdkRoot, "MSBuildUnitTestSdk");
+
+            try
+            {
+                Directory.CreateDirectory(testSdkDirectory);
+
+                string sdkPropsPath = Path.Combine(testSdkDirectory, "Sdk.props");
```
dsplaisted: I think this needs to be `Path.Combine(testSdkDirectory, "Sdk", "Sdk.props")` to match the path where the parser is going to look for the Sdk imports.

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectImportElement_Tests.cs
```diff
@@ -45,53 +45,56 @@ public void ReadNone()
         public void ReadInvalidMissingProject()
         {
             Assert.Throws<InvalidProjectFileException>(() =>
-            {
-                string content = @"
+                {
```
dsplaisted: Lots of whitespace changes in this file.

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectImportElement_Tests.cs
```diff
@@ -287,5 +291,47 @@ public void ImportWithRelativePath()
...
+            {
+                Directory.CreateDirectory(testSdkDirectory);
+
+                string sdkPropsPath = Path.Combine(testSdkDirectory, "Sdk.props");
+                string sdkTargetsPath = Path.Combine(testSdkDirectory, "Sdk.targets");
+
+                File.WriteAllText(sdkPropsPath, string.Empty);
+                File.WriteAllText(sdkTargetsPath, string.Empty);
+
+                using (new Helpers.TemporaryEnvironment("MSBUILDMAGICIMPORTDIRECTORY", testSdkRoot))
```
dsplaisted: Change to `MSBuildSDKsPath` environment variable.

src/Shared/UnitTests/ObjectModelHelpers.cs
rainersigwald: I envisioned a cascading set of usings but yeah, that's more elegant. Mind if we wait until after RC2?

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
rainersigwald: It wasn't enough in the original PR because I only put `Implicit` on `ProjectImportElement`. But you're right: if I just extend it to all `ProjectElement`s this becomes much cleaner.

src/Shared/UnitTests/ObjectModelHelpers.cs
jeffkl: I'll leave it up to you

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
rainersigwald: I can't think of a reason for this to fail either, but it seemed worth checking and throwing. Would you rather I just let a NRE happen later?

src/XMakeBuildEngine/Construction/ProjectImportElement.cs
rainersigwald: It throws an error: https://github.com/Microsoft/msbuild/pull/1403/files#diff-a9d9d41d4ebe72ee0285850966f63d3fR576

Test TK.

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
jeffkl: I did some searching and couldn't find an instance where it would return null unless the document couldn't be parsed.  I think its safe to remove the check personally.

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -186,6 +187,62 @@ private void ParseProjectElement(XmlElementWithLocation element)
...
+                {
+                    var slashIndex = sdk.LastIndexOf("/", StringComparison.InvariantCulture);
+                    string sdkName = slashIndex > 0 ? sdk.Substring(0, slashIndex) : sdk;
+
+                    // TODO: do something other than just ignore the version
+
+                    if (sdkName.Contains("/"))
+                    {
+                        ProjectErrorUtilities.ThrowInvalidProject(element.GetAttributeLocation(XMakeAttributes.sdk),
+                            "InvalidSdkFormatTooManySlashes");
```
rainersigwald: Self, actually add this string to the resx.

src/XMakeBuildEngine/Construction/ProjectImportElement.cs
rainersigwald: Good catch--remnant of a previous attempt.

rainersigwald:
Consensus has emerged that the default value of the SDK location should be `{MSBuild directory}\Sdks`. See https://github.com/dotnet/sdk/pull/444 for example.

I need to:
- [ ] Create an internal variable for that
- [ ] Make a property function to expose it
- [ ] Make a property with the same name in toolsets
  - [ ] Full Framework
  - [ ] .NET Core

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1714,8 +1714,18 @@ public void Save(Encoding saveEncoding)
...
```
rainersigwald: ah, I apparently hadn't committed and pushed that fix when I made it. You're right--the replacement is `CloneNode(true)`.

rainersigwald:

rainersigwald:
I'm going to `rebase -i` to a semi-coherent commit structure to resolve the merge to rc2

src/XMakeBuildEngine/Construction/ProjectImportElement.cs
rainersigwald: Also, I tried using a field in this class, but it caused problems with the preprocessor, since it walks only the XML tree.

src/Shared/XMakeAttributes.cs
```diff
@@ -40,9 +40,11 @@ internal static class XMakeAttributes
         internal const string taskName = "TaskName";
         internal const string continueOnError = "ContinueOnError";
         internal const string project = "Project";
+        internal const string @implicit = "_Implicit";
```
jeffkl: Can this be isImplicit so you can get rid of the `@`?  Should the attribute name just be `IsImplicit`?

src/XMakeBuildEngine/Definition/ToolsetLocalReader.cs
```diff
@@ -57,6 +57,7 @@ protected override IEnumerable<ToolsetPropertyDefinition> ToolsVersions
         protected override IEnumerable<ToolsetPropertyDefinition> GetPropertyDefinitions(string toolsVersion)
         {
             yield return new ToolsetPropertyDefinition(MSBuildConstants.ToolsPath, BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory, _sourceLocation);
+            yield return new ToolsetPropertyDefinition(MSBuildConstants.SdksPath, BuildEnvironmentHelper.Instance.MSBuildSDKsPath, _sourceLocation);
```
rainersigwald: Not by MSBuild itself, but the first iteration of some of the magic SDKs plan to do SDK-to-SDK references using the property. It's set here for .NET Core and in app.config for full.

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -186,6 +187,60 @@ private void ParseProjectElement(XmlElementWithLocation element)
             // so we have to set it now
             _project.SetProjectRootElementFromParser(element, _project);
 
+            if (element.HasAttribute(XMakeAttributes.sdk))
+            {
+
+                var sdksString = element.GetAttribute(XMakeAttributes.sdk);
+
+                var sdks = sdksString.Contains(";") ? sdksString.Split(';') : new[] {sdksString};
```
cdmihai: >sdksString.Split(';') [](start = 54, length = 21)

You could also look into using ExpressionShredder.SplitSemiColonSeparatedList. Though it might be overkill for this situation.

rainersigwald: worth thinking about for vnext

src/XMakeBuildEngine/Definition/ToolsetLocalReader.cs
cdmihai: I think you may also have to change `ToolsetConfigurationReader` to make the property flow from the app config too.

PR #1404, opened 2016-11-29
Preserve Culture metadata on satellite assemblies

AArnott:
Satellite assemblies are produced only with TargetPath metadata, which makes rediscovering their culture later in the build very difficult. Simply preserving it as a distinct metadatum makes scenarios such as NuGet packaging of Roslyn analyzers that are localized much easier.

BTW, I'm not sure about the target branch for this. I think it should ship xplat as well as in Dev15 RC.3. Is this the right branch for that?

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -5323,6 +5323,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <SatelliteDllsProjectOutputGroupOutputIntermediate Include="$(IntermediateOutputPath)%(EmbeddedResource.Culture)\$(TargetName).resources.dll"
                                                          Condition="'%(WithCulture)' == 'true'">
         <TargetPath>%(EmbeddedResource.Culture)\$(TargetName).resources.dll</TargetPath>
+        <Culture>%(EmbeddedResource.Culture)</Culture>
       </SatelliteDllsProjectOutputGroupOutputIntermediate>
     </ItemGroup>
 


cdmihai:
This makes it so that only the users of the `SatelliteDllsProjectOutputGroupOutputIntermediate ` item benefit from the culture.

I wonder if it's worth doing this at the more general level, so all users of the satellite assemblies get the culture.

This is the target that actually creates the satellites: https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L3376-L3421.
It unfortunately duplicates the satellite path twice (once in `Outputs`, once in the task call), instead of emitting a new item to capture that computation. The target you modified duplicates this logic again.

Can you please look into modifying the `GenerateSatelliteAssemblies` target so that it emits a new item representing the satellite assemblies? Then the target you modified can just reuse this item.

AArnott:
@cdmihai said:

> This makes it so that only the users of the SatelliteDllsProjectOutputGroupOutputIntermediate item benefit from the culture.

Not true, actually. The `SatelliteDllsProjectOutputGroupOutputIntermediate` item is immediately transformed to `SatelliteDllsProjectOutputGroupOutput` items, which is then returned as the result of the `SatelliteDllsProjectOutputGroup` output group target. So in fact where I've put this metadata is the _earliest_ upstream place I could put it to maximize benefit to all consumers. The output group is the only way to get this item list the way the .targets are written, and now every copy of that list has the metadata where it didn't before. :)

The `GenerateSatelliteAssemblies` target that you reference doesn't even produce items. It _might_ have target outputs, although I'm not too familiar with using the `%()` syntax in the outputs attribute as it has here. But that's not something anyone else in MSBuild itself can consume. So short of adding a new feature to that target to start producing items it never had before, I don't think there's anything to enhance about that target. And since enhancing the target I do in this PR unblocks my scenario, and I don't know of any other scenario that would require updating the target you suggested, I think this PR is probably the low-risk, simple PR that addresses the problem without being overly expensive or risky.

cdmihai:
OK, I was thinking about moving the creation of that item in `GenerateSatelliteAssemblies` so the path construction is not duplicated three times, but I see how that would be riskier than this approach. LGTM :)

PR #1405, opened 2016-11-29
Add NormalizePath() to IntrinsicFunctions

### LINKED ISSUE #55: File paths need better support on Unix
Because most project files assume that the directory path delimiter is a `\`, MSBuild contains a `FileUtilities.FixFilePath` method that is used by many tasks to ensure that paths are properly constructed when running on Unix systems.

However, some projects contain references to custom tasks that deal with file paths but do not have access to this method. (see, for example, the [DownloadFile](https://github.com/dotnet/coreclr/blob/82440897d45e16d706e80dc96918d362339d0d95/dir.targets#L4-22) task in coreclr)

After [some discussion](https://gitter.im/dotnet/coreclr?at=5517850d2278254a76090d89) in the coreclr chat room, it was proposed that this functionality might be best exposed as a [Property Function](https://msdn.microsoft.com/en-us/library/dd633440.aspx).

Such a property function would most likely need to be a member of the `MSBuild` set of property functions, which are defined in [`IntrinsicFunctions`](https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs). It would also need a name--`FixFilePath` might be acceptable, but I suspect that a more descriptive name would be desired.

I would be happy to submit a Pull Request if a member of the team signs off on this issue (and we agree on a name for the function)

jeffkl:
Rather than just "fix" the file path, this normalizes it and fixes the directory separators.  I did not make a unit test because `FileUtilities.NormalizePath()` is well tested.

Closes #55

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -323,7 +323,7 @@ namespace Microsoft.Build.Shared
                         // we reluctantly have to restrict things here.
                         if (length >= MaxPath)
                         {
-                            throw new PathTooLongException(path);
+                            throw new PathTooLongException();
                         }
 
                         // Avoid creating new strings unnecessarily
--- a/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
+++ b/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
@@ -381,14 +381,9 @@ namespace Microsoft.Build.Evaluation
         /// while ensuring it has a trailing slash.
         /// </summary>
         /// <param name="path">One or more directory paths to combine and normalize.</param>
-        /// <returns>A canonicalized full directory path with the correct directory separators and a trailing slash.  If no path is specified, the return value is <see cref="String.Empty"/>.</returns>
+        /// <returns>A canonicalized full directory path with the correct directory separators and a trailing slash.</returns>
         internal static string NormalizeDirectory(params string[] path)
         {
-            if (path.Length == 0 || path.All(String.IsNullOrWhiteSpace))
-            {
-                return String.Empty;
-            }
-
             return EnsureTrailingSlash(NormalizePath(path));
         }
 
--- a/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
+++ b/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
@@ -378,11 +378,11 @@ namespace Microsoft.Build.Evaluation
         /// <summary>
         /// Gets the canonicalized full path of the provided path and ensures it contains the correct directory separator characters for the current operating system.
         /// </summary>
-        /// <param name="path">The path to normalize.</param>
+        /// <param name="path">One or more paths to combine and normalize.</param>
         /// <returns>A canonicalized full path with the correct directory separators.</returns>
-        internal static string NormalizePath(string path)
+        internal static string NormalizePath(params string[] path)
         {
-            return FileUtilities.NormalizePath(path);
+            return FileUtilities.NormalizePath(Path.Combine(path));
         }
 
         public static string GetCurrentToolsDirectory()
--- a/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
+++ b/src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 using Microsoft.Build.Internal;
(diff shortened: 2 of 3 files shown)


danmosemsft:
`FileUtilities.NormalizePath()` can throw, what happens when use in a project file causes an exception?

jeffkl:
Excellent question.  If you pass it an empty string, you get an error during your build:
```
Microsoft (R) Build Engine version 15.1.418.63065

Build started 11/30/2016 8:10:23 AM.
Project "C:\Users\jeffkl\Desktop\55.proj" on node 1 (default targets).
C:\Users\jeffkl\Desktop\55.proj(11,14): error MSB4184: The expression "[MSBuild]::NormalizePath(*<>)" cannot be evaluated. Illegal characters in path.
Done Building Project "C:\Users\jeffkl\Desktop\55.proj" (default targets) -- FAILED.


Build FAILED.

"C:\Users\jeffkl\Desktop\55.proj" (default target) (1) ->
(Build target) ->
  C:\Users\jeffkl\Desktop\55.proj(11,14): error MSB4184: The expression "[MSBuild]::NormalizePath('')" cannot be evaluated. Parameter "path" cannot have zero length.

    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.26
```

If you have invalid characters you get this:
```
error MSB4184: The expression "[MSBuild]::NormalizePath(*<>)" cannot be evaluated. Illegal characters in path.
```

Invalid UNC path:
```
error MSB4184: The expression "[MSBuild]::NormalizePath(\\??\\)" cannot be evaluated. The UNC path should be of the form \\server\share.
```

_But_ if you specify a path that's too long, you get a very strange error:
```
C:\Users\jeffkl\Desktop\55.proj(12,14): error MSB4184: The expression "[MSBuild]::NormalizePath(12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890)" cannot be evaluated. 1234567890123456789012345678901234567890123456789012
34567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
901234567890
```

I'll figure out what that is all about.  Do the other errors seem like they are a good user experience?

rainersigwald:
@jeffkl put a non-digit character in there--we may be trying to coerce it to an int or something?

jeffkl:
Okay I fixed the PathTooLongException message so now it says: 

```
error MSB4184: The expression "[MSBuild]::NormalizePath(12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890)" cannot be evaluated. The specified path, file name, or both are too long.
 The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
```

jeffkl:
@rainersigwald 

> put a non-digit character in there--we may be trying to coerce it to an int or something?

A number doesn't appear to matter:

```xml
<Message Text="Result: '$([MSBuild]::NormalizePath(123))'" />
```
```
  Result: 'C:\Users\jeffkl\Desktop\123'
```

Anything else I should try?

jeffkl:
@AArnott hate to bother you, do you know why GitVersioning would crash like this?
https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/_Windows_NT_Desktop_prtest/95/consoleFull#153167369479494335-f7bd-47d0-8771-8661e00c2db2

It only happens in our CI environment and I can't get it to repro locally.  Before I spent any time looking into it I figured I'd ping you just in case you knew what the problem is or where to start.

rainersigwald:
@jeffkl I filed #1409 since in my feature branch I just disabled it. Would really like to get to the bottom of it, though.

AArnott:
@jeffkl: my best guess, since the failure is deep within libgit2sharp in reading a tree object, is that the git commit is corrupted/invalid, or that the server didn't initialize the .git directory correctly.
My first suggestion is to try to build again on a clean agent (that has no prior git clone).
My second suggestion is run `git fsck` on a clone with that commit in it to see if it's invalid somehow.

jeffkl:
@AArnott thanks a lot, I'll see what I can figure out.

jeffkl:
@AArnott I was able to figure out that our CI build definition changed recently to do a shallow clone and so GitVersioning obviously wouldn't be able to calculate commit height.  I changed the setting back in the build definition and everything is working again.  Just a heads up in case anyone else hits the issue.  Thanks again for pointing me in a direction.

jeffkl:

jeffkl:
I decided to add a NormalizeDirectory() as well which can be used to combine, normalize, and ensure a trailing slash.  Passing in an empty directory will return String.Empty rather than throwing an error.

AArnott:
@jeffkl Thanks for letting me know what it was. NB.GV does document that shallow cloning isn't supported, but it would be interesting for me to test this further so that instead of a strange failure like this it could give a more helpful message to reduce time like yours spent investigating. I've filed https://github.com/AArnott/Nerdbank.GitVersioning/issues/92 to track.

PR #1416, opened 2016-11-30
Do not reveal implicit elements through RawXml

rainersigwald:
The initial implementation caused errors when editing a project in Visual
Studio. The Roslyn Project System uses Project.Xml.RawXml to decide what
to show to the user, and handles edits by saving that file. When RawXml
had _Implicit elements, that made the first save create an invalid project
with a literal _Implicit attribute in its imports (as well as imports from
both the Sdk attribute and the now-present Import elements).

To fix this, do the same transformation for RawXml that was being done for
Save.

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -705,8 +705,10 @@ namespace Microsoft.Build.Construction
                 {
                     using (ProjectWriter projectWriter = new ProjectWriter(stringWriter))
                     {
-                        projectWriter.Initialize(XmlDocument);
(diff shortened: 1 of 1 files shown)


src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1785,6 +1779,20 @@ public void Save(Encoding saveEncoding)
 #endif
         }
 
+        private XmlDocument RemoveImplicits()
```
cdmihai: If this turns out to be a hot path, then the ProjectRootElement could be made aware if it has implicit elements and skip this call when it doesn't.

PR #1421, opened 2016-12-01
Update casing for Sdks

dsplaisted:
This change makes the casing of `Sdk` more consistent.

More importantly, it uses the lowercase version of the Sdk name when constructing the Sdk path.  This will enable Sdk names to be case-insensitive even when the file system is case sensitive.  For this to work we'll also need to make sure that we use lowercase versions of the Sdk names when we are deploying them to the Sdk folder.  This is the same way NuGet handles package IDs today.

@rainersigwald @AndyGerlicher

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -446,7 +446,7 @@ namespace Microsoft.Build.Shared
         /// <summary>
         /// Path to the Sdks folder for this MSBuild instance.
         /// </summary>
-        internal string MSBuildSDKsPath
+        internal string MSBuildSdksPath
         {
             get
             {
@@ -463,7 +463,7 @@ namespace Microsoft.Build.Shared
                 }
 
                 // Allow an environment-variable override of the default SDK location
-                return Environment.GetEnvironmentVariable("MSBuildSDKsPath") ?? defaultSdkPath;
+                return Environment.GetEnvironmentVariable(MSBuildConstants.SdksPath) ?? defaultSdkPath;
             }
         }
 
--- a/src/Shared/Constants.cs
+++ b/src/Shared/Constants.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Build.Shared
         /// <summary>
         /// Name of the property that indicates the root of the SDKs folder
         /// </summary>
-        internal const string SdksPath = "MSBuildSDKsPath";
+        internal const string SdksPath = "MSBuildSdksPath";
 
         /// <summary>
         /// The most current Visual Studio Version known to this version of MSBuild. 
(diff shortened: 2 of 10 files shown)


dsplaisted:
What do you mean by "corefx did not move to lowercasing"?

NuGet's lowercasing is mostly an implementation detail.  Package Ids are still capitalized correctly when they are referred to in `PackageReference`, `project.json`, or `packages.config`.  So I don't think it really conflicts with our standard conventions.

Another reason I think it makes sense to follow NuGet's behavior here is that the eventual acquisition story for MSBuild SDKs will probably involve publishing them as NuGet packages and using NuGet to download them to the Sdks folder.

rainersigwald:
> What do you mean by "corefx did not move to lowercasing"?

I mean that DLLs shipped from corefx have capital letters on disk, and that if you change the case on disk you'll get an assembly load error, as I expect on a case-sensitive file system.

> NuGet's lowercasing is mostly an implementation detail.

An implementation detail that [requires an escape hatch](https://github.com/NuGet/NuGet.Client/pull/814) and breaks the conceptual model "these paths are constructed from package IDs". I'm trying to avoid that fate here.

> the eventual acquisition story for MSBuild SDKs will probably involve publishing them as NuGet packages and using NuGet to download them to the Sdks folder.

I would expect this to be an explicit "deploy from {downloaded and unzipped package} to {destination location}", not "unzip from web into this folder".

jeffkl:
I agree with @rainersigwald, changing the case of what came with the template or the user typed in could spell trouble down the line.  At least if the user got the wrong case, they could fix it themselves.  If MSBuild is changing the case of things, they wouldn't be able to get around it...

dsplaisted:
I get that there's no way to be entirely consistent.  But are you folks saying that we shouldn't match NuGet's behavior here because NuGet's behavior was the wrong thing to do, or because there's a relevant difference between specifying an Sdk in the `Sdk` attribute and referencing a NuGet package?

rainersigwald:
> NuGet's behavior was the wrong thing to do

Yes. And it's not prevalent enough to serve as precedent for "broken but common so we'll go with it".

dsplaisted:
What about the `sdk` portion of the path in the Sdk folder?  I changed it to lowercase to match the casing of the folders inside NuGet packages (`lib`, `build`, etc).  In this case NuGet isn't converting something to lowercase (AFAIK), it's just specifying that the folder name should be lowercase.

jeffkl:
> What about the `sdk` portion of the path in the Sdk folder?

Good question, you also left `Sdk.props` proper cased, do we know if there's any enforcement in NuGet packages?  I could conceivably pack stuff into `NeT46` directories but I think it still works because things read the lock file.

That might be the only good long term solution is to read the assets file so we know where NuGet put the assets...

rainersigwald:
> What about the `sdk` portion of the path in the Sdk folder?

I'm fine with that for consistency. But for the timing of RC2 I'd rather not have to take the change.

AndyGerlicher:
Shiproom is in 2.5 hours so I would like to postpone these until post RC2. We can discuss at shiproom if you feel strongly against this.

Past that I do think commit 1 and 3 make a lot of sense. I lean towards what @rainersigwald said on the sdk name though.

AndyGerlicher:
Implementation on this has changed, closing. There's still an open question about should the package name be lower cased, but that detail should be in the resolver not in MSBuild itself (still to be developed).

PR #1441, opened 2016-12-05
Include .config files in AllowedReferenceRelatedFileExtensions

### LINKED ISSUE #1307: Include .config file when copying output files of a dependent project
Consider scenario where project A references project B and project B has an App.config file. The build copies .dll/.exe and .pdb files of B to the output directory of A, but doesn't currently copy .dll.config/.exe.config files that may contain binding redirects and other configuration.

I propose the build copies the configuration file as well, so that binding redirects are not lost. Since the latest CoreFX binaries heavily rely on presence of binding redirects having binding redirects flow seamlessly thru build system is becoming more and more important.

jeffkl:
Closes #1307

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -577,7 +577,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       .pdb;
       .xml;
       .pri;
-      .config
+      .dll.config;
+      .exe.config
     </AllowedReferenceRelatedFileExtensions>
 
     <!--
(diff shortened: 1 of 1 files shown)


jeffkl:

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -576,7 +576,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <AllowedReferenceRelatedFileExtensions Condition=" '$(AllowedReferenceRelatedFileExtensions)' == '' ">
       .pdb;
       .xml;
-      .pri
+      .pri;
+      .config
```
rainersigwald: I think it'd be better to explicitly have `.dll.config` and `.exe.config`.

AndyGerlicher: Change to `.exe.config` and `.dll.config`

jeffkl: Should I just do `.dll.config`?  Isn't `exe.config` already copied?

jeffkl:
Nevermind, after testing this it only worked when I used `.dll.config` and `.exe.config`.

PR #1448, opened 2016-12-06
Implementation of GetPathOfFileAbove

### LINKED ISSUE #1277: Add GetFileAbove intrinsic function
The CoreFX repo contains [over 1,000 instances](https://github.com/dotnet/corefx/search?utf8=%E2%9C%93&q=getdirectorynameoffileabove) of this pattern:
`<Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />`

it seems to me it ought to be possible to write this more compactly like

`<Import Project="$([MSBuild]::GetFileAbove(dir.props)" />`

This assumes that an intrinsic function can get passed the current file's location as an implicit parameter. I haven't looked at the code but I assume it can.

Once this is added, it should be fairly easy to search and replace to the new syntax.

jeffkl:
Only accepts a file name
Expander magically passes the directory containing IElementLocation.File before the method is called so users don't have to specify both arguments.

Closes #1277

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Expander.cs
+++ b/src/XMakeBuildEngine/Evaluation/Expander.cs
@@ -3085,14 +3085,18 @@ namespace Microsoft.Build.Evaluation
                         // Special case a few methods that take extra parameters that can't be passed in by the user
                         //
 
-                        if (_methodMethodName.Equals("GetPathOfFileAbove"))
+                        if (_methodMethodName.Equals("GetPathOfFileAbove") && args.Length == 1)
                         {
-                            // Append the IElementLocation as a parameter to GetPathOfFileAbove
+                            // Append the IElementLocation as a parameter to GetPathOfFileAbove if the user only
+                            // specified the file name.  This is syntactic sugar so they don't have to always
+                            // include $(MSBuildThisFileDirectory) as a parameter.
                             //
+                            string startingDirectory = String.IsNullOrWhiteSpace(elementLocation.File) ? String.Empty : Path.GetDirectoryName(elementLocation.File);
+
                             args = new []
                             {
                                 args[0],
-                                elementLocation,
+                                startingDirectory,
                             };
                         }
                     }
--- a/src/XMakeBuildEngine/Evaluation/Expander.cs
+++ b/src/XMakeBuildEngine/Evaluation/Expander.cs
@@ -3080,6 +3080,23 @@ namespace Microsoft.Build.Evaluation
                         args[0] = Convert.ChangeType(args[0], objectInstance.GetType(), CultureInfo.InvariantCulture);
                     }
 
+                    if (_receiverType == typeof(IntrinsicFunctions))
+                    {
+                        // Special case a few methods that take extra parameters that can't be passed in by the user
+                        //
+
+                        if (_methodMethodName.Equals("GetPathOfFileAbove"))
+                        {
+                            // Append the IElementLocation as a parameter to GetPathOfFileAbove
+                            //
+                            args = new []
+                            {
+                                args[0],
+                                elementLocation,
+                            };
+                        }
+                    }
+
                     // If we've been asked to construct an instance, then we
                     // need to locate an appropriate constructor and invoke it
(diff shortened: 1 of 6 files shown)


src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
```diff
@@ -306,6 +306,35 @@ internal static string GetDirectoryNameOfFileAbove(string startingDirectory, str
...
+        /// </summary>
+        /// <param name="file">The file to search for.</param>
+        /// <param name="elementLocation">An <see cref="IElementLocation"/>.  The directory of the <see cref="IElementLocation.File"/> is used as the starting
+        /// location of the file search.</param>
+        /// <returns>The full path of the file if it is found in a path above the <see cref="IElementLocation.File"/>.</returns>
+        internal static string GetPathOfFileAbove(string file, IElementLocation elementLocation)
+        {
+            if (String.IsNullOrWhiteSpace(elementLocation?.File))
+            {
+                ErrorUtilities.ThrowInternalError($"{nameof(elementLocation)} should not be null and it should have a file path associated with it.");
```
rainersigwald: Is this really an internal error? Wouldn't you hit this if you had an in-memory representation of a project that used this? That seems user-accessible (albeit still an error).

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -3080,6 +3080,23 @@ internal object Execute(object objectInstance, IPropertyProvider<T> properties,
...
+                        //
+
+                        if (_methodMethodName.Equals("GetPathOfFileAbove"))
+                        {
+                            // Append the IElementLocation as a parameter to GetPathOfFileAbove
+                            //
+                            args = new []
+                            {
+                                args[0],
+                                elementLocation,
```
rainersigwald: Definitely don't love this but I also can't think of a better way at the moment.

src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
cdmihai: What if the fallback location is the entry project's location? As far as I remember, an in-memory project uses the current directory by default.

src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
```diff
@@ -306,6 +306,35 @@ internal static string GetDirectoryNameOfFileAbove(string startingDirectory, str
...
+            }
+
+            // This method does not accept a path, only a file name
+            if(file.Any(i => i.Equals(Path.DirectorySeparatorChar) || i.Equals(Path.AltDirectorySeparatorChar)))
+            {
+                ErrorUtilities.ThrowArgument("InvalidGetPathOfFileAboveParameter", file);
```
cdmihai: Isn't there an error overload that accepts the location as well, so users would get the location of where this happens?

src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
jeffkl: I'd hate to see it import different files under different circumstances.  We should discuss this today.

src/XMakeBuildEngine/Evaluation/IntrinsicFunctions.cs
jeffkl: The error is thrown and caught by the expander.  It is then wrapped as an `InvalidProjectFileException` with the location of the property function.
``` C#
catch (TargetInvocationException ex)
{
    // We ended up with something other than a function expression
    string partiallyEvaluated = GenerateStringOfMethodExecuted(_expression, objectInstance, _methodMethodName, args);
    ProjectErrorUtilities.ThrowInvalidProject(elementLocation, "InvalidFunctionPropertyExpression", partiallyEvaluated, ex.InnerException.Message.Replace("\r\n", " "));
    return null;
}
```

jeffkl:
Okay this has been re-worked and is ready for approval...

PR #1459, opened 2016-12-08
Fix null-ref on project import fallback error message.

### LINKED ISSUE #1458: Project import fallback throws when property not defined
This issue comes from our Windows Watson failure telemetry and presents as an unhandled null-ref exception. The exception is generated from [this](https://github.com/Microsoft/msbuild/blob/ab6fed11b32ae1a89357b2c8cbaa9471876f77ff/src/XMakeBuildEngine/Evaluation/Evaluator.cs#L2655) method when trying to form an error message for an import not found. I'm not sure what the scenario really is here, but the issue occurs when there is fallback defined in the Toolset for a property that is not defined. In that case we crash when trying to form the error message if we couldn't find the import.

AndyGerlicher:
- Fix scenario where import fallback is specified in the toolset for a
   property that is not defined. This scenario works but will throw a
   null-ref exception trying to error if the imported project is not
   found.
 - Added checks for null and guards against empty strings being considered
   for fallback.
Fixes #1458

### DIFF
--- a/src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
+++ b/src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
@@ -735,6 +735,64 @@ namespace Microsoft.Build.UnitTests.Evaluation
             }
         }
 
+        [Fact]
+        public void FallbackImportWithFileNotFoundWhenPropertyNotDefined()
+        {
+            // Import something from $(UndefinedProperty)
+            string mainTargetsFileContent = @"
+               <Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003' >
+                   <Import Project='$(UndefinedProperty)\filenotfound.props' />
+                   <Target Name='Main' DependsOnTargets='FromExtn' />
+               </Project>";
+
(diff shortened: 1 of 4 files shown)


src/XMakeBuildEngine/UnitTests/Evaluation/ImportFromMSBuildExtensionsPath_Tests.cs
```diff
@@ -735,6 +735,64 @@ public void FallbackImportWithUndefinedProperty()
...
+                // The path to "extensions1" fallback should exist, but the file doens't need to
+                extnDir1 = GetNewExtensionsPathAndCreateFile("extensions1", Path.Combine("file.props"), string.Empty);
+
+                // Implement fallback for UndefinedProperty, but don't define the property.
+                var configFileContents = @"
+                <configuration>
+                  <configSections>
+                    <section name=""msbuildToolsets"" type=""Microsoft.Build.Evaluation.ToolsetConfigurationSection, Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"" />
+                  </configSections>
+                  <msbuildToolsets default=""14.1"">
```
cdmihai: Can that 14.1 version impact anything?

PR #1460, opened 2016-12-08
Enable Mono CI builds

radical:


netci.groovy
```diff
@@ -11,6 +11,8 @@ def project = GithubProject
 
         if (osName == 'Windows_NT') {
             runtimes.add('Desktop')
+        } else {
+            runtimes.add('Mono')
         }
 
         // TODO: Mono
```
rainersigwald: Delete stale comment 🎉

netci.groovy
```diff
@@ -55,15 +57,23 @@ def project = GithubProject
                 case 'OSX':
                     newJob.with{
                         steps{
-                            shell("./cibuild.sh --scope Test --target ${runtime}")
+                            buildCmd="./cibuild.sh --scope Test --target ${runtime}"
+                            if (runtime == "Mono") {
+                                buildCmd += " --host Mono"
```
rainersigwald: Isn't this implied by `--target`? I thought that was how we did it for `CoreCLR`.

radical: `--target` selects the configuration (like `Debug-MONO`) and `--host` selects coreclr/mono to use for running msbuild.

netci.groovy
```diff
@@ -11,6 +11,8 @@ def project = GithubProject
 
         if (osName == 'Windows_NT') {
             runtimes.add('Desktop')
+        } else {
+            runtimes.add('Mono')
```
radical: I was going for the quickest route for now, as I thought that I will have to figure out mono install location etc. But we don't need to do that, we can build the mono configuration using .net itself. so, `./cibuild.sh --scope Test --target Mono` should be fine. I will add that tomorrrow and update the PR.

radical:
Opening a new PR against master.

PR #1462, opened 2016-12-09
Fix mono build and tests - green build

radical:


src/XMakeTasks/UnitTests/Microsoft.Build.Tasks.UnitTests.csproj
```diff
@@ -185,11 +185,11 @@
     <Reference Include="System.Core" />
     <Reference Include="System.Xaml" />
     <Reference Include="System.Xml" />
-    <Reference Include="PresentationFramework" />
+    <Reference Include="PresentationFramework" Condition="'$(OS)' == 'Windows'" />
```
akoeplinger: I'd be surprised if the test depends on WPF even on Windows so this can probably just be removed.

src/XMakeBuildEngine/Evaluation/Expander.cs
```diff
@@ -3341,12 +3341,12 @@ private static Type GetTypeFromAssembly(string typeName, string candidateAssembl
 
                 // Try to load the assembly with the computed name
 #if FEATURE_GAC
-#pragma warning disable 618
+#pragma warning disable 618, 612
```
akoeplinger: FYI Marek recently fixed LoadWithPartialName's Obsolete attribute to have a message in mono master.

cibuild.sh
```diff
@@ -299,6 +299,12 @@ if [[ $BOOTSTRAP_ONLY = true ]]; then
     exit $?
 fi
 
+# Microsoft.Net.Compilers package is available now, so we can use the latest csc.exe
+if [ "$host" = "Mono" ]; then
+        CSC_EXE=`find $PACKAGES_DIR/microsoft.net.compilers -name csc.exe`
```
akoeplinger: what if there are multiple versions of the package in that dir?
maybe it's better to just hardcode the version here.

cibuild.sh
```diff
@@ -309,7 +309,7 @@ fi
 # warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.
 # warning MSB3026: Could not copy "XXX" to "XXX". Beginning retry 1 in 1000ms.
 # warning AL1053: The version '1.2.3.4-foo' specified for the 'product version' is not in the normal 'major.minor.build.revision' format
-_NOWARN="MSB3277;MSB3026;AL1053"
+_NOWARN="MSB3276;MSB3277;MSB3026;AL1053"
```
akoeplinger: why is it compiling against a 2.0 mscorlib?
if that's expected, you should also add a comment above why this warning was disabled

src/XMakeBuildEngine/Evaluation/Expander.cs
radical: Okay, but we'll need this here till we start always building against the the next mono version.

cibuild.sh
```diff
@@ -299,12 +299,18 @@ if [[ $BOOTSTRAP_ONLY = true ]]; then
...
+        CSC_EXE=`find $PACKAGES_DIR/microsoft.net.compilers -name csc.exe`
+        CSC_ARGS="/p:CscToolExe=csc.exe /p:CscToolPath=`dirname $CSC_EXE` /p:DebugType=portable"
+fi
+
 # The set of warnings to suppress for now
 # warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.
 # warning MSB3026: Could not copy "XXX" to "XXX". Beginning retry 1 in 1000ms.
 # warning AL1053: The version '1.2.3.4-foo' specified for the 'product version' is not in the normal 'major.minor.build.revision' format
-_NOWARN="MSB3277;MSB3026;AL1053"
+_NOWARN="MSB3276;MSB3277;MSB3026;AL1053"
```
rainersigwald: Please comment above, like the others (but better: add a justification :)).

cibuild.sh
radical: Hm good point. And once we have the mono CI enabled, breakages (referenced version change) should show up!

rainersigwald: I hate to agree but I think I do--when we change the compiler version there are likely to be two side by side here.

cibuild.sh
radical: References which depend on "mscorlib, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" [/Library/Frameworks/Mono.framework/Versions/4.8.0/lib/mono/2.0-api/mscorlib.dll]. (TaskId:188)

/Users/ankit/.nuget/packages/Microsoft.VisualStudio.Setup.Configuration.Interop/1.2.304-preview5/lib/net35/Microsoft.VisualStudio.Setup.Configuration.Interop.dll (TaskId:188)
            Project file item includes which caused reference "/Users/ankit/.nuget/packages/Microsoft.VisualStudio.Setup.Configuration.Interop/1.2.304-preview5/lib/net35/Microsoft.VisualStudio.Setup.Configuration.Interop.dll". (TaskId:188
)

/Users/ankit/.nuget/packages/Microsoft.VisualStudio.Setup.Configuration.Interop/1.2.304-preview5/lib/net35/Microsoft.VisualStudio.Setup.Configuration.Interop.dll (TaskId:188)

---

So, this is coming from `Microsoft.VisualStudio.Setup.Configuration.Interop.dll`. What would be a better fix:

1. ignore the warning
2. add remapping in app.config (?)
3. Set `$(AutoGenerateBindingRedirects)=true` in the csproj

radical: FYI, it is being referenced [here](https://github.com/Microsoft/msbuild/blob/xplat/src/Utilities/project.json#L5).

radical: Updated the PR to use (1). Though that doesn't sound like the best way :/

radical:
Disabled some tests for now, to get to zero tests failing on mono/osx, so that we can enable Mono on CI!

radical:
CI builds+tests are passing now!

cdmihai:
@Microsoft/msbuild-maintainers 
This needs another approval.

src/XMakeBuildEngine/UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -13,10 +13,10 @@ namespace Microsoft.Build.Engine.UnitTests
 {
     public class BuildEnvironmentHelper_Tests
     {
-#if FEATURE_RUN_EXE_IN_TESTS
-        private const string MSBuildExeName = "MSBuild.exe";
-#else
+#if USE_MSBUILD_DLL_EXTN
```
radical: No, it doesn't replace it.

PR #1466, opened 2016-12-09
Stop building legacy nuget packages

### LINKED ISSUE #1099: Stop building older NuGet packages
The new packages are being built out of `build\NuGetPackages` instead of `src\nuget`.  We need to remove all the logic to build the older style packages since the new ones are built as a separate step in the build.

cdmihai:
The new way to build the packages is via the script: `build\NuGetPackages\CreateNuGetPackages.proj`

Closes #1099


rainersigwald:
Can you link this to #1099?

cdmihai:
@jeffkl

PR #1471, opened 2016-12-13
Rename cross-targeting to multi-targeting

mhutch:
per https://github.com/dotnet/sdk/issues/486


src/XMakeCommandLine/MSBuildTaskHost/OutOfProcTaskHost.cs
```diff
@@ -20,7 +20,7 @@
 namespace Microsoft.Build.CommandLine
 {
     /// <summary>
-    /// This is the Out-Of-Proc Task Host for supporting Cross-Targeting tasks.
+    /// This is the Out-Of-Proc Task Host for supporting Multi-Targeting tasks.
```
rainersigwald: This one actually means cross-compiling, please revert.

src/XMakeTasks/Microsoft.CSharp.targets
```diff
@@ -17,12 +17,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
    <!-- 
-        We are doing a cross-targeting build if there is no list of target frameworks specified
+        We are doing a multi-targeting build if there is no list of target frameworks specified
         nor is there a current target framework being built individually. In that case, this import is
-        redirected to Microsoft.CSharp.CrossTargeting.targets.
+        redirected to Microsoft.CSharp.MultiTargeting.targets.
    -->
    <PropertyGroup Condition="'$(TargetFrameworks)' != '' and '$(TargetFramework)' == ''">
-      <IsCrossTargetingBuild>true</IsCrossTargetingBuild>
+      <IsMultiTargetingBuild>true</IsMultiTargetingBuild>
```
nguerrera: We may need to set both as compat shim as discussed in email,

src/XMakeTasks/Microsoft.CSharp.CrossTargeting.targets
```diff
@@ -10,15 +10,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 ***********************************************************************************************
 -->
 
+<!-- These targets are deprecated, forward to new targets -->
```
rainersigwald: Is this intended as a temporary measure so we can update the SDK gradually? Since we haven't shipped an official release with these targets I hope it'd be ok to just rename them atomically.

nguerrera:
@rainersigwald @dsplaisted I'm actually now wondering if we should rip the cross-targeting stuff out of msbuild. Now that the SDK has more control over imports than after common props & targets, it could handle all of it by itself without a built-in feature in msbuild.

nguerrera:
(The non x-targeting -> x-targeting P2P resolution would need to remain in msbuild, though).

rainersigwald:
@nguerrera I'm in favor.

src/XMakeTasks/Microsoft.CSharp.CrossTargeting.targets
mhutch: yes

mhutch:
I think I addressed all the feedback.

setup/files.swr
```diff
@@ -23,18 +23,21 @@ folder InstallDir:\MSBuild\15.0\Bin
...
   file source=$(X86BinPath)Microsoft.CSharp.CurrentVersion.targets
+  file source=$(X86BinPath)Microsoft.CSharp.MultiTargeting.targets
   file source=$(X86BinPath)Microsoft.CSharp.CrossTargeting.targets
   file source=$(X86BinPath)Microsoft.CSharp.targets
   file source=$(X86BinPath)Microsoft.NetFramework.CurrentVersion.props
   file source=$(X86BinPath)Microsoft.NetFramework.CurrentVersion.targets
   file source=$(X86BinPath)Microsoft.NetFramework.props
   file source=$(X86BinPath)Microsoft.NetFramework.targets
   file source=$(X86BinPath)Microsoft.VisualBasic.CurrentVersion.targets
+  file source=$(X86BinPath)Microsoft.VisualBasic.MultiTargeting.targets
```
AndyGerlicher: These changes needs to go in the amd64 folder as well (further down in the file). Use the same file source, just needs to be dropped to amd64.

mhutch:
OK, i think I addressed all remaining feedback.

AndyGerlicher:
Sorry to not be able to take this, at this point I think we have to close. This wouldn't meet the bar for this release, and I don't think it's worth changing at this point. If you think this is still worth discussing for Update 1 time-frame please re-open and we can discuss. Thanks!

mhutch:
@AndyGerlicher since this is still tracking for dotnet/sdk#486 and the NuGet side was merged already, i think it makes sense to reopen, however I don't have permissions to do so

rainersigwald:
We didn't get this past the bar for RTW and now it would be a breaking change from RTW. I think we're stuck with the "cross" name.

dsplaisted:
@rainersigwald if it's not a breaking change can we consider taking this?

mhutch:
Bump!

Multi-targeting has become the generally accepted term for this feature and it's odd that the targets use a different terminology in a fairly visible way.

kzu:
Is this in already? This renaming has already happened on NuGet as well as asp.net (see https://github.com/NuGet/Home/issues/4098)

AndyGerlicher:
We didn't have any plan for this. Last I heard this was discussed internally and decided against back when it was closed. I followed the issue you linked (and all the items that mention that), but I couldn't find anything that referred to something that was merged (as opposed to just closed). You're sure this already happened?

akoeplinger:
@AndyGerlicher https://github.com/NuGet/NuGet.Client/commit/efae789c75899ca0bdd063f34a4b73cd41510d56 mentions `Renamed buildCrossTargeting to buildMultiTargeting` in the commit message and does the renaming in sources.

PR #1486, opened 2016-12-15
Add AssemblyResolve handler for in-memory assemblies in CodeTaskFactory

### LINKED ISSUE #594: Path used for inline task reference is not honored
The path specified in a reference is respected during compilation, but not during execution. 

Repro Steps: Create an inline task, add a reference to a non-standard assembly and attempt to use it in a code fragment. 

This bug has also been reported on [StackOverflow](http://stackoverflow.com/questions/9455354/msbuild-inline-task-reference-non-standard-microsoft-assemblies).

Code that reproduces the error:

```
<Project ToolsVersion="14.0" DefaultTargets="TestBuild" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <UsingTask TaskName="Hello" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll" >
    <ParameterGroup />
    <Task>
      <Reference Include="$(MSBuildExtensionsPath)\Microsoft\VisualStudio\v14.0\Web\Microsoft.Web.XmlTransform.dll"/>
      <Code Type="Fragment" Language="cs">
        <![CDATA[
          var f = new Microsoft.Web.XmlTransform.XmlTransformationException("test");
          Log.LogMessage("test: " + f.Message);  
        ]]>
      </Code>
    </Task>
  </UsingTask>
  <Target Name="TestBuild">
    <Hello />
  </Target>
</Project>
```

### LINKED ISSUE #954: Willow MSBuild does not look up MSBuild extension path under VS folder when using MSBuildExtensionPath32
Running from a Willow command-prompt, with the following variations of imports:
``` XML
  <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
[...]
   <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets"" />
```

``` XML
<Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" />
```

and

``` XML
<Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\VSSDK\Microsoft.VsSDK.targets"/>
```

Under a Willow command-prompt, I would expect all of these to get found under C:\Program Files (x86)\Microsoft Visual Studio 15.0\MSBuild\Microsoft\VisualStudio\v15.0\VSSDK.

Instead I get:
``` 
D:\roslyn-project-system\build\Targets\VSL.Imports.targets(368,3): error MSB4019: The imported project "C:\Program File
s (x86)\MSBuild\Microsoft\VisualStudio\v15.0\VSSDK\Microsoft.VsSDK.targets" was not found. Confirm that the path in the
 <Import> declaration is correct, and that the file exists on disk. [D:\roslyn-project-system\src\ProjectSystemDogfoodS
etup\ProjectSystemDogfoodSetup.csproj] [D:\roslyn-project-system\build\build.proj]
```

jeffkl:
We could that since the assembly is loaded via `Assembly.Load(byte[])`, it is not allowed to use loaded dependencies because they came from a different load context.  This change keeps a static dictionary around with a mapping of assembly strong name and assembly.  If the inline task asks for the dependency at run time, the assembly resolve handler will return the loaded assembly if the strong name is the same.

Users will still experience runtime exceptions if they only list top-level dependencies in their `<Reference />` list.  To get around this, they'll need to list the full closure of dependencies that are not in the GAC or the MSBuild.exe directory.

Closes #594

### DIFF
--- a/src/XMakeTasks/CodeTaskFactory.cs
+++ b/src/XMakeTasks/CodeTaskFactory.cs
@@ -31,6 +31,37 @@ namespace Microsoft.Build.Tasks
     public class CodeTaskFactory : ITaskFactory
     {
         /// <summary>
+        /// This dictionary keeps track of custom references to compiled assemblies.  The in-memory assembly is loaded from a byte
+        /// stream and as such its dependencies cannot be found unless they are in the MSBuild.exe directory or the GAC.  They
+        /// cannot be found even if they are already loaded in the AppDomain.  This dictionary knows the FullName of the assembly
+        /// and a reference to the assembly itself.  In the <see cref="CurrentDomainOnAssemblyResolve"/> handler, the dictionary 
+        /// is used to return the loaded assemblies as a way to allow custom references that are not in the normal assembly Load
+        /// context.
+        /// </summary>
+        private static readonly IDictionary<string, Assembly> s_knownReferenceAssemblies = new ConcurrentDictionary<string, Assembly>(StringComparer.OrdinalIgnoreCase);
+
+        static CodeTaskFactory()
+        {
+            // The handler is not detached because it only returns assemblies for custom references that cannot be found in the normal Load context
+            AppDomain.CurrentDomain.AssemblyResolve += CurrentDomainOnAssemblyResolve;
+        }
+
+        /// <summary>
+        /// Handles the <see cref="AppDomain.AssemblyResolve"/> event to return assemblies loaded from custom references.
+        /// </summary>
+        /// <param name="sender"></param>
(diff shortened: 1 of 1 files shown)


cdmihai:

jeffkl:

PR #1487, opened 2016-12-15
Print process module path when waiting for debugger to attach

jeffkl:
Sometimes MSBuild is running under `MSBuild.exe` but sometimes it's running under `dotnet.exe`.  This change just prints out the path to the "main module" so its easier to determine what to attach the debugger to.


cdmihai:

PR #1492, opened 2016-12-16
Implementation of implicit imports during evaluation instead of parsing

### LINKED ISSUE #1447: Redesign implicit import implementation
As implemented (in the parser), implicit XML elements have caused a fair amount of surprise and bug tail (#1416, #1431, #1445) related to having differences between the ProjectElement tree exposed through the construction APIs and the Xml tree. We should redesign to avoid this.

I spoke to @jeffkl about this and he thought moving the transformation to be just before evaluation/preprocessing begins would be cleaner: Construction API consumers would see only the project file as it exists on disk, but the evaluator would get the complete tree as it expects. That sounds like a good plan to me.

We could also potentially go with @dsplaisted's preferred approach of implementing implicits entirely in the evaluator, but my first attempts to do so were stymied by assumptions about how the evaluation tree and the XML tree were intertwined.

jeffkl:
Had to do some hacky stuff in the preprocessor but all of that code is private and we can make it better in the future.

Closes #1447

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
@@ -455,11 +455,6 @@ namespace Microsoft.Build.Construction
         /// </remarks>
         internal void AddToXml(ProjectElement child)
         {
-            if (child.IsImplicit)
-            {
-                return;
-            }
-
             if (child.ExpressedAsAttribute)
             {
                 // todo children represented as attributes need to be placed in order too
@@ -485,7 +480,7 @@ namespace Microsoft.Build.Construction
                 // If none is found, then the node being added is inserted as the only node of its kind
 
                 ProjectElement referenceSibling;
-                Predicate<ProjectElement> siblingIsExplicitElement = _ => _.ExpressedAsAttribute == false && _.IsImplicit == false;
+                Predicate<ProjectElement> siblingIsExplicitElement = _ => _.ExpressedAsAttribute == false;
 
                 if (TrySearchLeftSiblings(child.PreviousSibling, siblingIsExplicitElement, out referenceSibling))
                 {
@@ -567,11 +562,6 @@ namespace Microsoft.Build.Construction
 
         internal void RemoveFromXml(ProjectElement child)
         {
-            if (child.IsImplicit)
-            {
-                return;
-            }
-
             if (child.ExpressedAsAttribute)
             {
                 XmlElement.RemoveAttribute(child.XmlElement.Name);
@@ -630,10 +620,7 @@ namespace Microsoft.Build.Construction
 
             _count++;
 
-            if (!child.IsImplicit)
-            {
-                MarkDirty("Add child element named '{0}'", child.ElementName);
-            }
+            MarkDirty("Add child element named '{0}'", child.ElementName);
         }
 
         /// <summary>
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -158,11 +158,6 @@ namespace Microsoft.Build.Construction
         private BuildEventContext _buildEventContext;
 
         /// <summary>
-        /// Xpath expression that will find any element with the implicit attribute
-        /// </summary>
-        private static readonly string ImplicitAttributeXpath = $"//*[@{XMakeAttributes.@implicit}]";
-
-        /// <summary>
         /// Initialize a ProjectRootElement instance from a XmlReader.
         /// May throw InvalidProjectFileException.
         /// Leaves the project dirty, indicating there are unsaved changes.
@@ -655,6 +650,25 @@ namespace Microsoft.Build.Construction
         }
 
         /// <summary>
+        /// 
+        /// </summary>
+        public string Sdk
+        {
+            [DebuggerStepThrough]
+            get
+            {
+                return ProjectXmlUtilities.GetAttributeValue(XmlElement, XMakeAttributes.sdk);
+            }
+
+            [DebuggerStepThrough]
+            set
+            {
+                ProjectXmlUtilities.SetOrRemoveAttribute(XmlElement, XMakeAttributes.sdk, value);
+                MarkDirty("Set project Sdk to '{0}'", value);
+            }
+        }
+
+        /// <summary>
         /// Gets or sets the value of TreatAsLocalProperty. If there is no tag, returns empty string.
         /// If the value being set is null or empty, removes the attribute.
         /// </summary>
@@ -710,10 +724,8 @@ namespace Microsoft.Build.Construction
                 {
                     using (ProjectWriter projectWriter = new ProjectWriter(stringWriter))
                     {
-                        var xmlWithNoImplicits = RemoveImplicits();
-
-                        projectWriter.Initialize(xmlWithNoImplicits);
-                        xmlWithNoImplicits.Save(projectWriter);
+                        projectWriter.Initialize(XmlDocument);
+                        XmlDocument.Save(projectWriter);
                     }
 
                     return stringWriter.ToString();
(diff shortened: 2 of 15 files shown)


jeffkl:

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
```diff
@@ -630,10 +620,7 @@ private void AddInitialChild(ProjectElement child)
 
             _count++;
 
-            if (!child.IsImplicit)
-            {
-                MarkDirty("Add child element named '{0}'", child.ElementName);
-            }
+            MarkDirty("Add child element named '{0}'", child.ElementName);
```
AndyGerlicher: This looks like a change in behavior. Does it need to happen for Remove just above this? Or others?

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -655,6 +650,25 @@ public string InitialTargets
         }
 
         /// <summary>
+        ///
```
AndyGerlicher: Fill in the xmldoc

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -849,6 +861,14 @@ public ElementLocation InitialTargetsLocation
...
```
AndyGerlicher: Here too

src/XMakeBuildEngine/Construction/ProjectElementContainer.cs
jeffkl: This is removed because I'm reverting the original implementation where the stuff is added to the XML DOM.  There is no `IsImplicit` property anymore so the `if` is removed.

The change above this is removed for the same reason.

jeffkl:
@dsplaisted They imports show up in the `Project` but not in the `ProjectRootElement`.  When you new up a `Project`, the evaluation occurs and the imports are added without a location.  The properties and items do show they are "imported" which I'm hoping is enough for CPS.  I'm not familiar enough with CPS to know all of the implications, do you know if them showing up in the Project but not the XML representation will break CPS?

The `Sdk` attribute on the `ProjectRootElement` is how they can be controlled but that's the only thing exposed.

cdmihai:
@jviau, is this OK for CPS (see last two comments)?

jviau:
@jeffkl @cdmihai CPS uses `ElementLocation` at the `ProjectRootElement` location for determining how to modify the project file. So this will be a breaking behavior change for us. With this change, for a `ProjectItemElement` at the `ProjectRootElement` level, how would we tell if it is from an SDK import, and if it is the top import or the bottom import?

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -655,6 +650,28 @@ public string InitialTargets
...
+        /// If  a value is specified, an Sdk.props is simplicity imported at the top of the project and an
+        /// Sdk.targets is simplicity imported at the bottom from the specified SDK.
+        /// If the value is null or empty, removes the attribute.
+        /// </summary>
+        public string Sdk
+        {
+            [DebuggerStepThrough]
+            get
+            {
+                return ProjectXmlUtilities.GetAttributeValue(XmlElement, XMakeAttributes.sdk);
```
cdmihai: As things get more involved with the SDKs, it might be worth creating a new SDKProjectElement which is represented as an attribute (reusing the way metadata as attributes represent themselves). This type would be a valid child for ProjectRootElements and ProjectImportElements. Then, the new type could gain responsibilities such as parsing the semicolon delimited string into a {name, version} object, doing data validation, etc

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -987,6 +987,42 @@ private void PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
                 DebuggerManager.BakeStates(Path.GetFileNameWithoutExtension(currentProjectOrImport.FullPath));
             }
 #endif
+            IList<ProjectImportElement> implicitImports = new List<ProjectImportElement>();
+
+            if (!String.IsNullOrWhiteSpace(currentProjectOrImport.Sdk))
+            {
+                // SDK imports are added implicitly where they are evaluated at the top and bottom as if they are in the XML
+                //
+                foreach (string sdk in currentProjectOrImport.Sdk.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
```
cdmihai: This splitting logic is an example of what could go in a new SDKProjectElement type, since most likely other parts of the codebase may want to parse the SDK string.

src/XMakeBuildEngine/Evaluation/Evaluator.cs
```diff
@@ -2307,7 +2355,13 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
                 return LoadImportsResult.ConditionWasFalse;
             }
 
-            return ExpandAndLoadImportsFromUnescapedImportExpression(directoryOfImportingFile, importElement, importElement.Project, throwOnFileNotExistsError, out projects);
+            string project = importElement.Project;
+            if (!String.IsNullOrWhiteSpace(importElement.Sdk))
+            {
+                project = Path.Combine(BuildEnvironmentHelper.Instance.MSBuildSDKsPath, importElement.Sdk, "Sdk", project);
```
cdmihai: This logic is duplicated. I guess that ultimately this bit of code would use an IAsserResolver and a potential SDKProjectElement to do its business.

src/XMakeBuildEngine/Evaluation/ProjectParser.cs
```diff
@@ -187,61 +187,6 @@ private void ParseProjectElement(XmlElementWithLocation element)
             // so we have to set it now
             _project.SetProjectRootElementFromParser(element, _project);
 
-            if (element.HasAttribute(XMakeAttributes.sdk))
```
cdmihai: Shouldn't this method (and the other thing that can have SDKs, ParseImportElement too) do some SDK string validation, such as ensure it's not empty?

src/XMakeBuildEngine/UnitTestsPublicOM/Construction/ProjectSdkImplicitImport_Tests.cs
```diff
@@ -29,24 +30,36 @@ public void SdkImportsAreInImportList()
                 string sdkPropsPath = Path.Combine(testSdkDirectory, "Sdk.props");
                 string sdkTargetsPath = Path.Combine(testSdkDirectory, "Sdk.targets");
 
-                File.WriteAllText(sdkPropsPath, "<Project />");
```
cdmihai: Can you also do a ProjectSdkExplicitImport_Tests class to test out the explit SDK imports? Or rename this class to refer to both implcit and explicit SDK elements.

jeffkl:
@dsplaisted 
> will this affect CPS and its understanding of provenance?

I need to understand CPS more to answer this.  If it uses the evaluation APIs first, the imports show up in the graph and everything has the `IsImported` property flag indicating that it came from an import.  If it uses the construction APIs for reading the project, then I might have to rethink more of this.  I've pinged Jacob to get more info but perhaps you can fill me in?

> Also, I'd suggest updating Evaluator.GetCurrentDirectoryForConditionEvaluation to return the Sdk directory if there is an Sdk attribute on an import, so that you can have something like Condition="Exists('Sdk\Sdk.props')" on an Sdk import.

Good idea, I'll do this as part of the SDK resolver work and I've added your comment as a note on the Design doc https://github.com/Microsoft/msbuild/issues/1493

jeffkl:
@cdmihai The ProjectParser does not parse any of the `ProjectRootElement` attributes (InitialTargets, DefaultTargets, etc).  They are all read by the evaluator as raw XML strings.  Let's talk in person about your comments when you're back.

jeffkl:
@dotnet-bot test OSX Build for Mono please

jeffkl:
@dotnet-bot test OSX Build for Mono please

rainersigwald:
Related to #1493.

PR #1497, opened 2016-12-19
Generate targets in metaproj

### LINKED ISSUE #1275: Solution targets need to be dynamic or at least extensible
There's currently a fixed set of targets emitted to the sln metaproj:

https://github.com/Microsoft/msbuild/blob/2db4d4b6821ff65987b2843d29cc5f02c4db59f5/src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs#L765-L769 (edited)

We will be needing other targets such as "pack" and "test" to work on solutions. 

@jeffkl suggested "It might be as simple as taking the target you specified at the command-line and injecting it into the generated project"

jeffkl:
This allows users to access any target when building a solution.  The target may or may not exist but this will create the corresponding targets in the `.metaproj`.

I had to pass along the target names from the build request all the way down to the solution generator via ReadOnlyCollections.

Closes #1275

### DIFF
--- a/src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs
+++ b/src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs
@@ -142,6 +142,9 @@ namespace Microsoft.Build.BackEnd
 
         #endregion
 
+        /// <summary>
+        /// The target names that were requested to execute.
+        /// </summary>
         internal IReadOnlyCollection<string> TargetNames { get; }
 
         /// <summary>
--- a/src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs
+++ b/src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs
@@ -7,7 +7,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.Globalization;
 using System.Text;
 using Microsoft.Build.Construction;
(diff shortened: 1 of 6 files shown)


src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -140,6 +142,8 @@ internal class BuildRequestConfiguration : IEquatable<BuildRequestConfiguration>
 
         #endregion
 
+        internal IReadOnlyCollection<string> TargetNames { get; }
```
AndyGerlicher: This could probably use a doc comment.

jeffkl:
@cdmihai Please review

src/XMakeBuildEngine/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
```
cdmihai: Who needs this namespace?

src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
```diff
@@ -104,6 +112,11 @@ internal class SolutionProjectGenerator
         private ILoggingService _loggingService;
 
         /// <summary>
+        /// The list of targets specified to use.
+        /// </summary>
+        private readonly IReadOnlyCollection<string> _targetNames = new Collection<string>();
```
cdmihai: Ah, not all paths write to this fields. Ignore.

---
In reply to: [93343487](https://github.com/Microsoft/msbuild/pull/1497#discussion_r93343487) [](ancestors = 93343487)

Issue #1472, opened 2016-12-13
Commenting out csproj elements moves them to top on save


333fred:
Related to #1442. Moved from https://github.com/dotnet/roslyn-project-system/issues/932.
@onotovy said:
> When I have a CSproj open for editing, if I comment-out a property group, when I hit save, it moves the commented area to the top of the file, right under project. That's startling and wrong as it moves items from where I want. Further, if it can change the meaning if they're uncommented and not restored to the correct location (which someone may not know where it belongs).

Again, I'm accessing the XML by going through the RawXml property: https://github.com/dotnet/roslyn-project-system/blob/dev15-rc2/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Utilities/IMsBuildAccessor.cs#L53.

FYI @AndyGerlicher @rainersigwald

Sarabeth-Jaffe-Microsoft:
Thank you for letting is know about this bug, @333fred. This sounds like a very worrisome developer experience. @cdmihai please look into this.

onovotny:
It's more than that...if you paste in an XML snippet that contains a comment and hit save, it appears to remove the comment from the editor. When you look on disk, it's really there.

cdmihai:
Can you please write out a minimal repro in terms of MSBuild APIs that exposes this issue? That way I can debug and see what happens. See my comment from the related issue: https://github.com/Microsoft/msbuild/issues/1442#issuecomment-266917716

cdmihai:
Should be fixed by #1507

Issue #1442, opened 2016-12-06
ProjectRootElement.RawXml Never contains updated comments


333fred:
As the title says, updated comments aren't reflected in the output of ProjectRootElement.RawXml. This causes https://github.com/dotnet/roslyn-project-system/issues/743. If there's a better api for getting what XML would be actually written out when saved, I'd be more than happy to use it, as RawXml seems to not be intended for that purpose.

dsplaisted:
I think what you want to use is [`ProjectRootElement.Save(TextWriter)`](https://github.com/Microsoft/msbuild/blob/fa4c4f350617d20ad78c68e0a44d7be449e96260/src/XMakeBuildEngine/Construction/ProjectRootElement.cs#L1836).  The [whitespace preservation tests](https://github.com/Microsoft/msbuild/blob/fa4c4f350617d20ad78c68e0a44d7be449e96260/src/XMakeBuildEngine/UnitTestsPublicOM/Construction/WhiteSpacePreservation_Tests.cs), for example, call `Project.Save(TextWriter)`, which just forwards the call to the method on `ProjectRootElement`.

333fred:
`RawXml` actually calls Save under the hood with a StringWriter. I tried it anyway just to be sure, and it has the same behavior.

rainersigwald:
I suspect the root of the problem here is that the MSBuild side of the object tree doesn't know anything about comments, so if we walk it to update we won't see comment nodes.

dsplaisted:
Strange.  The whitespace preservation tests use the same APIs to save to validate that not only are comments preserved, but that the whitespace around them is preserved (if preserveWhitespace is enabled).  I suspect the difference lies in how the project is loaded.

rainersigwald:
Could also be a post-clone problem.

333fred:
@dsplaisted , `RawXml`, and by extension `Save`, are also not showing the whitespace.

cdmihai:
@333fred / @dsplaisted 

Can you please write out a minimal repro that, using MSBuild's APIs shows this issue? I'm asking because the MSBuild APIs do not expose the xml comments, so I am curious how you get into this state. I would expect the editor to tear down and recreate the `Project` / `ProjectRootElement` objects after the contents got changed. Otherwise the underlying XmlElement object model does not get updated, hence `Save` does not reflect the updated comments.

jviau:
[...]

PR #1507, opened 2016-12-22
ProjectRootElement can reload itself

cdmihai:
This would enable IDEs to keep reloading a ProjectRootElement when the user changes the file instead of doing reference swapping or relying on `DeepCopyFrom` / `DeepClone` which do not preserve XML trivia (whitespace and comments).

This helps address #1442 and #1472.

Unclear points to discuss, probably best left for a future PR:
- Should the ProjectRootElementCache reload instead of forgetting entries when the white space preservation missmatches or `autoReloadFromDisk` is true?
  - pros
    - references stay the same so no weird stale issues (e.g. an evaluated Project gets stale if the Cache decides to forget the ProjectRootElements the Project references, and would stop receiving changed events.). An alternative to this is to forget entries but pipe through new events so evaluated projects swap their references.
  - cons
    - the Cache would throw the exceptions Reload throws, potentially breaking a lot of stuff that relies on `ProjectCollection` :). For example, all in-memory ProjectRootElements with a path would throw on Reload because their paths are not yet materialized. Or if there's syntax error, etc.

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -1928,7 +1928,13 @@ namespace Microsoft.Build.Construction
             // Reload should only mutate the state if there are no parse errors.
             ThrowIfDocumentHasParsingErrors(document);
 
+            // Do not clear the string cache.
+            // Based on the assumption that Projects are reloaded repeatedly from their file with small increments,
+            // and thus most strings would get reused
+            //this.XmlDocument.ClearAnyCachedStrings();
+
             this.RemoveAllChildren();
+
             ProjectParser.Parse(document, this);
 
             MarkDirty("Project reloaded", null);
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -266,7 +266,7 @@ namespace Microsoft.Build.Construction
 
         /// <summary>
         /// Initialize a ProjectRootElement instance from an existing document.
-        /// Helper constructor for the <see cref="ReloadFrom(XmlDocumentWithLocation)"/>> which needs to check if the document parses
+        /// Helper constructor for the <see cref="ReloadFrom(string,bool,System.Nullable{bool})"/>> mehtod which needs to check if the document parses
         /// </summary>
         /// <remarks>
         /// Do not make public: we do not wish to expose particular XML API's.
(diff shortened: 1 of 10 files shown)


cdmihai:
@jviau

src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs
```diff
@@ -194,7 +194,7 @@ internal ProjectRootElementCache(bool autoReloadFromDisk)
         /// <param name="preserveFormatting"><code>true</code> to the project was loaded with the formated preserved, otherwise <code>false</code>.</param>
         /// <returns>The ProjectRootElement instance if one exists.  Null otherwise.</returns>
         internal ProjectRootElement Get(string projectFile, OpenProjectRootElement openProjectRootElement, bool isExplicitlyLoaded,
-            bool preserveFormatting)
+            bool? preserveFormatting)
```
jviau: Change `ProjectRootElement.Open` and `ProjectRootElement.TryOpen` to pass `preserveFormatting: null` here when the overload without `preserveFormatting` is used.

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1856,6 +1870,84 @@ public ProjectRootElement DeepClone()
...
+            ThrowIfDocumentHasParsingErrors(document);
+
+            // Do not clear the string cache.
+            // Based on the assumption that Projects are reloaded repeatedly from their file with small increments,
+            // and thus most strings would get reused
+            //this.XmlDocument.ClearAnyCachedStrings();
+
+            this.RemoveAllChildren();
+
+            ProjectParser.Parse(document, this);
```
jviau: Was it not possible to take over the parsing result of the `ProjectRootElement` created in `ThrowIfDocumentHasParsingErrors` instead of re-parsing?

src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs
jviau: Change `ProjectRootElement.Open` and `ProjectRootElement.TryOpen` to pass `preserveFormatting: null` here when the overload without `preserveFormatting` is used.

src/XMakeBuildEngine/UnitTestsPublicOM/Definition/Project_Tests.cs
```diff
@@ -1439,6 +1439,207 @@ public void ImportedXmlModified()
                 File.Delete(path);
             }
         }
+        
+        /// <summary>
+        /// Adding an import to an existing PRE object and re-valuating should preserve the initial import PRE object
```
jviau: nit: "re-valuating" -> "re-evaluating"

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
cdmihai: The parser mutates the state of the root element (sets both XmlNode and ProjectElement elements) while it's parsing, so if it encounters an error you get a partially mutated root element. So we'll have to manually copy over the state from the test root element. Left it as a todo since it's as fast as the previous usage which also parsed twice, and it's worth doing after we flesh out the other aspects of Reload.

src/XMakeBuildEngine/Evaluation/ProjectRootElementCache.cs
cdmihai: At some point in the ProjectRootElement construction chain something has to choose either true or false for `PreserverFormatting`, otherwise the nullable bool flows all the way into the ProjectRootElement constructor (via very ambiguous detours through the cache callbacks). It gets interesting how the constructor should interpret the null value, maybe leave it off to the `System.Xml` default?

I've started thinking that we shouldn't expose format preservation in the cache, and instead change the ProjectRootElement such that you can mutate its `PreserveFormatting` property. Makes the Cache code easier, and follows the general pattern in the code that ProjectRootElement instances mutate and send change notifications rather than get evicted from the cache. The only cache eviction code is the `autoReloadFromDisk` feature, which is turned off either way (and maybe it should use Reload instead eviction :) ).

jviau: If `preserveFormatting` is not specified then the creation of the project should default to `false`, as that is the previous behavior. However, your second idea sounds good as well. The goal of the changes to preserve formatting is to be able to load it once with the formatting defaulted *or* specified, but then any subsequent loads you have the option to say "don't care, give me what you have already loaded". Without the nullable bool behavior exposed, calls to `Open` and `TryOpen` cannot say they just want whatever is already loaded.

cdmihai:
@AndyGerlicher @jviau 
I propagated the nullable `PreserveFormatting` to the public APIs and changed the Cache so that it reloads when the formatting setting does not match instead of replacing the object.

Will merge if tests pass.

PR #1516, opened 2017-01-06
Enable build summary text for more scenarios.

### LINKED ISSUE #1513: More control over printing with console logger
Reference: https://github.com/dotnet/cli/issues/5110

Currently there is no way to make a build and print only:
- warnings,
- errors,
- build status

This scenario is something which would improve experience in dotnet/cli.

The missing scenario would be that one of:
```
msbuild /v:n /clp:Summary;WarningsOnly
msbuild /v:m /clp:Summary;WarningsOnly
```
would print output which would look something like:
```
Microsoft (R) Build Engine version 15.1.458.808

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:03.78
```

Currently there is no way of removing info about all the targets being run in between.

AndyGerlicher:
Always show the Time Elapsed and Build Succeeded text whenever the user
specifies the 'summary' console property flag regardless of log verbosity
or presence of the ErrorsOnly / WarningsOnly flags.

Closes #1513

### DIFF
--- a/src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
+++ b/src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
@@ -1020,7 +1020,13 @@ namespace Microsoft.Build.BackEnd.Logging
 
             if (showOnlyWarnings || showOnlyErrors)
             {
-                _showSummary = false;
+                if (!_showSummary.HasValue)
+                {
+                    // By default don't show the summary when the showOnlyWarnings / showOnlyErrors is specified.
+                    // However, if the user explicitly specified summary or nosummary, use that.
+                    _showSummary = false;
+                }
+
                 this.showPerfSummary = false;
             }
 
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
```diff
@@ -1020,7 +1020,9 @@ public virtual void Initialize(IEventSource eventSource)
 
             if (showOnlyWarnings || showOnlyErrors)
             {
-                _showSummary = false;
+                if (!_showSummary.HasValue)
+                    _showSummary = false;
```
jeffkl: Add `{ }` here and a comment that we default to false if the user didn't specify `Summary`

src/XMakeBuildEngine/Logging/BaseConsoleLogger.cs
```diff
@@ -1020,7 +1020,9 @@ public virtual void Initialize(IEventSource eventSource)
...
+
                 this.showPerfSummary = false;
```
jeffkl: We we going to do the same thing for `PerformanceSummary`?

AndyGerlicher: I didn't because performance summary wasn't a `bool?`. `summary` has the option of `summary`, `nosummary`, or not specified. The perf one doesn't so I left as is.

PR #1519, opened 2017-01-06
Enable ngen on multiple architectures for all DLLs

AndyGerlicher:


TyOverby:
`vs.file.ngen=all` is illegal.  The only valid options are `yes` and `no`.  However, even with `vs.file.ngen=yes`, the file will only be ngen on 32-bit platforms.

I think that the correct solution would be 

`vs.file.ngenArchitecture = x86` for all .dll and .exe files in /bin and `vs.file.ngenArchitecture = x64` for all files in /bin/amd64

PR #1520, opened 2017-01-07
Only include DesignTimeFacadeAssemblies which are not in ReferencePath

### LINKED ISSUE #1345: ImplicitlyExpandDesignTimeFacades target should only add facades if a higher version isn't already referenced
The `ImplicitlyExpandDesignTimeFacades` target adds references (specifically `ReferencePath` items) to the System.Runtime surface facades under the Facades folder in the reference assembly folder.  However, we now also ship facades as part of NuGet packages.  The facades from a NuGet package may conflict with the facades that are implicitly added via this target.

An example of an issue due to this is in dotnet/sdk#367.  To repro this issue, create an ASP.NET Core web application targeting .NET Framework, update the `TargetFramework` property to `net461`, and run `dotnet publish`.  You'll get an error because a `Copy` task is trying to copy two different versions of `System.Runtime.Serialization.Primitives.dll` to the publish folder.

I discussed this with @weshaggard and it is likely that in the future we will ship updated facades in the targeting packs that should be chosen over previous versions in NuGet packages.  So we likely need a task that examines the `ReferencePath` items that come from `ResolveAssemblyReferences` and if there are any references with the same filename as an item in the Facades folder, pick the one with the higher Assembly Version.

cdmihai:
Resolves #1345

### DIFF
--- a/src/XMakeTasks/Microsoft.NETFramework.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.NETFramework.CurrentVersion.targets
@@ -108,7 +108,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
       <_DesignTimeFacadeAssemblies_Names Remove="@(_ReferencePath_Names)"/>
 
-      <ReferencePath Include="@(_DesignTimeFacadeAssemblies_Names->%(OriginalIdentity))">
+      <ReferencePath Include="@(_DesignTimeFacadeAssemblies_Names->'%(OriginalIdentity)')">
         <WinMDFile>false</WinMDFile>
         <CopyLocal>false</CopyLocal>
         <ResolvedFrom>ImplicitlyExpandDesignTimeFacades</ResolvedFrom>
--- a/src/XMakeTasks/Microsoft.NETFramework.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.NETFramework.CurrentVersion.targets
@@ -97,9 +97,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <ItemGroup Condition="'$(_HasReferenceToSystemRuntime)' == 'true'">
       <_DesignTimeFacadeAssemblies Include="%(DesignTimeFacadeDirectories.Identity)*.dll"/>
-      <!-- If the user has manually referenced these assemblies, remove them so we don't end up with duplicates -->
-      <ReferencePath Remove="@(_DesignTimeFacadeAssemblies)"/>
-      <ReferencePath Include="%(_DesignTimeFacadeAssemblies.Identity)">
+
+      <_DesignTimeFacadeAssemblies_Names Include="@(_DesignTimeFacadeAssemblies->'%(FileName)')">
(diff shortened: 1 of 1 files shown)


cdmihai:
@dsplaisted @ericstj

I guess you can delete the fix from the SDK now: https://github.com/dotnet/sdk/blob/7eb264f9cc5153b24793878307c4c410bd03ff68/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.targets#L285-L320

cdmihai:

ericstj:
I was taking another look at this the other day, and while I think this is the smallest tactical fix I noticed something that might be better.

ImplicitlyExpandDesignTimeFacades always adds all facades and causes conflicts, but I noticed that RAR is already getting the path to the facades via [$(TargetFrameworkDirectory)](https://github.com/Microsoft/msbuild/blob/052ad30f8f706a87e9fa9431fda1ec9c0b73893b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L1142).  I then see that RAR actually selects the subset of facades that are actually needed and outputs them via `ReferenceDependencyPaths`.  What surprised me is that we didn't pass this item to CSC, but other compilers did accept this item.  @davkean / @weshaggard do you remember why that is the case?

weshaggard:
>@weshaggard do you remember why that is the case?

No I don't recall why that is the case. I didn't even know RAR was getting the path to the facades at all.

cdmihai:
If that's the case, then it looks like `ImplicitlyExpandDesignTimeFacades` would become obsolete if only `ReferenceDependencyPaths` flowed into the the compiler.

ericstj:
@cdmihai yeah, I think so.  I just don't know why that wasn't done to begin with.  Maybe @davkean or someone with more history in this codebase could chime in.

davkean:
Yeah I don't history on this - I wasn't aware of it.

cdmihai:
Including some of the committers from [the new home of the csc target](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets#L71), maybe they know more.

@tannergooding, @agocke, @AArnott, do you know more about @ericstj's [proposal](https://github.com/Microsoft/msbuild/pull/1520#issuecomment-271342350) of deleting the [ImplicitlyExpandDesignTimeFacades](https://github.com/Microsoft/msbuild/blob/cb8c727a898e4439a3cd18f329504e436b0b7c00/src/XMakeTasks/Microsoft.NETFramework.CurrentVersion.targets#L91) target and changing the CSC invocation to include the output of RAR, [ReferenceDependencyPaths](https://github.com/Microsoft/msbuild/blob/052ad30f8f706a87e9fa9431fda1ec9c0b73893b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L1955), which supposedly contains the target framework facade dlls? If not, I'll start experimenting with it and see what breaks.

agocke:
@cdmihai That code is before my time, so unfortunately my advice is just try it and see :)

cdmihai:

AndyGerlicher:

AArnott:
I'm afraid I'm not this deeply familiar with resolving references. @crmann1 may know.

PR #1530, opened 2017-01-09
Update German translation

cdmihai:
Copied over from the internal MSBuild repository as the internal localization files are not used anymore.

### DIFF
--- a/src/Utilities/Resources/xlf/Strings.de.xlf
+++ b/src/Utilities/Resources/xlf/Strings.de.xlf
@@ -185,7 +185,7 @@
       </trans-unit>
       <trans-unit id="Tracking_SourceWillBeCompiledDependencyWasModifiedAt">
         <source>{0} will be compiled as {1} was modified at {2}.</source>
-        <target state="translated">{0} wird als {1} kompiliert, wurde geändert bei {2}.</target>
+        <target state="translated">{0} wird als {1} kompiliert. Letzte Änderung: {2}.</target>
         <note />
       </trans-unit>
       <trans-unit id="Tracking_SourceWillBeCompiled">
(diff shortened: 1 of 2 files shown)


src/Utilities/Resources/xlf/Strings.de.xlf
```diff
@@ -185,7 +185,7 @@
       </trans-unit>
       <trans-unit id="Tracking_SourceWillBeCompiledDependencyWasModifiedAt">
         <source>{0} will be compiled as {1} was modified at {2}.</source>
-        <target state="translated">{0} wird als {1} kompiliert, wurde geändert bei {2}.</target>
```
dasMulli: Does "as" mean "because" in this context? If so, I suspect a translation error..

dasMulli: (And `Tracking_SourceWillBeCompiledMissingDependency` will be wrong as well).
The current translation assumes that "as" is meant as identity.. ("like a ...", "into ...") and not "because"

rainersigwald: Thanks! Maybe we should update the English to "because"? It's clearer, and I _hope_ no one is scraping logs for this exact string--but they might be.

akoeplinger: Yep, I think that'd make the sentence clearer.

dasMulli: 👍 for changing the English text
If not, xlf has a note field for issues like this..

cdmihai:
Replaced "as" with "because" everywhere :)
In this case the translation becomes irrelevant, because in the next loc cycle all the localized strings will be marked as needing revision.

PR #1532, opened 2017-01-10
Fix issue where ToolTask will find directories instead of files

### LINKED ISSUE #1215: Search for ToolTask executable on path finds directories
The CLI repo has a tool task that [returns "git" from GenerateFullPathToTool](https://github.com/dotnet/cli/blob/b7ac730a7f0cf6013ed51bc45a299dec7efb8cfb/build_projects/dotnet-cli-build/GetCommitHash.cs#L27-L30)

On my machine, this attempts to execute C:\Program Files (x86)\Microsoft Visual Studio\VS15Preview\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer\git, which is a directory. 

There's an actual "git.exe" on my path, but it appears to lose no matter how my path is ordered. I presume because the directory matches the filename exactly whereas the executable adds an extension.

jeffkl:
The native function SearchPath will return a directory which isn't what we want.  The bug was only found on Windows because our FindOnPath() method was doing different things on Windows vs Non-windows.

This change unifies the implementation of FindOnPath() and only looks at files.  The method is only called by ToolTask since I removed the other reference to it so I moved it and the unit tests out of NativeMethodsShared and into ToolTask.

I've also written a test which failed before the fix and passes now.

Closes #1215


jeffkl:
@rainersigwald I have addressed your comments and fixed merge conflicts.

PR #1522, opened 2017-01-07
Update System.Collections.Immutable to 1.2.1

AndyGerlicher:
Update to match change in Roslyn tools.

PR #1534, opened 2017-01-10
Update nuspecs for #1522

rainersigwald:
These changes were made automatically via `cibuild.cmd --target
CoreCLR`, and update the checked-in specs to match the new references.


rainersigwald:
I spot-checked several references against the latest pushed package https://dotnet.myget.org/feed/msbuild/package/nuget/Microsoft.Build/15.1.0-preview-000516-03 and they look right.

rainersigwald:
@jeffkl yeah, this happened automagically (but only when building for CoreCLR). When I did #1522 I only ran cibuild for Full on my local machine before sending it off to CI.

jeffkl:
Ah that makes sense, the dependencies only matter for .NET Core because for full framework we only depend on ourselves and stuff in the GAC/Framework.

https://github.com/Microsoft/msbuild/blob/xplat/src/dir.targets#L60

PR #1535, opened 2017-01-10
Fail the CI build if there are any uncommitted changes

jeffkl:
I'll only be updating cibuild.cmd so ignore the other change as I'm using it to test this.

### DIFF
--- a/cibuild.cmd
+++ b/cibuild.cmd
@@ -156,6 +156,20 @@ if %ERRORLEVEL% NEQ 0 (
     goto :error
 )
 
+:: Only detect source control changes when running in the CI environment
+:: Detect if there are any changed files which should fail the build
+if DEFINED JENKINS_URL (
+    echo Detecting changed files...
+    git status
+    git --no-pager diff HEAD --word-diff=plain --exit-code
+    if ERRORLEVEL 1 (
+        echo.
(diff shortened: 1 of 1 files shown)


cibuild.cmd
```diff
@@ -156,6 +156,20 @@ if %ERRORLEVEL% NEQ 0 (
     goto :error
 )
 
+:: Only detect source control changes when running in the CI environment
+:: Detect if there are any changed files which should fail the build
+if DEFINED JENKINS_URL (
+    echo Detecting changed files...
+    git diff --cached HEAD --exit-code
```
rainersigwald: Isn't a failure here ignored since you go to the next line?

rainersigwald: You could maybe do `git diff --exit-code HEAD` which should show all diffs between the work tree and HEAD.

Torn on `--quiet` versus `--exit-code`--it'd be nice to get a list on failure, so I guess I prefer `--exit-code`.

jeffkl: I'm no git expert, but it worked when I tested it.  I'm going to try this instead:

`git --no-pager diff HEAD --word-diff=plain --exit-code` which seems to work...

cdmihai:
Also print out the diff for ease of understanding :)

jeffkl:
@dotnet-bot test Windows_NT for Desktop please

jeffkl:
@dotnet-bot test this please

PR #1536, opened 2017-01-10
Cache IsMono call

cdmihai:
VS has potentially expensive TypeResolve events, so cache IsMono. Apparently the c++ side of things even triggers a design time build during some TypeResolve calls.

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -454,13 +454,6 @@ namespace Microsoft.Build.Shared
 
         #region Member data
 
-        static NativeMethodsShared()
-        {
-            // There could be potentially expensive TypeResolve events, so cache IsMono.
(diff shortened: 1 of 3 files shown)


cdmihai:
@radical

src/Shared/NativeMethodsShared.cs
```diff
@@ -454,6 +454,12 @@ public SystemInformationData()
 
         #region Member data
 
+        static NativeMethodsShared()
```
rainersigwald: Use a `Lazy<T>` instead of this static constructor?

cdmihai: I'm not sure that helps, since a log of stuff calls into IsMono including anything that logs events.

rainersigwald: Fair enough.

cdmihai:

cdmihai:
@dotnet-bot test Windows_NT Build for Mono please

PR #1538, opened 2017-01-11
Binplace System.Collections.Immutable to Output

rainersigwald:
This file is required by the VSIX builder, and was placed in Output as
part of the transitive dependencies of
Microsoft.Build.Tasks.CodeAnalysis, but more recent versions don't have
that dependency, so it wasn't getting placed.

Added a workaround to ensure that Immutable appears in the folder.


rainersigwald:
This makes me sad, but I couldn't think of a much-better option. Another possibility would be to get Immutable from a different location in the swr file (like the Deployment directory?) but this seemed a little easier.

PR #1545, opened 2017-01-12
Filter out project names from potential user targets

cdmihai:


jeffkl:
The point of this is to act as a pass through of targets from the solution to the project. If there are "invalid" target names and one is specified, we should probably error.

rainersigwald:
They're not really "illegal" or "invalid", they're "already in use in the metaproj" and thus shouldn't be added (again, causing the error).

src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
```diff
@@ -58,7 +58,23 @@ internal class SolutionProjectGenerator
...
+            "Build",
+            "Clean",
+            "Rebuild",
+            "Publish"
+            );
+
+        /// <summary>
+        /// Internal implementation targets that can't be used by users
+        /// </summary>
+        private static readonly ISet<string> _illegaUserTargetNames = ImmutableHashSet.Create<string>(StringComparer.OrdinalIgnoreCase,
```
rainersigwald: rename `_illega` to `_solutionGeneratedTargetNames`

cdmihai:
Replaced by #1545

PR #1557, opened 2017-01-13
Better message when Exclude appears without Include

### LINKED ISSUE #1555: Bad error when missing “include” attribute on item
[Public copy of internal bug [https://devdiv.visualstudio.com/DevDiv/MSBuild/_workitems?_a=edit&id=270144](270144)]

Step to reproduce:
1.       Create “x.proj” file(add this file in attachment)
2.       Copy it to “C:\aardvark\aardvark”
3.       Build “x.proj” using command “msbuild C:\aardvark\aardvark\x.proj”
 
Expected result: 
Build failed. The expected console message should be MSB4035: The required attribute "Include" is empty or missing from the element <type>.
 
Actual result:
Build failed, the console message is unexpected.

image

cdmihai:
Resolves #1555

### DIFF
--- a/src/XMakeBuildEngine/Resources/Strings.resx
+++ b/src/XMakeBuildEngine/Resources/Strings.resx
@@ -616,6 +616,7 @@
   </data>
   <data name="IncludeRemoveOrUpdate" UESanitized="false" Visibility="Public">
     <value>MSB4232: Static items (items not under targets) must have one of the following operations: Include, Update, or Remove.</value>
+    <comment>{StrBegin="MSB4232: "} Include, Update, and Remove should not be localized and their casing should not be changed</comment>
   </data>
   <data name="MissingTaskError" UESanitized="false" Visibility="Public">
     <value>MSB4036: The "{0}" task was not found. Check the following: 1.) The name of the task in the project file is the same as the name of the task class. 2.) The task class is "public" and implements the Microsoft.Build.Framework.ITask interface. 3.) The task is correctly declared with &lt;UsingTask&gt; in the project file, or in the *.tasks files located in the "{1}" directory.</value>
--- a/src/XMakeBuildEngine/Resources/Strings.resx
+++ b/src/XMakeBuildEngine/Resources/Strings.resx
@@ -614,6 +614,9 @@
     <comment>{StrBegin="MSB4035: "}UE: This message is shown when a user leaves off a required attribute from a project element
     e.g. &lt;UsingTask AssemblyName="foo"&gt; -- this is missing the "TaskName" attribute.</comment>
   </data>
+  <data name="IncludeRemoveOrUpdate" UESanitized="false" Visibility="Public">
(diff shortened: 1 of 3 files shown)


src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -614,6 +614,9 @@
     <comment>{StrBegin="MSB4035: "}UE: This message is shown when a user leaves off a required attribute from a project element
     e.g. &lt;UsingTask AssemblyName="foo"&gt; -- this is missing the "TaskName" attribute.</comment>
   </data>
+  <data name="IncludeRemoveOrUpdate" UESanitized="false" Visibility="Public">
+    <value>MSB4232: Static items (items not under targets) must have one of the following operations: Include, Update, or Remove.</value>
+  </data>
```
jeffkl: Do you need a comment that `Include`, `Update`, and `Remove` should not be localized?

cdmihai: Good point

src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -614,6 +614,9 @@
     <comment>{StrBegin="MSB4035: "}UE: This message is shown when a user leaves off a required attribute from a project element
     e.g. &lt;UsingTask AssemblyName="foo"&gt; -- this is missing the "TaskName" attribute.</comment>
   </data>
+  <data name="IncludeRemoveOrUpdate" UESanitized="false" Visibility="Public">
+    <value>MSB4232: Static items (items not under targets) must have one of the following operations: Include, Update, or Remove.</value>
```
rainersigwald: Have we used "static items" in this sense before in user-facing messages? I think I'd rephrase to just `Items that are outside Target elements must . . .`, which matches https://msdn.microsoft.com/en-us/library/ms171453.aspx#Anchor_1

src/XMakeBuildEngine/Resources/Strings.resx
```diff
@@ -615,8 +615,8 @@
     e.g. &lt;UsingTask AssemblyName="foo"&gt; -- this is missing the "TaskName" attribute.</comment>
   </data>
   <data name="IncludeRemoveOrUpdate" UESanitized="false" Visibility="Public">
-    <value>MSB4232: Static items (items not under targets) must have one of the following operations: Include, Update, or Remove.</value>
-    <comment>{StrBegin="MSB4232: "} Include, Update, and Remove should not be localized and their casing should not be changed</comment>
+    <value>MSB4232: Items that are outside Target elements must have one of the following operations: Include, Update, or Remove.</value>
+    <comment>{StrBegin="MSB4232: "} Target, Include, Update, and Remove should not be localized and their casing should not be changed</comment>
```
rainersigwald: And `Item` (but does pluralization differ for different languages?).

cdmihai: If it's not an actual xml element / keyword then I wouldn't. I've checked with the other engine resources and they don't mark the term `item` as should-not-translate.

PR #1562, opened 2017-01-13
🔀 RI RC3 branch

rainersigwald:


rainersigwald:
Abandoning because some manual fixes are required and there doesn't seem to be a way to add them after starting the merge as a PR

PR #1565, opened 2017-01-13
Reload post condition should enforce HasUnsavedChanged = false

### LINKED ISSUE #1554: Reload post condition should enforce HasUnsavedChanged = false
The two Reload overloads that load from a file should make HasUnsavedChanged = false. This means incrementing the in-memory Version and then making _versionOnDisk = Version. This makes more sense because after a reload, all the evaluated Project objects depending on the PRE should be dirtied, but the actual PRE should have HasUnsavedChanged=false because it is up to date to the file system.

The Reload overload form an XmlReader should leave HasUnsavedChanges=true

cdmihai:
Resolves #1554

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -1806,7 +1806,7 @@ namespace Microsoft.Build.Construction
                         _lastWriteTimeWhenRead = fileInfo.LastWriteTime;
                     }
 
-                    _versionOnDisk = Version;
+                    MarkAsUpToDateToDisk();
                 }
             }
 #if MSBUILDENABLEVSPROFILING 
@@ -1856,6 +1856,11 @@ namespace Microsoft.Build.Construction
                 XmlDocument.Save(projectWriter);
             }
 
+            MarkAsUpToDateToDisk();
+        }
+
+        private void MarkAsUpToDateToDisk()
+        {
             _versionOnDisk = Version;
         }
 
@@ -1892,7 +1897,11 @@ namespace Microsoft.Build.Construction
             ErrorUtilities.VerifyThrowInvalidOperation(File.Exists(path), "FileToReloadFromDoesNotExist", path);
 
             Func<bool, XmlDocumentWithLocation> documentProducer = shouldPreserveFormatting => LoadDocument(path, shouldPreserveFormatting);
-            ReloadFrom(documentProducer, throwIfUnsavedChanges, preserveFormatting);
+
+            // Reloads from a file leave this object up to date wrt to that file, so mark the object as up to date
+            Action postReloadActions = MarkAsUpToDateToDisk;
+
+            ReloadFrom(documentProducer, throwIfUnsavedChanges, preserveFormatting, postReloadActions);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


cdmihai:
@jviau

jviau:
I think `Reload` should call `IncrementVersion()` and then `MarkAsUpToDateOnDisk()`

cdmihai:
@jviau It calls `MarkDirty()` which calls `IncrementVersion()` and fires up a bunch of dirty events.

cdmihai:
@dotnet-bot test this please.

src/XMakeBuildEngine/Construction/ProjectRootElement.cs
```diff
@@ -1892,7 +1897,11 @@ public void ReloadFrom(string path, bool throwIfUnsavedChanges = true, bool? pre
             ErrorUtilities.VerifyThrowInvalidOperation(File.Exists(path), "FileToReloadFromDoesNotExist", path);
 
             Func<bool, XmlDocumentWithLocation> documentProducer = shouldPreserveFormatting => LoadDocument(path, shouldPreserveFormatting);
-            ReloadFrom(documentProducer, throwIfUnsavedChanges, preserveFormatting);
+
+            // Reloads from a file leave this object up to date wrt to that file, so mark the object as up to date
+            Action postReloadActions = MarkAsUpToDateToDisk;
+
+            ReloadFrom(documentProducer, throwIfUnsavedChanges, preserveFormatting, postReloadActions);
```
cdmihai: You're right, looks like needless complexity. I'll remove it.
I was originally going for the "there's one reload method with the core logic and the rest just personalize it" approach, but here it does not look like it's worth it.

cdmihai:
@dotnet-bot test this please

src/XMakeBuildEngine/UnitTests/Evaluation/ProjectRootElementCache_Tests.cs
```diff
@@ -189,5 +189,61 @@ public void GetProjectRootElementChangedOnDisk2()
                 File.Delete(path);
             }
         }
+
+        /// <summary>
+        /// This test replicates the scenario in https://devdiv.visualstudio.com/DevDiv/_workitems?id=366077&_a=edit
```
rainersigwald: GitHub highlights an illegal ampersand here. I think you can just cut off that part of the URL and it'll still work.

src/XMakeBuildEngine/UnitTests/Evaluation/ProjectRootElementCache_Tests.cs
```diff
@@ -189,5 +189,61 @@ public void GetProjectRootElementChangedOnDisk2()
...
+
+                var nonReloadingProject = ProjectRootElement.Create(path);
+                nonReloadingCache.AddEntry(nonReloadingProject);
+
+                nonReloadingProject.Save();
+
+                var previousReloadingProject = ProjectRootElement.Open(path, reloadingCache, true, null);
+                Assert.NotSame(nonReloadingProject, previousReloadingProject);
+
+                for (int i = 0; i < 30; i++)
```
rainersigwald: What's the significance of 30? is it just "more than 3"?

cdmihai: Yup, it was the first number after 3 that came to mind :)
Since there's no thread sleeps in the test it should go pretty fast.

cdmihai:
@dotnet-bot test this please

PR #1569, opened 2017-01-13
🔀 RI vs15-rc3

rainersigwald:


cdmihai:
Maybe not bring the Reload partial revert, since the better fix is #1565.

PR #1578, opened 2017-01-17
Change Open and TryOpen to pass preserveFormatting: null

jviau:
Change the default behavior of `ProjectRootElement.Open(String, ProjectCollection)` and `ProjectRootElement.TryOpen(String, ProjectCollection)` to pass `preserveFormatting: null`. This way calls using that overload will not trigger project reloads and set preserveFormatting to false.

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectRootElement.cs
@@ -1115,7 +1115,7 @@ namespace Microsoft.Build.Construction
         public static ProjectRootElement Open(string path, ProjectCollection projectCollection)
         {
             return Open(path, projectCollection,
-                preserveFormatting: false);
+                preserveFormatting: null);
         }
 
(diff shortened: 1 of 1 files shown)


cdmihai:
There is one potential change in behaviour. Before the format preservation changes, the setting was essentially false. If these public APIs now use null instead of false, they will get whatever `preserveFormatting` value is in the cache. So existing code using these APIs will start getting project root elements with random `preserveFormatting` values, whereas in the past they only got false.

Is this a real issue though? It will break users who use the project root element to edit the msbuild file and then work on the raw string. For example, unit tests that assert the raw contents would break.

cdmihai:
From internal discussion with @rainersigwald 

The behaviour change risk is small, since it would only impact users who opened a project root element with `true` via the new API, and then called legacy Open again later on.
The benefits on the other hand are avoiding perf degradation due to needless Reloads from the `ProjectRootElementCache`.

Issue #1549, opened 2017-01-12
x64 MSBuild doesn't locate Unit Test assemblies


rainersigwald:
Moved from https://github.com/Microsoft/vsts-tasks/issues/3359 on behalf of @Pitterling and @AbhitejJohn.

Simplified repro steps seem to be:

* Create a Unit Test project
* Build it with `amd64\MSBuild.exe`

```
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64\msbuild.exe" "C:\Agent\_work\1\s\UnitTestProject1\UnitTestProject1.sln" 
fails with 
“Could not resolve this reference. Could not locate the assembly "Microsoft.VisualStu
dio.QualityTools.UnitTestFramework". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors.”
```

https://github.com/Microsoft/vsts-tasks/issues/3359#issuecomment-272146216

> MSbuild x64 does not seem to be resolving assemblies from Public Assemblies. This is however working in Dev14 and the x86 msbuild from Dev15 where the UnitTest Framework dll is being picked up from PublicAssemblies

rainersigwald:
As a workaround, using x86 MSBuild seems to work.

Pitterling:
@rainersigwald - tried with recent RC build 26127.00 but still run into this issue ?

PR #1583, opened 2017-01-18
Read AssemblyFolders.config from ToolsPath32

AndyGerlicher:
Visualy Studio will drop this file to the root of MSBuild only. On x64
MSBuild the file won't be found unless it looks in MSBuildToolsPath32.


rainersigwald:
Also related to #1549?

PR #1589, opened 2017-01-19
updated some doc links

mairaw:
I thought we might want to point people to the latest reference of MSBuild on docs. Also, removed locale and product family version from the msdn links so they go to the latest and to the locale specified by the user in their machine.


AndyGerlicher:
Looks good, thanks!

PR #1590, opened 2017-01-19
Ensure built-in and imported targets are respected by the solution pr…

### LINKED ISSUE #1587: Solution target generation does not take into account imported targets
A target added to a project via [solution project extensibility points](http://sedodream.com/2010/10/22/MSBuildExtendingTheSolutionBuild.aspx) is no longer accessible after #1497, because the injected call-arbitrary-target-in-all-projects target overrides the one imported from `After.SolutionName.sln.targets`. This produces errors like

```
S:\msbuild>bin\Bootstrap\15.0\Bin\MSBuild.exe src\MSBuild.sln /t:PackNugetPackages /v:q
Microsoft (R) Build Engine version 15.1.523.56541

S:\msbuild\src\Framework\Microsoft.Build.Framework.csproj : error MSB4057: The target "PackNugetPackages" does not exist in the project.
S:\msbuild\src\XMakeCommandLine\MSBuild.csproj : error MSB4057: The target "PackNugetPackages" does not exist in the project.
```

This regresses solutions that use a target at the solution level to restore NuGet packages, a common pattern for Xamarin.

jeffkl:
…oject generator.

The goal is to take whatever target the user specified from the command-line and add it to the metaproj.  However, we add targets and targets can be imported.

This change ensures that command-line targets are added last after everything else and they're only added if they don't already exist.  This removes the need to know what built-in targets are there.  If the user specifies `/t:Clean`, then it's still added as a built-in target and at the end when adding user specified targets it is ignored.

Another case is for projects.  We add a targets like "Project_Name", "Project_Name:Clean" where the project name becomes a target.  However we also add sub-targets which need to be respected.  These are also now added before the user specified targets are added.  If you specify `/t:Project_Name:Clean`, it won't be added to the project as a user specified target.

Unit test has been updated to hopefully test all built-in targets.

Closes #1587

### DIFF
--- a/src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
+++ b/src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
@@ -872,7 +872,7 @@ namespace Microsoft.Build.Construction
             // These are just dummies necessary to make the evaluation into a project instance succeed when 
             // any custom imported targets have declarations like BeforeTargets="Build"
             // They'll be replaced momentarily with the real ones.
-            foreach (string targetName in _defaultTargetNames.Union(_targetNames))
+            foreach (string targetName in _defaultTargetNames)
             {
                 traversalProject.AddTarget(targetName);
             }
@@ -897,7 +897,7 @@ namespace Microsoft.Build.Construction
                 );
 
             // Make way for the real ones                
-            foreach (string targetName in _defaultTargetNames.Union(_targetNames))
+            foreach (string targetName in _defaultTargetNames)
             {
                 traversalInstance.RemoveTarget(targetName);
             }
--- a/src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
+++ b/src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
@@ -66,17 +66,6 @@ namespace Microsoft.Build.Construction
             );
 
         /// <summary>
-        /// Internal implementation targets that can't be used by users
-        /// </summary>
-        internal static readonly ISet<string> _solutionGeneratedTargetNames = ImmutableHashSet.Create<string>(StringComparer.OrdinalIgnoreCase,
-            "ValidateSolutionConfiguration",
-            "ValidateToolsVersions",
-            "ValidateProjects",
-            "GetSolutionConfigurationContents",
-            "GetFrameworkPathAndRedistList"
-            ).Union(_defaultTargetNames);
-
-        /// <summary>
         /// Version 2.0
         /// </summary>
         private readonly Version _version20 = new Version(2, 0);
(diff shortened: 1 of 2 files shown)


src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
```diff
@@ -1942,6 +1917,13 @@ private void AddTraversalTargetForProject(ProjectInstance traversalProject, Proj
                 actualTargetName += ":" + targetToBuild;
             }
 
+            // Don't add the target again.  The user might have specified /t:Project:target which was already added but only this method knows about Project:Target so
+            // after coming up with that target name, it can check if it has already been added.
+            if (traversalProject.Targets.ContainsKey(actualTargetName))
```
rainersigwald: Under what circumstances is this not redundant with the filtering you do [here](https://github.com/Microsoft/msbuild/pull/1590/files#diff-9d49d0d91290d9f1e0761bdfc60d3533R777)?

src/XMakeBuildEngine/UnitTests/Construction/SolutionProjectGenerator_Tests.cs
```diff
@@ -2106,14 +2106,35 @@ public void IllegalUserTargetNamesDoNotThrow()
...
+
+                Assert.Equal(12, instances[0].TargetsCount);
             }
+
+
+            instances = SolutionProjectGenerator.Generate(solution, null, null, BuildEventContext.Invalid, null, new[] { "Foo" });
+
+            Assert.Equal(1, instances.Length);
+
+            Assert.Equal(14, instances[0].TargetsCount);
```
rainersigwald: Since it was a late regression, I'd love to see an explicit test of the after.sln.targets scenario. Would that be prohibitively hard to produce?

src/XMakeBuildEngine/Construction/Solution/SolutionProjectGenerator.cs
jeffkl: If the project name contains a `.` or if you have two projects with the same name (like foo.csproj and foo.vbproj) then only this method knows what the final target names will be.  

I tried to explain that in the comment.  Should I reword it so it makes more sense?

src/XMakeBuildEngine/UnitTests/Construction/SolutionProjectGenerator_Tests.cs
jeffkl: _Technically_, this test class does everything in memory.  I'd have to write the solution to disk, write an `after.solution.targets` to disk, and then run the test.  What do you think?

rainersigwald:
Retargeted to RTW.

src/XMakeBuildEngine/UnitTests/Construction/SolutionProjectGenerator_Tests.cs
jeffkl: Okay I added a test for this.

Issue #1576, opened 2017-01-17
Excludes or removes in project folder don't apply to recursive includes from a parent folder


dsplaisted:
Repro project: [ProjectFolder.zip](https://github.com/Microsoft/msbuild/files/712071/ProjectFolder.zip)

Consider the following project (named `Project.proj`):

```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <File Include="..\**" Exclude="Project.proj" />
  </ItemGroup>
  <Target Name="Build">
    <Message Text="@(File)"/>
  </Target>
</Project>
```

The `File` include is recursively including all files starting in the parent directory of the project.  The `Exclude` statement is expected to prevent the `Project.proj` file to be excluded.

However, when you build the project:

**EXPECTED:** The `File` item (and hence output message) does not include `Project.proj`
**ACTUAL:** `Project.proj` is included in the file item, as `..\ProjectFolder\Project.proj`

cdmihai:
This is not a regression. Both msbuild 14 and current appear to behave the same way here. This is because items produced by an include with globs integrate the fixed dir part of the glob and the relative expanded part into the item name. So `"..\**"` will capture the file `Project.proj` as `..\path\to\file\Project.proj`. Since Excludes are treated as strings (sometimes :( ), the string `..\path\to\file\Project.proj` does not match the string `Project.proj` and therefore the Exclude does not apply.

To make the Exclude work here you either 
- change it to be the same as Include would have produced it (prepend fixed dir part and glob expanded recursive dir part): `<File Include="..\**" Exclude="..\ProjectFolder\Project.proj" />`
- exclude with a glob pattern: `<File Include="..\**" Exclude="**\Project.proj" />`

onovotny:
@cdmihai I appreciate the technical detail, but that's just making the user alter their code to an internal weirdness. I'm not saying this is a regression, but I think we're saying that a user would expect `Exclude="Project.proj"` to work because `Project.proj` is a valid subset of anything matching `..\**`. The fact that certain things are expanded sometimes is irrelevant to user expectations.

dsplaisted:
I agree with @onovotny that this is counter to user expectations.  Given the improvements we've made in this release, people are going to be using wildcards more, so they will be more likely to run into this issue than in previous versions.

This seems like it might be related to #1598.

cdmihai:
[...]

PR #1612, opened 2017-01-25
Glob matching is path aware

### LINKED ISSUE #1598: GetItemProvenance matches items outside the project cone with globs inside
GetItemProvenance is currently matching items from outside of the project cone with globs that are purely within the project cone.

```XML
<Project>
  <ItemGroup>
    <Compile Include="**/*.cs" /> <!-- Inside the project cone -->
    <Compile Include="../shared/Class1.cs"> <!-- Outside the project cone -->
      <Link>Class1.cs</Link>
    </Compile>
  </ItemGroup>
</Project>
```

``` CSharp
var provenance = project.GetItemProvenance("../shared.Class1.cs", "Compile");
Assert.Equal(1, provenance.Count); // False! it is 2.
Assert.Equal(Provenance.StringLiteral, provenance[0].Provenance); // False! Provenance.Glob, from the "**/*.cs" glob
```

cdmihai:
Fixes #1598. The problem is worse than item provenance alone. `<I Include="..\a.foo" Exclude="**\*.foo"/>` excludes `..\a.foo` in xplat, but didn't in dev14. This PR fixes issues of this type.

### DIFF
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -1167,14 +1167,9 @@ namespace Microsoft.Build.Shared
                 out matchFileExpression, out needsRecursion, out isLegalFileSpec,
                 getFileSystemEntries);
 
-            if (isLegalFileSpec)
-            {
-                regexFileMatch = new Regex(matchFileExpression, RegexOptions.IgnoreCase);
-            }
-            else
-            {
-                regexFileMatch = null;
-            }
+            regexFileMatch = isLegalFileSpec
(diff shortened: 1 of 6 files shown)


dsplaisted:
Does this fix #1576 too?  Either way, it would probably be good to add tests to cover the scenarios from #1576, as it seems related to these changes.

cdmihai:
@dotnet-bot test this please

cdmihai:
@dsplaisted It does not fix  #1576. I added a test to document the behaviour.

Issue #1124, opened 2016-09-28
Unify evaluation time and build time item operations


cdmihai:
Right now evaluation time and build time item operations support different operations and behave differently. It would be nice to unify them, so users have the same experience when using both types of items.

- [ ] [Add Update attribute for dynamic item tags](https://github.com/Microsoft/msbuild/issues/889)
- [ ] [static Update should support dynamic Update's feature of not having any attributes](https://github.com/Microsoft/msbuild/issues/955)
- [ ] [Migrate keepmetadata and removemetadata to valid evaluation time item operations](https://github.com/Microsoft/msbuild/issues/1123)
- [ ] [Migrate build time item operations to use the same matching code as eval time operations](https://github.com/Microsoft/msbuild/issues/1122)

cdmihai:
Please update description with other differences, if any.

Nirmal4G:
Really want this happen. It's painful, especially for the beginners to have to know the differences!!!

Issue #955, opened 2016-08-26
Enhance OM: static Update should support features from dynamic Update


cdmihai:
Dynamic items support the following:

```xml
<Target Name="Build">
   <ItemGroup>
      <Compile>
         <metadata>value</metadata>
      </Compile>
   </ItemGroup>
</Target>
```

This updates the metadata of all Compile items. Static update operations should support this too

PR #1620, opened 2017-01-27
Update attribute can only appear in static items

### LINKED ISSUE #1618: Update construct seems to update all items regardless of match when run in target
``` XML
<Project>
  <ItemGroup>
     <X Include="A;B;C" />
     <X Update="B" Metadata="Just on B" />
     <X Update="NonExistent" Metadata="Useless" />
  </ItemGroup>

  <Target Name="Build">
    <Warning Text="Before in-target update: %(X.Identity) %(X.Metadata)" />
    
   <ItemGroup>
      <X Update="A" Metadata="Just on A" />
      <X Update="NonExistent" Metadata="Huh" />
   </ItemGroup>
      
   <Warning Text="After in-target update: %(X.Identity) %(X.Metadata)" />
 </Target>
</Project>
```

Expected after target runs:
* A has Metadata= "Just on A" and B has Metadata "Just on B", C has no Metadata=""

Actual:
* They all have Metadata="Huh"

```
C:\temp\test.proj(10,5): warning : Before in-target update: A
C:\temp\test.proj(10,5): warning : Before in-target update: B Just on B
C:\temp\test.proj(10,5): warning : Before in-target update: C
C:\temp\test.proj(17,4): warning : After in-target update: A Huh
C:\temp\test.proj(17,4): warning : After in-target update: B Huh
C:\temp\test.proj(17,4): warning : After in-target update: C Huh
```

@AndyGerlicher @rainersigwald

cdmihai:
Resolves #1618

### DIFF
--- a/src/XMakeBuildEngine/Construction/ProjectItemElement.cs
+++ b/src/XMakeBuildEngine/Construction/ProjectItemElement.cs
@@ -434,6 +434,8 @@ namespace Microsoft.Build.Construction
         {
             ErrorUtilities.VerifyThrowInvalidOperation(parent.Parent is ProjectTargetElement || (Include.Length > 0 || Update.Length > 0 || Remove.Length > 0), "OM_ItemsOutsideTargetMustHaveIncludeOrUpdateOrRemove");
             ErrorUtilities.VerifyThrowInvalidOperation(parent.Parent is ProjectRootElement || parent.Parent is ProjectTargetElement || parent.Parent is ProjectWhenElement || parent.Parent is ProjectOtherwiseElement, "OM_CannotAcceptParent");
+
+            ErrorUtilities.VerifyThrowInvalidOperation(Update.Length == 0 || parent.Parent is ProjectRootElement, "OM_UpdateOnlyUnderProject");
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


nguerrera:
Can we take this to ship room? I think we need this in RTW

rainersigwald:
@nguerrera I don't think it meets the bar now, but we can bring it up in standup today.

nguerrera:
Isn't it a now or never language change? Does bar sill have that. This feels like silent bad code gen equivalent for msbuild. It would be breaking to make it work as I expect if not an error now.

nguerrera:
Also bragging to go from non error to error

cdmihai:
Ideally, the _correct_ thing to do is to error now for RTW, to increase the user friendliness and debug-ability. And then, post RTW, allow Update for targets too. Along with everything else in #1124.

Prior to dev15, msbuild did have an `Update` only for dynamic items (items in targets). It works like this: #955. This is the code path that hits if you put an Update inside targets. The Update gets ignored and the old behaviour kicks in, which is equivalent to `<Compile Update="@(Compile)">`.

More details [here](https://msdn.microsoft.com/en-us/library/ms171453.aspx#BKMK_AttributesWithinTargets).

nguerrera:
Agree on error now support later.

I'm aware of the existing target mechanism, but it is cumbersome. Could you guys look at my use case in the bug comment and let me know how to express it with just a condition?

rainersigwald:
I was just surprised this didn't happen. Did we decide not to make this an error? Did we just miss shiproom? I don't remember at all :(

cdmihai:
Same here, no idea what happened. My vague memory is ship-room miss-alignment. Should we bring the error back? Or just wait until we port Update to targets?

Issue #1474, opened 2016-12-13
P2P ref from UWP picks the win8 version instead of uap10.0 ver of cross-compiled lib


nguerrera:
Moved from https://github.com/dotnet/roslyn-project-system/issues/937 on behalf of @onovotny:

> A UWP project isn't getting the correct output of a cross-compiled project. Instead it's getting the win8 version.
>
>Check out
>https://github.com/onovotny/Zeroconf/tree/847ef218c7f94a1dc0b0dfa35a4c11d954c97f98
>
>The ZeroconfTest.UWP project has a P2P ref to Zeroconf. Zeroconf has lots of targets. The product in the dll shows what TFM it was built for. When looking at what's in the UWP\bin\x86\Debug directory, I see the win8 version not the uap10.0 version and things break.
>

See https://github.com/dotnet/roslyn-project-system/issues/937#issuecomment-266874274 for the patch that fixed the issue. The problem is that it is $(NugetTargetMoniker) and not $(TargetFrameworkMoniker) that will result in nuget picking the correct "best" target for UAP.

onovotny:
I'm seeing this now in a .NET 4.5 test project referencing a project that multi-targets to the following:
`netstandard1.2;netstandard1.3;net45;win81;wpa81;MonoAndroid70;Xamarin.iOS10`

It's using the `netstandard1.2` version (the first in the list?) instead of the `net45` ver.

onovotny:
@nguerrera I'm seeing this across the board for "regular" projects referencing a multi-targeted one:

https://github.com/onovotny/microsoft-authentication-library-for-dotnet/tree/1336a410808341c5ab09efa04b096b82f476ade7

In there, every project seems to be choosing the `netstandard1.2` version based on the path I see in the properties window when I select the reference. Those should all be choosing "closer" platform-specific versions.

nguerrera:
@onovotny That's different and bizarre. Can you log a separate SDK bug and share repro steps?

onovotny:
@nguerrera done: https://github.com/dotnet/sdk/issues/535

Sarabeth-Jaffe-Microsoft:
@nguerrera -- have you made any progress on this fix? This is currently targeted for RC3 release.

nguerrera:
I will re-validate the fix tomorrow and send a PR.

onovotny:
This is still happening with RC3 for me.

nguerrera:
Yes. It's not fixed yet. PR coming shortly.

Issue #1390, opened 2016-11-23
Provide alternate way to specify the ReferringTargetFramework passed to referenced projects in _GetProjectReferenceTargetFrameworkProperties


dsplaisted:
Right now, the `_GetProjectReferenceTargetFrameworkProperties` [passes the `TargetFrameworkMoniker`](https://github.com/Microsoft/msbuild/blob/fbba93c46c3603d4eef64cbfe5aaf86a8e6b2e88/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L1518) property value as the `ReferringTargetFramework` property to referenced projects.  A deployment-type project might not be a traditional .NET project, so it doesn't make sense for it to have a `TargetFrameworkMoniker`.  However, there should still be a way for such a project to specify a value for `ReferringTargetFramework`, so that the best TargetFramework can be used if the referenced project is multi-targeted.

This could be as simple as having a `ReferringTargetFrameworkForProjectReferences` property that defaults to the `TargetFrameworkMoniker` if not otherwise specified and which is passed as the `ReferringTargetFramework` to referenced projects.

rainersigwald:
This is fixed with #1621's introduction of `ReferringTargetFrameworkForProjectReferences`, right @nguerrera, @dsplaisted?

nguerrera:
Yup.

PR #1621, opened 2017-01-27
Improve P2P handling for cross-targeting

nguerrera:
1. Deal with UWP case where NugetTargetMoniker is needed instead of TargetFrameworkMoniker
2. Provide mechanism to override the moniker that is passed to reference
3. Provide mechanism to pick any target framework of reference
4. Provide mechanism to skip the best TFM computation for known single-targeting case

@AndyGerlicher @rainersigwald

**Customer scenario**

*Primary:* 
Referencing an SDK/CPS-based project from UWP project would behave incorrectly (pick wrong TFM of reference) or fail with incorrect incompatibility error.

Also:
* There was no supported way to override the moniker sent to refs.
* There was no supported way to pick a different TFM than the one selected by the SDK for a cross-targeted ref, but sometimes you want to test a different one than the one deemed best for the caller.
* There was no supported way to turn off the cross-targeting protocol if you know ref is single targeted (which can now be used to improve performance if needed).

**Bugs this fixes:**
#1474 
#1390
dotnet/sdk#707
dotnet/roslyn-project-system#477

**Risk**

Low.

**Performance impact**

Negligible. 

**Is this a regression from a previous update?**

No.

**Root cause analysis:**

We missed the special cases around monikers for UWP.

**How was the bug found?**

Customer reported

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1512,17 +1512,55 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   -->
   <Target Name="_GetProjectReferenceTargetFrameworkProperties"
           Outputs="%(_MSBuildProjectReferenceExistent.Identity)">
-    <!-- As a mitigation for https://github.com/Microsoft/msbuild/issues/1276
-       .vcxproj is treated specially to avoid the cost of double-evaluating
-       for easily-identifiable-as-not-cross-targeting C++ projects. -->
+    <!--
+      Honor SkipGetTargetFrameworkProperties=true metadata on project references
+      to mean that the project reference is known not to target multiple frameworks
+      and the mechanism here for selecting the best one can be skipped as an optimization.
+
+      We give this treatment to .vcxproj by default since no .vcxproj can target more
+      than one framework.
+   -->
+   <ItemGroup>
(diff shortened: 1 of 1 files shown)


nguerrera:
@srivatsn

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1512,17 +1512,55 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <!--
+      Honor SkipGetTargetFrameworkProperties=true metadata on project references
+      to mean that the project reference is known not to target multiple frameworks
+      and the mechanism here for selecting the best one can be skipped as an optimization.
+
+      We give this treatment to .vcxproj by default since no .vcxproj can target more
+      than one framework.
+   -->
+   <ItemGroup>
+      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and '%(Extension)' == '.vcxproj'">
```
AndyGerlicher: Should this be or vcxproj?

rainersigwald: Makes sense to be able to opt into the double-eval even for a vcxproj if you really want to.

nguerrera: Yes. Intentional.

AndyGerlicher: Read that wrong. Read that as explicit skip and vcxproj.

PR #1630, opened 2017-01-30
Unify reference to System.Threading.Tasks.Dataflow

jeffkl:
For full framework we're pulling this reference from the GAC and on .NET Core its coming from a NuGet package.  This leaves us with references to different versions which is not ideal.


jeffkl:
In my opinion, its only required before we start shipping facades.  I think we decided to not ship facades for RTW so this can probably wait for update 1...

jeffkl:
@dotnet-bot test this please

jeffkl:

PR #1632, opened 2017-01-30
Update version for vs15-rtw branch

jeffkl:
This limits public releases to only this branch (vs15-rtw).  I removed the prerelease tag from `version.json`.  The NuGet package version is also changed to be major.minor.build instead of major.minor.0-prelease-build.

Versions in "public release" mode (anything built out of a branch ending with `-rtw`:
``` C#
[assembly: AssemblyFileVersion("15.1.542.25220")]
[assembly: AssemblyInformationalVersion("15.1.542+g8462223709")]
```

```
Microsoft.Build.15.1.542.nupkg
```

Finally, I edited the official build definition to not statically set `PublicRelease=true` and instead defer to GitVersioning to make the determination.


build/NuGetPackages/CreateNuGetPackages.proj
```diff
@@ -72,7 +72,7 @@
     This target is used to customize the versions generated by NerdBank.GitVersioning.  The $(PrereleaseVersion) is read from version.json and
     then the git commit height and build revision is appended.
   -->
-    <PropertyGroup>
+    <PropertyGroup Condition=" '$(PublicRelease)' == 'false' ">
```
rainersigwald: As a rule I prefer `!= true` and `== true` rather than dual explicit comparisons (allowing the unset case to default into one bucket or the other). But if NB.GV guarantees that it's explicitly set then this is fine.

version.json
```diff
@@ -1,7 +1,6 @@
 {
   "version": "15.1-preview",
   "publicReleaseRefSpec": [
-    "^refs/heads/master$",
-    "^refs/heads/xplat$"
+    "^refs/heads/.*-rtw"
```
rainersigwald: Not sure this'll last forever (I assume at some point we'll have something like `vs15-update1` that's a release branch) but good enough for now.

dir.targets
```diff
@@ -154,5 +154,12 @@
 
   </Target>
 
+  <Target Name="SetPublicReleaseVersion" AfterTargets="GetBuildVersion">
+    <PropertyGroup Condition=" '$(PublicRelease)' == 'true' ">
+      <!-- Removes the prerelease tag from version when this is a public version but still has the git commit-->
+      <AssemblyInformationalVersion>$(BuildVersionSimple)$(SemVerBuildSuffix)</AssemblyInformationalVersion>
```
rainersigwald: Does NB.GV not do this automagically? That was my interpretation of the docs there . . .

jeffkl: This is stripping out the pre-release tag `-preview` which GitVersioning does not do.

Without this:
```C#
[assembly: AssemblyInformationalVersion("15.1.543-preview+gdd5e8bc3f8")]
```

As far as I can tell, `PublicRelease` is about removing the git commit from stuff like the NuGet package.  The pre-release tag is never removed from what I could tell.

version.json
jeffkl: I agree.  I really wanted to go off tags instead but we don't tag stuff until after we RTM.  I think we'll just need to come up with a naming scheme and add to this as we go.  Although update 1 might be `vs15-update1-rtw` which would match this pattern...

build/NuGetPackages/CreateNuGetPackages.proj
jeffkl: Fixed!

dir.targets
rainersigwald: Well, that's . . . unintuitive. I always want the git commit, and want the prerelease tag easily toggleable. So be it.

jeffkl: I agree.  I think when we redo our projects, we should remove the prerelease tag from GitVersioning and add it ourselves if its not a public release.   It only applies to NuGet package anyway, I don't think it should show up in any assembly versions anyway.

rainersigwald: Works for me.

AndyGerlicher:
@AArnott any feedback for us on this? Just wanted to run this by you quickly before we go with it for RTW.

dir.targets
AArnott: An interesting perspective here. If what you're building is a prerelease, why wouldn't you want that to show up in your AssemblyInformationalVersion? Wouldn't it be useful for you or a customer to see a DLL and associate it with -beta or -rc or recognize that since it's missing it must be at or near RTW?

build/NuGetPackages/CreateNuGetPackages.proj
```diff
@@ -85,5 +85,9 @@
       <!-- Override the NuGet package version provided by Nerdbank.GitVersioning -->
       <NuGetPackageVersion>$(MajorMinorVersion).$(BuildNumber)$(PrereleaseVersion)</NuGetPackageVersion>
```
AArnott: I'd love to learn more about what you're not satisfied with here in NB.GV.

build/NuGetPackages/CreateNuGetPackages.proj
```diff
@@ -85,5 +85,9 @@
...
     </PropertyGroup>
+
+    <PropertyGroup Condition=" '$(PublicRelease)' == 'true' ">
+      <NuGetPackageVersion>$(MajorMinorVersion).$(BuildNumber)</NuGetPackageVersion>
```
AArnott: This looks like it will strip out the prerelease tag from your version for public releases, even if it was supposed to be a prerelease. With this as it is, how will you create a 15.1-beta package when -beta keeps getting stripped off of your public releases?

version.json
AArnott: We could chat about branching patterns that work well for other repos I've seen, if you're interested.

jeffkl:
I'll try to sum up what we want and you can tell me if it makes sense.  It'd great to do an in person chat with you today if you have time. @rainersigwald or @AndyGerlicher correct me if you feel like the versioning should work differently from what I describe...

We do our active development in `master` (which at the moment is `xplat` but we treat it as master).  We try to work exclusively out of forks, even if its coming from us.  So we create a branch in our fork, push, and do a pull request.  Pull requests do a CI build which does not build NuGet packages.  We merge into `master` and continue the flow.

We manually schedule "official" builds at random whenever we want to get some functionality into users' hands.  This build is in VSTS and it builds our full product along with NuGet packages and publishes them to our own NuGet feed.  Sometimes we'll queue another build of the same sources because something will go wrong with the build;  so our NuGet package version has to take this into account.

Major and minor is set manually in `version.json`.  We'll increment this after each release.

When we have a stable NuGet package and after we've declared RTM, we push packages those packages from the daily MyGet feed to public NuGet.org.

I also looked into using Git tags but we'd have to schedule another build for the "public release" logic to take effect and that means another insertion into VS after we've already declared RTM.

So what we want is this:

## Local dev builds and CI builds:

**Assembly Versions**
``` C#
[assembly: AssemblyFileVersion("15.1.542.25220")]
[assembly: AssemblyInformationalVersion("15.1.542-prerelease+g8462223709")]
```

**NuGet package version**
```
15.1.0-prerelease-000542-00
```

The NuGet package version has build number set to zero and instead we increment a padded integer in the prerelease tag so that users can use wildcards when consuming our daily packages `15.1.0.prerelease-*` and the build identifier is always zero (`-00`).

## Official builds from a non-RTM branch (even `master`):

The only difference here is the build identifier is set at the end of the prerelease tag of the NuGet package version.  It starts as `-01` and for every build we do from the same sources it increments.  We don't use this all the time but we've needed it before if something went wrong during the build.

**NuGet package version**
```
15.1.0-prerelease-000542-01
```

## RTM builds

With my pull request this discussion is about, we're treating "public release" as work out of a branch that we're going to ship from.  We branched for Visual Studio 2017 RTM but we're still doing development from it.  When I branch it to add changes, its a non-RTM branch and after the merge if I schedule an official build then it is in an RTM branch (my branch is `foo` and it gets merged back into `vs15-rtm`).

The key difference here is we need the prerelease tag stripped because builds from that branch are not pre-releases.  Eventually one of the builds will be declared our official RTM build but its declared _after_ the build is done so we can't change the version of it.

**Assembly Versions**
The prerelease tag is removed.
``` C#
[assembly: AssemblyFileVersion("15.1.542.25220")]
[assembly: AssemblyInformationalVersion("15.1.542+g8462223709")]
```

**NuGet package version**
The prelease tag is removed and the build number is now part of the version.  If our build died and NuGet packages were published, we'd have to delete the packages from our NuGet feed manually before launching another build.
```
15.1.542
```

One thought was to remove the prerelease tag from the version in `version.json` and have something like this instead:

```json
{
   "version":"15.1",
   "tags":{
      "": [
         "^refs/heads/.*-rtm"
      ],
      "-rc$1": [
         "^refs/heads/.*-rc([1-9]?)"
      ],
      "-preview": [
         "^refs/heads/.*"
      ]
   }
}
```

This is conveying that branches ending in `-rtm` should have no prerelease tag.  Branches ending in `-rc1` should have a `-rc` prerelease tag with the corresponding number from the capture group.  And everything else should have a `-preview` tag.

This would let us branch and have our versions automatically follow the flow.  Let me know what you think or if you'd like to discuss in person.

AArnott:
@jeffkl thanks for the through docs. I've sent a meeting invite. A few questions I'll ask right away, but feel free to wait till our meeting to answer.

1. Who wants to consume your daily builds without documentation of which build they picked up? Using wildcards seems "wild" to me because your build can break a moment after it succeeded, without any changes on your side. It makes me wonder who could be developing anything worthwhile and be able to tolerate that kind of non-reproducibility? 
1. Which nuget package receivers (packages.config, project.json, PackageReference) even support wildcards nowadays? I thought they were deprecated? (and was happy to be under that impression).
1. Regarding skuttled builds and the need to produce a unique package version, I never run into that problem for a few reasons: 
    1. I publish packages at the end of a build, when everything else was successful, so the only reason a build could fail by the time the package version has to be unique is if publishing itself failed.
    1. I set the msbuild determinstic property to true, so if I need to rebuild a commit, I know it's identical to the last time I built that commit. So if one of several packages was published when the build failed, I can build again and publish the rest.
    1. In the rare event that all else fails, I *could* push an empty commit to bump up the 3rd integer on the package. But in years of doing this, across all my repos, I've only seen that done once, and it was due to a misunderstanding on the part of the person who pushed it -- it didn't actually need to happen.
1. How do you build nuget packages? Why don't you want to validate that that build succeeds as part of your PR as well? You certainly don't have to publish those packages, but I would build them both for validation and to keep my dev and PR builds as close to my CI/release builds as possible.

jeffkl:
Looking forward to it.  I'll respond to a few questions because they come with GitHub links or detail that makes sense when its written down.

>Who wants to consume your daily builds without documentation of which build they picked up?

We were asked to support it here: https://github.com/Microsoft/msbuild/issues/1156

>Which nuget package receivers (packages.config, project.json, PackageReference) even support wildcards nowadays?

I think its still supported in project.json and PackageReference;  although personally I would never use them either.

> How do you build nuget packages?

We build MSBuild targeting .NET 4.6 full framework and .NET Core.  At the moment, we have separate builds for this on 5 different machines for the best performance:

1. Full framework on Windows
2. .NET Core on Windows
3. .NET Core on Ubuntu 14
4. .NET Core on Ubuntu 16
5. .NET Core on Mac OS X

The NuGet packages have to contain assemblies for full framework and .NET Core and since our CI build only builds one, it can't create the NuGet package.  The official build builds both in two steps and then a third to create packages.  We're hoping the next version of MSBuild can ship one a portable assembly which could make that a lot easier.

Also we don't use GitVersioning when building on non-Windows because of https://github.com/AArnott/Nerdbank.GitVersioning/issues/56 but that sounds like it might be fixed?

Anyways, see you this afternoon!

jeffkl:
@Microsoft/msbuild-maintainers I've changed this PR to only remove the prerelease tag in version.json and update the NuGet package version to be `15.1.[height]`.  Feel free to merge when you want builds from this branch to be "RTM" candidates.

PR #1635, opened 2017-01-31
Catch FormatException coming from Convert.ChangeType.

### LINKED ISSUE #1634: New String.TrimEnd overload breaks MSBuild `CoerceArguments` on netcoreapp2.0
CoreFX has added a new String.TrimEnd overload with https://github.com/dotnet/corefx/pull/15334.

This is breaking MSBuild running on netcoreapp2.0 when building a project that contains logic like the following [taken from dotnet/sdk](https://github.com/dotnet/sdk/blob/9dc299b606ba76d1860700ed27016920afdd25e2/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.TargetFrameworkInference.targets#L51):

```
<_ShortFrameworkIdentifier>$(TargetFramework.TrimEnd('.0123456789'))</_ShortFrameworkIdentifier>
```

I get the following error:

```
Microsoft.NET.TargetFrameworkInference.targets(51,4): error MSB4184: The expression ""netcoreapp2.0".TrimEnd(.0123456789)" cannot be evaluated. String must be exactly one character long. 
```

I've debugged the error down to [this logic in MSBuild](https://github.com/Microsoft/msbuild/blob/7a49764a4cf5a5aef2e16f5900edf61bd295dcc1/src/XMakeBuildEngine/Evaluation/Expander.cs#L3607-L3611):

```
                        else
                        {
                            // change the type of the final unescaped string into the destination
                            coercedArguments[n] = Convert.ChangeType(args[n], parameters[n].ParameterType, CultureInfo.InvariantCulture);
                        }
                    }
                }
                catch (InvalidCastException)
                {
                    // The coercion failed therefore we return null
                    return null;
```

The issue is `Convert.ChangeType` is throwing a `FormatException`, [just like it is documented to do](https://msdn.microsoft.com/en-us/library/ms130977(v=vs.110).aspx).  But this code only catches `InvalidCastException`.  Instead, since trying to convert a string to a single char fails, this method overload should be skipped and keep searching until it finds the overload that accepts `char[]`.

This is blocking the CLI "2.0" work from taking new builds of the .NET Core runtime.

eerhardt:
String.TrimEnd has a new overload for a single char in .NET Core 2.0. However, when MSBuild encounters this new overload first, it throws a FormatException which is not caught - breaking the build.

The fix is to catch the FormatException just like an InvalidCastException and continue searching the rest of the reflection MethodInfos for a match.

Fix #1634

@rainersigwald @AndyGerlicher 

This change isn't necessary for "1.0", but I need to get an MSBuild version with this fix to unblock taking a new corefx in dotnet/cli/master.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/Expander.cs
+++ b/src/XMakeBuildEngine/Evaluation/Expander.cs
@@ -3608,9 +3608,13 @@ namespace Microsoft.Build.Evaluation
                         }
                     }
                 }
+                // The coercion failed therefore we return null
                 catch (InvalidCastException)
                 {
-                    // The coercion failed therefore we return null
+                    return null;
+                }
+                catch (FormatException)
+                {
                     return null;
                 }
 


AndyGerlicher:
Once @jeffkl 's change goes in to bump the version I'd say we can open xplat for Update 1 changes.

eerhardt:
Any idea when that is?  Like I said this isn't critical for "1.0", but it is blocking taking new "2.0" runtimes in cli/master.

rainersigwald:
@eerhardt Packages versioned `15.2.0-preview-000002-01` on our myget feed should have this change. Let us know if you notice anything else break since you're the first ~victim~ consumer of the post-version-change packages.

PR #1642, opened 2017-02-01
Remove default usage of NugetTargetMoniker as it breaks classic PCL builds

nguerrera:
@AndyGerlicher @rainersigwald @cdmihai This is option 3 from my mail. I will also submit a competing PR to roll back my change entirely (option 1)

This way preserves the escape hatches, but leaves no functional change from the prior builds.


nguerrera:
Closing in favor of #1644

PR #1644, opened 2017-02-01
Do not use NTM if it has more than one value

nguerrera:
**Customer scenario**
Building a classic portable class library with any project references fails

**Bugs this fixes:**
[VSO 377419](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems?id=377419&_a=edit)

**Workarounds, if any**
Set ReferringTargetFrameworkForProjectReferences=$(TargetFrameworkMoniker) in your project file.

**Risk**
Low

**Performance impact**
Low

**Is this a regression from a previous update?**
Yes, introduced very recently.

**Root cause analysis:**
We missed that $(NugetTargetMoniker) can contain multiple values and validation did not run through classic PCL scenarios.

**How was the bug found?**
Dogfooding. (The dogfooders affected have tested with this fix applied and confirmed it eliminates the problem.)

@AndyGerlicher @rainersigwald @cdmihai

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1546,10 +1546,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <!--
       Select the moniker to send to each project reference  if not already set. NugetTargetMoniker (NTM) is preferred by default over 
       TargetFrameworkMoniker (TFM) because it is required to disambiguate the UWP case where TFM is fixed at .NETCore,Version=v5.0 and 
-      has floating NTM=UAP,Version=vX.Y.Z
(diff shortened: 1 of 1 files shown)


nguerrera:
@rainersigwald Right.

nguerrera:
Just to be clear, there was no regression of PCL -> netstandard. It hasn't worked yet. This just undoes the unintended impact on PCL.

cdmihai:
The actual failure was that the NTM property contained illegal property names, like `.NETStandard,Version`, and not that it contained multiple values. (`Microsoft.Common.CurrentVersion.targets(1556,5): error MSB5016: The name ".NETStandard,Version" contains an invalid character "."`) With this change, isn't there a risk that an NTM / TFM with illegal characters could still flow around?

nguerrera:
We're setting ReferringTargetFramework=NTM.  So the property name is always ReferringTargetFramework. The fact that it had multiple values caused the second value to become a second property and that's when the dot was considered an invalid property name.

PR #1652, opened 2017-02-01
Update file version to 15.2.  Leave assembly version at 15.1

jeffkl:
``` C#
[assembly: AssemblyFileVersion("15.2.0.50417")]
[assembly: AssemblyInformationalVersion("15.2.0-preview+gf1c458f508")]
```


AndyGerlicher:
This looks good. Annoying to have 15.2 be Update 1. But anything we do to fix that would also be weird. We could go 15.1.1000+ for Update 1, then 15.2 for Update 2. But that's also odd. I guess it'll correct itself in Dev16 :). And this shouldn't matter at all to anyone as long as our assembly version stays the same.

And agreed when we want to fork for update. At this point I would consider xplat our Update 1 branch though.

PR #1659, opened 2017-02-02
Compile MSBuildConstants into Framework so it can use the AssemblyVersion

jeffkl:
Added a #DEFINE to Tasks so its copy of MSBuildConstants is in a different namespace

Updated usages of the classes to avoid build errors..


jeffkl:
@dotnet-bot test this please

PR #1665, opened 2017-02-02
Enable clean + build of crosstargeting projects

### LINKED ISSUE #1065: Cross-targeting build /t:Clean;Build should work
See https://github.com/Microsoft/msbuild/pull/1063#issuecomment-248770988

rainersigwald:
Fixes #1065 by duplicating `DispatchToInnerBuilds` to the crosstargeting
version of `Clean`. This enables Clean to run before Build (or the
arbitrary user target set via the `InnerTargets` property), reducing the
surprise of command line builds that use `/t:Clean;Build` rather than
`/t:Rebuild`.

This change also changes the cross-targeting definition of `Rebuild` to
match the standard `Clean;Build` rather than invoking `Rebuild` on each
inner build. That makes it easier to author outer-build extension logic,
fixing NuGet/Home#4475.

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
+++ b/src/XMakeTasks/Microsoft.Common.CrossTargeting.targets
@@ -19,6 +19,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(CustomBeforeMicrosoftCommonCrossTargetingTargets)" Condition="'$(CustomBeforeMicrosoftCommonCrossTargetingTargets)' != '' and Exists('$(CustomBeforeMicrosoftCommonCrossTargetingTargets)')"/>
 
+  <Target Name="_ComputeTargetFrameworkItems" Returns="@(InnerOutput)">
+    <ItemGroup>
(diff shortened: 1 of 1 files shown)


rainersigwald:
cc @rohit21agrawal @rrelyea

This is a better long-term fix than https://github.com/NuGet/NuGet.Client/pull/1167, but it affects all multi-targeting SDK-using projects, rather than just consumers of `GeneratePackageOnBuild` so it's a bit riskier. Let's discuss options in tooling standup today.

rainersigwald:
@nguerrera Thoughts on risk for this fix (which also does `clean;build` vs the targeted mitigation on the NuGet side?

rainersigwald:
Retargeted to xplat based on the decision today to not take this for RTW.

@dotnet-bot test this please

rohit21agrawal:
Are we taking the nuget workaround for rtm ?

rainersigwald:
My understanding was that we would do neither for RTW.

rrelyea:
for rtm, we've decided we will live without either. happy to discuss tomorrow.

rainersigwald:
@dotnet-bot test this please

PR #1682, opened 2017-02-08
Add net46 dependencies to msbuild nuspecs

### LINKED ISSUE #1681: Add net46 dependencies in our nuspec files
Without them, publishing an app for net46 which depends on our Runtime packages fails.

cdmihai:
This enables users to publish their msbuild depending app for net46 and get the closure of dependencies needed for msbuild.

I just copied over the net46 only dependencies from the project.json files over to the nuspec files.

This change should affect neither VS (which depends on msbuild via the vsix) nor CLI (who only publishes msbuild packages for .net core, not full framework).

Resolves #1681

### DIFF
--- a/build/NuGetPackages/Microsoft.Build.Runtime.nuspec
+++ b/build/NuGetPackages/Microsoft.Build.Runtime.nuspec
@@ -25,6 +25,7 @@
         <dependency id="Microsoft.Build.Framework" version="[$version$]" />
         <dependency id="Microsoft.Build.Tasks.Core" version="[$version$]" />
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
+        <dependency id="Microsoft.VisualStudio.Setup.Configuration.Interop" version="1.2.304-preview5" />
       </group>
     </dependencies>
     <contentFiles>
--- a/build/NuGetPackages/Microsoft.Build.Runtime.nuspec
+++ b/build/NuGetPackages/Microsoft.Build.Runtime.nuspec
@@ -25,7 +25,6 @@
         <dependency id="Microsoft.Build.Framework" version="[$version$]" />
         <dependency id="Microsoft.Build.Tasks.Core" version="[$version$]" />
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
-        <dependency id="Microsoft.VisualStudio.Setup.Configuration.Interop" version="1.2.304-preview5" />
       </group>
     </dependencies>
     <contentFiles>
--- a/build/NuGetPackages/Microsoft.Build.nuspec
+++ b/build/NuGetPackages/Microsoft.Build.nuspec
@@ -63,7 +63,50 @@
       </group>
       <group targetFramework=".NETFramework4.6">
         <dependency id="Microsoft.Build.Framework" version="[$version$]" />
-        <dependency id="System.Collections.Immutable" version="1.3.1" />
+        <dependency id="Microsoft.Win32.Primitives" version="4.0.1" />
+        <dependency id="System.AppContext" version="4.1.0" />
+        <dependency id="System.Collections" version="4.0.11" />
+        <dependency id="System.Collections.Concurrent" version="4.0.12" />
+        <dependency id="System.Collections.Immutable" version="1.2.0" />
+        <dependency id="System.Collections.NonGeneric" version="4.0.1" />
+        <dependency id="System.Console" version="4.0.0" />
+        <dependency id="System.Diagnostics.Contracts" version="4.0.1" />
+        <dependency id="System.Diagnostics.Debug" version="4.0.11" />
+        <dependency id="System.Diagnostics.FileVersionInfo" version="4.0.0" />
+        <dependency id="System.Diagnostics.Process" version="4.1.0" />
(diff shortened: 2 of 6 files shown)


rainersigwald:
Aren't the netstandard side of the nuget dependencies automagically updated? Instead of fixing manually can we fix the magic to work for net46 too?

build/NuGetPackages/Microsoft.Build.Runtime.nuspec
```diff
@@ -25,6 +25,7 @@
         <dependency id="Microsoft.Build.Framework" version="[$version$]" />
         <dependency id="Microsoft.Build.Tasks.Core" version="[$version$]" />
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
+        <dependency id="Microsoft.VisualStudio.Setup.Configuration.Interop" version="1.2.304-preview5" />
```
rainersigwald: Isn't this not a runtime dependency because it gets compiled into our assembly?

cdmihai: Yea, just checked with ildasm and they're there.

cdmihai:
@rainersigwald @jeffkl @onovotny

Found a knob for the nuspec reference generator to also update the net46 dependencies. However, it added quite a lot more dependencies than I manually did :) Do these look right? They kind of look right ...

onovotny:
AFAIK, there's no need for the nuspec reference generator anymore with the advent of netstandard.library so the tool has been discontinued. 

Instead, the dependency is `netstandard.library` 1.6.0 and that's it. no need for a tool anymore.

jeffkl:
@onovotny if we have dependencies outside of `NETStandard.Library` (which we do) then nuspec reference generator is still useful right?

onovotny:
If you have dependencies outside of them, you can add them to the nuspec directly.....there shouldn't be that many?

Alternatively, the better option, is to use the new pack targets instead of using a manual nuspec. `msbuild /t:pack` using an SDK style project. I show how you can do that in my blog: https://oren.codes/2017/01/04/multi-targeting-the-world-a-single-project-to-rule-them-all/

Key point is that you can multi-target net46, netcoreapp and whatever else in a single project and then use the pack targets to generate the right package with the right dependency set.

rainersigwald:
@onovotny That's our goal, but it'll take some time to get there from our current state. We're tracking the work in https://github.com/Microsoft/msbuild/milestone/14.

I still think the right thing to do now is the current change to use the nuspec generator in full framework builds.

onovotny:
you're of course welcome to use the nuspec generator but I'm really not sure what the actual benefits are. It was really about determining which packages you needed from your references when it was unclear due to trimming metapackages.

Now most packages are netstandard.library + 3-4 other specific package versions, so the tool isn't really needed. Easier to just do by hand.

cdmihai:
We use the tool to have it automatically bump up nuspec versions when we bump up project.jsons and to include any extra stuff we've missed. Migrating to `Microsoft.Net.Sdk` would solve our "keep nuspecs synced" issue, but until that happens this will have to do.

@onovotny: Coming back to this PR, does [this commit](https://github.com/Microsoft/msbuild/pull/1682/commits/8414fe9c9178288c25da2b022d2e817da996bd10) look OK to you? I changed the NRG to also update dependencies for `.NETFramework4.6` and it added quite a few extra stuff that I would have expected to be implicit and just resolve from the GAC. Are those just facades that type forward into mscorlib?

onovotny:
Looking at it, it seems ok, and for .NET 4.6 most of those are either empty packages or type forwarders. What I've found though is that .NET 4.6 users (or really anyone using packages.config) gets really cranky when they install a package that has 20 dependencies. That's one reason why in many of the packages I've created lately, I'll multi-target to net46 where it doesn't have the netstandard packages.

the refgen tool is probably putting more than you need in this case for net46 but it shouldn't break things...just irritate consumers of the package :/

build/NuGetPackages/Microsoft.Build.nuspec
```diff
@@ -63,6 +63,50 @@
...
+        <dependency id="System.Reflection.Metadata" version="1.3.0" />
+        <dependency id="System.Reflection.Primitives" version="4.0.1" />
+        <dependency id="System.Reflection.TypeExtensions" version="4.1.0" />
+        <dependency id="System.Resources.ResourceManager" version="4.0.1" />
+        <dependency id="System.Runtime" version="4.1.0" />
+        <dependency id="System.Runtime.Extensions" version="4.1.0" />
+        <dependency id="System.Runtime.Handles" version="4.0.1" />
+        <dependency id="System.Runtime.InteropServices" version="4.1.0" />
+        <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.0.0" />
+        <dependency id="System.Runtime.Loader" version="4.0.0" />
```
eerhardt: @cdmihai @jeffkl - 

This dependency is wrong/bad.  `System.Runtime.Loader` doesn't support `net46`.  It only supports `net462`.

rainersigwald: https://github.com/Microsoft/msbuild/issues/1725

PR #1691, opened 2017-02-10
Mark System.Threading.Task.Dataflow as Private=true

cdmihai:
To make our microbuild definition pass again for xplat.


cdmihai:
It fixes the microbuild because we now get Dataflow from a nuget package and not from the GAC, so the dll does not get copied to the `Output` directory anymore because it's marked `private=false`. The VSIX packager expects it to be in `Output` and thus fails.

PR #1696, opened 2017-02-14
Add support for .net SDK 4.7. 

XianghanWang:
Add .net sdk 4.7 and set it as default version. It will be released with RS2 SDK in dev15 update1.


src/Shared/FrameworkLocationHelper.cs
```diff
@@ -1157,10 +1163,13 @@ private class VisualStudioSpec
             /// </summary>
             public string GetDotNetFrameworkSdkRegistryKey(Version dotNetSdkVersion)
             {
-                string sdkVersionFolder = "4.6"; // Default for back-compat
+                string sdkVersionFolder = "4.7"; // Default for back-compat
```
jeffkl: Should the default still be 4.6?

XianghanWang: The method is used to generate registryPath. On machine with Dev15.1, the default sdk is 4.7 so very likely 4.7 is the only sdk installed on the machine.

ref/net46/Microsoft.Build.Utilities.Core/Microsoft.Build.Utilities.Core.cs
```diff
@@ -314,7 +314,8 @@ public enum TargetDotNetFrameworkVersion
         Version46 = 7,
         Version461 = 8,
         Version462 = 10,
-        VersionLatest = 10,
+        Version47 = 11,
+        VersionLatest = 11,
```
davkean: This is a breaking change - and has broken teams (the Roslyn team) the last time this occurred. This pattern should be stopped.

XianghanWang: @davkean, could you share us how Roslyn got broken last time and what is your recommendation on the change? Thanks.

AndyGerlicher: In what way is it breaking? We've been just following the pattern done in the past knowing this needed to be revisited at some point.

We could make VersionLatest a special enum that we always treat as latest? We can't really get rid of it though.

davkean: You're making a breaking change in an in-place update.

How this broke Roslyn last time, was that we installed VS 2015 Update 2 onto the build machine - not realizing that this would cause us to pick up a new value of the enum for VersionLatest. This would run on another machine without this value and we were broken.

davkean: Here is the change: https://github.com/dotnet/roslyn/commit/a1e93ec18958ae3106605c76b48cf60d5f83421c.

You can't change enum values.

XianghanWang: @davkean agree the VersionLatest in enum isn't a good design. But when people install VS 2017 Update1 on clean machine, they will only get .net SDK 4.7 and Roslyn will break again as it looks for 4.6. We need fix that as well. 

My proposal is adding another ToolLocationHelper method or property to return latest version from reg key. When it is checked in, we might need change Roslyn to use the method/property instead of hardcoded Version46

Issue #1675, opened 2017-02-07
Crash deserializing BuildEventArgs-derived type in logging node


rainersigwald:
[This was reported by an internal user via email]

Microsoft (R) Build Engine version 15.1.523.56541
This is part of VS 26119.0.d15rel

Getting a repro only takes 2 minutes if you have the C++ tools and Windows 8.1 SDK installed:

Open a "Developer Command Prompt for VS 2017 RC"
1) git clone https://github.com/lz4/lz4.git D:\LZ4\src
2) cd D:\LZ4
3) devenv /upgrade .\src\visual\vs2010\lz4.sln
4) msbuild /p:Configuration=Release;Platform=x64 .\src\visual\VS2010\lz4.sln /t:Rebuild /m /p:BuildInParallel=true

The last command yields a crash in msbuild.exe:

```
Unhandled Exception: System.ArgumentException: Object of type 'Microsoft.VisualStudio.CodeAnalysis.AnalysisResults.AnalysisMessage' cannot be converted to type 'Microsoft.VisualStudio.CodeAnalysis.AnalysisResults.AnalysisMessage'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
   at System.Reflection.RtFieldInfo.UnsafeSetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, CultureInfo culture)
   at System.Runtime.Serialization.FormatterServices.SerializationSetValue(MemberInfo fi, Object target, Object value)
   at System.Runtime.Serialization.ObjectManager.CompleteObject(ObjectHolder holder, Boolean bObjectFullyComplete)
   at System.Runtime.Serialization.ObjectManager.DoNewlyRegisteredObjectFixups(ObjectHolder holder)
   at System.Runtime.Serialization.ObjectManager.RegisterObject(Object obj, Int64 objectID, SerializationInfo info, Int64 idOfContainingObj, MemberInfo member, Int32[] arrayIndex)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.RegisterObject(Object obj, ParseRecord pr, ParseRecord objectPr, Boolean bIsString)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.ParseObjectEnd(ParseRecord pr)
[...]

PR #1699, opened 2017-02-14
Work around #1675 for C++ Code Analysis scenarios

rainersigwald:
This addresses the repro case for #1675, but not its root cause. By
providing a `codeBase` for the CA assemblies, we ensure that any running
MSBuild can find them when deserializing a log message.

### DIFF
--- a/src/XMakeCommandLine/app.amd64.config
+++ b/src/XMakeCommandLine/app.amd64.config
@@ -52,6 +52,20 @@
           <assemblyIdentity name="XamlBuildTask" culture="neutral" publicKeyToken="31bf3856ad364e35" />
           <bindingRedirect oldVersion="4.0.0.0" newVersion="15.0.0.0" />
         </dependentAssembly>
+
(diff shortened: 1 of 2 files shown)


rainersigwald:
@mavasani As discussed over email.

I was a bit confused about what assemblies to involve here. I first tried with just `CodeAnalysis` which seemed to work, but I observed `FxCopTask `load failures in the Fusion log. So I added that, at which point we crashed again due to failing to load `CodeAnalysis.Sdk`. So I went with all three. But I don't know if that's right.

mavasani:
@rainersigwald - that seems correct. FXCopTask assembly can be invoked from MSBuild and has a dependency on CodeAnalysis SDK.

rainersigwald:
One other thing--the plan for dev15 is that those assemblies are only ever serviced in-place in the VS folder, right? They won't ever be loaded from a NuGet package or something? This would break that.

mavasani:
Yes, these assemblies are not published via NuGet packages, they ship in the box.

PR #1728, opened 2017-02-21
Add support for .net SDK 4.7 - VersionLatestOnMachine

AndyGerlicher:
Will likely close PR #1696 in favor of this one. Wanted to get the idea out there for review.

* Add support for .net SDK 4.7. It will release with Dev15.1.
* Update VersionLatest to remain at 4.6.2 (latest at RTM).
* Add VersionLatestOnMachine to identify latest MSBuild knows about.
* Fix issue with fallback (StackOverflow on unit tests).

@davkean Given what we have to work with, how does this sound? VersionLatest would be fixed per major update (4.6.2 for Dev15). VersionLatestOnMachine would mean use the latest whatever version it is (that MSBuild knows about, still has to be updated per sdk release). (I realize this isn't implemented, I just wanted to get the idea out for review in case anyone can think of something obvious we're missing.)

### DIFF
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -1970,6 +1970,10 @@ namespace Microsoft.Build.Utilities
                 case TargetDotNetFrameworkVersion.Version47:
                     return FrameworkLocationHelper.dotNetFrameworkVersion47;
 
+                case TargetDotNetFrameworkVersion.VersionLatestOnMachine:
+                    // VersionLatestOnMachine is a special value to indicate the highest version we know about.
+                    return FrameworkLocationHelper.dotNetFrameworkVersion47;
+
                 default:
                     ErrorUtilities.ThrowArgument("ToolLocationHelper.UnsupportedFrameworkVersion", version);
                     return null;
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -83,11 +83,16 @@ namespace Microsoft.Build.Utilities
         /// </summary>
         Version462 = 10,
 
+        /// <summary>
+        /// version 4.7
+        /// </summary>
+        Version47 = 11,
+
         // keep this up to date, this should always point to the last entry
         /// <summary>
         /// the latest version available at the time of release
         /// </summary>
-        VersionLatest = Version462
+        VersionLatest = Version47
     }
 
     /// <summary>
@@ -1952,6 +1957,9 @@ namespace Microsoft.Build.Utilities
 
                 case TargetDotNetFrameworkVersion.Version462:
                     return FrameworkLocationHelper.dotNetFrameworkVersion462;
+
+                case TargetDotNetFrameworkVersion.Version47:
+                    return FrameworkLocationHelper.dotNetFrameworkVersion47;
 
                 default:
                     ErrorUtilities.ThrowArgument("ToolLocationHelper.UnsupportedFrameworkVersion", version);
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -88,11 +88,20 @@ namespace Microsoft.Build.Utilities
         /// </summary>
         Version47 = 11,
 
-        // keep this up to date, this should always point to the last entry
+        // Keep this up to date, this should always point to the last entry
         /// <summary>
-        /// the latest version available at the time of release
+        /// The latest version available at the time of major release. This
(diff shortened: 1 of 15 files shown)


AndyGerlicher:

rainersigwald:
Could be `VersionLatestKnownToToolset` or something, maybe? Then I'd feel better about the semantics of hardcoding it per MSBuild version.

AndyGerlicher:
Renamed to VersionLatestEx. Couldn't think of anything good talking with Jeff. Just saw your name, don't really like any of the options...

rainersigwald:
Oh man, I hate `Ex` even more than the others though :-/

AndyGerlicher:
And we settled on `Latest`

XianghanWang:
maybe something like below to get the actual sdk version on machine? I didn't run the code to verify. Just some idea.

ToolLocationHelper.cs:
        internal static TargetDotNetFrameworkVersion GetLatestDotNetFrameworkVersionOnMachine()
        {
            string version = FrameworkLocationHelper.GetLatestDotNetSdkInstalledOnMachine();
            string versionName = string.Format("Version{0}", version?.Replace(".", ""));
            TargetDotNetFrameworkVersion result;
            if(Enum.TryParse(versionName, out result))
            {
                return result;
            }
            // or should we throw exception if nothing found?
            return TargetDotNetFrameworkVersion.VersionLatest;
        }
FrameworkLocationHelper.cs:
        private const string dotNetFrameworkSdkRootPath = @"SOFTWARE\Microsoft\Microsoft SDKs\\NETFXSDK";
        internal static string GetLatestDotNetSdkInstalledOnMachine()
        {
            return FindSubKeys(dotNetFrameworkSdkRootPath ).OrderBy(o=>o).LastOrDefault();
        }

        private static string[] FindSubKeys(
            string registryBaseKeyName,
            RegistryView registryView = RegistryView.Default)
        {
            using (RegistryKey baseKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, registryView))
            using (RegistryKey subKey = baseKey.OpenSubKey(registryBaseKeyName))
            {
                return subKey?.GetSubKeyNames();
            }
        }

src/Utilities/ToolLocationHelper.cs
```diff
@@ -83,11 +83,29 @@ public enum TargetDotNetFrameworkVersion
...
+        /// </summary>
+        VersionLatest = Version462,
+
+        /// <summary>
+        /// Sentinel value for the latest version that we currently know about. Similar to
+        /// VersionLatest except the compiled value in the calling application will not need
+        /// to change for the update in MSBuild to be used.
+        /// </summary>
+        /// <remarks>
+        /// This value was introduced in Visual Studio 15.1. It is incompatible with previous
```
cdmihai: nit: previous versions

src/XMakeTasks/GetFrameworkSDKPath.cs
```diff
@@ -38,15 +38,15 @@ public string Path
...
+                    s_path = ToolLocationHelper.GetPathToDotNetFrameworkSdk(TargetDotNetFrameworkVersion.Latest, VisualStudioVersion.VersionLatest);
 
                     if (String.IsNullOrEmpty(s_path))
                     {
                         Log.LogMessageFromResources(
                             MessageImportance.High,
                             "GetFrameworkSdkPath.CouldNotFindSDK",
-                            ToolLocationHelper.GetDotNetFrameworkSdkInstallKeyValue(TargetDotNetFrameworkVersion.VersionLatest, VisualStudioVersion.VersionLatest),
-                            ToolLocationHelper.GetDotNetFrameworkSdkRootRegistryKey(TargetDotNetFrameworkVersion.VersionLatest, VisualStudioVersion.VersionLatest)
+                            ToolLocationHelper.GetDotNetFrameworkSdkInstallKeyValue(TargetDotNetFrameworkVersion.Latest, VisualStudioVersion.VersionLatest),
```
cdmihai: >GetDotNetFrameworkSdkInstallKeyValue [](start = 47, length = 36)

Update the GetDotNetFrameworkSdkInstallKeyValue tests to assert it also works with Latest

XianghanWang:
LGTM

AndyGerlicher:
Just FYI I tried to make it query the registry and it seemed to mostly work, but I'm a little worried about some edge cases. Specifically around scenarios where we build a `Version` object and use that for comparison (and a key to a Dictionary). I don't think it's worth doing for this release, but I like the idea of changing `Latest` to mean query the registry for the latest, maybe Dev16.

PR #1732, opened 2017-02-21
Enable XmlPeek and XmlPoke tasks for .NET Core. 

### LINKED ISSUE #1731: XmlPoke task not available in .NET Core
**Repro:**

```xml
<Project>
   <Target Name="Build">
      <XmlPoke />
   </Target>
</Project>
```
dotnet msbuild /t:Build

**Error**

> poke.proj(3,7): error MSB4062: The "Microsoft.Build.Tasks.XmlPoke" task could not be loaded from the assembly Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a.  Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.

**Expected**
Same behavior as MSBuild.exe (full framework)

**Details**
dotnet.exe version: 1.0.0-rc4-004842
MSBuild version:  15.1.548.43366

cc @rainersigwald @jeffkl

jeffkl:
Had to add some additional dependencies and fix up some tests to use `netstandard` APIs.

Closes #1731

### DIFF
--- a/src/Tasks/XmlPoke.cs
+++ b/src/Tasks/XmlPoke.cs
@@ -223,7 +223,14 @@ namespace Microsoft.Build.Tasks
 
             if (iter.Count > 0)
             {
+#if RUNTIME_TYPE_NETCORE
+                using (Stream stream = File.Create(_xmlInputPath.ItemSpec))
+                {
+                    xmlDoc.Save(stream);
+                }
+#else
                 xmlDoc.Save(_xmlInputPath.ItemSpec);
+#endif
             }
 
             return true;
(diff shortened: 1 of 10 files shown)


src/Tasks/XmlPoke.cs
```diff
@@ -223,7 +223,14 @@ public override bool Execute()
 
             if (iter.Count > 0)
             {
+#if RUNTIME_TYPE_NETCORE
```
rainersigwald: Eww, this is unfortunate. And normally I'd say "define a specific feature flag instead!" but hopefully we can reunify this back in the .NETStandard 2.0 timeframe.

jeffkl: Yeah its weird because they're extension methods in .NET Core so I wasn't sure what the feature should be called?  So I just used `RUNTIME_TYPE_NETCORE` to ensure that older behavior is kept...

jeffkl:
@AndyGerlicher and @rainersigwald This is now ready for final review.  The unit tests had to be re-written.

PR #1733, opened 2017-02-21
Fix GetLastWriteFileUtcTime function for a directories

mfilippov:
If some additional item in GenerateResource Task will be a director ResourceCreateFile in GetLastWriteFileUtcTime return invalid handle and then resource item would be marked as dirty. I add FILE_FLAG_BACKUP_SEMANTICS file flag for avoiding it problem.

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -900,6 +900,10 @@ namespace Microsoft.Build.Shared
             {
                 fileModifiedTime = File.GetLastWriteTimeUtc(fullPath);
             }
+            else if (Directory.Exists(fullPath))
+            {
+                fileModifiedTime = Directory.GetLastWriteTimeUtc(fullPath);
+            }
 
             return fileModifiedTime;
         }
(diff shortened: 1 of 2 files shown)


rainersigwald:
What causes a directory to be passed in there? I didn't quite catch that.

Code change looks good, just want to make sure I understand the scenario.

mfilippov:
We have an error in custom generated targets. We use MSBuildThisFileDirectory instead of MSBuildThisFile in MSBuildAllProjects property. But it is a bug was undiscoverable :) I think my patch also fixed different behaviour on Unix and Windows.

AndyGerlicher:
Thanks for the contribution! This looks good in general, however given that it's pretty subtle and `GetLastWriteFileUtcTime` is called in many places, I would like to see a unit test added that covers the scenario that was broken previously. That seem doable? Then I'm good taking the change. Thanks!

mfilippov:
@AndyGerlicher Ok, I try to do it.

mfilippov:
@AndyGerlicher I'm not sure about this PR because I found GetLastWriteDirectoryUtcTime method special for directories. May be better way it is patch GenerateResource task?

AndyGerlicher:
That sounds reasonable to me (patching GenerateResource task). Less impact since this method is used in so many places.

AndyGerlicher:
@mfilippov did you look into another approach? As is with the test I don't object to taking this change. If you still think patching GenerateResource would be a better approach please close this PR. Thanks!

mfilippov:
Ok, I not have time for that yet.

PR #1747, opened 2017-02-24
Introduce a new BinaryLogger logger.

KirillOsenkov:
/binaryLogger[:output.binlog]

Serializes all build events to a compressed binary file. By default the file is in the current directory and named "msbuild.binlog". The binary log is a detailed description of the build process that can later be used to reconstruct text logs and used by other analysis tools. A binary log is usually 10-20x smaller than the most detailed text diagnostic-level log, but it contains more information. (Short form: /bl or /binlog)

Examples:
  /bl
  bl:output.binlog
  binlog:..\..\custom.binlog
  binaryLogger

For now this commit only includes the logger (writer), but the reader is disconnected. In subsequent PRs we will add an option to playback the binary log into any other set of existing loggers.

Support for the new format will be added to MSBuild Structured Log Viewer so that you can view the .binlog files.

BuildEventArgsWriter and BuildEventArgsReader are general purpose custom binary serializers for BuildEventArgs objects.

BinaryLogReplayEventSource is a way to playback events from the log to subscribers of IEventSource.

A set of unit-tests is to follow in a separate PR.

### DIFF
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -1350,6 +1350,29 @@ namespace Microsoft.Build.Execution
 }
 namespace Microsoft.Build.Logging
 {
+    public partial class BinaryLogger : Microsoft.Build.Framework.ILogger
+    {
+        public BinaryLogger() { }
+        public string Parameters { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public Microsoft.Build.Framework.LoggerVerbosity Verbosity { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public void Initialize(Microsoft.Build.Framework.IEventSource eventSource) { }
+        public void Shutdown() { }
+    }
+    public partial class BinaryLogReplayEventSource : Microsoft.Build.Logging.EventArgsDispatcher
+    {
+        public BinaryLogReplayEventSource() { }
+        public void Replay(string sourceFilePath) { }
+    }
+    public partial class BuildEventArgsReader
+    {
+        public BuildEventArgsReader(System.IO.BinaryReader binaryReader) { }
+        public Microsoft.Build.Framework.BuildEventArgs Read() { throw null; }
+    }
+    public partial class BuildEventArgsWriter
+    {
+        public BuildEventArgsWriter(System.IO.BinaryWriter binaryWriter) { }
+        public void Write(Microsoft.Build.Framework.BuildEventArgs e) { }
+    }
     public delegate void ColorResetter();
     public delegate void ColorSetter(System.ConsoleColor color);
     public partial class ConfigurableForwardingLogger : Microsoft.Build.Framework.IForwardingLogger, Microsoft.Build.Framework.ILogger, Microsoft.Build.Framework.INodeLogger
@@ -1401,6 +1424,25 @@ namespace Microsoft.Build.Logging
         public void Initialize(Microsoft.Build.Framework.IEventSource eventSource) { }
         public void Initialize(Microsoft.Build.Framework.IEventSource eventSource, int nodeCount) { }
         public void Shutdown() { }
+    }
+    public partial class EventArgsDispatcher : Microsoft.Build.Framework.IEventSource
+    {
+        public EventArgsDispatcher() { }
+        public event Microsoft.Build.Framework.AnyEventHandler AnyEventRaised { add { } remove { } }
+        public event Microsoft.Build.Framework.BuildFinishedEventHandler BuildFinished { add { } remove { } }
+        public event Microsoft.Build.Framework.BuildStartedEventHandler BuildStarted { add { } remove { } }
+        public event Microsoft.Build.Framework.CustomBuildEventHandler CustomEventRaised { add { } remove { } }
+        public event Microsoft.Build.Framework.BuildErrorEventHandler ErrorRaised { add { } remove { } }
+        public event Microsoft.Build.Framework.BuildMessageEventHandler MessageRaised { add { } remove { } }
+        public event Microsoft.Build.Framework.ProjectFinishedEventHandler ProjectFinished { add { } remove { } }
+        public event Microsoft.Build.Framework.ProjectStartedEventHandler ProjectStarted { add { } remove { } }
+        public event Microsoft.Build.Framework.BuildStatusEventHandler StatusEventRaised { add { } remove { } }
+        public event Microsoft.Build.Framework.TargetFinishedEventHandler TargetFinished { add { } remove { } }
+        public event Microsoft.Build.Framework.TargetStartedEventHandler TargetStarted { add { } remove { } }
+        public event Microsoft.Build.Framework.TaskFinishedEventHandler TaskFinished { add { } remove { } }
+        public event Microsoft.Build.Framework.TaskStartedEventHandler TaskStarted { add { } remove { } }
+        public event Microsoft.Build.Framework.BuildWarningEventHandler WarningRaised { add { } remove { } }
+        public void Dispatch(Microsoft.Build.Framework.BuildEventArgs buildEvent) { }
     }
     public partial class FileLogger : Microsoft.Build.Logging.ConsoleLogger
     {
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -1350,28 +1350,13 @@ namespace Microsoft.Build.Execution
 }
 namespace Microsoft.Build.Logging
 {
-    public partial class BinaryLogger : Microsoft.Build.Framework.ILogger
+    public sealed partial class BinaryLogger : Microsoft.Build.Framework.ILogger
     {
         public BinaryLogger() { }
         public string Parameters { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.Build.Framework.LoggerVerbosity Verbosity { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public void Initialize(Microsoft.Build.Framework.IEventSource eventSource) { }
         public void Shutdown() { }
-    }
-    public partial class BinaryLogReplayEventSource : Microsoft.Build.Logging.EventArgsDispatcher
-    {
-        public BinaryLogReplayEventSource() { }
-        public void Replay(string sourceFilePath) { }
-    }
-    public partial class BuildEventArgsReader
-    {
-        public BuildEventArgsReader(System.IO.BinaryReader binaryReader) { }
-        public Microsoft.Build.Framework.BuildEventArgs Read() { throw null; }
-    }
-    public partial class BuildEventArgsWriter
-    {
-        public BuildEventArgsWriter(System.IO.BinaryWriter binaryWriter) { }
-        public void Write(Microsoft.Build.Framework.BuildEventArgs e) { }
     }
     public delegate void ColorResetter();
     public delegate void ColorSetter(System.ConsoleColor color);
@@ -1424,25 +1409,6 @@ namespace Microsoft.Build.Logging
         public void Initialize(Microsoft.Build.Framework.IEventSource eventSource) { }
         public void Initialize(Microsoft.Build.Framework.IEventSource eventSource, int nodeCount) { }
         public void Shutdown() { }
-    }
-    public partial class EventArgsDispatcher : Microsoft.Build.Framework.IEventSource
-    {
-        public EventArgsDispatcher() { }
-        public event Microsoft.Build.Framework.AnyEventHandler AnyEventRaised { add { } remove { } }
-        public event Microsoft.Build.Framework.BuildFinishedEventHandler BuildFinished { add { } remove { } }
-        public event Microsoft.Build.Framework.BuildStartedEventHandler BuildStarted { add { } remove { } }
-        public event Microsoft.Build.Framework.CustomBuildEventHandler CustomEventRaised { add { } remove { } }
-        public event Microsoft.Build.Framework.BuildErrorEventHandler ErrorRaised { add { } remove { } }
-        public event Microsoft.Build.Framework.BuildMessageEventHandler MessageRaised { add { } remove { } }
-        public event Microsoft.Build.Framework.ProjectFinishedEventHandler ProjectFinished { add { } remove { } }
-        public event Microsoft.Build.Framework.ProjectStartedEventHandler ProjectStarted { add { } remove { } }
-        public event Microsoft.Build.Framework.BuildStatusEventHandler StatusEventRaised { add { } remove { } }
-        public event Microsoft.Build.Framework.TargetFinishedEventHandler TargetFinished { add { } remove { } }
-        public event Microsoft.Build.Framework.TargetStartedEventHandler TargetStarted { add { } remove { } }
-        public event Microsoft.Build.Framework.TaskFinishedEventHandler TaskFinished { add { } remove { } }
-        public event Microsoft.Build.Framework.TaskStartedEventHandler TaskStarted { add { } remove { } }
-        public event Microsoft.Build.Framework.BuildWarningEventHandler WarningRaised { add { } remove { } }
-        public void Dispatch(Microsoft.Build.Framework.BuildEventArgs buildEvent) { }
     }
     public partial class FileLogger : Microsoft.Build.Logging.ConsoleLogger
     {
--- a/ref/netstandard1.3/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/netstandard1.3/Microsoft.Build/Microsoft.Build.cs
@@ -1327,6 +1327,29 @@ namespace Microsoft.Build.Execution
 }
 namespace Microsoft.Build.Logging
 {
+    public partial class BinaryLogger : Microsoft.Build.Framework.ILogger
+    {
+        public BinaryLogger() { }
+        public string Parameters { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public Microsoft.Build.Framework.LoggerVerbosity Verbosity { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public void Initialize(Microsoft.Build.Framework.IEventSource eventSource) { }
+        public void Shutdown() { }
+    }
+    public partial class BinaryLogReplayEventSource : Microsoft.Build.Logging.EventArgsDispatcher
+    {
+        public BinaryLogReplayEventSource() { }
+        public void Replay(string sourceFilePath) { }
+    }
+    public partial class BuildEventArgsReader
+    {
+        public BuildEventArgsReader(System.IO.BinaryReader binaryReader) { }
+        public Microsoft.Build.Framework.BuildEventArgs Read() { throw null; }
+    }
+    public partial class BuildEventArgsWriter
+    {
+        public BuildEventArgsWriter(System.IO.BinaryWriter binaryWriter) { }
+        public void Write(Microsoft.Build.Framework.BuildEventArgs e) { }
+    }
     public delegate void ColorResetter();
     public delegate void ColorSetter(System.ConsoleColor color);
     public partial class ConfigurableForwardingLogger : Microsoft.Build.Framework.IForwardingLogger, Microsoft.Build.Framework.ILogger, Microsoft.Build.Framework.INodeLogger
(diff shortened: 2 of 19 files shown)


AndyGerlicher:
Looks like you're not far off on the .NET Core build. `Close()` isn't supported on the stream (but you can use `Dispose()`), and `Delegate.CreateDelegate()` isn't supported until we move to [.NET Standard 2.0](https://apisof.net/catalog/System.Delegate.CreateDelegate(Type,MethodInfo)). So have to find another way to implement that unfortunately.

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    BuildEventArgs instance = null;
+
+                    try
+                    {
+                        instance = reader.Read();
+                    }
+                    catch (Exception ex)
+                    {
+                        var text = $"Exception while reading log file:{Environment.NewLine}{ex.ToString()}";
```
jeffkl: Add a new string to our Strings.resx

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -0,0 +1,136 @@
...
+        private BinaryWriter binaryWriter;
+        private BuildEventArgsWriter eventArgsWriter;
+
+        private string FilePath { get; set; }
+
+        /// <summary>
+        /// The binary logger Verbosity is always maximum (Diagnostic). It tries to capture as much
+        /// information as possible.
+        /// </summary>
+        public LoggerVerbosity Verbosity { get; set; } = LoggerVerbosity.Diagnostic;
```
jeffkl: You might want to throw an exception in the setter if it must always be `Diagnostic`

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// The only supported parameter is the output log file path (e.g. "msbuild.binlog") 
+        /// </summary>
+        public string Parameters { get; set; }
+
+        /// <summary>
+        /// Initializes the logger by subscribing to events of IEventSource
+        /// </summary>
+        public void Initialize(IEventSource eventSource)
+        {
+            Environment.SetEnvironmentVariable("MSBUILDTARGETOUTPUTLOGGING", "true");
```
jeffkl: Is there another way to turn this on?  Changing my environment seems bad but I'm not sure if there's another way.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <summary>
+        /// Closes the underlying file stream.
+        /// </summary>
+        public void Shutdown()
+        {
+            if (stream != null)
+            {
+                // It's hard to determine whether we're at the end of decoding GZipStream
+                // so add an explicit 0 at the end to signify end of file
+                stream.WriteByte(0);
```
jeffkl: Can you use the value from your enum so you don't need this comment?

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -0,0 +1,136 @@
...
+                string helpKeyword;
+                string message = ResourceUtilities.FormatResourceString(out errorCode, out helpKeyword, "InvalidFileLoggerFile", FilePath, e.Message);
+                throw new LoggerException(message, e, errorCode, helpKeyword);
+            }
+
+            stream = new GZipStream(stream, CompressionLevel.Optimal);
+            binaryWriter = new BinaryWriter(stream);
+            eventArgsWriter = new BuildEventArgsWriter(binaryWriter);
+
+            binaryWriter.Write((byte)FileFormatVersion);
```
jeffkl: I would store your version as a byte in the first place, otherwise I could set it to something too big and it would break this logger right?

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /// <summary>
+        /// Processes the parameters given to the logger from MSBuild.
+        /// </summary>
+        /// <exception cref="LoggerException">
+        /// </exception>
+        private void ProcessParameters()
+        {
+            const string invalidParamSpecificationMessage = @"Need to specify a valid log file name, such as msbuild.binlog";
```
jeffkl: Add a new string please

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            string[] parameters = Parameters.Split(';');
+
+            if (parameters.Length != 1)
+            {
+                throw new LoggerException(invalidParamSpecificationMessage);
+            }
+
+            FilePath = parameters[0].TrimStart('"').TrimEnd('"');
```
jeffkl: We _might_ have a helper to do this, @rainersigwald might know

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -0,0 +1,606 @@
...
+
+        private void WriteMessage(LazyFormattedBuildEventArgs e)
+        {
+            string message = buildEventArgsMessageField.GetValue(e) as string;
+
+            var arguments = (object[])lazyFormattedArgumentsField.GetValue(e);
+            if (arguments != null && arguments.Length > 0)
+            {
+                // this is bad for performance but only used infrequently (when a message has arguments and hasn't been formatted previously)
+                message = (string)formatStringMethodInfo.Invoke(null, new object[] { CultureInfo.CurrentCulture, message, arguments });
```
jeffkl: Try this without reflection like we talked about.

src/MSBuild/CommandLineSwitches.cs
```diff
@@ -281,8 +282,9 @@ bool emptyParametersAllowed
...
-            new ParameterizedSwitchInfo(  new string[] { "warnasmessage", "nowarn" },           ParameterizedSwitch.WarningsAsMessages,         null,                           true,           "MissingWarnAsMessageParameterError",  true,   false ),
+            new ParameterizedSwitchInfo(  new string[] { "warnaserror", "err" },                ParameterizedSwitch.WarningsAsErrors,           null,                           true,           null,                                  true,   true   ),
+            new ParameterizedSwitchInfo(  new string[] { "warnasmessage", "nowarn" },           ParameterizedSwitch.WarningsAsMessages,         null,                           true,           "MissingWarnAsMessageParameterError",  true,   false  ),
+            new ParameterizedSwitchInfo(  new string[] { "binarylogger", "binlog", "bl" },      ParameterizedSwitch.BinaryLogger,               null,                           false,          null,                                  true,   false  ),
```
jeffkl: I don't think we have any other args that have 3 forms.  Its usually full and short.  So I would get rid of `binlog` 😄

src/MSBuild/XMake.cs
```diff
@@ -1454,6 +1455,14 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 int numberOfCpus = Environment.ProcessorCount;
                                 switchParameters = ":" + numberOfCpus;
                             }
+                            else if (String.Equals(switchName, "bl", StringComparison.OrdinalIgnoreCase) ||
+                                String.Equals(switchName, "binlog", StringComparison.OrdinalIgnoreCase) ||
```
jeffkl: Remove if you agree

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Sure. But nobody reads Verbosity and I'm the only person setting it, so don't think it's worth it.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: They read an environment variable so I didn't find another way. Good thing that this is per-process and doesn't impact machine state.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: We can later switch it to the same algorithm used by Write7BitEncodedInt (if the higher order bit is set, keep reading), so it's future proof.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Good idea, will do that.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Technically it's not even needed since the arg parser does it already. This is just for the case someone instantiates a logger outside of MSBuild.

src/MSBuild/CommandLineSwitches.cs
KirillOsenkov: I thought so too, but it's too compelling that the extension name is the same as the switch. Image msbuild.exe /binlog:foo.binlog

I thought it'd make for good mnemonics which would warrant a precedent. But I was on the fence myself.

jeffkl: Okay I would get rid of `bl` then and keep `binlog`...

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        /// <param name="sourceFilePath">The full file path of the binary log file</param>
+        public void Replay(string sourceFilePath)
+        {
+            using (var stream = new FileStream(sourceFilePath, FileMode.Open))
+            {
+                var gzipStream = new GZipStream(stream, CompressionMode.Decompress, leaveOpen: true);
+                var binaryReader = new BinaryReader(gzipStream);
+
+                byte fileFormatVersion = binaryReader.ReadByte();
```
cdmihai: Eventually, it would be nice to error here if the file formats do not match. Or alternatively, keep all the readers/writers for each file version and delegate to the correct one. This might be a common case. I suspect people would want to switch their CI builds to log the binary format rather than the text one, and then you'd have cases where people are assigned to investigate some old failed build. But this can be done whenever fileFormatVersion=2 comes along.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
cdmihai: This line shows that out of proc nodes also need to know about this setting: [https://github.com/Microsoft/msbuild/blob/xplat/src/Build/BackEnd/Components/Communications/LogMessagePacket.cs#L63](https://github.com/Microsoft/msbuild/blob/xplat/src/Build/BackEnd/Components/Communications/LogMessagePacket.cs#L63)
To fix this to not use environment variables, we'd have to come up with a way of propagating this information to the nodes. Probably by adding a new property to either BuildParameters or NodeConfiguration.

Another smell is that the environment var is read in two different places in the msbuild codebase. It could be unified and put into LoggingService.


---
In reply to: [103011070](https://github.com/Microsoft/msbuild/pull/1747#discussion_r103011070) [](ancestors = 103011070)

src/MSBuild/CommandLineSwitches.cs
KirillOsenkov: bl is short :) If you want to get rid of one I'd get rid of binlog. Will do that.

src/MSBuild/XMake.cs
KirillOsenkov: Let's discuss on Monday I'm curious what others think.

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
KirillOsenkov: Yes, when we read the format we should be able to handle all previous formats.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
KirillOsenkov: Yup, this is gone now.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Added a new string resource.

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
KirillOsenkov: Fixed.

src/MSBuild/XMake.cs
KirillOsenkov: Removed /binlog.

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
AndyGerlicher: Yeah I agree keep all the readers (and potentially writers) around and delegate to the right one. Current versions should be able to read all the old ones (when possible). In this case though we should throw if the version is greater than what we know about since there's no way it can read future versions.

ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -1402,6 +1425,25 @@ public partial class DistributedFileLogger : Microsoft.Build.Framework.IForwardi
         public void Initialize(Microsoft.Build.Framework.IEventSource eventSource, int nodeCount) { }
         public void Shutdown() { }
     }
+    public partial class EventArgsDispatcher : Microsoft.Build.Framework.IEventSource
```
AndyGerlicher: Is there a need to have these public? Seems like something to avoid when possible. Even the `BinaryLogger` class, although the `FileLogger` is public so maybe there's a reason?

AndyGerlicher:
Looks like a great start! My biggest worry here is that the log format is extremely terse. Obviously this is good for this purpose, but very difficult to debug. You did mention a second PR for unit tests, so hoping you have some dedicated to reader/writer specifically.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -21,10 +21,15 @@ public class BuildEventArgsWriter
...
-                typeof(LazyFormattedBuildEventArgs).GetMethod("FormatString", BindingFlags.Static | BindingFlags.NonPublic));
+
+        // TODO: Delegate.CreateDelegate will only become available on .NET Standard 2.0, for now just use a MethodInfo
+        //private static Func<CultureInfo, string, object[], string> formatStringDelegate =
+        //    (Func<CultureInfo, string, object[], string>)Delegate.CreateDelegate(
+        //        typeof(Func<CultureInfo, string, object[], string>),
+        //        typeof(LazyFormattedBuildEventArgs).GetMethod("FormatString", BindingFlags.Static | BindingFlags.NonPublic));
+
+        private static MethodInfo formatStringMethodInfo =
+            typeof(LazyFormattedBuildEventArgs).GetMethod("FormatString", BindingFlags.Static | BindingFlags.NonPublic);
```
nguerrera: You can use `MethodInfo.CreateDelegate` https://msdn.microsoft.com/en-us/library/hh194376(v=vs.110).aspx

jeffkl:

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
AndyGerlicher: That's not supported till .NET Standard 2.0 :(

KirillOsenkov: Thanks, good to know! But I just got rid of reflection entirely here.

nguerrera: > That's not supported till .NET Standard 2.0 :(

Just FYI, since it's all moot for this change, but MethodInfo.CreateDelegate has been supported replacement for Delegate.CreateDelegate since Windows 8, definitely pre .NET Standard 2.0.  There had been an effort to prevent reflection from being available in every platform, so Delegate could note depend on reflection, but the reverse was fine.

ref/net46/Microsoft.Build/Microsoft.Build.cs
KirillOsenkov: There's no need as such, I was just following the pattern with the other loggers, and I thought exposing the reader and writer might be useful for other purposes.

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
KirillOsenkov: I've just fixed this.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Actually you're right, just storing it as int is easier and better. I've just made the change.

KirillOsenkov:
If you guys are OK with it, I'd like to get this in. I'll continue iterating, adding more tests and will hook up the reader, but I'd like to do that in a separate PR.

ref/netstandard1.3/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -1327,6 +1327,29 @@ public enum TargetResultCode : byte
 }
 namespace Microsoft.Build.Logging
 {
+    public partial class BinaryLogger : Microsoft.Build.Framework.ILogger
```
jeffkl: You'll need to build against CoreCLR to regenerate these public APIs.

`cibuild.cmd --build-only --target CoreCLR`

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -0,0 +1,673 @@
...
+        private Dictionary<string, string> ReadStringDictionary()
+        {
+            int count = ReadInt32();
+
+            if (count == 0)
+            {
+                return null;
+            }
+
+            Dictionary<string, string> result = new Dictionary<string, string>();
```
rainersigwald: Worth calling `new Dictionary<string, string>(count)` to avoid reallocation?

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -0,0 +1,673 @@
...
+                result[key] = value;
+            }
+
+            return result;
+        }
+
+        private class TaskItem : ITaskItem
+        {
+            public string ItemSpec { get; set; }
+            public Dictionary<string, string> Metadata { get; } = new Dictionary<string, string>();
```
rainersigwald: Should this be one of the fancy collections we normally use? Might be nice for consistency but I doubt we need any of the additional functionality.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -0,0 +1,673 @@
...
+            {
+                string key = ReadString();
+                ITaskItem item = ReadItem();
+                list.Add(new DictionaryEntry(key, item));
+            }
+
+            return list;
+        }
+
+        private IEnumerable ReadItemList()
```
rainersigwald: Not sure I get the distinction between `ReadItems` and `ReadItemList`.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -0,0 +1,574 @@
...
+        {
+            this.binaryWriter = binaryWriter;
+        }
+
+        /// <summary>
+        /// Write a provided instance of BuildEventArgs to the BinaryWriter
+        /// </summary>
+        public void Write(BuildEventArgs e)
+        {
+            // the cases are ordered by most used first for performance
```
rainersigwald: How was "most used" determined?

src/MSBuild/MSBuild.csproj
```diff
@@ -18,6 +18,7 @@
     <RootNamespace>Microsoft.Build.CommandLine</RootNamespace>
     <AssemblyName>MSBuild</AssemblyName>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <UseVSHostingProcess>false</UseVSHostingProcess>
```
rainersigwald: ?

src/MSBuild/Resources/Strings.resx
```diff
@@ -620,6 +620,31 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+                       /bl
+                       /bl:output.binlog
+                       /bl:..\..\custom.binlog
+                       /binaryLogger
+    </value>
+    <comment>
+      LOCALIZATION: The following should not be localized:
+      1) "msbuild" 
+      2) the string "binlog" that describes the file extension
+      3) all switch names and their short forms e.g. /bl, /binlog and /binaryLogger
```
rainersigwald: one more missed reference to `/binlog`

KirillOsenkov: no, that's a file extension (.binlog)

src/MSBuild/MSBuild.csproj
KirillOsenkov: This turns off the annoying .vshost.exe process that we never need and that only gets in the way. It is never useful.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
KirillOsenkov: I did quite extensive perf tuning for MSBuild Structured Log Viewer, so I have some stats on the number of most common events. Messages by far the most common. Also simple logic: there are more tasks than targets, more targets than projects, more projects than builds.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
KirillOsenkov: MSBuild has 3 different implementations of ITaskItem that are significant. In one case we just have a list of ITaskItem, in another case it's a dictionary keyed by item type.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
KirillOsenkov: Nah.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
KirillOsenkov: Yes, will do that. Although reader performance is not important (or at least not as important). Writer importance is way more important and we try to not allocate there.

src/MSBuild/MSBuild.csproj
rainersigwald: Works for me. Can it be in its own commit with that description, please?

src/MSBuild/Resources/Strings.resx
rainersigwald: it is in point `2)` but in `3)` it's not.

KirillOsenkov: Oh, you're right. Sorry. Will fix.

src/MSBuild/MSBuild.csproj
KirillOsenkov: Sorry, already pushed...

KirillOsenkov: P.S. When we get to the new .csproj format this setting is already off by default, so it becomes a moot point.

PR #1760, opened 2017-02-27
FI to vs15.1: Make LargeAddressAware incremental

AndyGerlicher:
Cherry-pick 754c9d1 from xplat


rainersigwald:
@AndyGerlicher I think this is a bad regression and we should push to fix it at the earliest opportunity, so I targeted the 15.1 branch. Do you agree?

Issue #1751, opened 2017-02-24
MSBuild 15.0 is case sensitive for item transforms (regression).


ericstj:
Repro:
build https://gist.github.com/ericstj/bf0c3191800a2dd5a61d418eca4fffb9 with desktop Msbuild
`msbuild.exe test-casing-transforms.proj`
Output:
```
  ABC
  foo.ABC
  bar.ABC
```

Do the same with core msbuild
`dotnet build test-casing-transforms.proj`
Output:
```
  ABC
  bar.ABC
```

It would seem that MSBuild ignores the casing when doing an item reference, but uses strict casing when doing a transform.  Desktop MSBuild ignored the casing in both places.

rainersigwald:
Distressing. I don't observe a difference between .NET Core and full MSBuild, but I do see a difference between MSBuild 14 and MSBuild 15.1 . . .

```
S:\work\msb1751>msbuild /v:m
Microsoft (R) Build Engine version 15.1.548.43366
Copyright (C) Microsoft Corporation. All rights reserved.

  ABC
  bar.ABC

S:\work\msb1751>dotnet msbuild /v:m
Microsoft (R) Build Engine version 15.1.548.43366
Copyright (C) Microsoft Corporation. All rights reserved.

  ABC
  bar.ABC

S:\work\msb1751>"c:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe" /v:m
Microsoft (R) Build Engine version 14.0.25420.1
Copyright (C) Microsoft Corporation. All rights reserved.

  ABC
  foo.ABC
  bar.ABC

S:\work\msb1751>c:\Windows\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe /v:m
Microsoft (R) Build Engine version 4.6.1586.0
[Microsoft .NET Framework, version 4.0.30319.42000]
Copyright (C) Microsoft Corporation. All rights reserved.

  ABC
  foo.ABC
  bar.ABC
```

Case insensitivity is the correct behavior. I suspect this is a regression from lazy evaluation. Continuing to debug.

ericstj:
That could be too.  I was using MSBuild 14 when doing the desktop test.

ericstj:
[...]

PR #1766, opened 2017-02-28
Item references should be case-insensitive

rainersigwald:
Fix for #1751, which regressed with lazy item evaluation.

### DIFF
--- a/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
+++ b/src/XMakeBuildEngine/Evaluation/LazyItemEvaluator.cs
@@ -49,7 +49,13 @@ namespace Microsoft.Build.Evaluation
         /// </summary>
         private static CompareInfo s_invariantCompareInfo = CultureInfo.InvariantCulture.CompareInfo;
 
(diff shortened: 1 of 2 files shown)


rainersigwald:
@AndyGerlicher I think this is a bad regression and we should push to fix it at the earliest opportunity, so I targeted the 15.1 branch. Do you agree?

cdmihai:
The `MSBuildNameIgnoreCaseComparer` seems a bit overkill. When this flows into VS, it would be good to do an RPS run.

rainersigwald:
I liked that because it's made for this purpose, and is what is used for it elsewhere. It claims to have special performance improvements, which is probably not true with 10 years of additional development on `OrdinalIgnoreCase`, but maybe . . .

rainersigwald:
The CI build is failing because this branch doesn't have the changes to build on VS2017 but the CI definition has 2017. I tested on my machine by cherry-picking all of those changes into this branch, and tests (including the new ones) passed.

ericstj:
LGTM

PR #1779, opened 2017-03-01
Synchronize Dataflow dependency to what VS uses

cdmihai:
devenv.exe has a binding redirect for Dataflow of `oldVersion="4.0.0.0-4.5.65535.65535" newVersion="4.5.24.0"`. We need to have the same dependency, otherwise msbuild crashes when loaded in-proc inside devenv.exe


cdmihai:
@eerhardt you might want to wait for this one too.

cdmihai:

PR #1782, opened 2017-03-02
Add a set of unit-tests for serializing and deserializing BuildEventArgs.

KirillOsenkov:
Two minor fixes from code review feedback (Rainer).

The way the tests are written is objects are manually created, then serialized and deserialized back. The length of the bytes written needs to be the same as the number of bytes read back.

For each object type, a set of fields are compared on the original object and deserialized copy and we make sure the values are the same.

### DIFF
--- a/src/Build.UnitTests/BuildEventArgsSerialization_Tests.cs
+++ b/src/Build.UnitTests/BuildEventArgsSerialization_Tests.cs
@@ -1,0 +1,313 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


AndyGerlicher:
Looks like missing cases for `TaskCommandLine` and `CriticalBuildMessage`. Also, is this the right place to try writing an event without an EndOfFile or some other corruption that could occur if the writer crashed?

KirillOsenkov:
Good catch, added tests for `TaskCommandLine` and `CriticalBuildMessage`. Also added a test for abruptly ending stream.

KirillOsenkov:

Issue #1776, opened 2017-03-01
ResGen path too long on mono-osx


filipnavara:
I am receiving the following error with recent Mono / Visual Studio on OS X:

> /Library/Frameworks/Mono.framework/Versions/4.8.0/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(2881,5): error MSB3455: ResGen.exe may not run because the command line is 48351 characters long, which exceeds the maximum length of the command. To fix this problem, please either (1) remove unnecessary assembly references, or (2) make the paths to those references shorter. [/Users/filipnavara/Desktop/emclient/MailClient/MailClient.csproj

Apparently the unit test that is supposed to check for proper behaviour in this case is marked with [Trait("Category", "mono-osx-failing")] attribute.

_We hit this problem while trying to workaround a different Mono bug. The default resource generation code messes up embedded 8-bit PNGs with alpha channel, possibly due to issue in the  System.Drawing or System.Windows.Forms implementation that is outside the scope of this project. We have our own reimplementation of these two libraries on top of native OS X API so naturally our first idea was to build our own ResGen that references them and simply update the build process to point to it using the ResGenToolPath and ResGenExecuteAsTool properties. However this is not possible due to the broken command line splitting. Reimplementing the whole GenerateResource task seems to be too cumbersome for our purposes._

filipnavara:
Actually on second thought the problem is slightly different than I thought. ResGen 4.0 and newer supports response files. However the invocation code for that is wrapped with NativeSharedMethods.IsWindows conditions, which is also why the unit tests fails on OS X platforms.

Similarly the code for splitting the command line is wrapped with a condition:

if (!_resgenPath.Equals(Path.GetDirectoryName(NativeMethodsShared.GetLongFilePath(ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("resgen.exe", TargetDotNetFrameworkVersion.Version35))), StringComparison.OrdinalIgnoreCase))

Unfortunately this condition doesn't check for NativeSharedMethods.IsWindows, so on OS X the paths are not properly split (except for an exact path match for old tools) and the response files are not used either. The condition should be either modified to check for  NativeSharedMethods.IsWindows or alternatively the NativeSharedMethods.IsWindows condition should be removed completely.

AndyGerlicher:
Marking this up-for-grabs. Thanks for looking into the problem. If you'd like to submit a patch for this we would be happy to accept.

PR #1789, opened 2017-03-03
Fix command line length errors on Mono / OS X

filipnavara:
Currently the ResGen task tool implementation supports the response
file syntax only on Windows systems. The GenerateResource task should
mirror that behaviour. It is thus necessary to enable the long command
line splitting on non-Windows systems.

Issue #1776

### DIFF
--- a/src/Tasks/GenerateResource.cs
+++ b/src/Tasks/GenerateResource.cs
@@ -995,7 +995,8 @@ namespace Microsoft.Build.Tasks
 
             // We need to do a whole lot of work to make sure that we're not overrunning the command line ... UNLESS
             // we're running ResGen 4.0 or later, which supports response files. 
-            if (!_resgenPath.Equals(Path.GetDirectoryName(NativeMethodsShared.GetLongFilePath(ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("resgen.exe", TargetDotNetFrameworkVersion.Version35))), StringComparison.OrdinalIgnoreCase))
+            if (NativeMethodsShared.IsWindows &&
+                !_resgenPath.Equals(Path.GetDirectoryName(NativeMethodsShared.GetLongFilePath(ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("resgen.exe", TargetDotNetFrameworkVersion.Version35))), StringComparison.OrdinalIgnoreCase))
             {
                 ResGen resGen = CreateResGenTaskWithDefaultParameters();
 


AndyGerlicher:
Thanks for contributing! It would be great to have a unit test for this, but I'm not sure how reasonable that would be. Were you able to at least repro the error (command too long) and then verify this fixes the issue as expected?

filipnavara:
There's already unit test for it, but it is disabled on OS X (due to other errors related to the ResGen response files enabled only on Windows). I have verified that it fixes the error in our case.

radical:
IIUC, the limit on OSX (and other unices) is different from windows. On my macOS box:
```
$ getconf ARG_MAX
262144
```

.. which is significantly higher than the limit in the code:

```
private static int s_maximumCommandLength = 28000;
```

This limit can be obtained via the `sysconf(3)` libc function. It would be good to use something close to the actual limit on the system.

src/Tasks/GenerateResource.cs
```diff
@@ -995,7 +995,8 @@ private bool TransformResourceFilesUsingResGen(List<ITaskItem> inputsToProcess,
 
             // We need to do a whole lot of work to make sure that we're not overrunning the command line ... UNLESS
             // we're running ResGen 4.0 or later, which supports response files. 
-            if (!_resgenPath.Equals(Path.GetDirectoryName(NativeMethodsShared.GetLongFilePath(ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("resgen.exe", TargetDotNetFrameworkVersion.Version35))), StringComparison.OrdinalIgnoreCase))
+            if (NativeMethodsShared.IsWindows &&
```
radical: It might be easier to read/understand if:

```
bool usingResgenWithResponseFileSupport = NativeMethodsShared.IsWindows &&
                                   !_resgenPath.Equals(Path.GetDirectoryName(NativeMethodsShared.GetLongFilePath(ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("resgen.exe", TargetDotNetFrameworkVersion.Version35))), StringComparison.OrdinalIgnoreCase);

if (usingResgenWithResponseFileSupport)
{
    ...
```

And update the comment to mention that mono's resgen supports response file only on windows.

filipnavara: Just for the record, the Mono ResGen doesn't support response files on any platform. Neither does it support the /r:<reference> syntax. It is basically unusable by msbuild at this moment. We use a customized and extended version for our purposes. The response file requirement is something from ResGen.cs, where it is disabled for non-Windows system. This patch just aligns the "if" conditions there with the ones in GenerateResource.cs. I would be equally fine with removing the conditions in ResGen.cs and enabling the response files on all platforms, but I suspected someone put it there for a reason beyond my understanding.

radical: Ah, I misunderstood your commit message! Mono runs on windows too, so it would be good to explicitly mention that mono's resgen doesn't support response files.
- The condition needs to have a ` && NativeMethodsShared.IsMono &&` too.
- Update the comment to reflect the mono case.
- The same changes would be required in `ResGen.cs` too
- And IMHO, the commit message could be a little clearer, like:

=============

Fix command line length errors with resgen on mono/osx

Mono's `resgen` tool doesn't support response files. So, response files are
supported only when using .NET's `resgen` (4.0) tool.

This manifested as command line length errors for resgen, on mono/osx (see issue #1776).

```
/Library/Frameworks/Mono.framework/Versions/4.8.0/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(2881,5): error MSB3455: ResGen.exe may not run because the command line is 48351 characters long, which exceeds the maximum length of the command. To fix this problem, please either (1) remove unnecessary assembly references, or (2) make the paths to those references shorter. [/Users/filipnavara/Desktop/emclient/MailClient/MailClient.csproj
```

GenerateResource task only checked for `resgen.exe` path and on OSX, incorrectly
assumed that it must be `resgen 4.0` with support for response files. It 
invokes `ResGen` task, which has an additional condition `IsWindows` for using
response files, triggering the alternative code path but that fails due to command line 
length being too long.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   So, the `GenerateResource` and `ResGen` tasks should use response files only with
.NET's resgen tool.

filipnavara: The Mono's ResGen doesn't work with the GenerateResource task for more reasons than just the response files. It doesn't support the "/r:" reference arguments either. It would need significant changes to be used in this scenario in any case, thus adding NativeSharedMethods.IsMono condition doesn't make sense. Updating Mono ResGen would make sense. Also, I am not sure whether NativeSharedMethods.IsMono checks for Mono runtime or Mono toolchain, which may not match on Windows.

In my opinion the best option is to remove all the NativeSharedMethods.IsWindows conditions from both GenerateResource.cs and ResGen.cs. At the same time the disabled unit tests could be reenabled on OS X. The option to use custom ResGen tool was added with .NET 4 on Windows and thus any custom ResGen tool that was used had to support response files in the first place.

What I was trying to fix was invocation of our custom ResGen, which is based on Mono's ResGen. Any other case was already broken and remains broken even after applying the patch. I tried to go with the least invasive way, by enabling the command line splitting, where the conditions were not aligned over the code. However I am equally fine with aligning the conditions the other way and allowing response files on all platforms.

AndyGerlicher:
This looks ok to me, but I'd like a comment noting the limitations of this change and why it's needed like @radical was asking for.

With that are you ok with this change @radical ?

filipnavara:
Ok, I should have been more clear from the beginning about the problem that I am actually trying to fix. MSBuild allows to specify two properties - ResGenExecuteAsTool and ResGenToolPath - that force the GenerateResource task to run external tool instead of the internal resource compilation in the task itself.

On OS X this doesn't work on projects that contain many .resx files and it results in the following error:

> error MSB3455: ResGen.exe may not run because the command line is 48351 characters long, which exceeds the maximum length of the command. To fix this problem, please either (1) remove unnecessary assembly references, or (2) make the paths to those references shorter.

There are three fundamental problems with the code:

1. The command line limit that is enforced with the error message doesn't actually apply to Unix systems.
2. There's a code for splitting the execution of resgen.exe to separate invocations with command line shorter than the limit. This code is actually never used on non-Windows builds.
3. Windows resgen.exe uses response files for passing the command line, which is disabled on non-Windows builds.

I tried to investigate the problems above and I found the following:

1. The command line limit is arbitrarily set to 28000, which is slightly less than the actual Windows limit of 32k. The reason why it is set lower than 32k is because the above mentioned splitting code doesn't enforce the limit precisely (at least that's what the comments say).
2. The splitting code is not used on Windows either, since it applies only to ResGen 3.5, but the .targets file always sets ResGenExecuteAsTool to false in that case. Maybe I am missing something or maybe it's just legacy from old days when GenerateResource always executed resgen.exe on older .NET Framework versions.
3. The version history dates the change to the initial commits of the "xplat" branch and there's no specific explanation for why the response files are disabled on non-Windows systems.

Now I need to find a way how to make this working and I am willing to invest time into it, but I need consensus on what needs to be done. Problem 1 is independent from problem 2, 3, but if it gets fixed then I won't hit 2 and 3. The big question is, should response files be supported on non-Windows systems? If the answer is yes then I will modify ResGen.cs and remove NativeMethodsShared.IsWindows checks from there. If the answer is no, then the code in this pull request would solve the problem as-is by enabling the command line splitting on non-Windows system and never using response files.

PR #1793, opened 2017-03-03
Replay

KirillOsenkov:
Binary log replay functionality. 

Instead of specifying a project or solution, you can specify a .binlog file to "build" and it will replay all the events in the source log into the usual loggers.

Note that BuildLogReplayEventArgs and its base class EventArgsDispatcher need to be made public for the replay functionality in MSBuild.exe to work. I've made the class sealed and added <remarks> with explanations.

### DIFF
--- a/src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
+++ b/src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
@@ -10,7 +10,8 @@ namespace Microsoft.Build.Logging
     /// Provides a method to read a binary log file (*.binlog) and replay all stored BuildEventArgs
     /// by implementing IEventSource and raising corresponding events.
     /// </summary>
-    public class BinaryLogReplayEventSource : EventArgsDispatcher
+    /// <remarks>The class is public so that we can call it from MSBuild.exe when replaying a log file.</remarks>
+    public sealed class BinaryLogReplayEventSource : EventArgsDispatcher
     {
         /// <summary>
         /// Read the provided binary log file and raise corresponding events for each BuildEventArgs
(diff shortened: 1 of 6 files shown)


KirillOsenkov:

KirillOsenkov:
@dotnet-bot test Ubuntu16.04 Build for CoreCLR

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
```diff
@@ -10,65 +10,54 @@ namespace Microsoft.Build.Logging
...
-                    Dispatch(instance);
+                        Dispatch(instance);
+                    }
                 }
             }
+            catch (Exception ex)
+            {
+                string code;
+                string helpKeyword;
+                var text = ResourceUtilities.FormatResourceString(out code, out helpKeyword, "InvalidLogFileFormat", ex.Message);
```
cdmihai: there appears to be a FormatResourceString overload that does not have the out params

src/MSBuild/XMake.cs
```diff
@@ -598,26 +598,36 @@ string commandLine
...
                     DateTime t1 = DateTime.Now;
 
+                    // If the primary file passed to MSBuild is a .binlog file, play it back into passed loggers
+                    // as if a build is happening
+                    if (FileUtilities.IsBinaryLogFilename(projectFile))
+                    {
+                        ReplayBinaryLog(projectFile, loggers);
+                    }
+                    else // regular build
+                    {
```
cdmihai: The new else block is not indented.

KirillOsenkov:
There's already a test for replay here:
https://github.com/Microsoft/msbuild/blob/xplat/src/Build.UnitTests/BinaryLogger_Tests.cs#L46

PR #1798, opened 2017-03-06
Add versioned check for GetApiContractReferences

XianghanWang:
Correct paths for reference assemblies for systemManagementContract, part of the IOT extension SDK, are not resolved with insider preview SDK, 15003 and 15021. The files are in here: 
C:\Program Files (x86)\Windows Kits\10\References\<VERSION>\Windows.System.SystemManagementContract\3.0.0.0\Windows.System.SystemManagementContract.winmd. 
The reason is when we call GetApiContractReferences we didn't check the Versioned property of the sdk. So it returned the path without <VERSION>


src/Utilities/ToolLocationHelper.cs
```diff
@@ -985,7 +984,19 @@ string registryRoot
...
+                            {
+                                if (manifest.VersionedContent)
+                                {
+                                    extensionSdkReferences = GetApiContractReferences(extensionSdk.ApiContracts, matchingSdk.Path, manifest.PlatformVersion);
+                                }
+                                else
+                                {
+                                    extensionSdkReferences = GetApiContractReferences(extensionSdk.ApiContracts, matchingSdk.Path);
+                                }
+                            }
```
AndyGerlicher: If `TryGetPlatformManifest()` returns false, do we need to revert to the old behavior?

AndyGerlicher: It seems like that `else` should move up one level and it would only add the parameter if `TryGetPlatformManifest` returned `true` and `VersionedContent` was `true`, otherwise assume it's not?

AndyGerlicher:
Your branch is pretty out of sync. Can you run this:
```
git remote add upstream https://github.com/Microsoft/msbuild.git
git fetch --all
git reset upstream/xplat --hard
cherry-pick 0f7442d43e8d4ccddaaa0d4109357952019a80af
```

Verify:
```
git log -2
```
Should be your new commit + [latest xplat](https://github.com/Microsoft/msbuild/commits/xplat)

Push:
```
git push -f
```

Next time I'd recommend creating a branch on your end and starting the PR from that. Probably not worth it now since the PR is already open.

rainersigwald:
Shouldn't the target here be `vs15.1` instead of `xplat?

XianghanWang:
Create new PR based on vs15.1 branch.

PR #1802, opened 2017-03-07
Add versioned check for GetApiContractReferences

XianghanWang:
Correct paths for reference assemblies for systemManagementContract, part of the IOT extension SDK, are not resolved with insider preview SDK, 15003 and 15021. The files are in here: 
C:\Program Files (x86)\Windows Kits\10\References\<VERSION>\Windows.System.SystemManagementContract\3.0.0.0\Windows.System.SystemManagementContract.winmd. 
The reason is when we call GetApiContractReferences we didn't check the Versioned property of the sdk. So it returned the path without <VERSION>


XianghanWang:
@AndyGerlicher @rainersigwald could you help review the change? This is the PR based on vs15.1 branch. The previous xplat PR is canceled.

PR #1806, opened 2017-03-07
PVS-Studio: fixed vulnerabilitys

skhrenov:
[V3041](https://www.viva64.com/en/w/V3041/) The expression was implicitly cast from 'long' type to 'float' type. Consider utilizing an explicit type cast to avoid the loss of a fractional part. An example: double A = (double)(X) / Y;. Microsoft.Build CommunicationsUtilities.cs 593

[V3118](https://www.viva64.com/en/w/V3118/) Milliseconds component of TimeSpan is used, which does not represent full time interval. Possibly 'TotalMilliseconds' value was intended instead. MSBuild XMake.cs 564

[V3095](https://www.viva64.com/en/w/V3095/) The 'name' object was used before it was verified against null. Check lines: 230, 236. Microsoft.Build.Tasks GenerateBindingRedirects.cs 230
[V3095](https://www.viva64.com/en/w/V3095/) The 'publicKeyToken' object was used before it was verified against null. Check lines: 232, 236. Microsoft.Build.Tasks GenerateBindingRedirects.cs 232

V3095 = CWE-476 (NULL Pointer Dereference)


AndyGerlicher:
@skhrenov can you target this against `xplat` branch please? Master is not our current active branch.

.gitignore
```diff
@@ -229,3 +229,4 @@ Tools/
 
 # Visual Studio Code
 .vscode/
+/src/MSBuild.plog
```
AndyGerlicher: I assume this is some PVS-Studio file?

skhrenov: Yes. This is a PVS-Studio errors log.

skhrenov:
Re-created in the xplat branch.

PR #1809, opened 2017-03-07
Execute inner builds in parallel

### LINKED ISSUE #1064: Cross-targeting dispatch to inner builds should be parallelized
See https://github.com/Microsoft/msbuild/pull/1063#pullrequestreview-1059961

rainersigwald:
Instead of batching over `%(_TargetFramework.Identity)` (and thus
serializing calls to the MSBuild task), issue build requests for all
target frameworks in parallel. Since the escape-hatch parameter
`BuildInParallel` isn't defined in the outer build, I defined it here to
preserve the existing behavior of being able to serialize the build even
with /m by passing /p:BuildInParallel=false.

Fixes #1064

### DIFF
--- a/src/Tasks/Microsoft.Common.CrossTargeting.targets
+++ b/src/Tasks/Microsoft.Common.CrossTargeting.targets
@@ -12,6 +12,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <BuildInParallel Condition="'$(BuildInParallel)' == ''">true</BuildInParallel>
     <ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets Condition="'$(ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets)' == ''">true</ImportByWildcardBeforeMicrosoftCommonCrossTargetingTargets>
   </PropertyGroup>
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.CrossTargeting.targets\ImportBefore\*.targets"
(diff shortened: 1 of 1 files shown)


rainersigwald:
@nguerrera This is step 1 of making everything go in parallel.

src/Tasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -22,6 +23,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Target Name="_ComputeTargetFrameworkItems" Returns="@(InnerOutput)">
     <ItemGroup>
       <_TargetFramework Include="$(TargetFrameworks)" />
```
cdmihai: Renaming `_TargetFramework` to `_TargetFrameworks` would make it clear there are multiple tfms

rainersigwald: Convention is generally that items have singular names--like `Compile`, `ProjectReference`, and so on. I agree that it's a bit confusing here . . . think it's worth changing?

cdmihai:
Did this have any perf improvements? :)

rainersigwald:
I'm trying to measure on a project reported as being annoyingly slow in the old configuration, but installing VS with every checkbox so that I can build all of its flavors is entering its second hour 😬

src/Tasks/Microsoft.Common.CrossTargeting.targets
cdmihai: I don't think so, the scope of the `_TargetFramework ` since to be local to the `_ComputeTargetFrameworkItems` target. For now :)

rainersigwald:
Aha, a benchmark (ish).

On a 4-core VM (so this isn't the maximum speedup), I tested VS2017 RTW against this change on https://github.com/onovotny/Zeroconf (selected by , which builds for 8 TFs. It went from `00:00:34.85` to `00:00:14.45` total elapsed build time, and 4 nodes with 88% utilization.

@onovotny You get your wish from https://twitter.com/onovotny/status/808841868016570368 :)

PR #1826, opened 2017-03-08
Fix a unit-test.

KirillOsenkov:
A theory I have is that the file name needs to be unique to avoid two different processes accessing the same file name at the same time.


KirillOsenkov:
I'm going to abandon this since it clearly doesn't help.

PR #1827, opened 2017-03-08
Don't set console color from FileLogger

### LINKED ISSUE #1792: MSBuild changes console colors when logging to a file
We started running into build failures on Travis CI using MSBuild. It appears the logger produces thousands of unnecessary ANSI color codes.

Example:
https://travis-ci.org/aspnet/DotNetTools/jobs/206846997
The log output appears normal:
image

But if you look at the raw log file, you'll see that hundreds of color codes have been logged.
https://s3.amazonaws.com/archive.travis-ci.org/jobs/206846997/log.txt

image

cc @dougbu

rainersigwald:
Because FileLogger used the default value of the SetColor and ResetColor
delegates, it was quietly changing and resetting console color for every
log message that was considered by the FileLogger. On Windows, this was
unnecessary but not too noticeable. On Unix systems, though, console
colors are set by emitting an escape code to the output stream, so this
became observable.

Explicitly passing the Dont* mechanisms to prevent the unnecessary work
from happening. I removed the null check for the `write` delegate from
ConsoleLogger's constructor because it seemed cleaner to set the
delegate to `null` from FileLogger than to explicitly specify the
default `Console.Out.Write`.

Fixes #1792.

### DIFF
--- a/src/Build/Logging/ConsoleLogger.cs
+++ b/src/Build/Logging/ConsoleLogger.cs
@@ -103,7 +103,6 @@ namespace Microsoft.Build.Logging
             ColorResetter colorReset
         )
         {
-            ErrorUtilities.VerifyThrowArgumentNull(write, "write");
             _verbosity = verbosity;
             _write = write;
             _colorSet = colorSet;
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
LGTM!

AndyGerlicher:

PR #1828, opened 2017-03-09
Move an exception one level up.

KirillOsenkov:
To see more details from a unit-test exception let's not catch the exception inside the Replay() method but instead catch it at the caller site.

To still give a meaningful user message we need to share the resource for "InvalidLogFileFormat".


AndyGerlicher:

AndyGerlicher:
FYI the Windows failure is my issue (upgraded CI to VS2017 but my PR isn't in yet), ignore that.

PR #1829, opened 2017-03-09
PVS-Studio: fixed vulnerabilities

skhrenov:
[V3041](https://www.viva64.com/en/w/V3041/) The expression was implicitly cast from 'long' type to 'float' type. Consider utilizing an explicit type cast to avoid the loss of a fractional part. An example: double A = (double)(X) / Y;. Microsoft.Build CommunicationsUtilities.cs 615

[V3118](https://www.viva64.com/en/w/V3118/) Milliseconds component of TimeSpan is used, which does not represent full time interval. Possibly 'TotalMilliseconds' value was intended instead. MSBuild XMake.cs 629

[V3095](https://www.viva64.com/en/w/V3095/) The 'name' object was used before it was verified against null. Check lines: 229, 235. Microsoft.Build.Tasks GenerateBindingRedirects.cs 229
[V3095](https://www.viva64.com/en/w/V3095/) The 'publicKeyToken' object was used before it was verified against null. Check lines: 231, 235. Microsoft.Build.Tasks GenerateBindingRedirects.cs 231

V3095 = CWE-476 (NULL Pointer Dereference)

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -626,7 +626,7 @@ namespace Microsoft.Build.CommandLine
 
                     if (!String.IsNullOrEmpty(timerOutputFilename))
                     {
-                        AppendOutputFile(timerOutputFilename, elapsedTime.Milliseconds);
+                        AppendOutputFile(timerOutputFilename, (long)elapsedTime.TotalMilliseconds);
                     }
                 }
                 else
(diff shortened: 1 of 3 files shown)


src/MSBuild/XMake.cs
```diff
@@ -626,7 +626,7 @@ string commandLine
 
                     if (!String.IsNullOrEmpty(timerOutputFilename))
                     {
-                        AppendOutputFile(timerOutputFilename, elapsedTime.Milliseconds);
+                        AppendOutputFile(timerOutputFilename, (long)elapsedTime.TotalMilliseconds);
```
AndyGerlicher: This is frustratingly easy to do and such a very different result... That's great that it caught this!

AndyGerlicher:

rainersigwald:
Thanks!

PR #1835, opened 2017-03-09
RI 15.1 -> master

AndyGerlicher:


AndyGerlicher:
@rainersigwald if CI passes I would like to do this now to get a new build of master out. Holding off on #1836 for now.

PR #1839, opened 2017-03-09
Update EditBin target to use VsInstallRoot property

AndyGerlicher:
VS150COMNTOOLS is not defined in our MicroBuild environment.

### DIFF
--- a/dir.targets
+++ b/dir.targets
@@ -152,10 +152,18 @@
            produce the right outputs. Continuing on error to allow
            test results even on a misconfigured build machine. -->
       <EditBinContinueOnError Condition="'$(TF_BUILD)' == 'True' or '$(JENKINS_URL)' != ''">ErrorAndContinue</EditBinContinueOnError>
+
+      <!--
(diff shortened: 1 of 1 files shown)


AndyGerlicher:
Just so this doesn't get lost, we found that the VSTS MSBuild task doesn't use the Developer Console or something similar, so common environment variables are not set. `VS140COMNTOOLS` worked before because Visual Studio sets that machine-wide. With 2017 that is no longer the case since you can install more than one copy of Visual Studio. So in order to find VS I used the `VsInstallRoot` that MSBuild [sets](https://github.com/Microsoft/msbuild/blob/xplat/src/MSBuild/app.config#L81). To make things more complicated, our Jenkins CI build will build a copy of MSBuild and package it up as a "Bootstrap" MSBuild and then rebuild and run tests. That is a standalone environment and does not set `VsInstallRoot` like a normal copy of MSBuild would (we don't want Visual Studio instance assets to leak into our inner build).

I don't think there's a bug here, but it's not very straightforward!

AndyGerlicher:

KirillOsenkov:

KirillOsenkov:

PR #1840, opened 2017-03-09
Extend GetSDKReferenceFiles to allow specifying Platform Identifier and Version

AndyGerlicher:
In order to discover the correct API reference contracts, we must be able to query the manifest and determine the platform version. This change adds `TargetPlatformIdentifier` and `TargetPlatformVersion` to the `GetSDKReferenceFiles` task in order to have enough information to read the platform manifest.


src/Utilities/ToolLocationHelper.cs
```diff
@@ -964,6 +964,40 @@ public static string[] GetPlatformOrFrameworkExtensionSdkReferences
...
+        /// <param name="diskRoots">The disk roots used to gather installed SDKs.</param>
+        /// <param name="extensionDiskRoots">The disk roots used to gather installed extension SDKs.</param>
+        /// <param name="registryRoot">The registry root used to gather installed extension SDKs.</param>
+        public static string[] GetPlatformOrFrameworkExtensionSdkReferences
+            (
+                string extensionSdkMoniker,
+                string targetSdkIdentifier,
+                string targetSdkVersion,
+                string targetPlatformIdentifier,
+                string targetPlatformVersion,
```
rainersigwald: I don't like adding an overload that adds strings in the middle of the arguments. It's hard to move from the old call to this one, because the string arguments change meaning depending on how many there are.

src/XMakeTasks/GetSDKReferenceFiles.cs
```diff
@@ -766,6 +784,8 @@ private string[] GetReferencePathsFromManifest(ITaskItem sdk)
                     sdk.GetMetadata(GetInstalledSDKLocations.SDKNameMetadataName),
                     TargetSDKIdentifier,
                     TargetSDKVersion,
+                    TargetPlatformIdentifier,
+                    TargetPlatformVersion,
```
rainersigwald: These need to move too, since you moved the params.

src/Utilities/ToolLocationHelper.cs
```diff
@@ -989,6 +1023,7 @@ string registryRoot
                 else
                 {
                     string targetSdkPath = matchingSdk.Path;
+                    string platformVersion = GetPlatformVersion(matchingSdk, targetPlatformIdentifier, targetPlatformVersion);
```
rainersigwald: Does this need to be out here, or could it be inside the `if`s?

src/Utilities/ToolLocationHelper.cs
```diff
@@ -1516,12 +1569,12 @@ private static TargetPlatformSDK GetMatchingPlatformSDK(string targetPlatformIde
         /// Given a target platform identifier and version and locations in which to search, find the TargetPlatformSDK 
         /// object that matches.
         /// </summary>
-        private static TargetPlatformSDK GetMatchingPlatformSDK(string targetPlatformIdentifier, Version targetPlatformVersion, string[] diskRoots, string registryRoot)
+        private static TargetPlatformSDK GetMatchingPlatformSDK(string targetPlatformIdentifier, Version targetPlatformVersion, string[] diskRoots, string[] multiPlatformDiskRoots, string registryRoot)
```
rainersigwald: Ideally move this one to the end too?

src/Utilities/ToolLocationHelper.cs
XianghanWang: it needs to be cached here. the matching sdk will be changed in the if().

XianghanWang: it needs to be cached here. the matchingsdk will be changed in the if().

rainersigwald: Ah, I see. 👍

PR #1841, opened 2017-03-09
Log when a property is being assigned a new value.

KirillOsenkov:
Log a low priority message during evaluation when a property that was previously assigned is being assigned a new and different value. This is useful to determine when your property might be stomped over later.

Also adding Evaluation started and Evaluation finished messages for each project evaluation to logically isolate messages coming from the evaluation phase.

Sample log output:

```
Evaluation started ("C:\CodeCleanupTools\SortProjectItems\SortProjectItems.csproj")
Property reassignment: $(ProcessorArchitecture)="msil" (previous value: "AnyCPU") at C:\MSBuild\bin\Bootstrap\15.0\Bin\Microsoft.Common.CurrentVersion.targets (457,5)
Evaluation finished ("C:\CodeCleanupTools\SortProjectItems\SortProjectItems.csproj")
```

### DIFF
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -662,6 +662,55 @@ namespace Microsoft.Build.UnitTests.Evaluation
         }
 
         /// <summary>
+        /// Log when a property is being assigned a new value.
+        /// </summary>
+        [Fact]
+        public void LogPropertyAssignments()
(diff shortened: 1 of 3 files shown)


jaredpar:
Is this only available as a log entry or can it be made into a warning?

KirillOsenkov:
Just a message (I'm not familiar with how strict mode works). I'm guessing this would fire a lot even for trivial builds (just the way common targets are written).

PR #1843, opened 2017-03-10
Fix DescendingByElapsedTimeComparer

KirillOsenkov:
When replaying a binary log with multiproc I've discovered a crash in the DescendingByElapsedTimeComparer. In the case when both p1 and p2 were reentrant we were returning -1 always even if they were different objects.

This resulted in a crash:
```
There was an exception while reading the log file: Unable to sort because the IComparer.Compare() method returns inconsistent results. Either a value does not compare equal to itself, or one value repeatedly compared to another value yields different results. IComparer: 'Microsoft.Build.BackEnd.Logging.BaseConsoleLogger+PerformanceCounter+DescendingByElapsedTime'.
```

In the case when both performance counters are reentrant let's sort them by name such that it's stable.

The bug was always reproducible and after the fix I've never seen it again. The user impact if we don't fix is that they won't be able to replay logs with /m specified (because when printing perf summary all reentrant times are 00:00:00).

### DIFF
--- a/src/Build/Logging/BaseConsoleLogger.cs
+++ b/src/Build/Logging/BaseConsoleLogger.cs
@@ -978,7 +978,7 @@ namespace Microsoft.Build.BackEnd.Logging
                     }
                     else
                     {
-                        // both reentrant; sort properly to avoid throwing
+                        // both reentrant; sort stably by another field to avoid throwing
                         return string.Compare(p1.ScopeName, p2.ScopeName, StringComparison.Ordinal);
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/Build/Logging/BaseConsoleLogger.cs
```diff
@@ -966,16 +966,21 @@ public int Compare(object o1, object o2)
...
                     }
-                    else
+                    else if (!p1.reenteredScope && p2.reenteredScope)
                     {
                         // p2 was reentrant; sort first
                         return 1;
                     }
+                    else
+                    {
+                        // both reentrant; sort properly to avoid throwing
```
rainersigwald: nit: instead of "properly" say "stably by another field", so the motivation is clearer.

KirillOsenkov: Fixed

PR #1844, opened 2017-03-10
Support replaying logs with /m.

KirillOsenkov:
Previously we didn't support replaying logs when /m was specified (we were just ignoring all distributed logger records).

Now we also log into the central logger for each distributed logger record.

We also correctly initialize loggers that implement INodeLogger by passing the CPU count.

After this change replaying a log with /m is now supported.


AndyGerlicher:
Looks good. The test failures look worrisome, but I can't see how this change could cause that.

AndyGerlicher:

Issue #1794, opened 2017-03-03
Expose MSBuild globbing APIs


cdmihai:
So API users can expand msbuild globs or regex match them just like the engine does.

API reqs and specs:
- [x] Globs should match against strings (matching is what CPS actually needs, and what triggered the creation of these APIs)
- [ ] Globs should be able to expand files on disk. Issue tracked separately: #1908
- [x] Globs should be composites
- [x] Glob matching should be path aware ("../a.cs" should not match a against "**/*.cs")

am11:
Can it be extended to support negative glob patterns? For example, `.gitignore` supports it: https://git-scm.com/docs/gitignore#_pattern_format

```xml
<ItemGroup>
  <None Remove="!**/*.cs" /> <!-- remove all but the CS files -->
  <None Include="!**/*Actors*" /> <!-- include all but the paths with `Actors` in the name -->
</ItemGroup>
```

cdmihai:
@am11 this issue is for exposing whatever the MSBuild engine does now secretively in a public API. I have opened a separate issue with your idea: #1876 

Regarding negative globs, you can still express their intent via Exclude attributes and Remove elements. Though negative globs would indeed be shorter to express.

cdmihai:
First version of the APIs is out: https://github.com/Microsoft/msbuild/tree/master/src/Build/Globbing

PR #1859, opened 2017-03-14
Globbing APIs

cdmihai:
Addresses #1794.

This PR has the subset of the APIs that CPS will need. Left for future work:
- Add Expand method.
- Don't use regexes when the glob does not contain any wildcards
- Composites combine regexes for perf

### DIFF
--- a/src/Build/Globbing/CompositeGlob.cs
+++ b/src/Build/Globbing/CompositeGlob.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Build.Globbing
+{
+    /// <summary>
+    ///     A Composite glob
+    /// </summary>
+    public class CompositeGlob : IMSBuildGlob
+    {
+        /// <summary>
+        /// The direct children of this composite
+        /// </summary>
+        public IEnumerable<IMSBuildGlob> Globs { get; }
+
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="globs">Children globs</param>
+        public CompositeGlob(IEnumerable<IMSBuildGlob> globs)
+        {
+            Globs = globs;
+        }
+
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="globs">children globs</param>
+        public CompositeGlob(params IMSBuildGlob[] globs) : this(globs.AsEnumerable())
+        {}
+
+        /// <inheritdoc />
+        public bool IsMatch(string stringToMatch)
+        {
+            return Globs.AsParallel().Any(g => g.IsMatch(stringToMatch));
+        }
+    }
+}
--- a/src/Build/Globbing/MSBuildGlob.cs
+++ b/src/Build/Globbing/MSBuildGlob.cs
@@ -1,0 +1,266 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+
+
+using System;
+using System.IO;
+using System.Text.RegularExpressions;
+using Microsoft.Build.Shared;
+
+namespace Microsoft.Build.Globbing
+{
+    /// <summary>
+    ///     Represents a parsed MSBuild glob.
+    ///     An MSBuild glob is composed of three parts:
+    ///     - fixed directory part: "a/b/" in "a/b/**/*test*/**/*.cs"
+    ///     - wildcard directory part: "**/*test*/**/" in "a/b/**/*test*/**/*.cs"
+    ///     - file name part: "*.cs" in "a/b/**/*test*/**/*.cs"
+    /// </summary>
+    public class MSBuildGlob : IMSBuildGlob
+    {
+        internal readonly string _fileSpec;
+        internal readonly string _globRoot;
+        private readonly string _matchFileExpression;
+        internal readonly bool _needsRecursion;
+        private readonly Lazy<Regex> _regex;
+
+        /// <summary>
+        ///     The fixed directory part.
+        /// </summary>
+        public string FixedDirectoryPart { get; }
+
+        /// <summary>
+        ///     The wildcard directory part
+        /// </summary>
+        public string WildcardDirectoryPart { get; }
+
+        /// <summary>
+        ///     The file name part
+        /// </summary>
+        public string FilenamePart { get; }
+
+        /// <summary>
+        ///     Whether the glob was parsed sucsesfully from a string.
+        ///     Illegal glob strings contain:
+        ///     - invalid path characters (other than the wildcard characters themselves)
+        ///     - "..."
+        ///     - ":"
+        ///     In addition, the wildcard directory part:
+        ///     - cannot contain ".."
+        ///     - if ** is present it must appear alone between slashes
+        /// </summary>
+        public bool IsLegal { get; }
+
+        private MSBuildGlob(
+            string globRoot,
+            string fileSpec,
+            string fixedDirectoryPart,
+            string wildcardDirectoryPart,
+            string filenamePart,
+            string matchFileExpression,
+            bool needsRecursion,
+            bool isLegalFileSpec)
+        {
+            FixedDirectoryPart = fixedDirectoryPart;
+            WildcardDirectoryPart = wildcardDirectoryPart;
+            FilenamePart = filenamePart;
+            IsLegal = isLegalFileSpec;
+
+            _globRoot = globRoot;
+            _fileSpec = fileSpec;
+            _matchFileExpression = matchFileExpression;
+            _needsRecursion = needsRecursion;
+
+            // compile the regex since it's expected to be used multiple times
+            _regex =
+                new Lazy<Regex>(
+                    () => new Regex(_matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled),
+                    true);
+        }
+
+        /// <inheritdoc />
+        public bool IsMatch(string stringToMatch)
+        {
+            ErrorUtilities.VerifyThrowArgumentNull(stringToMatch, nameof(stringToMatch));
+
+            if (FileUtilities.PathIsInvalid(stringToMatch))
+            {
+                return false;
+            }
+
+            var normalizedString = NormalizeMatchInput(stringToMatch);
+
+            return _regex.Value.IsMatch(normalizedString);
+        }
+
+        /// <summary>
+        ///     Similar to <see cref="IsMatch" /> but also provides the match groups for the glob parts
+        /// </summary>
+        /// <param name="stringToMatch"></param>
+        /// <returns></returns>
+        public MatchInfoResult MatchInfo(string stringToMatch)
+        {
+            ErrorUtilities.VerifyThrowArgumentNull(stringToMatch, nameof(stringToMatch));
+
+            if (FileUtilities.PathIsInvalid(stringToMatch) ||
+                !IsLegal)
+            {
+                return MatchInfoResult.Empty;
+            }
+
+            var normalizedInput = NormalizeMatchInput(stringToMatch);
+
+            bool isMatch;
+            string fixedDirectoryPart, wildcardDirectoryPart, filenamePart;
+            FileMatcher.GetRegexMatchInfo(
+                normalizedInput,
+                _regex.Value,
+                out isMatch,
+                out fixedDirectoryPart,
+                out wildcardDirectoryPart,
+                out filenamePart);
+
+            return new MatchInfoResult(isMatch, fixedDirectoryPart, wildcardDirectoryPart, filenamePart);
+        }
+
+        private string NormalizeMatchInput(string stringToMatch)
+        {
+            var rootedInput = Path.Combine(_globRoot, stringToMatch);
+            var normalizedInput = FileUtilities.GetFullPathNoThrow(rootedInput);
+
+            // Degenerate case when the string to match is empty.
+            // Ensure trailing slash because the fixed directory part has a trailing slash.
+            if (stringToMatch == string.Empty)
+            {
+                normalizedInput = normalizedInput + Path.DirectorySeparatorChar;
+            }
+
+            return normalizedInput;
+        }
+
+        /// <summary>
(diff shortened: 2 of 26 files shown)


src/Build/Globbing/CompositeGlob.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    ///     A Composite glob
+    /// </summary>
+    public class CompositeGlob : IMsBuildGlob
+    {
+        /// <summary>
+        /// The direct children of this composite
+        /// </summary>
+        public IEnumerable<IMsBuildGlob> Globs { get; }
+        //private readonly Lazy<IEnumerable<string>> _expandedFiles;
```
AndyGerlicher: Remove commented out code

src/Build/Globbing/Extensions/IMSBuildGlobExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+
+using System.Collections.Generic;
+
+namespace Microsoft.Build.Globbing.Extensions
+{
+    /// <summary>
+    /// Extensions for <see cref="IMsBuildGlob"/>
+    /// </summary>
+    public static class IMsBuildGlobExtensions
```
AndyGerlicher: Not an interface so remove the I from the name/file. Also for more consistency, use `MSBuildGlobExtensions` (uppercase `S`) even though resharper doesn't like it.

src/Build/Globbing/GlobVisitor.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            var globWithGaps = glob as MsBuildGlobWithGaps;
+            if (globWithGaps != null)
+            {
+                VisitGlobWithGaps(globWithGaps);
+
+                Visit(globWithGaps.MainGlob);
+            }
+        }
```
AndyGerlicher: Just FYI this can be re-written with C# 7.0 feature:
```
if (glob is MSBuildGlob msbuildGlob) { VisitMSBuildGlob(msbuildGlob) }
if (glob is CompisiteGlob compositGlob) { . . . }
if (glob is null) { . . . }
```

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,271 @@
...
+        /// <returns></returns>
+        public static MsBuildGlob Parse(string globRoot, string fileSpec)
+        {
+            ErrorUtilities.VerifyThrowArgumentNull(globRoot, nameof(globRoot));
+            ErrorUtilities.VerifyThrowArgumentNull(fileSpec, nameof(fileSpec));
+            ErrorUtilities.VerifyThrowArgumentInvalidPath(globRoot, nameof(globRoot));
+
+            if (globRoot == string.Empty)
+            {
+                globRoot = Directory.GetCurrentDirectory();
```
AndyGerlicher: This ok to read here rather than at construction time? Could it have changed?

src/Build/Globbing/MSBuildGlobWithGaps.cs
```diff
@@ -0,0 +1,70 @@
...
+
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Build.Shared;
+
+namespace Microsoft.Build.Globbing
+{
+    /// <summary>
+    ///     A glob with gaps. The gaps are represented as other globs.
```
AndyGerlicher: Document an example of what a gap is please :)

AndyGerlicher:
Looks like a really good start. Some dead code commented out. What's the plan there? Future PR or future iteration? I'd like to not commit it with that.

src/Build/Globbing/Extensions/IMSBuildGlobExtensions.cs
cdmihai: In my mind the class name meant to say that it contains extension methods for the `IMSBuildGlob` type. I guess that when you take a step back and look at it from a distance it does not look like a good name anymore :)

cdmihai:
@KirillOsenkov Another on Ubuntu: https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/_Ubuntu16.04_CoreCLR_prtest/541/

src/Build/Globbing/MSBuildGlob.cs
cdmihai: This method is the construction time, where the string gets parsed into components, and the glob root gets hardcoded in the fixed dir component. Nothing mutates after this method.

src/Build/Globbing/Extensions/IMSBuildGlobExtensions.cs
rainersigwald: Agree with Andy: drop the `I`.

src/Build/Globbing/GlobVisitor.cs
rainersigwald: I look forward to applying this one across the codebase.

src/Build/Globbing/GlobVisitor.cs
```diff
@@ -0,0 +1,63 @@
...
+        protected virtual void VisitCompositeGlob(CompositeGlob compositGlob)
+        {
+        }
+
+        protected virtual void VisitMSBuildGlob(MsBuildGlob msbuildGlob)
+        {
+        }
+    }
+
+    internal class ParsedGlobCollector : GlobVisitor
```
rainersigwald: I'd go ahead and put this in its own file. Easier to keep up if every class has its own home.

src/Build/Globbing/GlobVisitor.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        private readonly ImmutableList<MsBuildGlob>.Builder _collectedGlobs = ImmutableList.CreateBuilder<MsBuildGlob>();
```
rainersigwald: Is it worth keeping a `Builder` around? I thought the idea behind the Immutable* collections was that you could use them in the Haskell way and be ok, and Builders were just an escape hatch in case you had to do some intense one-time manipulation?

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,271 @@
...
+
+        private MsBuildGlob(
+            string globRoot,
+            string fileSpec,
+            string fixedDirectoryPart,
+            string wildcardDirectoryPart,
+            string filenamePart,
+            string matchFileExpression,
+            bool needsRecursion,
+            bool isLegalFileSpec)
```
rainersigwald: Feels a bit odd to take `isLegal` into the ctor as an argument. What's the motivation behind producing an object that's not legal if the caller already knows that?

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,271 @@
...
+            _globRoot = globRoot;
+            _fileSpec = fileSpec;
+            _matchFileExpression = matchFileExpression;
+            _needsRecursion = needsRecursion;
+
+            // compile the regex since it's expected to be used multiple times
+            _regex =
+                new Lazy<Regex>(
+                    () => new Regex(_matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled),
+                    true);
```
rainersigwald: Doesn't this `true` just opt into the default no-thread-safety-mode-specified case?

KirillOsenkov:
@cdmihai I think this is from March 15, and my fix was merged this morning. So I think this should be fixed for future PRs.

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,271 @@
...
+                out fixedDirectoryPart,
+                out wildcardDirectoryPart,
+                out filenamePart);
+
+            return new MatchInfoResult(isMatch, fixedDirectoryPart, wildcardDirectoryPart, filenamePart);
+        }
+
+        private string NormalizeMatchInput(string stringToMatch)
+        {
+            var rootedInput = Path.Combine(_globRoot, stringToMatch);
```
rainersigwald: So the only times this is called with a relative path are when that relative path is rooted to the same place as the glob? Is that expressible as a guarantee somehow?

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,271 @@
...
+            var normalizedInput = FileUtilities.GetFullPathNoThrow(rootedInput);
+
+            // Degenerate case when the string to match is empty.
+            // Ensure trailing slash because the fixed directory part has a trailing slash.
+            if (stringToMatch == string.Empty)
+            {
+                normalizedInput = normalizedInput + Path.DirectorySeparatorChar;
```
rainersigwald: Call `EnsureTrailingSlash` or something? Or does `GetFullPathNoThrow` guarantee no trailing slash?

src/Build/Globbing/MSBuildGlob.cs
rainersigwald: Ah, because the ctor is private and only called in a static method below. 👍

src/Build/Globbing/MSBuildGlob.cs
rainersigwald: This worries me a bit, too. What's the calling pattern like? If at some point in the future we start parallelizing evaluation will this run into races?

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,259 @@
...
+            FileMatcher.GetFileSpecInfo(
+                fileSpec,
+                out fixedDirectoryPart,
+                out wildcardDirectoryPart,
+                out filenamePart,
+                out matchFileExpression,
+                out needsRecursion,
+                out isLegalFileSpec,
+                FileMatcher.s_defaultGetFileSystemEntries,
+                (fixedDirPart, wildcardDirPart, filePart) =>
```
rainersigwald: Can you promote this lambda to a named function, so the name can provide some documentation as to its purpose?

src/Build/Microsoft.Build.csproj
```diff
@@ -772,9 +778,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" Condition="!$(Configuration.EndsWith('MONO'))"/>
   -->
   <Import Project="..\dir.targets" />
-  <Target Name="CopyImmutableAndDataflowLocal"
-          BeforeTargets="ResolveAssemblyReferences"
-          AfterTargets="ResolveNuGetPackages">
+  <Target Name="CopyImmutableAndDataflowLocal" BeforeTargets="ResolveAssemblyReferences" AfterTargets="ResolveNuGetPackages">
```
rainersigwald: Why whitespace change?

(He asked rhetorically, knowing it's VS's fault)

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -0,0 +1,259 @@
...
+        ///     If empty, the current working directory is used.
+        ///     Cannot be null, and cannot contain invalid path arguments.
+        /// </param>
+        /// <param name="fileSpec">The string to parse</param>
+        /// <returns></returns>
+        public static MSBuildGlob Parse(string globRoot, string fileSpec)
+        {
+            ErrorUtilities.VerifyThrowArgumentNull(globRoot, nameof(globRoot));
+            ErrorUtilities.VerifyThrowArgumentNull(fileSpec, nameof(fileSpec));
+            ErrorUtilities.VerifyThrowArgumentInvalidPath(globRoot, nameof(globRoot));
```
rainersigwald: Doesn't this include throw-if-null?

src/Shared/FileMatcher.cs
```diff
@@ -45,6 +45,8 @@ internal static class FileMatcher
         /// </summary>
         private static readonly char[] s_invalidPathChars = Path.GetInvalidPathChars();
 
+        internal static RegexOptions DefaultRegexOptions = RegexOptions.IgnoreCase;
```
rainersigwald: Const?

src/Build/Globbing/MSBuildGlobWithGaps.cs
cdmihai: Added explanation and example

src/Build/Globbing/MSBuildGlob.cs
cdmihai: Docs suggest true enables thread safety: https://msdn.microsoft.com/en-us/library/dd783092(v=vs.110).aspx
They are a bit confusing, so this is what the code suggests (if I followed all the layers of indirection correctly)

True leads to this construction code which uses a lock: https://source.dot.net/#System.Private.CoreLib/shared/System/Lazy.cs,f8f9288e02c57df5

False leads to this one which does not use a lock:
https://source.dot.net/#System.Private.CoreLib/shared/System/Lazy.cs,f3a8e7f97a578b26,references

So I guess setting it to true enables thread safety.

src/Build/Globbing/MSBuildGlob.cs
cdmihai: EnsureTrailing slash does not work with empty strings, and GetFullPathNoThrow does not add a trailing slash. Moreover, if  stringToMatch has illegal path characters GetFullPathNoThrow does not do anything.

src/Build/Globbing/MSBuildGlob.cs
cdmihai: True, but I repeated it here to make it more clear in the code what's going on. Since it's just a null check I did not think it would be a big perf hit. I'll take it out if you think otherwise.

src/Build/Globbing/MSBuildGlob.cs
cdmihai: The only current driver of the API so far is GetAllGlobs, which gets them from ItemSpecs. ItemSpecs will always pass in the path of their project file as the glob root. Except for in-memory projects, which will have an empty path. Should I add an optional parameter for the current directory?

src/Build/Globbing/MSBuildGlob.cs
cdmihai: Excellent idea. I like self-documenting code :D

src/Build/Microsoft.Build.csproj
cdmihai: I usually unstage these, guess I missed one. Now, which commit did this ...

AndyGerlicher:
Before we merge we should finalize if this is in 15.1. If so it will need to go to that branch.

cdmihai:
@dotnet-bot test Windows_NT Build for Desktop please.

cdmihai:
Retargeted for future release, closing this PR

PR #1866, opened 2017-03-15
Ask ProjectReferences if they cross-target

### LINKED ISSUE #1276: Avoid double evaluation of all ProjectReferences
As of #1018, the `ResolveProjectReferences` flow (which is executed during solution load in VS and is therefore very perf-sensitive) is two-phase:
1. Run [`_GetProjectReferenceTargetFrameworkProperties`](https://github.com/Microsoft/msbuild/blob/2db4d4b6821ff65987b2843d29cc5f02c4db59f5/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L1494-L1528) to ask the reference "What TFM of yours should I build to match my current TFM?"
2. Run [`ResolveProjectReferences`](https://github.com/Microsoft/msbuild/blob/2db4d4b6821ff65987b2843d29cc5f02c4db59f5/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L1569-L1697) to ask the reference "What is your primary output for this TFM?"

Those evaluations have different sets of global properties (the first sets `ReferringTargetFramework` and the latter sets `SetTargetFramework`), so they must be evaluated separately. Neither called target takes very long, but the evaluation is not cheap. This means that time to `ResolveProjectReferences` has roughly doubled in cost for all projects.

Cost of crosstargeting should be on a pay-for-play basis as much as possible. Existing csproj -> existing csproj (ditto vcxproj) shouldn't be slowed down by this.

/cc @nguerrera

rainersigwald:
Before cross-targeting was implemented, a project had a single output
that was queried in ResolveProjectReferences by either `GetTargetPath`
(in single-project-build scenarios) or the default target (`Build`) in
command line builds.

A cross-targeting project doesn't have a single output--it has one for
each of its `TargetFrameworks`. A ProjectReference, then, has to first
figure out which output to ask for--the one that's the best fit for the
current project.

The initial implementation of this was simple: unconditionally call
`GetTargetFrameworkProperties` on every ProjectReference, passing a
`ReferringTargetFramework`. Then, if the project cross-targets, specify
the resulting TF on future calls to the ProjectReference. If the project
has a single TF, do not specify it to avoid building it once (from the
solution  build) and again (from a reference, explicitly specifying the
one TF).

Unfortunately, adding a global property for `ReferringTargetFramework`
will result in a separate evaluation for the project. This means that
almost _every_ project in a solution will be evaluated at least
twice--once directly, with no special global properties, and once for
each unique `ReferringTargetFramework` that refers to it. This can be
particularly onerous for C++ projects (.vcxproj), where evaluation time
is nontrivial and doubling it is noticeable. That was mitigated by
special-casing vcxproj in #1278, but that is inelegant and doesn't
address the root cause.

This change alters the calling pattern for `ProjectReference`s. Instead
of asking all references for the correct TF, it breaks the problem into
multiple phases:

* Ask all references whether they cross-target, and split into
cross-targeting and non-cross-targeting lists.
* Perform the `GetTargetFrameworkProperties` query only on projects that
reported that they cross-targeted.
* Update (in place) the existing `_MSBuildProjectReferenceExistent` item
to pass or undefine the relevant properties based on whether the
reference cross-targets (and what TF is the best match).

Fixes #1276 (when combined with an Sdk fix, forthcoming)

### DIFF
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1542,14 +1542,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove);$(PropertiesToRemoveWhenQueryingForDesiredProperties)"
         Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' != 'true'">
 
-      <Output TaskParameter="TargetOutputs" ItemName="_MSBuildProjectReferenceCrossTargetingResults" />
+      <Output TaskParameter="TargetOutputs" ItemName="_MSBuildProjectReferenceWithDesiredPropertiesResults" />
     </MSBuild>
 
     <ItemGroup>
-      <_MSBuildProjectReferenceCrossTargeting Include="@(_MSBuildProjectReferenceCrossTargetingResults->'%(OriginalItemSpec)')"
-                                              Condition="'%(_MSBuildProjectReferenceCrossTargetingResults.Identity)' == 'true'" />
+      <_MSBuildProjectReferenceWithDesiredProperties Include="@(_MSBuildProjectReferenceWithDesiredPropertiesResults->'%(OriginalItemSpec)')"
+                                              Condition="'%(_MSBuildProjectReferenceWithDesiredPropertiesResults.Identity)' == 'true'" />
       <_MSBuildProjectReferenceNoCrossTargeting Include="@(_MSBuildProjectReferenceExistent)"
-                                                Exclude="@(_MSBuildProjectReferenceCrossTargeting)" />
+                                                Exclude="@(_MSBuildProjectReferenceWithDesiredProperties)" />
 
     </ItemGroup>
   </Target>
@@ -1581,13 +1581,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <MSBuild
-        Projects="@(_MSBuildProjectReferenceCrossTargeting)"
+        Projects="@(_MSBuildProjectReferenceWithDesiredProperties)"
         Targets="GetReferenceProperties"
         BuildInParallel="$(BuildInParallel)"
-        Properties="%(_MSBuildProjectReferenceCrossTargeting.SetConfiguration); %(_MSBuildProjectReferenceCrossTargeting.SetPlatform); ReferringTargetFramework=$(ReferringTargetFrameworkForProjectReferences)"
+        Properties="%(_MSBuildProjectReferenceWithDesiredProperties.SetConfiguration); %(_MSBuildProjectReferenceWithDesiredProperties.SetPlatform); ReferringTargetFramework=$(ReferringTargetFrameworkForProjectReferences)"
         ContinueOnError="!$(BuildingProject)"
-        RemoveProperties="%(_MSBuildProjectReferenceCrossTargeting.GlobalPropertiesToRemove);$(PropertiesToRemoveWhenQueryingForDesiredProperties)"
-        Condition="'%(_MSBuildProjectReferenceCrossTargeting.SkipGetTargetFrameworkProperties)' != 'true'">
+        RemoveProperties="%(_MSBuildProjectReferenceWithDesiredProperties.GlobalPropertiesToRemove);$(PropertiesToRemoveWhenQueryingForDesiredProperties)"
+        Condition="'%(_MSBuildProjectReferenceWithDesiredProperties.SkipGetTargetFrameworkProperties)' != 'true'">
 
       <Output TaskParameter="TargetOutputs" ItemName="_ProjectReferenceTargetFrameworkProperties" />
     </MSBuild>
--- a/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
@@ -1499,33 +1499,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <!--
     ====================================================================================
-                                        _GetProjectReferenceTargetFrameworkProperties
+                                        _GetProjectReferencesThatCrossTarget
 
-    Builds the GetTargetFrameworkProperties target of all existent project references,
-    passing $(TargetFrameworkMoniker) as $(ReferringTargetFramework) and sets the
-    SetTargetFramework metadata of the project reference to the value that is returned.
+    Builds the IsCrossTargetingProject target of all existent project references,
+    removing any TargetFramework and RuntimeIdentifier properties that are set as
+    global properties in this project.
 
-    This allows a cross-targeting project to select how it should be configured to
-    build against the most appropriate target for the referring target framework.
+    This allows a cross-targeting project to indicate that it should be asked
+    for a compatible TargetFramework before building.
 
     ======================================================================================
   -->
-  <Target Name="_GetProjectReferenceTargetFrameworkProperties"
-          Outputs="%(_MSBuildProjectReferenceExistent.Identity)">
-    <!--
-      Honor SkipGetTargetFrameworkProperties=true metadata on project references
-      to mean that the project reference is known not to target multiple frameworks
-      and the mechanism here for selecting the best one can be skipped as an optimization.
-
-      We give this treatment to .vcxproj by default since no .vcxproj can target more
-      than one framework.
-   -->
-   <ItemGroup>
-      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and '%(Extension)' == '.vcxproj'">
-        <SkipGetTargetFrameworkProperties>true</SkipGetTargetFrameworkProperties>
-      </_MSBuildProjectReferenceExistent>
-   </ItemGroup>
-
+  <Target Name="_GetProjectReferencesThatCrossTarget">
     <!--
        Allow project references to specify which target framework properties to set and their values
        without consulting the referenced project. This has two use cases:
@@ -1544,6 +1529,47 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
 
     <!--
+      Honor SkipGetTargetFrameworkProperties=true metadata on project references
+      to mean that the project reference is known not to target multiple frameworks
+      and the mechanism here for selecting the best one can be skipped as an optimization.
+    -->
+    <MSBuild
+        Projects="@(_MSBuildProjectReferenceExistent)"
+        Targets="IsCrossTargetingProject"
+        BuildInParallel="$(BuildInParallel)"
+        Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform)"
+        ContinueOnError="!$(BuildingProject)"
+        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove);TargetFramework;RuntimeIdentifier"
+        Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' != 'true'">
+
+      <Output TaskParameter="TargetOutputs" ItemName="_MSBuildProjectReferenceCrossTargetingResults" />
+    </MSBuild>
+
+    <ItemGroup>
+      <_MSBuildProjectReferenceCrossTargeting Include="@(_MSBuildProjectReferenceCrossTargetingResults->'%(OriginalItemSpec)')"
+                                              Condition="'%(_MSBuildProjectReferenceCrossTargetingResults.Identity)' == 'true'" />
+      <_MSBuildProjectReferenceNoCrossTargeting Include="@(_MSBuildProjectReferenceExistent)"
+                                                Exclude="@(_MSBuildProjectReferenceCrossTargeting)" />
+
+    </ItemGroup>
+  </Target>
+
+  <!--
+    ====================================================================================
+                                        _GetProjectReferenceTargetFrameworkProperties
+
+    Builds the GetTargetFrameworkProperties target of all existent project references,
+    passing $(TargetFrameworkMoniker) as $(ReferringTargetFramework) and sets the
+    SetTargetFramework metadata of the project reference to the value that is returned.
+
+    This allows a cross-targeting project to select how it should be configured to
+    build against the most appropriate target for the referring target framework.
+
+    ======================================================================================
+  -->
+  <Target Name="_GetProjectReferenceTargetFrameworkProperties"
+          DependsOnTargets="_GetProjectReferencesThatCrossTarget">
+    <!--
       Select the moniker to send to each project reference  if not already set. NugetTargetMoniker (NTM) is preferred by default over 
       TargetFrameworkMoniker (TFM) because it is required to disambiguate the UWP case where TFM is fixed at .NETCore,Version=v5.0 and 
       has floating NTM=UAP,Version=vX.Y.Z. However, in other cases (classic PCLs), NTM contains multiple values and that will cause the MSBuild
@@ -1555,7 +1581,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <MSBuild
-        Projects="%(_MSBuildProjectReferenceExistent.Identity)"
+        Projects="@(_MSBuildProjectReferenceCrossTargeting)"
         Targets="GetTargetFrameworkProperties"
         BuildInParallel="$(BuildInParallel)"
         Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); ReferringTargetFramework=$(ReferringTargetFrameworkForProjectReferences)"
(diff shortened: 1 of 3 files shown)


rainersigwald:
I feel like I had a more elegant solution to the problem "how do I ensure that TF and RID get removed from global properties in non-cross-targeting projects" in my previous attempt (now lost due to my foolishness/angering of the Bitlocker gremlins).

Comments on general approach eagerly desired, as well as implementation improvements.

Especially interested in feedback from @nguerrera.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1605,6 +1633,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
   <!--
     ============================================================
+                                         IsCrossTargetingProject
+
+    Overrridden by cross-targeting projects to return "true" if
+    the project should have its GetTargetFrameworkProperties
+    target called with a ReferringTargetFramework. If "false",
+    the single output of the project is used.
+  -->
+  <Target Name="IsCrossTargetingProject" Returns="false" />
```
nguerrera: Does the Returns="true" version need to go in the SDK or should it be in Microsoft.Common.CrossTargeting.targets?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1555,38 +1581,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <UndefineProperties>@(_MSBuildProjectReferenceExistent->'%(UndefineProperties)');TargetFramework</UndefineProperties>
       </_MSBuildProjectReferenceExistent>
-    </ItemGroup>
 
-    <ItemGroup>
-      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.Identity)' == '%(Identity)' and '$(_ProjectReferenceTargetFrameworkProperties)' != ''">
-        <UndefineProperties Condition="$(_ProjectReferenceTargetFrameworkProperties.Contains(`ProjectIsRidAgnostic=true`))">%(_MSBuildProjectReferenceExistent.UndefineProperties);RuntimeIdentifier;ProjectIsRidAgnostic</UndefineProperties>
-        <!-- Unconditionally remove the property that was set as a marker to indicate that for this call we should remove RuntimeIdentifier -->
-        <UndefineProperties Condition="!$(_ProjectReferenceTargetFrameworkProperties.Contains(`ProjectIsRidAgnostic=true`))">%(_MSBuildProjectReferenceExistent.UndefineProperties);ProjectIsRidAgnostic</UndefineProperties>
+      <!-- If the project has only one RID, assume it's compatible with the current project and don't pass this one along. -->
```
nguerrera: "If the project has only one RID" doesn't match IsRidAgnostic == true, does it? I thought that is the case where the project has no mention of RIDs. cc @piotrpMSFT

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1555,38 +1581,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     </MSBuild>
 
     <ItemGroup>
-      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.Identity)' == '%(Identity)' and '$(_ProjectReferenceTargetFrameworkProperties)' != ''">
-        <SetTargetFramework>$(_ProjectReferenceTargetFrameworkProperties)</SetTargetFramework>
+      <!-- Build an item that has Identity matching _MSBuildProjectReferenceExistent and metadata for properties to set. -->
+      <_ProjectReferencesWithTargetFrameworkProperties Include="@(_ProjectReferenceTargetFrameworkProperties->'%(OriginalItemSpec)')" />
+      <_ProjectReferencesWithTargetFrameworkProperties Include="@(_MSBuildProjectReferenceNoCrossTargeting)">
+        <HasSingleTargetFramework>true</HasSingleTargetFramework>
+        <IsRidAgnostic>true</IsRidAgnostic>
```
nguerrera: I don't believe IsRidAgnostic=true holds for any non-cross-targeting project. 

If I have two single-targeted projects with project A having RIDs and proejct B having none, with A referencing B, then `dotnet build -r some-rid A.csproj` still needs to remove the global RID property from B's build.

rainersigwald: You're right, I broke that case. That should be fixed by the changes I'm making for https://github.com/dotnet/sdk/pull/993#discussion_r106276038.

rainersigwald: So thinking about this I think the question we want to ask isn't really "IsCrossTargetingProject?" but more like "ShouldCallGetMatchingProperties?". Then I can move that back into the sdk (from common.crosstargeting) and teach it to return "yes" in the multi-rid-but-single-tfm situation.

rainersigwald: Implemented this in the latest push here and in https://github.com/dotnet/sdk/pull/993.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1555,43 +1581,45 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <MSBuild
-        Projects="%(_MSBuildProjectReferenceExistent.Identity)"
-        Targets="GetTargetFrameworkProperties"
+        Projects="@(_MSBuildProjectReferenceCrossTargeting)"
+        Targets="GetReferenceProperties"
         BuildInParallel="$(BuildInParallel)"
         Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); ReferringTargetFramework=$(ReferringTargetFrameworkForProjectReferences)"
```
rainersigwald: Self, these metadata need to refer to the projects that are actually being built!

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1535,15 +1520,56 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      to mean that the project reference is known not to target multiple frameworks
+      and the mechanism here for selecting the best one can be skipped as an optimization.
+    -->
+    <MSBuild
+        Projects="@(_MSBuildProjectReferenceExistent)"
+        Targets="ShouldQueryForProperties"
+        BuildInParallel="$(BuildInParallel)"
+        Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform)"
+        ContinueOnError="!$(BuildingProject)"
+        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove);TargetFramework;RuntimeIdentifier"
```
nguerrera: Oh, I see that we do remove TargetFramework and RuntimeIdentifier from this evaluation. Nevermind comment on SDK side. 

Is this considered OK perf-wise because the common case is that neither are specified as global properties?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1535,15 +1520,56 @@ Copyright (C) Microsoft Corporation. All rights reserved.
           there is also a .NETFramework implementation.
 
        2. As an escape hatch for cases where the compatibility check performed by 
-          GetTargetFrameworkProperties is faulty.
+          GetReferenceProperties is faulty.
     -->
     <ItemGroup>
       <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SetTargetFramework)' != ''">
-        <SkipGetTargetFrameworkProperties>true</SkipGetTargetFrameworkProperties>
+        <SkipGetReferenceProperties>true</SkipGetReferenceProperties>
```
nguerrera: I'm concerned about breaking change here.

nguerrera: I see that they are still called TargetFrameworkProperties in places. And still go to %(SetTargetFramework). I think we should rename all or none, and lean towards none for compat.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: That's my operating theory, yes. Additionally, if the common case is that references single-target, then that will be the evaluation we want to use later, so this doesn't come at an additional cost (just moves it around a bit).

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Good catch. My find-and-replace was too broad.

rainersigwald:
## Perf improvements

These are measured on a set of projects that the C++ team has that mirror the structure of the customer project that originally caused us to put in the .vcxproj workaround. Timings aren't super scientific since this is on my dev box and N=1.

### RTW targets
```
Project Performance Summary:
        0 ms  S:\work\CppMixed\FB\FB.vcxproj             1 calls
                  0 ms  GetTargetPath                              1 calls
        0 ms  S:\work\CppMixed\Ut\Ut.vcxproj             1 calls
                  0 ms  GetTargetPath                              1 calls
        0 ms  S:\work\CppMixed\ACa\ACa.vcxproj           1 calls
                  0 ms  GetTargetPath                              1 calls
        1 ms  S:\work\CppMixed\Mt\Mt.vcxproj             1 calls
                  1 ms  GetTargetPath                              1 calls
      299 ms  S:\work\CppMixed\A\A.vcxproj               1 calls
                299 ms  _PerfIntellisenseInfo                      1 calls

Target Performance Summary:
        0 ms  BeforeResolveReferences                    1 calls
        0 ms  GetInstalledSDKLocations                   1 calls
        0 ms  FixupReferences                            1 calls
        0 ms  GetGeneratedFiles                          1 calls
        0 ms  AfterResolveReferences                     1 calls
        0 ms  FixupCLCompileOptions                      1 calls
        0 ms  GetTargetPath                              4 calls
        0 ms  GetFrameworkPaths                          1 calls
        0 ms  ResolveReferences                          1 calls
        0 ms  ComputeMIDLGeneratedCompileInputs          1 calls
        0 ms  InitGetClCommandLines                      1 calls
        0 ms  ImplicitlyExpandDesignTimeFacades          1 calls
        0 ms  GetProjectDirectories                      1 calls
        0 ms  ResolveSDKReferences                       1 calls
        0 ms  PrepareProjectReferences                   1 calls
        0 ms  _CheckForInvalidConfigurationAndPlatform   1 calls
        0 ms  ExpandSDKReferences                        1 calls
        0 ms  _RemoveNameMetadataFromProjectReferenceItems   1 calls
        0 ms  GetResolvedWinMD                           5 calls
        0 ms  _PerfIntellisenseInfo                      1 calls
        1 ms  ComputeReferenceCLInput                    1 calls
        1 ms  ComputeCLInputPDBName                      1 calls
        1 ms  _SplitProjectReferencesByFileExistence     1 calls
        2 ms  _GetProjectReferenceTargetFrameworkProperties   4 calls
        2 ms  _PrepareForReferenceResolution             5 calls
        3 ms  PrepareForBuild                            1 calls
        5 ms  GetReferenceAssemblyPaths                  1 calls
       10 ms  AssignProjectConfiguration                 1 calls
       18 ms  ResolveAssemblyReferences                  1 calls
       19 ms  GetProjectReferencesInfo                   1 calls
       50 ms  GetClCommandLines                          1 calls
      185 ms  ResolveProjectReferences                   1 calls

Task Performance Summary:
        0 ms  MakeDir                                    2 calls
        1 ms  ResolveNonMSBuildProjectOutput             1 calls
        3 ms  AssignProjectConfiguration                 1 calls
        5 ms  GetReferenceAssemblyPaths                  1 calls
       17 ms  ResolveAssemblyReference                   1 calls
       45 ms  CLCommandLine                              2 calls
      153 ms  MSBuild                                    1 calls

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.44
```

### RTW minus the .vcxproj workaround
```
Project Performance Summary:
        0 ms  S:\work\CppMixed\ACa\ACa.vcxproj           2 calls
                  0 ms  GetTargetPath                              1 calls
                  0 ms  GetTargetFrameworkProperties               1 calls
        1 ms  S:\work\CppMixed\Ut\Ut.vcxproj             2 calls
                  1 ms  GetTargetPath                              1 calls
                  0 ms  GetTargetFrameworkProperties               1 calls
        1 ms  S:\work\CppMixed\Mt\Mt.vcxproj             2 calls
                  1 ms  GetTargetPath                              1 calls
                  0 ms  GetTargetFrameworkProperties               1 calls
        1 ms  S:\work\CppMixed\FB\FB.vcxproj             2 calls
                  1 ms  GetTargetPath                              1 calls
                  0 ms  GetTargetFrameworkProperties               1 calls
      430 ms  S:\work\CppMixed\A\A.vcxproj               1 calls
                430 ms  _PerfIntellisenseInfo                      1 calls

Target Performance Summary:
        0 ms  BeforeResolveReferences                    1 calls
        0 ms  FixupReferences                            1 calls
        0 ms  GetGeneratedFiles                          1 calls
        0 ms  AfterResolveReferences                     1 calls
        0 ms  FixupCLCompileOptions                      1 calls
        0 ms  GetTargetPath                              4 calls
        0 ms  ResolveReferences                          1 calls
        0 ms  ComputeMIDLGeneratedCompileInputs          1 calls
        0 ms  InitGetClCommandLines                      1 calls
        0 ms  ImplicitlyExpandDesignTimeFacades          1 calls
        0 ms  GetProjectDirectories                      1 calls
        0 ms  GetTargetFrameworkProperties               4 calls
        0 ms  ResolveSDKReferences                       1 calls
        0 ms  GetProjectReferencesInfo                   1 calls
        0 ms  PrepareProjectReferences                   1 calls
        0 ms  ComputeCLInputPDBName                      1 calls
        0 ms  GetResolvedWinMD                           5 calls
        0 ms  ExpandSDKReferences                        1 calls
        0 ms  _PerfIntellisenseInfo                      1 calls
        1 ms  _RemoveNameMetadataFromProjectReferenceItems   1 calls
        1 ms  _SplitProjectReferencesByFileExistence     1 calls
        1 ms  GetInstalledSDKLocations                   1 calls
        1 ms  GetFrameworkPaths                          1 calls
        1 ms  _CheckForInvalidConfigurationAndPlatform   1 calls
        2 ms  PrepareForBuild                            1 calls
        2 ms  ComputeReferenceCLInput                    1 calls
        3 ms  _PrepareForReferenceResolution             5 calls
        5 ms  GetReferenceAssemblyPaths                  1 calls
       11 ms  AssignProjectConfiguration                 1 calls
       18 ms  ResolveAssemblyReferences                  1 calls
       50 ms  GetClCommandLines                          1 calls
      148 ms  _GetProjectReferenceTargetFrameworkProperties   4 calls
      184 ms  ResolveProjectReferences                   1 calls

Task Performance Summary:
        0 ms  MakeDir                                    2 calls
        1 ms  ResolveNonMSBuildProjectOutput             1 calls
        4 ms  AssignProjectConfiguration                 1 calls
        5 ms  GetReferenceAssemblyPaths                  1 calls
       18 ms  ResolveAssemblyReference                   1 calls
       44 ms  CLCommandLine                              2 calls
      295 ms  MSBuild                                    5 calls

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.57
```

### This PR
```
Project Performance Summary:
        0 ms  S:\work\CppMixed\Mt\Mt.vcxproj             2 calls
                  0 ms  GetTargetPath                              1 calls
                  0 ms  ShouldQueryForProperties                   1 calls
        1 ms  S:\work\CppMixed\ACa\ACa.vcxproj           2 calls
                  1 ms  GetTargetPath                              1 calls
                  0 ms  ShouldQueryForProperties                   1 calls
        1 ms  S:\work\CppMixed\FB\FB.vcxproj             2 calls
                  1 ms  GetTargetPath                              1 calls
                  0 ms  ShouldQueryForProperties                   1 calls
        1 ms  S:\work\CppMixed\Ut\Ut.vcxproj             2 calls
                  1 ms  GetTargetPath                              1 calls
                  0 ms  ShouldQueryForProperties                   1 calls
      307 ms  S:\work\CppMixed\A\A.vcxproj               1 calls
                307 ms  _PerfIntellisenseInfo                      1 calls

Target Performance Summary:
        0 ms  BeforeResolveReferences                    1 calls
        0 ms  FixupReferences                            1 calls
        0 ms  GetGeneratedFiles                          1 calls
        0 ms  AfterResolveReferences                     1 calls
        0 ms  FixupCLCompileOptions                      1 calls
        0 ms  GetFrameworkPaths                          1 calls
        0 ms  ResolveReferences                          1 calls
        0 ms  _CheckForInvalidConfigurationAndPlatform   1 calls
        0 ms  InitGetClCommandLines                      1 calls
        0 ms  GetInstalledSDKLocations                   1 calls
        0 ms  ShouldQueryForProperties                   4 calls
        0 ms  ResolveSDKReferences                       1 calls
        0 ms  GetProjectReferencesInfo                   1 calls
        0 ms  ComputeCLInputPDBName                      1 calls
        0 ms  ImplicitlyExpandDesignTimeFacades          1 calls
        0 ms  _PerfIntellisenseInfo                      1 calls
        0 ms  _RemoveNameMetadataFromProjectReferenceItems   1 calls
        1 ms  ExpandSDKReferences                        1 calls
        1 ms  PrepareProjectReferences                   1 calls
        1 ms  ComputeMIDLGeneratedCompileInputs          1 calls
        1 ms  GetProjectDirectories                      1 calls
        1 ms  GetTargetPath                              4 calls
        2 ms  _SplitProjectReferencesByFileExistence     1 calls
        2 ms  PrepareForBuild                            1 calls
        2 ms  GetResolvedWinMD                           5 calls
        3 ms  _PrepareForReferenceResolution             5 calls
        5 ms  GetReferenceAssemblyPaths                  1 calls
        5 ms  _GetProjectReferenceTargetFrameworkProperties   1 calls
       10 ms  AssignProjectConfiguration                 1 calls
       17 ms  ResolveAssemblyReferences                  1 calls
       21 ms  ComputeReferenceCLInput                    1 calls
       45 ms  GetClCommandLines                          1 calls
       48 ms  ResolveProjectReferences                   1 calls
      142 ms  _GetProjectReferencesThatNeedProperties    1 calls

Task Performance Summary:
        0 ms  ResolveNonMSBuildProjectOutput             1 calls
        0 ms  MakeDir                                    2 calls
        3 ms  AssignProjectConfiguration                 1 calls
        4 ms  GetReferenceAssemblyPaths                  1 calls
       17 ms  ResolveAssemblyReference                   1 calls
       40 ms  CLCommandLine                              2 calls
      152 ms  MSBuild                                    3 calls

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.45
```

## Results
. . . so there are more calls to the MSBuild task than RTW, but total time and MSBuild task times are similar, and both are better than when dropping the special-case for vcxproj (as we had it initially).

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1555,43 +1581,45 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <MSBuild
-        Projects="%(_MSBuildProjectReferenceExistent.Identity)"
-        Targets="GetTargetFrameworkProperties"
+        Projects="@(_MSBuildProjectReferenceCrossTargeting)"
```
cdmihai: Could non cross-targetting projects ever have special props for their target framework?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1555,43 +1581,45 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        Targets="GetReferenceProperties"
```
cdmihai: Also rename target to `_GetProjectReferenceProperties`?

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
nguerrera: It would be nice as escape hatch if the set of properties removed from ShouldQueryForProperties were configurable and defaulted TargetFramework;RuntimeIdentifier. I'm thinking that another global property might come along and bite us later on and an escape hatch here could help. This can be done/tracked separately though.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1544,6 +1529,47 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform)"
+        ContinueOnError="!$(BuildingProject)"
+        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove);TargetFramework;RuntimeIdentifier"
+        Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' != 'true'">
+
+      <Output TaskParameter="TargetOutputs" ItemName="_MSBuildProjectReferenceCrossTargetingResults" />
+    </MSBuild>
+
+    <ItemGroup>
+      <_MSBuildProjectReferenceCrossTargeting Include="@(_MSBuildProjectReferenceCrossTargetingResults->'%(OriginalItemSpec)')"
```
nguerrera: s/_MSBuildProjectReferenceCrossTargeting/_MSBuildProjectReferenceNeedingProperties/ (or something like that) as this is no longer coupled to cross-targeting cases alone.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: That's a very good idea.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
nguerrera: They can. Looks like a missing rename after discussing exactly those cases.

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
nguerrera: Indeed. ;)

src/XMakeTasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: It's "public" API surface so I don't like the initial underscore. I'm not sure I prefer `GetProjectReferenceProperties`. Maybe `GetDesiredProperties`?

rainersigwald:
MSBuild triage: we're not going to pursue this for now.

PR #1868, opened 2017-03-15
Enable binary logger on bootstrapped build.

AndyGerlicher:


AndyGerlicher:
@jeffkl also suggested we reply the log after (`MSBuild %MSBUILDBINLOGPATH% /v:m` or something). A good idea but might be confusing so I didn't do that yet.

PR #1871, opened 2017-03-16
A better GetAllGlobs

### LINKED ISSUE #1795: GetAllGlobs results should incorporate an IMsBuildGlob and include information from Remove elements
`Project.GetAllGlobs` returns a `GlobResult` for each glob introduced by a `ProjectItemElement`. `GlobResult.Excludes` only contains the excludes of the item, whereas CPS needs it to also include all the Remove elements that affect that item.

`GlobResult` should also include an `IMsBuildGlob` that could be used to test whether arbitrary strings would be matched by the include item element representing that `GlobResult`. Under the covers it will be an MsBuildGlobWithGaps, where the gaps incorporate information from the exclude and relevant remove elements.

In addition, there should be one glob result aggregating the information per item project element.

cdmihai:
Resolves #1795 

This is based on another PR: https://github.com/Microsoft/msbuild/pull/1859

### DIFF
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -500,10 +500,12 @@ namespace Microsoft.Build.Evaluation
 {
     public partial class GlobResult
     {
-        public GlobResult(Microsoft.Build.Construction.ProjectItemElement itemElement, string glob, System.Collections.Generic.IEnumerable<string> excludes) { }
+        internal GlobResult() { }
         public System.Collections.Generic.IEnumerable<string> Excludes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
-        public string Glob { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public System.Collections.Generic.IEnumerable<string> IncludeGlobs { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public Microsoft.Build.Construction.ProjectItemElement ItemElement { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public Microsoft.Build.Globbing.IMSBuildGlob MsBuildGlob { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public System.Collections.Generic.IEnumerable<string> Removes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
     }
     [System.FlagsAttribute]
     public enum NewProjectFileOptions
--- a/src/Build/Definition/Project.cs
+++ b/src/Build/Definition/Project.cs
@@ -1211,7 +1211,7 @@ namespace Microsoft.Build.Evaluation
             {
                 var excludeItemspec = new EvaluationItemSpec(itemElement.Exclude, _data.Expander, itemElement.ExcludeLocation);
 
-                excludeFragmentStrings = ConvertItemSpecToFragmentStrings(excludeItemspec).ToImmutableHashSet();
+                excludeFragmentStrings = excludeItemspec.FlattenFragmentsAsStrings().ToImmutableHashSet();
                 excludeGlob = excludeItemspec.ToMSBuildGlob();
             }
 
@@ -1278,34 +1278,7 @@ namespace Microsoft.Build.Evaluation
             {
                 cumulativeRemoveElementData.FragmentStrings.Add(removeFragment);
             }
-        }
 
-        /// <summary>
-        /// Given an ItemSpec, this method returns all the fragment strings that represent it.
-        /// "1;*;2;@(foo)" gets returned as ["1", "2", "*", "a", "b"], given that @(foo)=["a", "b"]
-        /// 
-        /// Order is not preserved.
-        /// </summary>
-        /// <returns></returns>
-        private IEnumerable<string> ConvertItemSpecToFragmentStrings(EvaluationItemSpec spec)
-        {
-            foreach (var valueString in spec.Fragments.OfType<ValueFragment>().Select(v => v.ItemSpecFragment))
-            {
-                yield return valueString;
-            }
-
-            foreach (var globString in spec.Fragments.OfType<GlobFragment>().Select(g => g.ItemSpecFragment))
-            {
-                yield return globString;
-            }
-
-            foreach (
-                var referencedItemString in
-                spec.Fragments.OfType<EvaluationItemExpressionFragment>().SelectMany(f => f.ReferencedItems).Select(v => v.ItemSpecFragment)
-            )
-            {
-                yield return referencedItemString;
-            }
         }
 
         /// <summary>
--- a/src/Build/Definition/Project.cs
+++ b/src/Build/Definition/Project.cs
@@ -1097,12 +1097,12 @@ namespace Microsoft.Build.Evaluation
         ///]
         /// </example>
         /// <remarks>
-        /// <see cref="GlobResult.MsBuildGlob"/> is a <see cref="IMsBuildGlob"/> that combines all globs in the include element and ignores
+        /// <see cref="GlobResult.MsBuildGlob"/> is a <see cref="IMSBuildGlob"/> that combines all globs in the include element and ignores
         /// all the fragments in the exclude attribute and all the fragments in all Remove elements that apply to the include element.
         /// 
         /// Users can construct a composite glob that incorporates all the globs in the Project:
         /// <code>
-        /// var uberGlob = new CompositeGlob(project.GetAllGlobs().Select(r => r.MsBuildGlob).ToArray());
+        /// var uberGlob = new CompositeGlob(project.GetAllGlobs().Select(r => r.MSBuildGlob).ToArray());
         /// uberGlob.IsMatch("foo.cs");
         /// </code>
         /// 
@@ -1131,7 +1131,7 @@ namespace Microsoft.Build.Evaluation
 
         private struct CumulatedRemoveElementData
         {
-            public ImmutableList<IMsBuildGlob>.Builder Globs { get; set; }
+            public ImmutableList<IMSBuildGlob>.Builder Globs { get; set; }
             public ImmutableHashSet<string>.Builder FragmentStrings { get; set; }
         }
 
@@ -1191,7 +1191,7 @@ namespace Microsoft.Build.Evaluation
             var includeGlob = new CompositeGlob(includeGlobFragments.Select(f => f.ToMsBuildGlob()).ToImmutableArray());
 
             IEnumerable<string> excludeFragmentStrings = Enumerable.Empty<string>();
-            IMsBuildGlob excludeGlob = null;
+            IMSBuildGlob excludeGlob = null;
 
             if (!string.IsNullOrEmpty(itemElement.Exclude))
             {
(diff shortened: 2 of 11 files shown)


src/Build/Definition/Project.cs
```diff
@@ -1117,32 +1129,161 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
-                .ToList();
+            if (projectItemElements.Count == 0)
+            {
+                return new List<GlobResult>();
+            }
+
+            // scan the project elements in reverse order and build globbing information for each include element
+            // based on the fact that relevant removes for a particular include element (element A) consist of:
+            // - all the removes seen by the next include statement of A's type (element B which appears after A in file order)
+            // - new removes between A and B
```
rainersigwald: Between A and B, or just the one for A?

src/Build/Definition/Project.cs
```diff
@@ -1117,32 +1129,161 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
 
-            var excludeSet =
-                new Lazy<IEnumerable<string>>(
-                    () =>
-                        excludeItemspec.Fragments.Where(
-                            // take out item references, we can't reason about them
-                            f => !(f is ItemExpressionFragment<ProjectProperty, ProjectItem>))
-                            .Select(f => f.ItemSpecFragment)
-                            .ToImmutableHashSet());
+            var includeGlobFragments = includeItemspec.Fragments.Where(f => f is GlobFragment).ToImmutableArray();
```
rainersigwald: Does this need to be immutable? looks like you could get away with just an array, which might be faster?

src/Build/Definition/Project.cs
```diff
@@ -1117,32 +1129,161 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
+
+            cumulatedRemoveElementData.Globs.Add(removeGlob);
+
+            foreach (var removeFragment in removeSpecFragmentStrings)
+            {
+                cumulatedRemoveElementData.FragmentStrings.Add(removeFragment);
+            }
+        }
+
+        private IEnumerable<string> GetRelevantFragmentStringsForExcludesAndRemoves(EvaluationItemSpec spec)
```
rainersigwald: Needs documentation:
* Reorders the fragments
* What fragments does it not return?
* "Relevant" to what?

src/Build/Evaluation/ItemSpec.cs
```diff
@@ -186,6 +189,15 @@ public IEnumerable<ItemFragment> FragmentsMatchingItem(string itemToMatch, out i
...
             return result;
         }
+
+        /// <summary>
+        /// Return an MSBuildGlob that represents this ItemSpec.
+        /// </summary>
+        /// <returns></returns>
+        public IMSBuildGlob ToMSBuildGlob()
+        {
+            return new CompositeGlob(Fragments.Select(f => f.ToMSBuildGlob()).ToImmutableArray());
```
rainersigwald: This calling pattern makes me think a ctor should take an `IEnumerable` and immutable-ize it itself.

src/Build/Definition/Project.cs
```diff
@@ -1117,32 +1129,161 @@ public List<GlobResult> GetAllGlobs(string itemType)
             return GetAllGlobs(GetItemElementsByType(GetEvaluatedItemElements(), itemType));
         }
 
+        private struct CumulatedRemoveElementData
```
cdmihai: Cumulative. As the items are scanned, that struct grows with every new piece of information on Remove operations. I'll rename it.

src/Build/Definition/Project.cs
```diff
@@ -1117,32 +1129,161 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
+        {
+            public ImmutableList<IMSBuildGlob>.Builder Globs { get; set; }
```
cdmihai: True, I could encapsulate it and provide CumulateRemoveGlob / GetImmutableSnapshot type of methods. However, since it's a private implementation struct I did not consider it too much (If I could have defined structs inside methods I would have done that). I'll play around with it and see hot it looks encapsulated.

src/Build/Definition/Project.cs
cdmihai: Removes that apply to A but not to B. Spacially, these are placed between A's element and B's element.

```xml
1. <A Include="A"/>
2. <A Remove="..."/> // this remove applies to the A includes
3. <A Include="B"/>
4. <A Remove="..."/> // this remove applies to the A and B includes
5. <A Include="C"/>
6. <A Remove="..."/> // this remove applies to the A, B, and C includes

```

So A's applicable removes are composed of:
- all the removes seen by the next include statement of A's type (element B which appears after A in file order). In this example that's Removes 4 and 6.
- new removes between A and B. In this example that's Remove 2.

src/Build/Definition/Project.cs
cdmihai: I used this as documentation: https://blogs.msdn.microsoft.com/dotnet/2013/06/24/please-welcome-immutablearrayt/
ImmutableArray is backed up by an actual array, and it's a struct. Since there does not seem to be a obvious perf difference from the regular array, I went with the more expressive choice. Is there something I'm missing? :)

rainersigwald: Ah, I didn't know that was quite so nice. Leave it, then--I like communicating the intent through the type.

cdmihai:
@rainersigwald Implemented your feedback.

src/Build/Definition/Project.cs
```diff
@@ -1117,33 +1129,169 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
+        // represents cumulated remove information for a particular item type
+        private struct CumulativeRemoveElementData
+        {
+            private ImmutableList<IMSBuildGlob>.Builder _globs;
+            private ImmutableHashSet<string>.Builder _fragmentStrings;
+
+            public IEnumerable<IMSBuildGlob> Globs => _globs.ToImmutable();
+            public IEnumerable<string> FragmentStrings => _fragmentStrings.ToImmutable();
+
+            public static CumulativeRemoveElementData New()
```
rainersigwald: In .NET this pattern is usually named `.Create()`.

src/Build/Definition/Project.cs
```diff
@@ -1117,33 +1129,169 @@ public List<GlobResult> GetAllGlobs(string itemType)
             return GetAllGlobs(GetItemElementsByType(GetEvaluatedItemElements(), itemType));
         }
 
+        // represents cumulated remove information for a particular item type
+        private struct CumulativeRemoveElementData
+        {
+            private ImmutableList<IMSBuildGlob>.Builder _globs;
+            private ImmutableHashSet<string>.Builder _fragmentStrings;
+
+            public IEnumerable<IMSBuildGlob> Globs => _globs.ToImmutable();
```
rainersigwald: If you're returning an `IEnumerable` is there any point in making it immutable? It's already readonly.

src/Build/Definition/Project.cs
```diff
@@ -1117,33 +1129,169 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
+            public static CumulativeRemoveElementData New()
+            {
+                return new CumulativeRemoveElementData
+                {
+                    _globs = ImmutableList.CreateBuilder<IMSBuildGlob>(),
+                    _fragmentStrings = ImmutableHashSet.CreateBuilder<string>()
+                };
+            }
+
+            public void CumulateInformationFromRemoveItemSpec(EvaluationItemSpec removeSpec)
```
rainersigwald: Nit: `Accumulate`

src/Build/Definition/Project.cs
```diff
@@ -1117,33 +1129,169 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
+            // 1. <A Include="A"/>
+            // 2. <A Remove="..."/> // this remove applies to the A includes
+            // 3. <A Include="B"/>
+            // 4. <A Remove="..."/> // this remove applies to the A and B includes
+            // 5. <A Include="C"/>
+            // 6. <A Remove="..."/> // this remove applies to the A, B, and C includes
+            // So A's applicable removes are composed of:
+            // 
+            // The applicable removes for the element at position 1. are composed of:
+            // - all the removes seen by the next include statement of A's type (element B which appears after A in file order). In this example that's Removes 4 and 6.
```
rainersigwald: Ahhhhh I think I understand now. Can you rename the item itself from `A` to `I` in the example and comments? I was getting confused between `A`-the-item-type and `A`-the-included-item.

I'm also a bit confused by what you mean by `element B` here. Line 3 that included `B`?

src/Build/Definition/Project.cs
```diff
@@ -1117,33 +1129,169 @@ public List<GlobResult> GetAllGlobs(string itemType)
...
+            // - new removes between A and B. In this example that's Remove 2.
+
+            // use immutable lists because there will be a lot of structural sharing between includes which share increasing subsets of corresponding remove elements
+            // item type -> aggregated information about all removes seen so far for that item type
```
rainersigwald: Should this go elsewhere now? there's no immutable right here.

ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -500,10 +500,12 @@ namespace Microsoft.Build.Evaluation
 {
     public partial class GlobResult
     {
-        public GlobResult(Microsoft.Build.Construction.ProjectItemElement itemElement, string glob, System.Collections.Generic.IEnumerable<string> excludes) { }
+        internal GlobResult() { }
         public System.Collections.Generic.IEnumerable<string> Excludes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
-        public string Glob { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
cdmihai: I could unbreak it by:
- keeping the constructor public and overloading it
- mark Glob as deprecated and have it return the first glob. After this change, there is one GlobResult per item element with globs. Before, there was one GlobResult per glob fragment, so one item element could have produced multiple GlobResults.

As an alternative, I could keep the old code and have the new variants as GetAllGlobs2 and GlobResult2 :)

I searched on Github and Index and found no other usage other than CPS. So breaking it looks like a viable option, since everything else introduces baggage.

src/Build/Globbing/CompositeGlob.cs
```diff
@@ -21,17 +22,17 @@ public class CompositeGlob : IMSBuildGlob
...
+            Globs = globs.ToImmutableArray();
         }
 
         /// <summary>
         ///     Constructor
         /// </summary>
-        /// <param name="globs">children globs</param>
-        public CompositeGlob(params IMSBuildGlob[] globs) : this(globs.AsEnumerable())
+        /// <param name="globs">Children globs. Input gets shallow cloned</param>
+        public CompositeGlob(params IMSBuildGlob[] globs) : this(globs.ToImmutableArray())
```
cdmihai: None: https://source.dot.net/#System.Collections.Immutable/System/Collections/Immutable/ImmutableArray.cs,390

rainersigwald: Sweet, works for me!

src/Build/Definition/Project.cs
cdmihai: Sanity. Just in case someone casts it back to builder and mutates it :)

The operation itself is not that expensive, it traverses the backing tree and toggles an `_frozen` bool to true.

src/Build/Definition/Project.cs
cdmihai: xml element B. I'll update the comment

src/Build/Definition/Project.cs
cdmihai: I'd keep it as a comment on the reason why immutable types are really good here.

cdmihai:
@mfilippov / @mhutch FYI, this PR will introduce a breaking change in GetAllGlobs / GlobResult, in case you are using it.

mfilippov:
@cdmihai Thanks for information.

mhutch:
@cdmihai thanks for the heads up!

cc @slluis.

PR #1879, opened 2017-03-16
Remove ProjectConfigurationsInferredFromUsage capability from managed targets

RaulPerez1:
The legacy project system works without it (since it's not CPS based) and there are changes in the pipeline to use a different capability to defined configurations differently. On the new project system the capability is defined by the project system (as a default). Additionally the capability was also defined in the design time targets so there is some redundancy. Having the old capabilities here conflicts with that and requires a workaround to add a remove statement elsewhere in the targets.


AndyGerlicher:

Odd, not sure I've ever seen a failure in `LogReferenceAndRedistConflictBetweenSdks`.

rainersigwald:
That confused me too, and I'm putting some better diagnosability in there for the future.

PR #1883, opened 2017-03-17
Globbing APIs to master

cdmihai:
Same as #1859 but oriented to master.


cdmihai:
Pushed directly to master, closing

PR #1885, opened 2017-03-17
Capture test output to Xunit MockEngine

rainersigwald:
https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/_Windows_NT_Desktop_prtest/640/ had a mysterious failure in `LogReferenceAndRedistConflictBetweenSdks` which isn't known to be flaky. But the log just indicated that a task failed, which isn't helpful.

I added the ability to keep log information in xunit's style in MockEngine, switched its internal log capture to use StringBuilder instead of super allocatey concatenation all over the place, and opted the class that failed into the new logging.

### DIFF
--- a/src/Shared/UnitTests/MockEngine.cs
+++ b/src/Shared/UnitTests/MockEngine.cs
@@ -75,41 +75,41 @@ namespace Microsoft.Build.UnitTests
 
         public void LogErrorEvent(BuildErrorEventArgs eventArgs)
         {
+            string message = string.Empty;
+
             if (eventArgs.File != null && eventArgs.File.Length > 0)
             {
-                if (_logToConsole)
-                    Console.Write("{0}({1},{2}): ", eventArgs.File, eventArgs.LineNumber, eventArgs.ColumnNumber);
-                _log += String.Format("{0}({1},{2}): ", eventArgs.File, eventArgs.LineNumber, eventArgs.ColumnNumber);
+                message += String.Format("{0}({1},{2}): ", eventArgs.File, eventArgs.LineNumber, eventArgs.ColumnNumber);
             }
 
-            if (_logToConsole)
-                Console.Write("ERROR " + eventArgs.Code + ": ");
-            _log += "ERROR " + eventArgs.Code + ": ";
+            message += "ERROR " + eventArgs.Code + ": ";
             ++_errors;
 
+            message += eventArgs.Message;
+
             if (_logToConsole)
-                Console.WriteLine(eventArgs.Message);
-            _log += eventArgs.Message;
+                Console.WriteLine(message);
+            _log += message;
             _log += "\n";
         }
 
         public void LogWarningEvent(BuildWarningEventArgs eventArgs)
         {
+            string message = string.Empty;
+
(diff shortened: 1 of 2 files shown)


src/Shared/UnitTests/MockEngine.cs
```diff
@@ -184,8 +190,16 @@ public int ColumnNumberOfTaskNode
...
-            set { _log = value; }
-            get { return _log; }
+            set
+            {
+                if (!string.IsNullOrEmpty(value))
+                {
+                    throw new ArgumentException("Expected log setter to be used only to reset the log to empty.");
+                }
+
+                _log.Clear();
```
cdmihai: I like what you did here :)

rainersigwald: I tried to just delete it but there was one use :(

src/Shared/UnitTests/MockEngine.cs
```diff
@@ -399,8 +413,7 @@ internal void AssertLogContains(string contains)
         {
             if (_upperLog == null)
             {
-                _upperLog = _log;
-                _upperLog = _upperLog.ToUpperInvariant();
+                _upperLog = _log.ToString().ToUpperInvariant();
```
cdmihai: I guess it's fine to write and forget about _upperLog because the engine is only used once? If the same engine is passed to multiple tasks, then _upperLog can get stale

src/Tasks.UnitTests/GetSDKReference_Tests.cs
```diff
@@ -187,15 +188,19 @@ private static string CreateFakeSDKReferenceAssemblyDirectory2(out string sdkDir
...
+
         private readonly string _sdkDirectory;
         private readonly string _sdkDirectory2;
         private readonly MockEngine.GetStringDelegate _resourceDelegate;
         private readonly GetAssemblyName _getAssemblyName = GetAssemblyName;
         private readonly GetAssemblyRuntimeVersion _getAssemblyRuntimeVersion = GetImageRuntimeVersion;
         private readonly string _cacheDirectory = Path.Combine(Path.GetTempPath(), "GetSDKReferenceFiles");
 
-        public GetSDKReferenceFilesTestFixture(FakeSdkStructure fakeSdkStructure)
+        public GetSDKReferenceFilesTestFixture(FakeSdkStructure fakeSdkStructure, ITestOutputHelper output)
```
cdmihai: Who calls this constructor? I would have expected the callsites to get updated as well.

rainersigwald: The xunit runner calls it (finds it by reflecting over the assembly somehow), so there are no direct callers.

cdmihai: Uuu, fancy!

src/Shared/UnitTests/MockEngine.cs
rainersigwald: I'm going to leave this since I'm not making it worse, but yeah, that might bite us in the future :(

PR #1898, opened 2017-03-22
Check for BOM in file when detecting Encoding.

### LINKED ISSUE #1768: Modifying a project adds BOM at the start of the project
_From @davkean on February 28, 2017 11:38_

1. File -> New Project
2. Check Create new Git repository
3. Right-click on project and choose Edit
4. CTRL+S

Expected: For no changes to the project
Actual: The project has changes if you run git diff you see:

``` XML
-<Project Sdk="Microsoft.NET.Sdk">
+<EF><BB><BF><Project Sdk="Microsoft.NET.Sdk">

   <PropertyGroup>
     <OutputType>Exe</OutputType>
```

_Copied from original issue: dotnet/roslyn-project-system#1651_

AndyGerlicher:
The XmlReader does not currently consider the BOM when detecting the
Encoding. This change checks the Stream for the UTF8 BOM and if not
present use UTF8 w/o BOM Encoding.

Fixes #1768

### DIFF
--- a/src/Build.OM.UnitTests/Construction/ProjectFormatting_Tests.cs
+++ b/src/Build.OM.UnitTests/Construction/ProjectFormatting_Tests.cs
@@ -3,11 +3,9 @@ using Microsoft.Build.Evaluation;
 using Microsoft.Build.UnitTests;
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Threading.Tasks;
 using System.Xml;
 using Microsoft.Build.Shared;
 using Xunit;
@@ -604,6 +602,73 @@ namespace Microsoft.Build.Engine.OM.UnitTests.Construction
             string actual = writer.ToString();
 
             VerifyAssertLineByLine(expected, actual);
+        }
+
+        [Fact]
+        public void VerifyUtf8WithoutBomTreatedAsUtf8()
+        {
+            string expected = ObjectModelHelpers.CleanupFileContents(@"<Project>
+</Project>");
+
+            Project project = new Project(NewProjectFileOptions.None);
+            StringWriter writer = new EncodingStringWriter(new UTF8Encoding(encoderShouldEmitUTF8Identifier: true));
+            project.Save(writer);
+
+            string actual = writer.ToString();
+
+            VerifyAssertLineByLine(expected, actual);
+        }
+
+        [Fact]
+        public void ProjectFileWithBomContainsBomAfterSave()
+        {
+            string content = ObjectModelHelpers.CleanupFileContents(@"<Project />");
+
+            var file = FileUtilities.GetTemporaryFile(".proj");
+            try
+            {
+                File.WriteAllText(file, content, new UTF8Encoding(encoderShouldEmitUTF8Identifier: true));
+                Assert.True(EncodingUtilities.FileStartsWithPreamble(file));
+
+                // Load and manipulate/save the project
+                var project = new Project(ProjectRootElement.Open(file, ProjectCollection.GlobalProjectCollection));
+                project.AddItem("Compile", "Program.cs");
+                project.Save();
+
+                // Ensure the file was really saved and that the file still contains a BOM
+                Assert.Contains("<Compile Include=\"Program.cs\" />", File.ReadAllText(file));
+                Assert.True(EncodingUtilities.FileStartsWithPreamble(file));
+            }
+            finally
+            {
+                FileUtilities.DeleteDirectoryNoThrow(Path.GetDirectoryName(file), false);
+            }
+        }
+
+        [Fact]
+        public void ProjectFileWithoutBomDoesNotContainsBomAfterSave()
+        {
+            string content = ObjectModelHelpers.CleanupFileContents(@"<Project><Target Name=""Build""/></Project>");
+
+            var file = FileUtilities.GetTemporaryFile(".proj");
+            try
+            {
+                File.WriteAllText(file, content, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false));
+                Assert.False(EncodingUtilities.FileStartsWithPreamble(file));
+
+                // Load and manipulate/save the project
+                var project = new Project(ProjectRootElement.Open(file, ProjectCollection.GlobalProjectCollection));
+                project.AddItem("Compile", "Program.cs");
+                project.Save();
+
+                // Ensure the file was really saved and that the file still contains a BOM
+                Assert.Contains("<Compile Include=\"Program.cs\" />", File.ReadAllText(file));
+                Assert.False(EncodingUtilities.FileStartsWithPreamble(file));
+            }
+            finally
+            {
+                FileUtilities.DeleteDirectoryNoThrow(Path.GetDirectoryName(file), false);
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


AndyGerlicher:
Note: The first commit was enough to detect the "correct" encoding (UTF8 w/o BOM and with BOM). The 2nd commit was an audit of comparisons of the document encoding to the default (UTF8) that now failed because the encoding was not strictly equal. Without the 2nd change the `<xml>` declaration with encoding UTF8 started showing up in project files on save.

src/Build.OM.UnitTests/Construction/ProjectFormatting_Tests.cs
```diff
@@ -605,5 +603,86 @@ public void ChangeItemTypeWithXmlHeader()
...
         }
+
+        [Fact]
+        public void VerifyUtf8WithoutBomTreatedAsUtf8()
+        {
+            string expected = ObjectModelHelpers.CleanupFileContents(@"<Project>
+</Project>");
+
+            Project project = new Project(NewProjectFileOptions.None);
+            StringWriter writer = new EncodingStringWriter(new UTF8Encoding(true));
```
rainersigwald: can you give the param name for the `true` param? I can't remember which direction it points.

(everywhere)

src/Build.OM.UnitTests/Construction/ProjectFormatting_Tests.cs
```diff
@@ -605,5 +603,86 @@ public void ChangeItemTypeWithXmlHeader()
...
+                Assert.True(FileStartsWithBom(file));
+
+                // Load and manipulate/save the project
+                var project = new Project(ProjectRootElement.Open(file, ProjectCollection.GlobalProjectCollection));
+                project.AddItem("Compile", "Program.cs");
+                project.Save();
+
+                // Ensure the file was really saved and that the file still contains a BOM
+                Assert.True(File.GetLastWriteTime(file).Subtract(DateTime.Now).TotalSeconds < 2);
+                Assert.True(File.ReadAllText(file).Contains("<Compile Include=\"Program.cs\" />"));
```
rainersigwald: `Assert.Contains(File.ReadAllText(file), "<Compile Include=\"Program.cs\" />")` ?

I think that'd give a nicer error if it failed.

src/Build.OM.UnitTests/Construction/ProjectFormatting_Tests.cs
```diff
@@ -605,5 +603,86 @@ public void ChangeItemTypeWithXmlHeader()
...
+        }
+
+        private static bool FileStartsWithBom(string file)
+        {
+            using (var stream = new FileStream(file, FileMode.Open, FileAccess.Read, FileShare.Read))
+            {
+                var bom = new byte[3];
+                stream.Read(bom, 0, 3);
+                var preamble = new UTF8Encoding(true).GetPreamble();
+                return !preamble.Where((p, i) => p != bom[i]).Any();
```
rainersigwald: Normally I say to LINQify things but I think this would be clearer as a for loop :)

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -609,12 +609,9 @@ public Encoding Encoding
                 {
                     XmlDeclaration declaration = XmlDocument.FirstChild as XmlDeclaration;
 
-                    if (declaration != null)
+                    if (declaration?.Encoding.Length > 0)
```
rainersigwald: pretty please can you put these cleanups in a separate commit?

src/Shared/EncodingUtilities.cs
```diff
@@ -53,6 +57,54 @@ static internal Encoding CurrentSystemOemEncoding
...
+            var position = stream.Position;
+            if (stream.Position != 0)
+                stream.Seek(0, SeekOrigin.Begin);
+
+            var detectedPreamble = new byte[3];
+            stream.Read(detectedPreamble, 0, 3);
+            var utf8Preamble = Encoding.UTF8.GetPreamble();
+            stream.Seek(position, SeekOrigin.Begin);
+
+            return !utf8Preamble.Where((p, i) => p != detectedPreamble[i]).Any();
```
rainersigwald: ditto `for `loop

src/Build/Xml/XmlReaderExtension.cs
```diff
@@ -34,12 +37,22 @@ private XmlReaderExtension(string file)
...
                 var encodingAttribute = Reader.GetAttribute("encoding");
                 _encoding = !string.IsNullOrEmpty(encodingAttribute)
                     ? Encoding.GetEncoding(encodingAttribute)
                     : _encoding;
+
+                // If the encoding is UTF8 but there was not a BOM detected in the file then we need
+                // to override the detected Encoding with UTF8 and encoderShouldEmitUTF8Identifier
+                // set to false. This is to prevent the BOM from being added if we save the file
+                // later using the detected encoding.
+                if (Equals(_encoding, Encoding.UTF8) && !bom)
```
cdmihai: Why is UTF8 special? What if users use something else in their non-bom project file?

src/Shared/ProjectWriter.cs
```diff
@@ -118,8 +118,8 @@ internal void Initialize(XmlDocument project, XmlDeclaration projectRootElementD
             }
 
             // don't write an XML declaration unless the project already has one or has non-default encoding
-            _writeXmlDeclaration = (projectRootElementDeclaration != null) ||
-                                   (_documentEncoding != null && !Equals(_documentEncoding, Encoding.UTF8));
+            _writeXmlDeclaration = projectRootElementDeclaration != null ||
```
cdmihai: Unrelated to PR: should the first expression be `projectRootElementDeclaration == null`? a.k.a write the xml declaration if there is no existing one.

AndyGerlicher: No, if there was an xml declaration in the file we always write it (`true`). If there wasn't we only write it if the encoding is not UTF8.

src/Build/Construction/ProjectRootElement.cs
AndyGerlicher: Done!

src/Build/Xml/XmlReaderExtension.cs
AndyGerlicher: It's special in this scenario because 1) it's the default encoding by us and Visual Studio and 2) can be constructed in a way that is with and without a BOM. And when the reader detects encoding it doesn't make that distinction and always returns an Encoding that indicates it should emit the BOM.

`UTF32Encoding` and `UnicodeEncoding` also have this and other differences, but I'm not going to address those now.

davkean: Help me understand why you can't sniff the encoding off the streamreader which does the right thing here?

davkean: We talked offline and figured out why `StreamReader` couldn't be trusted in the case - the parameter that detects encoding is called `detectEncodingFromByteOrderMarks` _not_ `detectEncoding`. ie It will only change the "default" encoding passed to the constructor if the byte-order-mark is present, not if it isn't.

If Andy changed the call to StreamReader to pass s_utf8NoBom by default, then this will do the right thing and he can end up reading the Encoding from the StreamReader instead of checking the preamble himself.

PR #1903, opened 2017-03-23
Remove apparently dead code

zebmason:
Have run my static analysis tool on your codebase and identified regions of apparently dead code
Note could only get build working for Debug, Release and the Mono variants of that due to my set-up not working for NetCore or PortProgress
May have removed some code that is called indirectly, did identify some such and rolled by the tools changes.

Please feel free to merge whatever you think is significant.

My progress in developing the tool is noted at https://github.com/zebmason/loc8 and commercial release is currently delayed whilst an appropriate web store solution is being  determined.


zebmason:
Sorry something has messed up the line endings, please ignore for the moment

AndyGerlicher:
Our codebase is going to be a challenging test case for a tool like that. If you truly want to delete dead code you would have to build the entire matrix of `Debug`, `Debug-NetCore`, `Debug-Mono`, `Release`, `Release-NetCore`, and `Release-Mono`. 

And even with that we still have methods that can be called via Reflection. For example `IntrinsicFunctions` may not be called directly but they do all need to exist. We'd definitely be open to removing true dead code, but you would have to make sure that the build passes for all 6 configurations and tests pass before we'd really look at it. Thanks for the interest in doing this though!

zebmason:
Might have another crack at it later on, at least I've identified a bug in my code that caused the line ending issues. I'll have a read round beforehand to resolve the NetCore builds. Unfortunately I think reflection will always defeat the tool.

PR #1904, opened 2017-03-23
Allow StreamReader to detect file Encoding w/o BOM.

AndyGerlicher:
Thanks @davkean 

Passing in UTF8 w/o BOM Encoding object to StreamReader will allow the
StreamReader class to correctly detect encoding when a BOM exists in
the file. Using the default (UTF8 with BOM) will cause issues when the
file is UTF8 encoded without a BOM (added on save).

The overall issue was fixed in 8e3adba by reading the Stream. This
change allows for that behavior without needing to check the preamble
again.

#1898, #1768

Note: I'm not reverting my other change since part of it still necessary and the unit tests still need to read the preamble to verify the fix and we still need other aspects of that change to ensure that UTF8 and UTF8 w/o BOM are treated the same in some cases.

### DIFF
--- a/src/Build/Xml/XmlReaderExtension.cs
+++ b/src/Build/Xml/XmlReaderExtension.cs
@@ -26,7 +26,6 @@ namespace Microsoft.Build.Internal
         }
 
         private static readonly Encoding s_utf8NoBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
-        private readonly Encoding _encoding;
         private readonly Stream _stream;
         private readonly StreamReader _streamReader;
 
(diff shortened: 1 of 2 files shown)


src/Build/Xml/XmlReaderExtension.cs
```diff
@@ -34,25 +34,20 @@ private XmlReaderExtension(string file)
...
+                _streamReader = new StreamReader(_stream, s_utf8NoBom, detectEncodingFromByteOrderMarks: true);
+
                 Reader = GetXmlReader(_streamReader, out _encoding);
-                var bom = _stream.StartsWithPreamble();
 
                 // Override detected encoding if xml encoding attribute is specified
                 var encodingAttribute = Reader.GetAttribute("encoding");
                 _encoding = !string.IsNullOrEmpty(encodingAttribute)
                     ? Encoding.GetEncoding(encodingAttribute)
                     : _encoding;
```
davkean: Shouldn't you be consulting the encoding from StreamReader.CurrentEncoding?

AndyGerlicher: It is set by that a bit farther down in the file and returned as an out param of GetXmlReader:
https://github.com/Microsoft/msbuild/pull/1904/files#diff-abe59969026298c02814a641a6e57fceL83

PR #1905, opened 2017-03-24
Remove deadcode

zebmason:
Have run my static analysis tool on your codebase and identified regions of apparently dead code
The build and tests appear to all pass on my machine and the line endings seem to be correct now.

Please feel free to merge whatever you think is significant.


jeffkl:
I like the idea behind this but there's way too much to verify.  I spot checked a few and they were incorrect while others were fine.  One of the changes modified our public API surface which scares me.  

Can you break this up by assembly?  That would make it a lot easier I think.

zebmason:
@jeffkl It is rather a brutal code review tool. Code can be dead for several reasons:
* a bug
* it exists to be used within a debugger
* that section is currently under development 
* it was never tidied up

I'm afraid you'll need review it yourselves as I haven't the relevant competency

AndyGerlicher:
I think Jeff's point is it would be a lot easier to review if there were say a PR per assembly. Unit tests for example we don't care much (if the same number of tests run and pass in all configurations then the dead code should be removed). Our other assemblies require much more careful review.

zebmason:
@AndyGerlicher @jeffkl  Good point, didn't consider that as my motivation was testing my code against yours. Had intended to ship via VS Marketplace by this point but sales are currently restricted for the foreseeable future. i.e it would be easier for you to run my software yourselves. Suppose I could try to work out some sort of licencing agreement if you could put me in touch with the appropriate MS person.

AndyGerlicher:
I don't think this is something we'd be interested in purchasing. We're happy that you're using us as a test case and will definitely consider taking one or more changes though, especially around test projects. Thanks for the contribution!

zebmason:
Must admit the tool is not really ready to ship yet as there were too many false positives on msbuild. Had some good results with [Roslyn](https://github.com/dotnet/roslyn/pull/17630) where it flagged a regression. I think I've pretty much finished gathering tests to work with so won't be working on any more C# repositories unless anyone particularly wants me to.

So if you want to mention it round your colleagues at Microsoft feel free. I could probably do with some consultancy work in the short term.

AndyGerlicher:
Team Triage:
Closing this for now. Appreciate the contribution, but in order to review this we would need to have it split by project. Our test code I would be happy to run this on if the tests passes, but we can't do that for our public API and any changes need to be carefully reviewed by us.

PR #1906, opened 2017-03-24
Improve running of tests in VS

jeffkl:
Don't version test assemblies.  When building with MSBuild 15.0. MSBuildRuntimeType wasn't set so our assemblies weren't being versioned.  With the move to MSBuild 15.0, we get errors when trying to debug the tests in Visual Studio.

Copy XUnit dependencies to VS output path when building in VS.

### DIFF
--- a/targets/DeployDependencies.proj
+++ b/targets/DeployDependencies.proj
@@ -129,6 +129,12 @@
           DestinationFolder="@(RuntimeProjectJson->'%(TestDeploymentDir)')"
           SkipUnchangedFiles="true"
           />
+
+    <Copy SourceFiles="@(ResolvedRuntimeFiles)"
+          DestinationFolder="$(OutputPath)"
+          SkipUnchangedFiles="true"
+          Condition=" '$(BuildingInsideVisualStudio)' == 'true' "
+          />
   </Target>
 
   <Target Name="SetUnixPermissions"
--- a/src/dir.targets
+++ b/src/dir.targets
@@ -77,7 +77,7 @@
 
   <!-- TODO: enable for all runtime hosts https://github.com/Microsoft/msbuild/issues/694 -->
   <Import Project="$(GitVersioningDir)dotnet\Nerdbank.GitVersioning.targets"
-          Condition="'$(MSBuildRuntimeType)' == 'Full'" />
+          Condition="'$(MSBuildRuntimeType)' == 'Full' And '$(IsTestProject)' != 'true' " />
   <PropertyGroup Condition="'$(MSBuildRuntimeType)' != 'Full'">
     <DefineConstants>$(DefineConstants);STATIC_VERSION_NUMBER</DefineConstants>
   </PropertyGroup>


jeffkl:

targets/DeployDependencies.proj
```diff
@@ -129,6 +129,12 @@
           DestinationFolder="@(RuntimeProjectJson->'%(TestDeploymentDir)')"
           SkipUnchangedFiles="true"
           />
+
+    <Copy SourceFiles="@(ResolvedRuntimeFiles)"
```
cdmihai: Does this mean we can just get rid of `DeploymentDir` in favor of `OutputPath`? AFAIK, initially, we wanted  `OutputPath` to only contain compilation output, `DeploymentDir` to contain `OutputPath` + dependencies + runtime, and `TestDeploymentDir` to contain `DeploymentDir` + test dependencies. I think the original idea was to do one .net core build which would produce one `OutputPath` (anycpu, contents would be the same) but different deployment dirs for each OS (win, linux, osx).

Since we never ended up doing the multi deployment, I'd be in favor of this PR also getting rid of [`DeploymentDir`](https://github.com/Microsoft/msbuild/blob/master/dir.props#L409) to eliminate duplicate crud. And then re-orient the [`RuntimeProjectJson`](https://github.com/Microsoft/msbuild/blob/6851538897f5d7b08024a6d8435bc44be5869e53/targets/DeployDependencies.proj#L90) to `OutputPath`.

A good test for this change would be the Microbuild build

jeffkl: I'm not sure.  When running tests from Visual Studio, we need runtime dependencies.  At some point, we'll be re-doing our projects and the output paths.

AndyGerlicher:
@mfilippov Do you know offhand if there's a way to get ReSharper to set the AppDomain base directory to something other than ReSharper install folder? When running tests in VS with ReSharper it seems to find Microsoft.Build.dll from that folder (looks like you guys ship that signed with a different key). It also doesn't use the app.config from the test projects so there are differences running the command-line runner and ReSharper xunit runner. Are these known issues? Or just because we happen to be Microsoft.Build.dll? Sorry to bug you just figure you might know.

mfilippov:
Seems to be xUnit specific plugin problem. @citizenmatt could you help?

jeffkl:
Here's some more context:

Our build output directory contains `Microsoft.Build.dll` as well as the unit tests.  When running under Visual Studio and the ReSharper test runner, the AppDomain BaseDirectory is 

`C:\Users\jeffkl\AppData\Local\JetBrains\Installations\ReSharperPlatformVs15_34e677e2\`

That directory contains `Microsoft.Build.dll` but is not our copy which is under test.  The public key token doesn't match so we get runtime exceptions.

We run tests from the command-line instead which appear to set the AppDomain BaseDirectory to our output folder instead.

citizenmatt:
We just use unit's APIs to run tests, and let xunit create the AppDomain. It uses the test assembly to create the AppDomain base. From what you say, this isn't Microsoft.Build.dll, so I don't see how the AppDomain is set to the install dir, unless you've disabled AppDomains, either in ReSharper's options, or in an xunit config file. This would mean that there is only the default AppDomain, which will always be the install dir, and we can't change this. Is this the case?

That said, I've just looked at the bin folder for ReSharper 2017.1, and it looks like we don't ship Microsoft.Build.dll any more. Perhaps you could try it with the [latest EAP preview build](https://www.jetbrains.com/resharper/eap/)?

jeffkl:
@citizenmatt I just installed from EAP and can confirm that `Microsoft.Build.dll` is no longer there, thanks!

However, the AppDomain.BaseDirectory is still `C:\Users\jeffkl\AppData\Local\JetBrains\Installations\ReSharperPlatformVs15_34e677e2\`

image

image


I checked and I don't think we're setting anything for xunit.  Any other ideas?

citizenmatt:
What are you looking at there? If xunit is creating the AppDomain, it names it after a GUID. You should get something like this (`AppDomain.CurrentDomain`):

screen shot 2017-03-24 at 7 16 19 pm

The ApplicationName you're seeing looks like it's the default domain.

jeffkl:
From Visual Studio, I'm debugging the test via a ReSharper Unit Test Session.  

image

It certainly looks like its using the test runner appdomain and not an appdomain for the test assembly.  Any tips on how I can troubleshoot why an AppDomain isn't being created?

citizenmatt:
AppDomains are being disabled by xunit itself. It's [looking to see if the xunit assembly ends with `.dotnet.dll`](https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Frameworks/v2/Xunit2Discoverer.cs#L245-L246). If so, it expects it to be a .NET Core test run, and .NET Core doesn't support AppDomains. Xunit will then disable AppDomain support, even though you're running on the .NET Framework.

Which means you won't be able to use the test assembly's app.config file - the host process' only AppDomain has already been created, and will be using its own config file, and AFAIK, you can't replace it.

PR #1918, opened 2017-03-28
PVS-Studio: fixed potential vulnerabilities

skhrenov:
[V3095](https://www.viva64.com/en/w/V3095/) The 'condition' object was used before it was verified against null. Check lines: 240, 244. Microsoft.Build.Engine Utilities.cs 240
[V3095](https://www.viva64.com/en/w/V3095/) The 'ReferenceInfo' object was used before it was verified against null. Check lines: 80, 90. Microsoft.Build.Tasks AxReference.cs 80
[V3095](https://www.viva64.com/en/w/V3095/) The 'packageFileName' object was used before it was verified against null. Check lines: 1439, 1448. Microsoft.Build.Tasks BootstrapperBuilder.cs 1439
[V3095](https://www.viva64.com/en/w/V3095/) The 'metadataNames' object was used before it was verified against null. Check lines: 243, 253. Microsoft.Build.Tasks CommandLineBuilderExtension.cs 243
[V3095](https://www.viva64.com/en/w/V3095/) The 'diskRoots' object was used before it was verified against null. Check lines: 2880, 2883. Microsoft.Build.Utilities ToolLocationHelper.cs 2880

V3095 -> CWE-476 (Medium) NULL Pointer Dereference

### DIFF
--- a/src/Tasks/AxReference.cs
+++ b/src/Tasks/AxReference.cs
@@ -77,7 +77,10 @@ namespace Microsoft.Build.Tasks
 
             var axImp = new ResolveComReference.AxImp();
 
-            axImp.ActiveXControlName = ReferenceInfo.strippedTypeLibPath;
+            if (ReferenceInfo != null)
+            {
+                axImp.ActiveXControlName = ReferenceInfo.strippedTypeLibPath;
+            }
 
             axImp.BuildEngine = BuildEngine;
             axImp.ToolPath = ToolPath;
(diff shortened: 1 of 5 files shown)


src/Tasks/AxReference.cs
```diff
@@ -77,7 +77,10 @@ internal bool GenerateWrapper(out ComReferenceWrapperInfo wrapperInfo)
 
             var axImp = new ResolveComReference.AxImp();
 
-            axImp.ActiveXControlName = ReferenceInfo.strippedTypeLibPath;
+            if (ReferenceInfo != null)
```
cdmihai: I would fail here instead of silently skipping the execution.

PR #1926, opened 2017-03-29
Enable RuntimeInformation intrinsic functions on all platforms

cdmihai:


eerhardt:
This basically fixes #539 right?  You can detect which OS you are on with this.

src/Build/project.json
```diff
@@ -1,4 +1,7 @@
 ﻿{
+  "dependencies": {
+    "System.Runtime.InteropServices.RuntimeInformation": "4.0.0"
```
jeffkl: Be sure to add the assembly to our custom copy logic here: https://github.com/cdmihai/msbuild/blob/fa2c0f8541ed96a303927131abe268952bf37706/src/Build/Microsoft.Build.csproj#L798

You might need to also add it to the build definition so that it ends up in our VSIX.

cdmihai: Good catch ... I'll also update the swr and microbuild if needed

cdmihai:
> This basically fixes #539 right? You can detect which OS you are on with this.

@eerhardt I think so, yes. In the future we could think about adding intrinsic functions like `IsLinux` / `IsOSX` / `IsWindows` to avoid screen wide lines.

cdmihai:
@dotnet-bot help

cdmihai:

cdmihai:

cdmihai:
@dotnet-bot help

radical:
@cdmihai Thanks! One item off my list ;)

build/NuGetPackages/Microsoft.Build.Runtime.nuspec
```diff
@@ -19,12 +19,14 @@
         <dependency id="Microsoft.Build.Framework" version="[$version$]" />
         <dependency id="Microsoft.Build.Tasks.Core" version="[$version$]" />
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
+        <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.0.0" />
```
jeffkl: You shouldn't need to add this to our runtime package because the dependency on Microsoft.Build transitively depends on `System.Runtime.InteropServices.RuntimeInformation`.

cdmihai: MSBuild.exe depends on RuntimeInformation via the MSBuildEnvironmentHelper, which is compiles into MSBuild.exe. Because of that, I thought it might be better to make the dependency explicit.
On the other hand, everything used by msbuild.exe package flows from the other packages. so I'll remove it for consistency's sake :)

PR #1928, opened 2017-03-30
Implement ability for project properties to suppress and elevate warnings

### LINKED ISSUE #1886: No way to set /nowarn:MSB3277 from Visual Studio
For a long time even Minimal verbosity build logs were polluted by warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved, and there were no way to suppress those as they were coming directly from MSBuild.

In MSBuild 15 FINALLY we got new option /nowarn:MSB3277 which supposed to suppress that and some other MSBuild warnings. That was the primary reason for our switch to VS 15 as soon as it become available. 

But now for the life of me I cannot figure out how this option could be set for Visual Studio builds! For the command line builds where we invoke MSBuild directly it is obvious, but most devs use VS and they seem out of luck - no MSBuild command line and no way to set /nowarn:MSB3277

I understand that VS most probably does not have UI to set that option and it is Ok. But I sincerely hope that there is a way to tweak the project file or machine wide configuration or SOMETHING to make this option effective for VS originated builds. 

Any advice, please?

Konstantin

jeffkl:
Introduce `MSBuildTreatWarningsAsErrors` property that when set to true will treat all warnings as errors when building a project.
Introduce `MSBuildWarningsAsErrors` property which is a semicolon delmited list of warning codes to treat as errors when building a project.
Introduce `MSBuildWarningsAsMessages` property which is a semicolon delimited list of warning codes to treat as low importance messages.

This allows users to control warnings at the project level via standard MSBuild properties.  They can be set in a project, an import, or from the command-line.

The limitation here is that the warnings can only include ones that occur during a build.  This is because the properties are not read until after parsing so warnings generated during parse/evaluation happen too soon.  For these warnings, users will only be able to treat them as errors/messages with the `/WarnAsError` and `/WarnAsMessage` command-line arguments.

Closes #1886

### DIFF
--- a/src/Build/BackEnd/Components/Logging/EventSourceSink.cs
+++ b/src/Build/BackEnd/Components/Logging/EventSourceSink.cs
@@ -228,6 +228,12 @@ namespace Microsoft.Build.BackEnd.Logging
             else if (buildEvent is ProjectFinishedEventArgs)
             {
                 this.RaiseProjectFinishedEvent(null, (ProjectFinishedEventArgs)buildEvent);
+
+                if (buildEvent.BuildEventContext != null && buildEvent.BuildEventContext.ProjectInstanceId != BuildEventContext.InvalidProjectInstanceId)
+                {
+                    WarningsAsErrorsByProject?.Remove(buildEvent.BuildEventContext.ProjectInstanceId);
+                    WarningsAsMessagesByProject?.Remove(buildEvent.BuildEventContext.ProjectInstanceId);
+                }
             }
             else if (buildEvent is BuildStartedEventArgs)
             {
--- a/src/Build/BackEnd/Components/Logging/EventSourceSink.cs
+++ b/src/Build/BackEnd/Components/Logging/EventSourceSink.cs
@@ -141,9 +141,27 @@ namespace Microsoft.Build.BackEnd.Logging
         }
 
         /// <summary>
+        /// A list of warnings to treat as errors for an associated <see cref="BuildEventContext.ProjectInstanceId"/>.  If the set associated with a ProjectInstanceId is null, nothing is treated as an error.  If an empty set, all warnings are treated as errors.
+        /// </summary>
+        public IDictionary<int, ISet<string>> WarningsAsErrorsByProject
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
         /// A list of warnings to treat as low importance messages.
         /// </summary>
         public ISet<string> WarningsAsMessages
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// A list of warnings to treat as low importance messages for an associated <see cref="BuildEventContext.ProjectInstanceId"/>.
+        /// </summary>
+        public IDictionary<int, ISet<string>> WarningsAsMessagesByProject
         {
             get;
             set;
@@ -233,7 +251,7 @@ namespace Microsoft.Build.BackEnd.Logging
             {
                 BuildWarningEventArgs warningEvent = (BuildWarningEventArgs) buildEvent;
 
-                if (WarningsAsMessages != null && WarningsAsMessages.Contains(warningEvent.Code))
+                if (ShouldTreatWarningAsMessage(warningEvent))
                 {
                     // Treat this warning as a message with low importance if its in the list
                     BuildMessageEventArgs errorEvent = new BuildMessageEventArgs(
@@ -257,7 +275,7 @@ namespace Microsoft.Build.BackEnd.Logging
                     this.RaiseMessageEvent(null, errorEvent);
 
                 }
-                else if (WarningsAsErrors != null && (WarningsAsErrors.Count == 0 || WarningsAsErrors.Contains(warningEvent.Code)))
+                else if (ShouldTreatWarningAsError(warningEvent))
                 {
                     // Treat this warning as an error if an empty set of warnings was specified or this code was specified
                     BuildErrorEventArgs errorEvent = new BuildErrorEventArgs(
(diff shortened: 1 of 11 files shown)


jeffkl:
FYI @jaredpar

jaredpar:
Nice!

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1079,6 +1080,10 @@ private void SetProjectCurrentDirectory()
 
             _projectLoggingContext = _nodeLoggingContext.LogProjectStarted(_requestEntry);
 
+            // Now that the project has started, parse a few known properties which indicate warning codes to treat as errors or messages
+            //
+            ConfigureWarningsAsErrorsAndMessages();
```
cdmihai: Hopefully this is the only place where MSBuild evaluates a project file during build. :)

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1246,5 +1251,59 @@ private void VerifyIsNotZombie()
...
+        {
+            // Gather needed objects
+            //
+            ProjectInstance project = _requestEntry?.RequestConfiguration?.Project;
+            BuildEventContext buildEventContext = _projectLoggingContext?.BuildEventContext;
+            ILoggingService loggingService = _projectLoggingContext?.LoggingService;
+
+            // Ensure everything that is required is available at this time
+            //
+            if (project != null && buildEventContext != null && loggingService != null && buildEventContext.ProjectInstanceId != BuildEventContext.InvalidProjectInstanceId)
```
cdmihai: Check this works on p2p node as well. Nodes initially assign a temp project ID (maybe the invalid one?) and I don't remember whether they get the global ID before or after they evaluate the project.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1246,5 +1251,59 @@ private void VerifyIsNotZombie()
...
+            {
+                if (String.Equals(project.GetPropertyValue(MSBuildConstants.TreatWarningsAsErrors)?.Trim(), "true", StringComparison.OrdinalIgnoreCase))
+                {
+                    // If <MSBuildTreatWarningsAsErrors was specified then an empty ISet<string> signals the IEventSourceSink to treat all warnings as errors
+                    //
+                    loggingService.AddWarningsAsErrors(buildEventContext.ProjectInstanceId, new HashSet<string>());
+                }
+                else
+                {
+                    ISet<string> warningsAsErrors = ParseWarningCodes(project.GetPropertyValue(MSBuildConstants.WarningsAsErrors));
```
cdmihai: What if the user defines TreatWarningsAsErrors=True and also WarningsAsErrors="a;b;c;d"? I'd expect msbuild to treat only the specified warnings as errors, as opposed to ignoring the warning list.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1246,5 +1251,59 @@ private void VerifyIsNotZombie()
...
+
+        private ISet<string> ParseWarningCodes(string warnings)
+        {
+            if (String.IsNullOrWhiteSpace(warnings))
+            {
+                return null;
+            }
+
+            return new HashSet<string>(
+                warnings.Split(new[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries)
```
cdmihai: I'd only keep semicolons, as they are the msbuild approved way of separating things. And then maybe use ExpressionShredder.SplitSemiColonSeparatedList

src/Build/BackEnd/Components/Logging/LoggingService.cs
```diff
@@ -481,14 +482,40 @@ public ISet<string> WarningsAsMessages
...
+                    _filterEventSource.WarningsAsErrorsByProject = new ConcurrentDictionary<int, ISet<string>>();
+                }
+
+                _filterEventSource.WarningsAsErrorsByProject.Add(projectInstanceId, new HashSet<string>(codes, StringComparer.OrdinalIgnoreCase));
+            }
+        }
+
+        public void AddWarningsAsMessages(int projectInstanceId, ISet<string> codes)
+        {
+            lock (_lockObject)
```
cdmihai: If you are using a concurrent dictionary, rather than locking, you can use a concurrent Lazy to initialize it, and then not use the lock.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1246,5 +1251,59 @@ private void VerifyIsNotZombie()
...
+                {
+                    ISet<string> warningsAsErrors = ParseWarningCodes(project.GetPropertyValue(MSBuildConstants.WarningsAsErrors));
+
+                    if (warningsAsErrors?.Count > 0)
+                    {
+                        loggingService.AddWarningsAsErrors(buildEventContext.ProjectInstanceId, warningsAsErrors);
+                    }
+                }
+
+                ISet<string> warningsAsMessages = ParseWarningCodes(project.GetPropertyValue(MSBuildConstants.WarningsAsMessages));
```
cdmihai: What if the user has the same warning code as both a message and an error? I'd expect error to win, on the grounds that errors are more serious than messages. Or maybe emit a warning saying that both were specified and the intent is ambiguous :)

src/Build/BackEnd/Components/Logging/EventSourceSink.cs
```diff
@@ -970,6 +988,75 @@ private void RaiseTelemetryEvent(object sender, TelemetryEventArgs buildEvent)
...
+        /// <summary>
+        /// Determines if the specified warning should be treated as an error.
+        /// </summary>
+        /// <param name="warningEvent">A <see cref="BuildWarningEventArgs"/> that specifies the warning.</param>
+        /// <returns><code>true</code> if the warning should be treated as an error, otherwise <code>false</code>.</returns>
+        private bool ShouldTreatWarningAsError(BuildWarningEventArgs warningEvent)
+        {
+            // This only applies if the user specified /warnaserror from the command-line or added an empty set through the object model
+            //
+            if (WarningsAsErrors != null)
```
cdmihai: I'd rename WarningsAsErrors/Messages to something like GlobalWarningsAsErrors/Messages

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
jeffkl: Do you foresee any performance degradation by adding this call?  I assumed it was safe since the project has "started" building...

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
jeffkl: At the moment, `MSBuildTreatWarningsAsErrors` will override any list.  This is the same way that argument works for the compilers.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
jeffkl: Suppressing a warning overrides elevating one.  This is so you can treat all warnings as errors but still turn a few of them off.

src/Build/BackEnd/Components/Logging/EventSourceSink.cs
jeffkl: The name is technically part of our public API already as a property to `BuildParameters`.  I could rename the internal property on `EventSourceSink` if you feel strongly about it.

src/Build/BackEnd/Components/Logging/LoggingService.cs
jeffkl: I tried this out but hit a snag.  I'm using the state of `null` as an indicator if any warnings should be treated as an error/message.  If I use a `Lazy<T>`, then I would have to expose it as such so that the caller can determine if `HasValue` is `true`.  Leaving it the way it is, the property remains of type `IDictionary` and a simple null check is done.  If its okay to have `IBuildEventSink` have a property that is of type `Lazy<T>`, then I'll do it.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
jeffkl: Done

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
cdmihai: It runs once for each project evaluation, so it should scale linearly with number of evaluations (Project nr  x (msbuild tasks per project)).

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
cdmihai: Cool, didn't know. Go with the same behaviour as the compilers then. Except for the warning list separator :)

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
cdmihai: Offline discussion: This actually happens only once per node.

src/Build/BackEnd/Components/Logging/EventSourceSink.cs
cdmihai: It was more of a nit. Thought it's worth making it clear in the ILoggingService that `WarningsAsErrors` is the global, master switch and not the local per project one.

PR #1931, opened 2017-03-31
Perf Regression: Stop using MSBuildNameIgnoreCaseComparer

AndyGerlicher:
Fix for regression in performance caused by #1766. Port of #1929.

Internal Visual Studio perf tests discovered a small regression in 15.1 that appears to be due to using MSBuildNameIgnoreCaseComparer.Default for string comparison in the lazy evaluator. The change was originally made because of a behavior regression that caused item references to be case-sensitive. Using StringComparer.OrdinalIgnoreCase seems to resolve the performance issue and maintains the correct behavior from previous versions of MSBuild.


AndyGerlicher:
Cherry-pick to master. Note this was going to go into 15.1 but will be reverted. It will not be included in 15.1.

AndyGerlicher:

PR #1933, opened 2017-04-03
Fix metadata reference in a property function, in an item metadata

### LINKED ISSUE #1932: [REGRESSION] Metadata reference in a property function doesn't get expanded
```
<Project>
        <ItemGroup>
            <ResxCode Include="Resources\Translation.resx">
              <WithNew>$([System.String]::new('%(RelativeDir)').TrimEnd('\'))</WithNew>
            </ResxCode>
        </ItemGroup>
        <Target Name="Main">
                <Message Text="WithNew: %(ResxCode.WithNew)"/>
        </Target>
</Project>
```

Expected result:
`WithNew: Resources/`

Actual result:
`WithNew: %(RelativeDir)`

Reverting `816d37362bc34a6dd754607cad8c601ae8f13000` fixes the issue:

```
commit 816d37362bc34a6dd754607cad8c601ae8f13000
Author: Mihai Codoban <micodoba@microsoft.com>
Date:   Thu Mar 23 13:09:56 2017 -0700

    Expand metadata properties earlier (#1896)

    * Expand metadata properties at operation construction instead of operation application

    Fixes #1359

    If properties are not expanded at operation construction, potential indirect item references via those properties will get missed, and thus evaluate to empty during operation application.

    Side effect change: if metadata properties are evalauted early, there's no use for operation application to expand them yet again, so cache the property expansion during operation construction and use the cached values during operation application. This should also be a perf improvement when metadata references are present within metadata, as that triggers the re-evaluation of all metadata elements for each item instance.

    * Add ItemSpec ToString

    Better debugging info
```

This breaks `netfx-System.StringResources.3.0.13` also.

radical:
Fixes #1932.

This partially reverts 816d37362bc34a6dd754607cad8c601ae8f13000,
specifically the part (quoted from the commit message):

` Side effect change: if metadata properties are evalauted early,
there's no use for operation application to expand them yet again, so
cache the property expansion during operation construction and use the
cached values during operation application. This should also be a perf
improvement when metadata references are present within metadata, as
that triggers the re-evaluation of all metadata elements for each item
instance.`

We can't use the metadata expanded at operation construction time as we
didn't expand referenced item metadata at that time. This breaks a case
where the item metadata has a value like:

`<MD>$([System.String]::new('%(RelativeDir)').TrimEnd('\'))</MD>`

So, we need to expand the original metadata string at operation
application time.

### DIFF
--- a/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Build.Evaluation
                 }
             }
 
-            protected void DecorateItemsWithMetadata(ICollection<I> items, ImmutableList<PartiallyEvaluatedMetadata> metadata)
+            protected void DecorateItemsWithMetadata(ICollection<I> items, ImmutableList<ProjectMetadataElement> metadata)
             {
                 if (metadata.Any())
                 {
@@ -110,14 +110,14 @@ namespace Microsoft.Build.Evaluation
 
                     // Do not expand properties as they have been already expanded by the lazy evaluator upon item operation construction.
                     // Prior to lazy evaluation ExpanderOptions.ExpandAll was used.
-                    const ExpanderOptions metadataExpansionOptions = ExpanderOptions.ExpandMetadata | ExpanderOptions.ExpandItems;
+                    const ExpanderOptions metadataExpansionOptions = ExpanderOptions.ExpandAll;
 
                     List<string> values = new List<string>(metadata.Count * 2);
 
-                    foreach (var metadatum in metadata)
+                    foreach (var metadataElement in metadata)
                     {
-                        values.Add(metadatum.ValueWithPropertiesExpanded);
-                        values.Add(metadatum.ConditionWithPropertiesExpanded);
+                        values.Add(metadataElement.Value);
+                        values.Add(metadataElement.Condition);
                     }
 
                     ItemsAndMetadataPair itemsAndMetadataFound = ExpressionShredder.GetReferencedItemNamesAndMetadata(values);
@@ -153,23 +153,23 @@ namespace Microsoft.Build.Evaluation
                         {
                             _expander.Metadata = item;
 
-                            foreach (var metadatum in metadata)
+                            foreach (var metadataElement in metadata)
                             {
 #if FEATURE_MSBUILD_DEBUGGER
                                 //if (DebuggerManager.DebuggingEnabled)
                                 //{
-                                //    DebuggerManager.PulseState(metadatumElement.Location, _itemPassLocals);
+                                //    DebuggerManager.PulseState(metadataElementElement.Location, _itemPassLocals);
                                 //}
 #endif
 
-                                if (!EvaluateCondition(metadatum.ConditionWithPropertiesExpanded, metadatum.Element, metadataExpansionOptions, ParserOptions.AllowAll, _expander, _lazyEvaluator))
+                                if (!EvaluateCondition(metadataElement.Condition, metadataElement, metadataExpansionOptions, ParserOptions.AllowAll, _expander, _lazyEvaluator))
                                 {
                                     continue;
                                 }
 
-                                string evaluatedValue = _expander.ExpandIntoStringLeaveEscaped(metadatum.ValueWithPropertiesExpanded, metadataExpansionOptions, metadatum.Element.Location);
+                                string evaluatedValue = _expander.ExpandIntoStringLeaveEscaped(metadataElement.Value, metadataExpansionOptions, metadataElement.Location);
 
-                                item.SetMetadata(metadatum.Element, evaluatedValue);
+                                item.SetMetadata(metadataElement, evaluatedValue);
                             }
                         }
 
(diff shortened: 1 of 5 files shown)


src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -110,14 +110,14 @@ protected void DecorateItemsWithMetadata(ICollection<I> items, ImmutableList<Par
 
                     // Do not expand properties as they have been already expanded by the lazy evaluator upon item operation construction.
                     // Prior to lazy evaluation ExpanderOptions.ExpandAll was used.
-                    const ExpanderOptions metadataExpansionOptions = ExpanderOptions.ExpandMetadata | ExpanderOptions.ExpandItems;
```
cdmihai: The bug is that recursive metadata is not expanded, however this line clearly states that metadata should get expanded at evaluation time. It's just properties that don't get expanded, as they have already been expanded previously during parsing time. Since `%(RelativeDir)` is metadata and not a property, the cause of the bug must be somewhere else. I'd rather keep the optimization and find the true cause.

radical: I have no real idea about this code :( But my cursory glance (and assumptions!) suggested that the property gets expanded at evaluation time, but since metadata is not being expanded at that time, the property just evaluates to the string `%(RelativeDir)`, which would mean that the whatever the property function wanted to do with it, cannot be done now. Isn't that still an issue?

You are correct that it says `ExpandMetadata` here, so even if it expanded to the `string` `%(RelativeDir)`, that should have expanded at this time.

I'll wait for the real fix :) 👍

cdmihai: Oh, you're right, I had not seen / realized that the metadata was inside a property function call.
My bad, the fix is good. Good bye optimization :)

radical: I'm wondering about my comment - the property function `$([System.String]::new('%(RelativeDir)').TrimEnd('\'))` would have been expanded at operation construction time and that should have resulted in `%(RelativeDir)`, is that correct? If so, then shouldn't that have been expanded at `operation application` time because we have ...

`const ExpanderOptions metadataExpansionOptions = ExpanderOptions.ExpandMetadata | ExpanderOptions.ExpandItems;`

... like you pointed out? I am asking out of curiosity, but didn't actually go and confirm this with a simple printf!

cdmihai: Good catch @radical!

After an adventurous debugging session, it seems that `%(RelativeDir)` does not get expanded at operation application because [the result of a property function is always escaped](https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Expander.cs#L3167-L3173). This means that `$([System.String]::new('%(RelativeDir)').TrimEnd('\'))` expands into `%25%28RelativeDir%29` which is then not detected as metadata during operation application.

The before-the-regression code and your fix work because if metadata expansion is delayed until operation application, it will get expanded with `ExpanderOptions.ExpandAll` which makes the `Expander` expand references in a [specific order](https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Expander.cs#L253-L255): 1) metadata, 2) properties, 3) items. Since metadata is expanded before properties, the property function gets the correct values.

Consequences:
```xml
<Project>
        
        <ItemGroup>
            <A Include="1;$(P)" Exclude="$(P)"/>
            <B Include="a\**">
              <WithMetArg>$([System.String]::new('%(RelativeDir)').TrimEnd('\'))</WithMetArg>
              <WithIndirection>$(P)</WithIndirection>
            </B>
            <C Include="$([System.String]::new('@(A)'))"/>
            <D Include="$([System.String]::new('$(P)'))"/>
        </ItemGroup>

        <PropertyGroup>
            <P>@(A)</P>
        </PropertyGroup>

        <Target Name="Main">
                <Message Text="WithMetArg: %(B.WithMetArg)"/>
                <Message Text="WithIndirection: %(B.WithIndirection)"/>
                <Message Text="Item as property function argument: @(C)"/>
                <Message Text="Property as property function argument: @(D)"/>
        </Target>
</Project>
```

Evaluates to the following on dev14
```
Microsoft (R) Build Engine version 14.0.25420.1

Build started 4/5/2017 1:39:25 PM.
Project "D:\projects\tests\projects\play\build.proj" on node 1 (default targets).
Main:
  WithMetArg: a\b
  WithIndirection: 1
  Item as property function argument: @(A)
  Property as property function argument: @(A)
Done Building Project "D:\projects\tests\projects\play\build.proj" (default targets).
```

Notice that the property reference inside the argument does get expanded, because arguments are actually expanded with the same `ExpanderOptions` as the initial string: https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Expander.cs#L3050

I'll add tests to capture this behaviour.

radical: Thanks for the all the digging+explanation! It makes sense now :) It would be useful to have this in the commit message with the tests that you will be adding. 👍

PR #1948, opened 2017-04-05
Tests for element reference as property function arguments

cdmihai:
Prompted by #1932 and https://github.com/Microsoft/msbuild/pull/1933#discussion_r110026336.
The asserts are taken from dev14's behaviour.

### DIFF
--- a/src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
@@ -337,7 +337,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
             var projectContent =
 @"<Project>
         <ItemGroup>
-            <A Include=`1;$(P)` Exclude=`$(P)`/>
+            <A Include=`1`/>
             <B Include=`$([System.String]::new('@(A)'))`/>
             <C Include=`$([System.String]::new('$(P)'))`/>
         </ItemGroup>
@@ -358,7 +358,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
             var projectContent =
 @"<Project>
         <ItemGroup>
-            <A Include=`1;$(P)` Exclude=`$(P)`/>
+            <A Include=`1` />
             <B Include=`B`>
                <M>$([System.String]::new(`%(Identity)`))</M>
             </B>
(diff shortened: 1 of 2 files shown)


cdmihai:
@radical

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -332,6 +332,78 @@ public void ItemOperationsShouldExpandIndirectItemReferences(string projectConte
         }
 
         [Fact]
+        public void OnlyPropertyReferencesGetExpandedInPropertyFunctionArgumentsInsideIncludeAttributes()
+        {
+            var projectContent =
+@"<Project>
+        <ItemGroup>
+            <A Include=`1;$(P)` Exclude=`$(P)`/>
```
radical: Why have the `Exclude` here? I am thinking that whatever `$(P)` would evaluate to (like some incorrect value due to a bug) then it won't affect `@(A)`, so doesn't that make this a no-op, in a sense?

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -332,6 +332,78 @@ public void ItemOperationsShouldExpandIndirectItemReferences(string projectConte
...
+
+        [Fact]
+        public void MetadataAndPropertyReferencesGetExpandedInPropertyFunctionArgumentsInsideMetadataElements()
+        {
+            var projectContent =
+@"<Project>
+        <ItemGroup>
+            <A Include=`1;$(P)` Exclude=`$(P)`/>
+            <B Include=`B`>
+               <M>$([System.String]::new(`%(Identity)`))</M>
```
radical: Should there be another `<M2>...` for non-built-in metadata?

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
cdmihai: Correct, it's superfluous. I forgot it in from experimentation time.

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
cdmihai: Wouldn't hurt, yeah. I'll add it.

cdmihai:

PR #1950, opened 2017-04-06
Differentiate version for RuntimeInformation and bring it back for property functions

cdmihai:
VS needs 4.0.1.0. If we use anything else, we won't get VS' ngened copy which then makes all of msbuild not be ngened.

This PR also brings back #1926 which got reverted in #1951

### DIFF
--- a/build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
+++ b/build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
@@ -67,7 +67,7 @@
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
         <dependency id="System.Collections.Immutable" version="1.3.1" />
         <dependency id="System.Reflection.Metadata" version="1.3.0" />
-        <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.0.0" />
+        <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.3.0" />
       </group>
     </dependencies>
   </metadata>
--- a/build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
+++ b/build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
@@ -67,6 +67,7 @@
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
         <dependency id="System.Collections.Immutable" version="1.3.1" />
         <dependency id="System.Reflection.Metadata" version="1.3.0" />
+        <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.0.0" />
       </group>
     </dependencies>
   </metadata>
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -2601,6 +2602,10 @@ namespace Microsoft.Build.UnitTests.Evaluation
             "$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)",
             "$$architecture$$"
         )]
+        [InlineData(
+            "$([MSBuild]::IsOsPlatform($$platform$$))",
+            "True"
+        )]
         public void PropertyFunctionRuntimeInformation(string propertyFunction, string expectedExpansion)
         {
             Func<string, string, string, string> formatString = (aString, platform, architecture) => aString
@@ -2610,20 +2615,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
             var pg = new PropertyDictionary<ProjectPropertyInstance>();
             var expander = new Expander<ProjectPropertyInstance, ProjectItemInstance>(pg);
 
-            var currentPlatformString = string.Empty;
-
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                currentPlatformString = "Windows";
-            }
-            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
-            {
-                currentPlatformString = "Linux";
-            }
-            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-            {
-                currentPlatformString = "OSX";
-            }
+            string currentPlatformString = Helpers.GetOSPlatformAsString();
 
             var currentArchitectureString = RuntimeInformation.OSArchitecture.ToString();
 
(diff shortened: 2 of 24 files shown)


cdmihai:

cdmihai:
Github apparently does not know an old branch from a new branch, closing and renaming my original branch.

cdmihai:
@eerhardt RuntimeInformation in property functions may be coming back :)

eerhardt:
> RuntimeInformation in property functions may be coming back :)

Yes, please.  I just hit this myself again where I need to do something different between macOS and Linux.

build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
```diff
@@ -67,6 +67,7 @@
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
         <dependency id="System.Collections.Immutable" version="1.3.1" />
         <dependency id="System.Reflection.Metadata" version="1.3.0" />
+        <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.3.0" />
```
rainersigwald: devenv.exe.config has a binding redirect for lower versions of this:

```xml
<dependentAssembly>
  <assemblyIdentity name="System.Runtime.InteropServices.RuntimeInformation" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>
  <bindingRedirect oldVersion="0.0.0.0-4.0.1.0" newVersion="4.0.1.0" />
</dependentAssembly>
```

Since this is outside of that range we can still ship and use the newer one, but we need to be a bit careful about it. Do we need this version specifically?

rainersigwald: Oh, I see: `System.Runtime.InteropServices.OSPlatform` only became available for netstandard in `4.0.2.0`. https://apisof.net/catalog/System.Runtime.InteropServices.OSPlatform. And the only package newer than that is `4.3.0`. Well, that's . . . something.

src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2612,16 +2625,15 @@ public void PropertyFunctionRuntimeInformation()
                 currentPlatformString = "OSX";
             }
```
rainersigwald: Maybe an `else Assert.True(false, "unrecognized current platform")`?

src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2585,19 +2585,32 @@ public void PropertyFunctionGetFolderPath()
...
-            var pg = new PropertyDictionary<ProjectPropertyInstance>();
+        /// The test exercises: RuntimeInformation / OSPlatform usage, static method invocation, static property invocation, method invocation expression as argument, call chain expression as argument
+        /// </summary>
+        [Theory]
+        [InlineData(
+            "$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Create($([System.Runtime.InteropServices.OSPlatform]::$$platform$$.ToString())))))",
+            "True"
+        )]
+        [InlineData(
+            @"$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::$$platform$$)))",
```
rainersigwald: This is pretty clunky, should we add `$([MSBuild]::IsOSPlatform(string))` that makes it easier?

src/Build/project.json
```diff
@@ -22,6 +23,7 @@
         "System.Reflection.Metadata": "1.3.0",
         "System.Reflection.TypeExtensions": "4.1.0",
         "System.Runtime": "4.1.0",
+        "System.Runtime.InteropServices.RuntimeInformation": "4.0.0",
```
rainersigwald: Why `4.0.0` here? Does the higher version drag in netcore 2 stuff?

src/Shared/UnitTests/App.config
```diff
@@ -30,6 +30,15 @@
         <assemblyIdentity name="Microsoft.Build.Utilities.Core" culture="neutral" publicKeyToken="b03f5f7f11d50a3a" />
         <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.1.0.0"/>
       </dependentAssembly>
+
+      <!-- RuntimeInformation is redirected because src/Build.UnitTests/project.json incorrectly resolves to 4.0.0.0 (package 4.0.0-rc3-24117-00) 
+      instead of 4.0.1.0 (package 4.3.0) for net4.6.
```
rainersigwald: This comment . . . my brain . . . why, versioning, why????

build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
cdmihai: Yes, and the nuget package 4.3.0 brings the 4.0.1.0 RuntimeInformation assembly. I had initially written 4.0.1.0 in all the jsons and then spent half an hour trying to reason why didn't get it :)

src/Build.UnitTests/Evaluation/Expander_Tests.cs
cdmihai: That sounds like a good idea. The only nit is that it requires a magic string. I guess `windows`, `osx`, and `linux` are good enough. Unless it gives people the illusion of being able to also use `unix`, or `ubuntu`. As long as we document it (somewhere) it should be fine. @AndyGerlicher ?

src/Build/project.json
cdmihai: Honestly I envisioned spending another day or so on this if I changed the .net core version, so decided to leave it alone :)
I'll give it a try and diff the lock files to see what happens.

src/Shared/UnitTests/App.config
cdmihai: It is really weird. The Utilities unit test have an almost identical project.json and an identical dependency tree path for RuntimeInformation, but they get 4.30 whereas the engine unit test project.json gets 4.0.0. I looked over it with @kingerja but could not figure it out.

Maybe upgrading our CLI version to the latest one that still supported project.jsons would fix it, but probably it's worth just waiting for the SDKs :)

src/Build.UnitTests/Evaluation/Expander_Tests.cs
rainersigwald: I was thinking maybe just attempt to cram it in as though they'd said `OSPlatform.Whatever`.

cdmihai: Like `$([MSBuild]::IsOSPlatform(OSPlatform.Linux))`? There's a repetition of `OSPlatform`.

rainersigwald: I was thinking of a dumb translation layer: If project file contains `$([MSBuild]::IsOSPlatform(foo))`, just pretend it instead said `$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::foo)))`

cdmihai: I think that doing that translation requires a bigger, non local code change than just a new intrinsic method `IsOSPlatform(string platform)` that contains all the logic. As @AndyGerlicher mentioned offline, we could blindly call `TryParse` on `OSPlatform` and if it does not parse, then throw an exception where we blindly enumerate the accepted enum values. This way we do not care about the underlying implementation. Does this sound good?

rainersigwald: Yeah, that'd totally work for me.

cdmihai: It turns out `OSPlatform` is a struct, not an Enum :(. Replace everything I said with reflection. We'll just blindly reflect for the members, and print available members if the user specified one does not exist.

cdmihai: I'll also add `IsUnixLike`. True means non windows.

cdmihai:
@eerhardt Now available:
`[MSBuild]::IsOsPlatform(string platformString)` where `platformString` must be a property name from `OSPlatform`
`[MSBuild]::IsOsUnixLike`

PR #1955, opened 2017-04-07
Add dummy global property for project instance based builds

### LINKED ISSUE #1748: Out of proc nodes can end up using stale project evaluation data
The following failing test mimics a codepath in CPS:
```c#
        [Fact]
        public void Test()
        {
            var mainProjectContents =
@"<Project>

  <Target Name=""BuildOther"" Outputs=""@(ReturnValue)"">
    <MSBuild Projects=""{0}"" Targets=""Foo"">
      <Output TaskParameter=""TargetOutputs"" ItemName=""ReturnValue"" />
    </MSBuild>
  </Target>

</Project>";


            var p2pProjectContents =
@"<Project>
  <PropertyGroup>
    <P>InitialValue</P>
  </PropertyGroup>

  <Target Name=""Foo"" Outputs=""$(P)""/>

</Project>";

            string p2pProjectPath = Path.GetTempFileName();
            File.WriteAllText(p2pProjectPath, p2pProjectContents);

            var mainRootElement = ProjectRootElement.Create(XmlReader.Create(new StringReader(string.Format(mainProjectContents, p2pProjectPath))), collection);

            mainRootElement.FullPath = Path.GetTempFileName();
            mainRootElement.Save();

            var mainProject = new Project(mainRootElement, new Dictionary<string, string>(), MSBuildConstants.CurrentToolsVersion, collection);
            var mainInstance = mainProject.CreateProjectInstance(ProjectInstanceSettings.Immutable).DeepCopy(isImmutable: false);

            var manager = new BuildManager();

            var request = new BuildRequestData(mainInstance, new[] { "BuildOther" });

            var mockLogger = new MockLogger();
            var parameters = new BuildParameters()
            {
                DisableInProcNode = true,
                Loggers = new []{mockLogger}
            };

            manager.BeginBuild(parameters);

            var submission = manager.PendBuildRequest(request);

            var results = submission.Execute();
            Assert.True(results.OverallResult == BuildResultCode.Success);
            Assert.Equal("InitialValue", results.ResultsByTarget["BuildOther"].Items.First().ItemSpec);

            // setup and build new instance

            var p2pProject = new Project(p2pProjectPath, new Dictionary<string, string>(), MSBuildConstants.CurrentToolsVersion, collection);
            var p2pInstance = p2pProject.CreateProjectInstance(ProjectInstanceSettings.Immutable).DeepCopy(isImmutable: false);

            p2pInstance.SetProperty("P", "NewValue");

            mockLogger.ClearLog();

            request = new BuildRequestData(p2pInstance, new[] { "Foo" });
            submission = manager.PendBuildRequest(request);
            results = submission.Execute();

            Assert.True(results.OverallResult == BuildResultCode.Success);
            Assert.Equal("NewValue", results.ResultsByTarget["Foo"].Items.First().ItemSpec);

            manager.EndBuild();
        }
    }
```

It fails in the last assert with:
```
      Assert.Equal() Failure
                 (pos 0)
      Expected: NewValue
      Actual:   InitialValue
                 (pos 0)
```

The first build builds `mainProject`, and its P2P dependency, `p2pProject`. This causes the [ConfigCache](https://github.com/Microsoft/msbuild/blob/xplat/src/Build/BackEnd/Components/Caching/ConfigCache.cs) in both the main node and the out of proc node (OOP node) to have an entry for `p2pProject`.

The second build creates a new project instance over `p2pProject`, mutates it, and then builds it. However, the msbuild Scheduler [observes that it already has results for `p2pProject`](https://github.com/Microsoft/msbuild/blob/xplat/src/Build/BackEnd/Components/Scheduler/Scheduler.cs#L1554)  from the previous build and returns the cached results, thus not building the mutated `ProjectInstance` when it should have (communication with the OOP node is not even attempted).

The apparent fix for this is to use the flag `Microsoft.Build.Execution.BuildRequestDataFlags.ReplaceExistingProjectInstance` when constructing the `BuildRequestData`, which tells the `BuildManager` to [swap the old `ProjectInstance` with the new one and clear the results cache for that configuration](https://github.com/Microsoft/msbuild/blob/xplat/src/Build/BackEnd/BuildManager/BuildManager.cs#L1306-L1309). However, this does not work because only the caches in the main node are updated, while the ones in the OOP node get stale. This leads to the Scheduler requesting a build on the OOP node which then proceeds to build the stale `ProjectInstance` from the first build.

Potential fixes:
- Particular fix, by making `Microsoft.Build.Execution.BuildRequestDataFlags.ReplaceExistingProjectInstance` work across nodes. When this flag is used, the Scheduler needs to react and request that all OOP nodes that have that particular configuration get updated. Probably via a `Microsoft.Build.BackEnd.NodePacketType.BuildRequestConfiguration` packet.
- More general fix, by making the node communication protocol account for syncing up stale configurations. This way, the BuildManager would not only keep [a record of what nodes have what configurations](https://github.com/Microsoft/msbuild/blob/xplat/src/Build/BackEnd/BuildManager/BuildManager.cs#L113), but would keep a record of what version of configurations each node has. Then, when the caches in the main root get updated, the scheduler calls all nodes with stale configurations and updates them. This would make the `ReplaceExistingProjectInstance` flag obsolete. Or as an optimization to reduce pipe chatter, it could only update a node before issuing builds to that node.
- Update the project identity during a build so it's not just <global props, toolsversion, project path>. It should also take full state of items and properties into account
  - Quick hack: add a dummy global property to project instance based builds to force new cache identity.

cdmihai:
Resolves #1748 (detailed investigation and reasoning there), This PR goes with the dummy global property fix.

### DIFF
--- a/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
@@ -3904,5 +3904,93 @@ namespace Microsoft.Build.UnitTests.BackEnd
             rootProject.Save();
             return new ProjectInstance(rootProject);
         }
+
+        [Fact]
+        public void ShouldBuildMutatedProjectInstanceWhoseProjectWasPreviouslyBuiltAsAP2PDependency()
+        {
+            var mainProjectContents =
+@"<Project>
+
+  <Target Name=""BuildOther"" Outputs=""@(ReturnValue)"">
+    <MSBuild Projects=""{0}"" Targets=""Foo"">
+      <Output TaskParameter=""TargetOutputs"" ItemName=""ReturnValue"" />
+    </MSBuild>
+  </Target>
+
+</Project>";
+
+
+            var p2pProjectContents =
+@"<Project>
+
+  <PropertyGroup>
+    <P>InitialValue</P>
+  </PropertyGroup>
+
+  <Target Name=""Foo"" Outputs=""$(P)""/>
+
+</Project>";
+
+            using (var testFiles = new Helpers.TestProjectWithFiles(string.Empty, new[] { "p2p", "main" }))
+            using (var collection = new ProjectCollection())
+            using (var manager = new BuildManager())
+            {
+                try
+                {
+                    var p2pProjectPath = testFiles.CreatedFiles[0];
+                    File.WriteAllText(p2pProjectPath, p2pProjectContents);
+
+                    var mainRootElement = ProjectRootElement.Create(XmlReader.Create(new StringReader(string.Format(mainProjectContents, p2pProjectPath))), collection);
+
+                    mainRootElement.FullPath = testFiles.CreatedFiles[1];
+                    mainRootElement.Save();
+
+                    // build p2p project as a real p2p dependency of some other project. This loads the p2p into msbuild's caches
+
+                    var mainProject = new Project(mainRootElement, new Dictionary<string, string>(), MSBuildConstants.CurrentToolsVersion, collection);
+                    var mainInstance = mainProject.CreateProjectInstance(ProjectInstanceSettings.Immutable).DeepCopy(isImmutable: false);
+
+                    var request = new BuildRequestData(mainInstance, new[] { "BuildOther" });
+
+                    var parameters = new BuildParameters()
+                    {
+                        DisableInProcNode = true,
+                    };
+
+                    manager.BeginBuild(parameters);
+
+                    var submission = manager.PendBuildRequest(request);
+
+                    var results = submission.Execute();
+                    Assert.Equal(BuildResultCode.Success, results.OverallResult);
+                    Assert.Equal("InitialValue", results.ResultsByTarget["BuildOther"].Items.First().ItemSpec);
+
+                    // build p2p directly via mutated ProjectInstances based of the same Project.
+                    // This should rebuild and the result shold reflect the in-memory changes and not reuse stale cache info
+
+                    var p2pProject = new Project(p2pProjectPath, new Dictionary<string, string>(), MSBuildConstants.CurrentToolsVersion, collection);
+
+                    for (var i = 0; i < 2; i++)
+                    {
+                        var p2pInstance = p2pProject.CreateProjectInstance(ProjectInstanceSettings.Immutable).DeepCopy(isImmutable: false);
+
+                        var newPropertyValue = $"NewValue_{i}";
+
+                        p2pInstance.SetProperty("P", newPropertyValue);
+
+                        request = new BuildRequestData(p2pInstance, new[] { "Foo" });
+                        submission = manager.PendBuildRequest(request);
+                        results = submission.Execute();
+
+                        Assert.Equal(BuildResultCode.Success, results.OverallResult);
+                        Assert.Equal(newPropertyValue, results.ResultsByTarget["Foo"].Items.First().ItemSpec);
+                    }
+                }
+                finally
+                {
+                    manager.EndBuild();
+                }
+            }
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1293,31 +1293,57 @@ private BuildRequestConfiguration ResolveConfiguration(BuildRequestConfiguration
...
             return resolvedConfiguration;
         }
 
+        private void ReplaceExistingProjectInstance(BuildRequestConfiguration newConfiguration, BuildRequestConfiguration existingConfiguration)
+        {
+            existingConfiguration.Project = newConfiguration.Project;
+            _resultsCache.ClearResultsForConfiguration(existingConfiguration.ConfigurationId);
+        }
+
+        private BuildRequestConfiguration AddMewConfiguration(BuildRequestConfiguration unresolvedConfiguration)
```
AndyGerlicher: Typo `AddMew`

AndyGerlicher:
Can you put a little more description of the issue/workaround in the commit please?

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -3904,5 +3904,93 @@ private ProjectInstance GenerateDummyProjects(string shutdownProjectDirectory, i
...
+                    {
+                        DisableInProcNode = true,
+                    };
+
+                    manager.BeginBuild(parameters);
+
+                    var submission = manager.PendBuildRequest(request);
+
+                    var results = submission.Execute();
+                    Assert.True(results.OverallResult == BuildResultCode.Success);
```
rainersigwald: `Assert.Equal`?

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -3904,5 +3904,93 @@ private ProjectInstance GenerateDummyProjects(string shutdownProjectDirectory, i
...
+
+                        var newPropertyValue = $"NewValue_{i}";
+
+                        p2pInstance.SetProperty("P", newPropertyValue);
+
+                        request = new BuildRequestData(p2pInstance, new[] { "Foo" });
+                        submission = manager.PendBuildRequest(request);
+                        results = submission.Execute();
+
+                        Assert.True(results.OverallResult == BuildResultCode.Success);
```
rainersigwald: `Equal`?

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -835,7 +835,7 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
                 // Create/Retrieve a configuration for each request
                 BuildRequestConfiguration buildRequestConfiguration = new BuildRequestConfiguration(submission.BuildRequestData, _buildParameters.DefaultToolsVersion, _buildParameters.GetToolset);
                 BuildRequestConfiguration matchingConfiguration = _configCache.GetMatchingConfiguration(buildRequestConfiguration);
-                BuildRequestConfiguration newConfiguration = ResolveConfiguration(buildRequestConfiguration, matchingConfiguration, (submission.BuildRequestData.Flags & BuildRequestDataFlags.ReplaceExistingProjectInstance) == BuildRequestDataFlags.ReplaceExistingProjectInstance);
+                BuildRequestConfiguration newConfiguration = ResolveConfiguration(buildRequestConfiguration, matchingConfiguration, submission.BuildRequestData.Flags.HasFlag(BuildRequestDataFlags.ReplaceExistingProjectInstance));
```
rainersigwald: ❤️

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1293,31 +1293,57 @@ private BuildRequestConfiguration ResolveConfiguration(BuildRequestConfiguration
...
+            else if (unresolvedConfiguration.Project != null && resolvedConfiguration.Project == null)
+            {
+                // Workaround for https://github.com/Microsoft/msbuild/issues/1748
+                // If the submission has a project instance but the existing configuration does not, it probably means that the project was 
+                // built on another node (e.g. the project was encountered as a p2p reference and scheduled to a node).
+                // Add a dummy property to force cache invalidation in the scheduler and the nodes.
+                // TODO find a better solution than a dummy property
+
+                var key = $"Dummy{Guid.NewGuid():N}";
+                unresolvedConfiguration.Properties[key] = ProjectPropertyInstance.Create(key, "v");
```
rainersigwald: Name the property `ProjectInstance{Guid.NewGuid():N}` or similar, and set the value to something slightly descriptive?

cdmihai:
Final squashing before merge. Keeping these commits separate.

PR #1967, opened 2017-04-11
🚨 Fix shellcheck warnings in cibuild.sh

rainersigwald:
Classes of change:

* Quoting variable expansions to handle spaces better
* Switching from backticks to `$(..)` expansions
* Avoiding `egrep` in favor of `grep -E`
* Removed unused `BOOTSTRAPPED_RUNTIME_HOST`

### DIFF
--- a/cibuild.sh
+++ b/cibuild.sh
@@ -20,14 +20,14 @@ restoreBuildTools(){
 # home is not defined on CI machines
 setHome()
 {
-    if [ -z ${HOME+x} ]
+    if [ -z "${HOME+x}" ]
     then
         export HOME=$HOME_DEFAULT
-        mkdir -p $HOME_DEFAULT
+        mkdir -p "$HOME_DEFAULT"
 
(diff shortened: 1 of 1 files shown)


cibuild.sh
```diff
@@ -99,7 +99,7 @@ get_current_linux_name() {
         echo "centos"
         return 0
     elif [ "$(cat /etc/*-release | grep -cim1 rhel)" -eq 1 ]; then
-        if [ "$(cat /etc/os-release | grep -cim1 'VERSION_ID="7\.')" -eq 1 ]; then
+        if [ "$(grep -cim1 'VERSION_ID="7\.' /etc/os-release)" -eq 1 ]; then
```
rainersigwald: This is the only thing I have real questions about. It's better (fewer processes, no need for a pipe), but loses parallelism. Anyone care one way or the other?

jeffkl: I don't think it matters, the script is pretty fast and only runs once per build...

Issue #1684, opened 2017-02-09
MSBuild does not create a directory for the log file


jeffkl:
When you specify a log file path that contains a folder that does not exist, MSBuild gives an error:

```
msbuild /flp:logfile=%TEMP%\foo\foo.log
Microsoft (R) Build Engine version 14.0.25420.1
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : Logger error MSB4104: Failed to write to log file "C:\Users\jeffkl\AppData\Local\Temp\foo\foo.log". Could not find a part of the path 'C:\Users\jeffkl\AppData\Local\Temp\foo\foo.log'.
```

I wish MSBuild would create `%TEMP%\foo` so that I don't have to.

patros:
Working on this one now

PR #1970, opened 2017-04-12
Fix for MSBuild does not create a directory for the log file (#1684)

patros:
Making Jeff's wishes come true :)

### DIFF
--- a/src/Build.UnitTests/FileLogger_Tests.cs
+++ b/src/Build.UnitTests/FileLogger_Tests.cs
@@ -321,6 +321,28 @@ namespace Microsoft.Build.UnitTests
         }
 
         /// <summary>
+        /// Logging to a file in a directory that doesn't exists
+        /// </summary>
+        [Fact]
+        public void BasicNoExistingDirectory()
+        {
+            string directory = Path.Combine(ObjectModelHelpers.TempProjectDir, Guid.NewGuid().ToString("N"));
+            string log = Path.Combine(directory, "build.log");
+            Assert.False(Directory.Exists(directory));
+            Assert.False(File.Exists(log));
+
+            try
+            {
+                SetUpFileLoggerAndLogMessage("logfile=" + log, new BuildMessageEventArgs("message here", null, null, MessageImportance.High));
+                VerifyFileContent(log, "message here");
+            }
+            finally
+            {
+                ObjectModelHelpers.DeleteDirectory(directory);
+            }
+        }
+
+        /// <summary>
         /// Gets a filename for a nonexistent temporary file.
         /// </summary>
         /// <returns></returns>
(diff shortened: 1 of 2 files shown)


patros:

src/Build/Logging/FileLogger.cs
```diff
@@ -86,6 +86,8 @@ private void InitializeFileLogger(IEventSource eventSource, int nodeCount)
 
             try
             {
+                Directory.CreateDirectory(Path.GetDirectoryName(Path.GetFullPath(_logFileName)));
```
jeffkl: I can't think of a case when `GetFullPath()` would return something bad or if `GetDirectoryName()` would return null but I think I'd prefer to see some more guarding around these calls.  Either check some of the return values or put this line in its own `try...catch` which ignores the exception and lets the exception come from `FileUtilities.OpenWrite()` instead.

jeffkl:
@patros it looks like a test is failing but [it uses](https://github.com/Microsoft/msbuild/blob/6851538897f5d7b08024a6d8435bc44be5869e53/src/Build.UnitTests/FileLogger_Tests.cs#L451) `Assert.True` instead of `Assert.Equals`.  I'll submit a fix for this and then hopefully we can see why the test is failing after your change.

jeffkl:
@patros The test that's failing, `DistributedLoggerBadPath` should just be removed because its excepting the file logger to throw an exception when the directory does not exist.  Please delete this test along with your changes.

rainersigwald:
Just debugged some unit test failures that turned out to be caused by this PR. It was an amusing story!

This PR enables just creating the directory for log files to live in. That's awesome. But when combined with the (now bogus) `DistributedLoggerBadPath` test, that meant that a directory `/DONTEXIST` was created on the OS X CI machines. Then, on subsequent runs _without_ this PR, `DistributedLoggerBadPath` failed--because the directory did exist, so the path wasn't bad after all!

I submitted some PRs with a temporary commit to delete that directory to clean up the 2 CI machines where this job ran.

patros:
Yeah, that's not good. Sorry for the run around @rainersigwald

src/Build.UnitTests/FileLogger_Tests.cs
```diff
@@ -321,6 +321,33 @@ public void BasicExistingFileAppend()
         }
 
         /// <summary>
+        /// Logging to a file in a directory that doesn't exists
+        /// </summary>
+        [Fact]
+        public void BasicNoExistingDirectory()
+        {
+            string directory = GetTempDirectoryName();
```
patros: Should I update the existing GetTempFilename to use the TempProjectDir too?

patros: Actually there are a few cases where tests in this area are using the current directory instead of the TempProjectDir. Would you like me to have a look at changing those too? If so I'll create another pull request and keep this one issue specific.

src/Build/Logging/FileLogger.cs
```diff
@@ -86,6 +86,15 @@ private void InitializeFileLogger(IEventSource eventSource, int nodeCount)
 
             try
             {
+                try
+                {
+                    Directory.CreateDirectory(Path.GetDirectoryName(Path.GetFullPath(_logFileName)));
```
rainersigwald: Hmm. I see what Jeff was going for here, and I like the idea of isolating a failure of the path-manipulation stuff, but I don't like throwing away possibly-useful information like "you don't have permission to create that folder" that could be thrown from `CreateDirectory()`. Maybe something like:

```C#
string logDirectory = null;
try
{
    logDirectory = Path.GetDirectoryName(Path.GetFullPath(_logFileName));
}
catch
{
    // Directory creation is best-effort; if finding its path fails don't create the directory
    // and possibly let OpenWrite() below report the failure
}

if (logDirectory != null)
{
    Directory.CreateDirectory(logDirectory);
}
```

How's that sound?

patros: Yeah I like that. It should still be caught and logged by the existing mechanism then.

PR #1971, opened 2017-04-12
Update DistributedLoggerBadPath to use Assert.Equals

jeffkl:
Assert.True() only tells you if the condition was false but Assert.Equals() will tell us why the strings were different.

FYI @patros

### DIFF
--- a/src/Build.UnitTests/FileLogger_Tests.cs
+++ b/src/Build.UnitTests/FileLogger_Tests.cs
@@ -448,14 +448,14 @@ namespace Microsoft.Build.UnitTests
                                                           + "mylogfile.log");
 
                 fileLogger.Initialize(new EventSourceSink());
-                Assert.True(
-                              string.Compare(
-                                             fileLogger.InternalFilelogger.Parameters,
-                                             ";ShowCommandLine;logfile="
-                                             + Path.Combine(
-                                                            Directory.GetCurrentDirectory(),
-                                                            Path.DirectorySeparatorChar + "DONTEXIST" + Path.DirectorySeparatorChar + "mylogfile2.log"),
-                                             StringComparison.OrdinalIgnoreCase) == 0);
+
+                Assert.Equal(
+                    fileLogger.InternalFilelogger.Parameters,
+                    ";ShowCommandLine;logfile="
+                        + Path.Combine(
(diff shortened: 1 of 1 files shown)


src/Build.UnitTests/FileLogger_Tests.cs
```diff
@@ -448,14 +448,14 @@ public void DistributedLoggerBadPath()
...
-                                                            Path.DirectorySeparatorChar + "DONTEXIST" + Path.DirectorySeparatorChar + "mylogfile2.log"),
-                                             StringComparison.OrdinalIgnoreCase) == 0);
+
+                Assert.Equal(
+                    fileLogger.InternalFilelogger.Parameters,
+                    ";ShowCommandLine;logfile="
+                        + Path.Combine(
+                        Directory.GetCurrentDirectory(),
+                        Path.DirectorySeparatorChar + "DONTEXIST" + Path.DirectorySeparatorChar + "mylogfile2.log"), 
+                    StringComparer.OrdinalIgnoreCase);
```
rainersigwald: Also swap, so it's `expected, actual`?

src/Build.UnitTests/FileLogger_Tests.cs
```diff
@@ -448,14 +448,14 @@ public void DistributedLoggerBadPath()
...
-                                                            Directory.GetCurrentDirectory(),
-                                                            Path.DirectorySeparatorChar + "DONTEXIST" + Path.DirectorySeparatorChar + "mylogfile2.log"),
-                                             StringComparison.OrdinalIgnoreCase) == 0);
+
+                Assert.Equal(
+                    fileLogger.InternalFilelogger.Parameters,
+                    ";ShowCommandLine;logfile="
+                        + Path.Combine(
+                        Directory.GetCurrentDirectory(),
+                        Path.DirectorySeparatorChar + "DONTEXIST" + Path.DirectorySeparatorChar + "mylogfile2.log"),
```
rainersigwald: Doesn't the leading `DirectorySeparatorChar` make that part an absolute path, making the combine useless? Could this just be `Path.Combine( Directory.GetCurrentDirectory(), "DONTEXIST", "mylogfile2.log")`?

jeffkl:
Nevermind, we're just going to delete this test.

PR #1977, opened 2017-04-14
✅ Enable out-of-proc build tests on .NET Core

rainersigwald:
There were a couple of minor changes to product code, but for the most part this was just about enabling tests again.

Product code changes:
* Ignore an API-set `EnableNodeReuse` when the feature isn't present.
* Clean up the way the `OnLinkStatusChanged` event is invoked to be new and shiny (not strictly needed here, yell at me and I can remove it).
* When searching for the type of a method to use as a property function, start the search at the full class name, to make it happier with the micro-assemblies used in .NET Core.


src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -440,7 +440,11 @@ public bool ResetCaches
         public bool EnableNodeReuse
         {
             get { return _enableNodeReuse; }
-            set { _enableNodeReuse = value; }
+            set {
+#if FEATURE_NODE_REUSE
+                _enableNodeReuse = value;
```
cdmihai: What about erroring here instead of silently ignoring the new value? Could be too disruptive though, since the silent behaviour should still lead to a successful build ...

src/Build/Evaluation/Expander.cs
```diff
@@ -3315,7 +3315,7 @@ private static Type GetTypeForStaticMethod(string typeName, string simpleMethodN
             private static Type GetTypeFromAssemblyUsingNamespace(string typeName)
             {
                 string baseName = typeName;
-                int assemblyNameEnd = baseName.LastIndexOf('.');
+                int assemblyNameEnd = baseName.Length;
```
cdmihai: Is this reasonably testable? If yes, a test would be good.

src/Build/BackEnd/BuildManager/BuildParameters.cs
rainersigwald: I couldn't decide about that either, and eventually went this way with this justification: if we error instead of ignore here, we push the burden of changing behavior onto our consumers. If someone tries to write a build invoker that works for both of our flavors, _they_ would have to know whether the version of MSBuild they're invoking supported it. By ignoring it, we can take that responsibility for ourselves.

Convincing?

src/Build/Evaluation/Expander.cs
rainersigwald: It's incidentally tested by some of the tests I enabled (couldn't find `System.Diagnostics.Process` otherwise) but an explicit test sounds reasonable.

rainersigwald:
Waiting on @AndyGerlicher who had some (more important) changes here, so I'll let those go first and merge this PR later.

src/Build/BackEnd/BuildManager/BuildParameters.cs
cdmihai: Quite :)

rainersigwald:
Ok, I think this is reviewable again. Any takers?

cdmihai:
Still LGTM

Issue #670, opened 2016-06-01
Investigate+fix tests failing on Ubuntu/NetCore


cdmihai:
The failing tests have been disabled with the trait `netcore-linux-failing` in #667.

YuliiaKovalova:
Close it as not relevant anymore.

PR #2013, opened 2017-04-26
Remove netcore-linux-failing from some now passing tests

patros:
Started looking at issue #670 and these tests with ```netcore-linux-failing``` pass for me. I'm assuming other work has since fixed these tests so I'm really creating this to run the tests on your build servers to confirm they now pass. If that's the case I'll also remove the ```netcore-osx-failing``` and check again.


patros:
Ok, looks like these tests are passing on linux and osx .net core now. Assuming I'm not missing some other problem I'll keep checking for now passing tests. There are a lot though. What approach would you like me to take? Keep this pull request open and just keep adding to it? I've only made changes to BuildManager_tests.cs so far, maybe a pull request per file so the pull requests are too long running?

rainersigwald:
@AndyGerlicher, are you done with the changes you wanted to make to `BuildManager_Tests.cs`? I was waiting for you on #1977, and this touches that file too (though it should be an easy merge).

patros:
@jeffkl cool. I'll leave this pr as is and create another one when I've had a bit of time to go through some more.

PR #2014, opened 2017-04-26
Implement importing a list of files

### LINKED ISSUE #890: Allow Import to accept a delimited list of Project files
I'd like to build up a list of project files to import in an ItemList and import them. So, for example, this:

```
<Import Project="Foo.props" />
<Import Project="Bar.props" />
```

 would be equivalent to this:

```
<ItemGroup>
  <Projects Include="Foo.props;Bar.props"/>
</ItemGroup>
<Import Project="@(Projects)" />
```

Seems a natural construct... 

Warm regards,
Chris

jeffkl:
Allows `<Import Project="import1.props;import2.props" />` to work.  This also updates `Exists` so that this will work:

```xml
<Import Project="import1.props;import2.props"
   Condition="Exists('import1.props;import2.props')" />
```

It also works with properties:

```xml
<Project>

  <PropertyGroup>
    <Imports>import1.props</Imports>
    <Imports>$(Imports);import2.props</Imports>
  </PropertyGroup>

  <Import Project="$(Imports)" Condition=" '$(Imports)' != '' And Exists($(Imports))" />
</Project>
```
**It does not work with items** because items are not available when evaluating `<Import />` elements.

Closes #890

### DIFF
--- a/src/Build.UnitTests/ExpressionTreeExpression_Tests.cs
+++ b/src/Build.UnitTests/ExpressionTreeExpression_Tests.cs
@@ -336,11 +336,6 @@ namespace Microsoft.Build.UnitTests
             "existsX",
             "!",
             "nonexistentfunction('xyz')",
-            "exists('a;b')", /* non scalar */
-            "exists(@(z))",
-            "exists('@(z)')",
-            "exists($(a_semi_b))",
-            "exists('$(a_semi_b)')",
             "exists(@(v)x)",
             "exists(@(v)$(nonexistent))",
             "exists('@(v)$(a)')",
--- a/src/Build.UnitTests/ExpressionTreeExpression_Tests.cs
+++ b/src/Build.UnitTests/ExpressionTreeExpression_Tests.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Build.UnitTests
         private readonly ITestOutputHelper output;
 
 
-        private static readonly string[] FilesWithExistenceChecks = { "a", "a;b", "a'b", ";", "'" };
+        private static readonly string[] FilesWithExistenceChecks = { "a", "c", "a;b", "a'b", ";", "'" };
 
         private readonly Expander<ProjectPropertyInstance, ProjectItemInstance> _expander;
 
@@ -137,6 +137,8 @@ namespace Microsoft.Build.UnitTests
             "exists(a)",
             "exists('a%3bb')", /* semicolon */
             "exists('a%27b')", /* apostrophe */
+            "exists('a;c')", /* items */
+            "exists($(a_semi_c))",
             "exists($(a_escapedsemi_b))",
             "exists('$(a_escapedsemi_b)')",
             "exists($(a_escapedapos_b))",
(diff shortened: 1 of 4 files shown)


src/Build/Evaluation/Conditionals/FunctionCallExpressionNode.cs
```diff
@@ -38,20 +38,11 @@ internal override bool BoolEvaluate(ConditionEvaluator.IConditionEvaluationState
...
-
                 try
                 {
-                    if (state.EvaluationDirectory != null && !Path.IsPathRooted(expandedValue))
-                    {
-                        expandedValue = Path.GetFullPath(Path.Combine(state.EvaluationDirectory, expandedValue));
-                    }
+                    return !ExpandArgumentAsFileList((GenericExpressionNode) _arguments[0], state)
+                        .DefaultIfEmpty()
+                        .Any(i => i == null || state.LoadedProjectsCache?.TryGet(i) == null && !FileUtilities.FileOrDirectoryExistsNoThrow(i));
```
cdmihai: However much it hurts me to say this, I'd advise against Linq and variable capturing lambdas inside evaluation code. :(
This is not the hottest of paths, but it's still evaluation, which runs a lot. RPS might hiccup here.

cdmihai: Should it be Any or All? I'd keep Exists to check for all, and maybe add an ExistsAny function to check for any.

cdmihai: In `Exists("foo.cs;bar.cs")` I'd expect Exists to check that both files exist.

---
In reply to: [113551386](https://github.com/Microsoft/msbuild/pull/2014#discussion_r113551386) [](ancestors = 113551386)

cdmihai: Oops, the conditions inside the Any are actually negated, didn't see that. Could add a comment warning readers :)

---
In reply to: [113551764](https://github.com/Microsoft/msbuild/pull/2014#discussion_r113551764) [](ancestors = 113551764,113551386)

cdmihai: From a discussion with Jeff, this is actually not that bad. The lambda object is created once per method call, and the Linq expression isn't that long.

---
In reply to: [113551198](https://github.com/Microsoft/msbuild/pull/2014#discussion_r113551198) [](ancestors = 113551198)

src/Build.UnitTests/ExpressionTreeExpression_Tests.cs
```diff
@@ -336,11 +336,6 @@ public class ExpressionTest : IDisposable
             "existsX",
             "!",
             "nonexistentfunction('xyz')",
-            "exists('a;b')", /* non scalar */
```
cdmihai: Make sure to add a test where the semicolon is escaped

jeffkl: This test already does that.

AndyGerlicher:
Looks good! We'll hold until Monday or Tuesday, this shouldn't go in the first preview of Update 3 and we haven't branched yet.

PR #2016, opened 2017-04-26
Use AssemblyName.Clone when possible.

AndyGerlicher:
AssemblyName.Clone is not available in early versions of .NET Core.
Until we can upgrade MSBuild to use .NET Core 2.0, we can't call the
Clone method directly. This change will attempt to call .Clone using
reflection and if not available fallback to the previous .NET Core
behavior.

Potential fix for #2015 (when running on .NET Core 2.0).

### DIFF
--- a/src/Shared/AssemblyNameExtension.cs
+++ b/src/Shared/AssemblyNameExtension.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Build.Shared
     /// between the two is done lazily on demand.
     /// </summary>
     [Serializable]
-    sealed internal class AssemblyNameExtension
+    internal sealed class AssemblyNameExtension
     {
         private AssemblyName asAssemblyName = null;
         private string asString = null;
@@ -592,11 +592,7 @@ namespace Microsoft.Build.Shared
 
             if (asAssemblyName != null)
             {
-#if FEATURE_ASSEMBLYNAME_CLONE
-                newExtension.asAssemblyName = (AssemblyName)asAssemblyName.Clone();
-#else
-                newExtension.asAssemblyName = new AssemblyName(asAssemblyName.FullName);
-#endif
+                newExtension.asAssemblyName = asAssemblyName.CloneIfPossible();
             }
 
             newExtension.asString = asString;
--- a/src/Shared/AssemblyUtilities.cs
+++ b/src/Shared/AssemblyUtilities.cs
@@ -11,6 +11,13 @@ namespace Microsoft.Build.Shared
     /// </summary>
     internal static class AssemblyUtilities
     {
+        // True when the cached method info objects have been set.
+        private static bool s_initialized;
+
+        // Cached method info
+        private static MethodInfo s_assemblyNameCloneMethod;
+        private static PropertyInfo s_assemblylocationProperty;
+
         public static string GetAssemblyLocation(Assembly assembly)
         {
 #if FEATURE_ASSEMBLY_LOCATION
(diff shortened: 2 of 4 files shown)


src/Shared/AssemblyNameExtension.cs
```diff
@@ -592,11 +592,7 @@ internal AssemblyNameExtension Clone()
 
             if (asAssemblyName != null)
             {
-#if FEATURE_ASSEMBLYNAME_CLONE
-                newExtension.asAssemblyName = (AssemblyName)asAssemblyName.Clone();
-#else
-                newExtension.asAssemblyName = new AssemblyName(asAssemblyName.FullName);
-#endif
+                newExtension.asAssemblyName = asAssemblyName.Clone2();
```
rainersigwald: `CloneIfPossible`?

src/Shared/AssemblyUtilities.cs
```diff
@@ -39,5 +39,23 @@ public static Type GetTypeInfo(this Type t)
...
+
+        public static AssemblyName Clone2(this AssemblyName assemblyNameToClone)
+        {
+#if FEATURE_ASSEMBLYNAME_CLONE
+            return assemblyNameToClone.Clone();
+#else
+            AssemblyName result = null;
+            try
+            {
+                MethodInfo method = assemblyNameToClone?.GetType().GetMethod("Clone");
```
rainersigwald: Seems like this `MethodInfo` should be in a static lazy so we don't have to reflect in the hot path?

src/Shared/AssemblyUtilities.cs
```diff
@@ -39,5 +39,23 @@ public static Type GetTypeInfo(this Type t)
...
+                result = (AssemblyName) method?.Invoke(assemblyNameToClone, null);
+            }
+            catch (MemberAccessException)
```
rainersigwald: For that matter, catching here seems prone to slow.

src/Shared/AssemblyUtilities.cs
pakrym: Not caching reflection results is quite slow too. You can use static field to store `typeof(AssemblyName).GetType().GetMethod("Clone")` result.

src/Shared/AssemblyUtilities.cs
```diff
@@ -39,5 +39,23 @@ public static Type GetTypeInfo(this Type t)
             return t;
         }
 #endif
+
+        public static AssemblyName Clone2(this AssemblyName assemblyNameToClone)
```
cdmihai: `CreateClone`

src/Shared/AssemblyUtilities.cs
```diff
@@ -39,5 +39,23 @@ public static Type GetTypeInfo(this Type t)
...
+            AssemblyName result = null;
+            try
+            {
+                MethodInfo method = assemblyNameToClone?.GetType().GetMethod("Clone");
+                result = (AssemblyName) method?.Invoke(assemblyNameToClone, null);
+            }
+            catch (MemberAccessException)
+            { }
+
+            return result ?? new AssemblyName(assemblyNameToClone.FullName);
```
cdmihai: You could look over Clone's implementation and copy it here? :)

src/Shared/AssemblyUtilities.cs
cdmihai: Good point. @AndyGerlicher if you do this, can you please do it for `GetAssemblyLocation` too? :)

AndyGerlicher:

src/Shared/AssemblyUtilities.cs
```diff
@@ -39,5 +45,37 @@ public static Type GetTypeInfo(this Type t)
...
+                   new AssemblyName(assemblyNameToClone.FullName);
+#endif
+        }
+
+        /// <summary>
+        /// Initialize static fields. Doesn't need to be thread safe.
+        /// </summary>
+        private static void Initialize()
+        {
+            if (s_initialized) return;
```
rainersigwald: Is this guaranteed to be run single-threaded?

rainersigwald: I guess it doesn't really matter if some things race to plunk the same value down

AndyGerlicher:

PR #2018, opened 2017-04-27
Project instance serialization

### LINKED ISSUE #1726: ProjectInstance based build fails with a "Target not found exception" when the instance is mutated
```c#
        private const string projectContents =
@"<Project>
  <PropertyGroup>
    <ImportIt>true</ImportIt>
  </PropertyGroup>

  <Import Project=""{0}"" Condition=""'$(ImportIt)' == 'true'""/>

  <Target Name=""Bazz"">
    <Message Text=""Buzz"" Importance=""High"" />
  </Target>

</Project>";

        private const string projectImport =
@"<Project>
  <Target Name=""Foo"">
    <Message Text=""Bar"" Importance=""High"" />
  </Target>
</Project>";

        [Fact]
        public void Test1()
        {
            string importPath = Path.GetTempFileName();
            File.WriteAllText(importPath, projectImport);

            var root = ProjectRootElement.Create(XmlReader.Create(new StringReader(string.Format(projectContents, importPath))), collection);

            root.FullPath = Path.GetTempFileName();
            root.Save();

            var project = new Project(root, new Dictionary<string, string>(), MSBuildConstants.CurrentToolsVersion, collection);
            var instance = project.CreateProjectInstance(ProjectInstanceSettings.Immutable).DeepCopy(isImmutable: false);

            var manager = new BuildManager();

            var request = new BuildRequestData(instance, new[] { "Foo" });
            var parameters = new BuildParameters()
            {
                DisableInProcNode = true,
            };

            manager.BeginBuild(parameters);
            var submission = manager.PendBuildRequest(request);

            var results = submission.Execute();
            Assert.True(results.OverallResult == BuildResultCode.Success);

            manager.EndBuild();
            manager.ResetCaches();

            project.SetProperty("ImportIt", "false");
            project.Save();

            manager.BeginBuild(parameters);
            request = new BuildRequestData(instance, new[] { "Foo" });
            submission = manager.PendBuildRequest(request);
            results = submission.Execute();
            Assert.True(results.OverallResult == BuildResultCode.Success);

            manager.EndBuild();
        }
```

The second builds fails with a "target not found exception". 

There seem to be two weird things happening here:
1. The build is failing with a target not found exception, even if the build is started on a ProjectInstance. This is probably because a target is getting conditioned to false. It is my understanding that if you start a build on a project instance, msbuild shouldn’t re-evaluate and just reuse the evaluated state from the project instance.

2. MSBuild is ignoring the properties that CPS set on a project instance after the project instance info is serialized and sent to the out of proc node.

cdmihai:
Resolves #1726 by entirely translating the ProjectInstance and making the out of proc nodes use the translated project entirely as opposed to what they did before: re-evaluate from disk and merge the state with a translated minimal state consisting of only properties and items.

Later in the day I'll resolve the warnings, add more tests, make some translations prettier, etc


cibuild.cmd
```diff
@@ -151,7 +151,7 @@ set MSBUILDBINLOGPATH=%~dp0msbuild_rebuild-%HOST%.binlog
 echo.
 echo ** Rebuilding MSBuild with locally built binaries
 
-call "%~dp0build.cmd" /t:%TARGET_ARG% /p:Configuration=%BUILD_CONFIGURATION% %LOCALIZED_BUILD_ARGUMENT% "/nowarn:%_NOWARN%" /warnaserror /bl:%MSBUILDBINLOGPATH%
```
cdmihai: I'll remove this later in the day when i solve the warnings

ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -1077,12 +1077,12 @@ public partial class ProjectInstance
...
-        public System.Collections.Generic.List<string> DefaultTargets { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public System.Collections.Generic.List<string> DefaultTargets { get { throw null; } }
         public string Directory { [System.Diagnostics.DebuggerStepThroughAttribute]get { throw null; } }
         public System.Collections.Generic.List<Microsoft.Build.Construction.ProjectItemElement> EvaluatedItemElements { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public string FullPath { [System.Diagnostics.DebuggerStepThroughAttribute]get { throw null; } }
         public System.Collections.Generic.IDictionary<string, string> GlobalProperties { [System.Diagnostics.DebuggerStepThroughAttribute]get { throw null; } }
-        public System.Collections.Generic.List<string> InitialTargets { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public System.Collections.Generic.List<string> InitialTargets { get { throw null; } }
```
cdmihai: I think these changed because I added backing fields. I don't think it binary breaks people, right?

AndyGerlicher: Yeah that's added when you use auto get/set and is not breaking in either direction.

src/Shared/Traits.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public Traits()
+        {
+            EscapeHatches = new EscapeHatches();
+        }
+    }
+
+    internal class EscapeHatches
+    {
+        public bool SerializeEntireProjectInstance => string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBUILDDONOTSERIALIZEENTIREPROJECTINSTANCE"));
```
AndyGerlicher: If you're going to make this a common share thing it would be good to do caching here. Maybe just make it Lazy?

src/Shared/NodePacketTranslator.cs
```diff
@@ -207,10 +208,27 @@ public void Translate(ref string[] array)
...
+                {
+                    set.Add(_reader.ReadString());
+                }
+            }
+
             /// <summary>
             /// Translates a list of strings
             /// </summary>
-            /// <param name="list">The list to be translated.</param>
+            /// <param name="list"></param>
```
AndyGerlicher: ?

src/Shared/Traits.cs
AndyGerlicher: Never mind, read that wrong :)

---
In reply to: [113763543](https://github.com/Microsoft/msbuild/pull/2018#discussion_r113763543) [](ancestors = 113763543)

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -455,6 +440,48 @@ public ProjectInstance Project
...
+                _transferredProperties != null // need to overwrite particular properties, so load project from file and overwrite properties
+            )
+            {
+                Project = loadProjectFromFile.Invoke();
+            }
+            else
+            {
+                // projectInstance was serialized over. Finish initialization with node specific state
+
+                _project.ProjectRootElementCache = buildParameters.ProjectRootElementCache;
```
cdmihai: Opinion needed.

I had to create these two ProjectInstance properties. Initially they only got into a ProjectInstance via the constructor and represent per node specific object. Now that the PI is travelling between nodes, they have to be able to be set after the PI is constructor.

Should I create a special `LateInitialize` method on ProjectInstance to give a clear meaning to this concept? Or just leave it like is, via the new properties?

src/Shared/NodePacketTranslator.cs
cdmihai: ? indeed. No idea what happened, must have been the code elves stealing my comment :)

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
rainersigwald: I could go either way, but I think I lean toward `LateInitialize`. Then the information about what's done there lives in the class that has the information, rather than just in a comment here.

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -1017,5 +1052,6 @@ private INodePacketTranslator GetConfigurationTranslator(TranslationDirection di
                 throw;
             }
         }
+
```
rainersigwald: big ol nit: whitespace :)

src/Build/Instance/TaskRegistry.cs
```diff
@@ -681,35 +704,57 @@ ElementLocation elementLocation
...
+                _name = name;
 
                 // The ReadOnlyDictionary is a *wrapper*, the Dictionary is the copy.
-                TaskIdentityParameters = taskIdentityParameters == null ? null : new ReadOnlyDictionary<string, string>(new Dictionary<string, string>(taskIdentityParameters, StringComparer.OrdinalIgnoreCase));
+                _taskIdentityParameters = taskIdentityParameters == null ? null : new ReadOnlyDictionary<string, string>(CreateTaskIdentityParametersDictionary(taskIdentityParameters));
+            }
+
+            private static IDictionary<string, string> CreateTaskIdentityParametersDictionary(IDictionary<string, string> initialState = null, int? initialCount = null)
+            {
+                ErrorUtilities.VerifyThrowInvalidOperation(initialState == null || initialCount == null, "only one parameter can be non null");
```
rainersigwald: Only or at least?

src/Build.UnitTests/Construction/ElementLocation_Tests.cs
```diff
@@ -181,6 +181,24 @@ public void SerializationTest()
...
+            IElementLocation location = ElementLocation.EmptyLocation;
+
+            TranslationHelpers.GetWriteTranslator().Translate(ref location, ElementLocation.FactoryForDeserialization);
+            IElementLocation deserializedLocation = null;
+            TranslationHelpers.GetReadTranslator().Translate(ref deserializedLocation, ElementLocation.FactoryForDeserialization);
+
+            Assert.Equal(location.File, deserializedLocation.File);
+            Assert.Equal(location.Line, deserializedLocation.Line);
+            Assert.Equal(location.Column, deserializedLocation.Column);
+            Assert.True(location.GetType().FullName.Contains("SmallElementLocation"));
```
cdmihai: good catch

src/Build.UnitTests/BackEnd/TaskRegistry_Tests.cs
```diff
@@ -1827,6 +1828,146 @@ public void NullEvaluate()
...
+        [Theory]
+        [MemberData(nameof(TaskRegistryTranslationTestData))]
+        public void TaskRegistryCanSerializeViaTranslator(List<ProjectUsingTaskElement> usingTaskElements, Toolset toolset)
+        {
+            var original = CreateTaskRegistryAndRegisterTasks(usingTaskElements, toolset);
+
+            original.Translate(TranslationHelpers.GetWriteTranslator());
+
+            var copy = TaskRegistry.FactoryForDeserialization(TranslationHelpers.GetReadTranslator());
+            Assert.True(new TaskRegistryComparers.TaskRegistryComparer().Equals(original, copy));
```
cdmihai: You're right, yeah. :)

src/Build.UnitTests/Instance/ProjectInstance_Internal_Tests.cs
```diff
@@ -553,6 +558,116 @@ public void Build()
             mockLogger.AssertLogContains(new string[] { "Building...", "Completed!" });
         }
 
+        [Theory]
+        [InlineData(
+            @"      <Project>
+                    </Project>
+                ")]
+        [InlineData(
+            @"      <Project InitialTargets=`t1` DefaultTargets=`t2` ToolsVersion=`{0}`>
```
cdmihai: Smaller case that's easier to debug in case things blow up :)
Should I just delete it and leave the more complex project?

src/Build/Instance/TaskRegistry.cs
cdmihai: At most one can be non null. Will update the error comment

src/Build.UnitTests/Instance/ProjectInstance_Internal_Tests.cs
rainersigwald: No, just add a comment :)

PR #2025, opened 2017-04-28
Fix file path in item metadata, when CWD != project directory

radical:


radical:
Build fixed, all tests passing.

src/Build.UnitTests/FixPathOnUnix_Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+            var project = projColln.LoadProject(buildProjectPath);
+            var result = project.Build("Build");
+
+            Assert.True(result);
+
+            logger.AssertLogContains($"ItemMetadata: Md0 = {Path.Combine("lib", "foo.dll")}");
+        }
+    }
+
+    public class TestTask : Task
```
rainersigwald: Can you name this more descriptively? `LogItemMetadata`?

src/Shared/FileUtilities.cs
```diff
@@ -401,8 +401,10 @@ internal static string FixFilePath(string path)
         /// first segment exists and is a directory.
         /// Use a native shared method to massage file path. If the file is adjusted,
         /// that qualifies is as a path.
+        ///
+        /// @baseDirectory is just passed to LooksLikeUnixFilePath, to help with the check
         /// </summary>
-        internal static string MaybeAdjustFilePath(string value)
+        internal static string MaybeAdjustFilePath(string value, string baseDirectory="")
```
rainersigwald: I think our style prefers `string.Empty` to `""`.

radical: I had tried that first and got `error CS1736: Default parameter value for 'baseDirectory' must be a compile-time constant`.

rainersigwald: Huh! 💫

src/Build.UnitTests/FixPathOnUnix_Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+                    <Target Name='Build'>
+                        <LogTaskPropertiesTask Items='@(Item0)' />
+                    </Target>
+                </Project>";
+
+            string buildProjectPath = ObjectModelHelpers.CreateFileInTempProjectDirectory("build.proj", buildProjectContents);
+            string mainProjectPath = ObjectModelHelpers.CreateFileInTempProjectDirectory("projectDirectory/main.proj", mainProjectContents);
+            ObjectModelHelpers.CreateFileInTempProjectDirectory(Path.Combine("projectDirectory", "lib", "foo.dll"), "just a text file");
+
+            var projColln = new ProjectCollection();
```
cdmihai: It's sad we don't have a superclass for all our tests, but:

- `ProjectCollection` is disposable, please put it in a using block
- Pass a newly created `ProjectRootElementCache` to the `ProjectCollection` constructor, and then put the cache in the using block as well. Without this step, the default cache is used, which may then interact with other tests miss-using the global xml cache.

cdmihai: Also, use the Disposable `Microsoft.Build.UnitTests.Helpers.TestProjectWithFiles` instead of the static method. It automatically clears the created files in the Dispose.

PR #2027, opened 2017-04-28
Update System.Reflection.Metadata nuget reference to 1.4.1 from 1.3.0

radical:
.. this matches what Roslyn is using.

### DIFF
--- a/build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
+++ b/build/NuGetPackages/Microsoft.Build.Tasks.Core.nuspec
@@ -36,7 +36,7 @@
         <dependency id="System.Linq.Parallel" version="4.0.1" />
         <dependency id="System.ObjectModel" version="4.0.12" />
         <dependency id="System.Reflection" version="4.1.0" />
-        <dependency id="System.Reflection.Metadata" version="1.3.0" />
+        <dependency id="System.Reflection.Metadata" version="1.4.1" />
         <dependency id="System.Reflection.TypeExtensions" version="4.1.0" />
         <dependency id="System.Resources.ResourceManager" version="4.0.1" />
         <dependency id="System.Resources.Writer" version="4.0.0" />
@@ -66,7 +66,7 @@
         <dependency id="Microsoft.Build.Framework" version="[$version$]" />
         <dependency id="Microsoft.Build.Utilities.Core" version="[$version$]" />
         <dependency id="System.Collections.Immutable" version="1.3.1" />
-        <dependency id="System.Reflection.Metadata" version="1.3.0" />
+        <dependency id="System.Reflection.Metadata" version="1.4.1" />
         <dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.3.0" />
       </group>
     </dependencies>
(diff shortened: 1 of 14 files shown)


radical:
First one:
```
Microsoft.Build.Exceptions.InternalLoggerException : The build stopped unexpectedly because the "BinaryLogger" logger failed unexpectedly during initialization.\n---- System.IO.FileLoadException : Could not load file or assembly 'System.IO.Compression, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

AFAICS, this is because now `System.Reflection.Metadata 1.4.1` is pulling in `System.IO.Compression, 4.3.0` (which has assembly version `4.1.1.0`). But `Microsoft.Buil.dll` is still built against the older version. Updating that, fixes this test

```
diff --git a/targets/runtimeDependencies/project.json b/targets/runtimeDependencies/project.json
index 00506d76..b26ad8b9 100644
--- a/targets/runtimeDependencies/project.json
+++ b/targets/runtimeDependencies/project.json
@@ -11,7 +11,7 @@
     "xunit": "2.1.0",
     "microsoft.xunit.netcore.extensions": "1.0.0-prerelease-00629-09",
     "System.IO.FileSystem": "4.0.1",
-    "System.IO.Compression": "4.1.0"
+    "System.IO.Compression": "4.3.0"
   },
   "frameworks": {
     "net46": {
```

AndyGerlicher:
Unfortunately updating a package like this is complicated. On the .NET Core side currently we're targeting 1.0, moving to this will pull in packages from 1.1. We're actually getting upgraded to 2.0 in the dotnet CLI so this likely wouldn't be an issue, but it will cause things like the issue you noted above.

On the Full Framework side we have to be inline with what `devenv.exe.config` redirects to. We potentially could go past that since they redirect older versions, but it's something we have to consider.

I'd like to hold off on this or other package updates if we can until we move to .NET Core 2.0 (with the SDK project migration). Date still TBD at this point.

AndyGerlicher:
Team Triage:
On the VS side this look correct. And this should not affect the CLI as they are already higher than this anyway (.NET Core 2.0). @radical can you fix the merge conflicts and we'll get this in? Thanks!

AndyGerlicher:
Odd failures. I think this is because it's upping the version for .NET Core which is bringing in other package updates as dependencies. Maybe scope the update for net46? Hopefully we should be able to solve these types of things when we're able to upgrade to .NET Core 2.0.

cdmihai:
@dotnet-bot test this please

Failures:
https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_OSX_CoreCLR_prtest/66/
https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Ubuntu14.04_CoreCLR_prtest/65/
https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Ubuntu16.04_CoreCLR_prtest/66/
https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_CoreCLR_prtest/71/

AndyGerlicher:
Team Triage:
Closing this for now. We're concerned that the update brings in too many updated references. When we move to a later version of .NET Core this issue should be resolved.

PR #2031, opened 2017-04-29
Switch to producing portable PDBs instead of classic PDBs.

KirillOsenkov:
Embed all source files directly into the PDBs.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -176,6 +176,7 @@
   <!-- Set the host specific OS and Architecture variables -->
   <PropertyGroup>
     <Deterministic>true</Deterministic>
+    <DebugType>portable</DebugType>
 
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
@tmat @gafter @jaredpar folks, is this the right usage for embedding sources into portable PDBs? Does the debugger and tooling support this? When I debug using these pdbs, will the source open automatically?

+==========================+
| Old .pdb size: | 3,096,064 |
| New .pdb size (including all embedded source): | 1,938,484 |
| New .pdb size (without source): | 716,556 |
+==========================+

KirillOsenkov:
Turns out this can't go in because the debugger doesn't yet support sources embedded in pdbs. Was worth a try though :P

PR #2037, opened 2017-05-01
Generate Portable PDBs and add SourceLink.

KirillOsenkov:
Switch to generating portable PDBs (instead of classic Windows PDBs). These are the future and since we're cross-platform as well it's good to switch.

Add https://github.com/ctaggart/SourceLink to the MSBuild repo. It injects a new target using `CompileDependsOn`. The `SourceLinkCreate` target determines the current git SHA, the origin remote URL and generates a `SourceLink.json` file that gets passed to `csc.exe` via the new `/sourcelink` switch. The file gets generated to: `C:\msbuild\bin\obj\Microsoft.Build.Framework\Debug\sourcelink.json`

Sample contents of the `SourceLink.json` file:

```
{"documents":{"C:\\msbuild\\*":"https://raw.githubusercontent.com/KirillOsenkov/msbuild/c57e0bd2d9014280d2f006826a1de85b6a32d481/*"}}
```

It will use the fork's URL for private builds but presumably it will take the official `https://github.com/Microsoft/MSBuild` for the CI server and official builds.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -176,6 +176,8 @@
   <!-- Set the host specific OS and Architecture variables -->
   <PropertyGroup>
     <Deterministic>true</Deterministic>
+    <DebugType>portable</DebugType>
+    <SourceLinkCreate>true</SourceLinkCreate>
 
     <!-- Possible Values: win7 / osx.10.10.  For linux, the value is calculated in cibuild.sh and passed at the command-line -->
     <RuntimeSystem Condition="'$(TargetPlatformIdentifier)' == 'Windows_NT'">win7</RuntimeSystem>
--- a/src/Build/project.json
+++ b/src/Build/project.json
@@ -3,6 +3,7 @@
     "net46": {
       "dependencies": {
         "Microsoft.VisualStudio.Setup.Configuration.Interop": "1.2.304-preview5",
+        "SourceLink.Create.CommandLine": "2.1.0",
         "System.Collections.Immutable": "1.3.1",
         "System.Threading.Tasks.Dataflow": "4.5.24.0"
       }
--- a/src/Framework/project.json
+++ b/src/Framework/project.json
@@ -4,6 +4,7 @@
     "netstandard1.3": {
       "dependencies": {
         "NETStandard.Library": "1.6.0",
+        "SourceLink.Create.CommandLine": "2.1.0",
         "System.Runtime.Serialization.Primitives": "4.1.1",
         "System.Threading.Thread": "4.0.0"
       }
--- a/src/MSBuild/project.json
+++ b/src/MSBuild/project.json
@@ -8,6 +8,7 @@
     "netcoreapp1.0": {
       "dependencies": {
         "Microsoft.NETCore.App": "1.0.1",
+        "SourceLink.Create.CommandLine": "2.1.0",
         "System.Collections.NonGeneric": "4.0.1",
         "System.Diagnostics.FileVersionInfo": "4.0.0",
         "System.Diagnostics.Process": "4.1.0",
--- a/src/MSBuildTaskHost/project.json
+++ b/src/MSBuildTaskHost/project.json
@@ -1,6 +1,7 @@
 {
   "dependencies": {
-    "Microsoft.VisualStudio.Setup.Configuration.Interop":  "1.2.304-preview5"
+    "Microsoft.VisualStudio.Setup.Configuration.Interop":  "1.2.304-preview5",
+    "SourceLink.Create.CommandLine": "2.1.0"
   },
   "frameworks": {
     "net35": { }
--- a/src/Tasks/project.json
+++ b/src/Tasks/project.json
@@ -3,6 +3,7 @@
     "net46": {
       "dependencies": {
         "Microsoft.VisualStudio.Setup.Configuration.Interop": "1.2.304-preview5",
+        "SourceLink.Create.CommandLine": "2.1.0",
         "System.Collections.Immutable": "1.3.1",
         "System.Reflection.Metadata": "1.3.0"
       }
--- a/src/Utilities/project.json
+++ b/src/Utilities/project.json
@@ -8,6 +8,7 @@
     "netstandard1.3": {
       "dependencies": {
         "NETStandard.Library": "1.6.0",
+        "SourceLink.Create.CommandLine": "2.1.0",
         "System.Collections.NonGeneric": "4.0.1",
         "System.Diagnostics.Process": "4.1.0",
         "System.Diagnostics.TraceSource": "4.0.0",
--- a/src/dir.targets
+++ b/src/dir.targets
@@ -111,5 +111,5 @@
 
   <!-- https://github.com/ctaggart/SourceLink - stamps the portable .pdbs with URLs to sources on GitHub 
   so that debugger can download the exact matching source files during debugging -->
-  <Import Project="$(PackagesDir)\sourcelink.create.commandline\2.1.0\build\SourceLink.Create.CommandLine.targets" Condition="'$(IsTestProject)' != 'true' And Exists('$(PackagesDir)\sourcelink.create.commandline\2.1.0\build\SourceLink.Create.CommandLine.targets')" />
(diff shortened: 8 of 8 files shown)


KirillOsenkov:
CC @tmat @ctaggart

KirillOsenkov:
Wow, it works:
image

It caches the files in a path like: `C:\Users\Kirill\AppData\Local\SourceServer\eab72c02195fa3683f709167f655c4f24ffd05d458c0717a70a16aa9fbdee67d\src\Shared\CommunicationsUtilities.cs`

tmat:
Yes, it does :) I'm using it in https://github.com/dotnet/symreader-converter repo:
https://github.com/dotnet/symreader-converter/blob/master/src/Directory.Build.targets#L6

AndyGerlicher:
Clearly it's been a while since we've updated our bootstrapped MSBuild. :( We won't be able to take this until we do the SDK work to upgrade our projects and build with a newer bootstrapper. The other option is to put it behind `'$(MSBuildRuntimeType)' == 'Full'` and have it only be available for our full framework build.

KirillOsenkov:
Just pushed a fix to only import the SourceLink targets on full framework.

KirillOsenkov:
@dotnetbot retest all please

KirillOsenkov:
@dotnetbot test OSX Build for CoreCLR please
@dotnetbot test Ubuntu14.04 Build for CoreCLR please
@dotnetbot test Ubuntu16.04 Build for CoreCLR please
@dotnetbot test Windows_NT Build for CoreCLR please

KirillOsenkov:
@dotnetbot test OSX Build for CoreCLR please

KirillOsenkov:
Hmm, any idea why the CoreCLR builds are failing?

cdmihai:
@dotnet-bot test this please

KirillOsenkov:
Let's get this in? Pretty please? ;)

ctaggart:
Out of curiosity, are you going to include the portable pdb in the nupkg package? How are you [testing](https://github.com/ctaggart/SourceLink#test) that it works?

rainersigwald:
@KirillOsenkov I see from https://github.com/dotnet/core/blob/master/Documentation/diagnostics/portable_pdb.md#supported-scenarios that "symbol server" is one of the currently-not-supported scenarios. Surely we can't break that scenario . . .

KirillOsenkov:
Great catch @rainersigwald! Where and how do we upload .pdbs currently?

rainersigwald:
(Sorry for internal links, external folks)

In our official build, we index to symweb: https://devdiv.visualstudio.com/DevDiv/MSBuild/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=4688, though that's apparently a deprecated process.

When we insert binaries to VS, we use the [insert Feature into VS](https://devdiv.visualstudio.com/DevDiv/_apps/hub/ms.vss-releaseManagement-web.hub-metatask?taskGroupId=06b047fe-e88a-43d1-8958-f36b1f1b88c9&_a=properties) flow with "Submit to Internet" turned on which results in (stripped, I think) symbols going to the public Microsoft symbol servers.

tmat:
To address both scenarios: having Portable PDBs in nupkg and publish Windows PDBs to symsrv, I'd recommend to use [Pdb2Pdb](https://github.com/dotnet/symreader-converter). 
The tool package is available here: https://dotnet.myget.org/feed/symreader-converter/package/nuget/Pdb2Pdb 
Using this tool you can switch your build to produce Portable PDBs. Your symbol server publishing script would convert them to Windows PDBs before running symsrv.

tmat:
Alternatively you can consider using embedded PDBs. Then you don't need to publish symbols at all.

rainersigwald:
@tmat embedded PDBs certainly sound tempting! Do you know if that would cause someone in VS to yell at us for increased image sizes?

KirillOsenkov:
Unfortunately the steps to publish symbols seem too complicated to me and I'd be scared to attempt to change anything in that area. 

Embedding the .pdbs is interesting, but it will increase the size of the MSBuild .dlls (albeit not by a lot). Not sure if this is an OK solution.

As an alternative, we could abandon this PR and try https://github.com/AArnott/PdbGit or https://github.com/GitTools/GitLink. Their approach is to stamp the Windows PDBs with instructions to download the sources from GitHub directly. The advantage there is that the PDBs are still compatible with uploading everywhere.

KirillOsenkov:
I wonder if there's any noticeable perf impact from embedding .pdbs beside image size (when images are loaded they eat memory). 

But wow, embedding the PDBs in terms of file sizes:

Assembly | Before | After
-----------|---------|-------
Microsoft.Build.Framework.dll | 78,336 | 90,624
Microsoft.Build.dll | 1,582,592 | 1,881,600
Microsoft.Build.Tasks.Core | 1,291,776 | 1,521,152
Microsoft.Build.Utilities.Core | 353,792 | 424,960

tmat:
The extra memory is reserved but not committed.

KirillOsenkov:
I've submitted a new PR: https://github.com/Microsoft/msbuild/pull/2104

It uses https://github.com/AArnott/PdbGit to stamp Windows classic .pdbs. That way we won't have to increase the size of the binaries.

Let me know what you think.

ctaggart:
> Unfortunately the steps to publish symbols seem too complicated to me and I'd be scared to attempt to change anything in that area.

Someone in Microsoft really needs to figure out the story for publishing Portable PDBs. I think @tmat is working on source link support for Windows PDB files as well as a tool to convert between the two formats. Publishing Windows only PDB files to your symbol servers isn't ideal when I'm running on dotnet core on Linux.

> When we insert binaries to VS, we use the insert Feature into VS flow with "Submit to Internet" turned on which results in (stripped, I think) symbols going to the public Microsoft symbol servers.

Have you tested this process? I think it modifies the Windows PDB files again and may not work.

The ideal scenario would be if you built and included Portable PDB files in the nuget packages and then tweaked the internal process to work, perhaps converting the Portable PDB to a Windows PDB if required for some of your legacy tools.

KirillOsenkov:
@ctaggart Cameron, first off, we do truly appreciate the work you're doing to improve the debugging experience for our ecosystem and reduce friction. 

Please don't get discouraged by temporary setbacks we're hitting. Let me explain. Inside Microsoft there's a company-wide symbol server called symweb where we publish pdbs. There are very old, complicated, archaic systems around signing and uploading symbols. You haven't seen the docs that @rainersigwald pointed me to, it's... complicated. Moreover these systems are directly impacting Visual Studio product, release processes, etc. We can't risk breaking all that by a quick change.

We do realize that long term we need to work to migrate off the legacy system, but right now nobody on this thread is funded or empowered to do that. We have about as much impact on these processes as you do. We all want the right thing, and we do love portable PDBs and source linking with all our heart.

However we have a problem at hand, the one that needs an urgent solution. I debug into MSBuild all the time, and I can save a lot of time if I don't have to go digging for the right sources every time I debug. Right now I'm proposing a tactical fix, that can quickly solve the immediate problem and move forward. At the same time, we will try to fight the internal battles to modernize symweb and move us off the legacy pdbs. However it's going to take a while so PdbGit seems like a good pragmatic temporary solution. We're just not there yet because of the constraints of this particular codebase (MSBuild) and how it builds and how it gets inserted into Visual Studio.

I hope this gives a good clarification, so let's now wait for the verdict from @AndyGerlicher and the rest of the MSBuild team.

rainersigwald:
^ Exactly!

However, I think there's a compromise position that can get us into a pretty good state until (Microsoft-internal) http://symweb and http://msdl.microsoft.com/download/symbols support Portable PDBs: we could use SourceLink and embed into our .NET Core binaries, while using PdbGit for full framework binaries to avoid the larger-binaries problem, while keeping symbols available and getting source downloads working nicely.

Does that seem like a reasonable approach @KirillOsenkov @AndyGerlicher?

KirillOsenkov:
3rd time is a charm: how about this?

https://github.com/Microsoft/msbuild/pull/2107

rainersigwald:
Let's do #2107.

Issue #1986, opened 2017-04-17
Support reference assemblies in build flows


rainersigwald:
The C# and VB compilers are building support for emitting a _reference assembly_ while compiling. The reference assembly contains the accessible interface of the assembly, but _no implementation details_. That means it changes less often than the full assembly--many common development activities don't change the interface, only the implementation. That means that incremental builds can be much faster--if you invert the sense of an `if` in a common library, today you must rebuild every assembly that references it, but with reference assemblies, you would only rebuild the library and copy it around to a few new output directories.

[Roslyn spec for reference assemblies](https://github.com/dotnet/roslyn/blob/features/refout/docs/features/refout.md). Initial implementation: https://github.com/dotnet/roslyn/pull/17558

These assemblies should be supported in MSBuild scenarios.

## Design criteria

- [ ] Create reference assemblies by default when using deterministic C#/VB compilation
- [ ] Use reference assemblies (when available) instead of implementation assemblies when invoking
  - [ ] the compiler
  - [ ] other tools? (resgen?)
- [ ] Targets that use ref assemblies are marked up to date and skipped when a reference's implementation assembly has changed but its ref assembly has not
- [ ] Final output folders must still contain implementation assemblies
- [ ] Project-to-project (p2p) references must be resolvable into both implementation and reference assemblies
- [ ] Design-time builds in Visual Studio build against the same assembly as the command line
- [ ] Build augmentations that are unaware of reference assemblies
  - [ ] continue to work (against implementation assemblies)
  - [ ] are able to learn about implementation assemblies if they want the incrementality benefits
- [ ] _Possible_ Consider how this interacts with NuGet reference assemblies. Can the concepts be unified?

rainersigwald:
Implementation option: find ref assembly by filename transformation convention.

For example: after RAR and before `CoreCompile`, sub `@(ReferencePath->'%(RootDir)%(Directory)\ref\%(Filename)%(Extension)')` into the `ReferencePath` item if it exists.

* ➕ minimal MSBuild dataflow changes
* ➖ convention based (the compiler option allows arbitrary naming of the ref asm)

rainersigwald:
Implementation option: pass ref assembly path as metadata on the primary compiler output

For example: after RAR and before `CoreCompile`, sub `@(ReferencePath->'%(RootDir)%(Directory)\ref\%(Filename)%(Extension)')` into the `ReferencePath` item if it exists.

* ➕ allows user flexibility in path conventions
[...]

PR #2039, opened 2017-05-01
Common.targets support for reference assemblies

rainersigwald:
Provides an item, `ReferencePathWithInterfaceOnlyAssemblies`, that
consists of the reference (interface-only) versions of assemblies where
available, allowing a consuming task to be incrementally up-to-date
after implementation-only changes have been made to a reference.

When `%(ReferenceAssembly)` metadata is unavailable for a given
reference, the new item contains the implementation assembly directly.

Additionally creates a new output item for the current project's
reference assembly by default when `$(Deterministic)` is `true`. If this
is created, it is copied to the output directory using the new
`CopyRefAssembly` task, which copies only if the ref assembly is
different from the current file. If present, the ref asm is passed to
the project's consumes in metadata.

See #1986, dotnet/roslyn#2184.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -4038,6 +4038,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             >
 
       <Output TaskParameter="DestinationPath" ItemName="ReferenceAssembly"/>
+      <Output TaskParameter="DestinationFiles" ItemName="FileWrites"/>
 
     </CopyRefAssembly>
 
@@ -4502,6 +4503,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <!--Record the main compile outputs.-->
     <ItemGroup>
       <FileWrites Include="@(IntermediateAssembly)" Condition="Exists('@(IntermediateAssembly)')"/>
+      <FileWrites Include="@(IntermediateRefAssembly)" Condition="'@(IntermediateRefAssembly)' != '' and Exists('@(IntermediateRefAssembly)')"/>
     </ItemGroup>
 
     <!-- Record the .xml if one was produced. -->
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2060,6 +2060,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       </ReferencePathWithRefAssemblies>
       <ReferencePathWithRefAssemblies Include="@(ReferencePath)"
                                       Condition="'$(CompileUsingReferenceAssemblies)' == 'false'" />
+
+      <!-- The project system normally looks only at the inputs to the compiler in its
+           fast up-to-date check, but if the implementation assembly of a ref is stale
+           we still want to rerun to ensure copies, etc. happen. -->
+      <UpToDateCheckInput Include="@(ReferencePathWithRefAssemblies->'%(OriginalPath)')" />
     </ItemGroup>
   </Target>
 
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1372,9 +1372,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       AfterResolveReferences
     </ResolveReferencesDependsOn>
   </PropertyGroup>
+
+  <!-- FindReferenceAssembliesForReferences needs to see a fully-populated @(ReferencePath),
+       so it should run after all of the subcomponents of ResolveReferences, including
+       user overrides of AfterResolveReferences and targets added by appending to
+       $(ResolveReferencesDependsOn) (like ImplicitlyExpandDesignTimeFacades). So list it
+       last explicitly, rather than adding it to the property.   -->
   <Target
       Name="ResolveReferences"
-      DependsOnTargets="$(ResolveReferencesDependsOn)"/>
+      DependsOnTargets="$(ResolveReferencesDependsOn);FindReferenceAssembliesForReferences"/>
 
   <!--
     ============================================================
@@ -2028,8 +2034,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="FilesWritten" ItemName="FileWrites"/>
       <Output TaskParameter="DependsOnSystemRuntime" PropertyName="DependsOnSystemRuntime"/>
     </ResolveAssemblyReference>
+  </Target>
 
-    <ItemGroup>
+  <!--
+    ============================================================
+
+                                        FindReferenceAssembliesForReferences
+
+    Given the list of references, create a list of assemblies to pass to the compiler that
+    includes reference assemblies rather than implementation assemblies where possible.
+
+        [IN]
+        @(ReferencePath) - List of assembly references as resolved paths with ReferenceAssembly metadata
+
+        [OUT]
+        @(ReferencePathWithRefAssemblies) - Paths to resolved reference (or implementation) assemblies.
+    ============================================================
+    -->
+  <Target Name="FindReferenceAssembliesForReferences"
+          DependsOnTargets="$(ResolveReferencesDependsOn)">
+      <ItemGroup>
       <ReferencePathWithRefAssemblies Include="@(ReferencePath->'%(ReferenceAssembly)')" />
     </ItemGroup>
   </Target>
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2032,6 +2032,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </Target>
 
   <ItemDefinitionGroup>
+    <!-- Reference assemblies are not produced in all cases, but it's easier to consume them
+         if this metadatum is always populated. This item definition ensures that it points
+         to the implementation assembly unless specified. -->
     <ReferencePath>
       <ReferenceAssembly>%(FullPath)</ReferenceAssembly>
     </ReferencePath>
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2030,6 +2030,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="SerializationAssemblyFiles" ItemName="_ReferenceSerializationAssemblyPaths"/>
       <Output TaskParameter="ScatterFiles" ItemName="_ReferenceScatterPaths"/>
       <Output TaskParameter="CopyLocalFiles" ItemName="ReferenceCopyLocalPaths"/>
+      <!-- Indicate to the IDE that updated (possibly transitive) references should trigger
+           an (incremental) build. -->
+      <Output TaskParameter="CopyLocalFiles" ItemName="UpToDateCheckInput"/>
       <Output TaskParameter="SuggestedRedirects" ItemName="SuggestedBindingRedirects"/>
       <Output TaskParameter="FilesWritten" ItemName="FileWrites"/>
       <Output TaskParameter="DependsOnSystemRuntime" PropertyName="DependsOnSystemRuntime"/>
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2030,7 +2030,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ResolveAssemblyReference>
 
     <ItemGroup>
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -350,6 +352,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <FinalDocFile Include="@(DocFileItem->'$(OutDir)%(Filename)%(Extension)')"/>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(Deterministic)' == 'true'">
```
rainersigwald: Needs an explicit "no ref asm" off switch, as well as defaulting for `Deterministic`.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -350,6 +352,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <!-- TODO: should this be configurable? Default path obeys conventions. -->
```
rainersigwald: I'm thinking "no", let the user define the item themselves if they want a nonstandard path.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Condition must also apply to `$(TargetRefPath)`.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2015,8 +2025,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     </ResolveAssemblyReference>
+
+    <ItemGroup>
+      <ReferencePathWithInterfaceOnlyAssemblies Include="@(ReferencePath->'%(ReferenceAssembly)')" />
+    </ItemGroup>
   </Target>
 
+  <ItemDefinitionGroup>
+    <ReferencePath>
+      <ReferenceAssembly>%(FullPath)</ReferenceAssembly>
```
rainersigwald: Comment this.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4007,6 +4027,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     </Copy>
 
+    <!-- Copy the build product (.dll or .exe). -->
+    <CopyRefAssembly
+        SourcePath="@(IntermediateRefAssembly)"
+        DestinationPath="$(TargetRefPath)"
+        Condition="'$(CopyBuildOutputToOutputDirectory)' == 'true' and '$(SkipCopyBuildProduct)' != 'true' and Exists('@(IntermediateRefAssembly)')"
+            >
+
+      <Output TaskParameter="DestinationPath" ItemName="ReferenceAssembly"/>
```
rainersigwald: Add output to the outputs-of-this-project item so clean works right.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -281,6 +281,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <!-- Example, c:\MyProjects\MyProject\bin\debug\MyAssembly.dll -->
     <TargetPath Condition=" '$(TargetPath)' == '' ">$(TargetDir)$(TargetFileName)</TargetPath>
 
+    <TargetRefPath Condition=" '$(TargetRefPath)' == '' and '$(Deterministic)' == 'true' ">$([System.IO.Path]::Combine(`$([System.IO.Path]::GetDirectoryName($([System.IO.Path]::GetFullPath(`$(TargetPath)`))))`, 'ref', `$(TargetFileName)`))</TargetRefPath>
```
jeffkl: Can this just use `$(TargetDir)`?  

Then you could use `$([MSBuild]::NormalizePath($(TargetDir), 'ref', $(TargetFileName)))`

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1797,6 +1806,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <TargetPathWithTargetPlatformMoniker Include="$(TargetPath)">
         <TargetPlatformMoniker>$(TargetPlatformMoniker)</TargetPlatformMoniker>
         <TargetPlatformIdentifier>$(TargetPlatformIdentifier)</TargetPlatformIdentifier>
+        <ReferenceAssembly Condition="'$(TargetRefPath)' != ''">$(TargetRefPath)</ReferenceAssembly>
```
jeffkl: Does this need to be conditioned?  It seemed unusual to me, most other places just copy the empty value.

src/Tasks/Microsoft.Common.CurrentVersion.targets
jeffkl: Will this be in `FileWrites`?  It probably needs to be in this item group almost always otherwise it will get deleted in between builds right?

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: That was my original plan, but it would break the case where `TargetPath` is explicitly specified, which seems to be legal from my reading of the prior set of conditions.

I'd be fairly ok with saying "if you want to do that odd thing, you must also specify this" but it needs to be an explicit choice.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: This is the only field that's optional, right? If this particular project doesn't have a ref asm for whatever reason, we don't want to populate this field because it'll break downstream projects.

I think this'll be clearer when I just use a big off switch for the entire feature, rather than checking whether that path is populated. Seem reasonable?

src/Tasks/Microsoft.Common.CurrentVersion.targets
jeffkl: I think if a user wants to set $(TargetPath) they should also have to set $(TargetRefPath) right?  There's no guarantee that appending `ref\$(TargetFileName)` to whatever they specified as `TargetPath` would make sense.  I'll let you decide...

src/Tasks/Microsoft.Common.CurrentVersion.targets
jeffkl: From what I've seen, setting metadata is _never_ conditioned because the getting of the value is checked later.  But if you like this pattern more I'll let you do what you think is best.

jcouv:
FYI @jasonmalinowski (this is the core change for MSBuild to handle ref assemblies).

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -154,6 +154,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <_DocumentationFileProduced>true</_DocumentationFileProduced>
     <_DocumentationFileProduced Condition="'$(DocumentationFile)'==''">false</_DocumentationFileProduced>
 
+    <!-- Whether or not a reference assembly is produced-->
+    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
```
cdmihai: Are users supposed to use this property to turn ref assemblies off?

rainersigwald: To not _create_ them in this project. They'll still be consumed if produced in other projects.

Better names welcome.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2015,8 +2028,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="FilesWritten" ItemName="FileWrites"/>
       <Output TaskParameter="DependsOnSystemRuntime" PropertyName="DependsOnSystemRuntime"/>
     </ResolveAssemblyReference>
+
+    <ItemGroup>
+      <ReferencePathWithInterfaceOnlyAssemblies Include="@(ReferencePath->'%(ReferenceAssembly)')" />
```
cdmihai: Is it OK to have empty strings for the ReferencaPath items that do not have ref assemblies?

cdmihai: Ah, the item def group is the next diff, ignore :)

---
In reply to: [114407016](https://github.com/Microsoft/msbuild/pull/2039#discussion_r114407016) [](ancestors = 114407016)

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: if not already defined!

AndyGerlicher:
@dotnet-bot test this please

rainersigwald:
@dotnet-bot test this please

(New CI jobs might be borked)

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Confirmed that after my last changes `Clean` now deletes the ref assembly from both bin and obj, and `IncrementalClean` leaves it alone. Is there another scenario to test?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1797,6 +1809,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <TargetPathWithTargetPlatformMoniker Include="$(TargetPath)">
         <TargetPlatformMoniker>$(TargetPlatformMoniker)</TargetPlatformMoniker>
         <TargetPlatformIdentifier>$(TargetPlatformIdentifier)</TargetPlatformIdentifier>
+        <ReferenceAssembly Condition="'$(ProduceReferenceAssembly)' != ''">$(TargetRefPath)</ReferenceAssembly>
```
jasonmalinowski: Should this condition be if ProduceReferenceAssembly == true? Otherwise if false we'll still be setting this.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2015,8 +2028,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      <ReferencePathWithRefAssemblies Include="@(ReferencePath->'%(ReferenceAssembly)')" />
+    </ItemGroup>
   </Target>
 
+  <ItemDefinitionGroup>
+    <!-- Reference assemblies are not produced in all cases, but it's easier to consume them
+         if this metadatum is always populated. This item definition ensures that it points
+         to the implementation assembly unless specified. -->
+    <ReferencePath>
+      <ReferenceAssembly>%(FullPath)</ReferenceAssembly>
```
jasonmalinowski: That is magic. ✨

src/Tasks/Microsoft.Common.tasks
```diff
@@ -172,5 +172,6 @@
     <!-- Roslyn tasks are now in an assembly owned and shipped by Roslyn -->
     <UsingTask TaskName="Microsoft.CodeAnalysis.BuildTasks.Csc"                       AssemblyFile="$(RoslynTargetsPath)\Microsoft.Build.Tasks.CodeAnalysis.dll" Condition="'$(MSBuildAssemblyVersion)' != ''" />
     <UsingTask TaskName="Microsoft.CodeAnalysis.BuildTasks.Vbc"                       AssemblyFile="$(RoslynTargetsPath)\Microsoft.Build.Tasks.CodeAnalysis.dll" Condition="'$(MSBuildAssemblyVersion)' != ''" />
+    <UsingTask TaskName="Microsoft.CodeAnalysis.BuildTasks.CopyRefAssembly"           AssemblyFile="$(RoslynTargetsPath)\Microsoft.Build.Tasks.CodeAnalysis.dll" Condition="'$(MSBuildAssemblyVersion)' != ''" />
```
jasonmalinowski: What's the expected behavior if the user has a newer version of the Roslyn package installed? I believe that package does UsingTasks on Csc/Vbc so those get overridden. Should we be doing the UsingTask too there so that way an upgrade also means you get a newer version here? (cc @jcouv)

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Yes 😳

src/Tasks/Microsoft.Common.tasks
rainersigwald: Excellent catch! I added it to the Roslyn package props in https://github.com/dotnet/roslyn/pull/19146/commits/ae50fe8cd9861e4f65709f123f2470d9ebb43728

jcouv: Thanks 😄

jcouv:
Relates to https://github.com/dotnet/roslyn/issues/18612

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Alas, it was *too* magic: #2165

Issue #2044, opened 2017-05-02
Potential perf optimisations in CompositeGlob


cdmihai:
[CompositeGlob](https://github.com/Microsoft/msbuild/blob/master/src/Build/Globbing/CompositeGlob.cs) could have the following potential improvements:

- Collapse recursive tree to leaves to reduce object count. For example, `CompositeGlob(GlobA, CompositeGlob(GlobB, GlobC))` could be internally rewritten to `CompositeGlob(GlobA, GlobB, GlobC)`. This is doable as long as all Composites in one subtree are the same.
- Merge regular expressions: `CompositeGlob(GlobA, GlobB, GlobC)` could be rewritten to `CompositeGlob(MergedGlob)`, where `MergedGlob`'s regex is the merger of the `GlobA`, `GlobB`, `GlobC ` regexes: `(GlobA.Regex)|(GlobBRegex)|(GlobC.Regex)`. Preliminary tests indicate this is faster by an order for less than ~50 regexes, but then gets exponentially slower.

PR #2045, opened 2017-05-02
Do not use PLINQ in CompositeGlob

cdmihai:
Threadpools are a scarce resource in VS, do not use them.

We should instead consider other optimisations: #2044


cdmihai:
@lifengl Look good?

lifengl:
looks good to me.  In the future, maybe composition glob can merge regular expressions...


________________________________

PR #2058, opened 2017-05-04
Initial stab at F# attribute generation for CoreCLR

baronfel:
This should help address dotnet/netcorecli-fsc#93.


AndyGerlicher:
Agreed with what @rainersigwald said in the referenced issue, I don't think this belongs in MSBuild. Can this be part of the F# SDK?

AndyGerlicher:
Closing this for now. Appreciate the contribution here, but I don't think it should be in MSBuild (as opposed to the F# SDK as suggested). Feel free to open with more info if you think we're mistaken or it ends up not being possible to sort in that way.

KevinRansom:
@AndyGerlicher can you please explain why it is okay to have VB and C# code generation here but not F#?

PR #2084, opened 2017-05-10
Fix custom localization and packing of satellite assemblies

AArnott:
## Add missing item type qualifier in Condition

This fixes SatelliteDllsProjectOutputGroup when SatelliteDllsProjectOutputGroupOutputIntermediate items already exist so that it doesn't double up and malform SatelliteDllsProjectOutputGroupOutputIntermediate items.

## Preserve FinalOutputPath metadata if preset 

When localization for an app occurs another way (like Microsoft's own LCL tooling) the FinalOutputPath isn't what is calculated here. We can preset it in the Intermediate items, but we need MSBuild to preserve that metadata when it is there.


AArnott:
I discovered these bugs when trying to validate the fix for https://github.com/NuGet/Home/issues/5160.
CC: @rohit21agrawal

AArnott:
@AndyGerlicher Am I targeting the right branch for a 15.3 release?

PR #2085, opened 2017-05-10
Spellcheck

nschonni:
Was trying to turn off my brain, so I ran https://github.com/EWSoftware/VSSpellChecker on the solution.
There are some whitespace trimming diffs, so looking at it with the `?w=1` might be a good idea.
This is mostly just comment fixes in the unit tests, but the XSD changes might have an affect on the translations.

### DIFF
--- a/src/Build.UnitTests/BackEnd/LoggingServicesLogMethod_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/LoggingServicesLogMethod_Tests.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Build.UnitTests.Logging
            );
         }
         /// <summary>
-        /// Verify an InternlErrorException is thrown when an empty MessageResourceName is passed in. 
+        /// Verify an InternlErrorException is thrown when an empty MessageResourceName is passed in.
         /// </summary>
         [Fact]
         public void LogErrorEmptyMessageResource()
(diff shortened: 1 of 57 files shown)


AndyGerlicher:
Thanks for the contribution!

src/Tasks.UnitTests/Touch_Tests.cs
```diff
@@ -23,7 +23,7 @@ sealed public class Touch_Tests
 
         internal static string myexisting_txt = NativeMethodsShared.IsWindows ? @"c:\touch\myexisting.txt" : @"/touch/myexisting.txt";
         internal static string mynonexisting_txt = NativeMethodsShared.IsWindows ? @"c:\touch\mynonexisting.txt" : @"/touch/mynonexisting.txt";
-        internal static string nonexisting_txt = NativeMethodsShared.IsWindows ? @"c:\touch-nonexisting\file.txt" : @"/touch-nonexisting/file.txt";
+        internal static string nonexisting_txt = NativeMethodsShared.IsWindows ? @"c:\touch-non existing\file.txt" : @"/touch-nonexistent/file.txt";
```
nschonni: Yeah, the first should have been changed to `nonexistent` as well

src/Build.UnitTests/BackEnd/LoggingServicesLogMethod_Tests.cs
```diff
@@ -1772,8 +1772,7 @@ public static IBuildComponent CreateLoggingService(LoggerMode mode, int nodeId,
             }
 
             /// <summary>
-            /// Override the method to log which event was p1446
-            /// rocessed so it can be verified in a test
+            /// Override the method to log which event was processed so it can be verified in a test
```
nschonni: Yeah, I wasn't sure if `p1446` was a code, but eventually say that it was an odd split of the word `processed`

PR #2091, opened 2017-05-11
Spelling 2: Electric Boogaloo

nschonni:
Think this is the last of it. Just took a quick second pass to see what I had accidentally hit "Ignore All" in #2085

### DIFF
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Build.UnitTests.OM.Definition
             ";
 
         /// <summary>
-        /// Since when the project file is saved it may be intented we want to make sure the indent characters do not affect the evaluation against empty.
+        /// Since when the project file is saved it may be intended we want to make sure the indent characters do not affect the evaluation against empty.
         /// We test here newline, tab, and carriage return.
         /// </summary>
         [Fact]
--- a/src/Shared/StringBuilderCache.cs
+++ b/src/Shared/StringBuilderCache.cs
@@ -5,28 +5,28 @@
 **
 **
 ** Purpose: provide a cached reusable instance of StringBuilder
-**          per thread  it's an optimisation that reduces the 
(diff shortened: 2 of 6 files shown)


src/Shared/StringBuilderCache.cs
```diff
@@ -5,28 +5,28 @@
 **
 **
 ** Purpose: provide a cached reusable instance of StringBuilder
-**          per thread  it's an optimisation that reduces the 
+**          per thread  it's an optimization that reduces the
```
rainersigwald: Who on the MSBuild team was ever British? Huh. :us:

src/Tasks.UnitTests/CreateProperty_Tests.cs
```diff
@@ -47,7 +47,7 @@ public void CreateBlankProperty()
 
         /// <summary>
         /// Make sure that I can use the CreateProperty task to create a property
-        /// that has a parseable semicolon in it.
+        /// that has a parsable semicolon in it.
```
rainersigwald: Was skeptical of this change but it does look like a sorta-standard: https://en.wiktionary.org/wiki/parsable.

nschonni: Yeah, it was spelled both ways in the repo so I did the same think and did a search to find which was "correct"

rainersigwald:
Thanks again!

src/Build.OM.UnitTests/Definition/Project_Tests.cs
```diff
@@ -81,7 +81,7 @@ public void Dispose()
             ";
 
         /// <summary>
-        /// Since when the project file is saved it may be intented we want to make sure the indent characters do not affect the evaluation against empty.
+        /// Since when the project file is saved it may be intended we want to make sure the indent characters do not affect the evaluation against empty.
```
eerhardt: `indented`?

nschonni: Yup, that would make more sense.

PR #2092, opened 2017-05-11
Use 32-bit registry for .NET SDK locations

### LINKED ISSUE #2047: When using 64bit version of MSBuild 15.0, $(SDK40ToolsPath) is not set
There's an unexpected difference between using MSBuild 15.0 from the base installation folder vs. calling it from the amd64/ subfolder: $(SDK40ToolsPath) is set only for the x86 version, the amd64 version does not set the variable. As a result, calling tools like `gacutil.exe` from those paths fails.

# Steps to reproduce

**1. Create a MSBuild project file 'test.proj':**
```
    <PropertyGroup>
        <GacUtilPath Condition="'$(GacUtilPath)' == ''">$(SDK40ToolsPath)gacutil.exe</GacUtilPath>
    </PropertyGroup>
    <Target Name="Build">
        <Exec Command="&quot;$(GacUtilPath)&quot; /?" />
    </Target>
</Project>
```

Specifying ToolsVersion="15.0" explicitly does not change the outcome.

**2. Good case: Start with 32bit MSBuild:**
```
& 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\MSBuild.exe' test.proj
```

**3. Bad Kitty Case: Start with 64bit MSBuild:**
```
& 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64\MSBuild.exe' test.proj
```

**Maybe related**: Issue #1920 "MSBuild v15.0 & Registry". The folder registration has changed for MSBuild 15.0, but I'm somewhat confused by the entries of the application configuration. (They are the same for both MSBuild versions, so why's there a difference?)

**Remarks**: I (seemingly incorrectly) reported this at [Connect](https://connect.microsoft.com/VisualStudio/feedback/details/3133352). Also since this is command-line only, it, unlike the comment there, is unlikely to be related to Visual Studio.

rainersigwald:
Fixes #2047 by always using the 32-bit registry to locate `$(SDK35ToolsPath)` and `$(SDK40ToolsPath)`.

### DIFF
--- a/src/MSBuild/app.amd64.config
+++ b/src/MSBuild/app.amd64.config
@@ -81,8 +81,8 @@
         <property name="MSBuildFrameworkToolsPath32" value="$(SystemRoot)\Microsoft.NET\Framework\v$(MSBuildRuntimeVersion)\" />
         <property name="MSBuildFrameworkToolsPath64" value="$(SystemRoot)\Microsoft.NET\Framework64\v$(MSBuildRuntimeVersion)\" />
         <property name="MSBuildFrameworkToolsRoot" value="$(SystemRoot)\Microsoft.NET\Framework\" />
-        <property name="SDK35ToolsPath" value="$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SDKs\Windows\v8.0A\WinSDK-NetFx35Tools-x86@InstallationFolder)" />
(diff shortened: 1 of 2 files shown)


rainersigwald:
@AndyGerlicher as we discussed in triage the other day, I considered adding a `$(Registry32:)` magic property, but it looked fairly complex (changes to the evaluator that could slow all property evaluations) for the benefit, so I'm switching to the existing `GetRegistryValueFromView` property function.

PR #2098, opened 2017-05-12
Do not leak dummy variable in user owned ProjectInstance object

cdmihai:
In addition, put a protective check in the two known msbuild string caches to preemptively avoid caching unique strings.

False positive caching is still hit in the following scenarios
- building successive in-memory mutations of the same `ProjectInstance` instance, in the same BeginBuild-EndBuild-ResetCaches session
- building different `ProjectInstance` instances over the same file, in the same BeginBuild-EndBuild-ResetCaches session

To fix these I think we have to actually remove the dummy variable and change `ConfigurationMetadata` (implements what it means to be a unique build in the engine) to account for the entire `ProjectInstance` state.

### DIFF
--- a/src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
+++ b/src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
@@ -485,6 +485,12 @@ namespace Microsoft.Build.BackEnd
             ErrorUtilities.VerifyThrow(IsLoaded, $"This {nameof(BuildRequestConfiguration)} must be loaded at the end of this method");
         }
 
+        internal void StampWithUniqueGlobalProperty()
+        {
+            var key = $"ProjectInstance{Guid.NewGuid():N}";
+            Properties[key] = ProjectPropertyInstance.Create(key, "Forces unique project identity in the MSBuild engine");
+        }
+
         /// <summary>
         /// Returns true if the default and initial targets have been resolved.
         /// </summary>
@@ -1056,6 +1062,5 @@ namespace Microsoft.Build.BackEnd
                 throw;
             }
         }
-
     }
 }
--- a/src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
+++ b/src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
@@ -485,10 +485,13 @@ namespace Microsoft.Build.BackEnd
             ErrorUtilities.VerifyThrow(IsLoaded, $"This {nameof(BuildRequestConfiguration)} must be loaded at the end of this method");
         }
 
-        internal void StampWithUniqueGlobalProperty()
+        internal void CreateUniqueGlobalProperty()
         {
+            // create a copy so the mutation does not leak into the ProjectInstance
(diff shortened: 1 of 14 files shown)


src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -490,6 +490,9 @@ internal void CreateUniqueGlobalProperty()
             // create a copy so the mutation does not leak into the ProjectInstance
             _globalProperties = new PropertyDictionary<ProjectPropertyInstance>(_globalProperties);
 
+            // if you change the key also change the following so that it does not leak into string caches:
+            // Microsoft.Build.OpportunisticIntern.BucketedPrioritizedStringList.TryIntern
+            // Microsoft.Build.Construction.ProjectStringCache.Add
             var key = $"ProjectInstance{Guid.NewGuid():N}";
```
rainersigwald: Can you define a constant and use it in all 3 places?

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -485,6 +485,18 @@ public void InitializeProject(BuildParameters buildParameters, Func<ProjectInsta
...
 
+        internal void CreateUniqueGlobalProperty()
+        {
+            // create a copy so the mutation does not leak into the ProjectInstance
+            _globalProperties = new PropertyDictionary<ProjectPropertyInstance>(_globalProperties);
+
+            // if you change the key also change the following so that it does not leak into string caches:
+            // Microsoft.Build.OpportunisticIntern.BucketedPrioritizedStringList.TryIntern
+            // Microsoft.Build.Construction.ProjectStringCache.Add
+            var key = $"ProjectInstance{Guid.NewGuid():N}";
```
rainersigwald: Should this name be prefixed with `MSBuild` like other reserved/magic properties?

src/Shared/OpportunisticIntern.cs
```diff
@@ -886,6 +886,28 @@ internal void ReportStatistics(string heading)
...
+                             candidate[5] == 'c' &&
+                             candidate[6] == 't' &&
+                             candidate[7] == 'I' &&
+                             candidate[8] == 'n' &&
+                             candidate[9] == 's' &&
+                             candidate[10] == 't' &&
+                             candidate[11] == 'a' &&
+                             candidate[12] == 'n' &&
+                             candidate[13] == 'c' &&
+                             candidate[14] == 'e'
```
rainersigwald: Oh, this probably makes the use-a-constant part hard. 😢

cdmihai: True, but I can iterate over the constant. The compiler hopefully will unroll it a sufficient number of times to actually remove the loop entirely.

rainersigwald: Probably best to check the IL output of the compiler. I suspect Roslyn doesn't do that kind of unrolling, and the [JIT doesn't](https://github.com/dotnet/coreclr/issues/11606).

AndyGerlicher:

PR #2099, opened 2017-05-13
Add a try/catch in BinaryLogger.

KirillOsenkov:
Just a safeguard to prevent the logger from crashing the build. I'm not even sure if it's necessary because of downstream catch blocks, but it feels right to contain exceptions here.

### DIFF
--- a/src/Build/Logging/BinaryLogger/BinaryLogger.cs
+++ b/src/Build/Logging/BinaryLogger/BinaryLogger.cs
@@ -83,7 +83,15 @@ namespace Microsoft.Build.Logging
 
         private void EventSource_AnyEventRaised(object sender, BuildEventArgs e)
         {
-            Write(e);
+            try
+            {
+                Write(e);
+            }
+            catch
+            {
+                // Exceptions here are unlikely but it'd be bad for a logger to crash the build.
+                // We can't log either because the logger is likely broken. Best just do nothing.
+            }
         }
 
         private void Write(BuildEventArgs e)


src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -83,7 +83,15 @@ public void Shutdown()
...
         private void EventSource_AnyEventRaised(object sender, BuildEventArgs e)
         {
-            Write(e);
+            try
+            {
+                Write(e);
+            }
+            catch
+            {
+                // Exceptions here are unlikely but it'd be bad for a logger to crash the build.
```
AndyGerlicher: Console.Write? Might be better than nothing so it's not completely lost?

rainersigwald:
I could have sworn I typed this reply already, but I guess I closed the tab or something.

I'm not sure this is better than just throwing. We already have attempts to catch any exception from a logger, package it semi-nicely for the user, and semi-gracefully fail the build. That seems better than a silent failure here.

cdmihai:
@KirillOsenkov did you see any weird exceptions bubbling up through there? :)

AndyGerlicher:
Team Triage:
Closing this. As @rainersigwald said we should have safeguards against failures here already and would rather not eat a generic exception with no handling.

KirillOsenkov:
So we do display a good error message, but we also fail the build:

```
C:\MSBuild\bin\Debug\Output\MSBuild.exe /bl C:\CodeCleanupTools\CodeCleanupTools.sln
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
Build started 5/16/2017 3:41:37 PM.
Project "C:\CodeCleanupTools\CodeCleanupTools.sln" on node 1 (default targets).
ValidateSolutionConfiguration:
  Building solution configuration "Debug|Mixed Platforms".
MSBUILD : error MSB4017: The build stopped unexpectedly because of an unexpected logger failure.
This is an unhandled exception from a logger -- PLEASE OPEN A BUG AGAINST THE LOGGER OWNER.
Microsoft.Build.Exceptions.InternalLoggerException: The build stopped unexpectedly because of an unexpected logger failure. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Build.Logging.BuildEventArgsWriter.Write(TargetFinishedEventArgs e) in C:\MSBuild\src\Build\Logging\BinaryLogger\BuildEventArgsWriter.cs:line 156
   at Microsoft.Build.Logging.BuildEventArgsWriter.Write(BuildEventArgs e) in C:\MSBuild\src\Build\Logging\BinaryLogger\BuildEventArgsWriter.cs:line 50
   at Microsoft.Build.Logging.BinaryLogger.Write(BuildEventArgs e) in C:\MSBuild\src\Build\Logging\BinaryLogger\BinaryLogger.cs:line 96
   at Microsoft.Build.Logging.BinaryLogger.EventSource_AnyEventRaised(Object sender, BuildEventArgs e) in C:\MSBuild\src\Build\Logging\BinaryLogger\BinaryLogger.cs:line 86
   at Microsoft.Build.Evaluation.ProjectCollection.ReusableLogger.AnyEventRaisedHandler(Object sender, BuildEventArgs e) in C:\MSBuild\src\Build\Definition\ProjectCollection.cs:line 2264
   at Microsoft.Build.Framework.AnyEventHandler.Invoke(Object sender, BuildEventArgs e)
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.RaiseAnyEvent(Object sender, BuildEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 925
   --- End of inner exception stack trace ---
   at Microsoft.Build.Exceptions.InternalLoggerException.Throw(Exception innerException, BuildEventArgs e, String messageResourceName, Boolean initializationException, String[] messageArgs) in C:\MSBuild\src\Build\Errors\InternalLoggerException.cs:line 236
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.RaiseAnyEvent(Object sender, BuildEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 957
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.RaiseStatusEvent(Object sender, BuildStatusEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 908
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.RaiseTargetFinishedEvent(Object sender, TargetFinishedEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 740
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.Consume(BuildEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 222
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.Consume(BuildEventArgs buildEvent, Int32 sinkId) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 192
   at Microsoft.Build.BackEnd.Logging.EventRedirectorToSink.Microsoft.Build.Framework.IEventRedirector.ForwardEvent(BuildEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventRedirectorToSink.cs:line 61
   at Microsoft.Build.BackEnd.Logging.CentralForwardingLogger.EventSource_AnyEventRaised(Object sender, BuildEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\CentralForwardingLogger.cs:line 123
   at Microsoft.Build.BackEnd.Logging.EventSourceSink.RaiseAnyEvent(Object sender, BuildEventArgs buildEvent) in C:\MSBuild\src\Build\BackEnd\Components\Logging\EventSourceSink.cs:line 925
```

Are we OK with this? Maybe I should add more logging in the catch block? Or am I being too paranoid?

KirillOsenkov:
Throwing a LoggerException results fails the build as well but doesn't display the stack:
```
C:\MSBuild\bin\debug\Windows_NT_Deployment\bin\debug\Windows_NT_Deployment\msbuild.exe /bl C:\CodeCleanupTools\CodeCleanupTools.sln
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
Build started 5/16/2017 3:45:55 PM.
Project "C:\CodeCleanupTools\CodeCleanupTools.sln" on node 1 (default targets).
ValidateSolutionConfiguration:
  Building solution configuration "Debug|Mixed Platforms".
MSBUILD : Logger error MSB1029: Awesome logger has crashed
```

cdmihai:
My primary concern with swallowing the exception is that bugs go unnoticed. On the other hand, having the logger fail the build is a bit annoying. However, if the binary logger fails silently, it would silently stop recording the events it failed for, which to me sounds more dangerous / messy than just failing the build (FrankenLog :) ). I think I prefer letting the user decide whether to file a bug and stop using the binary logger until the bug gets fixed, rather than having the user get incomplete logs without him knowing that's happening.

KirillOsenkov:
Good point. If the binary logger crashes, it'll print the stack and next time they'll just restart the build without the binary logger. I think I'm happy with that.

Issue #2083, opened 2017-05-09
SdkResolverContext needs to provide equivalent of MSBuildProjectFullPath


nguerrera:
We need an equivalent of $(MSBuildProjectFullPath) in the resolver context.  SdkResolverContext.ProjectFilePath is actually the path of the importing file, not the path to the project being built.  It turns out that if SolutionFilePath is unavailable and there are SDKs that import SDKs, then the CLI resolver has no way to pick all SDKs from a compatible version. See dotnet/cli#6553.

I think SdkResolverContext.ProjectFilePath should be MSBuildProjectFullPath and a separate property SdkResolverContext.ImportingProjectFilePath should be used for the current meaning. 

@AndyGerlicher @dsplaisted @johnbeisner

Issue #2082, opened 2017-05-09
SdkResolverContext.SolutionFilePath is sometimes *Undefined* instead of null


nguerrera:
See https://github.com/Microsoft/msbuild/blob/eeecf7bd7fb7af9be893b019828a3d9e28f9158d/src/Tasks/Microsoft.Common.CurrentVersion.targets#L311

It's not very predictable when this evaluation will impact the value of the context's SolutionFilePath. If you have a simple `dotnet new console` app, it won't happen at all but if you follow the repro steps in dotnet/cli#6553, you'll get a mix of `null` and `"*Undefined*"` in different resolver calls.

nguerrera:
@AndyGerlicher @dsplaisted @jbeisner

nguerrera:
We can workaround this in our resolver by treating "*Undefined*" as null, but I think it should be fixed so that resolvers either see the real solution path or null.

PR #2102, opened 2017-05-15
Add Version and fix ProjectPath in SdkResolverContext

### LINKED ISSUE #2097: SdkResolverContext should provide msbuild version
This is needed for dotnet/cli#6585

@AndyGerlicher

AndyGerlicher:
* Adds MSBuild version.
 * Fix ProjectFilePath to be the full path to the project (equivalent to
 '$(MSBuildProjectFullPath)').
 * Fix issue where SolutionFilePath can be set by Common targets to
 '*Undefined*'. When this is the case, set to null for consistency.

Closes #2097, #2083, #2082 

Looking into adding some tests, will have a followup commit.

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using Microsoft.Build.BackEnd.Logging;
 using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 
@@ -40,9 +41,10 @@ namespace Microsoft.Build.BackEnd
         /// <param name="buildEventContext">Build event context for logging.</param>
         /// <param name="sdkReferenceLocation">Location of the element within the project which referenced the SDK.</param>
         /// <param name="solutionPath">Path to the solution if known.</param>
+        /// <param name="projectPath">Path to the project being.</param>
         /// <returns>Path to the root of the referenced SDK.</returns>
         internal string GetSdkPath(SdkReference sdk, ILoggingService logger, BuildEventContext buildEventContext,
-            ElementLocation sdkReferenceLocation, string solutionPath)
+            ElementLocation sdkReferenceLocation, string solutionPath, string projectPath)
         {
             ErrorUtilities.VerifyThrowInternalNull(sdk, nameof(sdk));
             ErrorUtilities.VerifyThrowInternalNull(logger, nameof(logger));
@@ -58,7 +60,7 @@ namespace Microsoft.Build.BackEnd
                 var buildEngineLogger = new SdkLoggerImpl(logger, buildEventContext);
                 foreach (var sdkResolver in _resolvers)
                 {
-                    var context = new SdkResolverContextImpl(buildEngineLogger, sdkReferenceLocation.File, solutionPath);
+                    var context = new SdkResolverContextImpl(buildEngineLogger, projectPath, solutionPath, ProjectCollection.Version);
                     var resultFactory = new SdkResultFactoryImpl(sdk);
                     try
                     {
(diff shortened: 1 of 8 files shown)


src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
```diff
@@ -40,9 +41,10 @@ internal SdkResolution(SdkResolverLoader sdkResolverLoader)
         /// <param name="buildEventContext">Build event context for logging.</param>
         /// <param name="sdkReferenceLocation">Location of the element within the project which referenced the SDK.</param>
         /// <param name="solutionPath">Path to the solution if known.</param>
+        /// <param name="projectPath">Path to the project being.</param>
```
rainersigwald: you accidentally a word.

src/Framework/Sdk/SdkResolverContext.cs
```diff
@@ -22,5 +24,10 @@ public abstract class SdkResolverContext
         ///     Path to the solution file being built, if known. May be null.
         /// </summary>
         public virtual string SolutionFilePath { get; protected set; }
+
+        /// <summary>
+        ///     Version of MSBuild currently running.
```
rainersigwald: Clarify full git-height version vs. assembly version.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2370,12 +2370,15 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             if (importElement.ParsedSdkReference != null)
             {
-                // Try to get the solution path when available.
+                // Try to get the path to the solution and project being built. If the solution path equals "*Undefined*" 
+                // (hard coded in Microsoft.Common.CurrentVersion.targets) set to null for consistency.
                 var solutionPath = _data.GetProperty(SolutionProjectGenerator.SolutionPathPropertyName)?.EvaluatedValue;
+                var projectPath = _data.GetProperty(ReservedPropertyNames.projectFullPath)?.EvaluatedValue;
```
rainersigwald: Isn't there a `GetPropertyValue()`?

rainersigwald: Isn't there a `GetPropertyValue()`?

rainersigwald: Is getting this out of the reserved property the only way to get the information at this layer? I don't like that (even though AFAIK it's immutable it seems unclean).

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2370,12 +2370,15 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             if (importElement.ParsedSdkReference != null)
             {
-                // Try to get the solution path when available.
+                // Try to get the path to the solution and project being built. If the solution path equals "*Undefined*" 
+                // (hard coded in Microsoft.Common.CurrentVersion.targets) set to null for consistency.
```
rainersigwald: Instead of 

> If the solution path equals "*Undefined*" (hard coded in Microsoft.Common.CurrentVersion.targets) set to null for consistency.

Maybe

> The solution path is not directly known in MSBuild. It is passed in as a property either by the VS project system or by MSBuild's solution metaproject. Microsoft.Common.CurrentVersion.targets sets the value to `*Undefined*` when not set, and for backward compatibility, we shouldn't change that. But resolvers should be exposed to a string that's null or a full path, so correct that here.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2370,12 +2370,15 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
                 var solutionPath = _data.GetProperty(SolutionProjectGenerator.SolutionPathPropertyName)?.EvaluatedValue;
+                var projectPath = _data.GetProperty(ReservedPropertyNames.projectFullPath)?.EvaluatedValue;
+                if (solutionPath == "*Undefined*") solutionPath = null;
```
rainersigwald: Nit: put this line next to the acquisition of `solutionPath` (above `projectPath)`.

src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
```diff
@@ -100,6 +102,16 @@ internal SdkResolution(SdkResolverLoader sdkResolverLoader)
             return null;
         }
 
+        internal void InitializeForTests(IList<SdkResolver> resolvers, ILoggingService logger, BuildEventContext buildEventContext, ElementLocation location)
+        {
+            _resolvers = resolvers;
+        }
+
+        internal void ResetInitializeForTests()
```
rainersigwald: Where is this used/how is it (and `InitializeForTests`) related to this change?

rainersigwald:
@dotnet-bot test ci please

rainersigwald:
@dotnet-bot test this please

PR #2104, opened 2017-05-17
Enable PdbGit - stamp .pdbs with links to sources on GitHub.

KirillOsenkov:
https://github.com/AArnott/PdbGit rewrites the .pdb files post-compilation to include full URLs to GitHub for each source file. That way when the VS debugger is using the .pdb to find the source code, it will download the exact .cs source file that was used during compilation from the blob on GitHub.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -422,4 +422,6 @@
     <IBCMergeSubPath>x86/MSBuild</IBCMergeSubPath>
   </PropertyGroup>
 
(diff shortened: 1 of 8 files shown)


KirillOsenkov:
I've verified that this works fine and still produces classic Windows .pdbs. Let's take this PR instead of https://github.com/Microsoft/msbuild/pull/2037

ctaggart:
We want to enable source link support, not source server support. Please close this and help figure out the Microsoft internal symbol server situation for #2037, so that source linking in general will be successful.

rainersigwald:
Let's do #2107.

PR #2109, opened 2017-05-17
Embed the project and target source files from a build.

KirillOsenkov:
As the build is happening the BuildEventArgs will contain paths to the source files on disk encountered during the build: .proj files, .targets files, imported projects as well as potentially some source files.

It is valuable to capture these exact files and embed them into the binary log such that the log analysis has the actual sources using during the build at its disposal.

There are three options: don't capture sources, capture to a standalone .buildsources.zip archive and embed the zip archive directly into the binary log. The last option (embed) is the default. Logger parameters to change the default will be provided in a separate commit.

This commit adds the facility to write and read blobs to the log file, a SourceFileCollector that adds files to a .zip archive in the background and the integration pieces.


KirillOsenkov:
I've responded to the code review feedback, please take another look?

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        /// <summary>
+        /// Avoid visiting each file more than once.
+        /// </summary>
+        private readonly HashSet<string> _processedFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Extensions to exclude from the .buildsources.zip archive
+        /// </summary>
+        private static readonly HashSet<string> _extensionsToExclude = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
jeffkl: This isn't needed now right?

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
```diff
@@ -0,0 +1,159 @@
...
+            ".h",
+            ".cpp",
+            ".fs",
+            ".vb"
+        };
+
+        // this will form a chain of file write tasks, running sequentially on a background thread
+        private Task _currentTask = Task.CompletedTask;
+
+        public ProjectImportsCollector(string logFilePath, string sourcesArchiveExtension = ".buildsources.zip")
```
jeffkl:
```diff
+string sourcesArchiveExtension = ".projectimports.zip"
-string sourcesArchiveExtension = ".buildsources.zip"
```

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            lock (_fileStream)
+            {
+                // enqueue the task to add a file and return quickly
+                // to avoid holding up the current thread
+                _currentTask = _currentTask.ContinueWith(t =>
+                {
+                    try
+                    {
+                        AddFileCore(filePath);
```
jeffkl: If something goes wrong, will it matter that there's no indication?

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
```diff
@@ -0,0 +1,159 @@
...
+            if (filePath == null || _fileStream == null)
+            {
+                return;
+            }
+
+            lock (_fileStream)
+            {
+                // enqueue the task to add a file and return quickly
+                // to avoid holding up the current thread
+                _currentTask = _currentTask.ContinueWith(t =>
```
jeffkl: This looks cool, I've never seen this.  I'm trusting it works!

src/Build/Logging/BinaryLogger/ProjectImportsCollectorExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        private static string importingProjectPrefix = importingProject.Substring(0, importingProjectFilePathStart);
+
+        public static void IncludeSourceFiles(this ProjectImportsCollector projectImportsCollector, BuildEventArgs e)
+        {
+            if (e is ProjectStartedEventArgs)
+            {
+                var projectStarted = (ProjectStartedEventArgs)e;
+                projectImportsCollector.AddFile(projectStarted.ProjectFile);
+            }
+            else if (e is BuildMessageEventArgs)
```
jeffkl: Since you're using a message instead of a new `EventArgs`, will this parsing logic slow down the build?  You're processing every message right?  I thought you were going to add a new `EventArgs` type...

src/MSBuild/Resources/Strings.resx
```diff
@@ -631,9 +631,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
                      diagnostic-level log, but it contains more information.
                      (Short form: /bl)
 
+                     The binary logger by default collects the source text of
+                     project files, including all imported projects and target
+                     files encountered during the build. The optional 
+                     ProjectImports switch controls this behavior:
+
+                      ProjectImports=None     - Don't collect the sources
```
jeffkl:
```diff
-ProjectImports=None     - Don't collect the sources
+ProjectImports=None     - Don't collect the project imports
```

src/MSBuild/Resources/Strings.resx
```diff
@@ -631,9 +631,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+                      ProjectImports=Embed    - Embed sources in the log file
```
jeffkl:
```diff
-ProjectImports=Embed    - Embed sources in the log file
+ProjectImports=Embed    - Embed project imports in the log file
```

src/MSBuild/Resources/Strings.resx
```diff
@@ -631,9 +631,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+                      ProjectImports=ZipFile  - Save to output.buildsources.zip
```
jeffkl:
```diff
-ProjectImports=ZipFile  - Save to output.buildsources.zip
+ProjectImports=ZipFile  - Save to output.projectimports.zip
```

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2076,6 +2102,17 @@ private void EvaluateImportElement(string directoryOfImportingFile, ProjectImpor
                 PerformDepthFirstPass(importedProjectRootElement);
             }
 
+            if (!importedAnything && hasCondition)
```
cdmihai: Should there be an else branch here? If there is no condition and nothing was imported, maybe it was a glob import and the glob expanded to nothing. Or the import expression referenced properties which resolved to empty.

@jeffkl Sounds about right? Are there other cases where an import with no condition evaluates to empty?

cdmihai: Here are some variants. You might just want to drop the  `&& hasCondition` part
```xml
  <Import Project="Foo/*"/>
  <Import Project="$(EmptyProperty)"/>
  <Import Project="Foo/*" Condition="1 == 0"/>
  <Import Project="$(EmptyProperty)" Condition="1 == 1"/>
```

---
In reply to: [118298391](https://github.com/Microsoft/msbuild/pull/2109#discussion_r118298391) [](ancestors = 118298391)

KirillOsenkov: In that rare case I'm OK with not logging anything.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -107,17 +165,44 @@ private void ProcessParameters()
...
+                }
+                else if (string.Equals(parameter, "ProjectImports=Embed", StringComparison.OrdinalIgnoreCase))
+                {
+                    CollectProjectImports = ProjectImportsCollectionMode.Embedded;
+                }
+                else if (string.Equals(parameter, "ProjectImports=ZipFile", StringComparison.OrdinalIgnoreCase))
+                {
+                    CollectProjectImports = ProjectImportsCollectionMode.ZipFile;
+                }
+                else if (parameter.EndsWith(".binlog", StringComparison.OrdinalIgnoreCase))
```
cdmihai: What if it does not end with binlog? Why isn't that allowed? :)

src/Build/Logging/BinaryLogger/ProjectImportsCollectorExtensions.cs
KirillOsenkov: Right now the StartsWith call we have there is dwarfed by all the disk access happening during the build, so even if there is any slowdown at all I think it's OK to take the hit (since it only happens with /bl).

All that said, I am banking on @cdmihai adding better BuildEventContext for evaluation so I can detect off of that in the future. Or we could utilize the Sender string, which is now unused and is always "MSBuild". If we set the Sender = "Evaluation" and compare to that, that'd be cheaper.

I'm really hesitant to add a new EventArgs subtype because there are assumptions everywhere about the possible set of EventArgs like here:
https://source.dot.net/#Microsoft.Build/BackEnd/Components/Logging/EventSourceSink.cs,297
Who knows what switch default blocks we'll start hitting in other infrastructure (e.g. CloudBuild??)

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
KirillOsenkov: Good catch. Removed.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -33,12 +33,25 @@ public BuildEventArgsReader(BinaryReader binaryReader)
...
+        internal event Action<BinaryLogRecordKind, byte[]> OnBlobRead;
+
+        /// <summary>
         /// Reads the next log record from the binary reader. If there are no more records, returns null.
         /// </summary>
         public BuildEventArgs Read()
         {
             BinaryLogRecordKind recordKind = (BinaryLogRecordKind)ReadInt32();
 
+            while (IsBlob(recordKind))
```
cdmihai: Seems to me that the BinaryLogger is writing the blobs at the end of the stream but the reader is expecting them at the beginning of the stream. Does this actually work? Am I reading something wrong?

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -90,6 +103,21 @@ public BuildEventArgs Read()
             return result;
         }
 
+        /// <summary>
+        /// For now it's just the ProjectImportArchive.
+        /// </summary>
+        private static bool IsBlob(BinaryLogRecordKind recordKind)
+        {
+            return recordKind == BinaryLogRecordKind.ProjectImportArchive;
```
cdmihai: This will get interesting if more blob types appear :)

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
KirillOsenkov: Yeah I was thinking about that. We could issue an error and send it to the log, but it's on a different thread... The catch block has to be there because if an exception bubbles up from a thread like this nobody will catch it and it will tear down the process. Since it's non-essential functionality (worst case we lose a file) I thought it's OK to be on the safer side and just hope it never gets hit.

I have a system where I regularly exercise MSBuild and monitor all first-chance exceptions. If I notice anything really happening there we can think how to mitigate. I think it's OK for now.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: I want to restrict it because people have already mistakenly written .buildlog there and then the viewer was trying to read the binlog as a .buildlog: 
https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/68

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
KirillOsenkov: Think of it as `if (IsBlob(recordKind))`. If the current record is a blob, do the work. Then read again, and if it's another blob, do the work again. Read through all the blobs until we hit a proper record again. So basically "skip over a potential series of blobs".

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
KirillOsenkov: We'll deal with it then ;) We can convert this to a switch statement or anything else.

src/Build/Evaluation/Evaluator.cs
KirillOsenkov:
```
  <Import Project="$(EmptyProperty)"/>
  <Import Project="$(EmptyProperty)" Condition="1 == 1"/>
```
These two are invalid and will cause a build error.

But the glob scenario is interesting, I will log that.

KirillOsenkov: Fixed.

KirillOsenkov:
Thanks for the great feedback so far! Keep it coming :)

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
```diff
@@ -0,0 +1,140 @@
...
+                _fileStream = null;
+                _zipArchive = null;
+            }
+        }
+
+        public void AddFile(string filePath)
+        {
+            if (filePath == null || _fileStream == null)
+            {
+                return;
```
cdmihai: Should this be an error rather than silent return?

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
cdmihai: Oh, I see. This is called in a loop from the `BinaryLogReplayEventSource`. I had the wrong impression that this method reads everything. Ignore me :)

---
In reply to: [118305840](https://github.com/Microsoft/msbuild/pull/2109#discussion_r118305840) [](ancestors = 118305840)

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
KirillOsenkov: Not sure... I'd leave it as is unless you strongly feel there should be something.

src/Build/Evaluation/Evaluator.cs
cdmihai: Ah, yes, it would fail here: [https://github.com/Microsoft/msbuild/blob/4b50e21f36ccf87abf0ad0adc93f203d44a3d848/src/Build/Evaluation/Evaluator.cs#L2425-L2430](https://github.com/Microsoft/msbuild/blob/4b50e21f36ccf87abf0ad0adc93f203d44a3d848/src/Build/Evaluation/Evaluator.cs#L2425-L2430)

---
In reply to: [118308379](https://github.com/Microsoft/msbuild/pull/2109#discussion_r118308379) [](ancestors = 118308379)

src/Build/Logging/BinaryLogger/ProjectImportsCollector.cs
cdmihai: I think it should be an error. It is not a valid program state. You can use `ErrorUtilities.VerifyThrowInternalNull`

KirillOsenkov: As far as I can see, it is valid and can be hit.

Let's first deal with filePath. AddFile() is called from two places and in one of them we pass ProjectStartedEventArgs.ProjectFile, which can be null: http://source.dot.net/#Microsoft.Build.Framework/ProjectStartedEventArgs.cs,8d59ac5ef1c85c0b Unless we can prove that that is never null (which I don't think we can) we should expect null as a valid value for filePath.

_fileStream can be null if the collector was disabled due to a previous error (such as unable to write to the log file).

In non-essential functionality such as here I'd prefer to fail silently than risk crashing the build. Worst case, we don't include a file, who cares :)

KirillOsenkov:
This PR has now been superceded by https://github.com/Microsoft/msbuild/pull/2154

PR #2110, opened 2017-05-18
Fix regression with log file summary.

### LINKED ISSUE #1968: Msbuild ignores NoSummary parametr for filelog
If you use `/flp:LogFile=Test.log;NoSummary;ErrorsOnly` parameter, log should be empty (size = 0) unless there is no error in solution.

In msbuild 15.x there is always summary information:
```
Build succeeded.

Time Elapsed 00:00:03.34
```

See test - [MsbuildBug.zip](https://github.com/Microsoft/msbuild/files/915901/MsbuildBug.zip)

AndyGerlicher:
A change was (#1513) to allow for more control over log summary text.
Specifically, the ask was to allow for the use of ErrorsOnly /
WarningsOnly with Summary. That change also impacted the file log output
and in some cases caused build summary to output to the log file when
using ErrorsOnly.

This change defaults to not showing the summary text in the FileLogger
when ErrorsOnly or WarningsOnly is set (it can still be displayed with
the Summary flag). This reverts behavior back to pre-VS2017 behavior and
allows for an empty err/wrn file to be produced.

Closes #1968

### DIFF
--- a/src/Build.UnitTests/FileLogger_Tests.cs
+++ b/src/Build.UnitTests/FileLogger_Tests.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Collections.Generic;
 using System.Text;
 
+using Microsoft.Build.Engine.UnitTests;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Logging;
 using Microsoft.Build.Shared;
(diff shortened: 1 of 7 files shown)


src/Build.UnitTests/FileLogger_Tests.cs
```diff
@@ -342,6 +343,36 @@ public void BasicNoExistingDirectory()
             }
         }
 
+        [Theory]
+        [InlineData("warningsonly")]
```
cdmihai: What happens if both are set? :)

cdmihai:
:shipit:

src/Build/Logging/FileLogger.cs
```diff
@@ -269,7 +274,7 @@ private void ApplyFileLoggerParameter(string parameterName, string parameterValu
         /// <summary>
         /// Encoding for the output. Defaults to UTF-8.
         /// </summary>
-        private Encoding _encoding = Encoding.UTF8;
+        private Encoding _encoding = new UTF8Encoding(false);
```
KirillOsenkov: You don't need to pass false because the default ctor of UTF8Encoding passes false already:
http://referencesource.microsoft.com/#mscorlib/system/text/utf8encoding.cs,65

PR #2114, opened 2017-05-19
Avoid mangling backslashes that escape quotes for VB defines

rainersigwald:


rainersigwald:

(both failed in NuGet restore which I didn't touch)

rainersigwald:

(both failed in NuGet restore which I didn't touch)

cdmihai:

cdmihai:

(multiple commands per comment don't work :( )

src/Utilities.UnitTests/ToolTask_Tests.cs
```diff
@@ -780,5 +780,19 @@ public void GetProcessStartInfoCanOverrideEnvironmentVariables()
...
         }
+
+        [Fact]
+        public void VisualBasicLikeEscapedQuotesInCommandAreNotMadeForwardSlashes()
+        {
+            MyTool t = new MyTool();
+            MockEngine engine = new MockEngine();
+            t.BuildEngine = engine;
+            t.MockCommandLineCommands = NativeMethodsShared.IsWindows
+                                            ? "/C echo \"hello \\\"world\\\"\""
```
cdmihai: Can you make this a verbatim @"" string so it's easier on the eye? :)

src/Utilities/ToolTask.cs
```diff
@@ -1715,6 +1715,38 @@ public override bool Execute()
...
+            StringBuilder sb = new StringBuilder(input);
+
+            int length = sb.Length;
+
+            for (int i = 0; i < length; i++)
+            {
+                // Backslashes must be swapped, because we don't
+                // know what inputs are paths or path fragments.
+                // But it's a common pattern to have backslash-escaped
+                // quotes inside quotes--especially for the VB /define:
```
cdmihai: >VB /define: [](start = 60, length = 11)

Can you give a concrete example to nail down the idea? :)

src/Utilities.UnitTests/ToolTask_Tests.cs
rainersigwald: Since quotes are special in verbatim strings it's still confusing. The question is whether you prefer

```
 "-c echo \"hello \\\"world\\\"\""
```

or

```
@"-c echo ""hello \""world\"""""
```

In my head the former is marginally less confusing. But juuuuuuuust barely!

src/Utilities/ToolTask.cs
```diff
@@ -1715,6 +1715,38 @@ public override bool Execute()
...
+                // parameter. So don't replace a backslash immediately
+                // followed by a quote.
+                if (sb[i] == '\\' && (i == length - 1 || sb[i + 1] != '"'))
```
cdmihai: What are all the escapable characters? Like space, double quotes, single quotes, backticks, piping characters?. Is it worth doing the fix for them as well?

src/Utilities.UnitTests/ToolTask_Tests.cs
cdmihai: Read it from a file :p
I agree with you, non-verbatim form is easier to read.

src/Utilities/ToolTask.cs
```diff
@@ -1715,6 +1715,41 @@ public override bool Execute()
...
         /// <summary>
+        /// Replace backslashes with OS-specific path separators,
+        /// except when likely that the backslash is intentional.
+        /// </summary>
+        /// <remarks>
+        /// Not a static method so that an implementation can
+        /// override with more-specific knowledge of what backslashes
+        /// are likely to be correct.
+        /// </remarks>
+        virtual protected string ReplaceBackslashes(string input)
```
cdmihai: Now that it's a public type, I wonder if  `NormalizeString` might be a better name. Or `NormalizeToolArguments`

rainersigwald:
Re: allowing escaping of other characters.

I thought about this too. Ideally, of course, we wouldn't have to do any normalization here because the command-line inputs would already be normalized.

Right now the priority is unblocking VB, and this covers that. I'm concerned that expanding the scope of the change will break something else. For example, space is escapable in shells, and straightforwardly we'd leave `/outputFolder:bin\ /something` as an escaped space. But that's probably wrong--that to me looks like a path fragment that should become `/outputFolder:bin/` and `/something`.

Fortunately, `ToolTask` should encompass a single (top-level) process creation, so things like backticks and pipes shouldn't be in play here.

@AndyGerlicher thoughts on going for generality versus scoping to the known-problematic situation?

PR #2132, opened 2017-05-24
[WIP] add F# WriteCodeFragment

enricosada:
ref https://github.com/dotnet/netcorecli-fsc/issues/93

to generate the `assemblyinfo.fs` from fsproj properties.

@KevinRansom the plan is to do it like c# does for coreclr (where codedom doesnt exists anyway), like [here in src/Tasks/WriteCodeFragment.cs#L294-L313](https://github.com/Microsoft/msbuild/blob/2e302ddec9dab44f5f71993b2e5eaa67d7a58310/src/Tasks/WriteCodeFragment.cs#L294-L313) 

/cc @rainersigwald

### DIFF
--- a/src/Tasks.UnitTests/WriteCodeFragment_Tests.cs
+++ b/src/Tasks.UnitTests/WriteCodeFragment_Tests.cs
@@ -641,6 +641,32 @@ namespace Microsoft.Build.UnitTests
             File.Delete(task.OutputFile.ItemSpec);
         }
 
+        /// <summary>
+        /// Test with the VB language
+        /// </summary>
+        [Fact]
+        [Trait("Category", "mono-osx-failing")]
+        [Trait("Category", "fsharp")]
+        public void OneAttributeNoParamsFSharp()
+        {
+            WriteCodeFragment task = new WriteCodeFragment();
+            MockEngine engine = new MockEngine(true);
+            task.BuildEngine = engine;
+            TaskItem attribute = new TaskItem("System.AssemblyTrademarkAttribute");
+            task.AssemblyAttributes = new TaskItem[] { attribute };
+            task.Language = "f#";
+            task.OutputDirectory = new TaskItem(Path.GetTempPath());
+            bool result = task.Execute();
+
+            Assert.Equal(true, result);
+
+            string content = File.ReadAllText(task.OutputFile.ItemSpec);
+            Console.WriteLine(content);
+
+            CheckContentFSharp(content, "[<assembly: System.AssemblyTrademarkAttribute()>]");
+        }
+
+
         private static void CheckContentCSharp(string actualContent, params string[] expectedAttributes)
         {
             CheckContent(
@@ -661,6 +687,16 @@ namespace Microsoft.Build.UnitTests
                 "Option Explicit On",
                 "Imports System",
                 "Imports System.Reflection");
+        }
+
+        private static void CheckContentFSharp(string actualContent, params string[] expectedAttributes)
+        {
+            CheckContent(
+                actualContent,
+                expectedAttributes,
+                "//",
+                "open System",
+                "open System.Reflection");
         }
 
         private static void CheckContent(string actualContent, string[] expectedAttributes, string commentStart, params string[] expectedHeader)
--- a/src/Tasks/WriteCodeFragment.cs
+++ b/src/Tasks/WriteCodeFragment.cs
@@ -152,6 +152,12 @@ namespace Microsoft.Build.Tasks
         [SuppressMessage("Microsoft.Globalization", "CA1305:SpecifyIFormatProvider", MessageId = "System.IO.StringWriter.#ctor(System.Text.StringBuilder)", Justification = "Reads fine to me")]
         private string GenerateCode(out string extension)
         {
+            if (Language.ToLowerInvariant() == "f#")
+            {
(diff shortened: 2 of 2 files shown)


AndyGerlicher:
Thanks for opening a PR. Unfortunately we've already closed a similar PR (#2058) after expressing concerns that we would not like to have this logic in MSBuild. If you think this is not correct please open an issue to discuss.

enricosada:
@AndyGerlicher while long term this is not the best place, atm the `dotnet/sdk` use the `WriteCodeFragment` in the language agnostic [Microsoft.NET.GenerateAssemblyInfo.targets#L93-L96](https://github.com/dotnet/sdk/blob/8204b6b1f3ff2f9717c5d66ab903795271de2e9c/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets#L93-L96).

So maybe is possibile to make a special case for F# in `dotnet/sdk`, but msbuild contains both c# and VB atm, so may make sense to leave it there and add F#, to make it work for sdk 2.0?

@AndyGerlicher i need to open an issue to discuss, or can be discussed there? /cc @KevinRansom @cartermp

enricosada:
Adding @dsplaisted @eerhardt for an hint about this, because `dotnet/sdk` is using the `WriteCodeFragment` **in the language agnostic** [Microsoft.NET.GenerateAssemblyInfo.targets#L93-L96](https://github.com/dotnet/sdk/blob/8204b6b1f3ff2f9717c5d66ab903795271de2e9c/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets#L93-L96).

F# now is not supported in `WriteCodeFragment`, so the file is not generated.

Dunno if there is a better place to discuss, but pratically all the code needed is already there (i just need to add more unit tests, atm is just one), so maybe easier to see on what adding this mean for msbuild  @AndyGerlicher

KevinRansom:
@enricosada 

We can override : tps://github.com/dotnet/sdk/blob/8204b6b1f3ff2f9717c5d66ab903795271de2e9c/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets#L51

in our targets file and sdk apps will work fine.


But we should also send a PR to the dotnet sdk refactoring the current implementation a bit.
@nguerrera said he would welcome a PR that improved this :-)

The bit where the attributes are generated should be moved to a new target:  GetAssemblyAttributes or some such name.

And then : CoreGenerateAssemblyInfo just handles the write.

We can override CoreGenerateAssemblyInfo  to write the attributes with a build task added to the Microsoft.FSharp.Build library ... or something.

Does that make sense?

Enrico,
thanks for pushing on this, it's really great that you take the time to dive into all of the stuff.

Kevin

dasMulli:
> But we should also send a PR to the dotnet sdk refactoring the current implementation a bit.

There's already https://github.com/dotnet/sdk/pull/1007 which splits the target into `_CalculateAssemblyAttributes` and `CoreGenerateAssemblyInfo` but it probably won't land in its current form (see PR discussion. I'll be sending out a PR today for an alternative implementation)

enricosada:
Awesome @dasMulli , can you please add me and @KevinRansom and reference the f# issue  https://github.com/Microsoft/visualfsharp/issues/3113 in that PR, so we'll update.

src/Tasks.UnitTests/WriteCodeFragment_Tests.cs
```diff
@@ -641,6 +641,32 @@ public void OneAttributePositionalAndNamedParamsVisualBasic()
             File.Delete(task.OutputFile.ItemSpec);
         }
 
+        /// <summary>
+        /// Test with the VB language
```
saul: F#, not VB

src/Tasks.UnitTests/WriteCodeFragment_Tests.cs
```diff
@@ -641,6 +641,32 @@ public void OneAttributePositionalAndNamedParamsVisualBasic()
...
+            task.Language = "f#";
+            task.OutputDirectory = new TaskItem(Path.GetTempPath());
+            bool result = task.Execute();
+
+            Assert.Equal(true, result);
+
+            string content = File.ReadAllText(task.OutputFile.ItemSpec);
+            Console.WriteLine(content);
+
+            CheckContentFSharp(content, "[<assembly: System.AssemblyTrademarkAttribute()>]");
```
saul: Maybe test with multiple attributes and attributes with parameters?

src/Tasks/WriteCodeFragment.cs
```diff
@@ -152,6 +152,12 @@ public override bool Execute()
         [SuppressMessage("Microsoft.Globalization", "CA1305:SpecifyIFormatProvider", MessageId = "System.IO.StringWriter.#ctor(System.Text.StringBuilder)", Justification = "Reads fine to me")]
         private string GenerateCode(out string extension)
         {
+            if (Language.ToLowerInvariant() == "f#")
+            {
+                //no codedoom for F#, fallback to coreclr version
```
saul: "Code DOM" not "codedoom"

enricosada:
Thx for review @saul but this is going to Ben replaced with another, without writecodegragment

PR #2133, opened 2017-05-25
Fix issue where targets are executed after an error.

### LINKED ISSUE #2116: In certain cases, errors in one target don't stop a subsequent target from running and generating errors
In the .NET SDK, we have targets that run early on in the build process to verify things like whether the target framework is valid or not.  If we generate errors in those targets, we would expect targets that depend on the target that is generating an error not to run and not to generate errors.  However, this is not what is happening.  This means that the error that specifies the root cause of the problem can be lost amidst the many subsequent errors that might be generated.

See https://github.com/dotnet/sdk/issues/798 for more context.

Here are two minimal repros of the issue:

# GetReferenceAssemblyPaths error

```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
  
  <Target Name="_CheckForUnsupportedTargetFramework"
      BeforeTargets="_CheckForInvalidConfigurationAndPlatform;_CheckCompileDesignTimePrerequisite;GetFrameworkPaths;GetReferenceAssemblyPaths;GenerateBuildRuntimeConfigurationFiles">
    <Error Text="Error in _CheckForUnsupportedTargetFramework" />
  </Target>

  <Target Name="_CheckForInvalidConfigurationAndPlatform" BeforeTargets="$(BuildDependsOn);Build" />
  
  <PropertyGroup>
    <BuildDependsOn>
      EntityDeploy;
      CoreBuild;
    </BuildDependsOn>
  </PropertyGroup>
  
  <Target Name="Build" DependsOnTargets="$(BuildDependsOn)" Returns="$(TargetPath)" />

  <Target Name="CoreBuild" DependsOnTargets="PrepareForBuild;GenerateBuildRuntimeConfigurationFiles"/>

  <Target Name="PrepareForBuild" DependsOnTargets="GetReferenceAssemblyPaths"/>
  
  <Target Name="GetReferenceAssemblyPaths">
    <Error Text="GetReferenceAssemblyPaths Error" />
  </Target>

  <Target Name="EntityDeploy" Condition="'@(EntityDeploy)' != ''" />
</Project>
```

In this repro, I would hope that the error in `_CheckForUnsupportedTargetFramework` would prevent subsequent targets from being run.  Something about the `EntityDeploy` target being added to the start of `BuildDependsOn`, and possibly the fact that `BuildDependsOn` is being used both in `Build` as `DependsOnTargets` and in `_CheckForInvalidConfigurationAndPlatform` as `BeforeTargets` is causing `GetReferenceAssemblyPaths` to run even after a previous target has failed.

# GenerateBuildRuntimeConfigurationFiles error

```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
  
  <Target Name="_CheckForUnsupportedTargetFramework"
      BeforeTargets="_CheckForInvalidConfigurationAndPlatform;_CheckCompileDesignTimePrerequisite;GetFrameworkPaths;GetReferenceAssemblyPaths;GenerateBuildRuntimeConfigurationFiles">
    <Error Text="Error in _CheckForUnsupportedTargetFramework" />
  </Target>

  <Target Name="_CheckForInvalidConfigurationAndPlatform" BeforeTargets="$(BuildDependsOn);Build" />
  
  <PropertyGroup>
    <BuildDependsOn>
      CoreBuild;
    </BuildDependsOn>
  </PropertyGroup>
  
  <Target Name="Build" DependsOnTargets="$(BuildDependsOn)" Returns="$(TargetPath)" />

  <Target Name="CoreBuild" DependsOnTargets="PrepareForBuild;GenerateBuildRuntimeConfigurationFiles">
    <OnError ExecuteTargets="_CleanRecordFileWrites" />
  </Target>

  <Target Name="PrepareForBuild" DependsOnTargets="GetReferenceAssemblyPaths"/>
  
  <Target Name="GetReferenceAssemblyPaths">
    
  </Target>

  <Target Name="_CheckForCompileOutputs"/>
  <Target Name="_CleanGetCurrentAndPriorFileWrites" DependsOnTargets="_CheckForCompileOutputs" />
  <Target Name="_CleanRecordFileWrites" DependsOnTargets="_CleanGetCurrentAndPriorFileWrites"/>

  <Target Name="GenerateBuildRuntimeConfigurationFiles" BeforeTargets="_CheckForCompileOutputs">
    <Error Text="GenerateBuildRuntimeConfigurationFiles Error" />
  </Target>
  
</Project>
```

This is more straightforward.  When I run this I'd expect to only get the error from `_CheckForUnsupportedTargetFramework`.  However, I also get the error from `GenerateBuildRuntimeConfigurationFiles`.  It looks like the `OnError` task in `CoreBuild` is causing `_CleanRecordFileWrites` and its dependencies to run after an error is encountered.  I'm not sure exactly how to fix this though.

AndyGerlicher:
First attempt, looking for feedback. It works and fixes the error, but it scares me a bit!

This issue occurs when two targets are scheduled to be built (e.g.
ProduceError1 and ProduceError2) and the condition of the first target
causes it to not execute and a dependent target
(BeforeTargets='ProduceError1') fails. In this example, ProduceError2
still executes.

Fixes #2116

### DIFF
--- a/src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
@@ -1208,6 +1208,42 @@ namespace Microsoft.Build.UnitTests.BackEnd
             }
         }
 
+        [Fact]
+        public void SkippedTargetWithFailedDependenciesStopsBuild()
+        {
+            string projectContents = @"
+  <Target Name=""Build"" DependsOnTargets=""ProduceError1;ProduceError2"" />
+  <Target Name=""ProduceError1"" Condition=""false"" />
+  <Target Name=""ProduceError2"">
+    <ErrorTask2 />
+  </Target>
+  <Target Name=""_Error1"" BeforeTargets=""ProduceError1"">
+    <ErrorTask1 />
+  </Target>
+";
+
+            var project = CreateTestProject(projectContents, string.Empty, "Build");
+            TargetBuilder builder = (TargetBuilder)_host.GetComponent(BuildComponentType.TargetBuilder);
+            MockTaskBuilder taskBuilder = (MockTaskBuilder)_host.GetComponent(BuildComponentType.TaskBuilder);
+            // Fail the first task
+            taskBuilder.FailTaskNumber = 1;
+
+            IConfigCache cache = (IConfigCache)_host.GetComponent(BuildComponentType.ConfigCache);
+            BuildRequestEntry entry = new BuildRequestEntry(CreateNewBuildRequest(1, new[] { "Build" }), cache[1]);
+
+            var buildResult = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
+
+            IResultsCache resultsCache = (IResultsCache)_host.GetComponent(BuildComponentType.ResultsCache);
+            Assert.True(resultsCache.GetResultForRequest(entry.Request).HasResultsForTarget("Build"));
+            Assert.True(resultsCache.GetResultForRequest(entry.Request).HasResultsForTarget("ProduceError1"));
+            Assert.False(resultsCache.GetResultForRequest(entry.Request).HasResultsForTarget("ProduceError2"));
+            Assert.True(resultsCache.GetResultForRequest(entry.Request).HasResultsForTarget("_Error1"));
+
+            Assert.Equal(TargetResultCode.Failure, resultsCache.GetResultForRequest(entry.Request)["Build"].ResultCode);
+            Assert.Equal(TargetResultCode.Skipped, resultsCache.GetResultForRequest(entry.Request)["ProduceError1"].ResultCode);
+            Assert.Equal(TargetResultCode.Failure, resultsCache.GetResultForRequest(entry.Request)["_Error1"].ResultCode);
+        }
+
         #region IRequestBuilderCallback Members
 
         /// <summary>
--- a/src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
@@ -609,10 +609,16 @@ namespace Microsoft.Build.BackEnd
                     }
                 }
 
-                // Mark our parent for error execution
+                // Mark our parent for error execution when it is not Completed (e.g. Executing)
                 if (topEntry.ParentEntry != null && topEntry.ParentEntry.State != TargetEntryState.Completed)
                 {
                     topEntry.ParentEntry.MarkForError();
+                }
+                // In cases where we need to indicate a failure but the ParentEntry was Skipped (due to condition) it must be
+                // marked stop to prevent other targets from executing.
+                else if (topEntry.ParentEntry?.Result?.ResultCode == TargetResultCode.Skipped)
+                {
+                    topEntry.ParentEntry.MarkForStop();
                 }
             }
         }
--- a/src/Build/BackEnd/Components/RequestBuilder/TargetEntry.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TargetEntry.cs
@@ -753,6 +753,20 @@ namespace Microsoft.Build.BackEnd
         }
 
         /// <summary>
+        /// This method is used by the Target Builder to indicate that a child of this target has failed and that work should not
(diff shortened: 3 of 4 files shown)


src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
```diff
@@ -1208,6 +1208,43 @@ public void TestCancelWithNoEntriesAfterBuild()
...
+";
+
+            var project = CreateTestProject(projectContents, string.Empty, "Build");
+            TargetBuilder builder = (TargetBuilder)_host.GetComponent(BuildComponentType.TargetBuilder);
+            MockTaskBuilder taskBuilder = (MockTaskBuilder)_host.GetComponent(BuildComponentType.TaskBuilder);
+            // Fail the first task
+            taskBuilder.FailTaskNumber = 1;
+
+            IConfigCache cache = (IConfigCache)_host.GetComponent(BuildComponentType.ConfigCache);
+            BuildRequestEntry entry = new BuildRequestEntry(CreateNewBuildRequest(1, new[] { "Build", "ProduceError1", "_Error1" }), cache[1]);
```
rainersigwald: Do you need to explicitly request all three?

src/Build/BackEnd/Components/RequestBuilder/TargetEntry.cs
```diff
@@ -748,7 +748,20 @@ internal void EnterLegacyCallTargetScope(Lookup lookup)
         /// </summary>
         internal void MarkForError()
         {
-            ErrorUtilities.VerifyThrow(_state != TargetEntryState.Completed, "State must not be Completed. State is {0}.", _state);
+            // In the case that the Target was skipped due to a condition but a dependent target failed subsequent execution
+            // must stop. Detect that case and set the WorkUnitActionCode to stop and mark the TargetEntry for ErrorExecution.
```
rainersigwald: So:
* We're in this method because a dependent target failed
* But this target is already marked completed/skipped/continue

And therefore we want to update the current status to completed/skipped/stop?

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -609,11 +609,16 @@ private void PopDependencyTargetsOnTargetFailure(TargetEntry topEntry, TargetRes
...
                 }
 
-                // Mark our parent for error execution
+                // Mark our parent for error execution when it is not Completed (e.g. Executing) or when the target was skipped but
+                // needs to be marked for error to stop other targets from executing.
                 if (topEntry.ParentEntry != null && topEntry.ParentEntry.State != TargetEntryState.Completed)
                 {
                     topEntry.ParentEntry.MarkForError();
                 }
+                else if (topEntry.ParentEntry != null && topEntry.ParentEntry.Result?.ResultCode == TargetResultCode.Skipped)
```
rainersigwald:
```c#
topEntry.ParentEntry?.Result?.ResultCode == TargetResultCode.Skipped
```

src/Build/BackEnd/Components/RequestBuilder/TargetEntry.cs
```diff
@@ -753,6 +753,20 @@ internal void MarkForError()
...
 
         /// <summary>
+        /// This method is used by the Target Builder to indicate that a child of this target has failed and that work should not
+        /// continue in Completed / Skipped mode. We do not want to mark the state to run in ErrorExecution mode so that the
+        /// OnError targets do not run (the target was skipped due to condition so OnError targets should not run).
+        /// </summary>
+        internal void MarkForStop()
+        {
+            ErrorUtilities.VerifyThrow(_state == TargetEntryState.Completed, "State must be Completed. State is {0}.", _state);
+            ErrorUtilities.VerifyThrow(_targetResult.ResultCode == TargetResultCode.Skipped, "State must be Completed. State is {0}.", _state);
```
rainersigwald: copypasta: state/resultcode

AndyGerlicher:

dsplaisted:
@AndyGerlicher Thanks a bunch for fixing this!  I've verified that it fixes a lot of the issues with extra errors in the .NET SDK.

How will this flow into a release?  Does it need to go into a 15.3 branch in order to insert it into the CLI and VS?

rainersigwald:
@dsplaisted This went to `master` which is currently our branch for 15.3 preview 3 and should follow the normal code flows. @AndyGerlicher is working on getting it pulled into VS now.

PR #2135, opened 2017-05-25
Add ProjectEvaluationStartedEventArgs and ProjectEvaluationFinishedEventArgs

jeffkl:
Also updated binary log serialization to support it.  Loggers will receive the message through the `IEventSource.StatusEventRaised` event.  The messages will no longer show up in the text log.

An upcoming change will look at these events so that project evaluation can be displayed in the performance summary.

PR #2129, opened 2017-05-24
Evaluations have unique IDs

cdmihai:
Implements part of #2032 by adding an evaluation ID to evaluations and exposing the ID via new APIs `Project.LastEvaluationID` and `ProjectInstance.EvaluationID`. The Evaluation ID is available to event consumers via `BuildEventContext.EvaluationID` (currently only for events triggered during evaluation).

I also updated the binary logger to serialize the extra `BuildEventContext.EvaluationID`. Since it's not released yet, it's "okay" to break the format.

This enables CPS to tie log messages to project objects.

Future work that will be covered in another PR: Find a way to pipe the evaluation ID everywhere. This would help @KirillOsenkov's MSBuildStructuredLogger better group evaluation messages.

PR #2136, opened 2017-05-25
Bump up Binarry Logger file format version

cdmihai:
Due to new BuildEventContext.EvaluationID

It is okay to break people because the binary logger has not RTMed yet. After 15.3, each version bump needs to introduce some dispatching code so that the current version is backwards compatible. Worst case design is forking the reader / writer for each version number.


cdmihai:
Includes changes from #2135 and #2129

cdmihai:

cdmihai:
@dotnet-bot test Windows_NT Build for CoreCLR  please

PR #2137, opened 2017-05-25
Use base IsOSPlatform implementation.

### LINKED ISSUE #2127: [MSBuild]::IsOsPlatform implementation breaks API intent of RuntimeInformation
I just stumbled over https://github.com/Microsoft/msbuild/pull/1950. The implementation of the `[MSBuild]::IsOsPlatform` property function has a questionable implementation, it whitelists only members in the OSPlatform struct:

```csharp
private static Lazy<string> _validOsPlatforms = new Lazy<string>(
             () => typeof(OSPlatform).GetTypeInfo()
                 .GetProperties(BindingFlags.Static | BindingFlags.Public)
                 .Where(pi => pi.PropertyType == typeof(OSPlatform))
                 .Select(pi => pi.Name)
                 .Aggregate("", (a, b) => string.IsNullOrEmpty(a) ? b : $"{a}, {b}"),
             true);
```

That is against the API design of RuntimeInformation.IsOSPlatform(), after a long discussion in https://github.com/dotnet/corefx/issues/1017 and https://github.com/dotnet/apireviews/tree/master/2015/06-09-os-version#1017-os-version-information.

It was explicitly designed to allow new platforms to come online without needing to update the struct with new members.

The property function should probably just pass the string down: 

```csharp
internal static bool IsOsPlatform(string platformString)
{
   return RuntimeInformation.IsOSPlatform(OSPlatform.Create(platformString));
}
```

I'd also suggest renaming it to IsOSPlatform (upper-case S) to match with RuntimeInformation.IsOSPlatform().

The only quirk is that OSPlatform seem to do a case-sensitive comparison of the string, i.e. `WINDOWS` would work but `windows` won't.

It'd be quite unfortunate if MSBuild created _yet another_ slightly different API to do platform checks.

/cc @cdmihai @rainersigwald

AndyGerlicher:
We lose the ability to give a nicer error message when a platform
doesn't exist and case insensitivity, but ultimately not having a
diverged method of checking for what OS we're on is preferable.

Closes #2127

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -2637,7 +2637,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
         }
 
         [Fact]
-        public void IsOsPlatformShouldBeCaseSensitiveToParameter()
+        public void IsOsPlatformShouldBeCaseInsensitiveToParameter()
         {
             var pg = new PropertyDictionary<ProjectPropertyInstance>();
             var expander = new Expander<ProjectPropertyInstance, ProjectItemInstance>(pg);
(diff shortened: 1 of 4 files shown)


AndyGerlicher:

AndyGerlicher:

AndyGerlicher:

AndyGerlicher:

akoeplinger:
Awesome, thanks for the quick fix 👍

PR #2139, opened 2017-05-25
Find ref asms after ResolveReferences

### LINKED ISSUE #2134: [Breaking Change] Targets that run after ResolveReferences to manipulate references are broken after upgrading to 15.3
Version 15.3.0-pre.1.0+26524.3003.d15prerel

[This target](https://github.com/dotnet/project-system/blob/master/build/Targets/VSL.Imports.targets#L340) which we copied from a package that comes in the VS SDK is broken in this build:

``` XML
  <!-- This is a copy of the Microsoft.VisualStudio.SDK.EmbedInteropTypes NuGet package, but only the list of
       assemblies that we need. The package includes things like EnvDTE which are reasonable for consumers, but
       strange since we actually _implement_ DTE and use it as an exchange type with generics in a few places. -->
  <Target Name="LinkVSSDKEmbeddableAssemblies" DependsOnTargets="ResolveReferences" AfterTargets="ResolveReferences">
    <ItemGroup>
      <ReferencePath Condition="
              '%(FileName)' == 'Microsoft.VisualStudio.Shell.Embeddable'
           or '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.12.0'
           or '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.12.1.DesignTime'
           or '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime'
           or '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.15.0.DesignTime'
           or '%(Filename)' == 'Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime'
           or '%(FileName)' == 'Microsoft.VisualStudio.TextManager.Interop.12.1.DesignTime'
           or '%(FileName)' == 'Microsoft.VisualStudio.ProjectSystem.Interop'
           or '%(FileName)' == 'stdole'
           or '%(FileName)' == 'Microsoft.VisualStudio.CommandBars'
           or '%(FileName)' == 'NuGet.SolutionRestoreManager.Interop'
           or '%(FileName)' == 'NuGet.VisualStudio'
              ">
        <EmbedInteropTypes>true</EmbedInteropTypes>
      </ReferencePath>
    </ItemGroup>
  </Target>
```

ie Changes this target makes to the ReferencePath item are not respected at compile time.

This silently breaks the product until runtime, after building with this binary - it fails to load because Visual Studio doesn't ship with any of the .DesignTime dlls, hence why they are embedded. Manipulating `ReferencePath` is a pretty common extension point.

I'm guessing this was broken in the references assembly changes - haven't fully groked what changes we made there and where we landed with all this.

rainersigwald:
e3a3d69 added a new FindReferenceAssembliesForReferences target, which
ran just before ResolveReferences. Dogfooding this revealed targets in
the VS SDK (at least) that manipulate @(ReferencePath) and are hooked in
as AfterTargets of ResolveReferences. That was silently ignored,
corrupting the user's build.

Fixes #2134 by moving FindReferenceAssembliesForReferences later in the
process, so anyone hooking ResolveReferences will see the changes they
make in the compiler.

This is not a fully general solution to the problem that someone might
hook an arbitrary point and manipulate @(ReferencePath), but it unbreaks
an obvious hook point.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1372,14 +1372,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ResolveReferencesDependsOn>
   </PropertyGroup>
 
-  <!-- FindReferenceAssembliesForReferences needs to see a fully-populated @(ReferencePath),
-       so it should run after all of the subcomponents of ResolveReferences, including
-       user overrides of AfterResolveReferences and targets added by appending to
-       $(ResolveReferencesDependsOn) (like ImplicitlyExpandDesignTimeFacades). So list it
-       last explicitly, rather than adding it to the property.   -->
   <Target
       Name="ResolveReferences"
-      DependsOnTargets="$(ResolveReferencesDependsOn);FindReferenceAssembliesForReferences"/>
+      DependsOnTargets="$(ResolveReferencesDependsOn)"/>
 
   <!--
     ============================================================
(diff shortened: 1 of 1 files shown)


AndyGerlicher:

PR #2142, opened 2017-05-26
Add ProjectImportedEventArgs so loggers are notified when a project is imported

jeffkl:
This ties into the new evaluation context as well.

Event triggers if an import does not occur because the condition is false, an import does not occur because a wild card does not match, or an import occurs.


src/MSBuild.sln
```diff
@@ -607,44 +607,37 @@ Global
 		{11B5D53E-90E4-4BD5-9883-B5921F7DE854}.Release-NetCore|x64.Build.0 = Release|Any CPU
 		{11B5D53E-90E4-4BD5-9883-B5921F7DE854}.Release-NetCore|x86.ActiveCfg = Release|Any CPU
 		{11B5D53E-90E4-4BD5-9883-B5921F7DE854}.Release-NetCore|x86.Build.0 = Release|Any CPU
-		{53733ECF-0D81-43DA-B602-2AE9417F614F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{53733ECF-0D81-43DA-B602-2AE9417F614F}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{53733ECF-0D81-43DA-B602-2AE9417F614F}.Debug|Any CPU.ActiveCfg = Debug|x86
```
KirillOsenkov: This is for MSBuildTaskHost? Is this intentional?

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -231,16 +231,30 @@ private void Write(BuildMessageEventArgs e)
                 Write((CriticalBuildMessageEventArgs)e);
                 return;
             }
-            else if (e is TaskCommandLineEventArgs)
+            if (e is TaskCommandLineEventArgs)
```
KirillOsenkov: Nit: it'd be nice to have an empty line to separate the two if statements visually.

KirillOsenkov:
Will this tell me when an SDK style project is expanded?

KirillOsenkov:
It's awesome to have this, do you think we could get this in today so I can hack on the viewer over the weekend? I'd like to implement preprocessed view for projects.

jeffkl:
@KirillOsenkov sorry this didn't make it on Thursday because I hit an issue with testing it out.  I'm hoping to get things resolved in stand up today.

src/MSBuild.sln
jeffkl: No I'm just using a newer version of VS so I thought it only updated the version in the SLN...

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
jeffkl: Done

jeffkl:

KirillOsenkov:
So what about the SDK style projects? Will it report when imports are made for the SDK props and targets?

jeffkl:
@KirillOsenkov yes, all evaluated imports will be logged.

KirillOsenkov:
Awesome. So, let's get this in?

src/Build/Resources/Strings.resx
```diff
@@ -1604,13 +1604,30 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
     <comment>{StrBegin="MSB4231: "}</comment>
   </data>
 
+  <data name="UnsupportedOSPlatformString">
+    <value>MSB4239: The provided argument {0} for [MSBuild]::IsOsPlatform is not supported. Supported values are: {1}</value>
```
rainersigwald: This was removed, why is it coming back?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2375,6 +2375,29 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
             if (!EvaluateConditionCollectingConditionedProperties(importElement, ExpanderOptions.ExpandProperties,
                 ParserOptions.AllowProperties, _projectRootElementCache))
             {
+                if (!_evaluationLoggingContext.LoggingService.OnlyLogCriticalEvents)
```
rainersigwald: I think I changed my mind about having an escape hatch for this since yesterday. It's still not good enough if this causes an RPS regression, but it could be useful if it causes problems in some obscure scenario for a specific customer.

src/Build/Resources/Strings.resx
jeffkl: Sorry, bad merge.  I've fixed this

src/Build/Evaluation/Evaluator.cs
jeffkl: Escape hatch implemented

src/Build/Evaluation/Evaluator.cs
```diff
@@ -732,6 +734,8 @@ private static ProjectTargetInstance ReadNewTargetElement(ProjectTargetElement t
             _data.EvaluationID = NextEvaluationID();
             _evaluationLoggingContext = new EvaluationLoggingContext(loggingService, buildEventContext, _data.EvaluationID);
 
+            _logProjectImportedEvents = !_evaluationLoggingContext.LoggingService.OnlyLogCriticalEvents && !String.Equals("1", Environment.GetEnvironmentVariable("MSBUILDDONOTLOGIMPORTS"));
```
rainersigwald: Use the new fancy class from #2018?

jeffkl: 👍

KirillOsenkov:
Looks great!

KirillOsenkov:
BTW I've already implemented stitching preprocessed views together based on this. Works like a charm.

Issue #2165, opened 2017-06-02
Some transformations on @(ReferencePath) result in mangled references


rainersigwald:
Some builds fail with an error like

```
(CoreCompile target) ->
  CSC : error CS0006: Metadata file '%(FullPath)' could not be found
```

when building using VS 15.3 preview 2 builds (which are not yet public).

[This ItemReferenceGroup](https://github.com/Microsoft/msbuild/commit/e3a3d6998162665ab760518d0d0f2e338394b041#diff-02872751bc64e0ff785c9b6c1cd919b5R2082) sets a default value for the `ReferenceAssembly` metadata of `ReferencePath` items to `%(FullPath)`. The engine evaluates that as a literal string, but it can be successfully read through by item transformations like [`FindReferenceAssembliesForReferences`'s](https://github.com/Microsoft/msbuild/commit/e3a3d6998162665ab760518d0d0f2e338394b041#diff-02872751bc64e0ff785c9b6c1cd919b5R2060) `@(ReferencePath->'%(ReferenceAssembly)')`, because the evaluator recurses--it sees that literal, interprets it as a metadata reference, and reads the `FullPath` metadata.

However, not all transformations perform that recursion. An `ITask` implementation that uses `return new TaskItem(oldItem)`, for example, evidently escapes the metadatum string, and when it is expanded it turns into a literal `%(FullPath)` instead of the value of the `FullPath` metadatum.

`FindReferenceAssembliesForReferences` then populates `ReferencePathWithRefAssemblies` with a list of items whose identity is `%(FullPath)`, which isn't valid and confuses the compiler, producing the error above.

This is a regression introduced by e3a3d69 when implementing #1986.

rainersigwald:
You can work around the bug by setting the property

```xml
<CompileUsingReferenceAssemblies>false</CompileUsingReferenceAssemblies>
```

which avoids transforming `@(ReferencePath)` using the recursive metadata expansion.

rainersigwald:
Thanks for reporting this @KirillOsenkov and @ToddGrun.

FYI @dsplaisted--the task that hit this for them was StrongNamer. Filed https://github.com/dsplaisted/strongnamer/issues/14 in the hopes that anyone who sees it will find it there.

rainersigwald:
fyi @jcouv -- I'll also want to fix this in 15.3 p3.

PR #2167, opened 2017-06-02
Ref asm bugfixes

rainersigwald:
These look like good places to use ref assemblies if present--but we need to track down owners for the various tools here and confirm that they're correct.

@jcouv


rainersigwald:
Updated with a fix for #2165 and removal of the UpToDateCheckInputs items (which should be an extremely minor perf boost).

rainersigwald:
Per conversation with @AndyGerlicher, we're going to postpone the non-GenerateResources part of this fix until we can confirm that using ref assemblies is correct with tool owners.

The risks are that if it's wrong, we might break all consumers of the winmdexp and lc targets. If we continue with the current state, reference assemblies are less effective in some scenarios. That seems like a good tradeoff at this stage of 15.3.

jcouv:
Makes sense. Let's keep an open issue?

rainersigwald:
@jcouv Already had the commit so I just opened #2181 to take the changes that I noticed when inspecting common.targets.

PR #2171, opened 2017-06-05
Ignore missing imports

### LINKED ISSUE #1727: Add ability to ignore imports that are empty files
In certain cases when creating new imports within a UI, a file might not have any content yet.  For instance, if you create a new `Directory.Build.props` within a project, Visual Studio will immediately re-evaluate the project and since the newly created file is empty, the evaluation will fail and the project will unload.

I propose a new `ProjectLoadSetting` like `IgnoreEmptyImports` that treats an empty file as an empty project.  It would also treat empty XML files as empty if they only contained the XML declaration like:

```xml
<? xml version="1.0" encoding="utf-8" ?>
```

Since its a new project load setting, it would be opt-in and probably only used by UIs.  I've taken a first stab at this here:

https://github.com/Microsoft/msbuild/compare/xplat...jeffkl:IgnoreEmptyImports?expand=1

jeffkl:
New ProjectLoadSetting to control the feature

New code only runs after the project fails to load and the file existence is checked which should not impact performance.

Closes #1727

### DIFF
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -759,6 +759,7 @@ namespace Microsoft.Build.Evaluation
     public enum ProjectLoadSettings
     {
         Default = 0,
+        IgnoreEmptyImports = 16,
         IgnoreMissingImports = 1,
         RecordDuplicateButNotCircularImports = 2,
         RecordEvaluatedItemElements = 8,
--- a/ref/netstandard1.3/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/netstandard1.3/Microsoft.Build/Microsoft.Build.cs
@@ -747,6 +747,7 @@ namespace Microsoft.Build.Evaluation
     public enum ProjectLoadSettings
     {
         Default = 0,
+        IgnoreEmptyImports = 16,
         IgnoreMissingImports = 1,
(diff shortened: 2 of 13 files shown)


jeffkl:
@jviau would you want this for the next release?  This would fix the bug where Visual Studio immediately if you create an empty `Directory.Build.props` in a project.

jviau:
@jeffkl yeah this would be great to have.

I am wondering if we can even go a step further and have a way to handle/ignore invalid imports. I implemented a feature for CPS to auto-reload changed imports on disk, and to revert back to in-memory XML if the import is now invalid. The problem is I can't deal with the case where a condition import transitions from `false` to `true` and the newly imported file is invalid. My system relies on file changed notifications, and there is no way to know that some arbitrary change is going to bring in a new invalid import.

jeffkl:
@jviau I've opened https://github.com/Microsoft/msbuild/issues/2179 to track that, sounds like a good feature.

jeffkl:

jviau:
@jeffkl are the skipped imports exposed via any API? I would like to know what imports were skipped so I can watch them on disk and ask msbuild to re-evaluate when they change.

jeffkl:
@jviau yes we now log an event that's sent to `IEventSource.MessageRaised` event.  You can check the event args type to be `ProjectImportedEventArgs` and then cast it.  Let me know if you need more detail.

PR #2177, opened 2017-06-06
Always use response file for winmdexp

### LINKED ISSUE #2140: Use a response file in WinMDExp task
An internal team ran into a problem that caused `WinMDExp` to give the error

```
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(2352,7): warning MSB6002: The command-line for the "WinMDExp" task is too long. Command-lines longer than 32000 characters are likely to fail. Try reducing the length of the command-line by breaking down the call to "WinMDExp" into multiple calls with fewer parameters per call.
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(2352,7): error MSB6003: The specified task executable "winmdexp.exe" could not be run. The filename or extension is too long
1>Done executing task "WinMDExp" -- FAILED.
```

`WinMDExp` is a `ToolTaskExtension`, so the error is happening when it calls `CreateProcess` on `winmdexp.exe` with a very long command line.

Fortunately, `winmdexp.exe` can take a response file (according to its help), so this can be avoided entirely by calling `ToolTaskExtension.AddResponseFileCommands()` instead of `AddCommandLineCommands()`.

rainersigwald:
Fixes #2140 by always using a response file for winmdexp, avoiding
issues with too-long calls to CreateProcess.

Concerns:

- [x] Are there any arguments that _cannot_ go in the response file?
- [x] Do we need an option to go back to the old/well-tested way?
- [x] Or should this be opt-in instead?
- [x] Do all versions of `winmdexp.exe` support the response file? If not, should support be detected somehow?

### DIFF
--- a/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
+++ b/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
@@ -1166,7 +1166,7 @@ namespace Microsoft.Build.Tasks
         public bool UTF8Output { get { throw null; } set { } }
         [Microsoft.Build.Framework.RequiredAttribute]
         public string WinMDModule { get { throw null; } set { } }
-        protected internal override void AddCommandLineCommands(Microsoft.Build.Tasks.CommandLineBuilderExtension commandLine) { }
+        protected internal override void AddResponseFileCommands(Microsoft.Build.Tasks.CommandLineBuilderExtension commandLine) { }
         protected override string GenerateFullPathToTool() { throw null; }
         protected override bool SkipTaskExecution() { throw null; }
         protected override bool ValidateParameters() { throw null; }
--- a/src/Tasks.UnitTests/WinMDExp_Tests.cs
+++ b/src/Tasks.UnitTests/WinMDExp_Tests.cs
@@ -33,11 +33,11 @@ namespace Microsoft.Build.UnitTests
             CommandLine.ValidateHasParameter(
                 t,
                 "/reference:mscorlib.dll",
-                false);
+                useResponseFile: true);
             CommandLine.ValidateHasParameter(
                 t,
                 "/reference:Windows.Foundation.winmd",
-                false);
+                useResponseFile: true);
         }
 
         [Fact]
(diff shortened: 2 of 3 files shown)


ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
```diff
@@ -1166,7 +1166,7 @@ public partial class WinMDExp : Microsoft.Build.Tasks.ToolTaskExtension
         public bool UTF8Output { get { throw null; } set { } }
         [Microsoft.Build.Framework.RequiredAttribute]
         public string WinMDModule { get { throw null; } set { } }
-        protected internal override void AddCommandLineCommands(Microsoft.Build.Tasks.CommandLineBuilderExtension commandLine) { }
+        protected internal override void AddResponseFileCommands(Microsoft.Build.Tasks.CommandLineBuilderExtension commandLine) { }
```
cdmihai: Looked on Index and nobody's extending WinMDExp, so I guess breaking this is fine :)

rainersigwald: Ugh, didn't even occur to me that this wasn't sealed . . .

rainersigwald:
@dotnet-bot test this please

rainersigwald:
@dotnet-bot test this please

rainersigwald:
@yizhang82 Offline you mentioned that every version of winmdexp supported the response file. Do you have opinions about whether we need to provide an opt-in or opt-out for using it?

The C# and VB compilers [always use a response file](https://github.com/dotnet/roslyn/blob/cee39cf0423391e18a6b2fa56f564a68ce2d0e38/src/Compilers/Core/MSBuildTask/ManagedCompiler.cs#L666-L674), so there shouldn't be any risk around MSBuild's support of that scenario, just the tool's.

yizhang82:
@rainersigwald Always using response file should be fine. If you run into a bug we'll be happy to fix it in latest release.

rainersigwald:
Great, thanks. This should be good to go then, @AndyGerlicher.

PR #2181, opened 2017-06-06
Use ref assemblies in more locations

rainersigwald:
Need to confirm these changes with owners of:

- [x] `WinMDExp`
- [ ] `LC`

Continuation of #2167 for the post-15.3 timeframe.


rainersigwald:
@yizhang82 Can you confirm that WinMDExp only looks at the public interface of managed assemblies passed to it with `/reference:`? If so, this change will mean that the `ExportWindowsMDFile` can be skipped more often in incremental builds, because it will only run when the public interface of a reference changes, rather than when any change is made to the referenced assembly.

yizhang82:
@rainersigwald Yes, we only care about public contracts in the reference assemblies.

PR #2192, opened 2017-06-07
RAR performance improvements on Core

rainersigwald:
I recommend going through this commit-by-commit; a couple of commits are straightforward refactors that look complicated in the final diff.

There are two improvements here:
* When using the non-Full Framework `AssemblyInformation` extraction, read a file only once, populating `Dependencies` and `FrameworkName` with a single `FileStream`
* Read `AssemblyInformation` from a given path exactly once (instead of twice).

This chips away at #2015.

### DIFF
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -32,6 +32,15 @@ namespace Microsoft.Build.Utilities
 
         public readonly bool DoNotLogProjectImports = Environment.GetEnvironmentVariable("MSBUILDDONOTLOGIMPORTS") == "1";
 
+        public readonly bool CacheAssemblyInformation =
+#if FEATURE_ASSEMBLY_LOADFROM
+            // On full framework, default to old/well-tested/known-good behavior, but allow an opt in.
+            Environment.GetEnvironmentVariable("MSBUILDCACHERARASSEMBLYINFORMATION") == "1";
+#else
+            // On .NET Core, go for speed, but allow an opt out.
+            Environment.GetEnvironmentVariable("MSBUILDDONOTCACHERARASSEMBLYINFORMATION") != "1";
+#endif
+
         public readonly ProjectInstanceTranslationMode? ProjectInstanceTranslation = ComputeProjectInstanceTranslation();
 
         /// <summary>
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -32,14 +32,10 @@ namespace Microsoft.Build.Utilities
 
         public readonly bool DoNotLogProjectImports = Environment.GetEnvironmentVariable("MSBUILDDONOTLOGIMPORTS") == "1";
 
-        public readonly bool CacheAssemblyInformation =
-#if FEATURE_ASSEMBLY_LOADFROM
-            // On full framework, default to old/well-tested/known-good behavior, but allow an opt in.
-            Environment.GetEnvironmentVariable("MSBUILDCACHERARASSEMBLYINFORMATION") == "1";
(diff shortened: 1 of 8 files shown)


src/Shared/Traits.cs
```diff
@@ -32,6 +32,15 @@ internal class EscapeHatches
 
         public readonly bool DoNotLogProjectImports = Environment.GetEnvironmentVariable("MSBUILDDONOTLOGIMPORTS") == "1";
 
+        public readonly bool CacheAssemblyInformation =
+#if FEATURE_ASSEMBLY_LOADFROM
+            // On full framework, default to old/well-tested/known-good behavior, but allow an opt in.
+            Environment.GetEnvironmentVariable("MSBUILDCACHEASSEMBLYINFORMATION") == "1";
```
cdmihai: I'd put RAR in the names of the escape hatches.

src/Shared/Traits.cs
```diff
@@ -32,6 +32,15 @@ internal class EscapeHatches
...
+#else
+            // On .NET Core, go for speed, but allow an opt out.
+            Environment.GetEnvironmentVariable("MSBUILDDONOTCACHEASSEMBLYINFORMATION") == "1";
```
cdmihai: Shouldn't this be different instead of equals? Cache when ("Do not cache" is not 1)

src/Shared/Traits.cs
rainersigwald: Good thought.

src/Shared/Traits.cs
rainersigwald: 😞

src/Tasks/SystemState.cs
```diff
@@ -480,6 +481,7 @@ private void GetAssemblyMetadata
                 getAssemblyMetadata
                 (
                     path,
+                    cache,
```
cdmihai: So much indirection :)

src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
```diff
@@ -7079,7 +7079,7 @@ public void ReferenceTableDependentItemsInBlackListPrimaryWithSpecificVersion()
 
         private static ReferenceTable MakeEmptyReferenceTable(TaskLoggingHelper log)
         {
-            ReferenceTable referenceTable = new ReferenceTable(null, false, false, false, false, new string[0], null, null, null, null, null, null, SystemProcessorArchitecture.None, fileExists, null, null, null, null, null, null, null, null, null, new Version("4.0"), null, log, null, true, false, null, null, false, null, WarnOrErrorOnTargetArchitectureMismatchBehavior.None, false, false);
+            ReferenceTable referenceTable = new ReferenceTable(null, false, false, false, false, new string[0], null, null, null, null, null, null, SystemProcessorArchitecture.None, fileExists, null, null, null, null, null, null, null, null, null, new Version("4.0"), null, log, null, true, false, null, null, false, null, WarnOrErrorOnTargetArchitectureMismatchBehavior.None, false, false, null);
```
cdmihai: I sure wouldn't want to construct a new ReferenceTable :)

rainersigwald: If you hadn't been out yesterday afternoon I was going to send you that set of parameters and have you guess what the heck it was for :)

cdmihai: Factory method for a new Big Bang.

PR #2193, opened 2017-06-07
Don't force Verbosity to Diag with /bl

KirillOsenkov:
It turns out that some tasks log a ridiculous amount of information as part of log task inputs. We see upwards of 5GB of strings logged as task inputs to Nuget.

To alleviate this pressure let's remove this, since the users will always have the easy option to manually add /v:diag to force logging task inputs.

For now not having this option is better since otherwise binary logger can crash people's builds.

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -2738,11 +2738,6 @@ namespace Microsoft.Build.CommandLine
             BinaryLogger logger = new BinaryLogger();
             logger.Parameters = arguments;
 
-            // If we have a binary logger, force verbosity to diagnostic.
-            // The only place where verbosity is used downstream is to determine whether to log task inputs.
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
See also https://github.com/aspnet/KoreBuild/pull/256 and https://github.com/Microsoft/msbuild/issues/2168

KirillOsenkov:
Any chance of this making it into 15.3?

nguerrera:
If this is accepted, what is the command-line syntax for getting a binary log with task inputs without logging diag to the console?

KirillOsenkov:
Excellent question! I think there's the MSBUILDLOGTASKINPUTS environment variable. Or you could just do /bl /v:diag /noconlog.

But now I'm myself not convinced we should take this PR.

nguerrera:
How about having a max task items to log option and defaulting it to something reasonable?

PR #2197, opened 2017-06-08
@(TargetPathWithTargetPlatformMoniker) outside GetTargetPath

### LINKED ISSUE #2183: Projects that override GetTargetPath have outputs ignored in referencing projects
MSBuild version 15.3.255.36604 (Bundled with the VS 2017 15.3.0 Preview 2.0 [26605.1.d15rel]).
Fails to build the following project: https://github.com/NuGet/NuGet.Jobs
The same project builds fine with the MSBuild 15.1.1012.6693 (bundled with VS2017 15.2 (26430.12)).

Steps:
1. `mkdir NuGet.Jobs`
2. `cd NuGet.Jobs`
3. `git clone https://github.com/NuGet/NuGet.Jobs.git .`
4. `git checkout dev`
5. Restore nuget packages (pick one of two):
    * `build.cmd` (tries to restore nuget packages, then build)
    * `nuget restore NuGet.Jobs.sln` (needs `nuget.exe` present)
6. `"C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\15.0\bin\msbuild.exe" .\NuGet.Jobs.sln` (msbuild paths may vary)

The first error produced:
```
Tests.Logger.Job.cs(9,7): error CS0246: The type or namespace name 'NuGet' could not be found (are you missing a using directive or an assembly reference?) [E:\dev\Nuget.Jobs\src\Tests.Logger\Tests.AzureJobTraceListener.csproj]
```

`src\Tests.Logger\Tests.AzureJobTraceListener.csproj` references the `src\NuGet.Jobs.Common\NuGet.Jobs.Common.csproj` which declares the `NuGet` namespace the compiler complains about, so there shouldn't be any error.

Further examination of the command line it uses to compile the `Tests.AzureJobTraceListener.csproj` indicates that `NuGet.Jobs.Common.dll` is not passed as a `/reference` to the `csc.exe`. I'm not sure how to dig deeper from this point.

[msbuild.log.zip](https://github.com/Microsoft/msbuild/files/1056556/msbuild.log.zip) from the build produced with the `/flp:v=diag` command line argument.

rainersigwald:
Some projects (and some NuGet packages) override GetTargetPath, and they
can't be expected to keep up with changes in the metadata now returned
by it.

This change moves the computation of the item returned from GetTargetPath
and Build into its own target (recycling
GetTargetPathWithTargetPlatformMoniker), so that it will still be
created by a project that overwrites GetTargetPath.

Fixes #2183.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1793,8 +1793,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     ============================================================
                                         GetTargetPath
 
-    This stand-alone target returns the name of the build product (i.e. EXE, DLL)
-    that would be produced if we built this project.
+    This target returns an item containing the build product (i.e. EXE, DLL)
+    that would be produced if we built this project, with some relevant
+    metadata.
     ============================================================
     -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


rainersigwald:
FYI @agr

FYI @pilchie -- this is an alternate way to approach the [change you made two years ago](http://vstfdevdiv:8080/DevDiv2/DevDiv/_versionControl/changeset/1491539) to simplify `GetTargetPathWithTargetPlatformMoniker`. I recently changed the `Build` target to return the same metadata as the `GetTargetPath` target, so projects that overrode `GetTargetPath` were ignored when `ProjectReference`d because the item never got computed.

Pilchie:
Sounds good to me.

rainersigwald:
Got confirmation from @agr that this fixes the problem.

PR #2205, opened 2017-06-09
Revert "Always use response file for winmdexp (#2177)"

rainersigwald:
This reverts commit 184da1a84e1aaa3b2e4c0198154c66fa8858e0cd.

That commit compiled and produced reasonable output but the output is not accepted by (the current version of?) `winmdexp.exe`, which expects that arguments in the response file are newline rather than space separated.

@yizhang82 @AndyGerlicher


yizhang82:
@rainersigwald yes, that's what is expected since the first version.

PR #2210, opened 2017-06-12
Use BuildID rather than rev in package version

rainersigwald:
Since I switched to specifying the version of the VSTS build to match
the real build number, the `rev` portion of the ORIGINAL_BUILDVERSION
hasn't been increasing, because `01` is always available. Use BuildID
instead, which is guaranteed to be unique across a VSTS account.


jeffkl:
Ideally the BuildID would increment for each build per definition per day.  Is there no value that does that?

rainersigwald:
No, there's not, but I don't think I understand why that's desirable. Isn't this just a disambiguator for instances where the build fails for a reason that doesn't require a code patch, and thus any semi-unique string will do?

jeffkl:
Yeah the final NuGet version ends in a published package like https://www.nuget.org/packages/Microsoft.Build.Framework/15.3.0-preview-000117-01

We already append a 6 digit git height so another 6 digit unique build number seems over the top but maybe I'm over analyzing it.

Does this change just make the package version look like: `15.3.0-preview-000117-123456` ?

rainersigwald:
Yes, exactly. I'd also rather not have another 6 or 7-digit number, but I couldn't see another build variable that would be unique between builds on the same commit (since I accidentally broke revision) which had that behavior but now doesn't.

rainersigwald:
I accidentally took this to the 15.3 branch. I wasn't planning to do that. Merged it to master and pushed vs15.3 back to the previous commit.

PR #2213, opened 2017-06-12
Late bind CopyUpToDateMarker

rainersigwald:
PR #2180 caused a performance-test scenario internal to Microsoft to
start failing. The root cause was that the projects in the test
customized `$(IntermediateOutputPath)` _very_ late, after importing
CSharp.targets (and transitively Common.targets). That meant that the
path stored in `$(CopyOutOfDateMarker)` was no longer under
`$(IntermediateOutputPath)`, and that the directory wasn't created before
the new code attempted to touch the marker file.

Ideally, projects would set `IntermediateOutputPath` _before_ importing
common.targets, but that hasn't been strictly required in all cases
before so requiring it now is an unacceptable regression.

The late customization was handled for other intermediate outputs by
defining their paths with items (expanded after all properties are
expanded) rather than properties (expanded in order).

There are other properties defined by using `$(IntermediateOutputPath)`,
`$(_GenerateBindingRedirectsIntermediateAppConfig)` and
`$(WinMDExpOutputPdb)`. This PR doesn't change those since they weren't
recently regressed.


rainersigwald:
@panopticoncentral This will impact the FUTD check implementation. I'll add some notes in https://github.com/dotnet/project-system/pull/2414.

@jcouv FYI, related to ref assemblies--when I implemented the marker code, I imitated things that were subtly broken in some customized projects. This implementation imitates the things that work.

PR #2214, opened 2017-06-12
Use newline-delimited response file for winmdexp

### LINKED ISSUE #2140: Use a response file in WinMDExp task

rainersigwald:
Winmdexp requires response files with one argument per line, and can't
handle ToolTask's default one-long-line-with-spaces response files. So
create a new CommandLineBuilderExtension that appends newlines instead
of spaces between arguments and use it in winmdexp.

Reverts 0174f995. Fixes #2140.

### DIFF
--- a/ref/net46/Microsoft.Build.Utilities.Core/Microsoft.Build.Utilities.Core.cs
+++ b/ref/net46/Microsoft.Build.Utilities.Core/Microsoft.Build.Utilities.Core.cs
@@ -93,7 +93,7 @@ namespace Microsoft.Build.Utilities
         public void AppendFileNamesIfNotNull(string[] fileNames, string delimiter) { }
         protected void AppendFileNameWithQuoting(string fileName) { }
         protected void AppendQuotedTextToBuffer(System.Text.StringBuilder buffer, string unquotedTextToAppend) { }
-        protected void AppendSpaceIfNotEmpty() { }
+        protected virtual void AppendSpaceIfNotEmpty() { }
         public void AppendSwitch(string switchName) { }
         public void AppendSwitchIfNotNull(string switchName, Microsoft.Build.Framework.ITaskItem parameter) { }
         public void AppendSwitchIfNotNull(string switchName, Microsoft.Build.Framework.ITaskItem[] parameters, string delimiter) { }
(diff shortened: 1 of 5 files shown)


ref/net46/Microsoft.Build.Utilities.Core/Microsoft.Build.Utilities.Core.cs
```diff
@@ -93,7 +93,7 @@ public partial class CommandLineBuilder
         public void AppendFileNamesIfNotNull(string[] fileNames, string delimiter) { }
         protected void AppendFileNameWithQuoting(string fileName) { }
         protected void AppendQuotedTextToBuffer(System.Text.StringBuilder buffer, string unquotedTextToAppend) { }
-        protected void AppendSpaceIfNotEmpty() { }
+        protected virtual void AppendSpaceIfNotEmpty() { }
```
rainersigwald: Hmm. It looks like this is technically a breaking change according to https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md

Thoughts on whether that's fatal?

rainersigwald:
Closing in favor of #2216.

PR #2218, opened 2017-06-14
RI 15.3

rainersigwald:


rainersigwald:
Taking this without review since it passed CI and is just normal codeflow.

PR #2222, opened 2017-06-15
Return of RAR metadata caching

rainersigwald:
My previous attempt at this in #2192 had a bad side effect: it kept handles within `AssemblyInformation` objects live in the cache past the point of RAR returning, because nothing disposed them. They should have been cleaned up eventually, if finalizers ran after GC, but that wasn't sufficient to keep an internal test that did a rebuild in VS working--it failed copying from obj to bin because a the file in bin was held in use.

This is a cleaner approach: wrap the extraction process as it was done before, and cache an object consisting only of the results.

That "object consisting only of the results" seems like it should just be `AssemblyInformation` instead of a new type. I introduced the new type to minimize churn and maximize understandability of this change in the short term.

Relates to #2015.

### DIFF
--- a/src/Tasks/AssemblyDependency/AssemblyInformation.cs
+++ b/src/Tasks/AssemblyDependency/AssemblyInformation.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Globalization;
 using System.IO;
 using System.Runtime.InteropServices;
@@ -17,6 +18,7 @@ using System.Reflection.PortableExecutable;
 using System.Reflection.Metadata;
 #endif
 using System.Collections.Generic;
+using Microsoft.Build.Tasks.AssemblyDependency;
 
 namespace Microsoft.Build.Tasks
 {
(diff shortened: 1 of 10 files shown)


src/Tasks/AssemblyDependency/AssemblyInformation.cs
```diff
@@ -192,24 +194,25 @@ public FrameworkName FrameworkNameAttribute
...
         /// </summary>
         /// <param name="path">Path to the assembly.</param>
+        /// <param name="metadataCache">Cache of pre-extracted assembly metadata.</param>
         /// <param name="dependencies">Receives the list of dependencies.</param>
         /// <param name="scatterFiles">Receives the list of associated scatter files.</param>
         /// <param name="frameworkName">Gets the assembly name.</param>
         internal static void GetAssemblyMetadata
         (
             string path,
+            ConcurrentDictionary<string, AssemblyMetadata> metadataCache,
```
cdmihai: I'd name it assemblyMetadataCache

cdmihai:
>         internal static FrameworkName GetTargetFrameworkAttribute(string path)

Shouldn't GetTargetFrameworkAttribute get cached as well?

---
Refers to: src/Tasks/AssemblyDependency/AssemblyInformation.cs:222 in d541b3f. [](commit_id = d541b3fde95571060de73d4ef91576029c729490, deletion_comment = False)

rainersigwald:
> Shouldn't GetTargetFrameworkAttribute get cached as well? 

Yes, but it doesn't seem to be called anywhere. So I didn't bother.

cdmihai:
>         internal AssemblyInformation(string sourceFile)

Are there any other usages of the AssemblyInformation constructor besides the static method you created? If yes, it might be worth beeffing up AssemblyMetadata with the new usages.

---
Refers to: src/Tasks/AssemblyDependency/AssemblyInformation.cs:65 in d541b3f. [](commit_id = d541b3fde95571060de73d4ef91576029c729490, deletion_comment = False)

rainersigwald:
No, no other uses of the AssemblyInformation constructor, so I think this is reasonably complete.

rainersigwald:

rainersigwald:
I did some benchmarks with and without this change on top of current vs15.3 (520d1aa).

Testbed is cli `2.0.0-preview2-006497` + private MSBuilds on incremental build of a project generated with `dotnet new mvc`. 

scenario | mean | stdev
---------|------|-----
|no-cache|1580.35|102.2840601|
|with-cache|1528.2|134.2231449|

That's about a 3% gain, which is nothing to sneeze at but doesn't meet the bar for 15.3 at the moment, so I'll retarget this PR to master.

rainersigwald:

rainersigwald:

rainersigwald:
@dotnet-bot test this please

PR #2225, opened 2017-06-15
Hide import events under an opt in flag

cdmihai:
The logging of import events introduced a 100ms regression in c++
scenarios. Hiding them under an escape hatch for now. The binary logger
is the only thing that sets them on.

Later on, we should add another enum value to ProjectLoadSettings to
enable API users to run this. Or figure our how to flow debugger
verbosity into the Evaluator so we can branch on it.

### DIFF
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -3840,6 +3840,8 @@ namespace Microsoft.Build.UnitTests.OM.Definition
         }
 
         [Fact]
+        [Trait("Category", "netcore-osx-failing")] // https://github.com/Microsoft/msbuild/issues/2226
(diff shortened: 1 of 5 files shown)


AndyGerlicher:
@cdmihai looks like test failures but looks good once those are fixed.

@jeffkl we noticed that a lot of the regressed time was spent in `ExpandAndLoadImportsFromUnescapedImportExpressionConditioned` and `ExpandIntoStringAndUnescape`. Not sure if this is all of it, but should be a significant chunk. And it's the creation of the events, not the handling of them so that needs to be flagged. 

@jviau are you guys using this for 15.3? With this change we will not longer be emitting those imported events except for the binary logger. We'll likely have to add a parameter to Project construction (or something similar) in order to turn that on for you.

jviau:
@AndyGerlicher we are not using the skipped import event yet. We will post-15.3.

PR #2229, opened 2017-06-20
Search for satellites next to task assemblies

### LINKED ISSUE #2203: Task satellite assemblies are not used in .NET Core msbuild
[task-satellite.zip](https://github.com/Microsoft/msbuild/files/1062207/task-satellite.zip) has the contents of the c:\temp\task-satellite in transcript below.

We expect output of "Bonjour" instead of "Hello" in the final step because the task has a satellite assembly that replaces its neutral "Hello" message with "Bonjour" for CurrentUICulture=fr.

On machine with CurrentUICulture=fr-FR
```
C:\temp\task-satellite>type test.proj
<Project>
  <UsingTask TaskName="SomeTask" AssemblyFile="bin\Debug\netcoreapp1.1\task-satellite.dll" />
  <Target Name="Repro">
    <SomeTask />
  </Target>
</Project>

C:\temp\task-satellite>dir /s /b bin\Debug\netcoreapp1.1
C:\temp\task-satellite\bin\Debug\netcoreapp1.1\fr
C:\temp\task-satellite\bin\Debug\netcoreapp1.1\task-satellite.deps.json
C:\temp\task-satellite\bin\Debug\netcoreapp1.1\task-satellite.dll
C:\temp\task-satellite\bin\Debug\netcoreapp1.1\task-satellite.pdb
C:\temp\task-satellite\bin\Debug\netcoreapp1.1\fr\task-satellite.resources.dll

C:\temp\task-satellite>c:\src\cli\artifacts\win10-x64\stage2\dotnet.exe msbuild test.proj
Microsoft (R) Build Engine version 15.3.378.6360 pour .NET Core
Copyright (C) Microsoft Corporation. Tous droits réservés.

  Hello
```

I haven't debugged, but from quick code inspection, it seems [CoreCLRAssemblyLoader](https://github.com/Microsoft/msbuild/blob/master/src/Shared/CoreCLRAssemblyLoader.cs) doesn't probe for satellites correctly.

@AndyGerlicher @cdmihai This is a blocker for CLI localization. None of the dotnet/sdk messages are showing up localized as a result.  Please triage accordingly.

rainersigwald:
Fixes #2203 by including the culture folder in the search for assembly locations, if an assembly culture is defined.

### DIFF
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -100,26 +100,36 @@ namespace Microsoft.Build.Shared
                     return assembly;
                 }
 
-                foreach (var dependencyPath in _dependencyPaths)
+                foreach (var cultureSubfolder in string.IsNullOrEmpty(assemblyName.CultureName)
+                    // If no culture is specified, attempt to load directly from
+                    // the known dependency paths.
+                    ? new[] {string.Empty}
+                    // Search for satellite assemblies in culture subdirectories
+                    // of the assembly search directories, but fall back to the
+                    // bare search directory if that fails.
+                    : new[] {assemblyName.CultureName, string.Empty})
                 {
-                    foreach (var extension in _extensions)
+                    foreach (var dependencyPath in _dependencyPaths)
                     {
-                        var candidatePath = Path.Combine(dependencyPath,
-                            assemblyName.CultureName ?? string.Empty,
-                            $"{assemblyName.Name}.{extension}");
-                        if (IsAssemblyAlreadyLoaded(candidatePath) ||
-                            !File.Exists(candidatePath))
+                        foreach (var extension in _extensions)
                         {
-                            continue;
-                        }
+                            var candidatePath = Path.Combine(dependencyPath,
+                                cultureSubfolder,
+                                $"{assemblyName.Name}.{extension}");
+                            if (IsAssemblyAlreadyLoaded(candidatePath) ||
+                                !File.Exists(candidatePath))
+                            {
+                                continue;
+                            }
 
-                        AssemblyName candidateAssemblyName = AssemblyLoadContext.GetAssemblyName(candidatePath);
-                        if (candidateAssemblyName.Version != assemblyName.Version)
-                        {
-                            continue;
-                        }
+                            AssemblyName candidateAssemblyName = AssemblyLoadContext.GetAssemblyName(candidatePath);
+                            if (candidateAssemblyName.Version != assemblyName.Version)
+                            {
+                                continue;
+                            }
 
-                        return LoadAndCache(candidatePath);
+                            return LoadAndCache(candidatePath);
+                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -104,7 +104,9 @@ private Assembly TryResolveAssembly(AssemblyLoadContext context, AssemblyName as
                 {
                     foreach (var extension in _extensions)
                     {
-                        var candidatePath = Path.Combine(dependencyPath, $"{assemblyName.Name}.{extension}");
+                        var candidatePath = Path.Combine(dependencyPath,
+                            assemblyName.CultureName ?? string.Empty,
```
nguerrera: Separate question about line after this: how can we get this far with the assembly already loaded? It looks like it would fail to reuse the already loaded assembly in that case.

nguerrera: Maybe only if the previous load failed and we don't want to bother retrying? Worth a comment if so.

rainersigwald:

rainersigwald:

rainersigwald:

(internal Jenkins error?)

src/Shared/CoreCLRAssemblyLoader.cs
cdmihai: This is based on the assumption that only satellite assemblies have a culture set on them. If the assumption breaks this might lead to task dependencies not being found, if they have a culture set on them.

cdmihai: Add a comment saying the line is for satellite assemblies.

rainersigwald:
Self, update https://github.com/dotnet/cli/issues/6114 when this is completed.

src/Shared/CoreCLRAssemblyLoader.cs
rainersigwald: Wikipedia [claims](https://en.wikipedia.org/wiki/Assembly_(CLI)) this is ok

> In general, library and process assemblies should be culture neutral; the culture should only be used for satellite assemblies.

But I have no idea how true that is. @nguerrera do you know?

I could make the subdirectory an _additional_ search path instead . . .

cdmihai: I think the bullet proof approach would be to add the main dependency path at the end of the current chain. It would avoid perf issues, and still hit for main assemblies with a culture on them.

rainersigwald: Put it as the outermost loop so hopefully it's never used.

rainersigwald: That is confusing. @anurse @tmeschter do you recall the motivation behind the `IsAssemblyAlreadyLoaded` check by any chance? I assume not since it was almost a year ago . . .

tmeschter: If we got to this point then none of the assemblies we've already loaded match the one that is being requested. If one of them did then we wouldn't have been called at all. Also, we don't want to load the same assembly twice. So we skip the already loaded one and go looking for another.

I can't recall the details about when this would happen, but there was a reason for it.

PR #2232, opened 2017-06-20
sync xlf files

cdmihai:


cdmihai:
After this PR gets in I'll merge 15.3 into master, to make sure the resources don't go out of sync

PR #2235, opened 2017-06-20
Add binding redirect for System.IO.Compression to 4.1.2.0.

KirillOsenkov:
If a code task factory uses System.IO.Compression it will pick up 4.0.0.0 from the GAC and since we load our local 4.1.2.0 we end up with two versions of the .dll loaded side-by-side, causing havoc.

### DIFF
--- a/src/MSBuild/app.config
+++ b/src/MSBuild/app.config
@@ -39,6 +39,10 @@
           <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.1.0.0" />
         </dependentAssembly>
         <dependentAssembly>
+          <assemblyIdentity name="System.IO.Compression" culture="neutral" publicKeyToken="b77a5c561934e089" />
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
I've verified that this fixes https://github.com/Microsoft/msbuild/issues/2234

src/MSBuild/app.config
```diff
@@ -39,6 +39,10 @@
           <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="15.1.0.0" />
         </dependentAssembly>
         <dependentAssembly>
+          <assemblyIdentity name="System.IO.Compression" culture="neutral" publicKeyToken="b77a5c561934e089" />
+          <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="4.1.2.0" />
```
rainersigwald: I think this should be `oldVersion="0.0.0.0-4.1.2.0"` to match the devenv.exe.config equivalent.

KirillOsenkov: Fixed.

PR #2242, opened 2017-06-23
Version to 15.5

rainersigwald:
Changing version to match VS release this is expected to ship with.


rainersigwald:
I've "unlisted" all of the 15.6* versions from our myget feed. If no one yells we can delete them in a couple of days. AFAIK no one was depending on any of those versions.

PR #2246, opened 2017-06-26
Perf improvements

cdmihai:
Found these two by looking at low hanging profiler fruits. RPS results claim that they increase a scenario by 19%, and in local tests they improve the evaluation times of a 28 project solution by about 60ms.

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -34,6 +34,7 @@ using ReservedPropertyNames = Microsoft.Build.Internal.ReservedPropertyNames;
 using Constants = Microsoft.Build.Internal.Constants;
 using EngineFileUtilities = Microsoft.Build.Internal.EngineFileUtilities;
 using Microsoft.Build.Framework;
+using Microsoft.Build.Internal;
 using Microsoft.Build.Utilities;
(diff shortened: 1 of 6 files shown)


cdmihai:

cdmihai:

cdmihai:

cdmihai:

src/Build/Evaluation/Evaluator.cs
```diff
@@ -243,6 +244,11 @@ internal class Evaluator<P, I, M, D>
 #endif
 
         /// <summary>
+        /// The search paths are machine specific and should not change during builds
+        /// </summary>
+        private static readonly EngineFileUtilities.IOCache _fallbackSearchPathsCache = new EngineFileUtilities.IOCache();
```
rainersigwald: Talked about this with @AndyGerlicher on Friday. We're a bit concerned about this cache having process lifetime. VS is long-lived and someone could conceivably install things while it's opened.

Unfortunately, I don't know how to reduce this concern.

rainersigwald: We discussed this today. The directories that are cached here are: the VS instance's MSBuild directory, and the program files MSBuild directory (just those directories, not their subdirectories). Both should require VS being closed to be created, so the process-lifetime cache should be fine.

PR #2254, opened 2017-06-29
Resource extraction from ARM assemblies

### LINKED ISSUE #2253: GenerateResource cannot extract resources from ARM binaries
This was reported via feedback as https://developercommunity.visualstudio.com/content/problem/54934/uwp-resx-strings-not-merged-into-the-pri-file-for.html and as https://connect.microsoft.com/VisualStudio/feedback/details/2345310/. Internal bug [189984](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/189984).

If a UWP app project that builds an appx package refers to a DLL compiled for ARM that embeds resources, those resources are silently ignored rather than packaged into the PRI.

This is because of a `BadImageFormatException`

```
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadFrom(String assemblyFile, Evidence securityEvidence, Byte[] hashValue, AssemblyHashAlgorithm hashAlgorithm, Boolean forIntrospection, Boolean suppressSecurityChecks, StackCrawlMark& stackMark)
   at System.Reflection.Assembly.UnsafeLoadFrom(String assemblyFile)
   at Microsoft.Build.Tasks.ProcessResourceFiles.ReadAssemblyResources(String name, String outFileOrDir) in E:\A\_work\8\s\src\Tasks\GenerateResource.cs:line 2958
```
 
Thrown [here ](https://github.com/Microsoft/msbuild/blob/f45785fb104c30080024affe855cd10ad27c6d4e/src/Tasks/GenerateResource.cs#L2958) and silently ignored [here](https://github.com/Microsoft/msbuild/blob/f45785fb104c30080024affe855cd10ad27c6d4e/src/Tasks/GenerateResource.cs#L2997-L3003).

rainersigwald:
Fixes #2253 by avoiding "real" loading the assembly, instead loading it in the reflection-only context, which is allowed for assemblies that don't work in the current runtime.

### DIFF
--- a/src/Tasks/GenerateResource.cs
+++ b/src/Tasks/GenerateResource.cs
@@ -2333,7 +2333,7 @@ namespace Microsoft.Build.Tasks
 #if FEATURE_ASSEMBLY_LOADFROM
                 // Install assembly resolution event handler.
                 _eventHandler = new ResolveEventHandler(ResolveAssembly);
-                AppDomain.CurrentDomain.AssemblyResolve += _eventHandler;
+                AppDomain.CurrentDomain.ReflectionOnlyAssemblyResolve += _eventHandler;
 #endif
 
                 for (int i = 0; i < _inFiles.Count; ++i)
@@ -2354,7 +2354,7 @@ namespace Microsoft.Build.Tasks
             {
 #if FEATURE_ASSEMBLY_LOADFROM
                 // Remove the event handler.
-                AppDomain.CurrentDomain.AssemblyResolve -= _eventHandler;
+                AppDomain.CurrentDomain.ReflectionOnlyAssemblyResolve -= _eventHandler;
                 _eventHandler = null;
 #endif
             }
@@ -2415,7 +2415,7 @@ namespace Microsoft.Build.Tasks
                     {
                         if (candidateAssemblyName.CompareTo(requestedAssemblyName) == 0)
                         {
-                            return Assembly.UnsafeLoadFrom(_assemblyFiles[i].ItemSpec);
+                            return Assembly.ReflectionOnlyLoadFrom(_assemblyFiles[i].ItemSpec);
                         }
                     }
                 }
@@ -2432,7 +2432,7 @@ namespace Microsoft.Build.Tasks
                     {
                         if (String.Compare(requestedAssemblyName.Name, candidateAssemblyName.Name, StringComparison.CurrentCultureIgnoreCase) == 0)
                         {
-                            return Assembly.UnsafeLoadFrom(_assemblyFiles[i].ItemSpec);
+                            return Assembly.ReflectionOnlyLoadFrom(itemSpec);
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/Tasks/GenerateResource.cs
```diff
@@ -2432,6 +2432,12 @@ internal Assembly ResolveAssembly(object sender, ResolveEventArgs args)
                     {
                         if (String.Compare(requestedAssemblyName.Name, candidateAssemblyName.Name, StringComparison.CurrentCultureIgnoreCase) == 0)
                         {
+                            string itemSpec = _assemblyFiles[i].ItemSpec;
+
+                            if (requestedAssemblyName.Name.Contains("System.Runtime"))
```
akoeplinger: why `Contains()`? what happens if the requestedAssemblyName is System.Runtime.InteropServices.RuntimeInformation?

src/Tasks/GenerateResource.cs
```diff
@@ -2432,6 +2432,12 @@ internal Assembly ResolveAssembly(object sender, ResolveEventArgs args)
...
+                            {
+                                itemSpec = Path.Combine(Path.GetDirectoryName(typeof(object).Assembly.Location), "System.Runtime.dll");
```
akoeplinger: this assumes that System.Runtime.dll will be in the same location as mscorlib, I'm not sure this is always the case?

src/Tasks/GenerateResource.cs
rainersigwald: Great question! It "works" with the current code but I have very little confidence in it.

@ericstj I think this is a facade-related question.  I added this code based on some [internal source](http://index/?query=%60because%20the%20parent%20does%20not%20exist&rightProject=Microsoft.VisualStudio.IDE.ToolboxControlsInstaller&file=SharedUtilities.cs&line=393) that resolved the same `"Could not load type 'System.Object' from assembly 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' because the parent does not exist."` error that I got when `ReflectionOnlyLoadFrom`ing without it.

I have two questions:
1. In what cases should I load a different assembly, rather than the referenced one?
2. When the load should be redirected, where should it be redirected to?

ericstj: /cc @weshaggard 
I suspect you're getting that error when you cross type-universes.  IIRC reflection-only-load doesn't handle loading different core assemblies than what your running on.

I think the right fix here is to *stop using ReflectionOnlyLoad*.  Switch to something like System.Reflection.Metadata.  It looks like your usage of reflection is minimal so that might even speed up your code here.  Alternatively you could try to get CLR to fix ReflectionOnlyLoad with other core assembleis but I'm guessing that even if a fix was taken it would take a long time to make it to desktop where you could depend on it.

This hack is replacing application targeted assemblies with build-time assemblies, so I cannot really sign off on it as a safe thing to do.  With that caveat I will try and answer your questions.
1. I believe you're only working around the core-assembly issue, so simply using the System.Runtime facade instead of reference may be sufficient.
2.  Perhaps you should just try to load it by name and let the build runtime use whatever rules it has to (binding-redirect/unification/tpa-list) to locate the right one for the build framework.

rainersigwald: Interesting. I'm not tied to `ReflectionOnlyLoad`; it's just closest to what was done before. Since this didn't make 15.3, I'll explore `System.Reflection.Metadata`.

rainersigwald: @ericstj 

I was taking a look at doing resource extraction via `System.Reflection.Metadata` (rather than loading the assembly, as you suggested), and I’m pretty baffled. Do you know of a good contact that I could bug to see if what I want to do is possible?

The MSDN docs seem to just be the doc-comments, with no examples or motivations, and the only use I could find was in Roslyn, which is write-only.

I can successfully extract all the names of the embedded resources

The existing reflection-based code:
1.	[Loads the assembly](https://github.com/Microsoft/msbuild/blob/00ff8ba81a8046f87a4476d5f14113e0088c5ca4/src/Tasks/GenerateResource.cs#L2958) (this failing when loading an ARM assembly is what I really want to fix)
2.	[Gets a list of resource names](https://github.com/Microsoft/msbuild/blob/00ff8ba81a8046f87a4476d5f14113e0088c5ca4/src/Tasks/GenerateResource.cs#L3013)
3.	[Extracts each resource name from its stream](https://github.com/Microsoft/msbuild/blob/00ff8ba81a8046f87a4476d5f14113e0088c5ca4/src/Tasks/GenerateResource.cs#L3050-L3085) and saves it for extraction.

I can do steps 1 and 2 fairly reasonably:
```c#
using (var stream = new FileStream(args[0], FileMode.Open, FileAccess.Read,
    FileShare.Read | FileShare.Delete))
using (var peReader = new PEReader(stream))
{
    var r = peReader.GetMetadataReader();
    foreach (var resourceHandle in r.ManifestResources)
    {
        var resource = r.GetManifestResource(resourceHandle);
        if (resource.Implementation.IsNil)
        {
            string resName = r.GetString(resource.Name);

```

But I can't see how to translate a `System.Reflection.Metadata.ManifestResource` into a `Stream` suitable for passing into a `ResourceReader`. It has fields for `Implementation` (Nil in my simple test program) and `Offset` (0 in my simple test). I don't know how to turn those into bytes.

ericstj: @tmat and @nguerrera maintain it and are the guys I usually go to for questions like this.  

I believe we had some similar problem in a resource task and @sedarg implemented a fix a while ago, but he's no longer around.

nguerrera: https://gist.github.com/nguerrera/6864d2a907cb07d869be5a2afed8d764

Disclaimer: I wrote this a LONG time ago and it was against a pre-release before the API had taken its final form. I just fixed up the compile errors and did not re-test.

We need this to be an API in the library. I'll submit an API proposal.

nguerrera: https://github.com/dotnet/corefx/issues/23372

nguerrera: There was a bug in the gist in my port s/peImageStart/resourceStart when reading resource length. @safern reported that it works for him after he applied that. I've updated the gist.

safern: Implemented it here: https://github.com/dotnet/buildtools/pull/1630/files and tested it locally and works perfectly :)

Issue #2117, opened 2017-05-22
Msbuild 15.0 cannot build Coded UI test project from normal command prompt.


acesiddhu:
With msbuild 15.0 we are seeing that coded ui test projects doesn’t build if we use a normal command prompt. This used to work with msbuild 14.0.

Here is a repro

1)	Unzip attached coded ui test zip on local machine
[CodedUITestProject2.zip](https://github.com/Microsoft/msbuild/files/1018201/CodedUITestProject2.zip)

2)	Run c:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\MSBuild.exe " CodedUITestProject2.sln"  or codeduitestproject.csproj (use normal command prompt and not developer command prompt)

Ideally the project should have been build. We tried same steps with msbuild 14 and it works fine

After investigation we saw that the csproj takes a dependency for “VisualStudioVersion”. We see that this value is null when building with msbuild 15.0. Other point to note is that .sln defines “VisualStudioVersion = 15.0.26228.9”. shouldn’t this just work?

stffabi:
We are seeing exactly the same problem. Setting VisualStudioVersion property to `15.0` in the msbuild call seems to fix the problem. It seems like for normal csproj or vbproj projects the VisualStudioVersion get's set through the `Microsoft.Common.props`. This is normally imported as first project in csproj or vbproj files by:  `<Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />`

Unfortunately our csproj which contains the codedui test code doesn't have this imported. Also creating a new coded ui projcts from the "Add Project Wizard" doesn't add this import.

Does someon know if this is a bug in the template für coded ui test projects? Importing the `Microsoft.Common.props` for coded ui test projects seems to fix the problem. But I have no clue if this breaks something else.

stffabi:
@rainersigwald do you know if it would be safe to import `Microsoft.Common.props` for coded ui projects?

acesiddhu:
@stffabi you can add the import. it wont have any issues.

rainersigwald:
This is a duplicate of #1678 but has more detailed information, so I'm going to leave both open for now.

rainersigwald:
Fixed by #2258.

mgolois:
When will this be released?

[...]

PR #2258, opened 2017-06-30
Set VisualStudioVersion property when no subtoolset is defined.

jeffkl:
In previous versions of MSBuild, there is almost always a subtoolset that adds a VisualStudioVersion property.  Since there is most likely not a subtoolset now, we need to add VisualStudioVersion if its not already a property.


jeffkl:

rainersigwald:
This will fix https://github.com/Microsoft/msbuild/issues/1678 as well as #2117.

rainersigwald:

PR #2261, opened 2017-07-03
Do not cause an error or failed build when logging an expanded condition

### LINKED ISSUE #2259: Condition-false logging can throw because it doesn't short-circuit
Given a project that has conditions that can be fully evaluated without evaluating every expression within them because of the short-circuit nature of AND and OR in the condition expression language:

```xml
  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\init.props"
          Condition="'true' == 'false' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))' != ''" />
  <Target Name="Build">
    <Message Text="Build executed"
             Condition="'true' == 'false' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))' != ''"/>
  </Target>
</Project>
```

It's ok for `$(NonExistentProperty)` to expand to the empty string, even though that would cause an error evaluating `GetDirectoryNameOfFileAbove`, because the evaluator doesn't need that side of the expression to know the correct answer. A more realistic instance of this is something like `Condition="'$(EnlistmentRoot)' != '' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(EnlistmentRoot), init.props))' != '' "`.

The condition evaluator behaves correctly, but in some verbosities loggers will attempt to produce a string like `the condition {actual text from file} was evaluated as {text substituted with values}`, which attempts to run the right-hand side of the condition, which throws, failing the execution.

This has "always" been a problem (MSBuild 4 errors when executing the target from this test file), but is exacerbated by #2142, which causes the attempted full expansion of conditions on `<Import>` elements all the time. That is mitigated by #2225, so the increased problems now only occur when using the binary logger or opting into the logging behavior.

jeffkl:
In some cases, we evaluate a condition to false (which can be short circuited) but then expand the string fully.  The full condition can contain an error that wouldn't normally be seen if it evaluates to false early.

Fixes #2259

### DIFF
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4385,6 +4385,38 @@ namespace Microsoft.Build.UnitTests.Evaluation
             }
         }
 
+        /// <summary>
+        /// Tests that an import, target, or task with a condition that contains an error but it short circuited does not fail the build.  This can happen when you have a condition like:
+        /// 'true' == 'false' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))' != ''
+        /// 
+        /// The first condition is false so the second condition is not evaluated.  But in some cases we double evaluate the condition to log it.  The second evaluation will fail because it evaluates the whole string.
+        /// 
+        /// https://github.com/Microsoft/msbuild/issues/2259
+        /// </summary>
+        [Theory]
+        [InlineData("<Target Name=\"Build\" /><Import Project=\"$(NonExistentProperty)\" Condition=\"\'true\' == \'false\' And \'$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\' != \'\'\" />")]
+        [InlineData("<Target Name=\"Build\" Condition=\"\'true\' == \'false\' And \'$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\' != \'\'\" />")]
+        [InlineData("<Target Name=\"Build\"><Message Text=\"Build executed\" Condition=\"\'true\' == \'false\' And \'$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\' != \'\'\" /></Target>")]
+        public void ConditionWithShortCircuitAndErrorDoesNotFailBuild(string projectInnerXml)
+        {
+            string content = ObjectModelHelpers.CleanupFileContents($@"
+                             <Project ToolsVersion=""msbuilddefaulttoolsversion"" xmlns='msbuildnamespace'>
+                                {projectInnerXml}
+                             </Project>");
+
+            using (var env = TestEnvironment.Create())
+            {
+                env.SetEnvironmentVariable("MSBUILDLOGIMPORTS", "1");
+                Project project = new Project(XmlReader.Create(new StringReader(content)));
+
+                MockLogger logger = new MockLogger();
+
+                bool result = project.Build(logger);
+
+                Assert.True(result);
+            }
+        }
+
 #if FEATURE_HTTP_LISTENER
         /// <summary>
         /// HTTP server code running on a separate thread that expects a connection request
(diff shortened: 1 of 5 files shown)


src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4385,6 +4385,38 @@ public void ThrownInvalidProjectExceptionProperlyHandled()
             }
         }
 
+        /// <summary>
+        /// Tests that an import, target, or task with a condition that contains an error but it short circuited does not fail the build.  This can happen when you have a condition like:
```
rainersigwald: "it short circuited" -> "is short-circuited"

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4385,6 +4385,38 @@ public void ThrownInvalidProjectExceptionProperlyHandled()
...
+        /// 'true' == 'false' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))' != ''
+        /// 
+        /// The first condition is false so the second condition is not evaluated.  But in some cases we double evaluate the condition to log it.  The second evaluation will fail because it evaluates the whole string.
+        /// 
+        /// https://github.com/Microsoft/msbuild/issues/2259
+        /// </summary>
+        [Theory]
+        [InlineData("<Target Name=\"Build\" /><Import Project=\"$(NonExistentProperty)\" Condition=\"\'true\' == \'false\' And \'$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\' != \'\'\" />")]
+        [InlineData("<Target Name=\"Build\" Condition=\"\'true\' == \'false\' And \'$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\' != \'\'\" />")]
+        [InlineData("<Target Name=\"Build\"><Message Text=\"Build executed\" Condition=\"\'true\' == \'false\' And \'$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))\' != \'\'\" /></Target>")]
```
rainersigwald: What does the new message look like for each of these?

jeffkl:
```
Target "Build" skipped, due to false condition; ('true' == 'false' And '$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))' != '') was evaluated as ('true' == 'false' And '[MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props)' != '').
```

```
Task "Message" skipped, due to false condition; ('true' == 'false' And '$([MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props))' != '') was evaluated as ('true' == 'false' And '[MSBuild]::GetDirectoryNameOfFileAbove($(NonExistentProperty), init.props)' != '').
```

rainersigwald: Hmm. That's way better (thanks to not crashing) but I wonder if we could do even more. Would love to see it expand the arguments to the property function even if it can't to the function itself, so something like

`[MSBuild]::GetDirectoryNameOfFileAbove(, init.props)`

Think that's doable?

jeffkl: Its possible for property functions, I've done this in bec1635

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4385,6 +4385,38 @@ public void ThrownInvalidProjectExceptionProperlyHandled()
...
+        public void ConditionWithShortCircuitAndErrorDoesNotFailBuild(string projectInnerXml)
+        {
+            string content = ObjectModelHelpers.CleanupFileContents($@"
```
cdmihai: There's also a `Cleanup` extension method on strings.

PR #2263, opened 2017-07-05
Should not opt in the SHA-2 workaround for 3rd party crypto providers

Jinhuafei:
The code modified in this PR is used by clickonce to sign the manifest file. In order to support signing with SHA-2 cert, a workaround was added. However, this workaround should be opted out for 3rd party crypto providers. Otherwise, it causes verification failure(unknown publisher) when the manifest file is signed by a SHA-2 EV cert.


rainersigwald:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/462908 was approved at shiproom.

PR #2288, opened 2017-07-12
Avoid params char[] allocation

davkean:
This method got inlined a few methods, so it's hard to see, but I think this was this array creation, or probably also including an inlined: https://github.com/Microsoft/msbuild/pull/2284.

image


AndyGerlicher:

PR #2292, opened 2017-07-13
Avoid allocation on enumerating ReadOnlyEmptyList

davkean:
Enumerable.Empty<T>() returns a T[] that has an allocation free enumerator for empty.

This was allocating about 1.1 MB opening ProjectSystem.sln.

image

### DIFF
--- a/src/Shared/ReadOnlyEmptyList.cs
+++ b/src/Shared/ReadOnlyEmptyList.cs
@@ -8,6 +8,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 
 using Microsoft.Build.Shared;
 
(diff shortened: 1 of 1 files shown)


davkean:
I avoided a bigger change here - which is just to remove the concept of ReadOnlyEmptyList entirely, it can be just replaced with Array.Empty<T>.

rainersigwald:
On second thought, let's try to get rid of it entirely. I was poking around in that direction after some of your earlier bugs.

I'll try that and if I don't hit any roadblocks we can have the joy of deleting code.

rainersigwald:
Turned out to be pretty easy, so I think I'd prefer #2300 to this unless someone objects.

AndyGerlicher:
Sounds good, let's take #2300. Thanks @davkean for finding this!

PR #2294, opened 2017-07-13
Replace empty array creation with Array.Empty<T>

davkean:
Replaced all usage across the tree of empty array creation with Array.Empty<T> to remove unneeded allocations. Avoided tests to reduce the diff and because of diminishing returns.


davkean:
I looked through the tree to see if you are were relying on reference identity anywhere on strings (such as `lock (aStringField)`) but I could not see anything.

AndyGerlicher:

PR #2300, opened 2017-07-13
Remove ReadOnlyEmptyList in favor of Array.Empty

rainersigwald:
Both of these provide an empty immutable collection, but Array.Empty is
a framework type so it gets better sharing and allows deleting a class
from MSBuild, in addition to avoiding the allocation pointed out in
PR #2292.

Replaces all uses of the type, replacing direct uses of its enumerator
with `Enumerable.Empty<T>`. Since `ReadOnlyEmptyDictionary` is compiled
into taskhost and must work in 3.5, added conditional compilation to
return a distinct `new T[]` where necessary there.


rainersigwald:
Thanks for the idea, @davkean.

src/Build/Evaluation/EvaluatorMetadataTable.cs
```diff
@@ -43,18 +43,7 @@ public EvaluatorMetadataTable(string implicitItemType)
...
-            {
-                if (_metadata == null)
-                {
-                    return ReadOnlyEmptyList<EvaluatorMetadata>.Instance;
-                }
-
-                return _metadata.Values;
-            }
-        }
+        internal IEnumerable<EvaluatorMetadata> Entries => _metadata?.Values ?? (IEnumerable<EvaluatorMetadata>) Array.Empty<EvaluatorMetadata>();
```
cdmihai: You can use Enumerable.Empty to avoid the cast

rainersigwald: I sure can! I thought I tried something like that and this was required but I was Doing It Wrong.

davkean:
LGTM, deleting code is always a bonus.

PR #2301, opened 2017-07-13
When Copy fails to copy a file, report which process is locking the f…

KirillOsenkov:
…ile.

Adding a LockCheck utility that uses the Windows Restart Manager (only for Full/Desktop framework) to output which process is holding the file lock if a copy operation fails.

### DIFF
--- a/src/Shared/UnitTests/MockEngine.cs
+++ b/src/Shared/UnitTests/MockEngine.cs
@@ -37,7 +37,6 @@ namespace Microsoft.Build.UnitTests
         private int _warnings = 0;
         private int _errors = 0;
         private StringBuilder _log = new StringBuilder();
-        private string _upperLog = null;
         private ProjectCollection _projectCollection = new ProjectCollection();
         private bool _logToConsole = false;
         private MockLogger _mockLogger = null;
@@ -408,26 +407,21 @@ namespace Microsoft.Build.UnitTests
         /// First check if the string is in the log string. If not
         /// than make sure it is also check the MockLogger
         /// </summary>
-        /// <param name="contains"></param>
         internal void AssertLogContains(string contains)
         {
-            if (_upperLog == null)
-            {
-                _upperLog = _log.ToString().ToUpperInvariant();
-            }
-
             // If we do not contain this string than pass it to
             // MockLogger. Since MockLogger is also registered as
             // a logger it may have this string.
-            if (!_upperLog.Contains
-                (
-                    contains.ToUpperInvariant()
-                )
-              )
+            var logText = _log.ToString();
+            if (logText.IndexOf(contains, StringComparison.OrdinalIgnoreCase) == -1)
             {
                 if (_output == null)
                 {
-                    Console.WriteLine(_log.ToString());
+                    Console.WriteLine(logText);
+                }
+                else
+                {
+                    _output.WriteLine(logText);
                 }
 
                 _mockLogger.AssertLogContains(contains);
@@ -439,31 +433,25 @@ namespace Microsoft.Build.UnitTests
         /// First check if the string is in the log string. If not
         /// than make sure it is also not in the MockLogger
         /// </summary>
-        /// <param name="contains"></param>
         internal void AssertLogDoesntContain(string contains)
         {
+            var logText = _log.ToString();
+            
             if (_output == null)
             {
-                Console.WriteLine(_log);
+                Console.WriteLine(logText);
             }
-
-            if (_upperLog == null)
+            else
             {
-                _upperLog = _log.ToString().ToUpperInvariant();
+                _output.WriteLine(logText);
             }
 
-            Assert.False(_upperLog.Contains
-                (
-                    contains.ToUpperInvariant()
-                ));
+            Assert.Equal(-1, logText.IndexOf(contains, StringComparison.OrdinalIgnoreCase));
 
             // If we do not contain this string than pass it to
             // MockLogger. Since MockLogger is also registered as
             // a logger it may have this string.
-            _mockLogger.AssertLogDoesntContain
-            (
-                contains
-            );
+            _mockLogger.AssertLogDoesntContain(contains);
         }
 
         /// <summary>
--- a/src/Tasks/Copy.cs
+++ b/src/Tasks/Copy.cs
@@ -355,7 +355,7 @@ namespace Microsoft.Build.Tasks
             {
                 FileUtilities.DeleteNoThrow(destinationFileState.Name);
             }
-            
+
             linkCreated = createLink(sourceFileState.Name, destinationFileState.Name, errorMessage);
 
             if (!linkCreated)
@@ -680,7 +680,9 @@ namespace Microsoft.Build.Tasks
                     if (retries < Retries)
                     {
                         retries++;
-                        Log.LogWarningWithCodeFromResources("Copy.Retrying", sourceFileState.Name, destinationFileState.Name, retries, RetryDelayMilliseconds, e.Message);
+                        Log.LogWarningWithCodeFromResources("Copy.Retrying", sourceFileState.Name,
+                            destinationFileState.Name, retries, RetryDelayMilliseconds, e.Message,
+                            GetLockedFileMessage(destinationFileState.Name));
 
                         // if we have to retry for some reason, wipe the state -- it may not be correct anymore. 
                         destinationFileState.Reset();
@@ -691,7 +693,8 @@ namespace Microsoft.Build.Tasks
                     else if (Retries > 0)
                     {
                         // Exception message is logged in caller
-                        Log.LogErrorWithCodeFromResources("Copy.ExceededRetries", sourceFileState.Name, destinationFileState.Name, Retries);
+                        Log.LogErrorWithCodeFromResources("Copy.ExceededRetries", sourceFileState.Name,
+                            destinationFileState.Name, Retries, GetLockedFileMessage(destinationFileState.Name));
                         throw;
                     }
                     else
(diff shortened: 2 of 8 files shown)


AndyGerlicher:
I like this!!

@cdmihai is looking at how we can take this from the legal aspect (just to make sure everything is good). Looks like a few tests are asserting the number of errors from Copy that will need to change. It should assert that the write process has the lock too :).

KirillOsenkov:
Here's a sample output:
```
ERROR : File "C:\Users\Kirill\AppData\Local\Temp\tmp1E1D.tmp" is locked by: "TestDriven.Net 4.1.3526.0 (18400)"
```

Note that the error doesn't have an error code, the range of error codes there was exhausted and really this should probably be a message, but the benefit of this being an error is that this is right next to the errors (the first thing people look at during investigations).

KirillOsenkov:
Updated the unit-test.

KirillOsenkov:
👍

KirillOsenkov:
I've fixed the tests, but please review because I had made some fixes in the test infrastructure. Turns out it was caching the log text the first time AssertLogContains was called. This revealed hidden failures in two unrelated tests and I've filed an issue to fix those.

KirillOsenkov:
The more I think about it the more I don't like the design. There were two errors already, and that kind of annoyed me all these years, and now there's the third (which doesn't have the error code to boot). It feels like there should only be one error instead of three.

Maybe here, instead of throwing (which would result in the second error logged):
http://source.dot.net/#Microsoft.Build.Tasks.Core/Copy.cs,695
...just return false and make sure that we augment that error with the lock information. It will be finicky to get right but maybe worth it for a better end-user experience?

AndyGerlicher:
@KirillOsenkov agree with what you said, would be far better to extend the existing error message so we don't have to produce another. I do think this is a really great debugging feature though, excited to see it go in!

KirillOsenkov:
Thanks for helping with this!!

AndyGerlicher:
I added a commit that moved the file locked message to the existing error and warning. How's that look? Sample output trying to build MSBuild repo with ildasm open:
```
  Microsoft.Build.CommandLine.UnitTests -> D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Microsoft.Build.CommandLine.UnitTests.dll
D:\src\msbuild.fork3\bin\Bootstrap\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(4048,5): warning MSB3026: Could not copy "D:\src\msbuild.fork3\bin\ob
j\Debug\x86\Windows_NT\Microsoft.Build.Tasks\Microsoft.Build.Tasks.Core.dll" to "D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Microsoft.Build.Tasks.Core.
dll". Beginning retry 1 in 1000ms. The process cannot access the file 'D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Microsoft.Build.Tasks.Core.dll' becau
se it is being used by another process. The file is locked by: "Microsoft .NET Framework IL disassembler (6712)" [D:\src\msbuild.fork3\src\Tasks\Microsoft.Build.
Tasks.csproj]

... warning is repeated for tries 2 - 9 ...

D:\src\msbuild.fork3\bin\Bootstrap\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(4048,5): error MSB3027: Could not copy "D:\src\msbuild.fork3\bin\obj\
Debug\x86\Windows_NT\Microsoft.Build.Tasks\Microsoft.Build.Tasks.Core.dll" to "D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Microsoft.Build.Tasks.Core.dl
l". Exceeded retry count of 10. Failed. The file is locked by: "Microsoft .NET Framework IL disassembler (6712)" [D:\src\msbuild.fork3\src\Tasks\Microsoft.Build.
Tasks.csproj]
D:\src\msbuild.fork3\bin\Bootstrap\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(4048,5): error MSB3021: Unable to copy file "D:\src\msbuild.fork3\bin
\obj\Debug\x86\Windows_NT\Microsoft.Build.Tasks\Microsoft.Build.Tasks.Core.dll" to "D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Microsoft.Build.Tasks.Co
re.dll". The process cannot access the file 'D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Microsoft.Build.Tasks.Core.dll' because it is being used by ano
ther process. [D:\src\msbuild.fork3\src\Tasks\Microsoft.Build.Tasks.csproj]
  Dependency -> D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\Dependency.dll
  TaskUsageLogger -> D:\src\msbuild.fork3\bin\Samples\Debug\TaskUsageLogger\TaskUsageLogger.dll
  XmlFileLogger -> D:\src\msbuild.fork3\bin\Samples\Debug\XmlFileLogger\XmlFileLogger.dll
  TaskWithDependency -> D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\TaskWithDependency.dll
  PortableTask -> D:\src\msbuild.fork3\bin\Debug\x86\Windows_NT\Output\PortableTask.dll
```

src/Tasks/Copy.cs
```diff
@@ -725,19 +731,28 @@ private bool DoCopyWithRetries(FileState sourceFileState, FileState destinationF
...
             {
-                Log.LogErrorFromResources("Copy.FileLocked", destinationFile, processes);
+                var processes = LockCheck.GetProcessesLockingFile(file);
+                message = !string.IsNullOrEmpty(processes)
+                    ? ResourceUtilities.FormatResourceString("Copy.FileLocked", file, processes)
+                    : String.Empty;
+            }
+            catch (Exception)
+            {
+                // Never throw if we can't get an locking file.
```
KirillOsenkov: grammar: "an locking"

src/Tasks/Resources/Strings.resx
```diff
@@ -287,7 +287,7 @@
     <comment>{StrBegin="MSB3030: "} LOCALIZATION: {0} is a number.</comment>
   </data>
   <data name="Copy.FileLocked">
-    <value>File "{0}" is locked by: "{1}"</value>
+    <value>The file is locked by: "{1}"</value>
```
KirillOsenkov: Do you want to change this to {0}?

KirillOsenkov: Actually if you're still passing two arguments then it has to stay {1} (and {0} argument is never used)

AndyGerlicher:
Just a fun fact for the next person that asks why it doesn't contain the exe name of the process in the the warning/error message... The API that is called returns an [RM_PROCESS_INFO structure](https://msdn.microsoft.com/en-us/library/windows/desktop/aa373674(v=vs.85).aspx) and that does not contain the name of the exe, only the "user-friendly" name for the application.

KirillOsenkov:
We could retrieve the .exe name ourselves since we have the process Id (Process.GetProcessById()) but there may be a race condition when the process dies right before we query for it, so I think it's not worth the trouble. They have the Id so they can look it up themselves in the Task Manager ;)

AndyGerlicher:
I'm good with this as is. @KirillOsenkov could you rebase this please? Ideally a commit for the original intent, the test changes, and @cdmihai third party license commit. Or I can but I don't think I can force push your branch so I'd have to start a new PR.

KirillOsenkov:
Done, I hope my git skills didn't fail me...

KirillOsenkov:
FYI @cklutz we're adding the LockCheck capability to MSBuild, hope you like it.

cdmihai:

AndyGerlicher:
Yes @KirillOsenkov your git skilled failed you :)

Rainer thinks I can force push and I have that done. I'll do that.

src/Shared/UnitTests/MockEngine.cs
```diff
@@ -439,31 +433,25 @@ internal void AssertLogContains(string contains)
...
+                    _output.WriteLine(logText);
+                }
 
-            if (_upperLog == null)
-            {
-                _upperLog = _log.ToString().ToUpperInvariant();
+                // If we do not contain this string than pass it to
+                // MockLogger. Since MockLogger is also registered as
+                // a logger it may have this string.
+                _mockLogger.AssertLogDoesntContain(contains);
```
rainersigwald: Shouldn't there be an assert in here somewhere? If the log string doesn't contain the message, that part passes and it should check _mockLogger. If it does contain the message, it should print and fail without bothering.  Right?

src/Tasks/Resources/Strings.resx
```diff
@@ -267,11 +267,11 @@
     <comment>LOCALIZATION: {0} and {1} are paths.</comment>      
   </data>
   <data name="Copy.Retrying">
-    <value>MSB3026: Could not copy "{0}" to "{1}". Beginning retry {2} in {3}ms. {4}</value>
+    <value>MSB3026: Could not copy "{0}" to "{1}". Beginning retry {2} in {3}ms. {4} {5}</value>
     <comment>{StrBegin="MSB3026: "} LOCALIZATION: {0} and {1} are paths. {2} and {3} are numbers. {4} is an optional localized message.</comment>
```
rainersigwald: Should we update the comment to say that {5} is either blank or an instantiation of Copy.FileLocked?

src/Shared/UnitTests/MockEngine.cs
KirillOsenkov: Yeah this whole part was very fishy to me. Shall we file a separate bug to review MockEngine logic? It kind of gets out of scope of this PR :)

src/Tasks/Resources/Strings.resx
KirillOsenkov: I've fixed this, but I might have messed up Git again. Please double-check :)

KirillOsenkov:
Ouch I've screwed up in a terrible way. Now there's 8 commits :( Someone smart please fix my mess! I'm sorry!

KirillOsenkov:
What I did was git pull to bring the server changes back locally, and it created a merge commit. How do I bring a remote branch back locally after it's been force-pushed to?

PR #2304, opened 2017-07-14
Add target framework metadata to TargetPathWithTargetPlatformMoniker

dsplaisted:
...and check it in ImplicitlyExpandDesignTimeFacades

Fixes dotnet/sdk#1393.  What is happening with that bug is that the System.Runtime facades are not being added in the design-time build when you have a reference from a .NET Framework project to a .NET Standard 1.x project that doesn't have any output on disk yet.  This results in failed intellisense, which persists even after you build the solution and/or restart VS, because the design-time build is cached.

An alternative fix to this would be to set the `TargetPlatformIdentifier` of .NET Standard projects to `Portable`.  Then the existing logic that handles references to PCLs would kick in.  I would like to change the `TargetPlatformIdentifier` of .NET Standard projects to `Portable` in the future, as I think it's more correct than the current value of `Windows`.  However, the change in this PR should be less risky so I think it's the right way to fix the bug for now.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1834,6 +1834,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <TargetPathWithTargetPlatformMoniker Include="$(TargetPath)">
         <TargetPlatformMoniker>$(TargetPlatformMoniker)</TargetPlatformMoniker>
         <TargetPlatformIdentifier>$(TargetPlatformIdentifier)</TargetPlatformIdentifier>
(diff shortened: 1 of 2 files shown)


mhutch:
This approach looks good. Thanks!

DavidKarlas:
I'm seeing exception with this change at https://github.com/Microsoft/msbuild/blob/3986423/src/Build/Evaluation/Expander.cs#L3152 because `1` as parameter in `$(TargetFrameworkVersion.Substring(1))` is considered as `string` but should be `int` hence exception... I'm probably using a bit older MSBuild and running on Mono, maybe this is not issue on newer MSBuild...

PR #2312, opened 2017-07-17
Introduce Microsoft.Build.UnitTests.Shared class library

jeffkl:
Move RunnerUtilities for now and we can move more classes later.

### DIFF
--- a/src/Shared/UnitTests/FileUtilities_Tests.cs
+++ b/src/Shared/UnitTests/FileUtilities_Tests.cs
@@ -724,7 +724,7 @@ namespace Microsoft.Build.UnitTests
 
             try
             {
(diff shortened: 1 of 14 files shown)


src/Shared/UnitTests/FileUtilities_Tests.cs
```diff
@@ -724,7 +724,7 @@ public void GenerateTempFileNameWithExtension()
 
             try
             {
-                path = Shared.FileUtilities.GetTemporaryFile(".bat");
+                path = FileUtilities.GetTemporaryFile(".bat");
```
rainersigwald: So the idea is that with more changes like this, we can remove a bunch of the shared source files from compiling in every test project?

jeffkl:

PR #2324, opened 2017-07-19
Support newlines in property functions

### LINKED ISSUE #2323: Support newlines in property functions
I'd like to be able to place chained property functions onto new lines like:

```xml
<PropertyA>
    $(SomeProperty
    .Substring(0, 3)
    .Trim())
</PropertyA>
```

Right now the evaluator will thrown an exception because the newlines are not valid.  If we trimmed the parsed method name, the newlines and whitespace would be discarded and these chained property functions would look nicer in projects.

jeffkl:
Closes #2323

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -31,6 +31,7 @@ using ProjectItemInstanceFactory = Microsoft.Build.Execution.ProjectItemInstance
 using Xunit;
 using Microsoft.Build.BackEnd;
 using Microsoft.Build.Engine.UnitTests;
+using Shouldly;
 
 namespace Microsoft.Build.UnitTests.Evaluation
 {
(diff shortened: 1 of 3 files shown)


src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -3518,5 +3519,23 @@ public void PropertyFunctionEnsureTrailingSlash()
 
             Assert.Equal(path + Path.DirectorySeparatorChar, result);
         }
+
+        [Fact]
+        public void PropertyFunctionWithNewLines()
+        {
+            const string propertyFunction = @"$(SomeProperty
+.Substring(0, 10)
+.Substring(0, 5))";
```
rainersigwald: Also add some whitespace to the newlines for good measure?

src/Build/Evaluation/Expander.cs
```diff
@@ -3506,7 +3506,7 @@ private static void ConstructFunction(IElementLocation elementLocation, string e
                             functionArguments = ExtractFunctionArguments(elementLocation, expressionFunction, argumentsContent);
                         }
 
-                        remainder = expressionFunction.Substring(argumentsEndIndex + 1);
+                        remainder = expressionFunction.Substring(argumentsEndIndex + 1).Trim();
```
rainersigwald: I'm a bit concerned about allocations here but this should only cost a new `string` in the case where there _is_ whitespace so that should be ok.

src/Build.UnitTests/Evaluation/Expander_Tests.cs
rainersigwald: Does this test hit all 3 modified lines? If not, can you add cases until it does?

jeffkl: Yes the unit test covers all the changed lines.  The first changed line trims the first property/method parsed and the second and third trims the remainder for methods with and without arguments respectively.

src/Build/Evaluation/Expander.cs
jeffkl: Andy and I confirmed that `.Trim()` will not allocate if there is no trailing or ending whitespace.  So users won't pay a penalty unless they start adding newlines.

Issue #2314, opened 2017-07-18
Update and Remove should optimize on identity matches


cdmihai:
`Update` item operations provide a way to select items and set metadata on them.

When another item is used as selection (e.g. `<Foo Update="@(Bar) Metadata="Value">`), MSBuild compares all items from the operation item type (Foo) with all items from the referenced item type (Bar) and updates the intersection. This results in N^2 runtime.

However, when the operation item type is the same as the referenced item type (e.g. `<Foo Update="@(Foo) Metadata="Value">`) MSBuild could optimize and iterate over the items only once.

Suggested by @dsplaisted who found this is slow when there are many items in a self referencing update: https://github.com/Microsoft/msbuild/issues/2238#issuecomment-315937074

Optimizations:
- [x] self referencing Updates can short circuit matching altogether
- [ ] extend the optimization in Update to all item operations that use matching (for now only Remove)
- [ ] add a lookup dictionary to ItemExpressionFragment such that operations that need to do matching hit the lookup instead of looping through all the ItemExpressionFragment items

dsplaisted:
I was thinking that a more general solution would be to create a dictionary lookup if the ReferencedItem count of an ItemExpressionFragment was high enough.  That way you wouldn't get n^2 runtime in situations where you have two different lists with a lot of items, such as `<A Update="@(B)" />`.

dsplaisted:
Re-opening per comment: https://github.com/microsoft/msbuild/pull/5350#pullrequestreview-410513306

PR #2326, opened 2017-07-19
Self referencing Update operations skip matching

cdmihai:
Perf optimization: If the Update operation references itself (e.g. `<I Update="@(I)"/>`) then all items are updated and matching is not necessary.

Implements the simpler, more constrained optimization from #2314

### DIFF
--- a/src/Build/Evaluation/LazyItemEvaluator.UpdateOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.UpdateOperation.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using Microsoft.Build.Construction;
 using System.Collections.Immutable;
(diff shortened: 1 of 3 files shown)


src/Build/Evaluation/LazyItemEvaluator.UpdateOperation.cs
```diff
@@ -46,6 +63,28 @@ public override void Apply(ImmutableList<ItemData>.Builder listBuilder, Immutabl
...
             }
+
+            private static bool ItemSpecOnlyReferencesOneItemType(ItemSpec<P, I> itemSpec, string itemType)
+            {
+                if (itemSpec.Fragments.Count() != 1)
+                {
+                    return false;
+                }
+
+                var itemExpressionFragment = itemSpec.Fragments.First() as ItemExpressionFragment<P, I>;
```
rainersigwald: `.Single()`?

cdmihai:
@dsplaisted did some testing. From internal mail: 

> I tested Mihai’s fix on the project from ASP.NET that has about 10,000 items, and the fix reduced the build time from 12 minutes to 15 seconds.

davkean:
Just as a heads up, I suspect the linked metadata change (and hence this should fix it) resulted in a ~500% jump in the number of hangs in 15.3 when compared to 15.2. Please keep me abreast if you make any changes around this.

PR #2330, opened 2017-07-20
Remove unneeded allocations in ItemSpec.MatchesItem 

davkean:
This was allocating a capture class, Func and boxing the enumerator of Fragments:

image


davkean:
Most of the allocations in this method is coming from ComparePathsNoThrow (200 MB) - are we fixing that? @dsplaisted @cdmihai

davkean:
@AndyGerlicher @rainersigwald Can we merge this in?

PR #2331, opened 2017-07-20
Copy task: Use file sizes too, for skip checks, on !windows

radical:


AndyGerlicher:
I don't think this meets the bar for 15.3 this late. Changing the base to master to target next VS release.

PR #2333, opened 2017-07-20
Fix mono/mac build

radical:


AndyGerlicher:
Nice! Could we get CI enabled now?

radical:
@AndyGerlicher We could enable for build without tests. But tests are failing because `Shouldly`'s `net46` assembly has `kernel32.dll` pinvokes. It works with netcore/mac because their `netstandard` assembly doesn't have any such pinvokes!

PR #2339, opened 2017-07-21
Remove unneeded substring

davkean:
This was showing up as ~2% of all allocations in this trace: https://github.com/dotnet/sdk/pull/1432#issuecomment-316852561.

image

To make sure this wasn't causing additional overhead (as there's additional argument checking in the StringBuilder.Append(string, int, int) overload, I did a MicroBench mark:

``` C#
[MemoryDiagnoser]
public class SubString
{
    public string Field = "This is a real long long argument";

    [Benchmark]
    public void StringBuilder_Append()
    {
        for (int i = 0; i < 10000; i++)
        {
            StringBuilder builder = new StringBuilder();

            for (int j = 0; j < 5; j++)
            {
                builder.Append(Field, 20, 11);
            }
        }
    }

    [Benchmark]
    public void StringBuilder_AppendViaSubString()
    {
        for (int i = 0; i < 10000; i++)
        {
            StringBuilder builder = new StringBuilder();

            for (int j = 0; j < 5; j++)
            {
                builder.Append(Field.Substring(20, 11));
            }
        }
    }
}
```
Method |     Mean |     Error |    StdDev |     Gen 0 | Allocated |
--------------------------------- |---------:|----------:|----------:|----------:|----------:|
StringBuilder_Append | 1.325 ms | 0.0060 ms | 0.0053 ms |  589.8438 |   2.37 MB |
StringBuilder_AppendViaSubString | 1.896 ms | 0.0115 ms | 0.0108 ms | 1019.5313 |   4.08 MB |

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -632,7 +632,7 @@ namespace Microsoft.Build.Evaluation
                                     ProjectErrorUtilities.ThrowInvalidProject(elementLocation, "InvalidFunctionPropertyExpression", expressionFunction, AssemblyResources.GetString("InvalidFunctionPropertyExpressionDetailMismatchedParenthesis"));
                                 }
 
-                                argumentBuilder.Append(argumentsString.Substring(nestedPropertyStart, (n - nestedPropertyStart) + 1));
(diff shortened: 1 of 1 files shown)


davkean:
Also did a quick pass and didn't see anymore easy ones like this.

PR #2354, opened 2017-07-26
Update dependencies

rainersigwald:
This makes a couple of things easier for SDK conversion.

### DIFF
--- a/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
+++ b/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
@@ -2545,6 +2545,7 @@ namespace Microsoft.VisualStudio.Setup.Configuration
     {
         Complete = (uint)4294967295,
         Local = (uint)1,
+        NoErrors = (uint)8,
         None = (uint)0,
         NoRebootRequired = (uint)4,
         Registered = (uint)2,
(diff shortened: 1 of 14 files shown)


ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
```diff
@@ -2545,6 +2545,7 @@ public enum InstanceState : uint
     {
         Complete = (uint)4294967295,
         Local = (uint)1,
+        NoErrors = (uint)8,
```
jeffkl: Any idea why this is updated?

rainersigwald: `VisualStudio.Microsoft.VisualStudio.Setup.Configuration.Interop` works by embedding interop types into our assemblies. This type is one of theirs.

I wish it wasn't public in our assembly but I don't know of a way to avoid it.

jeffkl: If we ship the DLL, we might not need to embed the interop types...

rainersigwald: But we'd cause an RPS regression for having another DLL :(

PR #2361, opened 2017-07-28
Remove ObjectModelHelpers.AssertThrows

rainersigwald:
This was redundant with xunit's `Assert.Throws` so this reduces
unnecessary complexity.

Primarily done through replacing

```
ObjectModelHelpers\.AssertThrows\(typeof\(([^)]*)\), delegate { ([^;]*);\s*}\)
```

with

```
Assert.Throws<$1>( () => $2 )
```

### DIFF
--- a/src/Build.UnitTests/Collections/OMcollections_tests.cs
+++ b/src/Build.UnitTests/Collections/OMcollections_tests.cs
@@ -117,7 +117,10 @@ namespace Microsoft.Build.UnitTests.OM.Collections
             using (IEnumerator<ProjectItemInstance> enumerator = items.GetEnumerator())
             {
                 enumerator.MoveNext().ShouldBeFalse();
-                Should.Throw<InvalidOperationException>(() => _ = ((IEnumerator) enumerator).Current );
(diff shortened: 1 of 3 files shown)


src/Build.UnitTests/Collections/OMcollections_tests.cs
```diff
@@ -115,9 +115,9 @@ public void BasicItemDictionary()
             // Enumeration of empty collection
             using (IEnumerator<ProjectItemInstance> enumerator = items.GetEnumerator())
             {
-                Assert.Equal(false, enumerator.MoveNext());
-                ObjectModelHelpers.AssertThrows(typeof(InvalidOperationException), delegate { object o = ((IEnumerator)enumerator).Current; });
-                Assert.Equal(null, enumerator.Current);
+                Assert.False(enumerator.MoveNext());
+                Assert.Throws<InvalidOperationException>(() => ((IEnumerator) enumerator).Current);
```
jeffkl: `Should.Throw<T>()` instead?

rainersigwald: Sure.

PR #2363, opened 2017-07-28
Display an error if an ITaskFactory does not set the TaskType property.

### LINKED ISSUE #1508: Do not crash when a TaskFactory returns a null TaskType
When inspecting the MSBuild error reported in [this issue](https://youtrack.jetbrains.com/issue/RSRP-462378), the most exhausting part was to identify which value in `TaskExecutionHost.FindTaskInRegistry` was actually `null`. 

It was `TaskFactory.TaskType`, and the `.FullName`-access threw the `NullReferenceException` (see  [TaskExecutionHost.cs, line 987](https://github.com/Microsoft/msbuild/blob/cb8c727a898e4439a3cd18f329504e436b0b7c00/src/XMakeBuildEngine/BackEnd/TaskExecutionHost/TaskExecutionHost.cs#L987)), which crashed MSBuild with _"error MSB1025: An internal failure occurred while running MSBuild"_.

Please add a null-check before the `.FullName`-access to improve debuggging when a `TaskFactory` behaves wrong.

jeffkl:
If TaskType is null it can cause a NullReferenceException in our code.  This change ensures that ITaskFactory implementations return a value for this property.

Fixes #1508

### DIFF
--- a/src/Build/Instance/TaskRegistry.cs
+++ b/src/Build/Instance/TaskRegistry.cs
@@ -1411,6 +1411,12 @@ namespace Microsoft.Build.Execution
                                                 RegisteredName);
                                         }
                                     }
+
+                                    // Throw an error if the ITaskFactory did not set the TaskType property.  If the property is null, it can cause NullReferenceExceptions in our code
+                                    if (factory.TaskType == null)
+                                    {
+                                        ProjectErrorUtilities.ThrowInvalidProject(elementLocation, "TaskFactoryTaskTypeIsNotSet", factory.FactoryName);
+                                    }
                                 }
                                 finally
                                 {
@@ -1424,6 +1430,10 @@ namespace Microsoft.Build.Execution
                                     _taskFactoryWrapperInstance = null;
                                     return false;
                                 }
+                            }
+                            catch(InvalidProjectFileException)
+                            {
+                                throw;
                             }
                             catch (InvalidCastException e)
                             {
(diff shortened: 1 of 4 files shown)


src/Build/Instance/TaskRegistry.cs
```diff
@@ -1425,6 +1431,10 @@ private bool GetTaskFactory(TargetLoggingContext targetLoggingContext, ElementLo
                                     return false;
                                 }
                             }
+                            catch(InvalidProjectFileException)
+                            {
+                                throw;
```
rainersigwald: ?

jeffkl: The new `ProjectErrorUtilities.ThrowInvalidProject()` is in a `try...catch` which wraps the exception again.  This new `catch` is to allow the original exception to be thrown.

jeffkl:

src/Build/Instance/TaskRegistry.cs
rainersigwald: Comment, please :)

I'm also not sure I think it's worth the unique error code for this if the message is good. But it's already done so go with it.

jeffkl: Okay I've removed the unique error code and thrown an `InvalidOperationException` so the overall error contains the message

```
nulltasktype.proj(10,5): error MSB4175: The task factory "Microsoft.Build.UnitTests.BackEnd.TaskRegistry_Tests+NullTaskTypeTaskFactory" could not be loaded from the assembly "D:\msbuild\bin\Debug\AnyCPU\Windows_NT\Output\Microsoft.Build.Engine.UnitTests.dll".  The task factory must return a value for the "TaskType" property.
```

PR #2368, opened 2017-07-31
Fix #2367: incorrect evaluation of VisualStudioInstallRootDirectory

### LINKED ISSUE #2367: Incorrect value for MSBuildExtensionsPath when using Microsoft.Build.dll in 64bits
Using the following project:
https://github.com/emmanuelguerin/dotnet-builder/tree/bug64bits

When running the project in 32 bits mode using MsBuild inside Visual Studio 2017 Entreprise 15.2 (build configuration Debug32), the build script returns correct `MSBuildExtensionsPath` value:
```
Extension Path: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild
```

When running the project in 64 bits mode (build configuration Debug) and using the DLL inside:
`C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64`, the `MSBuildExtensionsPath` variable is incorrectly initialized to:

```
Extension Path: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\MSBuild
```

emmanuelguerin:
When using TryFromMSBuildAssembly with Microsoft.Build.dll inside the amd64 directory, the VisualStudioInstallRootDirectory
is incorrectly evaluated as if in 32 bits mode.

Fix for #2367

### DIFF
--- a/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
+++ b/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
@@ -386,6 +386,8 @@ namespace Microsoft.Build.Engine.UnitTests
         }
 
         [Fact]
+        [Trait("Category", "nonlinuxtests")]
+        [Trait("Category", "nonosxtests")]
         public void BuildEnvironmentVSFromMSBuildAssembly()
         {
             using (var env = new EmptyVSEnviroment())
@@ -403,6 +405,8 @@ namespace Microsoft.Build.Engine.UnitTests
         }
 
         [Fact]
+        [Trait("Category", "nonlinuxtests")]
+        [Trait("Category", "nonosxtests")]
         public void BuildEnvironmentVSFromMSBuildAssemblyAmd64()
         {
             using (var env = new EmptyVSEnviroment())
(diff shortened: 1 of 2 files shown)


AndyGerlicher:
This looks right. Thanks for the contribution! Before taking this I'd like to see a unit test for the scenario. There is a BuildEnvironment test class where all of this can be mocked. If you aren't able to get to that we can look into adding the test as well. Just let us know.

AndyGerlicher:
@emmanuelguerin I added a unit test for the scenario that fails without your change. I also had to force push your branch (you were several hundred commits behind master). Just FYI. And thanks again for the contribution!

src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -385,6 +385,44 @@ public void BuildEnvironmentNoneWhenNotAvailable()
             }
         }
 
+        [Fact]
+        [Trait("Category", "nonlinuxtests")]
+        [Trait("Category", "nonosxtests")]
```
rainersigwald: `[PlatformSpecific(Xunit.PlatformID.Windows)]` instead?

AndyGerlicher: Think I'd rather change these in bulk. When I tried to update the version of xunit and get rid of BuildTools a while back there were some discrepancies (don't remember details right now).

emmanuelguerin:
Thanks for the tests! Do you know in which version this fix will eventually get released?

PR #2372, opened 2017-07-31
Allow enumerating Project.Targets.Keys

rainersigwald:
The lack of this API causes CPS to fail when it calls project.Targets.Keys in some DDRIT scenarios here:  https://mseng.visualstudio.com/VSIDEProj/_git/VSIDEProj.CPS?path=%2Fsrc%2FMicrosoft.VisualStudio.ProjectSystem.Implementation%2FBuild%2FOutputGroupProviderBase.cs&line=75&lineStyle=plain&lineEnd=76&lineStartColumn=1&lineEndColumn=1

Removing this functionality was observable from outside MSBuild because
fields backed by a RetrievableEntryHashSet are exposed in the public
API. Specifically, 6881b07 caused a regression in TypeScript by throwing
when CPS tried to enumerate target names using `Project.Targets.Keys`.


src/Build.OM.UnitTests/Instance/ProjectTargetInstance_Tests.cs
```diff
@@ -130,6 +131,29 @@ public void FileLocationAvailableEvenAfterEdits()
             }
         }
 
+        [Fact]
+        public void ProjectTargetsAreEnumerableByName()
```
cdmihai: Another viable location for this test might be [Project_Tests](https://github.com/Microsoft/msbuild/blob/master/src/Build.OM.UnitTests/Definition/Project_Tests.cs). Though that class is already bloated.

rainersigwald:
I think I like this approach (implement the dropped `.Keys` without bringing back `ReadOnlyConvertingCollection` better. Anyone disagree?

rainersigwald:

PR #2377, opened 2017-08-01
Remove namespace check of every XML element

### LINKED ISSUE #2373: 0.5% of a design-time build is just checking that an element has a valid namespace
Looking at design-time builds of Roslyn (http://github.com/dotnet/roslyn), it looks like 0.5% of a build is just spent validating a namespace of an element:

image

Looking at the code it's doing a string compare - is there a way we can switch this is reference equality via a name table?

rainersigwald:
Fixes #2373 by only checking the top-level namespace of a project, not
the namespace of every element.

This is arguably incorrect since XML namespace _could_ be per-element,
but a) that never worked before and b) you'd have to manually edit a
project in a fairly odd way to get that, so the performance gain here
seems to outweigh the risk.

### DIFF
--- a/src/Build/Xml/ProjectXmlUtilities.cs
+++ b/src/Build/Xml/ProjectXmlUtilities.cs
@@ -45,7 +45,6 @@ namespace Microsoft.Build.Internal
 
                     case XmlNodeType.Element:
                         XmlElementWithLocation childElement = (XmlElementWithLocation)child;
-                        VerifyThrowProjectValidNamespace(childElement);
                         children.Add(childElement);
                         break;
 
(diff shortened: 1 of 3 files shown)


src/Build/Xml/ProjectXmlUtilities.cs
```diff
@@ -97,19 +96,6 @@ internal static void ThrowProjectInvalidChildElement(string name, string parentN
...
         /// <summary>
-        /// Verifies that an element is in the MSBuild namespace, otherwise throws an InvalidProjectFileException.
-        /// </summary>
-        internal static void VerifyThrowProjectValidNamespace(XmlElementWithLocation element)
-        {
-            // If a namespace was specified it must be the default MSBuild namespace.
-            if (!VerifyValidProjectNamespace(element))
-            {
-                ProjectErrorUtilities.ThrowInvalidProject(element.Location,
-                    "CustomNamespaceNotAllowedOnThisChildElement", element.Name, element.ParentNode?.Name);
```
davkean: I'm guessing this resource is now dead?

davkean:
👍

Issue #608, opened 2016-05-05
List conflicts in MSB3277 reference-version-conflict message


rainersigwald:
This is to keep track of this feature request from Connect: http://connect.microsoft.com/VisualStudio/feedback/details/2619450/unclear-warning-msb3277-found-conflicts-between-different-versions-of-the-same-dependent-assembly-that-could-not-be-resolved

> During compile I've got a warning: C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1819,5): 
> warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.
>  These reference conflicts are listed in the build log when log verbosity is set to detailed. 
> 
> I believe that compiler knows the name of assembly that has a problem, and it can be included in the messages without re-compile in diagnostics mode.
> 
> Could you please change the message to include  assemblies name, as it does in diagnostics mode.

This message is emitted in [ResolveAssemblyReferences](https://github.com/Microsoft/msbuild/blob/8b488bd508807f415193fe38d5084cd590931cfb/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs#L1069).

rainersigwald:
We discussed this in triage this week. We're in favor of clear, actionable error messages, but since the information is just a rebuild with `/v:diag` away, we really don't want to slow down the normal/success case by carrying around extra information just to support this error. A proposed solution to this problem should not slow down the no-error case.

cdmihai:
It seems like the information might be available, but just not logged: https://github.com/Microsoft/msbuild/blob/72c12de51702f7e214f1eba481f2f3f93b885616/src/XMakeTasks/AssemblyDependency/ResolveAssemblyReference.cs#L1039-L1042

rainersigwald:
_From @davkean in #1712:_

This error message has been the bane of .NET developers for years. Having worked on assembly resolution for ~6 years in the BCL team, I've diagnosed so many of these warnings for customers, and there's three issues that I see with it:

1. Developers don't know what this message means.

2. Developers once they know what it means, don't know how to fix it.

3. The warning is not actionable without building again with diagnostic build (if you can figure out how to do that). Even then you need to search for the issue in the log, and find the relevant portions.

Here's a typical log:

```
[...]

PR #2379, opened 2017-08-01
Include the name of the conflicting assembly as a part of warning MSB3277

theunrepentantgeek:
Modify the message MSB3277 to include the name of the conflicting assembly - a partial resolution of #608 . The message will be repeated once for each distinct assembly that is experiencing conflicts.

Old message:
```
MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.  
These reference conflicts are listed in the build log when log verbosity is set to detailed.
```

New message (sample):
```
MSB3277: Found conflicts between different versions of "Newtonsoft.Json" that could not be resolved.  
These reference conflicts are listed in the build log when log verbosity is set to detailed.
```

I found two unit tests that trigger this warning - `ConflictWithBackVersionPrimary()` and `ConflictWithBackVersionPrimaryWithAutoUnify()`, both found in `src\Tasks.UnitTests\AssemblyDependency\Miscellaneous.cs`; both tests have been modified to assert the correct message exists in the log.

### DIFF
--- a/src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
+++ b/src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
@@ -3770,6 +3770,7 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
             bool result = Execute(t);
 
             Assert.Equal(1, e.Warnings); // @"Expected one warning."
+            e.AssertLogContainsMessageFromResource(AssemblyResources.GetString, "ResolveAssemblyReference.FoundConflicts", "D");
 
             Assert.Equal(0, t.SuggestedRedirects.Length);
             Assert.Equal(3, t.ResolvedFiles.Length);
@@ -3808,6 +3809,7 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
             bool result = Execute(t);
 
             Assert.Equal(1, e.Warnings); // @"Expected one warning."
+            e.AssertLogContainsMessageFromResource(AssemblyResources.GetString, "ResolveAssemblyReference.FoundConflicts", "D");
 
             Assert.Equal(0, t.SuggestedRedirects.Length);
             Assert.Equal(3, t.ResolvedFiles.Length);
--- a/src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
+++ b/src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
@@ -3539,8 +3539,10 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
             Execute(t);
 
             Assert.Equal(3, t.ResolvedDependencyFiles.Length);
-            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V2_DDllPath)); // "Expected to find assembly, but didn't."
-            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_DDllPath)); // "Expected to find assembly, but didn't."
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V2_DDllPath));
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_DDllPath));
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V2_GDllPath));
+            
             Assert.Equal(1, t.SuggestedRedirects.Length);
             Assert.True(ContainsItem(t.SuggestedRedirects, @"D, Culture=neutral, PublicKeyToken=aaaaaaaaaaaaaaaa")); // "Expected to find suggested redirect, but didn't"
             Assert.Equal(1, e.Warnings); // "Should only be one warning for suggested redirects."
@@ -3939,8 +3941,11 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
 
             Assert.True(result); // @"Expected a success because this conflict is solvable."
             Assert.Equal(3, t.ResolvedFiles.Length);
-            Assert.True(ContainsItem(t.ResolvedFiles, s_myLibraries_V2_DDllPath)); // "Expected to find assembly, but didn't."
+            Assert.True(ContainsItem(t.ResolvedFiles, s_myLibraries_V2_DDllPath));
+
             Assert.Equal(2, t.ResolvedDependencyFiles.Length);
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_DDllPath));
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V2_GDllPath));
         }
 
 
@@ -4569,9 +4574,12 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
             Execute(t);
 
             Assert.Equal(2, t.ResolvedFiles.Length);
+
             Assert.Equal(4, t.ResolvedDependencyFiles.Length);
-            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_DDllPath)); // "Expected to find assembly, but didn't."
-            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_DDllPath)); // "Expected to find assembly, but didn't."
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_DDllPath));
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V2_DDllPath));
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V2_GDllPath));
+            Assert.True(ContainsItem(t.ResolvedDependencyFiles, s_myLibraries_V1_E_EDllPath));
 
             foreach (ITaskItem item in t.ResolvedDependencyFiles)
             {
--- a/src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
+++ b/src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
@@ -3820,6 +3820,47 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
         /// Consider this dependency chain:
         ///
         /// App
+        ///   References - B
+        ///        Depends on D version 2
+        ///   References - D, version 1
+        ///
+        /// Both D1 and D2 are CopyLocal. This is a warning because D1 is a lower version
+        /// than both D2 so that can't unify. These means that eventually when 
+        /// they're copied to the output directory they'll conflict.
+        /// </summary>
+        [Fact]
+        public void ConflictGeneratesMessageReferencingAssemblyName()
+        {
+            ResolveAssemblyReference t = new ResolveAssemblyReference();
+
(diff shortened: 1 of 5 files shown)


davkean:
👍 Note, this will increase the number of warnings a given project will have, so there is a theoretical break here if the developer has something that baselines warnings and fails if they increase. /warnaserror usage will be fine because they have the same code.

This is a great start, it the next step (not this PR) would be actually include the conflicts; similar to how NuGet has package conflicts in 15.3.

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -1037,7 +1036,9 @@ ArrayList generalResolutionExceptions
 
                                 if (conflictVictims.Length == 0)
                                 {
-                                    foundAtLeastOneUnresolvableConflict = true;
+                                    // This warning is logged regardless of AutoUnify since it means a conflict existed where the reference
+                                    // chosen was not the conflict victor in a version comparison, in other words it was older.
+                                    Log.LogWarningWithCodeFromResources("ResolveAssemblyReference.FoundConflicts", idealRemappingPartialAssemblyName.Name);
```
KirillOsenkov: If I'm reading this right, this is inside both the `for-i` and `for-j` loops. `idealRemappingPartialAssemblyName` only seems to depend on `i`, but not `j`.

I think maybe it's better to keep the bool `foundAtLeastOneUnresolvableConflict`, and set it to true as before, but log the warning outside of `for-j` but inside the `for-i` loop, so that it is output only once per each assembly name.

I didn't test my theory, it's all just from reading the code and psychic debugging. It'd be great to have unit-tests support or deny this.

theunrepentantgeek: I think you're right.

Noting that `conflictVictims` is also invariant across the `for-j` loop, I've promoted both variable and warning message out of the loop.

I tried to create a unit test that generated multiple identical messages but didn't manage to find a way.

theunrepentantgeek:
I've made some progress on getting a unit test written - largely have my head around the way `MockEngine` stubs out the filesystem to allow the unit tests to run in memory.

But, every scenario that I've tried ends up giving me a different error (`MSB3247`) instead.

rainersigwald:
@theunrepentantgeek I was running into very similar problems. This is what I got: https://github.com/rainersigwald/msbuild/tree/force-msb3277-test

But it doesn't seem to emit _any_ warnings, and I'm not entirely sure why.

theunrepentantgeek:
I've had a look at your branch @rainersigwald and I suspect we're trying to test different things.

Unless I'm reading them incorrectly (always a possibility), your tests are trying to generate a MSB3277 warning so you can check the message is the expected one. *Is this right?*

I found two existing tests `ConflictWithBackVersionPrimary()` and `ConflictWithBackVersionPrimaryWithAutoUnify()` (both in `Miscellaneous.cs`) that already triggered MSB3277 and then extended them to also check the message.

What I've been unable to do is to write a test where `idealRemapping.BindingRedirects` (see `ResolveAssemblyReferences.cs` around line 976) contains more than one item - to try and verify that moving the code out of the `for-j` loop was necessary.

KirillOsenkov:
I found this old bug: https://github.com/Microsoft/msbuild/issues/1996

It mentions that you can trigger this if the primary reference has the lower version than the victim one. Hope this helps??

theunrepentantgeek:
It's not possible for the property `idealRemapping.BindingRedirects` to contain more than one item, even though it's an array - the code that populates it only supplies a single element array. Here's the code from `ReferenceTable.cs`, around line # 1914, emphasis comment added:

``` csharp
foreach (AssemblyNameReference assemblyNameReference in assemblyNamesList)
{
    DependentAssembly remapping = new DependentAssembly();
    remapping.PartialAssemblyName = assemblyNameReference.assemblyName.AssemblyName;
    BindingRedirect bindingRedirect = new BindingRedirect();
    bindingRedirect.OldVersionLow = new Version("0.0.0.0");
    bindingRedirect.OldVersionHigh = assemblyNameReference.assemblyName.AssemblyName.Version;
    bindingRedirect.NewVersion = assemblyNameReference.assemblyName.AssemblyName.Version;

    // Populate BindingRedirects with a single element array
    remapping.BindingRedirects = new BindingRedirect[] { bindingRedirect };

    idealRemappingsList.Add(remapping);
}
```
While moving the log message out of the `for-j` loop was the right thing to do (it makes more sense when a future dev is reading `LogResults()`), this means that writing a unit test to show that moving the message prevents duplicates can't be done.

Arguably, this likely also means that the array property `BindingRedirects` should be changed to a simple reference - but that's a challenge for another time.

I **have** added an explicit test on the message in MSB3277 which you can see in commit b50c788491bdab26fe2baab82b5ad7377ca17b40.

KirillOsenkov:
Awesome detective work, thanks!

I found only one other location here where it's being assigned an array with potentially multiple elements, however I'm not sure that code ever runs:
http://source.dot.net/#Microsoft.Build.Tasks.Core/AppConfig/DependentAssembly.cs,132
http://source.dot.net/#Microsoft.Build.Tasks.Core/AppConfig/DependentAssembly.cs,ddfbec501dc56af9,references
http://source.dot.net/#Microsoft.Build.Tasks.Core/AppConfig/DependentAssembly.cs,c8cedbffa26ad5e0,references

Yeah I think you're right and it's impossible to get that message multiple times, so don't worry about testing that part I think. All changes look good to me.

KirillOsenkov:
I also just found this, not sure if relevant:
http://source.dot.net/#Microsoft.Build.Tasks.UnitTests/AppConfig_Tests.cs,37

theunrepentantgeek:
@KirillOsenkov: I had a look at  your test, and I have a suggestion.

In `GetAssemblyName()` (found in `ResolveAssemblyReferenceTestFixture.cs`), the assembly names you're returning for `Alpha.dll` and `Beta.dll` have zero version specified:

```
if (string.Compare(path, s_myLibraries_V1_DependsOnV1OfAlphaAndBetaDllPath, StringComparison.OrdinalIgnoreCase) == 0)
{
    return new AssemblyNameExtension("DependsOnV1OfAlphaAndBeta, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null");
}

if (string.Compare(path, s_myLibraries_V2_DependsOnV2OfAlphaAndBetaDllPath, StringComparison.OrdinalIgnoreCase) == 0)
{
    return new AssemblyNameExtension("DependsOnV2OfAlphaAndBeta, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null");
}
```

I suspect this would cause the second reference to merge with the first (same name, same version), leaving no conflict. What happens if you mark the first as `v1.0.0.0` and the second as `v2.0.0.0`?

theunrepentantgeek:
I've read through the build logs for the failing build on Windows and can't see what the problem might be - looks like a problem writing a file rather than my code change.

rainersigwald:

Agreed, that doesn't look related to your change.

theunrepentantgeek:
Is there anything more you want done on this PR @rainersigwald ? 
I believe it's got good unit test coverage now, but if there's something else you have in mind I can give it a shot.

rainersigwald:
Sorry, this got a bit lost in my to-do list.

What does the new test you created add over the ones that you modified?

What I was hoping for is a test that shows this warning being emitted for all of the assembly names that are in this state (each `idealAssemblyRemappings`). It sounds like you're saying that's not possible, but I don't understand why; the detailed RAR log should have log information for all such conflicts so it should be somewhere.

That's why I was trying to construct a new set of references; the current set has two versions of `D` but I couldn't see another one to drag in.

theunrepentantgeek:
@rainersigwald - I've added a new test `ConflictGeneratesMessageReferencingEachConflictingAssemblyName` to check the new message is correctly emitted when there are multiple active conflicts. I needed to add another conflicting reference (`G.dll`) to the fixture metadata.

The comment I made above ("It's not possible for the property idealRemapping.BindingRedirects to contain more than one item") refers to the collection *within* each `idealRemapping` item - the code in the public repo always populates that collection with a single item only.

theunrepentantgeek:
Found some tests that I'd broken by adding the new dependency in the fixture metadata - have updated those to pass. Also fixed a dumbass compilation error on my part.

src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
```diff
@@ -3538,7 +3538,7 @@ public void ConflictBetweenNonCopyLocalDependencies()
 
             Execute(t);
 
-            Assert.Equal(2, t.ResolvedDependencyFiles.Length);
+            Assert.Equal(3, t.ResolvedDependencyFiles.Length);
```
rainersigwald: Is the new one here `G`, added for the new test? Please add an assert for it as "documentation".

src/Tasks.UnitTests/AssemblyDependency/Miscellaneous.cs
```diff
@@ -3818,6 +3820,91 @@ public void ConflictWithBackVersionPrimaryWithAutoUnify()
...
+        /// Consider this dependency chain:
+        ///
+        /// App
+        ///   References - B
+        ///        Depends on D version 2
+        ///        Depends on G, version 2
+        ///   References - D, version 1
+        ///   References - G, version 1
+        ///
+        /// All of Dv1, Dv2, Gv1 and Gv2 are CopyLocal. We should get two conflict warnings, one for D and one for G.
```
rainersigwald: Nice!

theunrepentantgeek:
I've added assertions to the affected tests to ensure that `.ResolvedDependencyFiles` contains only the expected items.

BTW - I spotted a bug in the test `Regress265054`: it was asserting the same thing twice.

davkean:
@AndyGerlicher @jeffkl @cdmihai Is there anything holding this from getting merged?

AndyGerlicher:
Thanks for bringing this back to the top @davkean. I think everything is addressed. And perf looks good:

### Evaluation: Time (ms)
Test | Overall | Significant  | Value
:--- | :--- | :--- | :---
DotnetConsoleProject | :red_circle: | yes | 26.9986 -> 27.0686 (0.259%)
DotnetWebProject | ::ok_hand: | no | 41.4498 -> 41.4041 (-0.11%)
DotnetMvcProject | :white_check_mark: | yes | 78.7176 -> 78.5915 (-0.16%)
Picasso | :ok_hand: | no | 232.4183 -> 232.8356 (0.18%)
SmallP2POldCsproj | :white_check_mark: | yes | 35.6656 -> 35.5598 (-0.297%)
Generated_100_100_v150 | :ok_hand: | no | 1373.0152 -> 1376.4266 (0.248%)
LargeP2POldCsproj | :white_check_mark: | yes | 636.4005 -> 633.2157 (-0.5%)
roslyn | :white_check_mark: | yes | 5188.0676 -> 5165.7136 (-0.431%)

Test | Overall | Significant  | Value
:--- | :--- | :--- | :---
DotnetConsoleProject | :ok_hand: | no | 4594952 -> 4595428 (0.01%)
DotnetWebProject | ::ok_hand: | no | 6076905 -> 6076807 (-0.002%)
DotnetMvcProject | ::ok_hand: | no | 8345692 -> 8345574 (-0.001%)
Picasso | :ok_hand: | no | 31947819 -> 31947889 (0%)
SmallP2POldCsproj | :ok_hand: | no | 5073401 -> 5073424 (0%)
Generated_100_100_v150 | ::ok_hand: | no | 171881050 -> 171881048 (0%)
LargeP2POldCsproj | :ok_hand: | no | 67948686 -> 67948859 (0%)
roslyn | :white_check_mark: | yes | 571248416 -> 571243687 (-0.001%)

AndyGerlicher:
Thanks for the contribution!

PR #2384, opened 2017-08-02
Reuse RequestBuilder threads

benaadams:
RequestBuilder queues tasks with `TaskCreationOptions.LongRunning` which means their threads don't interfere with the threadpool. However it also means a new thread is created for every task and it is unbounded.

This means a Nuget restore of Rosyln creates 3274 short-lived Request Builder threads

This change introduces a TaskScheduler which caps the number of created threads (at processor count); and reuses them when they have stopped doing work rather than creating a new thread.

This means the same Nuget restore creates only 3 threads (on an 8 core machine, so upper bound 8)

I *believe* this change is safe as there is also a STA TaskScheduler which only allows one thread; so on the assumption that works; this should also work.

/cc @stephentoub

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -108,6 +109,16 @@ namespace Microsoft.Build.BackEnd
         /// Flag indicating whether this request builder has been zombied by a cancellation request.
         /// </summary>
         private bool _isZombie = false;
+
+        /// <summary>
+        /// The dedicated scheduler object.
+        /// </summary>
+        private static TaskScheduler _dedicatedScheduler;
+
+        /// <summary>
+        /// Gets the dedicated scheduler.
+        /// </summary>
+        private TaskScheduler DedicatedScheduler => _dedicatedScheduler ?? (_dedicatedScheduler = new DedicatedThreadsTaskScheduler());
 
         /// <summary>
         /// Creates a new request builder.
@@ -626,7 +637,7 @@ namespace Microsoft.Build.BackEnd
                         },
                         _cancellationTokenSource.Token,
                         TaskCreationOptions.None,
-                        AwaitExtensions.DedicatedThreadsTaskSchedulerInstance).Unwrap();
+                        DedicatedScheduler).Unwrap();
                 }
             }
         }
@@ -1308,6 +1319,58 @@ namespace Microsoft.Build.BackEnd
             }
             
             return new HashSet<string>(ExpressionShredder.SplitSemiColonSeparatedList(warnings), StringComparer.OrdinalIgnoreCase);
+        }
+
+        private sealed class DedicatedThreadsTaskScheduler : TaskScheduler
+        {
+            private readonly BlockingCollection<Task> _tasks = new BlockingCollection<Task>();
+            private int _availableThreads = 0;
+
+            protected override void QueueTask(Task task)
+            {
+                RequestThread();
+                _tasks.Add(task);
+            }
+
+            protected override bool TryExecuteTaskInline(Task task, bool taskWasPreviouslyQueued) => false;
+
+            protected override IEnumerable<Task> GetScheduledTasks() => _tasks;
+
+            private void RequestThread()
+            {
+                // Decrement available thread count; don't drop below zero
+                // Prior value is stored in count
+                var count = Volatile.Read(ref _availableThreads);
+                while (count > 0)
+                {
+                    var prev = Interlocked.CompareExchange(ref _availableThreads, count - 1, count);
+                    if (prev == count)
+                    {
+                        break;
+                    }
+                    count = prev;
+                }
+
+                if (count == 0)
+                {
+                    // No threads were available for request
+                    InjectThread();
+                }
+            }
+
+            private void InjectThread()
+            {
+                var thread = new Thread(() =>
+                {
+                    foreach (Task t in _tasks.GetConsumingEnumerable())
+                    {
+                        TryExecuteTask(t);
+                        Interlocked.Increment(ref _availableThreads);
+                    }
+                });
+                thread.IsBackground = true;
+                thread.Start();
+            }
         }
     }
 }
--- a/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
@@ -46,6 +46,11 @@ namespace Microsoft.Build.BackEnd
     internal class RequestBuilder : IRequestBuilder, IRequestBuilderCallback, IBuildComponent
     {
         /// <summary>
+        /// The dedicated scheduler object.
+        /// </summary>
+        private static readonly TaskScheduler s_dedicatedScheduler = new DedicatedThreadsTaskScheduler();
+
+        /// <summary>
         /// The event used to signal that this request should immediately terminate.
         /// </summary>
         private ManualResetEvent _terminateEvent;
@@ -109,16 +114,6 @@ namespace Microsoft.Build.BackEnd
         /// Flag indicating whether this request builder has been zombied by a cancellation request.
         /// </summary>
         private bool _isZombie = false;
-
-        /// <summary>
-        /// The dedicated scheduler object.
-        /// </summary>
-        private static TaskScheduler _dedicatedScheduler;
-
-        /// <summary>
-        /// Gets the dedicated scheduler.
-        /// </summary>
-        private TaskScheduler DedicatedScheduler => _dedicatedScheduler ?? (_dedicatedScheduler = new DedicatedThreadsTaskScheduler());
 
         /// <summary>
         /// Creates a new request builder.
@@ -637,7 +632,7 @@ namespace Microsoft.Build.BackEnd
                         },
                         _cancellationTokenSource.Token,
                         TaskCreationOptions.None,
-                        DedicatedScheduler).Unwrap();
+                        s_dedicatedScheduler).Unwrap();
                 }
             }
         }
--- a/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
@@ -625,8 +625,8 @@ namespace Microsoft.Build.BackEnd
                             return this.RequestThreadProc(setThreadParameters: true);
                         },
                         _cancellationTokenSource.Token,
-                        TaskCreationOptions.LongRunning,
-                        TaskScheduler.Default).Unwrap();
+                        TaskCreationOptions.None,
+                        AwaitExtensions.DedicatedThreadsTaskSchedulerInstance).Unwrap();
                 }
             }
         }
@@ -658,8 +658,10 @@ namespace Microsoft.Build.BackEnd
                 threadName = "RequestBuilder STA thread";
             }
 #endif
-
-            Thread.CurrentThread.Name = threadName;
+            if (string.IsNullOrEmpty(Thread.CurrentThread.Name))
+            {
+                Thread.CurrentThread.Name = threadName;
+            }
         }
 
         /// <summary>
--- a/src/Shared/AwaitExtensions.cs
+++ b/src/Shared/AwaitExtensions.cs
@@ -238,11 +238,8 @@ namespace Microsoft.Build.Shared
 
         private sealed class DedicatedThreadsTaskScheduler : TaskScheduler
         {
-            private static readonly int _maxThreads = Environment.ProcessorCount;
-
             private readonly BlockingCollection<Task> _tasks = new BlockingCollection<Task>();
             private int _availableThreads = 0;
-            private int _createdThreads = 0;
 
             protected override void QueueTask(Task task)
             {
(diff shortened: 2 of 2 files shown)


KirillOsenkov:
Naive question: why not just Task.Run() here?

benaadams:
> Naive question: why not just Task.Run() here?

¯\_(ツ)_/¯ I assume was some reason...

benaadams:
> Naive question: why not just Task.Run() here?

`¯\_(ツ)_/¯` I assume was some reason...

Can it be embedded in something else like VS? Then using the threadpool with blocking I/O (XmlDocument why do you have no async load..?) might cause issues?

onovotny:
@benaadams After this change, what is the performance impact? Does restore go faster (by how much)?

benaadams:
test Windows_NT Build for Full

benaadams:
>  what is the performance impact

Means I can profile MSBuild rather than the profiler getting weirded out by 3286 thread stacks 😄

onovotny:
Guess I'm wondering if this change *alone* has a material impact in improving wall-time since there'd presumably be less context switching and overhead?

stephentoub:
> a material impact in improving wall-time

Or a material regression.  One could imagine a situation where that would be true, e.g. where lots of the work items spend a lot of time blocking, such that the blocking will now be serialized to just three blocking in parallel.  In theory it could even cause deadlocks.  That said, I'm only speaking in generalities because I have no idea what the work items are actually doing.

onovotny:
@stephentoub in this case though, if the threads were blocking, wouldn't it have hit the peak of 8 threads and stopped there? Since it only hit 3 of 8, it would seem that the threads weren't blocking?

stephentoub:
> in this case though

Like I said, I've not looked at what the work is doing.  It's quite possible they're not blocking and entirely independent.  It's also possible there are dependencies that would manifest only if certain orderings occurred, e.g. imagine if you had a set of work items where some of them depended on previous ones finishing... if you queued them all in order, you might only end up with a few running concurrently, but if you queued them in the opposite order, you could end up with a ton of them blocking.

onovotny:
Fair enough, it's all speculation until tests and timings are run :)

benaadams:
> such that the blocking will now be serialized to just three blocking in parallel. 

It would be CPU count; it just goes up to three in this test; rather than the 8 it could go up to. Can increase to a multiple of cores; or remove the upper bound (for this it would have same outcome)

> In theory it could even cause deadlocks. 

I assume that's why it doesn't queue to threadpool and uses `TaskCreationOptions.LongRunning` currently, so the ThreadPool is free is its running in-proc with something.

For pure MSBuild it also has an STA mode which is purely single threaded; hence my caveat "if that works..."

It seems to be serialized in order, e.g. it doesn't kick off a future task that depends on results of previous task so waits for it and potentially deadlocking. However; don't know if MSBuild allows overlapping dependent tasks.

Main motivation for the upper limit was from coreclr/corefx builds which seem to spawn a huge number of parallel compiles and absolutely crush my machine; and assume spend a lot of time thread switching.

Will remove the upper bound to make it a "safe" change; can revisit it later..?

stephentoub:
> Will remove the upper bound to make it a "safe" change; can revisit it later..?

Isn't the upper bound what's preventing "a huge number of parallel compiles and absolutely crush my machine"?

benaadams:
> Isn't the upper bound what's preventing...

Yep; but not having it still produces the results I outlined in the PR summary; while maintaining previous behavior otherwise. 

So can address the cap in a different PR with an investigation - though likely I'll just raise an issue, as would need a much greater understanding of what MSBuild actually does (implementation-wise) than I have or can dedicate time to gaining.

onovotny:
@benaadams isn't the magic here really that you're reusing threads and not creating a new LongRunning one each time?

stephentoub:
> isn't the magic here

That would help if the problem was around the overhead associated with creating/destroying threads.  But it sounded like the problem was actually lots of threads all running/competing at the same time.  If that's not actually the problem, then what's the value of the PR?  If it's the cost of the thread overhead, are there numbers to justify the added complexity?

src/Shared/AwaitExtensions.cs
```diff
@@ -52,6 +63,28 @@ internal static TaskScheduler OneSTAThreadPerTaskSchedulerInstance
         }
 
         /// <summary>
+        /// Gets the dedicated scheduler.
+        /// </summary>
+        internal static TaskScheduler DedicatedThreadsTaskSchedulerInstance
```
cdmihai: Can you make this a field backed property to avoid the locking?

benaadams:
The MSBuild part of a NuGet restore of Rosyln creates and destorys 3274 builder threads; it only ever really uses CPU for 1 thread; which is the main reason behind this PR.

The cap was mostly an add-on to stop parallelizable tasks from going too wild; which coreclr/corefx seem to do for full builds (e.g. I assume most of the tests can be happily built in parallel).

However, that was a reach without much analysis, so removing that bit for the concerns @stephentoub highlighted (i.e. I don't actually know what it does)

benaadams:
Removed cap and moved to property rather than static; results the same

src/Shared/AwaitExtensions.cs
benaadams: Done

NickCraver:
I'm curious: what's the impact on overall start->end wall time here?

KirillOsenkov:
Keep in mind that each thread has a 1MB stack, so 3000 threads are 3GB of memory right there. Also makes profiling impossible since the profiler chokes getting the stacks of 3000 threads. Also starting a thread is expensive, it calls DllThreadAttach and stuff. I love this change and we do need to fix this.

I'm just thinking maybe it's worth just trying the default threadpool, maybe there's no real reason to have long running in the first place?

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -110,6 +111,16 @@ internal class RequestBuilder : IRequestBuilder, IRequestBuilderCallback, IBuild
...
 
         /// <summary>
+        /// The dedicated scheduler object.
+        /// </summary>
+        private static TaskScheduler _dedicatedScheduler;
+
+        /// <summary>
+        /// Gets the dedicated scheduler.
+        /// </summary>
+        private TaskScheduler DedicatedScheduler => _dedicatedScheduler ?? (_dedicatedScheduler = new DedicatedThreadsTaskScheduler());
```
cdmihai: `private static TaskScheduler _dedicatedScheduler = new DedicatedThreadsTaskScheduler();` and then the static property should just return the field. The CLR promises type creation is thread safe, so assignment expressions to static fields is thread safe. Therefore no locking needed.

benaadams:
> I'm curious: what's the impact on overall start->end wall time here?

Now I'm going to have to write a [time](https://linux.die.net/man/1/time) tool :P

benaadams:
Powershell! `Measure-Command `

stephentoub:
> However, that was a reach without much analysis, so removing that bit for the concerns @stephentoub highlighted

There are concerns to this change without the cap, too.  Previously threads would go away when they complete; now once a thread is added, it hangs out forever waiting for more work.  That means if, for example, you have a burst of work, you'll add a bunch of threads, and then even if they're unnecessary, they'll hang out waiting for more work until the process ends.  Maybe that's not a big deal, but any change like this really needs to be understood in a greater context of varied workloads that will stress it in different ways.

I'm not against a change like this in practice.  But writing thread pools is hard, and to write one for a specific use case really demands understanding what work is going to be executed, the various patterns involved, etc.

benaadams:
> I'm curious: what's the impact on overall start->end wall time here?

`NuGet restore C:\GitHub\roslyn\Roslyn.sln -NoCache`

Prior
`TotalSeconds      : 33.509692`
Post
`TotalSeconds      : 29.4461761`

Though I'm not sure I'd trust it; timing a process isn't a science 😉

benaadams:
> There are concerns to this change without the cap, too. Previously threads would go away when they complete; now once a thread is added, it hangs out forever waiting for more work....

😢 true. 

Ideally would move it to the ThreadPool but it uses `XmlDocument` which has no `LoadAsync` methods; (though `XmlReader` supports async); it isn't easy to make the equivalent as it delegates to the internal type `XmlLoader` which then uses lots of internal types.

Could use the `Load(string)` method and go async via StringBuilder, but allocs will be ugly

omariom:
> Could use the Load(string) method and go async via StringBuilder, but allocs will be ugly

MMF via UnmanagedMemoryStream?

omariom:
Ah.. that will be sync Io anyway

KirillOsenkov:
@benaadams https://github.com/KirillOsenkov/CodeCleanupTools/tree/master/Time

benaadams:
> The cap was mostly an add-on to stop parallelizable tasks from going too wild; which coreclr/corefx seem to do for full builds

Turns out this bit was wishful thinking as a coreclr build runs 8 x MSBuild on my 8 core machine...

The XmlDoc was a little distraction as the main issue is the Tasks can be anything; for out-of-proc processes, I can make the waits async rather than sync.

However, they can also be in proc and after going through a smattering of NuGet MSBuild target tasks on GitHub about half seem to use `.Wait` or `.WaitForExit`; so moving everything to the main threadpool may cause issues.

So question is whether to create a secondary threadpool; which seems like reinventing the wheel...

rainersigwald:
Hey, just wanted to say: we haven't forgotten about this but we'd like a core team member to do some investigation into exactly what's going on in the bad cases before merging this.

davkean:
I ran compared these changes against the previous commit using the internal tests for evaluation and got the following numbers:

DotnetConsoleProject | Evaluation: Time (ms) | :ok_hand: | no | 31.7814 -> 31.6024 (-0.563%)
DotnetConsoleProject | Evaluation: Memory | :ok_hand: | no | 5982291 -> 5982387 (0.002%)
DotnetWebProject | Evaluation: Time (ms) | :white_check_mark: | yes | 45.8522 -> 45.3494 (-1.097%)
DotnetWebProject | Evaluation: Memory | :ok_hand: | no | 7730825 -> 7730733 (-0.001%)
DotnetMvcProject | Evaluation: Time (ms) | :white_check_mark: | yes | 76.994 -> 76.0761 (-1.192%)
DotnetMvcProject | Evaluation: Memory | :ok_hand: | no | 10100507 -> 10100740 (0.002%)
Picasso | Evaluation: Time (ms) | ::ok_hand: | no | 295.5332 -> 295.1646 (-0.125%)
Picasso | Evaluation: Memory | :ok_hand: | no | 46007825 -> 46008835 (0.002%)
SmallP2POldCsproj | Evaluation: Time (ms) | :white_check_mark: | yes | 50.2176 -> 49.6575 (-1.115%)
SmallP2POldCsproj | Evaluation: Memory | :ok_hand: | no | 7771757 -> 7771703 (-0.001%)
Generated_100_100_v150 | Evaluation: Time (ms) | :ok_hand: | no | 1355.856 -> 1354.705 (-0.085%)
Generated_100_100_v150 | Evaluation: Memory | :ok_hand: | no | 193990613 -> 193990580 (0%)
LargeP2POldCsproj | Evaluation: Time (ms) | :ok_hand: | no | 876.6518 -> 879.356 (0.308%)
LargeP2POldCsproj | Evaluation: Memory | :ok_hand: | no | 114923256 -> 114923067 (0%)
roslyn | Evaluation: Time (ms) | :ok_hand: | no | 4886.8108 -> 4879.6933 (-0.146%)
roslyn | Evaluation: Memory | :ok_hand: | no | 565353594 -> 577630985 (2.172%)

lahma:
I do like a good cliffhanger, but did anyone get final conclusion about mergeability? It sounds like this would help with future performance analysis at least.

davkean:
Currently it doesn't seem to have moved the dile in actual build time/evaluation time. I'd like to understand if this reduces the amount of CPU burned by SleepEx as called out in https://github.com/Microsoft/msbuild/issues/2391 - that would involve looking at traces of before and after. @rainersigwald's not back until middle of December, I suspect it will need to wait until then.

KirillOsenkov:
Keep in mind that the motivation is also debuggability (like Ben's profiler issue). Also I'd be curious to look at the memory footprint (1 thread ~ 1 MB for the stack).

Ideally we should have some numbers with this change, without this change and with a simple Task.Run (as opposed to a custom TaskScheduler).

rainersigwald:
Over the course of the investigation into #3047, I think @cdmihai and I leveled up our knowledge of RequestBuilder's internals to grok both the problem and this solution.

In particular, I had a big nagging question when seeing the results of this change: _why isn't one thread enough?_

I knew that, logically, MSBuild builds only one project at a time per node. If the project hits certain conditions (MSBuild task calling into another project, task yielding), it can yield the node--freeing the node to build other projects, but still only one thing at a time.

That's all managed through a series of async continuations. Each BuildRequest starts a task that has continuations that will build the entire project. "Yielding the node" means awaiting a task that will be satisfied when some other BuildRequest has completed. Unblocking of requests is serialized, so that when a result comes in and N projects that were waiting for it can now build only one is unblocked at a time.

So when is there ever a need for more than one thread? When queuing new work, we take a couple of quick hops that can happen in parallel with normal processing. I don't think they _have_ to.

## Concerns I still have

* Is the culture being juggled in all the appropriate places, now that we're using fewer threads? I don't _think_ this would be worse than before.
* Would it actually be better to scope this down to one thread, period? Clearly our predecessors thought that was suboptimal since there's the `OneSTAThreadPerTaskSchedulerInstance` option. But as-is, the mostly-single-threaded workload will hop from thread to thread if a new thread is ever created. Still better than hopping to a new thread every time!
* Does this have some other hidden problems?

## Next steps

Over the weekend, I'm going to spin up a private build of this and subject it to VS's internal "DDRIT" and "RPS" tests to see if anything falls out. Assuming they all pass, I think I'll be ready to pull this into our master branch so we can get real-world usage of it in VS previews.

I'll also push a commit to this PR that I think clarifies some names that I found confusing when reviewing (preexisting, not new to this PR).

davkean:
@rainersigwald Nice! Thanks for the follow up.

rainersigwald:
> In particular, I had a big nagging question when seeing the results of this change: _why isn't one thread enough?_

At long last, I can explain this now! https://github.com/dotnet/msbuild/pull/6042 has the details.

While figuring this out I tracked down the original commit that introduced the `LongRunning` option ([Microsoft-internal link](http://vstfdevdiv:8080/DevDiv2/DevDiv/_versionControl/changeset/306503?_a=compare&path=%24%2FDevDiv%2Fpu%2FWin8Tools%2Fvsproject%2Fxmake%2FXMakeBuildEngine%2FBackEnd%2FComponents%2FRequestBuilder%2FRequestBuilder.cs)). It was introduced to fix a bug ([Microsoft-internal link](http://vstfdevdiv:8080/DevDiv2/DevDiv/_workitems/edit/203360)):

> ## [Build, Performance] Consider reusing the same RequestBuilder thread/s - 10x more thread starts over SP1
> 
> We are starting way more threads than before. I see 10x as many in MSTV Incremental.

That's . . . almost exactly what @benaadams fixed with this PR. It must have been really _really_ bad for a while there!

PR #2390, opened 2017-08-02
[mono][tests] Fix SolutionGenerator test on mono/osx

radical:
Under Mono/osx, even when using in-tree msbuild, solution before/after
files (eg. NuGet) get picked up from the installed mono. So, only count
targets coming from `.sln.metaproj` files, which would be the ones
generated by the SolutionGenerator.

The change should be safe for all platforms.

Fixes
Microsoft.Build.UnitTests.Construction.SolutionProjectGenerator_Tests.IllegalUserTargetNamesDoNotThrow

### DIFF
--- a/src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
+++ b/src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
@@ -2108,6 +2108,9 @@ EndGlobal
 
             ProjectInstance[] instances;
 
+            // Avoid any unexpected targets getting pulled in
+            var globalProperties = new Dictionary<string, string> { { "ImportByWildcardBeforeSolution", "false" } };
+
             foreach (string builtInTargetName in new[]
             {
                 null,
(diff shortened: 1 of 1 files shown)


cdmihai:
There might be an alternative implementation. Apparently setting the global property `ImportByWildcardBeforeSolution` to `false` [will skip those imports](https://github.com/Microsoft/msbuild/blob/master/src/Build/Construction/Solution/SolutionProjectGenerator.cs#L852-L856). Second arg in the `SolutionProjectGenerator.Generate` appears to take in global properties.

Though I do not know what's cleaner, setting that global property or making the `t.FullPath.Contains(".sln.metaproj")` assumption.

radical:
> Though I do not know what's cleaner, ...

Setting the global property, IMHO :)

radical:
BTW, we use `$(ImportByWildcardBeforeSolution)` for controlling both `Before` and `After` target imports, is that deliberate?

cdmihai:
> BTW, we use $(ImportByWildcardBeforeSolution) for controlling both Before and After target imports, is that deliberate?

Apparently it is deliberate: http://sedodream.com/2010/10/22/MSBuildExtendingTheSolutionBuild.aspx

PR #2400, opened 2017-08-06
Copy application base directory when creating app domain for task

### LINKED ISSUE #1900: ResolveComReference task fails when executing build from child AppDomain
I've run into following problem, when trying to execute build in child app domain of main process:

```
Invoke-MSBuild : C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(2545,5): 
error MSB4061: The "ResolveComReference" task could not be instantiated from "Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". 
Could not load file or assembly 'Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
The system cannot find the file specified.
```
The issue appears to be caused by fact that build is being run from child app domain of main process (powershell.exe in my case). `ResolveComReference` (extended from `AppDomainIsolatedTaskExtension`) does not use ApplicationBase of child appdomain, but instead uses ApplicationBase of parent PowerShell process.

Log from Fuslogvw.exe (Assembly Binding Log Viewer)
```
*** Assembly Binder Log Entry  (2017-03-22 @ 22:53:23) ***

The operation failed.
Bind result: hr = 0x80070002. The system cannot find the file specified.

Assembly manager loaded from:  C:\Windows\Microsoft.NET\Framework64\v4.0.30319\clr.dll
Running under executable  C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
--- A detailed error log follows. 

=== Pre-bind state information ===
LOG: DisplayName = Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
 (Fully-specified)
LOG: Appbase = file:///C:/Windows/System32/WindowsPowerShell/v1.0/
LOG: Initial PrivatePath = NULL
LOG: Dynamic Base = NULL
LOG: Cache Base = NULL
LOG: AppName = powershell.exe
Calling assembly : (Unknown).
===
LOG: This bind starts in default load context.
LOG: No application configuration file found.
LOG: Using host configuration file: 
LOG: Using machine configuration file from C:\Windows\Microsoft.NET\Framework64\v4.0.30319\config\machine.config.
LOG: GAC Lookup was unsuccessful.
LOG: Attempting download of new URL file:///C:/Windows/System32/WindowsPowerShell/v1.0/Microsoft.Build.DLL.
LOG: Attempting download of new URL file:///C:/Windows/System32/WindowsPowerShell/v1.0/Microsoft.Build/Microsoft.Build.DLL.
LOG: Attempting download of new URL file:///C:/Windows/System32/WindowsPowerShell/v1.0/Microsoft.Build.EXE.
LOG: Attempting download of new URL file:///C:/Windows/System32/WindowsPowerShell/v1.0/Microsoft.Build/Microsoft.Build.EXE.
LOG: All probing URLs attempted and failed.
```

kovalikp:
This fixes #1900 in scenario described there. I've tested the it locally, but I was not able to write unit test for the fix. Because ApplicationBase is being overwritten by test environment variables just few lines bellow.

### DIFF
--- a/src/Shared/TaskLoader.cs
+++ b/src/Shared/TaskLoader.cs
@@ -98,6 +98,7 @@ namespace Microsoft.Build.Shared
 
                         // Apply the appdomain settings to the new appdomain before creating it
                         appDomainInfo.SetConfigurationBytes(currentAppdomainBytes);
+                        appDomainInfo.ApplicationBase = appDomainSetup.ApplicationBase;
 
                         if (BuildEnvironmentHelper.Instance.RunningTests)
                         {


radical:
This is the fix I used for mono : https://github.com/mono/msbuild/commit/4719f9a3d141bfb43df3bf91dc4f8331dd8e1866 . Where is `appDomainSetup` coming from and will it point to the msbuild assemblies' path?

AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:

AndyGerlicher:
Sorry for the really long delay on getting back to this. This is one of those that looks like the right change and everything should be fine but it actually takes quite a while to test the implications. I poked around with it for a while and it caused some `AppDomainIsolatedTask`'s to stop working correctly in VS, which is the scenario I was worried about. Especially when you get to crazy tasks like WPF/Xaml where they do a double build and create their own AppDomain within a task, things get really complicated :(. Unfortunately I don't have the bandwidth right now to debug why the failure occurred, but it is a behavior change that could likely affect someone (even though it seems like the right value to be setting). The only way you could have found this is to write an `AppDomainIsolatedTask` and update VS with locally built binaries and integrate your task into a build in various places.

Thanks again for the contribution and if you do find a way to fix this without breaking-change impact we would reconsider.

Issue #2348, opened 2017-07-21
Excessive memory allocation comparing paths in evaluation.


AndyGerlicher:
Based on developer community feedback:
https://developercommunity.visualstudio.com/content/problem/80371/visual-studio-becomes-unresponsive-for-a-few-minut.html

This customer reported issue is showing some extreme build behavior https://devdiv.visualstudio.com/DevDiv/_workitems/edit/464792

Here we can see his profile shows 9GB of allocations for normalizing paths during the build (profile is available in the developer community comments)

Name | Inc % | Inc
-- | -- | --
microsoft.build.ni!FileUtilities.NormalizePathForComparisonNoThrow | 83.2 | 9,816,500,224.000
microsoft.build.ni!FileUtilities.ComparePathsNoThrow | 83.2 | 9,816,500,224.000
microsoft.build.ni!Microsoft.Build.Internal.EngineFileUtilities+<>c__DisplayClass6_0.<GetFileSpecMatchTester>b__0(System.String) | 83.2 | 9,816,500,224.000
microsoft.build.ni!ItemFragment.MatchCount | 83.2 | 9,816,500,224.000
microsoft.build!Microsoft.Build.Evaluation.ItemExpressionFragment`2+<>c__DisplayClass12_0[System.__Canon,System.__Canon].<MatchCount>b__0(class Microsoft.Build.Evaluation.ValueFragment) | 83.0 | 9,786,655,744.000
system.core.ni!System.Linq.Enumerable.Count[System.__Canon](System.Collections.Generic.IEnumerable`1, System.Func`2) | 83.0 | 9,786,655,744.000
microsoft.build!Microsoft.Build.Evaluation.ItemExpressionFragment`2[System.__Canon,System.__Canon].MatchCount(class System.String) | 83.0 | 9,786,655,744.000
microsoft.build.ni!Microsoft.Build.Evaluation.ItemSpec`2+<>c__DisplayClass17_0[System.__Canon,System.__Canon].<MatchesItem>b__0(Microsoft.Build.Evaluation.ItemFragment) | 83.0 | 9,786,655,744.000
microsoft.build.ni!Microsoft.Build.Evaluation.ItemSpec`2+<>c__DisplayClass17_0[System.__Canon,System.__Canon].<MatchesItem>b__0(Microsoft.Build.Evaluation.ItemFragment) | 0.3 | 29,843,840
[...]

Issue #2238, opened 2017-06-21
MSBuild File globbing is slow on large directory layouts (e.g. node_modules)


cdmihai:
Issue ported from:
- developer community (**workaround available**) https://developercommunity.visualstudio.com/content/problem/50110/visual-studio-is-extremely-slow-when-node-modules.html
- internal https://devdiv.visualstudio.com/DevDiv/MSBuild/_workitems?id=447520

cdmihai:
From internal investigation:

> When there's a large number of files, the Build FileMatcher.GetFilesRecursive allocates a huge amount of strings.
> 
> Name                                                                                                                                                             Inc %              Inc
>  Type System.String                                                                                                                                               55.8    1,446,726,272
> + microsoft.build.ni![COLD] Microsoft.Build.Shared.FileUtilities.PathsEqual(System.String, System.String)                                                         27.3      708,307,520
> |+ Microsoft.Build!Microsoft.Build.Shared.FileMatcher+<>c__DisplayClass29_0.<GetFilesRecursive>b__0(class System.String)                                          27.3      708,307,520
> | + microsoft.build.ni![COLD] Microsoft.Build.Shared.FileMatcher.GetFilesRecursive(System.Collections.Generic.IList`1, RecursionState, System.String, Boolean,    27.3      708,307,520
> |  + microsoft.build.ni!FileMatcher.GetFilesRecursive                                                                                                             27.3      707,986,112
> |  |+ microsoft.build.ni!FileMatcher.GetFiles                                                                                                                     27.3      707,986,112
> |  | + microsoft.build.ni!EngineFileUtilities.GetFileList                                                                                                         27.3      707,986,112
> |  |  + microsoft.build.ni!EngineFileUtilities.GetFileListEscaped                                                                                                 27.3      707,986,112
> |  |   + microsoft.build.ni!Microsoft.Build.Evaluation.LazyItemEvaluator`4+LazyItemOperation[System.__Canon,System.__Canon,System.__Canon,System.__Canon].Apply   27.3      707,986,112
> |  |    + microsoft.build.ni!Microsoft.Build.Evaluation.LazyItemEvaluator`4+MemoizedOperation[System.__Canon,System.__Canon,System.__Canon,System.__Canon].Appl   27.3      707,986,112
[...]

PR #2401, opened 2017-08-06
Reduce Allocs

benaadams:
Remove opportunistic array allocation from `GetTaskFiles`

Remove IEnumerator and closure capture allocations from `FunctionCallExpressionNode`
Remove IEnumerator allocations from `Evaluator.ReadNewTargetElement`
Remove IEnumerator allocations from `Evaluator.PerformDepthFirstPass`

Remove IEnumerator allocations from `Expander.ExpandIntoItemsLeaveEscaped`

Reduce `string`  allocations from `FileUtilities.PathsEqual`

Reduce boxing allocations from `ProjectErrorUtilities`

Remove `StringBuilder.Clear` allocation on discarded `StringBuilder`s

Saves 20% of allocations (496MB) in a Rosyln restore (2389MB to 1893MB)

/cc @davkean @KirillOsenkov

### DIFF
--- a/src/Build/Evaluation/Conditionals/FunctionCallExpressionNode.cs
+++ b/src/Build/Evaluation/Conditionals/FunctionCallExpressionNode.cs
@@ -167,7 +167,7 @@ namespace Microsoft.Build.Evaluation
                 return null;
             }
 
-            var list = new List<string>();
+            var list = new List<string>(capacity: expandedCount);
             for (var i = 0; i < expandedCount; i++)
             {
                 var item = expanded[i];
--- a/src/Build/Evaluation/Conditionals/FunctionCallExpressionNode.cs
+++ b/src/Build/Evaluation/Conditionals/FunctionCallExpressionNode.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using Microsoft.Build.Shared;
 
 using TaskItem = Microsoft.Build.Execution.ProjectItemInstance.TaskItem;
@@ -43,9 +42,21 @@ namespace Microsoft.Build.Evaluation
                     // Expand the items and use DefaultIfEmpty in case there is nothing returned
                     // Then check if everything is not null (because the list was empty), not
                     // already loaded into the cache, and exists
-                    return ExpandArgumentAsFileList((GenericExpressionNode) _arguments[0], state)
-                        .DefaultIfEmpty()
-                        .All(i => i != null && (state.LoadedProjectsCache?.TryGet(i) != null || FileUtilities.FileOrDirectoryExistsNoThrow(i)));
+                    var list = ExpandArgumentAsFileList((GenericExpressionNode) _arguments[0], state);
+                    if (list == null)
+                    {
+                        return false;
+                    }
+
+                    foreach (var item in list)
+                    {
+                        if (item == null || !(state.LoadedProjectsCache?.TryGet(item) != null || FileUtilities.FileOrDirectoryExistsNoThrow(item)))
+                        {
+                            return false;
+                        }
+                    }
+
+                    return true;
                 }
                 catch (Exception e) when (ExceptionHandling.IsIoRelatedException(e))
                 {
@@ -137,7 +148,7 @@ namespace Microsoft.Build.Evaluation
             return expandedValue;
         }
 
-        private IEnumerable<string> ExpandArgumentAsFileList(GenericExpressionNode argumentNode, ConditionEvaluator.IConditionEvaluationState state, bool isFilePath = true)
+        private List<string> ExpandArgumentAsFileList(GenericExpressionNode argumentNode, ConditionEvaluator.IConditionEvaluationState state, bool isFilePath = true)
         {
             string argument = argumentNode.GetUnexpandedValue(state);
 
@@ -147,11 +158,30 @@ namespace Microsoft.Build.Evaluation
                 argument = FileUtilities.FixFilePath(argument);
             }
 
-            return state.ExpandIntoTaskItems(argument)
-                .Select(i =>
-                    state.EvaluationDirectory != null && !Path.IsPathRooted(i.ItemSpec)
-                        ? Path.GetFullPath(Path.Combine(state.EvaluationDirectory, i.ItemSpec))
-                        : i.ItemSpec);
+
+            IList<TaskItem> expanded = state.ExpandIntoTaskItems(argument);
+            var expandedCount = expanded.Count;
+
+            if (expandedCount == 0)
+            {
+                return null;
+            }
+
+            var list = new List<string>();
+            for (var i = 0; i < expandedCount; i++)
+            {
+                var item = expanded[i];
+                if (state.EvaluationDirectory != null && !Path.IsPathRooted(item.ItemSpec))
+                {
+                    list.Add(Path.GetFullPath(Path.Combine(state.EvaluationDirectory, item.ItemSpec)));
+                }
+                else
+                {
+                    list.Add(item.ItemSpec);
+                }
+            }
+
+            return list;
         }
 
         /// <summary>
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -791,8 +791,9 @@ namespace Microsoft.Build.Evaluation
             // Pass1: evaluate properties, load imports, and gather everything else
             PerformDepthFirstPass(_projectRootElement);
 
-            List<string> initialTargets = new List<string>(_initialTargetsList.Count);
-            for (int i = 0; i < _initialTargetsList.Count; i++)
+            var initialTargetsListCount = _initialTargetsList.Count;
+            List<string> initialTargets = new List<string>(initialTargetsListCount);
+            for (var i = 0; i < initialTargetsListCount; i++)
             {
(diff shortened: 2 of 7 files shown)


src/Build/Evaluation/Conditionals/FunctionCallExpressionNode.cs
```diff
@@ -147,11 +158,30 @@ private IEnumerable<string> ExpandArgumentAsFileList(GenericExpressionNode argum
...
+
+            IList<TaskItem> expanded = state.ExpandIntoTaskItems(argument);
+            var expandedCount = expanded.Count;
+
+            if (expandedCount == 0)
+            {
+                return null;
+            }
+
+            var list = new List<string>();
```
jaredpar: Consider being explicit about the capacity here to optimze the internal array allocations

``` C#
new List<string>(capacity: expandedCount)
```

benaadams: Done

benaadams:
test OSX Build for CoreCLR

src/Build/Evaluation/Evaluator.cs
```diff
@@ -863,16 +867,20 @@ private static ProjectTargetInstance ReadNewTargetElement(ProjectTargetElement t
...
             }
 
             // If there was no DefaultTargets attribute found in the depth first pass, 
             // use the name of the first target. If there isn't any target, don't error until build time.
+
+            var targetElementsCount = _targetElements.Count;
             if (_data.DefaultTargets == null || _data.DefaultTargets.Count == 0)
             {
-                List<string> defaultTargets = new List<string>(_targetElements.Count);
+                List<string> defaultTargets = new List<string>(targetElementsCount);
```
benaadams: Still need to check count before indexing to the array?

src/Shared/ReuseableStringBuilder.cs
```diff
@@ -309,6 +306,9 @@ internal static void Release(StringBuilder returningBuilder)
                 // So the shared builder will be "replaced".
                 if (returningBuilder.Capacity < MaxBuilderSize)
                 {
+                    // ErrorUtilities.VerifyThrow(handouts.TryRemove(returningBuilder, out dummy), "returned but not loaned");
+                    returningBuilder.Clear(); // This is free: it just sets length to zero internally
```
benaadams: Setting length to 0 can [allocate a `char[]`](https://github.com/dotnet/coreclr/blob/eb1d52cc2e2a32ccbead0dac01c99a073640ee13/src/mscorlib/shared/System/Text/StringBuilder.cs#L534-L540) if the builder has moved passed its initial extent; which is definitely case if its being discarded

src/Build/Evaluation/Evaluator.cs
benaadams: Changed

src/Shared/FileUtilities.cs
```diff
@@ -1098,5 +1165,24 @@ internal static StreamReader OpenRead(string path, Encoding encoding = null, boo
                 return new StreamReader(fileStream, encoding, detectEncodingFromByteOrderMarks);
             }
         }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
benaadams: Added

src/Shared/ReuseableStringBuilder.cs
benaadams: Also why do work if you are just throwing it away 😄

src/Build/Evaluation/Evaluator.cs
```diff
@@ -962,9 +969,10 @@ private void PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
             if (!Traits.Instance.EscapeHatches.IgnoreTreatAsLocalProperty)
             {
                 IList<string> globalPropertiesToTreatAsLocals = _expander.ExpandIntoStringListLeaveEscaped(currentProjectOrImport.TreatAsLocalProperty, ExpanderOptions.ExpandProperties, currentProjectOrImport.TreatAsLocalPropertyLocation);
-
-                foreach (string propertyName in globalPropertiesToTreatAsLocals)
+                var globalPropertiesToTreatAsLocalsCount = globalPropertiesToTreatAsLocals.Count;
```
davkean: Add a comment here to avoid someone changing this to a foreach - is this to avoid the boxing the enumerator?

src/Shared/FileUtilities.cs
```diff
@@ -1054,23 +1055,89 @@ internal static string WithTrailingSlash(this string s)
...
+
+            for (var i = 0; i <= endA; i++)
+            {
+                var charA = (uint)path1[i];
+                var charB = (uint)path2[i];
+
+                if ((charA | charB) > 0x7F)
+                {
+                    // Non-ascii chars move to non fast path
+                    return PathsEqualNonAscii(path1, path2, i, endA - i + 1);
```
davkean: Ha, was looking to write this one myself.

src/Build/Evaluation/Evaluator.cs
benaadams: Yeah via `IList.GetEnumerator`; solving IEnumerable boxing would probably solve 75% of GC problems in .NET

Box `foreach` to stack or something...

src/Shared/ReuseableStringBuilder.cs
davkean: 👍

src/Build/Evaluation/Evaluator.cs
benaadams: Oh, caching the count is so its not an interface evaluation per loop iteration

davkean: Yeah, looked at that and thought - stupidly thought hey JIT will worry about that, then realized it's not an array.

benaadams: Won't even inline since its an interface 😢 

Added comments

src/Shared/ReuseableStringBuilder.cs
benaadams: Updated commit message

src/Shared/FileUtilities.cs
```diff
@@ -1098,5 +1165,27 @@ internal static StreamReader OpenRead(string path, Encoding encoding = null, boo
                 return new StreamReader(fileStream, encoding, detectEncodingFromByteOrderMarks);
             }
         }
+
+        // Method is simple set of function calls and may inline;
+        // we don't want it inlining into the tight loop that calls it as an exit case,
+        // so mark as non-inlining
```
rainersigwald: 👍

src/Shared/ReuseableStringBuilder.cs
rainersigwald: If you get a chance before we merge, can you also delete the misleading `//this is free` comment? _I_ was clearly misled by it!

benaadams:
Changed comment; only found it through tracing. I always thought it was free also

src/Shared/FileUtilities.cs
cdmihai: This might be a problem for other languages, where there is a high chance to get unicode characters in paths (e.g. user names), which would make globbing deep directories faster for English users, but keep it slow for non-English ones (#2238 suggests PathsEqual might be a big perf hit when globbing over node_modules).

Is there any issue with doing the expensive check just for the characters in the current iteration (using Char.ToUpper to get culture aware case insensitivity) rather than falling back to the expensive check for the rest of the strings?

cdmihai:
This PR might also fix the following issue regarding freezing VS when globbing deep directories: #2238

src/Shared/FileUtilities.cs
benaadams: Probably could work; however not sure what [Not fully implemented](https://github.com/dotnet/coreclr/blob/c7a10a52f959514428afe2ab6f4bade3ce1dcdbf/src/mscorlib/shared/System/Char.cs#L403-L410) means

benaadams: Also might be a problem for chars that can't be represented by a single UTF16 char? (i.e. would be multi char)

benaadams: Could raise a followup issue for non-ascii paths?

cdmihai: Sounds fine with me. I'll open a followup issue when the PR goes in.

davkean:
@benaadams Does this address: https://github.com/Microsoft/msbuild/issues/2348?

benaadams:
No, doesn't use this path. I did consider addressing it as that looks way more allocatey, but didn't have any data on it - so didn't now how much it was used.

davkean:
@benaadams Looks like this was addressed in: https://github.com/Microsoft/msbuild/pull/2326 anyway.

benaadams:
K, however `FileUtilities.ComparePathsNoThrow` is very allocatey

davkean:
It is, but I'm focusing on the big ones at the moment.

cdmihai:
@davkean 
#2326 (self referencing update operations) fixes #2314, not #2238. While it avoids doing needles path comparisons, it does not solve path equality for all code paths (i.e. when item operations are not self referencing update operations).

rainersigwald:
Thanks @benaadams!

davkean:
#2348 appears to be caused by the linked item regression as future traces on Preview 7, show that it no longer appears. So while FileUtilities.ComparePathsNoThrow is still very allocatey, it's not showing up in the traces that I'm looking at.

PR #2404, opened 2017-08-07
Allow using the TargetFrameworkRootPath property from MSBuild.exe.config for setup reference assemblies folder.

mfilippov:
This fix allows put all reference assemblies into a custom folder or NuGet package and use it in build time on CI.


mfilippov:
Also, it allowed pass path with reference assemblies from mono for .NET Core build with full framework target on Unix.

mfilippov:
@rainersigwald ping

PR #2405, opened 2017-08-07
Fix build when FEATURE_WIN32_REGISTRY disabled.

mfilippov:
This allows block registry request on Windows for better toolset isolation.

### DIFF
--- a/src/Build/Evaluation/IntrinsicFunctions.cs
+++ b/src/Build/Evaluation/IntrinsicFunctions.cs
@@ -133,7 +133,7 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
-        /// Perform a bitwise OR on the first and second (first | second) 
+        /// Perform a bitwise OR on the first and second (first | second)
         /// </summary>
         internal static int BitwiseOr(int first, int second)
         {
@@ -141,7 +141,7 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
-        /// Perform a bitwise AND on the first and second (first &amp; second) 
+        /// Perform a bitwise AND on the first and second (first &amp; second)
         /// </summary>
         internal static int BitwiseAnd(int first, int second)
         {
@@ -149,7 +149,7 @@ namespace Microsoft.Build.Evaluation
         }
 
         /// <summary>
-        /// Perform a bitwise XOR on the first and second (first ^ second) 
+        /// Perform a bitwise XOR on the first and second (first ^ second)
         /// </summary>
         internal static int BitwiseXor(int first, int second)
         {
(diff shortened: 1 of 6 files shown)


src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -258,6 +258,32 @@ internal static object GetRegistryValueFromView(string keyName, string valueName
...
+        {
+            return null; // FEATURE_WIN32_REGISTRY is off, need to mock the function names to let scrips call these property functions and get NULLs rather than fail with errors
+        }
+
+        /// <summary>
+        /// Get the value of the registry key from one of the RegistryView's specified
+        /// </summary>
+        internal static object GetRegistryValueFromView(string keyName, string valueName, object defaultValue, params object[] views)
+        {
+            return null; // FEATURE_WIN32_REGISTRY is off, need to mock the function names to let scrips call these property functions and get NULLs rather than fail with errors
```
rainersigwald: I think this should be `return defaultValue;` instead; that's what the registry-enabled code would do if the key didn't exist, right?

hypersw:
@rainersigwald: I made this change because we wanted to make our build process less dependent on differences of local machine configurations and have it more xcopy-configurable. Thanks to the new `msbuild.exe.config` toolsets feature, it's now possible to define paths to the tools with global properties, either to the portable locations or to select tools from the local machine. This works, more or less, with the standard .targets files shipped for building C#/XAML/C++/etc, though there're problems with unexpected local machine settings leaking into the build process.

Registry access is one such spot, and the major one. Even Program Files locations lookup often starts with a Registry read. Some .targets would use Registry either as the first chance to look for the tool, or as the last chance if a global property is not set. Either way, this breaks the isolation in a very covert fashion.

So I wished to try the MSBuild conditional compilation symbols for turning off Registry access. This promised to solve the Registry problems. However, the stock implementation which errors on any Registry read proved useless for our scenario — it's not possible to tell the whole lot of the standard .targets files to keep off the Registry (this includes targeting older .NET FX versions and C++ builds ver vc100 and later, not just the latest bits), so no build would ever finish successfully.

But there's the Linux example for which Registry reads would act as if the Registry is empty, and it works pretty good for builds. We wanted to have the same for Windows in this mode — build scripts just think they got only what is defined in the global properties, nothing in the Registry, and that allows for well-isolated builds. We have already successfully used such customized MSBuild packages for our builds.

Are there scenarios for building MSBuild with errors on Registry access? It seems that not many of the standard .targets would be usable that way. Moreover, when we started playing with these conditional compilation flags, the code wouldn't even compile for some of them because the conditioned-out code were out of sync, so I'm not sure they're much used at all.

mfilippov:
@rainersigwald defaultValue fixed

mfilippov:
@rainersigwald ping

mfilippov:
@rainersigwald @AndyGerlicher ping

AndreyAkinshin:
@rainersigwald @AndyGerlicher any updates about this pull request?
It will be really nice to have it in the master branch. I already checked this PR as a patch locally; everything works great. Do you have any blockers which prevent merging?

AndyGerlicher:
@dotnet-bot test this please

PR #2394, opened 2017-08-03
Enable tests in (opt-in) Mono CI

rainersigwald:

PR #2407, opened 2017-08-07
Build known target in MonoTest CI

rainersigwald:
I messed up in #2394 and didn't notice because I expected the MonoTest config to fail.

It did, but for the wrong reason: it wasn't building correctly and bailed out immediately.

That's because `cibuild.sh` didn't recognize the `MonoTest` runtime. Because it doesn't exist!

Fixing this by redundantly specifying `--target Mono` in mono flavor jobs. Since the last argument wins, this fixes the specification while keeping the code closer to what we want its final form to be.


rainersigwald:
@dotnet-bot test ci please

rainersigwald:

PR #2413, opened 2017-08-09
Revert GenAPI Package Update

AndyGerlicher:
Resolves a build error during GenAPI caused by the copy of
System.IO.FileSystem.dll in the new GenAPI package being bad (it's not
signed). Reverting back to the old version.


akoeplinger:
@AndyGerlicher could you please file an issue on https://github.com/dotnet/buildtools ? Now that you mention it I remember hitting the same bug when I was working with GenAPI and had to disable signing verification on the machine.

AndyGerlicher:
Still don't get why I wasn't hitting this on my desktop or why the CI didn't hit it. I definitely don't have strong name hijack installed. I filed the bug though hopefully that gets fixed.

PR #2414, opened 2017-08-09
[Do no merge] Implement new /restore argument which runs the Restore target prior to building

jeffkl:
This change introduces a new `/restore` argument which will build the `Restore` target with a different global property set prior to building the actual targets.

1. If the user specifies `/t:Restore /restore`, restore will run twice
2. `InitialTargets` will run multiple times
3. There is currently no property that indicates to target authors that a restore is being executed rather than a build

Unit tests will come after we've discussed this.


jeffkl:

jeffkl:
Sample project that creates an import during restore which will fail currently:
```xml

  <PropertyGroup>
    <RestoreFirstProps>restoreFirst.props</RestoreFirstProps>
  </PropertyGroup>
  
  <Import Project="$(RestoreFirstProps)" Condition="Exists($(RestoreFirstProps))"/>

  <Target Name="Build">
    <Error Text="PropertyA does not have a value defined" Condition=" '$(PropertyA)' == '' " />
    <Message Text="PropertyA's value is &quot;$(PropertyA)&quot;" />
    
    <Delete Files="$(RestoreFirstProps)" Condition="Exists($(RestoreFirstProps))" />
  </Target>

  <Target Name="Restore">
    <CreatePropsFile FilePath="$(RestoreFirstProps)" PropertyName="PropertyA" PropertyValue="Some value" />
  </Target>
</Project>
```

Output if you don't run restore first:
```
> msbuild restoreFirst.proj
Microsoft (R) Build Engine version 15.3.409.57025 for .NET Framework

Build started 8/10/2017 2:03:48 PM.
Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" on node 1 (default targets).
C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj(11,5): error : PropertyA does not have a value defined
Done Building Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" (default targets) -- FAILED.
```

If you run `MSBuild.exe /t:Restore` and then another build, everything works.  If you run `MSBuild.exe /t:Restore;Build`, it also fails:

```
> msbuild restoreFirst.proj /t:Restore;Build
Microsoft (R) Build Engine version 15.3.409.57025 for .NET Framework

Build started 8/10/2017 2:04:50 PM.
Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" on node 1 (Restore;Build target(s)).
Restore:
  Successfully created properties file 'restoreFirst.props'
C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj(11,5): error : PropertyA does not have a value defined
Done Building Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" (Restore;Build target(s)) -- FAILED.
```

This is because the project is not re-evaluated.

And finally, if you use the new command-line argument:

```
> MSBuild.exe restoreFirst.proj /restore
Microsoft (R) Build Engine version 15.5.67.25563 for .NET Framework

Build started 8/10/2017 2:05:35 PM.
Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" on node 1 (Restore target(s)).
Restore:
  Successfully created properties file 'restoreFirst.props'
Done Building Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" (Restore target(s)).

Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" on node 1 (default targets).
Build:
  PropertyA's value is "Some value"
  Deleting file "restoreFirst.props".
Done Building Project "C:\Users\jeffkl\Desktop\Stuff\restoreFirst.proj" (default targets).

```

jeffkl:

src/MSBuild/CommandLineSwitches.cs
```diff
@@ -103,6 +103,7 @@ internal enum ParameterizedSwitch
             WarningsAsErrors,
             WarningsAsMessages,
             BinaryLogger,
+            Restore,
```
cdmihai: Does Restore have any parameters? If it doesn't, shouldn't it go in the ParameterslessSwitch enum instead? Haven't read to code to understand the behavioural effects.

jeffkl: It's technically `true|false` like `/nodeReuse`.  This allows you to have the environment variable set but still specify `/restore:false`.  `/restore` is mapped to `/restore:true`

cdmihai: My bad, it has a bool parameter. Ignore comment.

---
In reply to: [136662597](https://github.com/Microsoft/msbuild/pull/2414#discussion_r136662597) [](ancestors = 136662597)

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1628,6 +1628,12 @@ private void ReportResultsToSubmission(BuildResult result)
 
                     _overallBuildSuccess = _overallBuildSuccess && (_buildSubmissions[result.SubmissionId].BuildResult.OverallResult == BuildResultCode.Success);
 
+                    if (submission.BuildRequestData != null && submission.BuildRequestData.Flags.HasFlag(BuildRequestDataFlags.ClearProjectRootElementCacheAfterBuild))
+                    {
+                        // Reset the project root element cache if specified which ensures that projects will be re-loaded from disk
+                        _buildParameters?.ProjectRootElementCache?.Clear();
```
cdmihai: Wonder if it would be better to also clear the ConfigCache and ResultsCache at this point (similar to what BuildManager.ResetCaches is doing). Anything held in the caches after restore completes is just memory waste.

cdmihai: Out of proc nodes have their own copy of the project root element cache. So resetting just the main node will not clear out the nodes, which would still lead to the stale cache issue :(

Either the build manager needs to shut down all nodes, or you need to implement another packet that tells the nodes to reset their caches.

cdmihai: Apparently out of proc nodes setup the cache to [reload from disk](https://github.com/Microsoft/msbuild/blob/master/src/Build/BackEnd/Node/OutOfProcNode.cs#L180). So if restore changes any existing imports, the nodes should detect the change and replace the cache entries with updated values.

PR #2415, opened 2017-08-10
Escape or avoid escaping paths in tests

rainersigwald:
Tests were failing on one of our CI environments because they were
built from a path containing a `@` character. As long as there were no
special characters in the paths used for the tests, the tests weren't
sensitive to whether the expected values were escaped properly.

Tested with

```
TMPDIR=/Users/raines/src/msbuild/tmp@ ./cibuild.sh --scope Tested
```


rainersigwald:
Example of failure this should fix: https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_OSX_CoreCLR_prtest/239/testReport/

PR #2437, opened 2017-08-16
Reduce allocations in ExpandExpressionCaptureIntoStringBuilder

davkean:
This reduces allocations by 3% by avoiding the Join by slightly increasing allocations in Append.

Before:
image

After:
image

Scenario is attempting to mimic a solution wide design-time build from here: https://github.com/dotnet/project-system/issues/2712.


davkean:
I'll probably look at trying to better "guess" up-front capabilities - to avoid the resizing that you can see in AppendHelper(String) and Append(char).

davkean:
Hmm, looks like I broke a lot of tests, investigating...

davkean:
Tests are passing now.

PR #2446, opened 2017-08-16
TaskItemSpecFilenameComparer cleanup

rainersigwald:
Avoid allocating strings for filenames, when we can take a slice of the `ItemSpec` instead.

Also some general style cleanup.

### DIFF
--- a/src/Tasks/AssemblyDependency/TaskItemSpecFilenameComparer.cs
+++ b/src/Tasks/AssemblyDependency/TaskItemSpecFilenameComparer.cs
@@ -2,10 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.IO;
 using System.Collections;
-using Microsoft.Build.Framework;
 using System.Collections.Generic;
+using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 
 namespace Microsoft.Build.Tasks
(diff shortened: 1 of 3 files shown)


davkean:
One day I'm going to write some allocation-free Path manipulation that returns a "segment" (think ArraySegment<T>) for strings...

src/Tasks/AssemblyDependency/TaskItemSpecFilenameComparer.cs
```diff
@@ -47,21 +47,38 @@ public int Compare(object o1, object o2)
...
+            {
+                yFilenameStart = 0;
+            }
 
-            int fileComparison = String.Compare(f1, f2, StringComparison.OrdinalIgnoreCase);
+            int fileComparison = string.Compare(xItemSpec,
+                xFilenameStart,
+                yItemSpec,
+                yFilenameStart,
+                int.MaxValue, // all characters after the start index
```
davkean: I never knew you could do that and I worked on the BCL. :)

rainersigwald: I just learned!

1. Assume the API is just `string1, start1, string2, start2, comparisonType`
2. Get kinda baffled by the red squiggle.
2. Oh, it requires a length.
3. Write code to get the length of one side and use that.
4. Get annoyed at having to do that.
5. Look at the implementation and discover that it already does match-portion length extraction and [takes a minimum](https://github.com/dotnet/coreclr/blob/720800cb23024f36c3f281fb586107e028b14b75/src/mscorlib/src/System/String.Comparison.cs#L612-L613) with that and this param.
5. Rejoice!

PR #2454, opened 2017-08-18
Check only the major version when finding VS

### LINKED ISSUE #2369: msbuild nuget package unable to open vs2017 csproj files
I'm referencing the nuget packages `Microsoft.Build` and `Microsoft.Build.Tasks.Core`. It doesn't matter if I chose stable or prerelease, both versions run into the same error.

I'm trying to open a csproj file for evaluation:
```csharp
var projectFile = @"C:\projects\test\test.csproj";
var project = new Microsoft.Build.Evaluation.Project(projectFile);
var projectName = Path.GetFileNameWithoutExtension(projectFile);
var outputFile = project.GetPropertyValue("TargetPath");
var outputName = Path.GetFileName(outputFile);
```
I'm getting `Microsoft.Build.Exceptions.InvalidProjectFileException: 'The tools version "15.0" is unrecognized. Available tools versions are "12.0", "14.0", "2.0", "3.5", "4.0".'`

When searching the web for this error message I'm only coming up with things related to the VS 2017 installation and how to look up the msbuild packaged with it; unfortunately this doesn't help me with my instance of the exception because I'm invoking msbuild as a library and not as a process.

Do I have to tell the nuget assemblies where the VS 2017 installation is? How do I do this? (I was assuming the nuget assemblies can work stand alone, but if a VS installation is required that works too, it's just not discoverable for me what to do here.)

rainersigwald:
Fixes #2369 by relaxing the check to find any Visual Studio 15.*
install. This is required because Visual Studio 2017 Update 3 reports as
15.3.something, which wouldn't be found by the old code. Before Visual
Studio 15.2, a bug in VS setup caused it to report 15.0 (erroneously,
but compatibly with the old MSBuild behavior).

### DIFF
--- a/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
+++ b/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
@@ -227,15 +227,17 @@ namespace Microsoft.Build.Engine.UnitTests
             }
         }
 
-        [Fact]
+        [Theory]
         [Trait("Category", "nonlinuxtests")]
         [Trait("Category", "nonosxtests")]
-        public void BuildEnvironmentDetectsVisualStudioFromSetupInstance()
+        [InlineData("15.0")]
+        [InlineData("15.3")]
+        public void BuildEnvironmentDetectsVisualStudioFromSetupInstance(string visualStudioVersion)
         {
             using (var env = new EmptyVSEnviroment())
             {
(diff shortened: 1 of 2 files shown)


rainersigwald:
Confirmed that on a machine with 15.2 and 15.3 installed, I get bad toolset expansion (4.0) on released bits and correct expansion (installed VS) with this change.

rainersigwald:

(looks like a corrupt sync, somehow?)

PR #2459, opened 2017-08-21
Avoid allocating char[] in GetEnvironmentVariables

### LINKED ISSUE #2402: Large allocation via GetEnvironmentVariables
`GetEnvironmentVariables` allocates a lot via new `strings` and `GetEnvironmentCharArray` 

e.g. for a Rosyln restore:

Could these be cached rather than constantly looked up and reconstructed; with only deltas applied and removed?

/cc @davkean

rainersigwald:
Fixes #2402 by walking the OS environment block return via pointer math,
rather than first copying it to a `char[]` and walking that with safe
code.

### DIFF
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -184,6 +184,7 @@ namespace Microsoft.Build.Internal
                     {
                         pStringBlockEnd++;
                     }
+                    long stringBlockLength = pStringBlockEnd - pStringBlock;
 
                     // Copy strings out, parsing into pairs and inserting into the table.
                     // The first few environment variable entries start with an '='!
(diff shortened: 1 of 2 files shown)


src/Shared/CommunicationsUtilities.cs
```diff
@@ -154,55 +154,6 @@ private static int FileVersionHash
         [DllImport("kernel32.dll", EntryPoint = "RtlMoveMemory")]
         internal static unsafe extern void CopyMemory(char* destination, char* source, uint length);
```
davkean: Do you call this anymore?

davkean:
Looks good.

src/Shared/CommunicationsUtilities.cs
rainersigwald: Indeed I don't. And that means I can remove the whole #define. That's fulfilling, right before going on leave!

PR #2467, opened 2017-08-24
Single-quote Exec command line on *nix

rainersigwald:
Triple-quotes seem to have been expanded as 
`{empty quoted string}{single-quoted string}{empty quoted string}`
rather than special
escaping. That confused an updated version of System.Diagnostic.Process,
so simplifying the quoting.

See https://github.com/dotnet/corefx/issues/23496#issuecomment-324519381 for context.

### DIFF
--- a/src/Tasks/Exec.cs
+++ b/src/Tasks/Exec.cs
@@ -606,10 +606,10 @@ namespace Microsoft.Build.Tasks
             if (NativeMethodsShared.IsUnixLike)
             {
                 commandLine.AppendSwitch("-c");
-                commandLine.AppendTextUnquoted(" \"\"\"");
+                commandLine.AppendTextUnquoted(" \"");
                 commandLine.AppendTextUnquoted("export LANG=en_US.UTF-8; export LC_ALL=en_US.UTF-8; . ");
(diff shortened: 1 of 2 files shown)


eerhardt:
Thanks @rainersigwald for the quick fix.

@nguerrera @dsplaisted - if we take the MSBuild build with this fix into dotnet/cli, we should be able to enable `Exec` on .NET Core 2.1 again.  I'm assuming it will be faster to get a new MSBuild build, than it will be to wait for the fix in CoreFX.

nguerrera:
Works for me. Thanks!

rainersigwald:
@eerhardt @nguerrera Are y'all ok with taking an MSBuild 15.5? I asked @dsplaisted about that yesterday and wasn't sure if it was ok to do so.

(No breaking changes, but some changes that are slightly higher risk. We want to move forward but don't want to drag y'all around unnecessarily.)

nguerrera:
This would go to master which is actually post 15.5 so no worries there.

rainersigwald:
👍 I'll build + publish a package as soon as someone else on my team gets to work and approves this.

PR #2458, opened 2017-08-21
Get TargetFrameworks for ProjectReferences in parallel

rainersigwald:
Fixes #1785 with backward compatibility by accepting a) the current return format of `GetTargetFrameworkProperties`, or b) a new format that consists of a single item with metadata, like

```xml
<ProjectBuildInstructions Include="$(MSBuildThisFileFullPath)">
  <DesiredTargetFrameworkProperties>TargetFramework=$(NearestTargetFramework)</DesiredTargetFrameworkProperties>
  <HasSingleTargetFramework>$(_HasSingleTargetFramework)</HasSingleTargetFramework>
  <IsRidAgnostic>$(_IsRidAgnostic)</IsRidAgnostic>
</ProjectBuildInstructions>
```

I had to do some tricks to keep backward compatibility with the old semicolon-delimited return value, but I think that's a hard requirement, because it'll be possible to use the MSBuild 15.5 engine with SDK 2.0.0, which will still emit the old value.

Note that this doesn't directly improve things much, except in the case of building a single project with many ProjectReferences in a multiproc MSBuild environment. But I think this is a good foundation for building a fix for #1276 on top of.

PR #2472, opened 2017-08-24
Identify inner build in consuming project

rainersigwald:
This is a proposed ProjectReference protocol change and the required new target (+ dependency on NuGet.Frameworks).

This allows solving the double-evaluation-of-ProjectReferences problem by ensuring that a) every referencing project calls into the same cached `GetTargetFrameworks`-with-no-TF-set evaluation of the reference, and b) single-targeted projects will use the same evaluation for their build.

It is also cacheable, but that's not done here.

Builds on #2458, since that simplified the ProjectReference flow.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1522,8 +1522,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     ======================================================================================
   -->
-  <Target Name="_GetProjectReferenceTargetFrameworkProperties"
-          Outputs="%(_MSBuildProjectReferenceExistent.Identity)">
+  <Target Name="_GetProjectReferenceTargetFrameworkProperties">
     <!--
       Honor SkipGetTargetFrameworkProperties=true metadata on project references
       to mean that the project reference is known not to target multiple frameworks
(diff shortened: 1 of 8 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1555,6 +1541,60 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    Builds the GetTargetFrameworkProperties target of all existent project references,
+    passing $(TargetFrameworkMoniker) as $(ReferringTargetFramework) and sets the
+    SetTargetFramework metadata of the project reference to the value that is returned.
+
+    This allows a cross-targeting project to select how it should be configured to
+    build against the most appropriate target for the referring target framework.
+
+    ======================================================================================
+  -->
+  <Target Name="_GetProjectReferenceTargetFrameworkProperties">
```
rainersigwald: This is no longer needed if the above multitargeting handling is comprehensive. If we go with the try-the-new-way-and-don't-fail-if-it-doesn't-work approach, it'd serve as the compat fallback.

rainersigwald:
Left to do:

- [ ] Figure out if the breaking change is ok (or implement `<MSBuild Dont
- [ ] Figure out TF-compat-oracle story
- [ ] Productize error messages
- [ ] Compat testing on cli
- [ ] Perf

cdmihai:
@dotnet-bot test this please

AndyGerlicher:
Closing in favor of #2595

PR #2508, opened 2017-09-11
Update solution to match project configs

davkean:
Solution is out of sync of project configs.

VS wanted to keep changing the solution to below, and got tired of undoing the change.


AndyGerlicher:
I don't see any changes in this?

davkean:
Oh weird - looks like the changes in master already have it, and when I merged the conflict this resulted in zero changes.

PR #2513, opened 2017-09-13
Major performance improvements (.NET Core + Linux)

rartemev:
Those following changes improve msbuild performance at least in 2 times. The build time in our case decreased **from 31 minutes to 13 minutes**.

In CoreCLR project we faced with major performance problem in Linux environment related to msbuild where building our C# tests (more then 10000 stand-alone C#/IL projects) takes a lot of time, much more than on Windows.
I started to investigate and figured out that most of time root msbuild process spends spinning trying to acquire some locks. The perf report shows that fact:

```
  59.63%  dotnet   libcoreclr.so                         [.] Object::EnterObjMonitorHelperSpin
  18.21%  dotnet   libcoreclr.so                         [.] AwareLock::Contention
   1.05%  dotnet   libcoreclr.so                         [.] LookupMapBase::GetValueFromCompressedMap
   0.85%  dotnet   ld-2.23.so                            [.] __tls_get_addr
   0.77%  dotnet   libcoreclr.so                         [.] CMiniMd::vSearchTable
   0.71%  dotnet   libcoreclr.so                         [.] JIT_MonReliableEnter_Portable
   0.68%  dotnet   libcoreclr.so                         [.] StgBlobPoolReadOnly::GetBlob
   0.63%  dotnet   [kernel.kallsyms]                     [k] update_blocked_averages
   0.62%  dotnet   libcoreclr.so                         [.] CMiniMdBase::SearchTableForMultipleRows
   0.55%  dotnet   libcoreclr.so                         [.] ThreadPoolNative::NotifyRequestComplete
   0.49%  dotnet   libcoreclr.so                         [.] MetaDataImport::Enum
```
In total it almost 80% of CPU time. I got the exact methods that spend most of time, here they are:
```
LoadedType Microsoft.Build.Shared.TypeLoader$AssemblyInfoToLoadedTypes::GetLoadedTypeByTypeName(string)
LoadedType Microsoft.Build.Shared.TypeLoader::GetLoadedType(class System.Collections.Concurrent.ConcurrentDictionary<Func<System.Type, object ,bool>, System.Collections.Concurrent.ConcurrentDictionary<AssemblyLoadInfo, TypeLoader$AssemblyInfoToLoadedTypes>>, string, AssemblyLoadInfo)
```

In those methods there are bunch of locks which led to such problem, but there also were different methods that were using `lock` statement and `Queue<T>`.

To solve the main problem I used `ConcurrentDicrionary` that spreads workload over the number of different locks inside Dictionary.
To solve the problem with Queues I replaced all those cases with `ConcurrentQueue` which in theory may work even without any lock. I also fixed some cases of inappropriate use of locks.

With my changes perf tool shows a picture like:
```
   6.81%  dotnet   libcoreclr.so                         [.] WKS::enter_spin_lock
   4.29%  dotnet   libcoreclr.so                         [.] AwareLock::Contention
   1.85%  dotnet   libcoreclr.so                         [.] Module::FindDomainFile
   1.67%  dotnet   libcoreclr.so                         [.] JIT_GetRuntimeType
   1.64%  dotnet   libcoreclr.so                         [.] Object::EnterObjMonitorHelperSpin
   1.61%  dotnet   libcoreclr.so                         [.] CMiniMd::vSearchTable
   1.55%  dotnet   libcoreclr.so                         [.] MDInternalRO::EnumInit
   1.48%  dotnet   libcoreclr.so                         [.] MethodTable::GetModule
   1.48%  dotnet   libcoreclr.so                         [.] ModuleHandle::GetMetadataImport
   1.22%  dotnet   ld-2.23.so                            [.] __tls_get_addr
   1.17%  dotnet   libcoreclr.so                         [.] CMiniMd::vSearchTableNotGreater
```
So you can see both  `AwareLock::Contention` and `Object::EnterObjMonitorHelperSpin` take just (4.29% + 1.64%) 5.93% of CPU time instead of ~80%. The method `WKS::enter_spin_lock` is the part of GC and its top position might be explained by the memory consumption.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Text;
 using System.Threading;
 using System.Diagnostics;
@@ -92,7 +93,7 @@ namespace Microsoft.Build.BackEnd
         /// Operations on this queue must be synchronized since it is accessible by multiple threads.
         /// Use a lock on the packetQueue itself.
         /// </remarks>
-        private Queue<INodePacket> _packetQueue;
+        private ConcurrentQueue<INodePacket> _packetQueue;
         #endregion
 
         #region Constructors and Factories
@@ -332,11 +333,8 @@ namespace Microsoft.Build.BackEnd
             ErrorUtilities.VerifyThrow(null != _packetQueue, "packetQueue is null");
             ErrorUtilities.VerifyThrow(null != _packetAvailable, "packetAvailable is null");
 
-            lock (_packetQueue)
-            {
-                _packetQueue.Enqueue(packet);
-                _packetAvailable.Set();
-            }
+            _packetQueue.Enqueue(packet);
+            _packetAvailable.Set();
         }
 
         /// <summary>
@@ -367,7 +365,7 @@ namespace Microsoft.Build.BackEnd
                 _packetPump.Name = "InProc Endpoint Packet Pump";
                 _packetAvailable = new AutoResetEvent(false);
                 _terminatePacketPump = new AutoResetEvent(false);
-                _packetQueue = new Queue<INodePacket>();
+                _packetQueue = new ConcurrentQueue<INodePacket>();
 #if FEATURE_THREAD_CULTURE
                 _packetPump.CurrentCulture = _componentHost.BuildParameters.Culture;
                 _packetPump.CurrentUICulture = _componentHost.BuildParameters.UICulture;
@@ -432,30 +430,10 @@ namespace Microsoft.Build.BackEnd
                             break;
                         case 1:
                             {
-                                // Figure out how many packets are currently in the queue to process.  We
-                                // will only process the number in the queue at this moment so that we don't
-                                // get into a condition where the sending thread continues to pump packets in
-                                // as fast as we can send them, potentially starving us from detecting a
-                                // terminate event.
-                                int packets = 0;
-                                lock (_packetQueue)
+                                INodePacket packet;
+                                while (_packetQueue.TryDequeue(out packet))
                                 {
-                                    packets = _packetQueue.Count;
-                                }
-
-                                while (packets > 0)
-                                {
-                                    // Grab the first packet in the queue.
-                                    INodePacket packet = null;
-
-                                    lock (_packetQueue)
-                                    {
-                                        ErrorUtilities.VerifyThrow(_packetQueue.Count > 0, "Packet Queue count is zero.");
-                                        packet = _packetQueue.Dequeue();
-                                    }
-
                                     _peerEndpoint._packetFactory.RoutePacket(0, packet);
-                                    --packets;
                                 }
                             }
 
--- a/src/MSBuildTaskHost/Concurrent/ConcurrentDictionary.cs
+++ b/src/MSBuildTaskHost/Concurrent/ConcurrentDictionary.cs
@@ -1,0 +1,534 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.Build.Shared.Concurrent
+{
+    // The following class is back-ported from .NET 4.X CoreFX library because
+    // MSBuildTaskHost requires 3.5 .NET Framework. Only GetOrAdd method kept.
+    internal class ConcurrentDictionary<TKey, TValue>
+    {
+        /// <summary>
+        /// Tables that hold the internal state of the ConcurrentDictionary
+        ///
+        /// Wrapping the three tables in a single object allows us to atomically
+        /// replace all tables at once.
+        /// </summary>
+        private sealed class Tables
+        {
+            internal readonly Node[] _buckets; // A singly-linked list for each bucket.
+            internal readonly object[] _locks; // A set of locks, each guarding a section of the table.
+            internal volatile int[] _countPerLock; // The number of elements guarded by each lock.
+
+            internal Tables(Node[] buckets, object[] locks, int[] countPerLock)
+            {
+                _buckets = buckets;
+                _locks = locks;
+                _countPerLock = countPerLock;
+            }
+        }
+
+        private volatile Tables _tables; // Internal tables of the dictionary
+        private IEqualityComparer<TKey> _comparer; // Key equality comparer
+        private readonly bool _growLockArray; // Whether to dynamically increase the size of the striped lock
+        private int _budget; // The maximum number of elements per lock before a resize operation is triggered
+
+        // The default capacity, i.e. the initial # of buckets. When choosing this value, we are making
+        // a trade-off between the size of a very small dictionary, and the number of resizes when
+        // constructing a large dictionary. Also, the capacity should not be divisible by a small prime.
+        private const int DefaultCapacity = 31;
+
+        // The maximum size of the striped lock that will not be exceeded when locks are automatically
+        // added as the dictionary grows. However, the user is allowed to exceed this limit by passing
+        // a concurrency level larger than MaxLockNumber into the constructor.
+        private const int MaxLockNumber = 1024;
+
+        // Whether TValue is a type that can be written atomically (i.e., with no danger of torn reads)
+        private static readonly bool s_isValueWriteAtomic = IsValueWriteAtomic();
+
+        /// <summary>
+        /// Determines whether type TValue can be written atomically
+        /// </summary>
+        private static bool IsValueWriteAtomic()
+        {
+            //
+            // Section 12.6.6 of ECMA CLI explains which types can be read and written atomically without
+            // the risk of tearing.
+            //
+            // See http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-335.pdf
+            //
+            Type valueType = typeof(TValue);
+            if (!valueType.IsValueType)
+            {
+                return true;
+            }
+
+            switch (Type.GetTypeCode(valueType))
+            {
+                case TypeCode.Boolean:
+                case TypeCode.Byte:
+                case TypeCode.Char:
+                case TypeCode.Int16:
+                case TypeCode.Int32:
+                case TypeCode.SByte:
+                case TypeCode.Single:
+                case TypeCode.UInt16:
+                case TypeCode.UInt32:
+                    return true;
+                case TypeCode.Int64:
+                case TypeCode.Double:
+                case TypeCode.UInt64:
+                    return IntPtr.Size == 8;
+                default:
+                    return false;
+            }
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see
+        /// cref="ConcurrentDictionary{TKey,TValue}"/>
+        /// class that is empty, has the default concurrency level, has the default initial capacity, and
+        /// uses the default comparer for the key type.
+        /// </summary>
+        public ConcurrentDictionary(IEqualityComparer<TKey> comparer = null)
+        {
+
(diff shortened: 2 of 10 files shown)


rartemev:
@jeffkl PTAL

src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
```diff
@@ -332,11 +333,8 @@ private void EnqueuePacket(INodePacket packet)
             ErrorUtilities.VerifyThrow(null != _packetQueue, "packetQueue is null");
             ErrorUtilities.VerifyThrow(null != _packetAvailable, "packetAvailable is null");
 
-            lock (_packetQueue)
```
jeffkl: I'm not sure if it matters since `Enqueue()` is blocking but `_packetAvailable.Set()` used to be in a `lock`.  Do you know if calling `Set()` in this manner requires some thread synchronization?

src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
```diff
@@ -437,25 +435,11 @@ private void PacketPumpProc()
...
                                 // as fast as we can send them, potentially starving us from detecting a
                                 // terminate event.
-                                int packets = 0;
-                                lock (_packetQueue)
-                                {
-                                    packets = _packetQueue.Count;
-                                }
 
-                                while (packets > 0)
+                                INodePacket packet;
```
jeffkl: Now that this code is simplified, you can probably delete the comment above it.  The comment almost indicates that the `while()` should be checking for the termination signal but the code isn't doing that.

src/MSBuildTaskHost/MSBuildTaskHost.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
-    <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
```
jeffkl: The `MSBuildTaskHost` targets .NET 3.5 so that we can host task libraries that were built against older frameworks.  I don't think we can change its target framework so @AndyGerlicher will need to weigh in

src/MSBuildTaskHost/TypeLoader.cs
```diff
@@ -276,79 +244,54 @@ internal LoadedType GetLoadedTypeByTypeName(string typeName)
...
-                                {
-                                    _typeNameToType.Add(typeName, type);
-                                }
-                            }
-                        }
-
-                        // We could not find the type based on the passed in type name, we now need to see if there is a type which 
-                        // will match based on partially matching the typename. To do this partial matching we need to get the public types in the assembly
-                        if (type == null && !_haveScannedPublicTypes)
+                        if (Interlocked.Read(ref _haveScannedPublicTypes) == 0)
```
jeffkl: Would `Interlocked.Exchange()` be better?

src/MSBuildTaskHost/MSBuildTaskHost.csproj
rartemev: I thought about it, but actually Concurrent collections were introduced since 4.0, so this change is necessary.

src/MSBuildTaskHost/TypeLoader.cs
rartemev: No, it wouldn't. In case there more than 1 thread we have to make sure that they will wait till the first thread finished initialization.

src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
rartemev: According to documentation https://msdn.microsoft.com/en-us/library/system.threading.eventwaithandle.set(v=vs.110).aspx it doesn't. Also in my imagination of implementation such even it also doesn't.

src/MSBuildTaskHost/MSBuildTaskHost.csproj
rartemev: I am not familiar with your process but you probably ship msbuild with required framework. I don't actually understand why we should care about what run msbuild? It is rather question what framework used by application, isn't it? In my optinion run-framework and target-framework are different ones.

AndyGerlicher: @jeffkl is correct. The entire purpose of the TaskHost is to target .NET 3.5. I'm not opposed to forking implementation (uhg) or adding a stub `ConcurrentQueue` class that has uses locking and is slow (perf penalty for using really old projects is probably ok). I've wanted to do the latter for `Lazy<T>` as well since I always find myself using that and having to revert it...

AndyGerlicher:
This looks really good at first glance. Thanks so much for taking the time to do this!

src/MSBuildTaskHost/MSBuildTaskHost.csproj
rartemev: The reason why I had to change `MSBuildTaskHost` too is that on Linux for some reason this code is used. If you know the way to fix it it I am okey with keeping .NET 3.5.

AndyGerlicher: MSBuildTaskHost doesn't ship on .NET Core. Perhaps on Mono? But if it does there it would be the same issue as on Windows. And it should never be used outside of legacy targeting scenarios so perf is not a primary concern there.

rartemev: Actually I can just add simple implementation of ConcurrentDictionary there, if you are okey I will fix it that way.

davkean:
@rartemev This is awesome. Do you know what gains you got on Windows?

Either way, I'm going to pull this down and see if this makes a difference in the spin waiting I was seeing.

src/Shared/TypeLoader.cs
```diff
@@ -345,79 +312,55 @@ internal LoadedType GetLoadedTypeByTypeName(string typeName)
             {
                 ErrorUtilities.VerifyThrowArgumentNull(typeName, "typeName");
 
-                // Only one thread should be doing operations on this instance of the object at a time.
-                lock (_lockObject)
+                if (Interlocked.Read(ref _haveScannedPublicTypes) == 0)
```
RussKeldorph: Aren't there better/accepted ways to implement the singleton pattern in C#?  Lazy<T> was only added in .NET 4, but we're already adding a dependency on that, no?

RussKeldorph:
Indeed awesome.  This degree of performance improvement will have a huge impact on CoreCLR build perf and will finally allow us to unify the Windows and non-Windows developer experiences there.  @Microsoft/msbuild-maintainers: Please help us resolve the issues and get this change in ASAP.  Of course, getting this into a toolset consumed by CoreCLR is possibly harder than fixing this issue, but that's a separate problem.

@vancem @brianrob FYI

AndyGerlicher:
We've done some testing today and we don't see any real performance changes for Windows. Which is surprising. @rartemev do you know if the issue was compounded by anything in particular in your project? Or was it just in Linux in general where you are seeing issues that are resolved by this? Looking at the code it seems odd it wouldn't get somewhat better everywhere. I'm also curious do you get gains without the change to `TypeLoader.cs`?

And yes even if it is only for Linux we can get this in and over to the CLI. The only thing we have to resolve before this can go in is the TaskHost must target 3.5 (which means no default Concurrent or Lazy types).

src/MSBuildTaskHost/MSBuildTaskHost.csproj
rartemev: Added `ConcurrentDictionary` and `ConcurrentQueue` from CoreFX

src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
rartemev: Done

rartemev:
@AndyGerlicher as I said I was testing building tests in CoreCLR repo. So if you want to reproduce my result you should go this way:

1. Clone CoreCLR repo
2. Build runtime. `build.cmd skiptests` or `build.sh`
3. Replace msbuild in `Tools` with new one
4. Run build. `build-tests.cmd` or `build-tests.sh`

The reason why I got started with this issue that we had very different times on Windows and Linux. Since I am neither Linux nor Windows engineer I don't know why it goes like this. But the fact is on Linux we have much worse picture. So this PR is attempt to solve the problem at least on Linux.

rartemev:
@dotnet-bot test this please

rartemev:
@dotnet-bot test this please

rartemev:
I backed TaskHost  to .NET 3.5
@AndyGerlicher @jeffkl PTAL

src/Shared/TypeLoader.cs
```diff
@@ -345,79 +312,55 @@ internal LoadedType GetLoadedTypeByTypeName(string typeName)
...
-                                }
-                            }
-                        }
-
-                        // We could not find the type based on the passed in type name, we now need to see if there is a type which 
-                        // will match based on partially matching the typename. To do this partial matching we need to get the public types in the assembly
-                        if (type == null && !_haveScannedPublicTypes)
+                        if (Interlocked.Read(ref _haveScannedPublicTypes) == 0)
                         {
                             ScanAssemblyForPublicTypes();
```
cdmihai: In the previous version of this method, ScanAssemblyForPublicTypes was only called when Type.GetType failed, whereas here it is called all the time. Are there any weird issues if you move this inside the GetOrAdd lambda and have it called only when Type.GetType fails?

src/MSBuildTaskHost/Concurrent/ConcurrentDictionary.cs
```diff
@@ -0,0 +1,534 @@
+﻿using System;
```
cdmihai: @AndyGerlicher Do these classes need a copyright header? They are copied from corefx, so I am not quite sure ...

src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
cdmihai: I wonder if weird check-then-act style issues can appear here. Before, enqueing and setting the event happened atomically, but now they don't anymore.

One possible new interleaving is when the producer enqueues a packet but before it gets to set the event, the consumer drains the queue from a previous event signal. Then the producer signals the event, and the consumer does a dry run on the empty queue. This appears to be harmless.

I wonder though if there's an interleaving where the consumer misses notifications and packets get stuck on the queue. As far as I can tell it shouldn't happen, but another confirmation would be nice. Would have been nice if ConcurrentQueue offered an ElementAdded event.

src/Shared/TypeLoader.cs
rartemev: There might be an issue if we have 2 or more threads that working with different buckets and both have failed Type.GetType at the same time so since dictionary is not concurrent we will have a race inside `GetOrAdd`.

src/Build/BackEnd/Components/Communications/NodeEndpointInProc.cs
rartemev: With lock here the same situation is still possible. If we have 2+ threads put packets, consumer also might drain the whole queue.

src/Shared/TypeLoader.cs
cdmihai: What is the state they would be racing on? I assume the buckets do not share any state, and then the state accessed by `ScanAssemblyForPublicTypes` is protected by `_lockObject`

rartemev: Fixed

AndyGerlicher:
Looks almost ready to go. I fixed up history a bit but waiting on @rartemev for permissions to push to his branch before this goes in.

PR #2517, opened 2017-09-13
Add MsBuild.Engine.Corext package

cdmihai:
This package aggregates to MsBuild.Corext. Because MsBuild.Corext is hand made, updating it to new versions is difficult. MsBuild.Engine.Corext alleviates this issue by upgrading the master package with the deliverables of this repo (engine binaries and the common build files)


cdmihai:
@dfederm

PR #2525, opened 2017-09-14
Fix NullReferenceException in SdkResolution

jeffkl:
If an SdkResolver returns null, then we throw a NullReferenceException

Also cleaned up the test a little

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
@@ -63,7 +63,12 @@ namespace Microsoft.Build.BackEnd
                     try
                     {
                         var result = (SdkResultImpl)sdkResolver.Resolve(sdk, context, resultFactory);
-                        if (result != null && result.Success)
+                        if (result == null)
+                        {
+                            continue;
+                        }
+
+                        if (result.Success)
                         {
                             LogWarnings(loggingContext, sdkReferenceLocation, result);
                             return result.Path;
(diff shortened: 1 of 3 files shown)


jeffkl:
@dotnet-bot test this please

src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
```diff
@@ -63,10 +63,15 @@ internal SdkResolution(SdkResolverLoader sdkResolverLoader)
...
+                        if (result == null)
+                        {
+                            continue;
+                        }
+
+                        if (result.Success)
                         {
                             LogWarnings(loggingContext, sdkReferenceLocation, result);
-                            return result.Path;
+                            return result.Path;
```
AndyGerlicher: With @rainersigwald out I'll have to be the one to point out your extra space here :P

jeffkl: 👍

src/Build/BackEnd/Components/SdkResolution/SdkResolution.cs
```diff
@@ -63,7 +63,12 @@ internal SdkResolution(SdkResolverLoader sdkResolverLoader)
                     try
                     {
                         var result = (SdkResultImpl)sdkResolver.Resolve(sdk, context, resultFactory);
-                        if (result != null && result.Success)
+                        if (result == null)
+                        {
+                            continue;
```
cdmihai: Can you update the docs for the resolver execute method to specify that returning null means the intent to skip?

PR #2531, opened 2017-09-15
Clear more caches after /restore if necessary

jeffkl:
1. Clear file enumerations cache because when the environment variable `MsBuildCacheFileEnumerations` is set, wild card expansions are cached.
2. Clear file existence cache because when the environment variable `MsBuildCacheFileExistence` is set, file existence is cached

/cc @dfederm

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -1180,12 +1180,14 @@ namespace Microsoft.Build.Shared
             return false;
         }
 
+#if !CLR2COMPATIBILITY
         /// <summary>
-        /// 
+        /// Clears the file existence cache.
         /// </summary>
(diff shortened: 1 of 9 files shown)


src/Shared/FileUtilities.cs
```diff
@@ -1179,5 +1179,13 @@ private static bool PathsEqualNonAscii(string strA, string strB, int i, int leng
 
             return false;
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        internal static void ClearFileExistenceCache()
+        {
+            FileExistenceCache.Clear();
```
cdmihai: FileExistenceCache is not there in MSBuildTaskHost. IfDef the clear method in `CLR2COMPATIBILITY`

PR #2537, opened 2017-09-19
Shorten intermediate app.config path to avoid MAX_PATH

### LINKED ISSUE #1786: PathTooLongException with _GenerateBindingRedirectsIntermediateAppConfig
We just hit a `PathTooLongException` on EF because this variable was producing the following path.

> C:\Projects\EntityFramework\test\Microsoft.EntityFrameworkCore.SqlServer.Design.FunctionalTests\obj\Debug\netcoreapp1.1\Microsoft.EntityFrameworkCore.SqlServer.Design.FunctionalTests.csproj.Microsoft.EntityFrameworkCore.SqlServer.Design.FunctionalTests.dll.config

That filename seems rather excessive. Does it really need ***both*** `$(MSBuildProjectFile)` and `$(TargetFileName)` in there?

nguerrera:
Use the same file name as the output directory $(TargetFileName).config. This is sufficiently unique because $(TargetFileName) is also written to the intermediate directory without further disambiguation.

Fix #1786

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -349,7 +349,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </PropertyGroup>
   <PropertyGroup>
     <IntermediateOutputPath Condition="!HasTrailingSlash('$(IntermediateOutputPath)')">$(IntermediateOutputPath)\</IntermediateOutputPath>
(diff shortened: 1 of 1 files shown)


nguerrera:
Are there tests for GenerateBindingRedirects?

cdmihai:
There are: https://github.com/Microsoft/msbuild/blob/master/src/Tasks.UnitTests/GenerateBindingRedirects_Tests.cs
Though current tests assert the functionality of the task and not how it integrates within the common targets.

PR #2538, opened 2017-09-20
Fixup package

cdmihai:
I based the initial package on my local output directory, but it turns out it was in a bad state. Respecified the dependencies against a clean build. I also twiked the nuspec to make it easier to consume from our VSTS build definition.


cdmihai:
Going to push this to an experimental branch and test out the entire pipeline before sending the PR.

PR #2539, opened 2017-09-20
add initial support for BSD clones

wfurt:
This is initial support for https://github.com/dotnet/corefx/issues/1626

This change will recognize FreeBSD and other BSD clones as UnixLike.
That as minimum allows to handle slashes properly.


AndyGerlicher:
Team Triage:
Change looks pretty good, but we're concerned that calling `OSPlatform.Create` is called every time this property is accessed. This is something that should be `Lazy` and only calculated once. If that is corrected we'll be able to accept this change. Thanks!

wfurt:
Thanks @AndyGerlicher . Original proposal was to add freebsd in https://github.com/dotnet/corefx/issues/23989
It was rejected for now with suggestion to use this construct. I was told it is struct so it won't have cost of creating new object. @stephentoub may have better explanation or suggestions.

stephentoub:
Right, initializing the struct OSPlatform is very cheap.

AndyGerlicher:
Looks not too bad, but it is a very hot path. `IsUnixLike` which calls this is called in many places internally, especially during evaluation (and we're really trying to increase perf for evaluation). I can take care of it if you don't want to add it though.

Method | Mean | Error | StdDev | Allocated
-- | -- | -- | -- | --
OsPlatformCreate | 8.772 ns | 0.0122 ns | 0.0102 ns | 0 B
OsPlatformLazy | 2.220 ns | 0.0032 ns | 0.0030 ns | 0 B

wfurt:
Platform should not change, right? Would it make sense to cache it as bool instead of computing it every time? Maybe that what your lazy suggestion is...?

stephentoub:
@AndyGerlicher, the difference is two null checks.  If you really care about that level of perf, then you can certainly cache the OSPlatform struct, but you shouldn't use a Lazy to do it.  To construct the Lazy is more expensive in both time and space than just calling OSPlatform.Create:
```C#
private static readonly OSPlatform FreeBSD = OSPlatform.Create("FreeBSD");
```

wfurt:
would it make sense to cache whole isUnixLike @stephentoub ???

stephentoub:
> would it make sense to cache whole isUnixLike

If we're concerned about that level of perf optimization, then yes, caching IsUnixLike would be better than caching the individual components of it.

AndyGerlicher:
Yeah I think caching `IsUnixLike` would be the way to go. That's called all over (including in evaluator at least once on every item). I don't think any of the others are used enough to matter.

AndyGerlicher:
I didn't have permissions to push to your remote. Can you pull in my commit? If you fetch upstream you'll have that commit I added it to the Microsoft/msbuild repo.
https://github.com/Microsoft/msbuild/commit/b4690b0f2eaa58bdfb2528c8e688a2b9376246d1

Thanks!

wfurt:
Thanks Andy. Done.

PR #2541, opened 2017-09-21
Add support to skip missing targets in MSBuild task

### LINKED ISSUE #2471: Provide a way to check if a target exists in a project without errors
### Proposal

There should be a way to check if a target exists or not in a project without the user seeing warnings/errors when it does not.

### Scenario

NuGet restore walks all project references by using the `<MSBuild>` element to call a target in child projects. Child projects that do not import common props/targets are missing the restore target display an error to the user. Often it is expected that the target will not exist, for example if the user creates a custom project file and references it.

Since Restore isn't able to gracefully probe projects the user sees a lot of extra noise when the target is missing. Restore also isn't able to tell the difference between a project that doesn't support restore, and one that has invalid XML or other issues.ore from failing when it should.

### Possible ways to solve this

* `<MSBuild>` could provide an option to skip projects where the target doesn't exist
* Provide a way to filter a set of projects based on the existence of a target or property

AndyGerlicher:
Note: Tests still to come. Since this is a lot of cascading changes I wanted to get it out for review before I spent too much time on tests.

Add optional `SkipNonexistentTargets` attribute to the `MSBuild` task. This allows for the ability to make changes to our project to project reference protocol which would be an otherwise breaking change.

Related to #2472

Closes #2471

### DIFF
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -1000,6 +1000,7 @@ namespace Microsoft.Build.Execution
         None = 0,
         ProvideProjectStateAfterBuild = 2,
         ReplaceExistingProjectInstance = 1,
+        SkipNonexistentTargets = 16,
     }
     public partial class BuildResult
     {
--- a/src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
@@ -1244,6 +1244,29 @@ namespace Microsoft.Build.UnitTests.BackEnd
             Assert.Equal(TargetResultCode.Failure, resultsCache.GetResultForRequest(entry.Request)["_Error1"].ResultCode);
         }
 
+        [Fact]
+        public void SkipNonexistentTargetsDoesNotExecuteOrCacheTargetResult()
+        {
+            string projectContents = @"<Target Name=""Build"" />";
+
+            var project = CreateTestProject(projectContents, string.Empty, "Build");
+            TargetBuilder builder = (TargetBuilder)_host.GetComponent(BuildComponentType.TargetBuilder);
+            MockTaskBuilder taskBuilder = (MockTaskBuilder)_host.GetComponent(BuildComponentType.TaskBuilder);
+            taskBuilder.FailTaskNumber = 1;
+
+            IConfigCache cache = (IConfigCache)_host.GetComponent(BuildComponentType.ConfigCache);
+            BuildRequestEntry entry = new BuildRequestEntry(CreateNewBuildRequest(1, new[] { "NotFound" }, BuildRequestDataFlags.SkipNonexistentTargets), cache[1]);
+            var buildResult = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
+
+            IResultsCache resultsCache = (IResultsCache)_host.GetComponent(BuildComponentType.ResultsCache);
+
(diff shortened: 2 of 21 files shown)


ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -1000,6 +1000,7 @@ public enum BuildRequestDataFlags
         None = 0,
         ProvideProjectStateAfterBuild = 2,
         ReplaceExistingProjectInstance = 1,
+        SkipMissingTargets = 22,
```
jeffkl: Looks like `SkipMissingTargets` should be `0x10`

jeffkl: Oh and be sure to build CoreCLR locally to generate ref sources for .NET Core

AndyGerlicher:
@dotnet-bot test this please

src/Build/BackEnd/Components/BuildRequestEngine/FullyQualifiedBuildRequest.cs
```diff
@@ -25,68 +23,42 @@ namespace Microsoft.Build.BackEnd
...
-
-        /// <summary>
         /// Initializes a build request.
         /// </summary>
         /// <param name="config">The configuration to use for the request.</param>
         /// <param name="targets">The set of targets to build.</param>
         /// <param name="resultsNeeded">Whether or not to wait for the results of this request.</param>
-        public FullyQualifiedBuildRequest(BuildRequestConfiguration config, string[] targets, bool resultsNeeded)
+        /// <param name="flags">Flags specified for the build request.</param>
+        public FullyQualifiedBuildRequest(BuildRequestConfiguration config, string[] targets, bool resultsNeeded, BuildRequestDataFlags flags = BuildRequestDataFlags.None)
```
cdmihai: Maybe just send the bool to the FullyQualifiedBuildRequest? I don't think it needs the entire enum right?

src/Tasks.UnitTests/MSBuild_Tests.cs
```diff
@@ -1341,5 +1342,56 @@ public void TargetNameIsCaseInsensitive()
...
+        [Fact]
+        public void NonexistentTargetFailsAfterSkipped()
+        {
+            ObjectModelHelpers.DeleteTempProjectDirectory();
+            ObjectModelHelpers.CreateFileInTempProjectDirectory(
+                "SkipNonexistentTargets.csproj",
+                @"<Project>
+                    <Target Name=`t` >
+	                    <MSBuild Projects=`SkipNonexistentTargets.csproj` Targets=`t_nonexistent` SkipNonexistentTargets=`true` />
+                        <MSBuild Projects=`SkipNonexistentTargets.csproj` Targets=`t_nonexistent` SkipNonexistentTargets=`false` />
```
cdmihai: Maybe build again with SkipNonexistentTargets=`true` to ensure it actually gets turned off.

cdmihai: Oops, the other way around. Have three msbuild task invocations with false, true, false

---
In reply to: [141172470](https://github.com/Microsoft/msbuild/pull/2541#discussion_r141172470) [](ancestors = 141172470)

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
```diff
@@ -1244,6 +1244,29 @@ public void SkippedTargetWithFailedDependenciesStopsBuild()
             Assert.Equal(TargetResultCode.Failure, resultsCache.GetResultForRequest(entry.Request)["_Error1"].ResultCode);
         }
 
+        [Fact]
+        public void SkipNonexistentTargetsDoesNotExecuteOrCacheTargetResult()
```
cdmihai: :thumbsup:

PR #2543, opened 2017-09-21
Fix BinaryLogger not creating directories when it doesn't exist

### LINKED ISSUE #2479: Directory for binary log not created
Start msbuild with binary log enabled:

`msbuild.exe /bl:logs\out.binlog`

If the `logs` folder does not exist, msbuild fails.

This means `/bl` cannot easily replace `/flp:filename=` which does ensure directory exists.

samneirinck:
Fixes #2479

### DIFF
--- a/src/Build.UnitTests/BinaryLogger_Tests.cs
+++ b/src/Build.UnitTests/BinaryLogger_Tests.cs
@@ -64,5 +64,25 @@ namespace Microsoft.Build.UnitTests
                 Environment.SetEnvironmentVariable(MSBUILDTARGETOUTPUTLOGGING, originalTargetOutputLogging);
             }
         }
+
+        [Fact]
+        public void NonExistingDirectory()
+        {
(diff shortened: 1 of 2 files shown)


src/Build.UnitTests/BinaryLogger_Tests.cs
```diff
@@ -64,5 +64,25 @@ public void TestBinaryLoggerRoundtrip()
...
+            string directory = Path.Combine(ObjectModelHelpers.TempProjectDir, Guid.NewGuid().ToString("N"));
+            string log = Path.Combine(directory, "build.binlog");
+            Assert.False(Directory.Exists(directory));
+            Assert.False(File.Exists(log));
+
+            var binaryLogger = new BinaryLogger { Parameters = log };
+
+            try
+            {
+                ObjectModelHelpers.BuildProjectExpectSuccess(s_testProject, binaryLogger);
```
cdmihai: Maybe repeat the asserts that the directory and file exist

KirillOsenkov:
Wow, thanks for fixing this! 👍

PR #2566, opened 2017-09-29
Include Sdk Resolver in bootstrapped MSBuild

dsplaisted:
- Include Sdk Resolver in bootstrapped MSBuild.  This should mean the bootstrapped MSBuild can successfully build SDK-Style projects that use the 2.0 SDK or higher
- Add binary logging to build scripts

@AndyGerlicher @cdmihai @jeffkl for review

### DIFF
--- a/targets/BootStrapMSBuild.proj
+++ b/targets/BootStrapMSBuild.proj
@@ -31,6 +31,7 @@
     <ItemGroup>
       <InstalledVersionedExtensions Include="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\**\*.targets" />
       <InstalledVersionedExtensions Include="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\**\*.props" />
+      <SdkResolverFiles Include="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Bin\SdkResolvers\**\*.*" />
       <InstalledSdks Include="$(MSBuildExtensionsPath)\Sdks\**\*.*" />
       <InstalledStaticAnalysisTools Include="$(VsInstallRoot)\Team Tools\Static Analysis Tools\**\*.*" />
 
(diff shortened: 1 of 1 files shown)


AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:
This _was_ failing because the Windows CI image was really old and didn't have 15.3 where `/bl` was introduced. Now it's failing because the new CI image doesn't have the older Windows SDK. Can't win apparently! I have a fix in #2568 to disable the test.

cibuild.cmd
```diff
@@ -125,6 +126,7 @@ echo ** Moving bootstrapped MSBuild to the bootstrap folder
 taskkill /F /IM vbcscompiler.exe
 
 set MSBUILDLOGPATH=%~dp0msbuild_move_bootstrap-%HOST%.log
+set MSBUILDBINARYLOGPATH=%~dp0msbuild_move_bootstrap-%HOST%.binlog
```
cdmihai: The file logger seems unnecessary now right?. Can you please remove it?

dsplaisted: @cdmihai You're saying we should no longer create a text log?

cdmihai: Initially I thought that having the binary log makes the text one obsolete. But now that I think of it, having two loggers piped through the msbuild app makes for a nice test case. So let's leave both of them :)

dsplaisted:
@dotnet-bot test Windows_NT Build for Full

AndyGerlicher:
Still failing because on CoreCLR CI we download a (super old) copy of MSBuild built for .NET Core and use that for the initial build. That won't have `/bl` until we upgrade to the new SDK. So I think we can't do `/bl` in the first build.

AndyGerlicher:
FYI @dsplaisted  I took over your branch and dropped the `/bl` commits. Don't think we can take them at the moment for the initial build.

dsplaisted:
@AndyGerlicher Thanks!

PR #2567, opened 2017-09-29
Include .NET Framework facades if there is a reference to netstandard.dll

### LINKED ISSUE #2199: ImplicitlyExpandDesignTimeFacades must trigger off of netstandard.dll once it is inbox
Today ImplicitlyExpandDesignTimeFacades conditions the injection of facades based only on a reference to System.Runtime.

That works for contract-based PCLs, but not for netstandard libraries.  We'll want to also inject facades when netstandard.dll is referenced once netstandard.dll goes inbox.

/cc @alexghi

dsplaisted:
Fixes #2199.  This is needed for .NET 4.7.1 to support references to .NET Standard 2.0 libraries.

@AndyGerlicher @cdmihai @jeffkl @nguerrera @ericstj for review

### DIFF
--- a/src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
+++ b/src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
@@ -37,6 +37,11 @@ namespace Microsoft.Build.Tasks
         private const string SystemRuntimeAssemblyName = "System.Runtime";
 
         /// <summary>
+        /// additional key assembly used to trigger inclusion of facade references. 
+        /// </summary>
+        private const string NETStandardAssemblyName = "netstandard";
+
+        /// <summary>
         /// Delegate to a method that takes a targetFrameworkDirectory and returns an array of redist or subset list paths
         /// </summary>
         /// <param name="targetFrameworkDirectory">TargetFramework directory to search for redist or subset list</param>
@@ -881,6 +886,17 @@ namespace Microsoft.Build.Tasks
             get;
             private set;
         }
+
+        /// <summary>
+        /// Whether the assembly or any of its primary references depends on netstandard
+        /// </summary>
+        [Output]
+        public String DependsOnNETStandard
+        {
+            get;
+            private set;
+        }
+
 
         #endregion
         #region Logging
@@ -2242,16 +2258,25 @@ namespace Microsoft.Build.Tasks
                     }
 
                     bool useSystemRuntime = false;
+                    bool useNetStandard = false;
                     foreach (var reference in dependencyTable.References.Keys)
                     {
                         if (string.Equals(SystemRuntimeAssemblyName, reference.Name, StringComparison.OrdinalIgnoreCase))
                         {
                             useSystemRuntime = true;
+                        }
+                        if (string.Equals(NETStandardAssemblyName, reference.Name, StringComparison.OrdinalIgnoreCase))
+                        {
+                            useNetStandard = true;
+                        }
+                        if (useSystemRuntime && useNetStandard)
+                        {
                             break;
                         }
                     }
 
-                    if (!useSystemRuntime && !FindDependencies)
+
+                    if ((!useSystemRuntime || !useNetStandard) && !FindDependencies)
                     {
                         // when we are not producing the dependency graph look for direct dependencies of primary references.
                         foreach (var resolvedReference in dependencyTable.References.Values)
@@ -2266,10 +2291,15 @@ namespace Microsoft.Build.Tasks
                                         useSystemRuntime = true;
                                         break;
                                     }
+                                    if (string.Equals(NETStandardAssemblyName, dependentReference.Name, StringComparison.OrdinalIgnoreCase))
+                                    {
+                                        useNetStandard = true;
+                                        break;
+                                    }
                                 }
                             }
 
-                            if (useSystemRuntime)
+                            if (useSystemRuntime && useNetStandard)
                             {
                                 break;
                             }
@@ -2277,6 +2307,7 @@ namespace Microsoft.Build.Tasks
                     }
 
                     this.DependsOnSystemRuntime = useSystemRuntime.ToString();
+                    this.DependsOnNETStandard = useNetStandard.ToString();
 
 #if FEATURE_BINARY_SERIALIZATION
                     WriteStateFile();
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2040,7 +2040,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="SuggestedRedirects" ItemName="SuggestedBindingRedirects"/>
       <Output TaskParameter="FilesWritten" ItemName="FileWrites"/>
       <Output TaskParameter="DependsOnSystemRuntime" PropertyName="DependsOnSystemRuntime"/>
-      <Output TaskParameter="DependsOnNETStandard" PropertyName="DependsOnNETStandard"/>
+      <Output TaskParameter="DependsOnNETStandard" PropertyName="_DependsOnNETStandard"/>
     </ResolveAssemblyReference>
   </Target>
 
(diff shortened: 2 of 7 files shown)


src/Tasks/Microsoft.NETFramework.CurrentVersion.targets
```diff
@@ -93,6 +93,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <PropertyGroup>
       <!-- Does one of our dependencies reference a System.Runtime-based portable library? -->
       <_HasReferenceToSystemRuntime Condition="'$(DependsOnSystemRuntime)' == 'true'">true</_HasReferenceToSystemRuntime>
+      <_HasReferenceToSystemRuntime Condition="'$(DependsOnNETStandard)' == 'true'">true</_HasReferenceToSystemRuntime>
```
nguerrera: Do we need all of the facades if there is a reference to netstandard.dll but no reference to system.runtime? Seems like there could be a perf win if only netstandard.dll is in the closure if we only pass along the netstandard.dll facade.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2040,6 +2040,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="SuggestedRedirects" ItemName="SuggestedBindingRedirects"/>
       <Output TaskParameter="FilesWritten" ItemName="FileWrites"/>
       <Output TaskParameter="DependsOnSystemRuntime" PropertyName="DependsOnSystemRuntime"/>
+      <Output TaskParameter="DependsOnNETStandard" PropertyName="DependsOnNETStandard"/>
```
dsplaisted: The property name `DependsOnNETStandard` is also [used by the Microsoft.NET.Build.Extensions targets](https://github.com/dotnet/sdk/blob/d9955fcf3440f8596f4e63aa37a8dbcc477d97f4/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets#L45).  Should we choose a different name here, or let them overlap?  I believe that if both the extensions target and RAR are writing to this property, they would agree on its value.  Thoughts? @ericstj

dsplaisted:
Also @AlexGhiondea for review

src/Tasks/Microsoft.NETFramework.CurrentVersion.targets
nguerrera: If it turns out we can't do that, can we at least rename this to _NeedsFacades or something? We shouldn't have _HasReferenceToSystemRuntime=true with no reference to System.Runtime.

ericstj: /cc @weshaggard @alexghi
I believe we don't need everything if only netstandard.dll is referenced.  

This would only be true for the targeting packs: in those cases the facades referencing non-facades.  We couldn't make that assumption in the Extensions targets in the SDK since netstandard referenced other assemblies that were injected (eg: netstandard.dll > System.ValueTuple, netstandard.dll > System.Net.Http, etc).

build/NuGetPackages/Microsoft.Build.Framework.nuspec
```diff
@@ -15,11 +15,11 @@
     <tags>MSBuild</tags>
     <dependencies>
       <group targetFramework=".NETStandard1.3">
+        <dependency id="System.Resources.ResourceManager" version="4.0.1" />
         <dependency id="System.Collections" version="4.0.11" />
         <dependency id="System.Diagnostics.Debug" version="4.0.11" />
         <dependency id="System.Globalization" version="4.0.11" />
         <dependency id="System.Linq" version="4.1.0" />
-        <dependency id="System.Resources.ResourceManager" version="4.0.1" />
```
AndyGerlicher: Revert this file please. It does this sometimes, not sure why. We've tried checking it in this way but it will just switch back.

src/Tasks/Microsoft.Common.CurrentVersion.targets
ericstj: Indeed they would agree on the value.  The extensions write to it earlier (honoring user specified value) and consume the value immediately.  RAR writes to it later ignoring a previously set value.  I don't think its a problem that they both use it, more of a style thing.

IMHO these should be `_` properties anyway since MSBuild isn't honoring a value previously set by the user (like we do in extensions for DependsOnNETStandard).

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -2242,16 +2258,25 @@ out _copyLocalFiles
...
                     {
                         if (string.Equals(SystemRuntimeAssemblyName, reference.Name, StringComparison.OrdinalIgnoreCase))
                         {
                             useSystemRuntime = true;
+                        }
+                        if (string.Equals(NETStandardAssemblyName, reference.Name, StringComparison.OrdinalIgnoreCase))
+                        {
+                            useNetStandard = true;
+                        }
+                        if (useSystemRuntime && useNetStandard)
```
weshaggard: `should this be `useSystemRuntime || useNetStandard`

dsplaisted: @weshaggard No, we only want to break out of the loop if we've found both of them.  Otherwise we won't properly detect the case where both System.Runtime and netstandard are in the dependency closure.

AndyGerlicher:
Just discussed with Mihai. Looks good overall, but we're a little concerned with perf. It looks like it will exit the loop later than it would have before waiting for `useSystemRuntime && useNetStandard`. We're going to run some tests to try to quantify if this is an issue or not. As an alternative could we pass in a value of whether or not we care about useNetStandard or not for legacy projects?

dsplaisted:
@AndyGerlicher How would "legacy projects" be defined?  The logic is needed for .NET 4.7.1 or higher, but if there's a meaningful different in perf it's probably not OK to only fix the regression for .NET 4.7 and earlier.

In any case, I don't think there will be a meaningful effect on perf.  I think ResolveAssemblyReferences is going to iterate through all the references at least once anyway (it has to resolve the references, after all).  Also, lots of existing projects won't have any transitive references to System.Runtime, so in many cases it wouldn't be breaking out of this specific loop early anyway.

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -2266,17 +2291,23 @@ out _copyLocalFiles
                                         useSystemRuntime = true;
                                         break;
                                     }
+                                    if (string.Equals(NETStandardAssemblyName, dependentReference.Name, StringComparison.OrdinalIgnoreCase))
+                                    {
+                                        useNetStandard = true;
+                                        break;
+                                    }
```
brcaswell: where is this logic in current code base? ...and does it still reflect this logic, `break` with `useSystemRuntime` assignment without checking subsequent `dependentReferences` iteration for a dependency on `NETStandardAssemblyName` and assigning `useNetStandard`?

i.e.
In a scenario where `useSystemRuntime` is `true` and `useNetStandard` is `false` the outer scope `(!useSystemRuntime || !useNetStandard)` condition logic is `true`, with warrants entering this scope with the intent to determine whether `useNetStandard` should be `true`, right?  doesn't the `break` provide a scenario where it won't\didn't?

brcaswell: potentially, this should be\have been (may be now)
```
if (!useSystemRuntime && string.Equals(SystemRuntimeAssemblyName, dependentReference.Name, StringComparison.OrdinalIgnoreCase))
{
	useSystemRuntime = true;
} 
else if (!useNetStandard && string.Equals(NETStandardAssemblyName, dependentReference.Name, StringComparison.OrdinalIgnoreCase))
{
	useNetStandard = true;
}

if (useNetStandard && useSystemRuntime)
{
	break;
}
```

brcaswell: this depends on whether it's possible that `dependencyTable.References.Values` can have both, can it?

Issue #2392, opened 2017-08-03
For large projects with globs, MSBuild takes up to 30% - 50% of a build searching the disk


davkean:
I'm been analyzing two projects to try and understand the reasoning behind long design-time builds. Some projects design-time builds are taking upwards of 3 seconds to run - when the expectation is that don't take more than 50 - 100 ms. This results in very observable lag times from when you make a change (such as adding a reference) and it seen by the compiler, or time before you get useful intellisense the first time you open a solution. It also makes evaluation inside the VS process itself slow, resulting in observable UI delays if we put this on the UI thread, or the UI thread is blocked on evaluation in another thread.

Looking at these builds under PerfView, I can see that half the build is dedicated to evaluating the project, of which the majority is caused by MSBuild searching the disk:

The second project was taking 3 seconds for a design-time build, it's build evaluation time looked like this:
![image](https://user-images.githubusercontent.com/1103906/28906824-94151262-785d-11e7-8105-ce9717d395bf.png)

There were two main projects that I was looking, the first http://github.com/dotnet/roslyn, was taking up to 1 second for a design-time build, it's build evaluation time looked like this:
![image](https://user-images.githubusercontent.com/1103906/28906776-3008a374-785d-11e7-8164-61457bf0de6b.png)

The first project, Roslyn, does nothing but import the default globs from the SDK.

The second project, is a lot more complex, and includes the following syntax:

``` XML
       <ItemGroup>
              <None Include="**\*.iPhone.cs" />
              <None Include="**\*.Google.cs" />
              <None Include="**\*.Microsoft.cs" />
              <None Include="**\*.WinRT.cs" />
              <None Include="**\*.Xamarin.cs" />
       </ItemGroup>

       <ItemGroup>
              <Compile Remove="**\*.iPhone.cs" />
              <Compile Include="**\*.iPhone.cs" Condition="'$(TargetDevice)' == 'iPhone'" />

              <Compile Remove="**\*.Google.cs" />
              <Compile Include="**\*.Google.cs" Condition="$(TargetDevice)' == 'Google'" />
              
              <Compile Remove="**\*.Xamarin.cs" />
              <Compile Include="**\*.Xamarin.cs" Condition="'$(TargetDevice)' == 'Google' or '$(TargetDevice)' == 'iPhone'" />

              <Compile Remove="**\*.Microsoft.cs" />
              <Compile Include="**\*.Microsoft.cs" Condition="$(TargetDevice)' == 'Microsoft'" />
[...]

PR #2572, opened 2017-10-01
Optimizations for FileMatcher.GetFiles 

maca88:
Here are the optimization changes that were discussed in #2392, which are divided into three commits:

1. Usage of `Directory.EnumerateFiles` and `Directory.EnumerateDirectories` and checking excludes with a custom wildcard matching algorithm.
2. Adding directory caching passed by `LazyItemEvaluator`
3. Traversing subdirectories in parallel by using `Parallel.ForEach`

For the parallel change I did some modifications in order to prevent spawning too much threads. With the old modification the traversing created up to 9 threads on my computer, with the new one, it creates up to 4 threads. I found this to be a good balance between performance and threads being used, at least on my computer (Intel I7 860).

[This](https://github.com/maca88/msbuild/commit/24ccceeafb1d7969107fc02c272bcd29909eea4d#diff-7a21f5e73644f08fe8b0dc9763a632aaR30) test was used on [NHibernate.Test](https://github.com/nhibernate/nhibernate-core/tree/master/src/NHibernate.Test) project and [Roslyn](https://github.com/dotnet/roslyn) root folder. For NHibernate also the evaluation was tested by using the [Project](https://github.com/Microsoft/msbuild/blob/master/src/Build/Definition/Project.cs#L50) class.

Latest master:
```
NH:
File spec: **/*, Call time: 2650ms, Returned files: 14
File spec: **/*.cs, Call time: 1556ms, Returned files: 3509
File spec: **/*.resx, Call time: 1548ms, Returned files: 0
File spec: **\*.hbm.xml, Call time: 258ms, Returned files: 827
File spec: **\*.jpg, Call time: 246ms, Returned files: 1
Total time: 6261ms

NHibernate.Test Project Evaluation: Total time: 7391ms

Roslyn:
File spec: **/*, Call time: 5572ms, Returned files: 5485
File spec: **/*.cs, Call time: 3398ms, Returned files: 8631
File spec: **/*.resx, Call time: 3187ms, Returned files: 70
File spec: **\*.hbm.xml, Call time: 516ms, Returned files: 0
File spec: **\*.jpg, Call time: 524ms, Returned files: 0
Total time: 13199ms
```

This PR:
```
NH:
File spec: **/*, Call time: 172ms, Returned files: 14
File spec: **/*.cs, Call time: 114ms, Returned files: 3509
File spec: **/*.resx, Call time: 63ms, Returned files: 0
File spec: **\*.hbm.xml, Call time: 68ms, Returned files: 827
File spec: **\*.jpg, Call time: 64ms, Returned files: 1
Total time: 483ms

NHibernate.Test Project Evaluation: Total time: 2097ms

Roslyn:
File spec: **/*, Call time: 748ms, Returned files: 5485
File spec: **/*.cs, Call time: 373ms, Returned files: 8631
File spec: **/*.resx, Call time: 186ms, Returned files: 70
File spec: **\*.hbm.xml, Call time: 170ms, Returned files: 0
File spec: **\*.jpg, Call time: 177ms, Returned files: 0
Total time: 1656ms
```

All tests were ran in Release configuration on Intel I7 860 processor and the average of 5 runs was taken.

### DIFF
--- a/src/Build.UnitTests/Microsoft.Build.Engine.UnitTests.csproj
+++ b/src/Build.UnitTests/Microsoft.Build.Engine.UnitTests.csproj
@@ -249,6 +249,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup Condition="'$(NetCoreBuild)' != 'true'">
+    <Reference Include="System.Collections.Immutable, Version=1.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Xml" />
--- a/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -43,6 +44,11 @@ namespace Microsoft.Build.Evaluation
 
                 _itemSpec.Expander = _expander;
             }
+
+            /// <summary>
+            /// Cache used for caching IO operation results
+            /// </summary>
+            protected ConcurrentDictionary<string, ImmutableArray<string>> EntriesCache => _lazyEvaluator._entriesCache;
 
             public virtual void Apply(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
             {
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections;
 using System.Collections.Concurrent;
 using System.IO;
 using System.Text;
@@ -15,6 +16,8 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using System.Globalization;
 using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
 using Microsoft.Build.Utilities;
 
 namespace Microsoft.Build.Shared
@@ -71,8 +74,8 @@ namespace Microsoft.Build.Shared
         /// <param name="pattern">The file pattern.</param>
         /// <param name="projectDirectory"></param>
         /// <param name="stripProjectDirectory"></param>
-        /// <returns>The array of filesystem entries.</returns>
-        internal delegate string[] GetFileSystemEntries(FileSystemEntity entityType, string path, string pattern, string projectDirectory, bool stripProjectDirectory);
+        /// <returns>An immutable array of filesystem entries.</returns>
+        internal delegate ImmutableArray<string> GetFileSystemEntries(FileSystemEntity entityType, string path, string pattern, string projectDirectory, bool stripProjectDirectory);
 
         internal static void ClearFileEnumerationsCache()
         {
@@ -119,39 +122,39 @@ namespace Microsoft.Build.Shared
         /// <param name="projectDirectory">The directory for the project within which the call is made</param>
         /// <param name="stripProjectDirectory">If true the project directory should be stripped</param>
         /// <returns></returns>
-        private static string[] GetAccessibleFileSystemEntries(FileSystemEntity entityType, string path, string pattern, string projectDirectory, bool stripProjectDirectory)
+        private static ImmutableArray<string> GetAccessibleFileSystemEntries(FileSystemEntity entityType, string path, string pattern, string projectDirectory, bool stripProjectDirectory)
         {
             path = FileUtilities.FixFilePath(path);
-            string[] files = null;
             switch (entityType)
             {
-                case FileSystemEntity.Files: files = GetAccessibleFiles(path, pattern, projectDirectory, stripProjectDirectory); break;
-                case FileSystemEntity.Directories: files = GetAccessibleDirectories(path, pattern); break;
-                case FileSystemEntity.FilesAndDirectories: files = GetAccessibleFilesAndDirectories(path, pattern); break;
+                case FileSystemEntity.Files: return GetAccessibleFiles(path, pattern, projectDirectory, stripProjectDirectory);
+                case FileSystemEntity.Directories: return GetAccessibleDirectories(path, pattern);
+                case FileSystemEntity.FilesAndDirectories: return GetAccessibleFilesAndDirectories(path, pattern);
                 default:
                     ErrorUtilities.VerifyThrow(false, "Unexpected filesystem entity type.");
                     break;
             }
-
-            return files;
+            return ImmutableArray<string>.Empty;
         }
 
         /// <summary>
-        /// Returns an array of file system entries matching the specified search criteria. Inaccessible or non-existent file
+        /// Returns an immutable array of file system entries matching the specified search criteria. Inaccessible or non-existent file
         /// system entries are skipped.
         /// </summary>
         /// <param name="path"></param>
         /// <param name="pattern"></param>
-        /// <returns>Array of matching file system entries (can be empty).</returns>
-        private static string[] GetAccessibleFilesAndDirectories(string path, string pattern)
+        /// <returns>An immutable array of matching file system entries (can be empty).</returns>
+        private static ImmutableArray<string> GetAccessibleFilesAndDirectories(string path, string pattern)
         {
-            string[] entries = null;
-
             if (Directory.Exists(path))
             {
                 try
                 {
-                    entries = Directory.GetFileSystemEntries(path, pattern);
+                    return (ShouldEnforceMatching(pattern)
+                        ? Directory.EnumerateFileSystemEntries(path, pattern)
+                            .Where(o => IsMatch(Path.GetFileName(o), pattern, true))
+                        : Directory.EnumerateFileSystemEntries(path, pattern)
+                    ).ToImmutableArray();
                 }
                 // for OS security
                 catch (UnauthorizedAccessException)
@@ -164,17 +167,37 @@ namespace Microsoft.Build.Shared
                     // do nothing
                 }
             }
-
-            if (entries == null)
-            {
-                entries = Array.Empty<string>();
-            }
-
-            return entries;
+            return ImmutableArray<string>.Empty;
         }
 
         /// <summary>
-        /// Same as Directory.GetFiles(...) except that files that
+        /// Determine if the given search pattern will match loosely on Windows
+        /// </summary>
+        /// <param name="searchPattern">The search pattern to check</param>
+        /// <returns></returns>
+        private static bool ShouldEnforceMatching(string searchPattern)
+        {
+            if (!NativeMethodsShared.IsWindows || searchPattern == null)
+            {
+                return false;
+            }
+            // NOTE: Windows matches loosely in three cases (in the absence of the * wildcard in the extension):
+            // 1) if the extension ends with the ? wildcard, it matches files with shorter extensions also e.g. "file.tx?" would
+            //    match both "file.txt" and "file.tx"
+            // 2) if the extension is three characters, and the filename contains the * wildcard, it matches files with longer
+            //    extensions that start with the same three characters e.g. "*.htm" would match both "file.htm" and "file.html"
+            // 3) if the ? wildcard is to the left of a period, it matches files with shorter name e.g. ???.txt would match
+            //    foo.txt, fo.txt and also f.txt
+            return searchPattern.IndexOf("?.", StringComparison.Ordinal) != -1 ||
+                   (
+                       Path.GetExtension(searchPattern).Length == (3 + 1 /* +1 for the period */) &&
+                       searchPattern.IndexOf('*') != -1
+                   ) ||
+                   searchPattern.EndsWith("?", StringComparison.Ordinal);
+        }
+
+        /// <summary>
+        /// Same as Directory.EnumerateFiles(...) except that files that
         /// aren't accessible are skipped instead of throwing an exception.
         /// 
         /// Other exceptions are passed through.
@@ -184,7 +207,7 @@ namespace Microsoft.Build.Shared
         /// <param name="projectDirectory">The project directory</param>
         /// <param name="stripProjectDirectory"></param>
         /// <returns>Files that can be accessed.</returns>
-        private static string[] GetAccessibleFiles
+        private static ImmutableArray<string> GetAccessibleFiles
         (
             string path,
             string filespec,     // can be null
@@ -198,15 +221,24 @@ namespace Microsoft.Build.Shared
                 string dir = ((path.Length == 0) ? s_thisDirectory : path);
 
                 // get all files in specified directory, unless a file-spec has been provided
-                string[] files = (filespec == null)
-                    ? Directory.GetFiles(dir)
-                    : Directory.GetFiles(dir, filespec);
-
+                IEnumerable<string> files;
+                if (filespec == null)
+                {
+                    files = Directory.EnumerateFiles(dir);
+                }
+                else
+                {
+                    files = Directory.EnumerateFiles(dir, filespec);
+                    if (ShouldEnforceMatching(filespec))
+                    {
+                        files = files.Where(o => IsMatch(Path.GetFileName(o), filespec, true));
+                    }
+                }
                 // If the Item is based on a relative path we need to strip
                 // the current directory from the front
                 if (stripProjectDirectory)
                 {
-                    RemoveProjectDirectory(files, projectDirectory);
+                    files = RemoveProjectDirectory(files, projectDirectory);
                 }
                 // Files in the current directory are coming back with a ".\"
                 // prepended to them.  We need to remove this; it breaks the
@@ -215,25 +247,25 @@ namespace Microsoft.Build.Shared
                 // didn't itself contain a ".\".
                 else if (!path.StartsWith(s_thisDirectory, StringComparison.Ordinal))
                 {
-                    RemoveInitialDotSlash(files);
+                    files = RemoveInitialDotSlash(files);
                 }
 
-                return files;
+                return files.ToImmutableArray();
             }
             catch (System.Security.SecurityException)
             {
                 // For code access security.
-                return Array.Empty<string>();
+                return ImmutableArray<string>.Empty;
             }
             catch (System.UnauthorizedAccessException)
             {
                 // For OS security.
-                return Array.Empty<string>();
+                return ImmutableArray<string>.Empty;
             }
         }
 
         /// <summary>
-        /// Same as Directory.GetDirectories(...) except that files that
+        /// Same as Directory.EnumerateDirectories(...) except that files that
         /// aren't accessible are skipped instead of throwing an exception.
         /// 
         /// Other exceptions are passed through.
@@ -241,7 +273,7 @@ namespace Microsoft.Build.Shared
         /// <param name="path">The path.</param>
         /// <param name="pattern">Pattern to match</param>
         /// <returns>Accessible directories.</returns>
-        private static string[] GetAccessibleDirectories
+        private static ImmutableArray<string> GetAccessibleDirectories
         (
             string path,
             string pattern
@@ -249,15 +281,19 @@ namespace Microsoft.Build.Shared
         {
             try
             {
-                string[] directories = null;
+                IEnumerable<string> directories = null;
 
                 if (pattern == null)
                 {
-                    directories = Directory.GetDirectories((path.Length == 0) ? s_thisDirectory : path);
+                    directories = Directory.EnumerateDirectories((path.Length == 0) ? s_thisDirectory : path);
                 }
                 else
                 {
-                    directories = Directory.GetDirectories((path.Length == 0) ? s_thisDirectory : path, pattern);
+                    directories = Directory.EnumerateDirectories((path.Length == 0) ? s_thisDirectory : path, pattern);
+                    if (ShouldEnforceMatching(pattern))
+                    {
+                        directories = directories.Where(o => IsMatch(Path.GetFileName(o), pattern, true));
+                    }
                 }
 
                 // Subdirectories in the current directory are coming back with a ".\"
@@ -267,20 +303,20 @@ namespace Microsoft.Build.Shared
                 // didn't itself contain a ".\".
                 if (!path.StartsWith(s_thisDirectory, StringComparison.Ordinal))
                 {
-                    RemoveInitialDotSlash(directories);
+                    directories = RemoveInitialDotSlash(directories);
                 }
 
-                return directories;
+                return directories.ToImmutableArray();
             }
             catch (System.Security.SecurityException)
             {
                 // For code access security.
-                return Array.Empty<string>();
+                return ImmutableArray<string>.Empty;
             }
             catch (System.UnauthorizedAccessException)
             {
                 // For OS security.
-                return Array.Empty<string>();
+                return ImmutableArray<string>.Empty;
             }
         }
 
@@ -372,7 +408,7 @@ namespace Microsoft.Build.Shared
                     else
                     {
                         // getFileSystemEntries(...) returns an empty array if longPath doesn't exist.
-                        string[] entries = getFileSystemEntries(FileSystemEntity.FilesAndDirectories, longPath, parts[i], null, false);
+                        ImmutableArray<string> entries = getFileSystemEntries(FileSystemEntity.FilesAndDirectories, longPath, parts[i], null, false);
 
                         if (0 == entries.Length)
                         {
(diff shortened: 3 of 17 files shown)


src/Shared/FileMatcher.cs
```diff
@@ -1548,14 +1630,30 @@ string fileToMatch
...
         {
+            var getFileSystemEntries = entriesCache == null
+                ? s_defaultGetFileSystemEntries
+                : (type, path, pattern, directory, projectDirectory) =>
+                {
+                    // Cache only directories, for files we won't ever hit the cache while open a project
+                    if (type == FileSystemEntity.Directories)
+                    {
+                        return entriesCache.GetOrAdd($"{path};{pattern ?? "*"}",
+                            s => new CachedEnumerable<string>(s_defaultGetFileSystemEntries(type, path, pattern,
```
radical: Given that `s_defaultGetFileSystemEntries` gets the whole array of entries in one go (https://github.com/Microsoft/msbuild/blob/master/src/Shared/FileMatcher.cs#L256), wouldn't it be better to use something like `ImmutableArray<T>`, initialize with the full array and avoid the locks and contention in `CachedEnumerable`?

maca88: The main idea of `CachedEnumerable` was to save one iteration through the `IEnumerable<string>` retrived by [Directory.EnumerateDirectory](https://github.com/Microsoft/msbuild/pull/2572/files#diff-64d47c5cbb59959b7b2fa6963672d596R291). After some testing I realized that there are no noticable differences by initialize the `IEnumerable` with `ToImmutableArray` method prior returning. I will remove the `CachedEnumerable` and cache an `ImmutableArray<string>` instead.

radical: Oh, but you are switching that to using `EnumerateDirectories` later so this doesn't matter.

radical: So, if you are reading all of it here itself then we can switch back to using `Directory.GetDirectories` and to `string[]` from `IEnumerable<string>`.

Also, it would be good to use `git rebase -i` to rework the commits whenever you want to change or fix something in them. That way they look like clean commits and makes it easier to review.

radical: Just thinking aloud - `Directory.EnumerateDirectories` should be helping the amount of memory being used at least, in case of *lot* of directories. Didn't it?

cdmihai:
Looking into the tests.

maca88:
The problem is with the usage of `File.GetFileName` that doesn't treat a backslash as a director separator on Unix, so `Path.GetFileName("Folder\MyFile.cs")` will return `"Folder\MyFile.cs"` instead of `"MyFile.cs"`. My plan is to add a custom `GetFileName` that will work the same on Windows and Unix. Also, I suspect that `Path.GetDirectoryName` [here](https://github.com/Microsoft/msbuild/pull/2572/files#diff-64d47c5cbb59959b7b2fa6963672d596R827) will not work in all cases, I will do some investigation tomorrow.

cdmihai:
I think the main issue here is the testing mock file system returns paths with backslashes in them, regardless of the OS. I'll try and make it normalize slashes.

maca88:
Fixed directory matching and flattern the `AggregateException` in case the current settings for the parallel traversing will be altered to have nested `Parallel.ForEach` calls. With the current settings only one `Parallel.ForEach` is used per `GetFiles` call, as [MaxTasksPerIteration](https://github.com/Microsoft/msbuild/pull/2572/files#diff-64d47c5cbb59959b7b2fa6963672d596R2141) is set to the same value as `MaxTasks`.

maca88:
There seems to be an inconsistent behavior with excludes and case sensibility. In the current master, when an exclude is matched via regex the matching is [case-insensitive](https://github.com/Microsoft/msbuild/blob/master/src/Shared/FileMatcher.cs#L1578) and when is matched with the [searchesToExcludeInSubdirs](https://github.com/Microsoft/msbuild/blob/master/src/Shared/FileMatcher.cs#L797) dictionary, is case-sensitive, as the dictionary [does not use](https://github.com/Microsoft/msbuild/blob/master/src/Shared/FileMatcher.cs#L1840) a case-insensitive comparer. For example, if we modify [this line](https://github.com/Microsoft/msbuild/blob/master/src/Shared/UnitTests/FileMatcher_Tests.cs#L1034) to `@"**\bIN\**\*.cs"` the test still succeeds, but if `@"src\Common\**"` is changed to `@"src\CoMMon\**"` the test fails. What should be the correct behavior? If case-insensitive is the answer, then I will modify the pattern matching algorithm to support it. Also, with this PR, an exclude like `@"src\Co??on\**"` will exclude files from `"src\Common"`, but in the in current master, it won't. Is this behavior desired, or should we change it in order to behave the same as in the current master?

cdmihai:
@maca88 

Regarding case sensitivity, I checked [the version](https://github.com/Microsoft/msbuild/blob/0c226f4938a0b75c0047586f7a4f371ff5cac7de/src/Shared/FileMatcher.cs) that shipped with VS 2015 Update 3. Back then the exclude specs were not sent to GetFiles, the disk got walked once for include, and once for each exclude, and the include file set was obtained by subtracting the excluded files from the included ones. The only matching that happened was the Regex one, which was case insensitive. So to keep with previous behaviour, both matches need to be case insensitive.

Regarding the double question mark match, master seems to work, it excludes `src/common/a.cs` with `src/co??on/**`:

```
e:\projects\tests\projects\play>tree /f /a

E:.
|   build.proj
|   MaxPathFinder.exe
|
+---Common
|       a.cs
|
\---src
    |   c.cs
    |
    +---Common
    |       a.cs
    |
    \---Comon
            b.cs


e:\projects\tests\projects\play>type build.proj
    <ItemGroup>
        <Compile Include="**\*.cs" Exclude="src\Co??on\**"/>
        <Compile2 Include="**\*.cs" Exclude="**\Co??on\**"/>
        <Compile3 Include="**\*.cs" Exclude="Co??on\**"/>
    </ItemGroup>

    <Target Name="Build">
        <Message Text=" Compile: @(Compile)"/>
        <Message Text="Compile2: @(Compile2)"/>
        <Message Text="Compile3: @(Compile3)"/>
    </Target>
</Project>

e:\projects\tests\projects\play>e:\projects\msbuild\bin\Bootstrap\MSBuild\15.0\Bin\MSBuild.exe
Microsoft (R) Build Engine version 15.5.98.54482 for .NET Framework

Build started 10/5/2017 3:25:12 PM.
Project "e:\projects\tests\projects\play\build.proj" on node 1 (default targets).
Build:
   Compile: Common\a.cs;src\c.cs;src\Comon\b.cs
  Compile2: src\c.cs;src\Comon\b.cs
  Compile3: src\c.cs;src\Common\a.cs;src\Comon\b.cs
Done Building Project "e:\projects\tests\projects\play\build.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.36
```

src/Shared/FileMatcher.cs
```diff
@@ -164,17 +169,39 @@ private static string[] GetAccessibleFilesAndDirectories(string path, string pat
...
-
-            return entries;
+            // NOTE: Windows matches loosely in three cases (in the absence of the * wildcard in the extension):
+            // 1) if the extension ends with the ? wildcard, it matches files with shorter extensions also e.g. "file.tx?" would
+            //    match both "file.txt" and "file.tx"
+            // 2) if the extension is three characters, and the filename contains the * wildcard, it matches files with longer
+            //    extensions that start with the same three characters e.g. "*.htm" would match both "file.htm" and "file.html"
+            // 3) if the ? wildcard is to the left of a period, it matches files with shorter name e.g. ???.txt would match
+            //    foo.txt, fo.txt and also f.txt
+            string extensionPart = Path.GetExtension(searchPattern);
```
davkean: Please don't allocate here unless needed, make sure this occurs only when if the search for "?." succeeds.

src/Shared/FileMatcher.cs
```diff
@@ -142,16 +147,16 @@ private static string[] GetAccessibleFileSystemEntries(FileSystemEntity entityTy
...
         {
-            string[] entries = null;
-
             if (Directory.Exists(path))
             {
                 try
                 {
-                    entries = Directory.GetFileSystemEntries(path, pattern);
+                    return ShouldEnforceMatching(pattern)
+                        ? Directory.EnumerateFileSystemEntries(path, pattern).Where(o => IsMatch(Path.GetFileName(o), pattern))
```
davkean: You need to think about exception semantics, no longer are exceptions isolated here - they can be thrown when the "enumerable" is iterated.

src/Shared/FileMatcher.cs
```diff
@@ -372,7 +412,7 @@ GetFileSystemEntries getFileSystemEntries
                     else
                     {
                         // getFileSystemEntries(...) returns an empty array if longPath doesn't exist.
-                        string[] entries = getFileSystemEntries(FileSystemEntity.FilesAndDirectories, longPath, parts[i], null, false);
+                        string[] entries = getFileSystemEntries(FileSystemEntity.FilesAndDirectories, longPath, parts[i], null, false).ToArray();
```
davkean: This can throw.

src/Shared/FileMatcher.cs
```diff
@@ -547,16 +587,20 @@ private static void PreprocessFileSpecForSplitting
...
         /// <param name="paths">Paths to remove .\ from.</param>
-        private static void RemoveInitialDotSlash
+        private static IEnumerable<string> RemoveInitialDotSlash
         (
-            string[] paths
+            IEnumerable<string> paths
         )
         {
-            for (int i = 0; i < paths.Length; i++)
+            foreach (string path in paths)
```
davkean: This can throw.

src/Shared/FileMatcher.cs
```diff
@@ -576,53 +620,58 @@ internal static bool IsDirectorySeparator(char c)
...
+        internal static IEnumerable<string> RemoveProjectDirectory
         (
-            string[] paths,
+            IEnumerable<string> paths,
             string projectDirectory
         )
         {
             bool directoryLastCharIsSeparator = IsDirectorySeparator(projectDirectory[projectDirectory.Length - 1]);
-            for (int i = 0; i < paths.Length; i++)
+            foreach (string path in paths)
```
davkean: This can throw.

src/Shared/FileMatcher.cs
```diff
@@ -1905,21 +2123,42 @@ DirectoryExists directoryExists
              * This is because it's cheaper to add items to an IList and this code
```
davkean: Comment is outdated.

src/Shared/FileMatcher.cs
```diff
@@ -1905,21 +2123,42 @@ DirectoryExists directoryExists
...
+                    searchesToExcludeInSubdirs,
+                    taskOptions);
+            }
+            // Catch exceptions that are thrown inside the Parallel.ForEach
+            catch (AggregateException ex)
+            {
+                // Flatten to get exceptions than are thrown inside a nested Parallel.ForEach
+                if (ex.Flatten().InnerExceptions.All(ExceptionHandling.IsIoRelatedException))
+                {
+                    return CreateArrayWithSingleItemIfNotExcluded(filespecUnescaped, excludeSpecsUnescaped);
```
davkean: Oh, is this to handle the exceptions? If so, why we do we still need the catch blocks above?

src/Shared/FileMatcher.cs
```diff
@@ -1905,21 +2123,42 @@ DirectoryExists directoryExists
...
             /*
              * Now get the files that match, starting at the lowest fixed directory.
              */
             try
             {
+                // Setup the values for calculating the MaxDegreeOfParallelism option of Parallel.ForEach
+                // Set to use only half processors when we have 4 or more of them, in order to not be too aggresive
+                // By setting MaxTasksPerIteration to the maximum amount of tasks, which means that only one
+                // Parallel.ForEach will run at once, we get a stable number of threads being created.
+                var maxTasks = Math.Max(1, Environment.ProcessorCount / 2);
```
davkean: @lifengl What do you think about this when it runs inside VS?

src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -44,6 +45,11 @@ protected LazyItemOperation(OperationBuilder builder, LazyItemEvaluator<P, I, M,
                 _itemSpec.Expander = _expander;
             }
 
+            /// <summary>
+            /// Cache used for caching IO operation results
+            /// </summary>
+            protected ConcurrentDictionary<string, IEnumerable<string>> EntriesCache => _lazyEvaluator._entriesCache;
```
davkean: @cdmihai Help me understand the lifetime of this cache? Just for a single Include?

cdmihai: it spans the entire item evaluation phase, so covers all items outside of targets

---
In reply to: [143077078](https://github.com/Microsoft/msbuild/pull/2572#discussion_r143077078) [](ancestors = 143077078)

src/Shared/FileMatcher.cs
cdmihai: All of these will be caught via the aggregate exception in GetFiles right?

---
In reply to: [143076202](https://github.com/Microsoft/msbuild/pull/2572#discussion_r143076202) [](ancestors = 143076202)

src/Shared/FileMatcher.cs
maca88: Ouch, I completely missed that. I will wrap the enumerator into a safe enumerator that will omit those two exceptions, in order to preserve the same behavior, prior this change.

src/Shared/FileMatcher.cs
maca88: Actually, they will be, but this is not correct as we don't want to return "empty" result just because one of the subdirectories, failed to get its files. The catching of the `AggreateException` is supposed to do the same as the  [other catch](https://github.com/Microsoft/msbuild/blob/master/src/Shared/FileMatcher.cs#L1924), when the exception is thrown inside a `Parallel.ForEach`, which will always throw an `AggreateException` containing the actual exception. The iteration over `paths` should not throw for `SecurityException` and `UnauthorizedAccessException` in order to preserve the old behavior, I will correct this.

src/Shared/FileMatcher.cs
maca88: Check the [comment](https://github.com/Microsoft/msbuild/pull/2572#discussion_r143182574) above for catching the `AggregateException`.

src/Shared/FileMatcher.cs
maca88: > no longer are exceptions isolated here

Sadly, this is not actually true, when `EnumerateFileSystemEntries`/`EnumerateFiles`/`EnumerateDirectories` methods are called, they can also throw because the file enumerator is initialized by finding the first file/directory and if an error occurs, the exception is [immediately thrown](https://github.com/dotnet/coreclr/blob/da8b7181dbe362fbfe64c7ef72ff3a4a7db68f24/src/mscorlib/src/System/IO/FileSystemEnumerable.cs#L217). So we have to handle exceptions when one of those three methods is called and on the MoveNext call.

maca88:
@cdmihai 

> So to keep with previous behaviour, both matches need to be case insensitive.

I will modify the algorithm to support case-insensitive matching as fast as possible, correct the initialization of `searchesToExcludeInSubdirs` dictionary to use a case-insensitive comparer and add tests for complex patterns using the real file system.

> Regarding the double question mark match, master seems to work

My bad, I didn't noticed that the error was due to the mock file system, which doesn't support complex patterns.

src/Shared/FileMatcher.cs
maca88: [Fixed](https://github.com/maca88/msbuild/blob/b5574289426c99e9567a43fe17d4731e1f4f5380/src/Shared/FileMatcher.cs#L259)

src/Shared/FileMatcher.cs
maca88: [Fixed](https://github.com/maca88/msbuild/blob/b5574289426c99e9567a43fe17d4731e1f4f5380/src/Shared/FileMatcher.cs#L221)

src/Shared/FileMatcher.cs
maca88: [Fixed](https://github.com/maca88/msbuild/blob/b5574289426c99e9567a43fe17d4731e1f4f5380/src/Shared/FileMatcher.cs#L321)

src/Shared/FileMatcher.cs
maca88: [Updated](https://github.com/maca88/msbuild/blob/b5574289426c99e9567a43fe17d4731e1f4f5380/src/Shared/FileMatcher.cs#L2188)

maca88:
@davkean

The exception handling should now work like before, with two exceptions:
1. If an exception is thrown inside a `Parallel.ForEach` and it is not handled by `ExceptionHandling.IsIoRelatedException`, then an `AggregateException` will be [thrown](https://github.com/maca88/msbuild/blob/b5574289426c99e9567a43fe17d4731e1f4f5380/src/Shared/FileMatcher.cs#L2227) instead of the actual exception/s.
2. By replacing the `Directory.GetXXX` with `Directory.EnumerateXXX` we may get one or more files/directories before the first `Security`/`UnauthorizedAccess` exception is thrown, in other words, if the third file is not accessible, with `Directory.EnumerateXXX` we would get the first two, but with `Directory.GetXXX` we wouldn't get any.

cdmihai:

Flaky restore

src/Shared/FileMatcher.cs
lifengl: That seems to be fine to me.  In dev 14 time, during the solution loading time, we use about 2.2 cores inside VS, and design time build uses about 1 core in the background.  However, in .net core project, after we disabled the async tree loading, and with very long time in the evaluation, it falls back to almost one core, especially during the project evaluation phase.  So this change will make the evaluation more efficient during that phase.

In the project editing phase, or project reloading time, we may have multiple evaluation in the background.  Ideally, if we can control that based on the priority of the evaluation (foreground or background), it will be better.  But it looks like this setting is not too aggressive.

src/Shared/FileMatcher.cs
```diff
@@ -372,7 +478,7 @@ GetFileSystemEntries getFileSystemEntries
                     else
                     {
                         // getFileSystemEntries(...) returns an empty array if longPath doesn't exist.
-                        string[] entries = getFileSystemEntries(FileSystemEntity.FilesAndDirectories, longPath, parts[i], null, false);
+                        string[] entries = getFileSystemEntries(FileSystemEntity.FilesAndDirectories, longPath, parts[i], null, false).ToArray();
```
lifengl: Minor: ToList() is usually slightly faster, (they are almost same to copy data to a buffer, but ToArray() often copies one more time to fit the exact size.

lifengl:
That is really nice work, @maca88.  I saw so many traces that the slow globbing code consumed 10 seconds or even 30 seconds in some projects.  This change will make a huge difference.

src/Build.UnitTests/Microsoft.Build.Engine.UnitTests.csproj
```diff
@@ -249,6 +249,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup Condition="'$(NetCoreBuild)' != 'true'">
+    <Reference Include="System.Collections.Immutable, Version=1.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```
maca88: I've had to add this reference in order to build the project. Does anyone know why?

src/Shared/FileMatcher.cs
maca88: With the [recent change](https://github.com/Microsoft/msbuild/pull/2572/commits/1322acf14ed4e26a88bd3bcc927ea13b94426dd0), the `ToArray` is not needed anymore.

src/Shared/FileMatcher.cs
```diff
@@ -705,128 +826,183 @@ private static void GetFilesRecursive
...
-                excludeNextSteps = new RecursiveStepResult[searchesToExclude.Count];
-                for (int i = 0; i < searchesToExclude.Count; i++)
+                if (excludeNextSteps == null)
                 {
-                    excludeNextSteps[i] = GetFilesRecursiveStep(
-                        searchesToExclude[i],
-                        projectDirectory,
-                        stripProjectDirectory,
-                        getFileSystemEntries);
+                    files = files ?? new List<string>();
```
lifengl: you can adjust the logic here slightly to prevent the duplicated code (line 845-847) and (line 861-862)

src/Shared/FileMatcher.cs
```diff
@@ -705,128 +826,183 @@ private static void GetFilesRecursive
...
-                        listOfFiles.Add(file);
-                    }
+                    files = files ?? new List<string>();
+                    files.Add(file);
                 }
             }
+            // Add all matched files at once to reduce thread contention
+            if (files != null && files.Any())
+            {
+                listOfFiles.PushRange(files.ToArray());
```
lifengl: Instead of copying the list to an array, and then create a link list node for each of them (inside the stack code), I think it can be Stack<List<string>>(), so you just push the non-empty list there.  In the code consuming it, it will be:
stack.SelectMany(list => list).xxxx..., so we can prevent the extra allocation, in case the list is big.

src/Build.UnitTests/Microsoft.Build.Engine.UnitTests.csproj
cdmihai: I think it's because ImmutableArray got exposed to the tests (FileMatcher_Tests, via the new cache). Then FileMatcher is included in two test dlls.

maca88: I understand that the two test dlls need the dependency, what I don't understand is that adding the dependency to the `project.json` don't work for `Engine.UnitTests`, but [it works](https://github.com/Microsoft/msbuild/pull/2572/commits/1322acf14ed4e26a88bd3bcc927ea13b94426dd0#diff-c06dddefcbd40b425acebaac4b74d20eR15) for the `Tasks.UnitTests` project. Its seems like that one of the `Engine.UnitTests` dependency needs a lower version of the Immutable dll, but from the logs it is not clear which one.

src/Shared/FileMatcher.cs
maca88: [Done](https://github.com/Microsoft/msbuild/pull/2572/commits/b5e4a47c725f3dbbec16b3fda6af9fd69175ea58)

src/Shared/FileMatcher.cs
maca88: [Done](https://github.com/Microsoft/msbuild/pull/2572/commits/b5e4a47c725f3dbbec16b3fda6af9fd69175ea58)

AndyGerlicher:

src/Shared/FileMatcher.cs
```diff
@@ -705,128 +758,181 @@ private static void GetFilesRecursive
...
-                        stripProjectDirectory,
-                        getFileSystemEntries,
-                        newSearchesToExclude,
-                        searchesToExcludeInSubdirs);
+            // Calcuate the MaxDegreeOfParallelism value in order to prevent too much tasks being running concurrently.
+            int dop = 0;
+            // Lock only when we may be dealing with multiple threads
+            if (taskOptions.MaxTasks > 1 && taskOptions.MaxTasksPerIteration > 1)
+            {
+                lock (s_directoryLock)
```
davkean: Can you explain what you are protecting against?

radical: .. and rename the variable to reflect that. (s_taskOptionsLock?)

maca88: This lock prevents multiple threads that were created from different `Parallel.ForEach` calls from modifying the `dop` variable in order to assure, that we are not going to use more tasks that have been specified in `TaskOptions`. We have to consider that this method is recursive, which means that we could have one `Parallel.ForEach` that may call another `Parallel.ForEach` inside its iteration. [I've added](https://github.com/Microsoft/msbuild/pull/2572/commits/0fec532e6551c12008ed12b0dc223736edf354a0) an additional condition for locking, as with the current settings the lock is not needed. To demonstrate how this works, I've created [a simulation](https://dotnetfiddle.net/GVvg02), where `TaskOptions` is configured to have up to 4 `Parallel.ForEach` running concurrently. By commenting the `lock` statement, we can see that after several runs, we may get 5 `Parallel.ForEach` running concurrently.

davkean: 👍

davkean: Can we just lock the taskoptions instead of creating another obhect?

maca88: Good idea! By doing so, we will have the locking per call, which is more correct than having it static, as the purpose of the locking is just to modify the `TaskOptions`.

maca88:
@cdmihai

I have corrected the case sensitivity behavior, so that everything is now matched case-insensitive. I've added a unit test for each correction related to matching excludes, which would fail on the current master. The overall performance after adding the case-insensitive matching remains more or less the same as before.

AndyGerlicher:
@dotnet-bot test this please

cdmihai:
@maca88 

> There are some specific cases when an include or exclude pattern won't be matched case-insensitive on Unix. By quick checking the version that shipped with VS 2015 Update 3, seems that it has the same problems. Should we correct them or should we leave as it is, for backward compatibility?

Leave them as is, to reduce scope creep :) I'll open an issue linking to your test once this PR gets in.

cdmihai:
FYI, we're planning to merge this in tomorrow morning (10am PDT). @maca88 can you please squash your changes if you get the chance? (If not I'll do it, I just need to make sure you are preserved as commit author)

davkean:
@maca88 can you please contact me offline davkean@microsoft.com?

maca88:
@cdmihai 
I've squashed my changes into one commit on top of the mock file system correction.

AndyGerlicher:
@maca88 this is a very substantial contribution, probably the biggest we've had on MSBuild outside Microsoft. Thank you so much for seeing this through and addressing the numerous comments and iterations! This will definitely be a big part of the perf gains we'll get out of 15.5!

JesperTreetop:
> My plan is to add a custom GetFileName that will work the same on Windows and Unix. 

I just came here to ooh and aah at the performance improvements but I really hope [the quoted part](https://github.com/Microsoft/msbuild/pull/2572#issuecomment-333973915) (in an earlier comment by @maca88) wasn't implemented because there are definitely Unix file systems where you are allowed to have backslashes in filenames, and trying to turn them into slashes could be a logic error at best and point to completely the wrong file at worst. There's definitely security consequences.

maca88:
@JesperTreetop the quoted part wasn't implemented, I updated the missleading part of the comment in order to avoid confusion.

PR #2577, opened 2017-10-03
Avoid iterator/list allocations when parsing projects

davkean:
For each project and for every container construct in a project (ItemGroup, ImportGroup, etc) we were allocating a `List<XmlElementWithLocation>` and a `XmlChildEnumerator`. In large solutions with large import graphs this was adding up to a non-trivial amount of allocations (1.7% of all allocations during evaluation in one trace).

Replaces these with a struct-based enumerable/enumerator that manually walks the child elements.

### DIFF
--- a/src/Build/Xml/ProjectXmlUtilities.XmlElementChildIterator.cs
+++ b/src/Build/Xml/ProjectXmlUtilities.XmlElementChildIterator.cs
@@ -1,0 +1,106 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Xml;
+using Microsoft.Build.Construction;
+
+namespace Microsoft.Build.Internal
+{
+    partial class ProjectXmlUtilities
+    {
+        // Iterating an element's nodes allocates an non-trivial amount of data in certain
+        // large solutions with lots of targets, so we have our own struct-based iterator
+        // that avoids unneeded GC pressure.
+        //
+        // Deliberately not implement IEnumerable/IEnumerator to avoid accidental boxing
+        internal struct XmlElementChildIterator
+        {
+            private readonly XmlElementWithLocation _element;
+            private readonly bool _throwForInvalidNodeTypes;
+            private XmlElementWithLocation _current;
+            private bool _isFirst;
+
+            internal XmlElementChildIterator(XmlElementWithLocation element, bool throwForInvalidNodeTypes)
+            {
+                _element = element;
+                _throwForInvalidNodeTypes = throwForInvalidNodeTypes;
+                _current = null;
+                _isFirst = true;
+            }
+
+            public bool MoveNext()
+            {
+                if (_isFirst)
+                {
+                    _isFirst = false;
+                    _current = GetNextNode(_element.FirstChild);
+                }
+                else if (_current != null)
+                {
+                    _current = GetNextNode(_current.NextSibling);
+                }
+
+                return _current != null;
+            }
+
+            public XmlElementChildIterator GetEnumerator()
+            {
+                return this;
+            }
+
+            public XmlElementWithLocation Current
+            {
+                get
+                {
+                    if (_isFirst || _current == null)
+                        throw new InvalidOperationException();
+
+                    return _current;
+                }
+            }
+
+            public void Reset()
+            {
+                throw new NotSupportedException();
+            }
+
+            private XmlElementWithLocation GetNextNode(XmlNode child)
+            {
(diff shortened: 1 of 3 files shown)


davkean:
This is the allocations in one trace I looked at:

image

cdmihai:

davkean:
Here are the numbers:

DotnetConsoleProject | Evaluation: Time (ms) | :ok_hand: | no | 34.7853 -> 34.844 (0.169%)
DotnetConsoleProject | Evaluation: Memory | :white_check_mark: | yes | 5715452 -> 5566833 (-2.6%)
DotnetWebProject | Evaluation: Time (ms) | :ok_hand: | no | 49.8234 -> 49.7828 (-0.081%)
DotnetWebProject | Evaluation: Memory | :white_check_mark: | yes | 7450200 -> 7281210 (-2.268%)
DotnetMvcProject | Evaluation: Time (ms) | :ok_hand: | no | 84.5729 -> 84.2088 (-0.431%)
DotnetMvcProject | Evaluation: Memory | :white_check_mark: | yes | 9734911 -> 9566142 (-1.734%)
Picasso | Evaluation: Time (ms) | :ok_hand: | no | 319.7085 -> 320.2459 (0.168%)
Picasso | Evaluation: Memory | :white_check_mark: | yes | 44420561 -> 43906740 (-1.157%)
SmallP2POldCsproj | Evaluation: Time (ms) | :white_check_mark: | yes | 54.113 -> 53.7609 (-0.651%)
SmallP2POldCsproj | Evaluation: Memory | :white_check_mark: | yes | 7557055 -> 7426286 (-1.73%)
Generated_100_100_v150 | Evaluation: Time (ms) | :ok_hand: | no | 1504.9799 -> 1506.4059 (0.095%)
Generated_100_100_v150 | Evaluation: Memory | :white_check_mark: | yes | 187470244 -> 185709264 (-0.939%)
LargeP2POldCsproj | Evaluation: Time (ms) | :white_check_mark: | yes | 970.1681 -> 964.5588 (-0.578%)
LargeP2POldCsproj | Evaluation: Memory | :white_check_mark: | yes | 110015017 -> 109549813 (-0.423%)
roslyn | Evaluation: Time (ms) | :red_circle: | yes | 5399.7462 -> 5430.1103 (0.562%)
roslyn | Evaluation: Memory | :ok_hand: | no | 569358676 -> 574803028 (0.956%)

src/Build/Xml/ProjectXmlUtilities.XmlElementChildIterator.cs
```diff
@@ -0,0 +1,106 @@
...
+                get
+                {
+                    if (_isFirst || _current == null)
+                        throw new InvalidOperationException();
+
+                    return _current;
+                }
+            }
+
+            public void Reset()
```
dsplaisted: Do you need to have a Reset method if you're not implementing the interfaces?  Is Reset part of the pattern that foreach looks for?

src/Build/Xml/ProjectXmlUtilities.XmlElementChildIterator.cs
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    _current = GetNextNode(_current.NextSibling);
+                }
+
+                return _current != null;
+            }
+
+            public XmlElementChildIterator GetEnumerator()
+            {
+                return this;
```
dsplaisted: This may be a standard pattern, but is it correct to have the enumerable and the enumerator be the same object?  It seems like the contract would be that you can call GetEnumerator() multiple times on the same enumerable and you should get different enumerators / iterators with different state.

src/Build/Xml/ProjectXmlUtilities.XmlElementChildIterator.cs
davkean: Good point, I'll check.

src/Build/Xml/ProjectXmlUtilities.XmlElementChildIterator.cs
davkean: It's a pattern used by LINQ- http://index/?query=Enumerable.Select&rightProject=System.Core&file=System%5CLinq%5CEnumerable.cs&line=90. I was a little worried about increasing the size to handle this case (because it will never be hit), but I noticed they use a "state" field to track this and I might do the same thing.

davkean:
@dsplaisted I addressed your comments, changes:

- I assert the state machine - I make it illegal to double iterate, if we need to do that it wouldn't be too hard to clone the iterator.
- I removed Reset, it's not needed.
- I fixed up the check for illegal children - it was using the wrong child for errors.

davkean:
tag @Pilchie

PR #2579, opened 2017-10-03
Use reflection to find CultureInfo.GetCultures

### LINKED ISSUE #2349: How can I build a .NET Core project against custom locales using msbuild?
I want to build `*.resx` for locale that platform does not support in default. With the help of @tarekgh and @jefgen I installed custom locale on windows, and I succeeded in building `<custom-locale>.resx` with VS2017 (https://github.com/dotnet/corefx/issues/22135). 

Now I need to build it on linux using dotnet cli. 

Looking at the msbuild code quickly, https://github.com/Microsoft/msbuild/blob/master/src/Tasks/CultureInfoCache.cs#L31, it seems that msbuild currently only uses the `hardcoded` locale names for coreclr.

How can we solve this situation?

In my guess, the simplest way is to (a) add all the locales I need to the cache, and the other way is (b) to make msbuild to get the locale information from the platform for coreclr as well. But, to do so, it is likely that msbuild should be modified to target the `netstandard2.0`.

Is my guess right? Which way would be better?

cdmihai:
Closes #2349

### DIFF
--- a/src/Shared/AssemblyUtilities.cs
+++ b/src/Shared/AssemblyUtilities.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Globalization;
+using System.Linq;
 using System.Reflection;
 
 namespace Microsoft.Build.Shared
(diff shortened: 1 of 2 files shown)


cdmihai:
Copied the msbuild binaries under dotnet CLI and manually checked that the method gets called right.

lemmaa:
@cdmihai  Thank you!

PR #2580, opened 2017-10-03
Set document full path sooner to reduce allocations

jeffkl:
Building a single project I saw 17,537 `ElementLocation` objects being created.  It turns out that code in `XmlElementWithLocation` and `XmlAttributeWithLocation` was making new objects because the document full path is not set in time.

This change sets the document full path sooner.  This reduced the number of ElementLocation objects being created from 17,537 to just 9,237 (47% less)

FYI @davkean

### DIFF
--- a/src/Build/Construction/ProjectRootElement.cs
+++ b/src/Build/Construction/ProjectRootElement.cs
@@ -2093,7 +2093,11 @@ namespace Microsoft.Build.Construction
         {
             ErrorUtilities.VerifyThrowInternalRooted(fullPath);
 
-            var document = new XmlDocumentWithLocation {PreserveWhitespace = preserveFormatting};
+            var document = new XmlDocumentWithLocation
+            {
+                FullPath = fullPath,
+                PreserveWhitespace = preserveFormatting
+            };
 #if (!STANDALONEBUILD)
             using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildProjectLoadFromFileBegin, CodeMarkerEvent.perfMSBuildProjectLoadFromFileEnd))
 #endif
@@ -2110,7 +2114,6 @@ namespace Microsoft.Build.Construction
                         document.Load(xtr.Reader);
                     }
 
-                    document.FullPath = fullPath;
                     _projectFileLocation = ElementLocation.Create(fullPath);
                     _directory = Path.GetDirectoryName(fullPath);
 


jeffkl:

src/Build/ElementLocation/XmlDocumentWithLocation.cs
```diff
@@ -115,7 +115,20 @@ private enum ReadOnlyLoadFlags
...
+            {
+                if (!FullPathHasChanged && !String.Equals(_fullPath, value, StringComparison.OrdinalIgnoreCase))
+                {
+                    FullPathHasChanged = true;
+                }
+                _fullPath = value;
+            }
+        }
+
+        internal bool FullPathHasChanged
```
davkean: It seems to me that once you "SaveAs" you pay the cost of checking the full path forever?

jeffkl: At the moment, you pay the price always.  With my change, projects loaded from disk no longer create a new ElementLocation every time.  In-memory projects that are saved will pay the cost once.

src/Build/ElementLocation/XmlElementWithLocation.cs
```diff
@@ -91,7 +91,8 @@ internal ElementLocation Location
             {
                 // Caching the element location object saves significant memory
                 XmlDocumentWithLocation ownerDocumentWithLocation = (XmlDocumentWithLocation)OwnerDocument;
-                if (!String.Equals(_elementLocation.File, ownerDocumentWithLocation.FullPath, StringComparison.OrdinalIgnoreCase))
+
+                if (ownerDocumentWithLocation.FullPathHasChanged && !String.Equals(_elementLocation.File, ownerDocumentWithLocation.FullPath, StringComparison.OrdinalIgnoreCase))
```
davkean: Are these equals showing up? I'd expect this to be a reference equality check in the equal case?

jeffkl: Without my change, `_elementLocation.File` is `String.Empty` for all elements and attributes while `OwnerDocument.FullPath` is `null` or a path.

With my change, we now check a bool first to see if we should compare the paths second.  And again, if the files were loaded from disk and haven't been renamed, the string comparison won't happen anymore and the `ElementLocation` objects won't be recreated

jeffkl:
@dotnet-bot test this please

src/Build/ElementLocation/XmlElementWithLocation.cs
davkean: I'm not sure I follow, under what circumstances would FullPath be different from ElementLocation.File and you wouldn't want to recreate the element location?

jeffkl: If they are different, you definitely want to recreate ElementLocation.  However, without my change, we weren't setting FullPath prior to loading the document so all ElementLocations would have no `File` and then after loading, we'd set FullPath and after that any access to `Location` would recreate the `ElementLocation`.  If we set `FullPath` first and the file path doesn't change, the `ElementLocation` doesn't get recreated anymore.

The optimized case is evaluating a project from disk, it will no longer recreate ElementLocation the first time you access the property (which happens all over the place).

davkean: I get the first change - what I don't get is why you need the FullPathHasChanged check? Is that indicated by the fact that the owner document and my "file" are different?

cdmihai: I think the only purpose of FullPathHasChanged is to avoid the string equals in the location getters, in the case where full path does not change. If the full path changes, `FullPathHasChanged` gets stuck to true and all subsequent location getters get to execute the getters.

cdmihai:

src/Build/ElementLocation/XmlElementWithLocation.cs
davkean: Yeah we talked offline, it's to avoid the string comparison. In the traces I saw this isn't showing up *at all* in CPU time. This isn't doing a full string comparison, it's almost always just a reference equality check, and hence I think it can be removed.

jeffkl: I have confirmed that `_elementLocation.File` and `ownerDocumentWithLocation.FullPath` are the same reference so when loading projects from disk this check adds trivial overhead.  I've reverted that commit.

AndyGerlicher:
Perf test looks good!

### Evaluation: Time (ms)
:--- | :--- | :--- | :---
DotnetConsoleProject | ::ok_hand: | no | 26.7886 -> 26.7749 (-0.051%)
DotnetWebProject | :white_check_mark: | yes | 41.1882 -> 41.0733 (-0.279%)
DotnetMvcProject | :white_check_mark: | yes | 78.5604 -> 78.277 (-0.361%)
Picasso | :white_check_mark: | yes | 229.9646 -> 228.0485 (-0.833%)
SmallP2POldCsproj | :red_circle: | yes | 35.3142 -> 35.7567 (1.253%)
Generated_100_100_v150 | :white_check_mark: | yes | 1370.3112 -> 1334.3888 (-2.621%)
LargeP2POldCsproj | :white_check_mark: | yes | 629.7093 -> 619.125 (-1.681%)
roslyn | :red_circle: | yes | 5113.0268 -> 5131.6278 (0.364%)

:--- | :--- | :--- | :---
DotnetConsoleProject | :white_check_mark: | yes | 4475598 -> 4366933 (-2.428%)
DotnetWebProject | :white_check_mark: | yes | 5944325 -> 5829216 (-1.936%)
DotnetMvcProject | :white_check_mark: | yes | 8204810 -> 8087376 (-1.431%)
Picasso | :white_check_mark: | yes | 30716322 -> 30444251 (-0.886%)
SmallP2POldCsproj | :white_check_mark: | yes | 4878339 -> 4793840 (-1.732%)
Generated_100_100_v150 | :white_check_mark: | yes | 164364704 -> 163511058 (-0.519%)
LargeP2POldCsproj | :white_check_mark: | yes | 64470631 -> 64229667 (-0.374%)
roslyn | :white_check_mark: | yes | 560865733 -> 560474327 (-0.07%)

Issue #2561, opened 2017-09-27
Large amount of allocations during evaluation due to boxing of KeyedObject


davkean:
The call to RetrievableHashSet.Get(string) and RetrievableHashSet.Get(string, int, int) is boxing a huge amount of KeyedObject instances:

![image](https://user-images.githubusercontent.com/1103906/30942545-2274d59a-a42f-11e7-8b9d-454baa0f1a13.png)

These are boxed, only to be compared as a "string" anyway. We should avoid the boxing if the enumerator already implements IConstraintedEqualityComparer<string>/IEqualityComparer<string>.

davkean:
Here it is consuming 1.4% (348 MB):

![image](https://user-images.githubusercontent.com/1103906/31112888-306052a6-a862-11e7-977d-14846b1c7b9b.png)

PR #2581, opened 2017-10-04
Avoid boxing KeyedObject 

davkean:
Fixes: #2561

We were boxing/allocating KeyedObject on every string lookup to RetrievableEntryHashSet, resulting in 0.4% -> 1.4% of allocations during a build. 

To avoid that, changed RetrievableEntryHashSet to take `IEqualityComparer<string>` instead of `IEqualityComparer<IKeyed>`. Effectively, we were comparing strings underneath anyway - just moved to always comparing them so instead of `string` -> `IKeyed` we go `IKeyed` -> `string`. This removed all allocations on lookups to this collection.

This reduces allocations by 1.1% in a mixed solution containing .NET Standard/.NET Framework projects, 1.4% in a major partner's solution.

tag @pilchie


src/Build/Collections/RetrievableEntryHashSet/HashSet.cs
```diff
@@ -592,8 +574,8 @@ public virtual void OnDeserialization(Object sender)
             }
 
             int capacity = _siInfo.GetInt32(CapacityName);
-            _comparer = (IEqualityComparer<IKeyed>)_siInfo.GetValue(ComparerName, typeof(IEqualityComparer<IKeyed>));
-            _constrainedComparer = _comparer as IConstrainedEqualityComparer<IKeyed>;
+            _comparer = (IEqualityComparer<string>)_siInfo.GetValue(ComparerName, typeof(IEqualityComparer<string>));
```
davkean: Note these types are serializable - so this breaks the serialization format, we do we expect this to be round tripped with other MSBuild versions?

davkean:
Before allocations:
image

After allocations:
image

I ran the script - but I'm having trouble reducing noise, and don't really trust the numbers - can someone else buddy test this?

DotnetConsoleProject | Evaluation: Time (ms) | :ok_hand: | no | 25.515 -> 25.7422 (0.89%)
DotnetConsoleProject | Evaluation: Memory | :white_check_mark: | yes | 4613228 -> 4501474 (-2.422%)
DotnetWebProject | Evaluation: Time (ms) | :ok_hand: | no | 37.5473 -> 37.678 (0.348%)
DotnetWebProject | Evaluation: Memory | :white_check_mark: | yes | 6041199 -> 5915877 (-2.074%)
DotnetMvcProject | Evaluation: Time (ms) | :red_circle: | yes | 70.27 -> 71.2918 (1.454%)
DotnetMvcProject | Evaluation: Memory | :white_check_mark: | yes | 8303910 -> 8168895 (-1.626%)
Picasso | Evaluation: Time (ms) | :red_circle: | yes | 217.4518 -> 221.2058 (1.726%)
Picasso | Evaluation: Memory | :white_check_mark: | yes | 31965154 -> 30774454 (-3.725%)
SmallP2POldCsproj | Evaluation: Time (ms) | :ok_hand: | no | 34.7612 -> 35.0058 (0.704%)
SmallP2POldCsproj | Evaluation: Memory | :white_check_mark: | yes | 5107335 -> 4919683 (-3.674%)
Generated_100_100_v150 | Evaluation: Time (ms) | ::ok_hand: | no | 1208.8444 -> 1206.1295 (-0.225%)
Generated_100_100_v150 | Evaluation: Memory | :white_check_mark: | yes | 170179508 -> 162790513 (-4.342%)
LargeP2POldCsproj | Evaluation: Time (ms) | :white_check_mark: | yes | 615.7841 -> 607.5492 (-1.337%)
LargeP2POldCsproj | Evaluation: Memory | :white_check_mark: | yes | 68907375 -> 65464188 (-4.997%)
roslyn | Evaluation: Time (ms) | ::ok_hand: | no | 4490.2411 -> 4473.2247 (-0.379%)
roslyn | Evaluation: Memory | :red_circle: | yes | 499197850 -> 509861789 (2.136%)

Even when I was testing the same branch side-by-side I was seeing regressions/improvements.

src/Build/Collections/RetrievableEntryHashSet/HashSet.cs
AndyGerlicher: Shouldn't be. There are a few instances where caches are created and read in between builds so the format is important. I think I've added a version field to those now to invalidate them, so shouldn't be a big deal. But I don't think this is used in any of those (I will double check). For the main usage of binary serialization there's a handshake takes place that will ensure it's running exactly the same version of MSBuild on both ends of the pipe.

AndyGerlicher:
I have a dedicated physical machine running Server SKU with as much disabled as I can and I still get differences between my A/B versions even when they're actually the same commit. But usually the significance is low (P < .05 using T-Test) or it's really minor (~1ms hard to really say either way). For memory, I added `GC.GetTotalMemory()` before and after running. I think it's decent on the small projects but I don't know how much stock to put in that on the bigger projects like Roslyn. I don't believe it would be very accurate when GCs are happening. 

Just from the numbers though I would say this looks good (significant memory decrease). I've got another run going I'll try this out after that one finishes.

PR #2582, opened 2017-10-04
Initialize collections with up-front counts

davkean:
We already know how many elements these collections are going to contain, size them accordingly.

tag @pilchie


davkean:
It's really scary how many of these "properties" on these objects that I'm looking at allocate...

PR #2585, opened 2017-10-04
add F# support for the WriteCodeFragment task

### LINKED ISSUE #2584: WriteCodeFragment should generate code for F#

brettfo:
Enables the `<WriteCodeFragment />` task for F#.

Since F# doesn't have CodeDOM support, both the CodeDOM and CoreCLR branches end up calling the same function to do the generation.  This should enable things like assembly name/version attributes to be generated.

Due to some oddities in how F# code must be structured, there are two things to point out in the generated code:

1.  Files must (not really, but it's complicated) start with a `namespace` declaration.  This value doesn't matter and won't end up in the compiled assembly, so the value of `FSharp` was selected.
2.  F# files that contain no types/modules must have some executable code and the F# way of `do`ing this is to add a `do()` expression at the bottom of the file.

Otherwise the code generated for CoreCLR C# was used as the model and a VB-specific test was copied.

Fixes #2584.


AndyGerlicher:
This seems like it should be in the F# SDK? Ideally even C# / VB wouldn't be here either, but I'd at least rather not expand when possible. Any reason it couldn't go there?

brettfo:
No reason this couldn't go in the F# SDK; I was just mimicking what C#/VB does.  I'd be happy to have this PR closed and put it there if that's what you'd prefer.

brettfo:
Closing in favor of Microsoft/visualfsharp#3702

PR #2588, opened 2017-10-05
Remove unnecessary List<T> instantiation

davkean:
This was the cause of a large amount (1.2%) of allocations in a large project, some of which end up on the large object heap.

Large project from a partner:
image

Smaller project with mix of .NET Framework/.NET Standard projects:
image

### DIFF
--- a/src/Build/Evaluation/LazyItemEvaluator.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.cs
@@ -401,7 +401,7 @@ namespace Microsoft.Build.Evaluation
             return ret;
         }
 
-        public IEnumerable<ItemData> GetAllItems()
+        public IEnumerable<ItemData> GetAllItemsDeferred()
         {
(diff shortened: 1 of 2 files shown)


davkean:
tag @pilchie

src/Build/Evaluation/LazyItemEvaluator.cs
```diff
@@ -401,13 +401,10 @@ private LazyItemList GetItemList(string itemType)
             return ret;
         }
 
-        public IList<ItemData> GetAllItems()
+        public IEnumerable<ItemData> GetAllItems()
```
Pilchie: Might be worth at least commenting that this is deferred, so that callers can be careful about multiple enumerations.  Or consider changing the name to make it more clear.

davkean: I'll change the name.

PR #2592, opened 2017-10-05
Ensure ResolveProjectReferences runs before RAR

### LINKED ISSUE #2591: Intellisense missing on project references (design-time build) if you have a COM reference
An internal team reported an issue where referenced types were red underlined in Visual Studio but the build (command-line or VS) reported no errors. All projects in the sln were new SDK projects (most multi targeting).

image

AndyGerlicher:
Fixes a case where RAR was called early in the build before project
references were resolved. As far as I know this only happened in
design-time builds, but this makes the dependency explicit (if
ResolveProjectReferences has not run all project dependencies will not
be correctly resolved in RAR).

Fixes #2591

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2046,6 +2046,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     -->
   <PropertyGroup>
     <ResolveAssemblyReferencesDependsOn>
+      ResolveProjectReferences;
+      FindInvalidProjectReferences;
       GetFrameworkPaths;
       GetReferenceAssemblyPaths;
       PrepareForBuild;


AndyGerlicher:

davkean:
Shouldn't FindInvalidProjectReferences also be running to unresolve them for invalid references?

AndyGerlicher:
@davkean I confirmed it does normally and it should, and doesn't in the case I looked at. Not convinced this is the right place to set this, but it does have the desired effect. And those targets will run before `ResolveAssemblyReferences` anyway so this should not affect performance.

davkean:
I wonder if this is the cause of https://github.com/dotnet/project-system/issues/2735?

davkean:
Yep, looks like this is the cause of https://github.com/dotnet/project-system/issues/2735.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1939,6 +1939,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     -->
   <PropertyGroup>
     <ResolveAssemblyReferencesDependsOn>
+      ResolveProjectReferences;
+      FindInvalidProjectReferences;
+      PrepareProjectReferences;
```
davkean: I don't understand why PrepareProjectReferencs is listed after "ResolveProjectReferences" - it's a dependency of ResolveProjectReferences.

rainersigwald: I'd also rather avoid listing it redundantly.

davkean:
@AndyGerlicher Can we have this picked up again? We're seeing this being hit more and more: https://github.com/dotnet/project-system/issues/3233 and https://github.com/dotnet/project-system/issues/3211.

AndyGerlicher:

davkean:
Thanks!

Issue #2064, opened 2017-05-05
Build error MSB4067 (UnrecognizedChildElement) when building a solution with rptproj or dwproj projects


AndrewBennet:
When using MSBuild 15.0, cleaning or building a solution which contains rptproj or dwproj projects gives the following error:

> Reports.rptproj(3,3): error MSB4067: The element <State> beneath element <Project> is unrecognized.

When using MSBuild 14.0, the behaviour is better: it logs warnings (MSB4078) about unsupported projects:

> warning MSB4078: The project file "Reports.rptproj" is not supported by MSBuild and cannot be built.

The latter behaviour is more desirable (for me, at least): I would like to able to build a solution in a configuration which includes the .rptproj files. This would allow me to include these projects when building the solution in Visual Studio, and to be able to use the same configuration when building the solution via an MSBuild script.

Is this change of behaviour a bug? The message certainly seems like it is attempting to parse the rptproj file into some structure to which it doesn't belong. If it is not, is there a way to downgrade the error MSB4067 to a warning, or to skip certain projects when building a solution? The `/ignoreprojectextensions:.rptproj` option does not prevent the error.

Note: this issue relates to [this Stack Overflow post](https://stackoverflow.com/questions/43735390/msbuild-15-0-visual-studio-2017-throws-error-msb4067-for-solutions-with-ssdt-p).

DaveSlinn:
I recently added our .rptproj back into our solution file once I saw the availability of the Reporting Services extension for VS 2017.  Our solution builds fine on the dev boxes, but is broken on our build server - even though I have installed the Reporting Services Projects on the build machine as well.  It's concerning that no one from the team has commented for almost 2 months.

MgSam:
This is a big breaking change. We too now have broken solutions on our build machines because of this. Can someone please triage?

rainersigwald:
Thanks for the report!

MSBuild has [some code](https://github.com/Microsoft/msbuild/blob/07c3b3392dc05e0e3aba18bdec235a374aa72301/src/Build/Construction/Solution/ProjectInSolution.cs#L279-L362) that attempts to detect whether projects referred to from the solution are MSBuild projects or some other format.

[...]

PR #2593, opened 2017-10-05
Check schema when performing a sln build.

AndyGerlicher:
Note this doesn't fix the `.dwproj` case. I need to figure out something slightly different for that, but I wanted to send this out because I need to move on for now.

* Check that the schema matches either empty or the MSBuild default schema
when building a solution file.

* Special case .rptproj files. This file format looks like a standard
MSBuild file but specifies ToolsVersion=2.0 and a default XML schema.
Since the XML parser treats that schema as equivalent to empty and empty
is now allowed, there's no good way to determine if it's an MSBuild file
until we try to parse. This results in an error in MSBuild 15 rather
than a warning in MSBuild 14 for solutions with a .rptproj file. To
solve this, the sln build will now allow projects where the schema is
the default MSBuild schema or empty and ToolsVersion is not 2.0.

Partial fix for #2064


jeffkl:
@AndyGerlicher should this fix still go in now that this type of project is supported?

https://github.com/Microsoft/msbuild/issues/2064#issuecomment-336843520

PR #2594, opened 2017-10-05
Cleanup bootstrap build

AndyGerlicher:
Needed for #2595 

* Download NuGet.Build.Tasks from package (myget feed).
* Cleanup bootstrap. Avoid copying BuildTools blindly into the output folder.
* Correct some path properties. A lot of things just worked because we copied so much.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -11,7 +11,7 @@
     <add key="nuget.org v2" value="https://nuget.org/api/v2/" />
     <add key="myget.org roslyn nightly" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
     <add key="dotnet.myget.org buildtools v3" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
-    <!-- <add key="myget.org nugetbuild" value="https://www.myget.org/F/nugetbuild/api/v3/index.json" /> -->
+    <add key="myget.org nugetbuild" value="https://dotnet.myget.org/F/nuget-build/api/v3/index.json" />
     <add key="dotnet.myget.org dotnet-core v3" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="myget.org aspnet vnext v3" value="https://www.myget.org/F/aspnetvnext/api/v3/index.json" />
     <add key="myget.org aspnet vnext" value="https://www.myget.org/F/aspnetvnext/" />
--- a/dir.props
+++ b/dir.props
@@ -43,6 +43,7 @@
     <MicroBuildVersion>0.2.0</MicroBuildVersion>
     <GitVersioningVersion>1.6.35</GitVersioningVersion>
     <NuSpecReferenceGeneratorVersion>1.4.2</NuSpecReferenceGeneratorVersion>
+    <NuGetVersion>4.5.0-preview1-4518</NuGetVersion>
   </PropertyGroup>
 
    <!-- Common repo directories -->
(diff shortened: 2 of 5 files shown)


targets/BootStrapMSBuild.proj
```diff
@@ -46,6 +46,23 @@
...
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.dependencyresolver.core\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.frameworks\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.librarymodel\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.packaging\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.packaging.core\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.projectmodel\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.protocol\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.versioning\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuspec.referencegenerator\$(NuGetVersion)\lib\net45\**\*.*" />
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\Newtonsoft.Json\9.0.1\lib\net45\**\*.*" />
```
AndyGerlicher: This is really bad, but I hate to do anything fancy here in hopes that SDK migration will happen and this will need to change completely anyway.

targets/BootStrapMSBuild.proj
```diff
@@ -46,6 +46,23 @@
       <InstalledMicrosoftExtensions Include="$(MSBuildExtensionsPath)\Microsoft\**\*.*" />
 
       <InstalledNuGetFiles Include="$(MSBuildExtensionsPath)\Microsoft\NuGet\*" />
+      <!--<NuGetCommonExtensions Include="$(VsInstallRoot)\Common7\IDE\CommonExtensions\Microsoft\NuGet\**\*.*" />-->
```
jeffkl: Should this be removed?

NuGet.Config
```diff
@@ -11,7 +11,7 @@
     <add key="nuget.org v2" value="https://nuget.org/api/v2/" />
     <add key="myget.org roslyn nightly" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
     <add key="dotnet.myget.org buildtools v3" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
-    <!-- <add key="myget.org nugetbuild" value="https://www.myget.org/F/nugetbuild/api/v3/index.json" /> -->
+    <add key="myget.org nugetbuild" value="https://dotnet.myget.org/F/nuget-build/api/v3/index.json" />
```
cdmihai: Remove comment. Seems unneeded.

targets/BootStrapMSBuild.proj
```diff
@@ -46,6 +46,23 @@
...
+      <!-- TODO: Need NuGet.Build.Tasks.dll, would be not to not list out all of its dependencies! -->
```
cdmihai: Comment is unclear.

targets/BootStrapMSBuild.proj
```diff
@@ -46,6 +46,23 @@
...
+      <NuGetCommonExtensions Include="$(ProjectDir)packages\nuget.build.tasks\$(NuGetVersion)\lib\net45\**\*.*" />
```
cdmihai: What was wrong with the initial way of getting them from under VS?

Potential problem with grabbing them from their packages is if they have a dependency closure that isn't covered by the enumerated packages.

dir.props
```diff
@@ -56,8 +57,10 @@
     <GitVersioningDir>$(PackagesDir)\Nerdbank.GitVersioning\$(GitVersioningVersion)\build\</GitVersioningDir>
 
     <!-- Tell build tools to use the full framework dlls that contain build tasks (like the nuget assets task) -->
-    <BuildToolsTaskDir>$(ToolsDir)net45/</BuildToolsTaskDir>
-    <PackagingTaskDir>$(ToolsDir)net45/</PackagingTaskDir>
+    <BuildToolsTaskDir Condition="'$(MSBuildRuntimeType)'=='Full'">$(ToolsDir)net45\</BuildToolsTaskDir>
+    <BuildToolsTaskDir Condition="'$(MSBuildRuntimeType)'!='Full'">$(ToolsDir)</BuildToolsTaskDir>
```
cdmihai: Does this work in our .net core builds because the bootstrap msbuild coming from build tools is so old that MSBuildRuntimeType is empty and therefore different from "Full"?

AndyGerlicher: We don't download MSBuild when on `Full` (machine install on the image). We only download it when on `Core`. So == Core would not work, but != Full should.

targets/BootStrapMSBuild.proj
AndyGerlicher: We need a new task from NuGet so we need to latest version. Getting it from VS won't work until that dependency has flowed through. Plus this is better anyway not to copy stuff from VS.

NuGet.Config
AndyGerlicher: I un-commented this (and updated the feed URL) :)

PR #2600, opened 2017-10-06
Remove LINQ operations

davkean:
LazyOperation.Apply [is very allocation hungry](https://github.com/Microsoft/msbuild/issues/2599) and requires a small refactoring to prevent the gigabytes of data that its allocating while opening a partner's solution. To make reviews as easy as possible - I'm breaking off small portions on the road to the refactoring to try reduce regressions/review pain.

Below is unlinqifying RemoveOperation.SelectItems.

### DIFF
--- a/src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Build.Evaluation
             // todo Perf: do not match against the globs: https://github.com/Microsoft/msbuild/issues/2329
             protected override ICollection<I> SelectItems(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
             {
-                return SelectItemsMatchingItemSpec(listBuilder, _itemElement.RemoveLocation).ToImmutableHashSet();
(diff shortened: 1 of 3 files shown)


src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
```diff
@@ -20,7 +20,14 @@ public RemoveOperation(OperationBuilder builder, LazyItemEvaluator<P, I, M, D> l
...
             {
-                return SelectItemsMatchingItemSpec(listBuilder, _itemElement.RemoveLocation).ToImmutableHashSet();
+                var items = ImmutableHashSet.CreateBuilder<I>();
+                foreach (ItemData item in listBuilder)
+                {
+                    if (_itemSpec.MatchesItem(item.Item))
+                        items.Add(item.Item);
+                }
+
+                return items.ToImmutableHashSet();
```
davkean: @cdmihai Any ideas why this is an immutable hashset?

cdmihai: Because it's main use is items.Contains in SaveItems

---
In reply to: [143151572](https://github.com/Microsoft/msbuild/pull/2600#discussion_r143151572) [](ancestors = 143151572)

PR #2606, opened 2017-10-09
Remove allocs

davkean:
Remove 6 allocations from ExpandAndLoadImports, approx. 0.1% of allocations.

Removes a `string[]`, `ConcatIterator`, `List<T>` and Array enumerator from `pathsToSearch` creation.
Removes a `string` and boxed `List<string>.Enumerator` from the Join.


davkean:
tag @pilchie

PR #2607, opened 2017-10-09
Remove unneeded ToArray

davkean:
String.Join added an IEnumerable<string> overload in 4.0. This doesn't reduce the # allocations, but it does reduce the size of the allocations and prevent the copying.


davkean:
tag @pilchie

davkean:
tag @pilchie

PR #2608, opened 2017-10-09
Remove unneeded StringBuilder alloc

davkean:
Note, I left the behavior "as is" which is that it always logs inputs/outputs regardless of whether they exist, ie the previous null checks would never have returned false because they were always assigned a value.

While this uses a StringBuilder under the covers, it pulls it from mscorlib's pool.


davkean:
tag @pilchie

PR #2609, opened 2017-10-09
Avoid object[] alloc when logging

davkean:
Calls to FormatResourceString were allocating object[] - replace with calls to GetResourceString.


src/Shared/ResourceUtilities.cs
```diff
@@ -171,6 +172,13 @@ internal static string FormatResourceString(out string code, out string helpKeyw
             return ExtractMessageCode(true /* msbuildCodeOnly */, FormatString(GetResourceString(resourceName), args), out code);
         }
 
+        [Obsolete("Use GetResourceString instead.", true)]
```
davkean: Note, new code start here.

davkean:
tag @pilchie

src/Shared/ResourceUtilities.cs
Pilchie: Wait, how does this work?  Shouldn't this method at least have a single `string` parameter?

davkean: Whoops, I must have removed the string parameter. I used this method to find the original usages so looks like I must have removed it when I added the obsolete.

PR #2610, opened 2017-10-09
Init smarter and rewrite loops

davkean:
Type all fields as List<T> and then just foreach over them. This avoids the box.

Also initialize capability in a few places.


davkean:
tag @pilchie

PR #2626, opened 2017-10-11
Option to not evaluate elements with false conditions

### LINKED ISSUE #2502: Early evaluation of invalid DefaultCompileItems recursive itemgroup performance issue during incremental builds
Using the solution provided here: https://github.com/jeromelaban/MicrosoftIssues/tree/master/MsBuildEarlyEvalIssue, 

it is possible to determine that the first invocation of [LazyItemEvaluator.GetAllItems](https://github.com/Microsoft/msbuild/blob/dd134d0a917b870b6a6e805984909a8c341a8083/src/Build/Evaluation/LazyItemEvaluator.cs#L404) evaluates [this Compile item group](https://github.com/dotnet/sdk/blob/e83933917a2f3e3d10c6d18180ba6fab514117c8/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.DefaultItems.props#L26) while **DefaultLanguageSourceExtension** and **DefaultItemExcludes** are not defined.

This has the effect of recursively and unconditionally search all files in the project's folder, even in the bin and obj folders, and the **ret** variable in the GetAllItems method contains all these files.

When building Xamarin.Android projects (because of Google SDKs java imports), or other project that a **very** deep file structure in the obj or bin folder, this recursive enumeration takes quite a long time to complete, and its result is seemingly discarded afterwards.

I'm also assuming this can be a big problem perf in VS, where design-time builds reload each targetframework to evaluate its content, where in some profiling sessions, **ntfs.sys** shows up regularly in perfview.

cdmihai:
Up until now project based evaluations evaluated certain elements (properties / groups, items / groups, item definition groups) even when their condition was false. This PR adds an option to turn this off.

Tested with [this project](https://github.com/jeromelaban/MicrosoftIssues/blob/master/MsBuildEarlyEvalIssue/XTargetProject/XTargetProject.csproj) and dropped a node_modules inside. Respecting conditions takes ~600ms, not respecting them takes ~1200ms.

CPS will have to turn this on via ProjectLoadSettings.DoNotEvaluateElementsWithFalseCondition

Closes #2502

### DIFF
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -108,7 +108,8 @@ namespace Microsoft.Build.Utilities
                 return null;
             }
 
-            if (bool.TryParse(value, out var result))
+            bool result;
+            if (bool.TryParse(value, out result))
             {
                 return result;
             }
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Build.Utilities
         /// <summary>
         /// Force whether Project based evaluations should evaluate elements with false conditions.
         /// </summary>
-        public readonly bool? EvaluateElementsWithFalseConditionInProjectEvaluation = ParseNullableBoolFromEnvironmentVariable(nameof(EvaluateElementsWithFalseConditionInProjectEvaluation));
+        public readonly bool? EvaluateElementsWithFalseConditionInProjectEvaluation = ParseNullableBoolFromEnvironmentVariable("MSBUILDEVALUATEELEMENTSWITHFALSECONDITIONINPROJECTEVALUATION");
 
         /// <summary>
         /// Always use the accurate-but-slow CreateFile approach to timestamp extraction.
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -50,6 +50,11 @@ namespace Microsoft.Build.Utilities
     internal class EscapeHatches
     {
         /// <summary>
+        /// Force whether Project based evaluations should evaluate elements with false conditions.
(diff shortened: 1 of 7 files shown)


cdmihai:
@davkean @lifengl

src/Shared/Traits.cs
```diff
@@ -50,6 +50,11 @@ public Traits()
     internal class EscapeHatches
     {
         /// <summary>
+        /// Force whether Project based evaluations should evaluate elements with false conditions.
+        /// </summary>
+        public readonly bool? EvaluateElementsWithFalseConditionInProjectEvaluation = ParseNullableBoolFromEnvironmentVariable(nameof(EvaluateElementsWithFalseConditionInProjectEvaluation));
```
AndyGerlicher: Should this start with MSBUILD to match the other flags?

cdmihai: I guess it would be nicer to respect convention. I'll change it.
My intent was to reduce the amount of information, by keeping the member name the same as the environment variable.

src/Shared/Traits.cs
```diff
@@ -94,6 +99,26 @@ internal class EscapeHatches
         // it's unneeded (at least by the 16.0 timeframe).
         public readonly bool UseCaseSensitiveItemNames = Environment.GetEnvironmentVariable("MSBUILDUSECASESENSITIVEITEMNAMES") == "1";
 
+        private static bool? ParseNullableBoolFromEnvironmentVariable(string environmentVariable)
```
davkean: Why is this option a 'true' or 'false' but others are ints?

PR #2637, opened 2017-10-12
Add support for .NET SDK 4.7.1

AndyGerlicher:
Following the pattern from #1728


ref/net46/Microsoft.Build.Utilities.Core/Microsoft.Build.Utilities.Core.cs
```diff
@@ -317,6 +317,7 @@ public enum TargetDotNetFrameworkVersion
         Version461 = 8,
         Version462 = 10,
         Version47 = 11,
+        Version471 = 12,
```
cdmihai: Is it OK for VersionLatest to point to 462?

AndyGerlicher: Yeah, since it's only an update changing `VersionLatest` could be a breaking change. That's why I added `Latest` last time so that consumers could target the appropriate flag.

PR #2641, opened 2017-10-13
Restrict effect of DoNotEvaluateElementsWithFalseCondition

cdmihai:
This scopes down the effect of the flag introduced in #2629.


cdmihai:
@lifeng I prepared this to have it ready in case you find OK.

PR #2643, opened 2017-10-15
Exp/evaluation profiler

saranshmanu:


AndyGerlicher:
This isn't quite polished enough and would have to be completely turned off by default to be considered going into master to avoid regressing performance. This was pushed to a publish branch so that others could use it but it's not ready to go.

PR #2639, opened 2017-10-12
Target item includes use same code as Project item includes

cdmihai:
Fixes #2621 

This brings the exclude optimization (sending excludes together with includes, rather than file walking twice and doing set difference) to target based item includes as well, so it should make target item globbing a bit faster.

PR #2557, opened 2017-09-27
Normalize include and exclude specs before file walking

cdmihai:
Avoids expensive string normalization on each discovered directory.

Improves performance (and probably memory) by not calling path aware comparisons and string normalizations during the recursive file walk. Instead, the user specs are normalized before the walk begins.

Also closes #1874 

Results based on building NHibernate.Test.csproj referenced in https://github.com/Microsoft/msbuild/issues/2392#issuecomment-330029024

image

PR #2644, opened 2017-10-16
Fix exclude normalization mismatch in target item computation

cdmihai:
This is another implementation of #2639 which introduced a big perf regression.

Background:
Computing items under target elements is duplicated functionality with computing items under project elements. The clones used to be quite similar in the past, but have greatly diverged in the past year: initially, both clones would compute includes separately from excludes, and then obtain the final list of items by subtracting the excludes from the includes (excludes would hit the disk if they contained globs). Items under project underwent multiple modifications, like lazy item computation to avoid hitting the disk, and computing wildcard excludes together with wildcard includes to further avoid hitting the disk.

#2557 started normalizing items (for item elements under project elements) if they originate from globbing (e.g. `c:\a\\b//**` would produce `c:\a\b\c\f.cs` instead of the previous `c:\a\\b//c\f.cs`). This regressed target item computation because includes are computed separate from excludes there, and now include items originating from globbing would get normalized, whereas the excludes which did not contain wildcards would not get normalized, causing mismatches. #2639 fixed this by removing part of the code duplication, but hit a major [perf regression](https://github.com/Microsoft/msbuild/pull/2639#issuecomment-336602427).
This PR is a quick fixup until we do something better (i.e. fix the perf regression in the common code). It keeps the separate computation of excludes and includes, but also pipes the excludes together with includes, so that whenever includes get normalized, excludes get too and the exclusion works.

### DIFF
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -13,7 +13,20 @@ namespace Microsoft.Build.Utilities
     /// </summary>
     internal class Traits
     {
-        public static Traits Instance = new Traits();
+        private static Traits _instance = new Traits();
+        public static Traits Instance
+        {
+            get
+            {
+#if DEBUG
+                if (Environment.GetEnvironmentVariable("MSBUILDRELOADTRAITSONEACHACCESS") == "1")
+                {
+                    return new Traits();
+                }
+#endif
+                return _instance;
+            }
+        }
 
         public EscapeHatches EscapeHatches { get; }
 
--- a/src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
@@ -13,6 +13,7 @@ using System.Linq;
 using System.Xml;
 using Microsoft.Build.Construction;
 using Microsoft.Build.Engine.UnitTests;
+using Microsoft.Build.Engine.UnitTests.Globbing;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Shared;
 using InvalidProjectFileException = Microsoft.Build.Exceptions.InvalidProjectFileException;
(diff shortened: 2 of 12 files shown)


cdmihai:
Perf test results. Evaluation is just noise, because nothing changed there. Build is a bit slower. This is expected, because the target item computation does more now (it pushes excludes together with includes, in addition of computing the excludes separately).

### Evaluation: Time (ms)
:--- | :--- | :--- | :---
DotnetConsoleProject | :red_circle: | yes | 90.6055 -> 90.9269 (0.355%)
DotnetWebProject | :red_circle: | yes | 133.016 -> 133.6116 (0.448%)
DotnetMvcProject | :white_check_mark: | yes | 157.7966 -> 156.5746 (-0.774%)
Picasso | :red_circle: | yes | 738.8228 -> 782.5352 (5.916%)
SmallP2POldCsproj | :red_circle: | yes | 129.7989 -> 131.248 (1.116%)
SmallP2PNewCsproj | :red_circle: | yes | 238.2213 -> 240.3197 (0.881%)
Generated_100_100_v150 | ::ok_hand: | no | 1604.0993 -> 1579.3942 (-1.54%)
LargeP2POldCsproj | :red_circle: | yes | 2161.7559 -> 2192.0117 (1.4%)
roslyn | ::ok_hand: | no | 9580.2518 -> 9532.1202 (-0.502%)

:--- | :--- | :--- | :---
DotnetConsoleProject | ::ok_hand: | no | 4893744 -> 4893259 (-0.01%)
DotnetWebProject | ::ok_hand: | no | 6566959 -> 6566485 (-0.007%)
DotnetMvcProject | ::ok_hand: | no | 7412307 -> 7412210 (-0.001%)
Picasso | ::ok_hand: | no | 34311107 -> 34311077 (0%)
SmallP2POldCsproj | :ok_hand: | no | 5898319 -> 5898493 (0.003%)
SmallP2PNewCsproj | :ok_hand: | no | 10356413 -> 10356421 (0%)
Generated_100_100_v150 | ::ok_hand: | no | 143467602 -> 143467447 (0%)
LargeP2POldCsproj | :white_check_mark: | yes | 83945155 -> 83942822 (-0.003%)
roslyn | :ok_hand: | no | 401583183 -> 401895702 (0.078%)

### Build: Time (ms)
:--- | :--- | :--- | :---
DotnetConsoleProject | :red_circle: | yes | 22.8794 -> 23.0854 (0.9%)
DotnetWebProject | :ok_hand: | no | 28.1029 -> 28.2645 (0.575%)
DotnetMvcProject | :white_check_mark: | yes | 31.4909 -> 31.0981 (-1.247%)
Picasso | :red_circle: | yes | 3106.249 -> 3133.7586 (0.886%)
SmallP2POldCsproj | :red_circle: | yes | 148.9675 -> 150.4191 (0.974%)
SmallP2PNewCsproj | :red_circle: | yes | 156.5939 -> 157.7253 (0.723%)

### Build: Memory (bytes)
:--- | :--- | :--- | :---
DotnetConsoleProject | :red_circle: | yes | 1156694 -> 1162606 (0.511%)
DotnetWebProject | :red_circle: | yes | 1282700 -> 1286292 (0.28%)
DotnetMvcProject | :red_circle: | yes | 1453606 -> 1457712 (0.282%)
Picasso | :red_circle: | yes | 121490168 -> 121917941 (0.352%)
SmallP2POldCsproj | :red_circle: | yes | 6104333 -> 6171347 (1.098%)
SmallP2PNewCsproj | :ok_hand: | no | 6401056 -> 6403831 (0.043%)

src/Shared/Traits.cs
```diff
@@ -13,7 +13,20 @@ namespace Microsoft.Build.Utilities
...
     internal class Traits
     {
-        public static Traits Instance = new Traits();
+        private static Traits _instance = new Traits();
+        public static Traits Instance
+        {
+            get
+            {
+#if DEBUG
+                if (Environment.GetEnvironmentVariable("MSBUILDRELOADTRAITSONEACHACCESS") == "1")
```
AndyGerlicher: I get why you did this, but I'm a little worried. Getting this environment variable will start eating up a lot of CPU on debug builds. We should try to think of another way to do this only for tests.

Maybe check executing assembly? BuildEnvironmentHelper already has a RunningTests property. Could do something similar to that.

PR #2650, opened 2017-10-18
Don't add OriginalItemSpec when it already exists

### LINKED ISSUE #2649: OriginalItemSpec metadata incorrect on project references
Previous Behavior
 image

 
New Behavior
image

 
The reference is:
```XML
  <ItemGroup>
    <ProjectReference Include="..\ClassLibrary1\ClassLibrary1.csproj" />
  </ItemGroup>
 ```
This change breaks the check in [DependenciesRuleHandlerBase.cs#L129](https://github.com/dotnet/project-system/blob/7267defa7fb7d5790378e21f19ca740a21a36e1c/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Tree/Dependencies/Subscriptions/DependenciesRuleHandlerBase.cs#L129) which compares the Original Item Specs from ProjectReference and ResolvedProjectReference to determine whether a change should be processed. It returns false and hence the design time change is not processed.

### PR DESCRIPTION
**Pull request by AndyGerlicher**
Not positive this is the best way to fix this issue, but it seems to solve this class of issue. Not 100% sure that this wouldn't subtly change behavior though. But it seems like if you have `OriginalItemSpec` you probably don't want it reset every time you alter the item.

There is special logic to add OriginalItemSpec as metadata when copying
items. When that metadata already exists, don't overwrite it.

Closes #2649

### DIFF
--- a/src/Build/Instance/ProjectItemInstance.cs
+++ b/src/Build/Instance/ProjectItemInstance.cs
@@ -1368,6 +1368,11 @@ namespace Microsoft.Build.Execution
                         {
                             // Utilities.TaskItem's don't know about item definition metadata. So merge that into the values.
                             destinationItem.SetMetadata(metadatum.Name, GetMetadataEscaped(metadatum.Name));
+                            if (metadatum.Name == "OriginalItemSpec")
+                            {
+                                // We copied OriginalItemSpec, don't re-apply it later.
+                                addOriginalItemSpec = false;
+                            }
                         }
                     }
                 }

### CODE REVIEW

**Comment by AndyGerlicher** [comment]
Trying to write an overview of the issue.
https://github.com/Microsoft/msbuild/blob/57ae27cf0812bf262c6f8be456ed527a1542dba7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1589-L1594

When that task runs, it takes `_ProjectReferenceTargetFrameworkPossibilities` and returns `AnnotatedProjects` which contains information we need. The `AnnotatedProjects` array is created by constructing new `TaskItem`s from the original input `ITaskItem` array:
https://github.com/NuGet/NuGet.Client/blob/dda020578d856bedc3646f39fd826b483a58fa63/src/NuGet.Core/NuGet.Build.Tasks/GetReferenceNearestTargetFrameworkTask.cs#L71

The code in this PR sets the `OriginalItemSpec` metadata to the `_includeEscaped` value of the source `TaskItem`. There's explicit code in there to add that. Which makes some amount of sense. The odd thing here is that the source `TaskItem` object already had an `OriginalItemSpec` so the value gets overwritten. I would _think_ the `OriginalItemSpec` should be preserved. And it gets a little more complicated in the next bit of code if that value isn't preserved.

After that task runs, we need to decorate `_MSBuildProjectReferenceExistent` with metadata based on what we learned from the task. So we update `AnnotatedProjects` and then remove/replace what is in `_MSBuildProjectReferenceExistent`.

https://github.com/Microsoft/msbuild/blob/57ae27cf0812bf262c6f8be456ed527a1542dba7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1604-L1627

Because `OriginalItemSpec` on `AnnotatedProjects` is lost and is essentially set to the Include value, that original value is gone.

**Comment by cdmihai** [comment]
Another option would be to reset the absolute path back to the original value right after the MSBuild task

```
<_ProjectReferenceTargetFrameworkPossibilities Include="@(_ProjectReferenceTargetFrameworkPossibilities->'%(OriginalItemSpec)')">
```

File: src/Build/Instance/ProjectItemInstance.cs [hunk 1]
```diff
@@ -1368,6 +1368,11 @@ public void CopyMetadataTo(ITaskItem destinationItem, bool addOriginalItemSpec)
                         {
                             // Utilities.TaskItem's don't know about item definition metadata. So merge that into the values.
                             destinationItem.SetMetadata(metadatum.Name, GetMetadataEscaped(metadatum.Name));
+                            if (metadatum.Name == "OriginalItemSpec")
```
cdmihai: If the intent of `OriginalItemSpec` is to contain the value of the latest item metadata was copied from, then this would break that behaviour, as it would not start propagating the previous value.

cdmihai: :tired_face:

The logic is duplicated in:
https://github.com/Microsoft/msbuild/blob/master/src/Utilities/TaskItem.cs#L367
https://github.com/Microsoft/msbuild/blob/master/src/Shared/TaskParameter.cs#L682

Searching for `SetMetadata("OriginalItemSpec"` finds other clones all over the place that patch this metadata.

PR #2658, opened 2017-10-21
Avoid cloning AssemblyName

davkean:
This is causing 11% of all allocations building Roslyn.sln.

image

### DIFF
--- a/src/Shared/AssemblyUtilities.cs
+++ b/src/Shared/AssemblyUtilities.cs
@@ -70,10 +70,9 @@ namespace Microsoft.Build.Shared
             name.SetPublicKeyToken(assemblyNameToClone.GetPublicKeyToken());
             name.Version = assemblyNameToClone.Version;
             name.Flags = assemblyNameToClone.Flags;
-            name.ContentType = assemblyNameToClone.ContentType;
             name.ProcessorArchitecture = assemblyNameToClone.ProcessorArchitecture;
 
-#if FEATURE_ASSEMBLYNAME_CULTUREINFO
+#if !RUNTIME_TYPE_NETCORE
             name.CultureInfo = assemblyNameToClone.CultureInfo;
             name.HashAlgorithm = assemblyNameToClone.HashAlgorithm;
             name.VersionCompatibility = assemblyNameToClone.VersionCompatibility;
--- a/src/Shared/AssemblyUtilities.cs
+++ b/src/Shared/AssemblyUtilities.cs
@@ -17,7 +17,6 @@ namespace Microsoft.Build.Shared
         private static bool s_initialized;
 
         // Cached method info
-        private static MethodInfo s_assemblyNameCloneMethod;
         private static PropertyInfo s_assemblylocationProperty;
         private static MethodInfo s_cultureInfoGetCultureMethod;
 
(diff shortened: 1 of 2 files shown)


src/Shared/AssemblyUtilities.cs
```diff
@@ -65,12 +65,19 @@ public static AssemblyName CloneIfPossible(this AssemblyName assemblyNameToClone
             name.SetPublicKey(assemblyNameToClone.GetPublicKey());
             name.SetPublicKeyToken(assemblyNameToClone.GetPublicKeyToken());
             name.Version = assemblyNameToClone.Version;
+            name.Flags = assemblyNameToClone.Flags;
+            name.CultureName = assemblyNameToClone.CultureName;
```
radical: Setting `CultureName` will throw NIE on mono. You can set `AssemblyName.CultureInfo` instead. See https://github.com/mono/msbuild/commit/d4c871554368485fde1dbaed09922319dd7f4ea7 and https://github.com/mono/msbuild/commit/42ba59a920754c24968dafe82622a59d27385c1d .

akoeplinger: @radical could you please file a bugzilla so we get this implemented?

cdmihai: Assigning it when `!MONO`. CultureInfo gets set via `FEATURE_ASSEMBLYNAME_CULTUREINFO`

src/Shared/AssemblyUtilities.cs
```diff
@@ -55,22 +54,41 @@ public static Type GetTypeInfo(this Type t)
...
+            AssemblyName name = new AssemblyName();
+            name.Name = assemblyNameToClone.Name;
+            name.SetPublicKey(assemblyNameToClone.GetPublicKey());
+            name.SetPublicKeyToken(assemblyNameToClone.GetPublicKeyToken());
+            name.Version = assemblyNameToClone.Version;
+            name.Flags = assemblyNameToClone.Flags;
+            name.ContentType = assemblyNameToClone.ContentType;
+            name.ProcessorArchitecture = assemblyNameToClone.ProcessorArchitecture;
+
+#if FEATURE_ASSEMBLYNAME_CULTUREINFO
```
radical: Should this section be restricted to setting `name.CultureInfo` only?

src/Shared/AssemblyUtilities.cs
```diff
@@ -55,22 +54,41 @@ public static Type GetTypeInfo(this Type t)
...
+            //
+            // While the array themselves are mutable - throughout MSBuild they are only ever
+            // read from.
+            AssemblyName name = new AssemblyName();
+            name.Name = assemblyNameToClone.Name;
+            name.SetPublicKey(assemblyNameToClone.GetPublicKey());
+            name.SetPublicKeyToken(assemblyNameToClone.GetPublicKeyToken());
+            name.Version = assemblyNameToClone.Version;
+            name.Flags = assemblyNameToClone.Flags;
+            name.ContentType = assemblyNameToClone.ContentType;
```
davkean: ContentType is a flag, this isn't needed and was deliberately omitted.

src/Shared/AssemblyUtilities.cs
cdmihai: I was thinking about that. All the members under the flag are not supported on .net core. Since we already have a mess of API flags, I didn't want to increase the proliferation. I'll switch it with `RUNTIME_TYPE_NETCORE` instead.

PR #2662, opened 2017-10-24
Stop reading the ToolsVersion from project XML during build

### LINKED ISSUE #2604: Resolving ToolsVersion costs 1% of a build and 2% of all allocations
I thought we don't respect/use ToolsVersion anymore?

Memory:
image

CPU:
image

jeffkl:
We don't use ToolsVersion and getting the attibute from XML is costly.  It has to allocate a streeam, a byte[] to read the BOM, and then read elements and attributes.

Fixes #2604

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildManager.cs
+++ b/src/Build/BackEnd/BuildManager/BuildManager.cs
@@ -833,7 +833,7 @@ namespace Microsoft.Build.Execution
                 }
 
                 // Create/Retrieve a configuration for each request
-                BuildRequestConfiguration buildRequestConfiguration = new BuildRequestConfiguration(submission.BuildRequestData, _buildParameters.DefaultToolsVersion, _buildParameters.GetToolset);
+                BuildRequestConfiguration buildRequestConfiguration = new BuildRequestConfiguration(submission.BuildRequestData, _buildParameters.DefaultToolsVersion);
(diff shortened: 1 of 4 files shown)


davkean:
The change itself looks good, I cannot comment on the compat impact.

cdmihai:
Did a bit of code hunting to see the impact. `data.ExplicitToolsVersionSpecified` from `ResolveToolsversion` does not always come from the project file. For example, on command line builds, ExplicitToolsVersionSpecified is only set if the /toolsversion flag is used. So with this change, the toolsversion from the actual file, if present, does not flow into MSBuild anymore.

Regarding the concept of toolsversion itself, it seems to flow everywhere in MSBuild :(. But most of the logic only has to do with toolsversion management and not actual functionality branching. The only worrying part that I found is that toolsversion is part of the unique compound key that identifies projects in msbuild's caches (alongside the project path and global properties), so there's the risk of different code paths under weird circumstances where there's multiple sources of toolsversion, one from the file and another one from different sources (command line, project collection, etc).

Maybe it would be better to hide the logic under an escape hatch Trait rather than delete it? That way potentially broken projects can get unstuck. If however nobody breaks, maybe it would be better to delete the concept of toolsversion altogether from MSBuild, since there's a lot of logic around it to just to manage and propagate it anywhere, but not real use.

PR #2595, opened 2017-10-05
Double Evaluation Fix

AndyGerlicher:
Replaces #2472 and #2458

Still WIP. I need to get the NuGet task pulled down for bootstrap build.

PR #2663, opened 2017-10-24
Explicitly import NuGet.targets

AndyGerlicher:
NuGet.targets was added as a hard dependency in #2595, this change will
explicitly import it in Microsoft.Common.CurrentVersion.targets.

NuGet change: https://github.com/NuGet/NuGet.BuildTasks/pull/44


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -5869,6 +5869,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <!-- This import is temporary and will be removed once it is moved into the silverlight targets -->
   <Import Project="$(MSBuildToolsPath)\Microsoft.Data.Entity.targets" Condition="'$(TargetFrameworkIdentifier)' == 'Silverlight' and Exists('$(MSBuildToolsPath)\Microsoft.Data.Entity.targets')"/>
 
+  <!-- Import NuGet.targets (required for GetReferenceNearestTargetFrameworkTask and used for Restore functionality) -->
+  <PropertyGroup>
+    <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'=='' and '$([MSBuild]::IsRunningFromVisualStudio())'=='true'">$(MSBuildToolsPath32)\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets</NuGetRestoreTargets>
+    <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'==''">$(MSBuildToolsPath)\NuGet.targets</NuGetRestoreTargets>
```
radical: Is the location being changed from `$(MSBuildExtensionsPath)\NuGet.targets` ? (https://github.com/Microsoft/msbuild/blob/master//targets/bootstrapDependencies/Core/15.0/Microsoft.Common.targets/ImportAfter/Microsoft.NuGet.ImportAfter.targets#L15)

AndyGerlicher: Good catch, thanks! This shouldn't be needed at all anymore. This file was just because that was the way we imported `NuGet.targets` and on .NET Core we didn't have the file locally to copy it. Now that the import is explicit this file isn't needed.

AndyGerlicher: And to the location, it should be either in the VS install location (when `IsRunningFromVisualStudio` is true) or next to MSBuild in the .NET Core case (`$(MSBuildToolsPath)`). It should be the case that `MSBuildToolsPath` and `MSBuildExtensionsPath` are the same, but not necessarily. This work ok with mono?

cibuild.cmd
```diff
@@ -158,7 +158,10 @@ if /i "%TARGET%"=="CoreCLR" (
 :: warning MSB3026: Could not copy "XXX" to "XXX". Beginning retry 1 in 1000ms.
 :: warning MSB3073: Exec task failure (when set to be a warning) -- needed to keep from failing on dev desktops that don't have C++ tools
 :: warning AL1053: The version '1.2.3.4-foo' specified for the 'product version' is not in the normal 'major.minor.build.revision' format
-SET _NOWARN=MSB3277;MSB3026;MSB3073;AL1053
+:: warning MSB4011: Imported target already imported. This will happen when Microsoft.NuGet.ImportAfter.targets installed on the local
+::                  machine still has the import to NuGet.targets which is now in Microsoft.Common.CurrentVersion.targets. This can be
+::                  removed when we want to require VS15.5+ to build locally.
+SET _NOWARN=MSB3277;MSB3026;MSB3073;AL1053;MSB4011
```
jeffkl: You will need to add the code to `cibuild.sh` as well

jeffkl: Or not, on cross plat I guess we won't get a double import warning?

src/Tasks/Microsoft.Common.CurrentVersion.targets
radical: We deploy to `MSBuildExtensionsPath`. And that differs from `MSBuildToolsPath`. If the two are the same for netcore, then we can change it to use `MSBuildExtensionsPath`, I guess?

AndyGerlicher: Is it possible to change? Prefer to not use `MSBuildExtensionsPath`. Since they're fully required now, it's not really an extension. More like Roslyn is:

https://github.com/Microsoft/msbuild/blob/master/src/MSBuild/app.config#L101

The checked in file was just to get tests working not for any deployment strategy.

radical: re:change, sure, I'll do that. I am wondering if I should be removing it from `MSBuildExtensionsPath` or have it in both locations (it doesn't import anything, currently!).

PR #2671, opened 2017-10-26
Fix issue with list of warnings or errors to treat differently can be added to the list multiple times

### LINKED ISSUE #2667: Build failure with use of MSBuildWarningsAsMessages
tm1.proj:
```
<Project>
        <Target Name="Main">
                <MSBuild Projects="tm2.proj" Targets="Main" />
        </Target>
</Project>
```

tm2.proj:
```
<Project>
        <PropertyGroup>
                <MSBuildWarningsAsMessages>MSB9004</MSBuildWarningsAsMessages>
        </PropertyGroup>

        <Target Name="Main">
                <MSBuild Projects="$(MSBuildThisFileFullPath)" Targets="Foo" />
        </Target>

        <Target Name="Foo">
                <Message Text="Hello from Foo" />
        </Target>
</Project>
```

Build with msbuild:

At second invocation of `MSBuild` task to call `Foo` target (from the *diag* log):

```
<snip>
Building with tools version "15.0".
Done Building Project "/Users/ankit/test/msbuild/tm2.proj" (Foo target(s)) -- FAILED.
Done executing task "MSBuild" -- FAILED. (TaskId:3)
Done building target "Main" in project "tm2.proj" -- FAILED.: (TargetId:3)
Done Building Project "/Users/ankit/test/msbuild/tm2.proj" (Main target(s)) -- FAILED.
Done executing task "MSBuild" -- FAILED. (TaskId:2)
Done building target "Main" in project "tm1.proj" -- FAILED.: (TargetId:2)
Done Building Project "/Users/ankit/test/msbuild/tm1.proj" (default targets) -- FAILED.
```

Adding a few CWLs to msbuild and we get:

```
System.ArgumentException: The key already existed in the dictionary.
  at System.Collections.Concurrent.ConcurrentDictionary`2[TKey,TValue].System.Collections.Generic.IDictionary<TKey,TValue>.Add (TKey key, TValue value) [0x00015] in /Users/builder/data/lanes/4992/mono-mac-sdk/external/bockbuild/builds/mono-x64/external/
corefx/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs:1303
  at Microsoft.Build.BackEnd.Logging.LoggingService.AddWarningsAsMessages (System.Int32 projectInstanceId, System.Collections.Generic.ISet`1[T] codes) [0x00188] in /Users/ankit/dev/msbuild/src/Build/BackEnd/Components/Logging/LoggingService.cs:539
  at Microsoft.Build.BackEnd.RequestBuilder.ConfigureWarningsAsErrorsAndMessages () [0x000ed] in /Users/ankit/dev/msbuild/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs:1311
  at Microsoft.Build.BackEnd.RequestBuilder+<BuildProject>d__57.MoveNext () [0x0010d] in /Users/ankit/dev/msbuild/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs:1094
```

(The line number is a little off because of extra debug statements in my copy).
https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/Components/Logging/LoggingService.cs#L517

`_filterEventSource.WarningsAsMessagesByProject.Add(projectInstanceId, new HashSet<string>(codes, StringComparer.OrdinalIgnoreCase));`

1. So, `MSBuild` task from tm1 invokes tm2's `Main` target and this gets a unique `projectInstanceId`, which is used to populate the above `_filterEventSource.WarningsAsMessagesByProject`. 
2. And then tm2, uses `MSBuild` task to invoke `Foo` target on the same project file, and even this has the same `projectInstanceId` and so the above line of code fails as the entry already exists in `_filterEventSource.WarningsAsMessagesByProject`.

The Id is generated in relation to the contextId (AFAICS), and you would get a different Id if the build different like if any extra properties were passed to the `MSBuild` task.

What is the fix here? Should this be using `projectContextId` instead? or should new `projectInstanceId` be generated every time? I might be talking nonsense about Ids too, I am still trying to read the code and make sense of it!

jeffkl:
When building two or more projects in the same build episode, the list of warnings to treat as errors or warnings to treat as messages is parsed multiple times.

This fixes the issue by checking if the list is already populated for a given project instance.

Fixes #2667

### DIFF
--- a/src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
+++ b/src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
-using System.Diagnostics;
 using System.Linq;
 using Microsoft.Build.Framework;
 using Microsoft.Build.UnitTests;
@@ -82,6 +80,45 @@ namespace Microsoft.Build.Engine.UnitTests
             MockLogger logger = ObjectModelHelpers.BuildProjectExpectSuccess(GetTestProject(warningsAsMessages: ExpectedEventCode));
 
             VerifyBuildMessageEvent(logger);
+        }
+
+        /// <summary>
+        /// https://github.com/Microsoft/msbuild/issues/2667
+        /// </summary>
+        [Fact]
+        public void TreatWarningsAsMessagesWhenBuildingSameProjectMultipleTimes()
+        {
+            try
+            {
+                ObjectModelHelpers.CreateFileInTempProjectDirectory("project1.proj", $@"
+                <Project ToolsVersion=""msbuilddefaulttoolsversion"" xmlns=""msbuildnamespace"">
+                    <Target Name=""Build"">
+                        <MSBuild Projects=""project2.proj"" Targets=""Build"" />
+                    </Target>
+                </Project>");
+
+                ObjectModelHelpers.CreateFileInTempProjectDirectory("project2.proj", $@"
+                <Project ToolsVersion=""msbuilddefaulttoolsversion"" xmlns=""msbuildnamespace"">
+                    <PropertyGroup>
+                        <MSBuildWarningsAsMessages>{ExpectedEventCode}</MSBuildWarningsAsMessages>
+                    </PropertyGroup>
+                    <Target Name=""Build"">
+                        <MSBuild Projects=""$(MSBuildThisFileFullPath)"" Targets=""AnotherTarget"" />
+                    </Target>
+                    <Target Name=""AnotherTarget"">
+                        <Message Text=""WarningsAsMessages: '$(MSBuildWarningsAsMessages)'"" />
+                        <Warning Text=""{ExpectedEventMessage}"" Code=""{ExpectedEventCode}"" />
+                    </Target>
+                </Project>");
+
+                MockLogger logger = ObjectModelHelpers.BuildTempProjectFileExpectSuccess("project1.proj");
+
+                VerifyBuildMessageEvent(logger);
+            }
+            finally
+            {
+                ObjectModelHelpers.DeleteTempProjectDirectory();
+            }
         }
 
         [Fact]
--- a/src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
+++ b/src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
@@ -22,6 +22,40 @@ namespace Microsoft.Build.Engine.UnitTests
             ObjectModelHelpers.BuildProjectExpectSuccess(GetTestProject(treatAllWarningsAsErrors: false));
         }
 
+        /// <summary>
+        /// https://github.com/Microsoft/msbuild/issues/2667
+        /// </summary>
+        [Fact]
+        public void TreatWarningsAsErrorsWhenBuildingSameProjectMultipleTimes()
+        {
+            using (TestEnvironment testEnvironment = TestEnvironment.Create())
+            {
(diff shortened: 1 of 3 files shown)


src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
```diff
@@ -84,6 +82,45 @@ public void TreatWarningsAsMessagesWhenSpecified()
...
+        /// <summary>
+        /// https://github.com/Microsoft/msbuild/issues/2667
+        /// </summary>
+        [Fact]
+        public void TreatWarningsAsMessagesWhenBuildingSameProjectMultipleTimes()
+        {
+            try
+            {
+                ObjectModelHelpers.CreateFileInTempProjectDirectory("project1.proj", $@"
+                <Project ToolsVersion=""msbuilddefaulttoolsversion"" xmlns=""msbuildnamespace"">
```
radical: Do we need the `ToolsVersion` and `xmlns` attributes here?

src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
```diff
@@ -84,6 +82,45 @@ public void TreatWarningsAsMessagesWhenSpecified()
             VerifyBuildMessageEvent(logger);
         }
 
+        /// <summary>
+        /// https://github.com/Microsoft/msbuild/issues/2667
+        /// </summary>
+        [Fact]
+        public void TreatWarningsAsMessagesWhenBuildingSameProjectMultipleTimes()
```
radical: Maybe another test for `WarningsAsErrors`, to be thorough?

jeffkl: Done

src/Shared/UnitTests/TestEnvironment.cs
```diff
@@ -407,10 +406,40 @@ public class TransientTestProjectWithFiles : TransientTestState
             CreatedFiles = Helpers.CreateFilesInDirectory(TestRoot, files);
         }
 
+        internal MockLogger BuildProjectExpectFailure(IDictionary<string, string> globalProperties = null, string toolsVersion = null)
```
radical: Nitpicking a bit, I guess. Why `internal` and not public? This is a tests only class.

radical: Also, are these two wrappers useful? And it's not obvious from the name that this would return a logger.

AndyGerlicher:
As-is this would target 15.6. That ok? We can go through escrow if needed for 15.5.

jeffkl:
@radical you mention this is failing NuGet restore in some cases.  Do you know how big of an impact we're going to see?  I'm assuming its low since you're the first person to report it...

src/Shared/UnitTests/TestEnvironment.cs
jeffkl: `MockLogger` is internal so the methods had to be internal as well.  These methods reflect other functionality we have in `ObjectModelHelpers`

jeffkl:

radical:
@jeffkl I am able to reproduce this with a simple solution which has project A depending on a library project B. And B has `<MSBuildWarningsAsMessages>MSB9001</MSBuildWarningsAsMessages>`. `msbuild /t:Restore` fails for such a solution! Note that I didn't even have to add any nuget references to this! I think hitting this might be not be rare for anyone using `MSBuildWarningsAsMessages`. Though usage of `MSBuildWarningsAsMessages` might not be that common itself, I don't know!

jeffkl:
Thanks @radical.  We're going to attempt to get this in 15.5 pending approval

PR #2672, opened 2017-10-26
Import NuGet.targets in CrossTargeting build

AndyGerlicher:
This case was missed and is failing CI for https://github.com/dotnet/cli/pull/7914

When doing a CrossTargeting build, we need to import NuGet for Restore. This was done by importing the file that used to be dropped with the CLI which would import the real NuGet.targets.


AndyGerlicher:
@nguerrera look ok to you? Fixes the PR I sent to CLI.

PR #2674, opened 2017-10-26
Don't enumerate GAC directories if the GAC path doesn't exist

### LINKED ISSUE #2673: GetReferenceAssemblyPaths task fails on Mono when "gac" path does not exist
In `NativeMethods+AssemblyCacheEnum`, the `s_gacPath` field is initialized like so ([src](https://github.com/Microsoft/msbuild/blame/7f40c7cef6714a78769d3c1158657aca23ebdeef/src/Tasks/NativeMethods.cs#L1455)):

```C#
private static string s_gacPath = Path.Combine(NativeMethodsShared.FrameworkBasePath, "gac");
```

This uses `FrameworkBasePath`, which is initialized with `FrameworkCurrentPath`, which is initialized with:

```C#
AssemblyUtilities.GetAssemblyLocation(typeof(string).GetTypeInfo().Assembly)
```

So, the GAC path is assumed to be the location of the assembly containing `System.String` with "gac" appended to it.

Later, when running on non-Windows, `s_gacPath`  is used to enumerate directories in order to initialize the `AssemblyCacheEnum._gacDirectories` field like so ([src](https://github.com/Microsoft/msbuild/blame/7f40c7cef6714a78769d3c1158657aca23ebdeef/src/Tasks/NativeMethods.cs#L1535-L1543)):

```C#
if (!string.IsNullOrWhiteSpace(assemblyName))
{
    _assemblyNameVersion = new AssemblyName(assemblyName);
    _gacDirectories = Directory.EnumerateDirectories(s_gacPath, _assemblyNameVersion.Name);
}
else
{
    _gacDirectories = Directory.EnumerateDirectories(s_gacPath);
}
```

For a normal Mono layout, this works fine. However, on a non-standard Mono layout (such as the one used in OmniSharp), it fails because `"<location-of-assembly-containing-system-string>/gac"` might not exist. The exception thrown is a `System.IO.DirectoryNotFoundException` from the `GetReferenceAssemblyPaths` task.

DustinCampbell:
Fixes #2673

On non-Windows, MSBuild assumes that the GAC path is the location of the assembly containing `System.String` plus `"/gac"`. This works fine on a standard Mono layout. However, on a non-standard Mono layout (such as the one used by OmniSharp), this throws a `System.IO.DirectoryNotFoundException` when trying enumerate the GAC directories.

This PR fixes the issue by checking with the GAC path exists before enumerating directories within it. If it doesn't, it is assumed that there are no GAC directories to search.

### DIFF
--- a/src/Tasks/NativeMethods.cs
+++ b/src/Tasks/NativeMethods.cs
@@ -1532,14 +1532,21 @@ typedef enum _tagAssemblyComparisonResult
                 }
                 else
                 {
-                    if (!string.IsNullOrWhiteSpace(assemblyName))
+                    if (Directory.Exists(s_gacPath))
                     {
-                        _assemblyNameVersion = new AssemblyName(assemblyName);
-                        _gacDirectories = Directory.EnumerateDirectories(s_gacPath, _assemblyNameVersion.Name);
+                        if (!string.IsNullOrWhiteSpace(assemblyName))
+                        {
+                            _assemblyNameVersion = new AssemblyName(assemblyName);
+                            _gacDirectories = Directory.EnumerateDirectories(s_gacPath, _assemblyNameVersion.Name);
+                        }
+                        else
+                        {
+                            _gacDirectories = Directory.EnumerateDirectories(s_gacPath);
+                        }
                     }
                     else
                     {
-                        _gacDirectories = Directory.EnumerateDirectories(s_gacPath);
+                        _gacDirectories = Array.Empty<string>();
                     }
                 }
             }


AndyGerlicher:
15.6 ok? If 15.5 I can change the target branch but need to take to escrow so a statement on impact if we don't do this would be great.

DustinCampbell:
15.6 is fine. I've been meaning to get this in for awhile. :smile:

src/Tasks/NativeMethods.cs
```diff
@@ -1532,14 +1532,21 @@ private void InitializeEnum(String assemblyName)
...
+                        }
+                        else
+                        {
+                            _gacDirectories = Directory.EnumerateDirectories(s_gacPath);
+                        }
                     }
                     else
                     {
-                        _gacDirectories = Directory.EnumerateDirectories(s_gacPath);
+                        _gacDirectories = Array.Empty<string>();
```
jeffkl: Is it better to set this to `Enumerable.Empty<string>()` or `Array.Empty<string>()`?

DustinCampbell: They're the same: https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Enumerable.cs#L13

AndyGerlicher: Method | Mean | Error | StdDev | Allocated
-- | -- | -- | -- | --
EmptyArray | 21.02 ns | 0.0624 ns | 0.0584 ns | 0 B
EmptyEnumerable | 19.25 ns | 0.1355 ns | 0.1268 ns | 0 B

Odd they're moderately different. Maybe it avoids a cast? This for sure doesn't matter, but I ran that real quick before I saw your link.

DustinCampbell: LOL

jeffkl: The more you know

PR #2676, opened 2017-10-26
Reduce AssemblyName allocations

cdmihai:
In mvc rebuild (133mb total allocations), reduces AssemblyName allocations 18.3% ->  0.5% and AssemblyNameExtensions 4.5% -> 0.3%. In a Roslyn rebuild (3.2gb total allocations), it reduces AssemblyName 2.3% -> 0.2% and AsssemblyNameExtension 0.6% -> 0.1%.

Commit [#1](https://github.com/Microsoft/msbuild/commit/8d5b139b9fabe58d288a65d4c34fe887fe812ae5) is safe to do, because the AssemblyName was double cloned ...
Commit [#2](https://github.com/Microsoft/msbuild/pull/2676/commits/b292563136731a53858defbac2fda9b1d056d7af) replaces calls to [AssemblyNameExtension.EqualsImpl](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Shared/AssemblyNameExtension.cs#L670-L733) and [AssemblyNameExtension.CompareBaseNameToImpl](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Shared/AssemblyNameExtension.cs#L516-L546) with a simple name, culture, and key comparison. This _looks_ safe to do because:
- the versions don't matter because both assemblies [get the same version](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Tasks/AssemblyDependency/ReferenceTable.cs#L2379)
- [AssemblyNameExtension.asString](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Shared/AssemblyNameExtension.cs#L63) does not matter because the comparison assembly [is instantiated](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Tasks/AssemblyDependency/ReferenceTable.cs#L2375) via a constructor that leaves `asString` null
- [ignoreVersion ](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Shared/AssemblyNameExtension.cs#L712)and [considerRetargetableFlag ](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Shared/AssemblyNameExtension.cs#L727) are [false](https://github.com/Microsoft/msbuild/blob/0a47f62a138b32cb406d058fe7f49c0494aa6c84/src/Shared/AssemblyNameExtension.cs#L646)

### DIFF
--- a/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
+++ b/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
@@ -19,6 +20,13 @@ namespace Microsoft.Build.Engine.UnitTests
 #else
         private const string MSBuildExeName = "MSBuild.exe";
 #endif
+
+        [Fact]
+        public void ShouldBeRunningInTests()
+        {
+            BuildEnvironmentHelper.Instance.RunningTests.ShouldBeTrue();
+        }
+
         [Fact]
         public void GetExecutablePath()
         {
--- a/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
+++ b/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
@@ -22,12 +22,6 @@ namespace Microsoft.Build.Engine.UnitTests
 #endif
 
         [Fact]
-        public void ShouldBeRunningInTests()
-        {
-            BuildEnvironmentHelper.Instance.RunningTests.ShouldBeTrue();
-        }
-
-        [Fact]
         public void GetExecutablePath()
         {
             var msbuildPath = Path.GetDirectoryName(FileUtilities.ExecutingAssemblyPath);
@@ -272,11 +266,7 @@ namespace Microsoft.Build.Engine.UnitTests
             }
         }
 
-#if RUNTIME_TYPE_NETCORE
-        [Fact(Skip = "https://github.com/Microsoft/msbuild/issues/669")]
-#else
         [Fact]
-#endif
         public void BuildEnvironmentDetectsRunningTests()
         {
             BuildEnvironmentHelper.Instance.RunningTests.ShouldBeTrue();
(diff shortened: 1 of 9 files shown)


src/Shared/AssemblyNameExtension.cs
```diff
@@ -734,29 +734,13 @@ private bool EqualsImpl(AssemblyNameExtension that, bool ignoreVersion, bool con
...
-                bCulture = CultureInfo.InvariantCulture;
-            }
-
-            if (aCulture.LCID != bCulture.LCID)
-            {
-                return false;
-            }
-
-            return true;
+            return a?.CultureInfo?.LCID == b?.CultureInfo?.LCID;
```
davkean: This isn't the same comparison as before - if one is Invariant and one is null, this used to return true, now it returns false.

cdmihai: You're right. I made the .net core implementation follow the full framework one and not the other way around

cdmihai:
Apparently `BuildEnvironmentHelper.Instance.RunningTests` does not work on .net core :(. Looking into it.

src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -19,6 +20,13 @@ public class BuildEnvironmentHelper_Tests
 #else
         private const string MSBuildExeName = "MSBuild.exe";
 #endif
+
+        [Fact]
+        public void ShouldBeRunningInTests()
```
AndyGerlicher: https://github.com/cdmihai/msbuild/blob/ca62c0d4a3d9ba003176c4069a976978c59949d7/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs#L275-L284

😄

cdmihai: Best kind of bugfix. :)

davkean:
Can we get this merged in?

AndyGerlicher:
Will submit this plus the others on Monday for escrow.

PR #2686, opened 2017-10-31
Add an issue template

cdmihai:


.github/issue_template.md
```diff
@@ -0,0 +1,12 @@
+### Steps to reproduce
+
```
jeffkl: Maybe some place holder text with an example or two?

.github/issue_template.md
```diff
@@ -0,0 +1,12 @@
...
+### Expected  behavior
+
+
+### Actual behavior
+
+
+### Environment data
+`msbuild /version` output:
+If aplicable, version of tool that invokes MSBuild (Visual Studio, dotnet CLI, etc):
```
jeffkl:
```diff
-aplicable
+applicable
```

.github/issue_template.md
```diff
@@ -0,0 +1,41 @@
+### Steps to reproduce
+
+Example repro steps:
+
+- zip file with project file, directory contents, and script that invokes msbuild.
```
cdmihai: @jeffkl Is it too much to ask for an actual zip file? I assume it would save us time redoing the repro, and also assume the issue creator has the files locally already.

jeffkl: I would make it a suggestion to either paste a project sample or attach a project.  Just word it as a suggestion so that people don't feel like they have to provide that much info just to file something.  But that way they'll know what sort of info we'll need to really dig into some of the issues

kasper3:
Should the branch `issueTemplate` be deleted from main repo?

PR #2690, opened 2017-11-01
Only import NuGet.targets when it exists

AndyGerlicher:
Workaround might be needed until a better solution is found.

* Some scenarios may not have NuGet.targets (e.g. C++ only workload in VS)


emgarten:
LGTM

Issue #2693, opened 2017-11-02
FileUtilities.HasExtension allocating 3.7% of a solution build


davkean:
We should replace this with EndsWith:

![image](https://user-images.githubusercontent.com/1103906/32354855-e1a5d718-c07f-11e7-8df8-0f7850dfec61.png)

davkean:
This is fixed in: https://github.com/Microsoft/msbuild/commit/5a8bd79d4f0422af985c307b378a09360e8bda0c

PR #2695, opened 2017-11-03
Avoid allocating in HasExtension

davkean:
Also changed comparison to ordinal.

This was causing 3.7% of a solution build of customer solution.

Fixes: #2693 

Before:
image

After:
image

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -532,14 +532,23 @@ namespace Microsoft.Build.Shared
         /// <returns></returns>
         internal static bool HasExtension(string fileName, string[] allowedExtensions)
         {
-            string fileExtension = Path.GetExtension(fileName);
-            foreach (string extension in allowedExtensions)
+            Debug.Assert(allowedExtensions != null && allowedExtensions.Length > 0);
+
+            // Easiest way to invoke invalid path chars
+            // check, which callers are relying on.
+            if (Path.HasExtension(fileName))
             {
-                if (String.Compare(fileExtension, extension, PathComparison) == 0)
+                foreach (string extension in allowedExtensions)
                 {
-                    return true;
+                    Debug.Assert(!String.IsNullOrEmpty(extension) && extension[0] == '.');
+
+                    if (fileName.EndsWith(extension, PathComparison))
+                    {
+                        return true;
+                    }
                 }
             }
+
             return false;
         }
 
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -535,7 +535,7 @@ namespace Microsoft.Build.Shared
             string fileExtension = Path.GetExtension(fileName);
             foreach (string extension in allowedExtensions)
             {
-                if (String.Compare(fileExtension, extension, StringComparison.CurrentCultureIgnoreCase) == 0)
+                if (String.Compare(fileExtension, extension, StringComparison.OrdinalIgnoreCase) == 0)
                 {
                     return true;
                 }
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -535,7 +535,7 @@ namespace Microsoft.Build.Shared
             string fileExtension = Path.GetExtension(fileName);
             foreach (string extension in allowedExtensions)
             {
-                if (String.Compare(fileExtension, extension, StringComparison.OrdinalIgnoreCase) == 0)
+                if (String.Compare(fileExtension, extension, PathComparison) == 0)
                 {
                     return true;
                 }
--- a/src/Shared/UnitTests/FileUtilities_Tests.cs
+++ b/src/Shared/UnitTests/FileUtilities_Tests.cs
@@ -241,7 +241,6 @@ namespace Microsoft.Build.UnitTests
         [InlineData("foo.txt",      new[] { ".TXT" })]
         [InlineData("foo.txt",      new[] { ".TXT"})]
         [InlineData("foo.txt",      new[] { ".EXE", ".TXT" })]
-        [InlineData("foo",          new[] { "" })]
         public void HasExtension_WhenFileNameHasExtension_ReturnsTrue(string fileName, string[] allowedExtensions)
         {
             var result = FileUtilities.HasExtension(fileName, allowedExtensions);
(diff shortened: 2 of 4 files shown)


src/Shared/FileUtilities.cs
```diff
@@ -532,10 +532,9 @@ internal static string GetDirectory(string fileSpec)
         /// <returns></returns>
         internal static bool HasExtension(string fileName, string[] allowedExtensions)
         {
-            string fileExtension = Path.GetExtension(fileName);
             foreach (string extension in allowedExtensions)
             {
-                if (String.Compare(fileExtension, extension, StringComparison.CurrentCultureIgnoreCase) == 0)
+                if (fileName.EndsWith(extension, StringComparison.OrdinalIgnoreCase))
```
radical: Should this also confirm that `fileName[0] == '.'` ?

davkean: Path.GetExtension(".exe") returns ".exe" - what would your check cover?

I'll split it into a separate commit, the comparison changed because you don't want this `HasExtension("foo.ini", ".Ini")` to return false when you are running under `tr-TR`.

radical: > Path.GetExtension(".exe") returns ".exe" - what would your check cover?

Oops, I meant `extension[0] == '.'`.

davkean:
@jeffkl @cdmihai @radical Made significant changes that need re-reviewing. Needed to handle the test case failure which was expecting this method to throw on invalid chars.

src/Shared/FileUtilities.cs
```diff
@@ -532,14 +532,23 @@ internal static string GetDirectory(string fileSpec)
...
+
+            // Easiest way to invoke invalid path chars
+            // check, which callers are relying on.
+            if (Path.HasExtension(fileName))
             {
-                if (String.Compare(fileExtension, extension, StringComparison.CurrentCultureIgnoreCase) == 0)
+                foreach (string extension in allowedExtensions)
                 {
-                    return true;
+                    Debug.Assert(!String.IsNullOrEmpty(extension) && extension[0] == '.');
```
radical: IIUC, this (method) is meant for internal use but gets exposed via RAR's extension checking (or others in future?). Isn't this causing a slight difference in behavior now, because earlier it would accept `extension` to be only `.foo` but now it would even accept `.bar.foo`? Just thinking aloud. If it is fine or we don't care about it then maybe we should add a test for this case too?

radical: And similar concern about the `'.'` check too - is this something we care about at runtime? Or maybe we want to be lax on that because it is rare to not have the `'.'` and the performance concern is more important.

jeffkl:
The change looks good to me.  I don't understand why the tests are failing due to the method not throwing though?

Also, should we re-arrange the order of allowed extensions?  Wouldn't `.dll` be the most common case?

https://github.com/Microsoft/msbuild/blob/75e5117ab8be40faad4601904600aceae6bf9ee1/src/Tasks/Microsoft.Common.CurrentVersion.targets#L581-L585

``` C#
private string[] _allowedAssemblyExtensions = new string[] { ".winmd", ".dll", ".exe" };
```

https://github.com/Microsoft/msbuild/blob/master/src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs#L80

jeffkl:
Also, since we're setting the array as part of construction, then immediately setting it from the `$(AllowedReferenceAssemblyFileExtensions)` property, doesn't that mean we're allocating more than we need?  Would be help to do this instead:

``` C#
private string[] _allowedAssemblyExtensions = null;

public string[] AllowedAssemblyExtensions
{
    get
    {
        if (_allowedAssemblyExtensions == null)
        {
            _allowedAssemblyExtensions = new[] {".dll", ".winmd", ".exe"};
        }
        return _allowedAssemblyExtensions;
    }
    set { _allowedAssemblyExtensions = value; }
}
```

Or better yet, don't set a default in XML which allocates.

src/Shared/UnitTests/FileUtilities_Tests.cs
```diff
@@ -234,16 +235,65 @@ public void GetDirectoryWithTrailingSlash()
...
+        public void HasExtension_WhenFileNameDoesNotHaveExtension_ReturnsFalse(string fileName, string[] allowedExtensions)
+        {
+            var result = FileUtilities.HasExtension(fileName, allowedExtensions);
+
+            Assert.False(result);
+        }
+
         [Fact]
-        public void HasExtension()
+        public void HasExtension_WhenInvalidFileName_ThrowsArgumentException()
```
cdmihai: Windows only test: use `[PlatformSpecific(Xunit.PlatformID.Windows)]`

src/Shared/FileUtilities.cs
davkean: @cdmihai Did you reply to these concerns before merging? These look like legitimate concerns.

cdmihai: Initially I thought it would be OK, but now I realize it will catch files like `foo.config.exe`, which is not OK. I'll fix this up to only compare starting from the last '.' and send a PR.
Regarding the `.`, the original code was not checking for it, so I don't whether it would be good to start failing it now.

cdmihai: Ah, I remember now. With both the previous version of the code and the current version, `.exe` would still catch `foo.appconfig.exe`, because `Path.GetExtension("foo.appconfig.exe")` returns ".exe". The only difference now is that before, if you told RAR to start collecting extensions like ".config.exe" it would not get them, but now it would. So this is only a change in behaviour if somebody modified the inputs to RAR to include multi dot extensions and expected RAR to not find them, which means they expect RAR to do a no-op for those inputs. Since that seems plausible, and since it does not affect RAR's output, I decided it's fine.

Let me know if I missed something.

radical: > before, if you told RAR to start collecting extensions like ".config.exe" it would not get them, but now it would.

Yep, that is exactly what I was talking about.

> So this is only a change in behaviour if somebody modified the inputs to RAR to include multi dot extensions ...

Makes sense. Probably nitpicking: But if it is so specific to RAR, then maybe it shouldn't be in `FileUtilities`. Or should be renamed
to reflect that. Or add a comment to warn future non-RAR callers.

cdmihai: Or replace the other potentially existing extension checking code clones to reuse this one :)

PR #2720, opened 2017-11-14
Add flag to ignore faulted imports

### LINKED ISSUE #2179: Add ability to ignore invalid imports
Similar to #1727, we should add a `ProjectLoadSettings.IgnoreInvalidImports` which will evaluate a project even if one or more imports have invalid XML or something else wrong with them.

From @jviau 
> I am wondering if we can even go a step further and have a way to handle/ignore invalid imports. I implemented a feature for CPS to auto-reload changed imports on disk, and to revert back to in-memory XML if the import is now invalid. The problem is I can't deal with the case where a condition import transitions from false to true and the newly imported file is invalid. My system relies on file changed notifications, and there is no way to know that some arbitrary change is going to bring in a new invalid import.

jviau:
Add `ProjectLoadSettings.IgnoreFaultedImports` flag, behaving like `ProjectLoadSettings.IgnoreEmptyImports`, where an import that throws `InvalidProjectFileException`  will be logged and removed from that evaluation.

This change is to help improve CPS auto-reload of imports. Right now there is no common point for CPS to handle exceptions thrown from evaluation. With auto-reload of imports, it is not uncommon for an invalid project file to be saved, and attempted to reload. Our current design is to validate the file before reloading it, but this is expensive as it requires validating the XML as well as seeing if any project configuration would throw. Even then it is not a perfect system as we can only validate imports we know exist. For example, a conditioned import that goes from false to true and is invalid can still cause an error dialog or even crash VS. With this change we can tolerate any faulted import.

I would like to make 2 changes after this as well:
1. The event args for a skipped import should be an error or at least a warning.
2. Need to be able to identify what imports were skipped. This can either be a new property on the `Project` class, or the log messages can be uniquely identifiable (new type?) so CPS can track them via log messages.

/cc @lifengl @davkean 

Fixes #2179

### DIFF
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -4060,33 +4060,20 @@ namespace Microsoft.Build.UnitTests.OM.Definition
 
                     Project unused = new Project(pre, null, null, collection, ProjectLoadSettings.IgnoreFaultedImports);
 
-                    IEnumerable<ProjectImportedEventArgs> eventArgs = logger.AllBuildEvents.Where(i => i is ProjectImportedEventArgs).Cast<ProjectImportedEventArgs>();
+                    ProjectImportedEventArgs eventArgs = logger.AllBuildEvents.SingleOrDefault(i => i is ProjectImportedEventArgs) as ProjectImportedEventArgs;
 
-                    ProjectImportedEventArgs ignoreImportArg = eventArgs.FirstOrDefault();
-                    ProjectImportedEventArgs faultArg = eventArgs.LastOrDefault();
+                    Assert.NotNull(eventArgs);
 
-                    Assert.NotNull(ignoreImportArg);
-                    Assert.NotNull(faultArg);
-                    Assert.NotSame(ignoreImportArg, faultArg);
+                    Assert.Equal(import.Project, eventArgs.UnexpandedProject);
 
-                    Assert.Equal(import.Project, ignoreImportArg.UnexpandedProject);
-                    Assert.Equal(import.Project, faultArg.UnexpandedProject);
+                    Assert.Null(eventArgs.ImportedProjectFile);
 
-                    Assert.Null(ignoreImportArg.ImportedProjectFile);
-                    Assert.Null(faultArg.ImportedProjectFile);
+                    Assert.Equal(pre.FullPath, eventArgs.ProjectFile);
 
-                    Assert.Equal(pre.FullPath, ignoreImportArg.ProjectFile);
-                    Assert.Equal(pre.FullPath, faultArg.ProjectFile);
+                    Assert.Equal(6, eventArgs.LineNumber);
+                    Assert.Equal(3, eventArgs.ColumnNumber);
 
-                    Assert.Equal(6, ignoreImportArg.LineNumber);
-                    Assert.Equal(3, ignoreImportArg.ColumnNumber);
-
-                    Assert.Equal(2, faultArg.LineNumber);
-                    Assert.Equal(1, faultArg.ColumnNumber);
-
-                    logger.AssertLogContains(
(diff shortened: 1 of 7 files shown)


src/Build/Definition/ProjectLoadSettings.cs
```diff
@@ -48,6 +48,11 @@ public enum ProjectLoadSettings
         /// By default, evaluations performed via <see cref="Project"/> evaluate and collect elements whose conditions were false (e.g. <see cref="Project.ItemsIgnoringCondition"/>).
         /// This flag turns off this behaviour. <see cref="Project"/> members that collect such elements will throw when accessed.
         /// </summary>
-        DoNotEvaluateElementsWithFalseCondition = 32
+        DoNotEvaluateElementsWithFalseCondition = 32,
+
+        /// <summary>
+        /// Ignore faulted target files when evaluating the project
+        /// </summary>
+        IgnoreFaultedImports = 64,
```
jeffkl: I like the name `IgnoreInvalidImports` better.

jviau: Renamed to `Invalid`

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2668,6 +2668,32 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
                                 continue;
                             }
 
+                            // If IgnoreFaultedImports is enabled, log all other non-handled exceptions and continue
```
jeffkl: Update comment `IgnoreInvalidImports`

jviau:
@jeffkl  and @cdmihai, what are your thoughts on the 2 additional changes I want to make? Should ignored imports be logged as an error or warning, and should they be a new event args derived from error or warning? And what are the thoughts of adding the property `Project.IgnoreImports`?

src/Build.OM.UnitTests/Definition/Project_Tests.cs
```diff
@@ -4035,6 +4035,50 @@ public void ProjectImportedEventEmptyFile()
...
+        public void ProjectImportedEventFaultedFile()
+        {
+            using (var env = TestEnvironment.Create(_output))
+            {
+                env.SetEnvironmentVariable("MSBUILDLOGIMPORTS", "1");
+
+                const string contents = @"<?xml version=""1.0"" encoding=""utf-8""?>
+<Project>BROKEN</Project>
+";
+                string importPath = ObjectModelHelpers.CreateFileInTempProjectDirectory(Guid.NewGuid().ToString("N"), contents, Encoding.UTF8);
```
cdmihai: You can use Env.CreateFile as well here, to ensure it gets deleted afterwards. TestEnvironment is new, so many older tests are not as clean as they could be.

jviau: I pretty much copy pasted the `IgnoreEmptyImports` test. Do you want me to update that one as well?

cdmihai: Sure, thanks!

---
In reply to: [150973447](https://github.com/Microsoft/msbuild/pull/2720#discussion_r150973447) [](ancestors = 150973447)

cdmihai: Sure, thanks!

jeffkl:
Would it be enough to add some properties to the `ProjectImportedEventArgs` that you could read?  To me the point of the flags is to "ignore" certain conditions.  Logging a warning or error would not exactly be "ignoring" the issue

jviau:
True, the naming does indicate a regular event may be better. Adding `bool Ignored` to `ProjectImportedEventArgs` is sufficient. I am still a bit partial to logging them as errors, as I want to show any errors the user would receive if they trying to build the project from command line. I can always re-log them as errors on the CPS side though.

src/Build.OM.UnitTests/Definition/Project_Tests.cs
jviau: I updated both tests to use `env.CreateFile`

jviau:

jviau:
I'll make a separate PR for adding the extra information to `ProjectImportedEventArgs`

PR #2722, opened 2017-11-16
Skipped import logging

jviau:
Improves logging when an import is ignored.

1. Sets the `ImportedProjectFile` property
2. ~Adds `ProjectImportedEventArgs.ImportException` when an import is ignored. Had to be type `Exception` because `Microsoft.Build.Framework` does not reference `Microsoft.Build`, where that exception lives.~ Exception fails to serialize for .netcore, using `bool ImportIgnored` instead.
3. Logs for `IgnoreMissingImports`

### DIFF
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -4105,7 +4105,7 @@ namespace Microsoft.Build.UnitTests.OM.Definition
                     MockLogger logger = new MockLogger();
                     collection.RegisterLogger(logger);
 
-                    Project unused = new Project(pre, null, null, collection, ProjectLoadSettings.IgnoreInvalidImports);
+                    Project unused = new Project(pre, null, null, collection, ProjectLoadSettings.IgnoreMissingImports);
 
                     ProjectImportedEventArgs eventArgs = logger.AllBuildEvents.SingleOrDefault(i => i is ProjectImportedEventArgs) as ProjectImportedEventArgs;
 
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -4020,7 +4020,7 @@ namespace Microsoft.Build.UnitTests.OM.Definition
                     ProjectImportedEventArgs eventArgs = logger.AllBuildEvents.SingleOrDefault(i => i is ProjectImportedEventArgs) as ProjectImportedEventArgs;
 
                     Assert.NotNull(eventArgs);
-                    Assert.NotNull(eventArgs.ImportException);
+                    Assert.True(eventArgs.ImportIgnored);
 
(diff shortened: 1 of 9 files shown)


src/Build.OM.UnitTests/Definition/Project_Tests.cs
```diff
@@ -4082,6 +4084,49 @@ public void ProjectImportedEventInvalidFile()
         }
 
         [Fact]
+        public void ProjectImportedEventMissingFile()
```
cdmihai: If it does not already exist, can you also add a test for a glob based import where the glob expands to nothing? In that case, the event should still be there, but `ImportedProjectFile` should be empty.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -259,6 +259,7 @@ private void Write(ProjectImportedEventArgs e)
         {
             Write(BinaryLogRecordKind.ProjectImported);
             WriteMessageFields(e);
+            Write(e.ImportIgnored);
```
cdmihai: Ah, almost forgot. Please increment Microsoft.Build.Logging.BinaryLogger.FileFormatVersion and append to the comment. :)

cdmihai: FYI @KirillOsenkov

src/Build.OM.UnitTests/Definition/Project_Tests.cs
jviau: Do you think `ImportIgnored` should be true in this case?

cdmihai: I agree with you. Can you please add a comment about this on ImportIgnored?

jviau: Done

KirillOsenkov:
See https://github.com/Microsoft/msbuild/pull/2729 on how to make the reader backwards compatible.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -132,6 +132,7 @@ private BuildEventArgs ReadProjectImportedEventArgs()
             var fields = ReadBuildEventArgsFields();
             // Read unused Importance, it defaults to Low
             ReadInt32();
+            bool ignored = ReadBoolean();
```
KirillOsenkov: After https://github.com/Microsoft/msbuild/pull/2729 goes in, we need to surround this place with an if statement. This doesn't need to block on that PR but it's important to do it soon after.

KirillOsenkov: Basically 
```
bool ignored = false;

// the import ignored field was added in file format version 3
if (fileFormatVersion > 2)
{
    ignored = ReadBoolean();
}
...
e.ImportIgnored = ignored;
```

PR #2726, opened 2017-11-16
Make it easier to opt into "Core" satellite assembly gen

tmeschter:
Currently the `GenerateSatelliteAssemblies` target handles generating satellite assemblies, _unless_ we're using the .NET Core version of MSBuild. In this case we skip `GenerateSatelliteAssemblies` and this is instead handled by the `CoreGenerateSatelliteAssemblies` target in the .NET SDK. This is because `GenerateSatelliteAssemblies` utilizes alink (al.exe) which has no counterpart for .NET Core. Instead, `CoreGenerateSatelliteAssemblies` uses csc.exe to create the assembly.

There are times where it would be helpful to use the .NET Core approach, even when we aren't using the .NET Core MSBuild. For example, al.exe has some key limitations. For example, it has [no support for public-signing](https://github.com/dotnet/roslyn/issues/23191) and [poor handling of AssemblyInformationalVersionAttribute](https://github.com/dotnet/roslyn/issues/23190). These issues are currently blocking the build of the dotnet/roslyn repo from generating its own satellite assemblies.

Opting _into_ `CoreGenerateSatelliteAssemblies` is easily done by setting the `GenerateSatelliteAssembliesForCore` property to `true`. However, there is currently no good way to opt _out_ of `GenerateSatelliteAssemblies`. This commit replaces the broad check that we're not using the "Core" runtime type with a specific check that `GenerateSatelliteAssembliesForCore` is not set to `true`.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -3550,7 +3550,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Target Name="GenerateSatelliteAssemblies"
           Inputs="$(MSBuildAllProjects);@(_SatelliteAssemblyResourceInputs);$(IntermediateOutputPath)$(TargetName)$(TargetExt)"
           Outputs="$(IntermediateOutputPath)%(Culture)\$(TargetName).resources.dll"
-          Condition="'@(_SatelliteAssemblyResourceInputs)' != '' and '$(MSBuildRuntimeType)' != 'Core'">
+          Condition="'@(_SatelliteAssemblyResourceInputs)' != '' and '$(GenerateSatelliteAssembliesForCore)' != 'true'">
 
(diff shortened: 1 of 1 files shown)


tmat:
Why isn't this the default?

cdmihai:
@tmat Good question

@tmeschter Does the AL based satellite generation have any features not present in the csc based satellite generation? If not, should we just replace GenerateSatelliteAssemblies with CoreGenerateSatelliteAssemblies? I do not see any back compat issues, but I am also not very intimately familiar with this.

tmeschter:
Except for a very limited set of consumers (currently just dotnet/roslyn) there's no value in using CoreGenerateSatelliteAssemblies over GenerateSatelliteAssemblies. There will almost certainly be back-compat issues (example: until yesterday CoreGenerateSatelliteAssemblies only supported C#).

At some point making it the default might be the right thing to do, but not right now.

PR #2729, opened 2017-11-17
Add ability to replay .binlogs of older versions.

KirillOsenkov:
This allows for backwards compatibility, when a newer version of MSBuild is still able to replay .binlog files created with an older version of MSBuild.

### DIFF
--- a/src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
+++ b/src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
@@ -582,9 +582,10 @@ namespace Microsoft.Build.Logging
             int projectInstanceId = ReadInt32();
 
             // evaluationId was introduced in format version 2
+            int evaluationId = BuildEventContext.InvalidEvaluationId;
             if (fileFormatVersion > 1)
             {
-                int evaluationId = ReadInt32();
+                evaluationId = ReadInt32();
             }
 
             var result = new BuildEventContext(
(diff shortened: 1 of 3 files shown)


KirillOsenkov:
This sets up the pattern for incrementing the .binlog format version.

Every change to the reader needs to be wrapped in an if statement like `if (fileFormatVersion > version where this field was introduced)`.

Even though we've never shipped the file format version 1, it didn't have evaluation context Id, so I'm adding this as a usage example.

KirillOsenkov:
Unfortunately we can't have forward compatibility, where an older version of MSBuild is able to replay .binlogs created with a newer version of MSBuild. We can never anticipate how the format will change since it's not self-describing, i.e. it doesn't carry a schema around (for performance and compactness reasons).

Good news is that I will always keep the MSBuild Structured Log Viewer up-to-date and it will always be able to display .binlogs of any version. I just need to be notified of all changes to the format. Ideally just open an issue over at https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/new every time the binlog format is incremented.

src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -578,7 +580,12 @@ private BuildEventContext ReadBuildEventContext()
             int taskId = ReadInt32();
             int submissionId = ReadInt32();
             int projectInstanceId = ReadInt32();
-            int evaluationId = ReadInt32();
+
+            // evaluationId was introduced in format version 2
+            if (fileFormatVersion > 1)
+            {
+                int evaluationId = ReadInt32();
```
cdmihai: BuildEventContext has a constructor without the evaluationID but all the rest. Or just initialize the id with BuildEventContext.InvalidEvaluationId

KirillOsenkov: LOL this is how we know it's been a long Friday :)

PR #2739, opened 2017-11-22
Fix for BuildProjectReferences=false case in ResolveProjectReferences

erikma:
When BuildProjRefs is false the IDE <MSBuild> targets in ResolveProjectReferences should not trigger and build dependencies.


erikma:
Canceling PR - on further testing this breaks other targets downstream. Searching for a better solution.

PR #2748, opened 2017-11-24
Fix evaluation logging for multi-proc builds

panopticoncentral:
When working on a build log viewer, I noticed that there are two issues with evaluation log events in multi-proc builds:

* Because the evaluation ID was a static on the evaluator, each node reused evaluation IDs, resulting in non-unique and overlapping evaluation IDs on events in the log.
* The evaluation ID was not serialized for messages and so was getting lost when de-serialized in the central node.

I know there are a lot of considerations around this, so let me know if something looks wrong about what I've done.

cc: @KirillOsenkov


KirillOsenkov:
Great find! This explain this issue my users have been reporting:
https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/132

src/Framework/BuildEventArgs.cs
```diff
@@ -311,7 +312,8 @@ internal virtual void CreateFromStream(BinaryReader reader, int version)
                 {
                     int submissionId = reader.ReadInt32();
                     int projectInstanceId = reader.ReadInt32();
-                    buildEventContext = new BuildEventContext(submissionId, nodeId, projectInstanceId, projectContextId, targetId, taskId);
+                    int evaluationId = reader.ReadInt32();
```
cdmihai: For back compat, you should bump up the format version number, and read the new value only when the format version is greater than or equal the new version value.

cdmihai:
ProjectInstance.Initialize has some [postconditions ](https://github.com/Microsoft/msbuild/blob/master/src/Build/Instance/ProjectInstance.cs#L2459-L2463) that an evaluation ID should have been assigned. Can you please add the equivalent for [Project.Reevaluate](https://github.com/Microsoft/msbuild/blob/master/src/Build/Definition/Project.cs#L2641) as well?

cdmihai:
Can you please add a test for the out of proc case? You can use [this test](https://github.com/Microsoft/msbuild/blob/master/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs#L3304) as an example.

src/Framework/BuildEventArgs.cs
panopticoncentral: This is binary serialization which looks like the version number is derived from the CLR version. It looks like I shouldn't mess with that?

panopticoncentral:
@cdmihai I pushed changes, can you take a look?

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -3564,5 +3564,70 @@ public void OutOfProcProjectInstanceBasedBuildDoesNotReloadFromDisk(bool shouldS
...
+                    new[] { "MainTarget" },
+                    null
+                );
+
+                var submission = _buildManager.PendBuildRequest(buildRequestData);
+
+                var result = submission.Execute();
+
+                Assert.Equal(BuildResultCode.Success, result.OverallResult);
+                Assert.True(_logger.AllBuildEvents.OfType<ProjectEvaluationStartedEventArgs>().GroupBy(args => args.BuildEventContext.EvaluationId).All(g => g.Count() == 1));
```
cdmihai: You should also assert that there are exactly two groups, right? There should have been two evaluations, each with different evaluations IDs

src/Framework/BuildEventArgs.cs
cdmihai: You're right. I had confused this with the binary logging serialization, but that one's not affected by this PR.

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
cdmihai: Looks like this is failing on .net core. Can you please take a look? Ping me if you need another set of eyes on it.

panopticoncentral: @cdmihai What is the best way to debug a CoreCLR test? I see how to build it but not sure the easiest way to get it under the debugger.

panopticoncentral: Nevermind, looked at the docs and figured it out. I think I've fixed the bug.

panopticoncentral:

panopticoncentral:
@cdmihai Looks like everything is green, what do you think?

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
cdmihai: Can you also assert there's two groups? Or better yet, instead of grouping, assert that there are exactly two evaluation ids, and they're different.

AndyGerlicher:

AndyGerlicher:
@panopticoncentral Added a merge commit, hopefully it's correct?

panopticoncentral:
:+1:

panopticoncentral:
Thanks!

PR #2749, opened 2017-11-25
Open the .binlog file as read-only during playback.

KirillOsenkov:
This avoids a possible File Access exception when reading the binlog from a network share where you don't have write access.

### DIFF
--- a/src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
+++ b/src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Build.Logging
         /// <param name="sourceFilePath">The full file path of the binary log file</param>
         public void Replay(string sourceFilePath)
         {
-            using (var stream = new FileStream(sourceFilePath, FileMode.Open))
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
Interesting that the default file access is ReadWrite.

src/Build/Logging/BinaryLogger/BinaryLogReplayEventSource.cs
```diff
@@ -19,7 +19,7 @@ public sealed class BinaryLogReplayEventSource : EventArgsDispatcher
         /// <param name="sourceFilePath">The full file path of the binary log file</param>
         public void Replay(string sourceFilePath)
         {
-            using (var stream = new FileStream(sourceFilePath, FileMode.Open))
+            using (var stream = new FileStream(sourceFilePath, FileMode.Open, FileAccess.Read))
```
jaredpar: Should we make FileShare explicit as well while we're changing this?

KirillOsenkov: Would you suggest just adding `FileShare.Read` there?

cdmihai: That's the default, and it seems sensible to me.

PR #2761, opened 2017-11-29
WIP Update to netstandard2.0

AndyGerlicher:
For transparency, here is the current WIP for the upgrade that @dsplaisted has been working on:

> I’ve pushed my work in progress for targeting .NET Core 2.0 to the exp/update-toolset branch.
> 
> I was able to get the bootstrap build working for .NET Framework.  If you run cibuild.cmd, by default it will create a bootstrapped version of MSBuild, then do another build using that version and run the tests.  (I think the second build will probably not actually do anything due to incrementality.)
> 
> The path to the bootstrapped MSBuild.exe for Full Framework is artifacts\Debug\bootstrap\net46\MSBuild\15.0\Bin\MSBuild.exe.


DotnetCLIVersion.txt
```diff
@@ -1 +1 @@
-1.0.0-preview3-003686
\ No newline at end of file
+2.0.2
```
4creators: This should be upgraded to 2.0.3 as 2.0.2 contains security bug.

AndyGerlicher:
@dsplaisted will create a new one.

PR #2762, opened 2017-11-29
Fix reference to Microsoft.Build.Tasks.CodeAnalysis

AndyGerlicher:
* Update to use NuGet package
 * Update sournces feed to use Roslyn nightly

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -3,6 +3,7 @@
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
+    <RestoreSources>$(RestoreSources);https://dotnet.myget.org/F/roslyn/api/v3/index.json</RestoreSources>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


build/Versions.props
```diff
@@ -37,6 +38,7 @@
   <PropertyGroup>
     <MicrosoftNetCompilersVersion>2.4.0</MicrosoftNetCompilersVersion>
     <MicrosoftNetCoreCompilersVersion>2.6.0-beta3-62309-01</MicrosoftNetCoreCompilersVersion>
+    <MicrosoftCodeAnalysisBuildTasksVersion>3.0.0-beta1-61516-01</MicrosoftCodeAnalysisBuildTasksVersion>
```
akoeplinger: According to https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.CodeAnalysis.Build.Tasks the 3.0.0-beta1-61516-01 version was pushed on Thu, 16 Mar 2017 while e.g. 2.6.0-beta3-62329-03 was pushed yesterday, is this the correct version to use?

AndyGerlicher: Doubt it. This entire branch needs to get reviewed. It just didn't build as-is so I wanted to get that in quick. Good catch though, will add it to the list to ensure they're in sync (if possible).

PR #2764, opened 2017-12-01
Add collect referenced nuget packages target

jmarolf:
adding CollectReferencedNuGetPackages so consumers that partially import VS targets are not broken.

VS Bug [526825](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/526825)

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -731,7 +731,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
                                         _CheckForInvalidConfigurationAndPlatform
 
     This target checks for errors in statically defined properties.  By setting BeforeTargets, we try
-    to ensure that the target runs before any build related targets.  
+    to ensure that the target runs before any build related targets.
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -5656,6 +5656,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <!-- Get the resolved SDK reference items -->
   <Target Name="GetResolvedSDKReferences" DependsOnTargets="ResolveSDKReferences" Returns="@(ResolvedSDKReference)"/>
 
+  <!-- Target will return no results by default but will be overridden by other targets elsewhere.
```
Pilchie: I don't think this comment is technically accurate.  This target isn't overridden anywhere, it's just that other targets may happen to populate `ReferencedNuGetPackages` as a side-effect, and we want to collect them as a return value if they do.

AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:
Per email thread changing to master for 15.6 (FYI I had to rebase and push your branch).

PR #2786, opened 2017-12-08
Fix issue with processing Directory.Build.rsp

jeffkl:
# 😢 😒 🐛 ⛑ 🔨 🔧 🕐 🕑 🕒 🕓 🕔 💡 👌 😴 💤 
The logic was broken because the `found` variable is not set to true when switches are read from the Directory.Build.rsp.  If `found` is `false`, then the method is not called recursively and the switches from the response file are not added to the list of parameters used.


jeffkl:

PR #2788, opened 2017-12-08
Add profile data to binary log file

AndyGerlicher:
Quick stab at adding this. @KirillOsenkov  look right?

FYI @smera

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -988,7 +988,7 @@ namespace Microsoft.Build.Evaluation
                 }
             }
 
-            _evaluationProfiler.AssertEmpty();
+            ErrorUtilities.VerifyThrow(_evaluationProfiler.IsEmpty(), "Evaluation profiler stack is not empty.");
             _evaluationLoggingContext.LogBuildEvent(new ProjectEvaluationFinishedEventArgs(ResourceUtilities.GetResourceString("EvaluationFinished"), projectFile)
             {
                 BuildEventContext = _evaluationLoggingContext.BuildEventContext,
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -393,13 +393,7 @@ namespace Microsoft.Build.Evaluation
                 DataCollection.CommentMarkProfile(8812, beginProjectEvaluate);
 #endif
                 Evaluator<P, I, M, D> evaluator = new Evaluator<P, I, M, D>(data, root, loadSettings, maxNodeCount, environmentProperties, itemFactory, toolsetProvider, projectRootElementCache, projectInstanceIfAnyForDebuggerOnly, sdkResolution);
-                IDictionary<string, object> projectLevelLocalsForBuild;
-                using (evaluator._evaluationProfiler.TrackPass(EvaluationPass.TotalEvaluation))
-                {
-                    projectLevelLocalsForBuild = evaluator.Evaluate(loggingService, buildEventContext);
-                }
-
-                return projectLevelLocalsForBuild;
+                return evaluator.Evaluate(loggingService, buildEventContext);
 #if MSBUILDENABLEVSPROFILING 
             }
             finally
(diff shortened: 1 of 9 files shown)


src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -21,10 +21,9 @@ public sealed class BinaryLogger : ILogger
         //   - new record kinds: ProjectEvaluationStarted, ProjectEvaluationFinished
         // version 3:
         //   - new ProjectImportedEventArgs.ImportIgnored
-        // version 4:
-        //   - new TargetSkippedEventArgs
-        //   - new TargetStartedEventArgs.BuildReason
-        internal const int FileFormatVersion = 4;
```
KirillOsenkov: Are you erasing version 4? Probably just a merge conflict.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -129,6 +130,18 @@ private void Write(ProjectEvaluationFinishedEventArgs e)
             Write(BinaryLogRecordKind.ProjectEvaluationFinished);
             WriteBuildEventArgsFields(e);
             Write(e.ProjectFile);
+
+            Write(e.ProfilerResult.HasValue);
+            if (e.ProfilerResult.HasValue && e.ProfilerResult.Value.ProfiledLocations.Count > 0)
```
KirillOsenkov: >e.ProfilerResult.Value.ProfiledLocations.Count > 0 [](start = 45, length = 50)

if it HasValue, but the Count is zero, this will confuse the reader. The condition here must match what we're writing in the line above.

KirillOsenkov:
Looks great!

Just make sure you don't erase format version 4 in BinaryReader and double-check that HasValue and Count always match.

KirillOsenkov:
Bonus points for writing a manual serialization, this is how I would do it too. Decouple the file format from the current data structure implementation.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Don't forget to fix this.

AndyGerlicher:
@smera @KirillOsenkov 

This took way too long to figure out. Turns out the `}` of the using was happening after the evaluation finished event fired. So the data was getting mutated after the binary logger wrote out the data and there were inconsistencies between runs.

And Kirill, good catch. I made it more explicit now even if the count is 0 is should deserialize to the same state.

smera:
Good catch!

src/MSBuild/XMake.cs
```diff
@@ -2249,15 +2260,23 @@ internal static bool ProcessRestoreSwitch(string[] parameters)
         /// and also returns the created logger. Otherwise, the collection of loggers is not affected and null
         /// is returned
         /// </remarks>
-        internal static ProfilerLogger ProcessProfileEvaluationSwitch(string[] parameters, ArrayList loggers)
+        internal static ProfilerLogger ProcessProfileEvaluationSwitch(string[] parameters, ArrayList loggers, out bool enableProfiler)
         {
+            enableProfiler = false;
```
smera: >enableProfiler = false; [](start = 12, length = 23)

nit: push inside if

src/MSBuild/XMake.cs
```diff
@@ -1086,7 +1089,7 @@ ProfilerLogger profilerLogger
 
                     // Propagate the profiler flag into the project load settings so the evaluator
                     // can pick it up
-                    if (profilerLogger != null)
+                    if (profilerLogger != null || enableProfiler)
```
smera: >(profilerLogger != null || enableProfile [](start = 23, length = 40)

shouldn't this be 'enableProfiler'  directly?

src/Framework/Profiler/EvaluationLocation.cs
```diff
@@ -163,31 +163,33 @@ public override bool Equals(object obj)
...
+        public override int GetHashCode()
         {
-            return $"{EvaluationDescription ?? string.Empty}\t{File ?? string.Empty}\t{Line?.ToString() ?? string.Empty}\t{ElementName ?? string.Empty}";
+            unchecked
+            {
+                var hashCode = (int) EvaluationPass;
+                hashCode = (hashCode * 397) ^ (EvaluationDescription != null ? EvaluationDescription.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (File != null ? File.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ Line.GetHashCode();
+                hashCode = (hashCode * 397) ^ (ElementName != null ? ElementName.GetHashCode() : 0);
```
smera: >(ElementName != null ? ElementName.GetHashCode() [](start = 46, length = 48)

just curious, why the change? The default equality comparer does exactly that (the null check).

And I think the original code was what resharper generates by default

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -24,7 +24,9 @@ public sealed class BinaryLogger : ILogger
         // version 4:
         //   - new TargetSkippedEventArgs
         //   - new TargetStartedEventArgs.BuildReason
-        internal const int FileFormatVersion = 4;
+        // version 5:
+        //   - new EvaluationFinished.ProfilerResult
+        internal const int FileFormatVersion = 5;
```
smera: >= 5; [](start = 45, length = 4)

quick question. I have another small PR in progress that modifies the evaluation location object to add specific tracking for globbing.
This actually changes the structure of the location and turns the 'is element'  field into an enum.
Not a big deal, but we will need to bump the version of the binary logger again and do the appropriate checks.
Maybe we want this other PR to be pushed first so we don't need to do this?

src/MSBuild/XMake.cs
AndyGerlicher: I didn't do this since it really should conceptually be either. If the logger is attached (somehow) it needs to be set. Or if we want to set it explicitly via the other mechanism.

src/Framework/Profiler/EvaluationLocation.cs
AndyGerlicher: I suspected this could be the cause of the data loss I was seeing. Both the GetHashCode and equality were not completely correct (GetHashCode was missing members). I regenerated with resharper now it's more like it was before. Honestly don't have any preference, I just 1) don't want to write it and 2) I want it to take all the fields into account :). Especially in this case where it's used as a Dictionary key.

smera: Yeah, I don't have any preference either, was just curious about the change.

I thought the data loss was due to the event creation being inside a using(). The missing members shouldn't have had any impact on that in practical terms (even though the correct equality is what you did). The missing field was IsElement, which has a strong correlation with the item name/condition, so the previous equality was likely giving the same results as the corrected one

---
In reply to: [156247662](https://github.com/Microsoft/msbuild/pull/2788#discussion_r156247662) [](ancestors = 156247662)

src/MSBuild/XMake.cs
smera: The slightly weird thing to me is that having a profiler logger != null and a disabled profiler at the same time doesn't seem to make a lot of sense.

So IMO ideally 'enableProfiler' should represent the fact that the profiler should be on (which I think it is already happening), and you shouldn't need to inspec the profiler logger.

Anyway, this is very minor, feel free to ignore

---
In reply to: [156247321](https://github.com/Microsoft/msbuild/pull/2788#discussion_r156247321) [](ancestors = 156247321)

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -1352,7 +1352,8 @@ public void ProcessProfileEvaluationInvalidFilename(string filename)
         {
             try
             {
-                MSBuildApp.ProcessProfileEvaluationSwitch(new string[] {filename}, new ArrayList());
+                bool enableProfiler = false;
+                MSBuildApp.ProcessProfileEvaluationSwitch(new string[] {filename}, new ArrayList(), out enableProfiler);
```
cdmihai: nit: Use Should.Throw

src/MSBuild/XMake.cs
cdmihai: What about adding an assert: if profiler != null, then enabledProfiler == true.

---
In reply to: [156247321](https://github.com/Microsoft/msbuild/pull/2788#discussion_r156247321) [](ancestors = 156247321)

src/Build/Evaluation/Evaluator.cs
```diff
@@ -938,64 +935,66 @@ private static ProjectTargetInstance ReadNewTargetElement(ProjectTargetElement t
...
-                {
-                    BuildEventContext = _evaluationLoggingContext.BuildEventContext,
-                    ProjectFile = projectFile,
-                    ProfilerResult = (_loadSettings & ProjectLoadSettings.ProfileEvaluation) != 0 ? (ProfilerResult?)_evaluationProfiler.ProfiledResult : null
-                });
+                    _data.FinishEvaluation();
+                }
             }
+
+            _evaluationProfiler.AssertEmpty();
```
cdmihai: nit: change call to IsEmpty and do the assertion here.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
cdmihai: I'd rather we didn't have an extra version. @smera, can you rebase your work on this PR so you can reuse the version?

---
In reply to: [156224871](https://github.com/Microsoft/msbuild/pull/2788#discussion_r156224871) [](ancestors = 156224871)

smera: Sure, as soon as it is ready to be pushed I'll rebase on top of this one, and then I can push both together

---
In reply to: [156255138](https://github.com/Microsoft/msbuild/pull/2788#discussion_r156255138) [](ancestors = 156255138,156224871)

PR #2792, opened 2017-12-09
Enable cache for RAR

AndyGerlicher:
Preliminary numbers on a basic netcoreapp2.0 -> netstandard2.0 library sln:

Tip of `exp/update-toolset` branch rebuild:
```
      439 ms  ResolveAssemblyReferences                  2 calls
      437 ms  ResolveAssemblyReference                   2 calls
```

This branch with RAR *.Core.cache file:
```
      238 ms  ResolveAssemblyReferences                  2 calls
      237 ms  ResolveAssemblyReference                   2 calls
```
This branch after deleting *.Core.cache:
```
      312 ms  ResolveAssemblyReferences                  2 calls
      311 ms  ResolveAssemblyReference                   2 calls
```

So even without the cache file it's still after with binary serialization turned on. Not sure why (there's a lot of code that differs). Also haven't done anything other than manual testing.

### DIFF
--- a/src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
@@ -361,13 +361,8 @@ namespace Microsoft.Build.UnitTests.BackEnd
                 1,
                 Directory.GetCurrentDirectory(),
                 null,
-#if FEATURE_THREAD_CULTURE
                 Thread.CurrentThread.CurrentCulture,
                 Thread.CurrentThread.CurrentUICulture,
-#else
-                CultureInfo.CurrentCulture,
-                CultureInfo.CurrentCulture,
-#endif
 #if FEATURE_APPDOMAIN
                 null,
 #endif
@@ -376,11 +371,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
                 @"c:\my project\myproj.proj",
                 _continueOnErrorDefault,
                 "TaskName",
-#if FEATURE_ASSEMBLY_LOADFROM
                 @"c:\MyTasks\MyTask.dll",
-#else
-                new AssemblyName("MyTask"),
-#endif
                 parameters);
 
             ((INodePacketTranslatable)config).Translate(TranslationHelpers.GetWriteTranslator());
--- a/src/Directory.BeforeCommon.targets
+++ b/src/Directory.BeforeCommon.targets
@@ -65,6 +65,7 @@
     <DefineConstants Condition="'$(MachineIndependentBuild)' != 'true'">$(DefineConstants);FEATURE_REGISTRY_SDKS</DefineConstants>
     <DefineConstants Condition="'$(MachineIndependentBuild)' != 'true'">$(DefineConstants);FEATURE_REGISTRYHIVE_DYNDATA</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_RESGEN</DefineConstants>
+    <DefineConstants>$(DefineConstants);FEATURE_RESGENCACHE</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_RESOURCE_EXPOSURE</DefineConstants>
     <!-- System.Resources.ResourceManager.GetResourceSet() method is currently only in full framework -->
     <DefineConstants>$(DefineConstants);FEATURE_RESOURCEMANAGER_GETRESOURCESET</DefineConstants>
--- a/src/Directory.BeforeCommon.targets
+++ b/src/Directory.BeforeCommon.targets
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants>$(DefineConstants);STANDALONEBUILD</DefineConstants>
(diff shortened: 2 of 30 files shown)


src/Directory.BeforeCommon.targets
```diff
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants>$(DefineConstants);STANDALONEBUILD</DefineConstants>
+    <DefineConstants>$(DefineConstants);STANDALONEBUILD;FEATURE_BINARY_SERIALIZATION</DefineConstants>
```
cdmihai: nit: Remove it from the net4* pile? Or is the pile intent to document what's on net4*. In which case there should be a similar pile for .net core?

src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
```diff
@@ -361,13 +361,8 @@ public void TestTranslationWithValueTypesInDictionary()
                 1,
                 Directory.GetCurrentDirectory(),
                 null,
-#if FEATURE_THREAD_CULTURE
```
cdmihai: Maybe it's not worth doing this until we get the green light for removing ns13? If we can't remove ns13, we'll have to put this back. Same goes for the other removed flags.

src/Shared/AssemblyNameExtension.cs
```diff
@@ -958,5 +994,25 @@ internal bool PartialNameCompare(AssemblyNameExtension that, PartialComparisonFl
             }
             return true;
         }
+
+        public void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            if (asAssemblyName != null)
```
cdmihai: Probably nice to have it sync with [https://github.com/Microsoft/msbuild/blob/master/src/Shared/AssemblyUtilities.cs#L62](https://github.com/Microsoft/msbuild/blob/master/src/Shared/AssemblyUtilities.cs#L62)

AndyGerlicher:
https://github.com/dotnet/corefx/issues/25930

src/Shared/AssemblyNameExtension.cs
```diff
@@ -958,5 +1008,30 @@ internal bool PartialNameCompare(AssemblyNameExtension that, PartialComparisonFl
...
+            {
+                info.AddValue("asAssemblyName", true);
```
danmosemsft: It probably doesn't matter a lot but these could just be something like "AN" not "asAssemblyName" which adds up, with 100 assembly names that's about 15KB of extra writes, 30KB if it's written out as UTF16 (not sure).

For similar reasons you might consider "interning" some well known values like the 3 or 4 common PK and PKT almost all our assemblies have -- writing out a marker for those instead would save a lot of bytes if you write out many of these assembly names.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2012,7 +2012,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         expensive to write the newly created cache file.
         -->
     <PropertyGroup>
-      <ResolveAssemblyReferencesStateFile Condition="'$(BuildingProject)'=='true'">$(IntermediateOutputPath)$(MSBuildProjectFile)ResolveAssemblyReference.cache</ResolveAssemblyReferencesStateFile>
+      <!-- Note: Binary serialization not guaranteed to be consistent across full framework and .NET Core, so we need to use a different cache file. -->
```
danmosemsft: .NET Framework and .NET Core do promise to interchange serialized data (if the type is binary serializable in both). What you mean is your special AssemblyName serializatoin scheme is not shared by older MSBuild's.

danmosemsft: This is only necessary if you expect both MSBuild's to build the same projectwith the same obj folder. Neither side crash if they can't deserialize, they just warn and overwrite it.

src/Shared/AssemblyNameExtension.cs
```diff
@@ -121,6 +122,55 @@ internal AssemblyNameExtension(string assemblyName, bool validate) : this()
...
+        /// <remarks>This is required because AssemblyName is not Serializable on .NET Core.</remarks>
+        /// </summary>
+        /// <param name="info"></param>
+        /// <param name="context"></param>
+        private AssemblyNameExtension(SerializationInfo info, StreamingContext context)
+        {
+            var hasAsAssemblyName = info.GetBoolean("asAssemblyName");
+            if (hasAsAssemblyName)
+            {
+                var name = info.GetString("asAssemblyName.Name");
```
danmosemsft: hopefully there are unit tests that would fail if there was a mistake in the serializatoin/deserialization? it's very easy to have a tpyo or suchlike.

AndyGerlicher:

AndyGerlicher:
All the tests are green here, ready for final review.

AndyGerlicher:
One thing to note I removed the split to use a different cache for the Core version. I don't *think* we need that, but I could be wrong. They should cache the same things and the serialization changes to support Core are for both (I updated the format version).

AndyGerlicher:

AndyGerlicher:

danmosemsft:
Yay!

PR #2794, opened 2017-12-10
Allow XmlPeek into XML containing DTD

### LINKED ISSUE #2141: xmlpeek does not allow DTD
We are trying to use the external entity in XML. When we run msbuild, it calls the xmlpeek task, which seems does not allow DTD by default according to the error from xmlpeek task attached below.

error MSB3733: Input file "..." cannot be opened. For security reasons DTD is prohibited in this XML document. To enable DTD processing set the DtdProcessing property on XmlReaderSettings to Parse and pass the settings into XmlReader.Create method.

Is there anyway to change the property of DtdProcessing for the xmlpeek task.

kkm000:
Added property DisallowDtd in case the user wants to prohibit loading XML
with a DTD clause for additional security reasons. DTD is ignored by
default, but does not prevent loading the XML.

Closes #2141

### DIFF
--- a/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
+++ b/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
@@ -1217,8 +1217,8 @@ namespace Microsoft.Build.Tasks
     public partial class XmlPeek : Microsoft.Build.Tasks.TaskExtension
     {
         public XmlPeek() { }
-        public bool DisallowDtd { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public string Namespaces { get { throw null; } set { } }
+        public bool ProhibitDtd { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public string Query { get { throw null; } set { } }
         [Microsoft.Build.Framework.OutputAttribute]
         public Microsoft.Build.Framework.ITaskItem[] Result { get { throw null; } }
--- a/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
+++ b/ref/net46/Microsoft.Build.Tasks.Core/Microsoft.Build.Tasks.Core.cs
@@ -1217,6 +1217,7 @@ namespace Microsoft.Build.Tasks
     public partial class XmlPeek : Microsoft.Build.Tasks.TaskExtension
     {
         public XmlPeek() { }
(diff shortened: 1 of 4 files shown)


AndyGerlicher:
Looks good!
The only thing that stands out to me is the name. `DisallowDtd` set to true sounds somewhat confusing? And it's setting `DtdProcessing.Prohibit`, which is kind of similar but a bit different. Maybe `ProhibitDtdException`? `ProhibitDtdViolations`? Anyone else have any good ideas?

kkm000:
> And it's setting DtdProcessing.Prohibit, which is kind of similar but a bit different

I'd say it's not really different. This is the default setting for `XmlReader`, which is [to throw an exception on DTD being present in the parsed XML](https://docs.microsoft.com/en-us/dotnet/api/system.xml.dtdprocessing.prohibit?view=netframework-4.7.1). This is exactly the issue reported in #2141. As I understand, the [decision was to allow](https://github.com/Microsoft/msbuild/issues/2141#issuecomment-305028579) and ignore DTD in the task, so the *task*'s default is to ignore DTD, unless this parameter is set to true.

So, `ProhibitDtd` would be more consistent? It does not prohibit an exception or a violation, it just prohibits a DTD clause itself, and would match the enum value name the closest, I think.

Or maybe better not to introduce the new property at all, and just always ignore DTD even if present? And  I cannot think of any case when this ProhibitDtd property of the task (revert to the old behavior of throwing on DTD presence) would be of any use. I just went on with an instinct "if there is an option, why not expose it," but this is not always the best design, naturally. So, your call. :)

(As a side note, my understanding is that a security concern is the third enum value, `Parse`, that may cause an HTTP request to download the DTD document. Neither `Prohibit` or `Ignore` are sensitive).

kkm000:
@cdmihai, @AndyGerlicher: I take the PR approval as a decision to go on with the extra property as implemented. Let me quickly push a rename to it anyway. @AndyGerlicher is certainly right that the name would be better if it matches the BCL API enum name, so at the very least those familiar with XmlReader would stand a less chance of being confused. So let me rename it `ProhibitDtd`, not DisallowDtd as I named it.

Just a few minutes please.

AndyGerlicher:
Sounds good, think that name would be less confusing. Thanks!

AndyGerlicher:
Looks good, and thanks for updating!

kkm000:
@AndyGerlicher And thank you and your team, it's really impressive how Microsoft's development process have changed with so many projects becoming open source! But this is only one part of the equation; the fact that you process the submissions immediately and the new continuous updating of the tools is also necessary, and gets all the kudos I can give. A very important point is not only that someone like me can submit a fix, but also how they would get the fix back in their tools very quickly, literally 2-4 weeks time frame. This is not at all comparable with the previous e. g. Visual Studio service releases that came no more often that 1-2 times a year, and then  you'd realize the bug you submitted was confirmed but not fixed just because of a backlog. Now I feel it's my turn to go through the up-for-grabs issues and fix those that others people are really waiting for, whenever I have free cycles (myself, I never used the XmlPeek task, but there indeed was some user's impatience in the issues' comments :) ). All in all, this new development flow is just amazing (and I've been a Microsoft dev tools junkie since the times of VC++ 4.2, so *I* know), thank you guys very much for it!

On your part, please triage and mark "up-for-grabs" as many issues as reasonable for outsiders to stab on. It does attract attention.

PR #2799, opened 2017-12-12
Add globbing information to the profiler

smera:
Adds glob-specific profiling under /prof. Every top-level glob call will have an entry in the profiler report, plus a single entry with an aggregated item representing the overall evaluation time spent in glob calls.

### DIFF
--- a/ref/netstandard1.3/Microsoft.Build.Framework/Microsoft.Build.Framework.cs
+++ b/ref/netstandard1.3/Microsoft.Build.Framework/Microsoft.Build.Framework.cs
@@ -547,18 +547,19 @@ namespace Microsoft.Build.Framework.Profiler
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public partial struct EvaluationLocation
     {
-        public EvaluationLocation(Microsoft.Build.Framework.Profiler.EvaluationPass evaluationPass, string evaluationDescription, string file, System.Nullable<int> line, Microsoft.Build.Framework.IProjectElement element) { throw null;}
-        public EvaluationLocation(Microsoft.Build.Framework.Profiler.EvaluationPass evaluationPass, string evaluationDescription, string file, System.Nullable<int> line, string condition) { throw null;}
-        public EvaluationLocation(Microsoft.Build.Framework.Profiler.EvaluationPass evaluationPass, string evaluationDescription, string file, System.Nullable<int> line, string elementName, string elementOrCondition, bool isElement) { throw null;}
+        public EvaluationLocation(Microsoft.Build.Framework.Profiler.EvaluationPass evaluationPass, string evaluationDescription, string file, System.Nullable<int> line, string elementName, string description, Microsoft.Build.Framework.Profiler.EvaluationLocationKind kind) { throw null;}
+        public string Description { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public string ElementName { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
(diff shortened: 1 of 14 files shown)


AndyGerlicher:
I force pushed your branch for rebase to new master. I also added a commit to update the binary logger format.

ref/netstandard1.3/Microsoft.Build.Framework/Microsoft.Build.Framework.cs
```diff
@@ -547,36 +547,45 @@ namespace Microsoft.Build.Framework.Profiler
...
+    }
+    public enum EvaluationLocationKind : byte
+    {
+        Condition = (byte)1,
+        Glob = (byte)2,
+        Item = (byte)0,
     }
     public enum EvaluationPass : byte
     {
-        InitialProperties = (byte)1,
```
AndyGerlicher: We should mark these explicitly so the value doesn't change when we add something new.

AndyGerlicher:
If this gets in very soon I don't see a need to bump the file version. We need to do another build for CLI insertion I'd like to get this one in so version 5 might be around for a while (in preview form). FYI @KirillOsenkov

AndyGerlicher:
@smera You put WIP... did you need to add anything else? Or can we merge this?

src/Framework/Profiler/EvaluationLocation.cs
```diff
@@ -77,47 +93,56 @@ public struct EvaluationLocation
         public string ElementName { get; }
 
         /// <nodoc/>
-        public string ElementOrCondition { get; }
+        public string Description { get; }
```
cdmihai: What's the difference between Description and EvaluationDescription?

AndyGerlicher:
Merging this to take the change to the binary format. Please open a new PR to address the feedback. If we have to change the format again that's fine but we'd need to rev the version at that point. This is going in an official build for the CLI.

src/Framework/Profiler/EvaluationLocation.cs
```diff
@@ -37,6 +39,19 @@ public enum EvaluationPass : byte
         Targets
     }
 
+    /// <summary>
+    /// The kind of the evaluated location being tracked
+    /// </summary>
+    public enum EvaluationLocationKind : byte
```
cdmihai: There's more kinds than these 3. Like properties, imports, etc. Now, all of the others will get defaulted to item, which is incorrect.

Ways to fix:
- exhaustively describe kind (look at all the usings and generate a kind for each location)
- the above but using broader terms (e.g. if property evaluation has 2 usings, name them both as Property)
- Introduce an Unspecified kind for anything that's not these current 3, and in the future we'll specify on demand. Should not break binary compat.

smera:
The idea was to add IDs and parent IDs to each entry, so in the future the binary logger can display things in a tree-like view.
This change doesn't need to get in now. It does change the structure of the message, so we'd need to bump the binary logger version 

---
In reply to: [351132774](https://github.com/Microsoft/msbuild/pull/2799#issuecomment-351132774) [](ancestors = 351132774)

src/Framework/Profiler/EvaluationLocation.cs
smera: Before this change we had just item and condition. I added glob, and that's why the enum.
I don't think we need more kinds at this point, so maybe this is just renaming 'item' to something else. Maybe 'element'?

---
In reply to: [156447130](https://github.com/Microsoft/msbuild/pull/2799#discussion_r156447130) [](ancestors = 156447130)

src/Framework/Profiler/EvaluationLocation.cs
smera: Renamed EvaluationDescription to EvaluationPassDescription, and Description to ElementDescription

---
In reply to: [156443221](https://github.com/Microsoft/msbuild/pull/2799#discussion_r156443221) [](ancestors = 156443221)

PR #2805, opened 2017-12-13
Fix tests in update-toolset branch

dsplaisted:


src/Utilities.UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -454,7 +454,7 @@ public void HighestVersionOfTargetFrameworkIdentifierRootMultipleVersions()
         [Fact]
         public void VerifyinternalGetPathToDotNetFrameworkSdkFile40()
         {
-            string tempDirectory = Path.Combine(Path.GetTempPath(), "VGPTDNFSFN40");
+            string tempDirectory = Path.Combine(Path.GetTempPath(), "VGPTDNFSF40");
```
cdmihai: Did this hit maxpath? :)

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -19,22 +19,6 @@ internal class BuildEnvironmentHelper
         // MSBuildConstants.CurrentToolsVersion
         private const string CurrentToolsVersion = "15.0";
 
-        // Duplicated in InternalErrorException.cs. Update both when changing.
```
cdmihai: What about the duplicated part? Is that still needed?

PR #2806, opened 2017-12-13
WIP Libraries can reference Microsoft.Build.CommandLine.MSBuildApp

cdmihai:
Update the nuspec such that the executables are treated by nuget as referenceable assemblies.

I tested that building a console app which depends on Microsoft.Build.Runtime:
- targeting .net core this works: `Microsoft.Build.CommandLine.MSBuildApp.Main(new []{@"path to some project"});`
- targeting net46 and invoking the MSBuild.exe under bin on some other project works. (this more of a regression test)


cdmihai:
@livarcocc

PR #2814, opened 2017-12-15
Breaking Change in FileMatcher: ab*ba pattern matches aba by mistake

zhilich:
Pull request is created based on issue #2815
Performance optimization #2392 led to a breaking change in FileMatcher logic. Let's examine the following example with a single web.config file in working directory:
```
<MyFiles Include="**\*" Exclude="web.*.config">
</MyFiles>
```

Before the change @(MyFiles) included web.config because exclude pattern didn't match it. However after the change web.config was excluded from the item group. This is because implementation of FileMatcher.IsMatch() doesn't consider a situation when tail after wildcard (*) is longer than the remaining part of the input during "tail check". This is a pretty common scenario when several file extensions are used.

To be more specific FileMatcher.IsMatch("aba", "ab*ba") should return false instead of true.

I suggest a simple fix to cancel tail check when pattern tail length exceeds input tail length.

### DIFF
--- a/src/Shared/UnitTests/FileMatcher_Tests.cs
+++ b/src/Shared/UnitTests/FileMatcher_Tests.cs
@@ -36,7 +36,8 @@ namespace Microsoft.Build.UnitTests
                 "file1.txt",
                 "file1.txtother",
                 "fie1.txt",
-                "fire1.txt"
+                "fire1.txt",
+                "file.bak.txt"
             };
 
             try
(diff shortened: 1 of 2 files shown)


src/Shared/UnitTests/FileMatcher_Tests.cs
```diff
@@ -57,7 +57,8 @@ public void GetFilesPatternMatching()
                     { "???.*", 1},
                     { "????.*", 3},
                     { "*.???", 4},
-                    { "f??e1.txt", 2}
+                    { "f??e1.txt", 2},
+                    { "file.*.txt", 0 }
```
cdmihai: Add a file that also matches.

PR #2816, opened 2017-12-15
Implement a /restoreProperty command-line argument for specifying global properties

jeffkl:
If a user does not specify any /restoreProperty arguments, the global properties specified by /property arguments will be used

If a user specifies any /restoreProperty arguments, only those properties will be set during restore

Usage looks like this:

```
  /restoreProperty:<n>=<v>
                     Set or override these project-level properties only
                     during restore and do not use properties specified
                     with the /property argument. <n> is the property
                     name, and <v> is the property value. Use a
                     semicolon or a comma to separate multiple properties,
                     or specify each property separately.
                     (Short form: /rp)
                     Example:
                       /restoreProperty:IsRestore=true;MyProperty=value
```

### DIFF
--- a/src/MSBuild/Resources/Strings.resx
+++ b/src/MSBuild/Resources/Strings.resx
@@ -687,6 +687,23 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_32_RestorePropertySwitch" UESanitized="false" Visibility="Public">
+    <value>  /restoreProperty:&lt;n&gt;=&lt;v&gt;
+                     Set or override these project-level properties only
+                     during restore and do not use properties specified
+                     with the /property argument. &lt;n&gt; is the property
+                     name, and &lt;v&gt; is the property value. Use a
+                     semicolon or a comma to separate multiple properties,
+                     or specify each property separately.
+                     (Short form: /rp)
+                     Example:
+                       /restoreProperty:IsRestore=true;MyProperty=value
+    </value>
+    <comment>
+      LOCALIZATION: "/restoreProperty" and "/rp" should not be localized.
+      LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
+    </comment>
+  </data>
   <data name="InvalidConfigurationFile" Visibility="Public">
     <value>MSBUILD : Configuration error MSB1043: The application could not start. {0}</value>
     <comment>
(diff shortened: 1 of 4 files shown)


jeffkl:
> what problem does this solve?

Some command-line builds do a restore first and then a build second.  When I introduced `/restore`, a few users pointed out that they run restores like this:

```
msbuild /t:Restore /p:Property1=Value
msbuild /t:Build /p:Property1=Something else
```
Since the invocations are separate, it lets them do this.  With `/restore`, the same global properties are used for both "builds".  @AndyGerlicher and I discussed that we might as well give people the option to use different global properties.

> Does this really deserve a short form? Do we expect people to type it on the command line?

I think so, but I only get a single vote

> I think I'd expect /property to always be set, with a /restoreProperty possibly overriding it during the /restore build. Is there a case where that doesn't work right?

I could merge the two, but then users couldn't use completely different properties even if they wanted to.  With the current implementation, they are free to do whatever they want but they have to set the property twice if they want restore and build to share only some of the properties.  I didn't want to have more command-line arguments to control how the properties were set (like `/mergeRestoreProperties:[true|false]`)

rainersigwald:
Just talked to @AndyGerlicher about this offline, and the motivating case for this (specify a single TFM for build but auto-restoring with no TF specified because that would break NuGet) makes a pretty good case for the-properties-should-be-entirely-distinct.

jeffkl:
Should I get rid of the short form?

src/MSBuild/XMake.cs
```diff
@@ -2029,6 +2034,9 @@ bool recursing
                     // figure out which properties have been set on the command line
                     globalProperties = ProcessPropertySwitch(commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.Property]);
 
+                    // figure out restore properties have been set on the command line
```
rainersigwald: nit: `figure out which restore-only properties...`

src/MSBuild/Resources/Strings.resx
```diff
@@ -687,6 +687,23 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_32_RestorePropertySwitch" UESanitized="false" Visibility="Public">
```
rainersigwald: There's a `HelpMessage_32_` way later in the file (not sure why not right here . . .). Make this 33?

jeffkl: Good catch!

PR #2819, opened 2017-12-20
Add new MSBuildVersion property

### LINKED ISSUE #2350: Detecting and failing on old MSBuild versions
The [dotnet/roslyn](http://github.com/dotnet/roslyn) repo requires the MSBuild from at least VS2017 update 2 to compile. This the earliest version that has a number of SDK bug fixes that we depend on. When built with earlier versions of MSBuild we end up with some fairly cryptic error messages like

```
RuntimeIdentifier must be set for .NETFramework executables
```

This means customers with RTM builds of our product come up to our repos, build and get crpytic error messages. Feel like that makes our build process look broken / flaky. Trying to come up with a better experience here. 

The ideal is that when you run msbuild on our solution you end up with one or more messages that say:

```
This project requires at least MSBuild from VS2017 Update 2. 
```

That would be actionable by customers and save a lot of headache on our part. Think this would be really easy to do in a BeforeBuild Target / Task. But I can't figure out the best way to map an installation of MSBuild to say VS2017 update 2. Looking on my machine I can see the file version is 15.1.1012.6693. So I could just say if the version is greater than or equal to that then you're at least update 2. Feels a bit hacky though. Is there a better way of doing a version comparison on MSBuild binaries?

jeffkl:
`MSBuildVersion` will be major.minor.build version of MSBuild according to the `[assembly:AssemblyFileVersion]` attribute (currently `15.6.X`).  This allows build owners to implement errors if a user is using an older version of MSBuild that might not be able to build their source code.

I've implemented an `MSBuildAssemblyFileVersion` class which gets the assembly attribute with a Lazy singleton.  This should create very little overhead and not add any perf cost.

Closes #2350

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -1486,7 +1486,7 @@ namespace Microsoft.Build.Evaluation
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.buildNodeCount, _maxNodeCount.ToString(CultureInfo.CurrentCulture)));
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.programFiles32, FrameworkLocationHelper.programFiles32));
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.assemblyVersion, Constants.AssemblyVersion));
-            builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.version, MSBuildAssemblyFileVersion.Instance.MajorMinor));
+            builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.version, MSBuildAssemblyFileVersion.Instance.MajorMinorBuild));
 
             // Fake OS env variables when not on Windows
             if (!NativeMethodsShared.IsWindows)
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -1477,7 +1477,7 @@ namespace Microsoft.Build.Evaluation
         {
             string startupDirectory = BuildParameters.StartupDirectory;
 
-            List<P> builtInProperties = new List<P>(12);
+            List<P> builtInProperties = new List<P>(19);
 
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.toolsVersion, _data.Toolset.ToolsVersion));
             builtInProperties.Add(SetBuiltInProperty(ReservedPropertyNames.toolsPath, _data.Toolset.ToolsPath));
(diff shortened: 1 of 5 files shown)


jeffkl:
FYI @jaredpar

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1477,7 +1477,7 @@ private ICollection<P> AddBuiltInProperties()
         {
             string startupDirectory = BuildParameters.StartupDirectory;
 
-            List<P> builtInProperties = new List<P>(12);
+            List<P> builtInProperties = new List<P>(19);
```
rainersigwald: Nice catch.

src/Build/Resources/MSBuildAssemblyFileVersion.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Gets the <see cref="AssemblyFileVersionAttribute"/> of Microsoft.Build.dll.
+    /// </summary>
+    internal sealed class MSBuildAssemblyFileVersion
+    {
+        private static readonly Lazy<MSBuildAssemblyFileVersion> MSBuildAssemblyFileVersionLazy = new Lazy<MSBuildAssemblyFileVersion>(GetMSBuildAssemblyFileVersion, isThreadSafe: true);
+
+        private MSBuildAssemblyFileVersion(string majorMinor)
+        {
+            MajorMinor = majorMinor;;
```
rainersigwald: nit: double semicolon

src/Build/Resources/MSBuildAssemblyFileVersion.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            Version version;
+
+            if (String.IsNullOrEmpty(versionString) || !Version.TryParse(versionString, out version))
+            {
+                // Fall back to the constant AssemblyVersion
+                return new MSBuildAssemblyFileVersion(Constants.AssemblyVersion);
+            }
+
+            return new MSBuildAssemblyFileVersion($"{version.Major}.{version.Minor}");
```
rainersigwald: This seems reasonable but we might want to go three-part here, so that a build extension author can say "you n need to install 15.23.1 because it has an engine fix we depend on". We probably don't want to go all the way to 4-part, since the NB.GV-generated 4th part is based on the commit hash and not comparable.

src/Build/Resources/MSBuildAssemblyFileVersion.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        /// <summary>
+        /// Gets the assembly file version in the format major.minor.
+        /// </summary>
+        public string MajorMinor { get; set; }
+
+        private static MSBuildAssemblyFileVersion GetMSBuildAssemblyFileVersion()
+        {
+            string versionString = typeof(MSBuildAssemblyFileVersion).GetTypeInfo()?.Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version;
```
rainersigwald: super nit: this line is long. insert some newlines?

src/Build/Resources/MSBuildAssemblyFileVersion.cs
jeffkl: At the moment, we're not using NBGV for .NET Core so we only set it as `15.6`.  After we've moved to the new SDK, we'll be using NBGV for .NET Core and I could change this to 3 part.  But would it be better to only be 2 part for this:

```xml
<Error
    Condition=" '$(MSBuildVersion)' &lt; '15.7' "
    Text="You must build this code base with MSBuild 15.7 or above"
 />
```

Since we don't have a separate property for major, minor, and build, string comparison is the only choice.  I'm worried a 3 part version would make this harder, what do you think?

rainersigwald: Won't that go into version-comparison mode, and be able to handle up to four-part versions?

jeffkl: Rainer is right, we `Version.TryParse()` all strings in conditions so there is no harm in the version being major.minor.build.

At the moment. .NET core versions will `0` for the build number until we move to the new SDK.

jeffkl:
Per @AndyGerlicher, I'll update the logic to just use the `ThisAssembly.AssemblyFileVersion` constant after we've moved to .NET Core 2.0

PR #2820, opened 2017-12-22
[mono][osx] Fix test hangs

radical:
Some tests caused the test run to hang under mono/osx. This is because
of an `InternalErrorException` being thrown and msbuild trying to launch
and connect to a debugger. We do try to avoid the debugger code path
(`ConsiderDebuggerLaunch`), but it depends on
`BuildEnvironmentHelper.RunningTests` which incorrectly gets set to
`false`.

When running a test runner under mono, the process name is not helpful
to determine if we are running tests as that would be `mono`. Getting
command line arguments, OTOH, do help with this.

So, we use GetCommandLineArgs, also, to check for test runners. We bring
back the old `GetProcessFromCommandLine()` method. And also, add this as
a parameter to the `ResetInstance_ForUnitTestsOnly(..)` method.

Example test that triggered this:
    `Microsoft.Build.UnitTests.ErrorUtilities_Tests.VerifyThrowFalse`

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -294,12 +294,22 @@ namespace Microsoft.Build.Shared
             return FileUtilities.CombinePaths(visualStudioRoot, "MSBuild", CurrentToolsVersion, "Bin", "MSBuild.exe");
         }
 
+        private static string GetProcessFromCommandLine()
+        {
+#if FEATURE_GET_COMMANDLINE
+            return Environment.GetCommandLineArgs()[0];
+#else
+            return null;
+#endif
+        }
+
         private static bool CheckIfRunningTests()
         {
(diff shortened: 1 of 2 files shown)


src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -294,12 +294,22 @@ private static string GetMSBuildExeFromVsRoot(string visualStudioRoot)
             return FileUtilities.CombinePaths(visualStudioRoot, "MSBuild", CurrentToolsVersion, "Bin", "MSBuild.exe");
         }
 
+        private static string GetProcessFromCommandLine()
+        {
+#if FEATURE_GET_COMMANDLINE
+            return Environment.GetCommandLineArgs()[0];
```
rainersigwald: Does the Mono runtime strip the `mono` prefix from this? .NET Core does not, IIRC.

PR #2822, opened 2017-12-22
RI vs15.5

rainersigwald:
The only meaningful change was a port of a change that's already in master, but this will help with git bookkeeping--the prior release should always be an ancestor of a subsequent release.


rainersigwald:
NOTE NOTE NOTE: I was lazy and did this through the web UI, and to resolve conflicts I **moved the `vs15.5` branch pointer**. If we need to service 15.5, we need to create a new branch.

AndyGerlicher:
Shouldn't we just reset vs15.5 to e4c819961e3154152cd0c47681527784632aa55f and force push?

PR #2824, opened 2017-12-25
Fix #828: Upgrade "Mismatched assemblies" MSB3274 to error

### LINKED ISSUE #828: Upgrade "Mismatched assemblies" MSB3274 to error.
Originally asked for here:
https://github.com/dotnet/roslyn/issues/5001

This warning should be upgraded to an error and fail the build. The warning is logged [here](https://github.com/Microsoft/msbuild/blob/f90dcff1db3b10c3b136bed1a23ce710f7c579cd/src/XMakeTasks/AssemblyDependency/ReferenceTable.cs#L2927).

eriawan:
This is to fix #828, update MSB3274 warning to error.

cc @AndyGerlicher , @rainersigwald

### DIFF
--- a/src/Tasks/AssemblyDependency/ReferenceTable.cs
+++ b/src/Tasks/AssemblyDependency/ReferenceTable.cs
@@ -3014,7 +3014,7 @@ namespace Microsoft.Build.Tasks
         {
             if (displayPrimaryReferenceMessage)
             {
-                _log.LogWarningWithCodeFromResources("ResolveAssemblyReference.PrimaryReferenceOutsideOfFrameworkUsingAttribute", reference.PrimarySourceItem.ItemSpec /* primary item spec*/, reference.FrameworkNameAttribute /*Version of dependent assemby*/, targetedFramework);
+                _log.LogErrorWithCodeFromResources("ResolveAssemblyReference.PrimaryReferenceOutsideOfFrameworkUsingAttribute", reference.PrimarySourceItem.ItemSpec /* primary item spec*/, reference.FrameworkNameAttribute /*Version of dependent assemby*/, targetedFramework);
             }
             else
             {


eriawan:
@AndyGerlicher 

Any feedback about this? Please confirm.

cc @rainersigwald

cdmihai:
@dotnet-bot test this please

src/Tasks/AssemblyDependency/ReferenceTable.cs
```diff
@@ -3014,7 +3014,7 @@ internal void LogHigherVersionUnresolveDueToAttribute(bool displayPrimaryReferen
         {
             if (displayPrimaryReferenceMessage)
             {
-                _log.LogWarningWithCodeFromResources("ResolveAssemblyReference.PrimaryReferenceOutsideOfFrameworkUsingAttribute", reference.PrimarySourceItem.ItemSpec /* primary item spec*/, reference.FrameworkNameAttribute /*Version of dependent assemby*/, targetedFramework);
+                _log.LogErrorWithCodeFromResources("ResolveAssemblyReference.PrimaryReferenceOutsideOfFrameworkUsingAttribute", reference.PrimarySourceItem.ItemSpec /* primary item spec*/, reference.FrameworkNameAttribute /*Version of dependent assemby*/, targetedFramework);
```
cdmihai: Can you please add a unit for this, if there's no failing ones?

AndyGerlicher:
@danmosemsft do you or someone from your team have any input on this? It seems like the right thing to do, but it could be a breaking change. Can you envision a scenario where this would not break at runtime and thus we should leave as a warning? We've made a similar "breaking" change when we used to silently upgrade you to a newer framework if you didn't have the target installed which broke horribly at runtime (Roslyn would use `Array.Empty` when it didn't exist in the target framework). So there is precedence, but I'm leaning towards not taking this in an update since it could break someone.

danmosemsft:
I don't but perhaps @weshaggard or @eerhardt do.

weshaggard:
While I agree it may not be correct I can certainly imagine scenarios that today just work but after this error will break. 

For better or worse people mix-and-match framework assemblies in projects. It is certainly possible to write a library targeting a higher framework, lets say .NET 4.7, that only uses APIs that exist in .NET 4.6 and then consume that library in a .NET 4.6 application where everything just works. They would get this warning but nothing bad would happen at runtime because the library just happened to not use any APIs that don't exist in .NET 4.6

AndyGerlicher:
@cdmihai setup a repro of this, and it looks like if the warning hits the reference gets excluded. So you can't actually use any of the types in there anyway since they're not passed to the compiler if this hits. So it seems like this should have no runtime affect on anyway, it would just error in the build instead of warn.

So maybe it could break someone relying on this for build ordering? :)

eerhardt:
Would it be a lot of work to add an "escape hatch" in the off chance that this DOES break someone?  That way they can set a flag saying "give me back the old behavior, it was working just fine".

> if the warning hits the reference gets excluded

Does the reference still get copied to the output directory?

src/Tasks/AssemblyDependency/ReferenceTable.cs
eriawan: @cdmihai thanks for the feedback!

OK, I will add one unit test for this.

cdmihai:
@eerhardt 

> Does the reference still get copied to the output directory?

p2p is not present in RAR output, which makes CSC fail with not finding the reference (when there's a compile time dependency on the p2p), p2p not copied to output (probably because p2p does not make it out of RAR).

cdmihai:

Testing flakiness for https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_Full_prtest/715/testReport/junit/Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests/VerifyTargetFrameworkAttribute/HigherVersionDirect/

cdmihai:
@eriawan 

Apparently there is a test catching this: https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_Full_prtest/723/testReport/junit/Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests/VerifyTargetFrameworkAttribute/HigherVersionDirect/
You just need to change warnings to errors in it.

Unfortunately there is another test there that looks like it should fail but doesn't (asserts that indirect reference to higher TFM should warn): https://github.com/Microsoft/msbuild/blob/4161a7f410771aab0aa7d79ff58032e53a43a850/src/Tasks.UnitTests/AssemblyDependency/VerifyTargetFrameworkAttribute.cs#L219
Can you please look into it? 

@eerhardt 

> Would it be a lot of work to add an "escape hatch" in the off chance that this DOES break someone? That way they can set a flag saying "give me back the old behavior, it was working just fine".

Apparently there's a RAR parameter that turns off TFM matching: https://github.com/Microsoft/msbuild/blob/4161a7f410771aab0aa7d79ff58032e53a43a850/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2088 Is that good enough?

pkulikov:
@weshaggard I have remark on that one
> It is certainly possible to write a library targeting a higher framework, lets say .NET 4.7, that only uses APIs that exist in .NET 4.6 and then consume that library in a .NET 4.6 application where everything just works. They would get this warning but nothing bad would happen at runtime because the library just happened to not use any APIs that don't exist in .NET 4.6

That warning sometimes results in failed build and error [CS0246](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0246) in the **Output** Visual Studio window, but not in the **Error List** that shows only the warning.

I've checked the following solution (Visual Studio 2017, version 15.5.4):
- .NET Framework Console app that targets 4.6.1
- .NET Framework Class Library that targets 4.7.1 with some empty class. So, I assume, no .NET APIs used at all.

Two scenarios:
- The console app project references the class library project, **BUT** does not use the class defined at the library. Everything builds and runs, the class library dll is not copied to the console app output folder.
- The console app project references the class library project, **AND** uses the class defined at the library. Build fails. However, as I mentioned above, **Error List** shows no errors, only the warning, while **Output** window reports CS0246 error.

@AndyGerlicher the reference to the library was not excluded.

rainersigwald:
Let's try to get this in for dev16. I'm going to close/reopen to pick up new tests.

danmosemsft:
As a general rule if you aren't certain that this won't have bad results for some customers then I would expose a property just in case, to change it back to a warning. We had precedent for something similar (it was called `$(GenerateResourceNeverLockTypeAssemblies)` and the dev was confident it would never ever be needed, but we included it just in case, and it turned out that many people's builds broke and they were saved by the flag from being dead in the water)

rainersigwald:
Team triage: Thank you for the contribution! Since we haven't been able to figure out how breaking this would be, we're going to close this and reconsider it for the next major version, instead of possibly introducing a break in a patch release.

eriawan:
@rainersigwald 

I agree. Do you think this merge would fit for 16.2 timeframe?

PR #2830, opened 2017-12-28
Need to check assembly version when doing IsFrameworkAssembly check

Jinhuafei:
Clickonce tooling relies on IsFrameworkAssembly method in Microsoft.Build.Tasks.Core.dll to determine if the assembly is part of .net framework redist. If so, Clickonce tooling won't publish that assembly. However, IsFraemworkAssembly method doesn't check the assembly version which causes assembly loading failure at runtime.

**Example**
If a project references System.Runtime nupkg 4.3, ClickOnce publish doesn’t include System.Runtime.dll into the publish file list.

### DIFF
--- a/src/Tasks/RedistList.cs
+++ b/src/Tasks/RedistList.cs
@@ -450,15 +450,7 @@ namespace Microsoft.Build.Tasks
 
         private AssemblyNameExtension GetAssemblyNameExtension(string assemblyName)
         {
-            AssemblyNameExtension assemblyNameExtension = null;
-
-            if(!_assemblyNameToAssemblyNameExtension.TryGetValue(assemblyName, out assemblyNameExtension))
-            {
-                assemblyNameExtension = new AssemblyNameExtension(assemblyName);
-                _assemblyNameToAssemblyNameExtension.TryAdd(assemblyName, assemblyNameExtension);
-            }
-
-            return assemblyNameExtension;
+            return _assemblyNameToAssemblyNameExtension.GetOrAdd(assemblyName, (key) => new AssemblyNameExtension(key));
         }
 
         /// <summary>
--- a/src/Tasks/RedistList.cs
+++ b/src/Tasks/RedistList.cs
@@ -15,7 +15,6 @@ using System.Collections.Concurrent;
 using System.Collections.ObjectModel;
 using System.Linq;
 
-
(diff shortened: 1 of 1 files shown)


src/Tasks/RedistList.cs
```diff
@@ -159,7 +159,10 @@ internal int Count
         public bool IsFrameworkAssembly(string assemblyName)
         {
             AssemblyEntry entry = GetUnifiedAssemblyEntry(assemblyName);
-            if (entry != null && !String.IsNullOrEmpty(entry.RedistName))
+            AssemblyNameExtension assembly = new AssemblyNameExtension(assemblyName);
```
AndyGerlicher: I'm a little worried about perf here. AssemblyName is pretty slow, so creating a new `AssemblyNameExtension` every time could be an issue. 

Looking at the details of `GetUnifiedAssemblyEntry`, it always creates an `AssemblyNameExtension` anyway. What about return that as an overload with an out variable? Reusing it would eliminate extra overhead.

Jinhuafei: Most of the cases, we won't hit the code in GetUnifiedAssemblyEntry that creates AssemblyNameExtension. If we really want to reuse it, we need to use a dictionary to hold the instances.

Jinhuafei: Also, do we really call IsFrameworkAssembly a lot, just like calling GetUnifiedAssemblyEntry?

src/Tasks/RedistList.cs
```diff
@@ -437,6 +448,19 @@ private AssemblyEntry GetUnifiedAssemblyEntry(string assemblyName)
             return unifiedEntry;
         }
 
+        private AssemblyNameExtension GetAssemblyNameExtension(string assemblyName)
+        {
+            AssemblyNameExtension assemblyNameExtension = null;
+
+            if(!_assemblyNameToAssemblyNameExtension.TryGetValue(assemblyName, out assemblyNameExtension))
```
cdmihai: You can use `ConcurrentDictionary.GetOrAdd` to make the check and instantiation atomic.

AndyGerlicher:
I also put a `Debugger.Launch` and this code doesn't hit building our repo or tests. So I'm good with perf concerns. Please switch to GetOrAdd though and we'll accept. Thanks!

PR #2840, opened 2018-01-04
👷 Ignore temporary ProjectImports zip file

rainersigwald:
This file is created during cibuilds and may persist if MSBuild is
aborted abnormally. It's a build output like other logs and shouldn't be
checked in.


rainersigwald:
I noticed this file popping up in VS's Git pane when running CIBuild on the command line.

PR #2848, opened 2018-01-08
[WIP] Native globbing via a file system abstraction

smera:
Introduce a file system abstraction and propagate it to FileMatcher, responsible for globbing.
Use a windows-specific native enumeration when available.

TODO: validate perf, put this under a flag and propagate the file system abstraction to the rest of the evaluator


rainersigwald:
Thanks. I'm going to close this since we haven't been able to dedicate the time to figure out if it's the right path forward. I'm adding a note to #3586 that this is something that could help there.

Issue #2803, opened 2017-12-13
NuGet-based SDK resolver design spec


jeffkl:
# Overview

At the moment, there are two SDK resolvers, both of which assume that the SDKs are on disk.  It would be very useful to have SDKs resolved as NuGet packages as well which would allow third party SDKs to be acquired.

Related issues:
* Adding a custom SDK resolver #2278 
* Proposal: implement a SdkResolver for NuGet packages [#5220](https://github.com/NuGet/Home/issues/5220)

## User Experience
Users would have to opt into the new behavior by specifying a version of the SDK to use.  The resolver would never automatically get the latest because this would introduce very non-deterministic builds.

Users could control the version on a per-project basis:
```xml
<Project Sdk="My.Custom.Sdk/2.3.4">
  ...
</Project>
```
In this example, a NuGet package named `My.Custom.Sdk` version `2.3.4` would be retrieved and imported.

It would scale much better if users could specify a version per SDK for the whole repository.  They could do this in their `global.json`:

```json
{
  "sdk" : {
    "version:" "1.0.0"
  },
  "msbuild-sdks": {
    "My.Custom.Sdk" : "2.3.4"
  }
}
```

The `sdk` object controls which version of the .NET CLI to use and the `msbuild-sdks` object controls which versions of the MSBuild SDKs to use.  If the resolver detected that multiple versions of the same SDK are requested for the same build episode, a warning would be logged and the first version specified would be used.

```
MSB4528: warning: A version of the SDK 'My.Custom.Sdk' was already specified in 'D:\Foo\global.json'.  The version specified at 'C:\Foo\bar.csproj (1.7)' was ignored.
```

In Visual Studio, users would see projects in a Loading state while SDKs are resolved.  Any errors that occur during acquisition would be reported in the Output window just like other package restore or build errors.  Projects would fail to load if an SDK could not be resolved.  Users could fix the issue and reload the project in Visual Studio.

![image](https://user-images.githubusercontent.com/17556515/33953328-7c672580-dfe9-11e7-81e9-a06e6f190f96.png)

Since the SDKs would be NuGet packages, the standard caching mechanisms and configurations would be in place.  Users would control which feeds to use and other NuGet settings in their `NuGet.config`.

## Implementation

- [x] 1. The current SdkResolvers infrastructure within MSBuild would need to be improved to support caching and parallelization.  Currently, the same SDK is resolved multiple times per build episode which can slow down evaluations.  #2847 
[...]

PR #2849, opened 2018-01-09
Allow SDK resolvers to preserve state

jeffkl:
SDK resolvers can get and set the `State` property on the `SdkResolverContext` they receive.  This state is preserved for the same resolver during the same build submission.  When there is no build submission, the state is not preserved.  Resolvers will need to take account for these considerations.

This will allow the NuGet-based SDK resolver to only look for and parse `global.json` once per build.

Address item 2 in #2803

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildManager.cs
+++ b/src/Build/BackEnd/BuildManager/BuildManager.cs
@@ -1664,9 +1664,6 @@ namespace Microsoft.Build.Execution
 
                     // Clear all cached SDKs for the submission
                     SdkResolverService.ClearCache(submission.SubmissionId);
-
-                    // Clear SDK resolver cache for the submission
-                    BackEnd.SdkResolution.SdkResolverService.Instance.ClearCache(submission.SubmissionId);
                 }
 
                 if (_buildSubmissions.Count == 0)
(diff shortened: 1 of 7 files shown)


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1664,6 +1664,9 @@ private void CheckSubmissionCompletenessAndRemove(BuildSubmission submission)
 
                     // Clear all cached SDKs for the submission
                     SdkResolverService.ClearCache(submission.SubmissionId);
+
+                    // Clear SDK resolver cache for the submission
```
cdmihai: This line seems very similar to the previous one. A code comment on why this is would be nice.

jeffkl: Good point, and after I took a second look at it, I decided this call should be moved.  I've also improved the comment

jeffkl:
@nguerrera, this PR allows resolvers to maintain state for a particular build submission.  I'll be using it in my NuGet-based SDK resolver so that `global.json` is only searched for and deserialized once per build.  Do you have any feedback on this implementation?

nguerrera:
In yesterday's sync, @AndyGerlicher said VS will cache all SDK resolver results per build. But (without digging in to code) this suggests that I'll be called multiple times per build and it is up to me to use a state object for caching. Which is it? Basically, will we get a perf win without doing anything or do we need to go implement the caching using this state feature?

jeffkl:
For each build and unique SDK name, all resolvers are called in order.  Once an SDK has been resolved, no resolvers are called again for that SDK.  If another SDK comes along that is not in the cache, then all SDK resolvers are called again.

The point of this "state" object is so that a resolver can discover something about the environment (like a path to global.json) only once.  But it will still only be called one for each unique SDK.  So if your entire project tree only references `Microsoft.NET.Sdk`, the resolver is only called once anyway.

Issue #2278, opened 2017-07-11
Adding a custom SDK resolver


natemcmaster:
**Problem**
The default SDK resolver loading only looks for resolvers in $(MSBuildToolsDirectory32)/SdkResolvers ([source][source-link]).

[source-link]: https://github.com/Microsoft/msbuild/blob/7b9ebbae516f7a9b19e6b4de195a648073bf3ab9/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs#L24

Most of the time, this path is going to be `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\SdkResolvers`. Writing files here requires admin permissions.

_My scenario:_ I'm looking for a way to share a common set of targets and props across ASP.NET Core's 500+ csproj files, which are scattered among repos. We've tried PackageReference, but continually run into limitations. I've written and tested a custom SDK resolvers. It seems promising, but there is no easy way to make this available for MSBuild for VS and dotnet.exe without getting elevated permissions to write into the VS folder and/or `C:\Program Files\dotnet` folder.

**Request**
Provide a way to register custom SDK resolvers without needing to write into the MSBuild tools folder.

**Ideas**

  - Similar to `<UsingTask>`, add  `<UsingSdkResolver>` directive that takes an assembly path.
  - Add another search path, such as $(MSBuildUserExtensionsPath)/$(MSBuildRuntimeType)/SdkResolvers

onovotny:
I would like to use this to make it easier to add support for other project types, like Rust. Ideally, a directory level configuration could point to a path containing an SDK. It'd work like directory.build.props where parent dirs are searched. I'd put that along the solution file.

Also, there should be a way (NuGet?) to distribute sdks. The aforementioned config file could point to a NuGet package with an SDKs folder in it, then it'd be automatically downloaded and added to the search path.

Then something like Visual Rust can package its targets as an SDK and distribute it on NuGet.

dasMulli:
@onovotny fyi there is an issue for a NuGet resolver at https://github.com/NuGet/Home/issues/5220

onovotny:
@dasMulli Thanks for the pointer!

My ask when designing this feature is that it's not tied to .NET. That is, it should ideally support any project type that uses CPS to enable it to have its own SDK. The example I'm thinking of right now is Rust, but it can apply to any other language/platform.

natemcmaster:
[...]

PR #2850, opened 2018-01-10
Implement a NuGet-based SDK resolver

jeffkl:
An SDK resolver that activates only when a user has specified a version.  This resolver is ordered to come before the DotNetSdkResolver but if a user does not specify a version (the default), then the NuGet SDK resolver is skipped.

The NuGet SDK resolver can be disabled with an environment variable just in case its causing issues or a user wants to override its functionality.

NuGet assemblies are loaded at runtime from their installed location.  They either come come [VisualStudioInstallRoot\Common7\IDE\CommonExtensions\Microsoft\NuGet or next to MSBuild.exe.  There is also an environment variable that can be set to specify a folder to use instead.

I've also placed references to NuGet into private subclasses so that assemblies are not loaded unless they are needed.  This means that if you have no version specified and no `global.json` with versions, `Newtonsoft.Json` and `NuGet.*.dll` are not loaded.  You only pay JIT penalty if you're using the functionality and hopefully we'll make up that time later in the build through caching.

Addresses item 6 in #2803 
Related to #2278


jeffkl:
FYI @nguerrera @emgarten @mishra14 @zhili1208

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// Represents a NuGet-based SDK resolver.  It is very important that this class does not reference any NuGet assemblies
+    /// directly as an optimization to avoid loading them unless they are needed.  The current implementation only loads
+    /// Newtonsoft.Json if a global.json is found and it contains the msbuild-sdks section and a few NuGet assemblies to parse
+    /// a version.  The remaining NuGet assemblies are then loaded to do a restore.
+    /// </summary>
+    internal sealed class NuGetSdkResolver : NuGetSdkResolverBase
+    {
+        public override string Name => nameof(NuGetSdkResolver);
+
+        public override int Priority => 2500;
```
AndyGerlicher: To be explicit here for everyone reviewing, the `DefaultSdkResolver` is 10,000 and `DotNetMSBuildSdkResolver` is 5,000. So at 2,500 `NuGetSdkResolver` will run before any others (that we ship).

src/Build/Microsoft.Build.csproj
```diff
@@ -810,6 +819,7 @@
     <Compile Include="Evaluation\LazyItemEvaluator.EvaluatorData.cs" />
     <None Include="project.json" />
   </ItemGroup>
+  <ItemGroup />
```
rainersigwald: nit: remove

src/Build/project.json
```diff
@@ -3,6 +3,8 @@
     "net46": {
       "dependencies": {
         "Microsoft.VisualStudio.Setup.Configuration.Interop": "1.14.114",
+        "NuGet.Commands": "4.5.0",
+        "NuGet.Protocol": "4.5.0",
```
rainersigwald: Do we need to get the NuGet team to add us to their automatically-drop-new-NuGet-assemblies PRs?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolverBase.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    /// <summary>
+    /// Acts as a base class for the NuGet-based SDK resolver and handles assembly resolution to dynamically locate NuGet assemblies.
+    /// </summary>
+    internal abstract class NuGetSdkResolverBase : SdkResolverBase
+    {
+        /// <summary>
+        /// The sub-folder under the Visual Studio installation where the NuGet assemblies are located.
+        /// </summary>
+        public const string PathToNuGetUnderVisualStudioRoot = @"Common7\IDE\CommonExtensions\Microsoft\NuGet";
```
rainersigwald: Is this documented? How will we know if it changes?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+                                if (!TryGetMSBuildSdkPackageInfo(fallbackPackagePathResolver, installedPackage.Name, installedPackage.Version, out installedPath, out installedVersion))
+                                {
+                                    // This should never happen
+                                    errors.Add("Package was installed but could not be found");
+                                }
+                            }
+                            else
+                            {
+                                // This should never happen
+                                errors.Add("Package was supposed to be installed but wasn't");
```
AndyGerlicher: Move the 2 error messages to resources please.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolverBase.cs
```diff
@@ -0,0 +1,98 @@
...
+                string assemblyPath = Path.Combine(NuGetAssemblyPathLazy.Value, $"{assemblyName.Name}.dll");
+
+                if (File.Exists(assemblyPath))
+                {
+#if !FEATURE_APPDOMAIN
+                    return assemblyLoadContext.LoadFromAssemblyPath(assemblyPath);
+#elif !CLR2COMPATIBILITY
+                    return Assembly.UnsafeLoadFrom(assemblyPath);
+#else
+                    return Assembly.LoadFrom(assemblyPath);
```
rainersigwald: How do these loads work in the face of deployed NuGet DLLs that are a different version from what we built against? Is it different between Core and full framework?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
jeffkl: I was having trouble with the wording,  I can't see how the conditions would be hit in the first place.  NuGet says it successfully restored the package but you have to do a secondary lookup of where it was put.  Any suggestions for the messages?

src/Build/project.json
rainersigwald: Is the Newtonsoft.Json reference coming to us transitively through this? Since we use it explicitly I'd like to see it mentioned explicitly.

src/Build/BackEnd/Components/SdkResolution/NuGet/GlobalJsonReader.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        public const string GlobalJsonFileName = "global.json";
+
+        public const string MSBuildSdksPropertyName = "msbuild-sdks";
+
+        /// <summary>
+        /// Walks up the directory tree to find the first global.json and reads the msbuild-sdks section.
+        /// </summary>
+        /// <returns>A <see cref="Dictionary{String,String}"/> of MSBuild SDK versions from a global.json if found, otherwise <code>null</code>.</returns>
+        public static Dictionary<string, string> GetMSBuildSdkVersions(SdkResolverContextBase context)
```
rainersigwald: What's the relationship between our finding and interpreting this file and `dotnet` CLI's? Are we doing anything to make sure they follow the same logic?

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -32,8 +32,8 @@ public void AssertDefaultLoaderReturnsDefaultResolver()
 
             var resolvers = loader.LoadResolvers(_loggingContext, new MockElementLocation("file"));
 
-            Assert.Equal(1, resolvers.Count);
-            Assert.Equal(typeof(DefaultSdkResolver), resolvers[0].GetType());
+            resolvers.Count.ShouldBe(2);
+            resolvers[1].GetType().ShouldBe(typeof(DefaultSdkResolver));
```
rainersigwald: Also check `resolvers[0]`?

src/Build/BackEnd/Components/SdkResolution/NuGet/GlobalJsonReader.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Microsoft.Build.BackEnd.SdkResolution.NuGet
+{
+    /// <summary>
+    /// Reads MSBuild related sections from a global.json.
+    /// </summary>
+    internal static class GlobalJsonReader
+    {
+        public const string GlobalJsonFileName = "global.json";
+
+        public const string MSBuildSdksPropertyName = "msbuild-sdks";
```
rainersigwald: Is this new? Is it [documented](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json)?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkLogger.cs
```diff
@@ -0,0 +1,95 @@
...
+                    break;
+
+                case LogLevel.Warning:
+                    _warnings.Add(data);
+                    break;
+
+                case LogLevel.Error:
+                    _errors.Add(data);
+                    break;
+            }
```
rainersigwald: No `default` to throw if someone adds a log level?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkLogger.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        public void Log(ILogMessage message) => Log(message.Level, message.Message);
+
+        public Task LogAsync(LogLevel level, string data) => Task.Run(() => Log(level, data));
+
+        public Task LogAsync(ILogMessage message) => Task.Run(() => Log(message));
```
rainersigwald: Is this going to bite us in VS threadpool scenarios? Might be safer to just do

```c#
public Task LogAsync(LogLevel level, string data)
{
    Log(level, data);
    return Task.CompletedTask;
}

public Task LogAsync(ILogMessage message)
{
    Log(message);
    return Task.CompletedTask;
}
```

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+    {
+        public override string Name => nameof(NuGetSdkResolver);
+
+        public override int Priority => 2500;
+
+        public override SdkResultBase Resolve(SdkReference sdk, SdkResolverContextBase context, SdkResultFactoryBase factory)
+        {
+            object parsedSdkVersion;
+
+            // This resolver only works if the user specifies a version in a project or a global.json.  Ignore invalid versions, there may be another resolver that can handle the version specified
```
rainersigwald: nit: make this multiline

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
AndyGerlicher: Not concerned much with the message if it shouldn't ever happen. Maybe just `Error resolving package '{0}'. NuGet restored successfully but the package was not found.` Or something like that? Just think it should have the package name and be moved to resource.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                msbuildSdkVersions = (Dictionary<string, string>) context.State;
+            }
+            else
+            {
+                msbuildSdkVersions = GlobalJsonReader.GetMSBuildSdkVersions(context);
+            }
+
+            // Save the SDK versions in case this resolver is called again for another SDK in the same build
+            context.State = msbuildSdkVersions;
```
rainersigwald: roll this into the else?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+                FallbackPackagePathResolver fallbackPackagePathResolver = new FallbackPackagePathResolver(NuGetPathContext.Create(settings));
+
+                string installedPath, installedVersion;
+
+                // Attempt to find a package if its already installed
+                if (!TryGetMSBuildSdkPackageInfo(fallbackPackagePathResolver, sdk.Name, parsedSdkVersion, out installedPath, out installedVersion))
+                {
+                    try
+                    {
+                        context.Logger.LogMessage($"Resolving SDK '{sdk}'...");
```
rainersigwald: Does the logger automatically add the resolver name for additional context?

rainersigwald: Should this be localized?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+
+                        // Asynchronously run the restore without a commit which find the package on configured feeds, download, and unzip it without generating any other files
+                        IReadOnlyList<RestoreResultPair> results = RestoreRunnerEx.RunWithoutCommit(
+                                context.ProjectFilePath,
+                                sdk.Name,
+                                parsedSdkVersion.ToFullString(),
+                                settings,
+                                new NuGetSdkLogger(context.Logger, warnings, errors))
+                            .Result;
```
rainersigwald: Does this need a `ConfigureAwait`? (I'm really asking; I don't grok when that is needed.)

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
rainersigwald: along with other comments in this file, Mr. Wide Monitor :)

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+                                }
+                            }
+                            else
+                            {
+                                // This should never happen
+                                errors.Add("Package was supposed to be installed but wasn't");
+                            }
+                        }
+                    }
+                    catch (Exception e)
```
rainersigwald: This is a larger scope than I'm really comfortable with for a catch-all exception handler. Is there no NuGet exception base class?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                // Cast the NuGet version since the caller does not want to consume NuGet classes directly
+                NuGetVersion parsedSdkVersion = (NuGetVersion) nuGetVersion;
+
+                // Stores errors and warnings for the result
+                ICollection<string> errors = new List<string>();
+                ICollection<string> warnings = new List<string>();
+
+                // Load NuGet settings and a path resolver
+                ISettings settings = Settings.LoadDefaultSettings(context.ProjectFilePath);
```
rainersigwald: Under what circumstances would this need _non_-default settings?

src/Build/BackEnd/Components/SdkResolution/NuGet/RestoreRunnerEx.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+using ILogger = NuGet.Common.ILogger;
+
+namespace Microsoft.Build.BackEnd.SdkResolution.NuGet
+{
+    /// <summary>
+    /// An extension of the NuGet.Commands.RestoreRunner class that contains APIs we do not yet have.
```
rainersigwald: Link to bugs on NuGet for the desired functionality so we can delete this?

src/Build/BackEnd/Components/SdkResolution/NuGet/RestoreRunnerEx.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <param name="projectPath">The full path to the project.</param>
+        /// <param name="id">The ID of the package.</param>
+        /// <param name="version">The version of the package.</param>
+        /// <param name="settings">The NuGet settings to use.</param>
+        /// <param name="logger">An <see cref="ILogger"/> to use for logging.</param>
+        /// <returns></returns>
+        public static Task<IReadOnlyList<RestoreResultPair>> RunWithoutCommit(string projectPath, string id, string version, ISettings settings, ILogger logger)
+        {
+            // NuGet requires at least one framework, we use .NET Standard here just to get the API to do work.  The framework is not actually used.
+            List<NuGetFramework> targetFrameworks = new List<NuGetFramework>
```
rainersigwald: make this a `static readonly` field?

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -20,10 +22,27 @@ internal class SdkResolverLoader
...
+                }
+            }
+
             var potentialResolvers = FindPotentialSdkResolvers(
                 Path.Combine(BuildEnvironmentHelper.Instance.MSBuildToolsDirectory32, "SdkResolvers"));
 
-            if (potentialResolvers.Count == 0) return resolvers;
+            if (potentialResolvers.Count == 0)
+            {
+                return resolvers.Count == 1 ? resolvers : resolvers.OrderBy(t => t.Priority).ToList();
```
rainersigwald: Was this optimization noticeably faster?

src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
```diff
@@ -118,7 +120,24 @@ public SdkResult GetSdkResult(int submissionId, SdkReference sdk, LoggingContext
...
                     try
                     {
-                        SdkResult result = (SdkResult) sdkResolver.Resolve(sdk, context, resultFactory);
+                        SdkResult result;
+
+                        try
+                        {
+                            result = (SdkResult) sdkResolver.Resolve(sdk, context, resultFactory);
+                        }
+                        catch (Exception e) when (sdkResolver is NuGetSdkResolver && e is FileNotFoundException)
```
rainersigwald: Oof, this is specific. But I can't think offhand of a better way.

src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
```diff
@@ -118,7 +120,24 @@ public SdkResult GetSdkResult(int submissionId, SdkReference sdk, LoggingContext
...
+                        {
+                            // MSB4243: The NuGet-based SDK resolver failed to run because NuGet assemblies could not be located.  Check your installation of MSBuild or set the environment variable "{0}" to the folder that contains the required NuGet assemblies. {1}
+                            loggingContext.LogWarning(null, new BuildEventFileInfo(sdkReferenceLocation), "CouldNotRunNuGetSdkResolver", NuGetSdkResolverBase.NuGetAssemblyPathEnvironmentVariableName, e.Message);
+                            continue;
+                        }
+                        catch(Exception e)
+                        {
+                            // MSB4242: The SDK resolver "{0}" failed to run. {1}
+                            loggingContext.LogWarning(null, new BuildEventFileInfo(sdkReferenceLocation), "CouldNotRunSdkResolver", sdkResolver.Name, e.Message);
```
rainersigwald: What were we doing for this before? Is it ok that this is no longer fatal?

src/Build/project.json
```diff
@@ -14,19 +16,26 @@
     "netstandard1.5": {
       "dependencies": {
         "NETStandard.Library": "1.6.0",
+        "NuGet.Commands": "4.5.0",
+        "NuGet.Protocol": "4.5.0",
```
rainersigwald: nit: since these are the same across frameworks it'd be nice to have them in a common node.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
jeffkl: Done

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
jeffkl: The unit test is only verifying that `DefaultSdkResolver` was added.  Should I check for both here or make a new test?

src/Build/BackEnd/Components/SdkResolution/NuGet/GlobalJsonReader.cs
jeffkl: It is so new, that it isn't documented yet.

src/Build/BackEnd/Components/SdkResolution/NuGet/GlobalJsonReader.cs
jeffkl: Unfortunately, the code that finds and parses `global.json` in the dotnet CLI is in unmanaged code ([here](https://github.com/dotnet/core-setup/blob/77152705cd56e7d89f52585959ec3559d91084c4/src/corehost/cli/fxr/fx_muxer.cpp#L894)).

I followed the same overall pattern to find `global.json` but the logic could get out of sync.

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
rainersigwald: I'd change it to `AssertDefaultLoaderReturnsDefaultResolver`**`s`** and check for both here.

src/Build/BackEnd/Components/SdkResolution/NuGet/GlobalJsonReader.cs
rainersigwald: That's cool, but it needs to be. Can we track that work somewhere?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkLogger.cs
jeffkl: This would cause runtime exceptions as soon as NuGet added a new log level.  Would it be better to just have a case for `Error` and `Warning` and everything else is a message?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkLogger.cs
jeffkl: Not sure, but better safe than sorry.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
jeffkl: Wait, so you're telling me not everyone has a 34" monitor by default?  😭

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
jeffkl: The "default" settings means to find the settings in the default way.  Reference: https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied

At this time, there would be no way for a user to override how settings are loaded.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolverBase.cs
jeffkl: I have no idea, maybe @emgarten knows?

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolverBase.cs
jeffkl: They are effectively assembly binding redirects.  If we compile against 4.3 we can safely load 4.5 instead.  As long as there are no breaking API changes, it should work.

src/Build/BackEnd/Components/SdkResolution/NuGet/RestoreRunnerEx.cs
jeffkl: Done. https://github.com/NuGet/Home/issues/5919

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
jeffkl: I left it in to keep existing functionality.  @AndyGerlicher had the optimization in there first, possibly by accident

src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
jeffkl: Yeah we went back and forth on this.  We're explicitly adding the resolver so we felt comfortable with the special case here.  I'll add a comment to explain the reasoning

src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
jeffkl: Yes before it was a fatal error.  When testing things out, a fatal build error when something was wrong with a resolver seemed like overkill.  What do you think?

src/Build/project.json
jeffkl: I'm not sure what this is?

jeffkl: Yes, it comes through [NuGet.ProjectModel](https://www.nuget.org/packages/NuGet.ProjectModel/)

Technically the version doesn't matter since we only load it from where NuGet is at.  We only have one call which I highly doubt would ever be broken.  But if you feel strongly about it...

src/Build/project.json
jeffkl: Okay

src/Build/project.json
```diff
@@ -14,19 +16,26 @@
...
         "System.Collections.Immutable": "1.2.0",
         "System.Collections.NonGeneric": "4.0.1",
         "System.Diagnostics.Contracts": "4.0.1",
         "System.Diagnostics.FileVersionInfo": "4.0.0",
         "System.Diagnostics.Process": "4.1.0",
         "System.Diagnostics.TraceSource": "4.0.0",
+        "System.IO": "4.1.0",
```
jeffkl: These had to be added explicitly because the reference to NuGet caused some unwanted upgrades.

src/Build/BackEnd/Components/SdkResolution/NuGet/GlobalJsonReader.cs
jeffkl: I've added it as an item in my spec: https://github.com/Microsoft/msbuild/issues/2803

src/Build/project.json
rainersigwald: They have a bot that creates PRs like https://github.com/dotnet/sdk/pull/1851. I was wondering if we needed to get them to target us too. Hopefully not, since we don't need the very latest . . . until we adopt their version of the resolverEx thing.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolverBase.cs
emgarten: This is considered a well known location and it is defined in the VSIX manifest, we have it set here also: https://github.com/NuGet/NuGet.BuildTasks/blob/1d8af3499f94a32c1d128a42faceade39c1f4337/src/Microsoft.NuGet.Build.Tasks/ImportBeforeAfter/Microsoft.NuGet.ImportAfter.targets#L14

I would not expect this to change, if it did it would be very hard to find where it moved to.

src/Build/BackEnd/Components/SdkResolution/NuGet/NuGetSdkResolver.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// </summary>
+    internal sealed class NuGetSdkResolver : NuGetSdkResolverBase
+    {
+        public override string Name => nameof(NuGetSdkResolver);
+
+        public override int Priority => 2500;
+
+        public override SdkResultBase Resolve(SdkReference sdk, SdkResolverContextBase context, SdkResultFactoryBase factory)
+        {
+            object parsedSdkVersion;
```
cdmihai: Why object?

cdmihai: Ah, comment in the nested class :)

---
In reply to: [160777249](https://github.com/Microsoft/msbuild/pull/2850#discussion_r160777249) [](ancestors = 160777249)

jnm2:
I've been waiting to hear when I can write a build script that works without requiring any specific machine state by pulling MSBuild and all SDKs from NuGet. I'm specifically interested in the legacy csproj SDKs, the new csproj SDKs, and .vcxproj SDKs.

@rainersigwald [told me](https://github.com/Microsoft/msbuild/issues/1677#issuecomment-278164451) to wait for https://github.com/Microsoft/msbuild/issues/1493 which has been inactive for eight months.  Does this PR bring us any closer to that? Is progress being made and I'm just still out of the loop?

rainersigwald:
@jnm2 Thanks for pointing out that that issue was super stale; I updated it a bit.

This improves the situation for custom SDKs, which can now be acquired from NuGet.

I believe there is no planned work to make the .NET Core SDK available via NuGet, so you must still deploy it separately (or get MSBuild through the Core SDK, which is a single unzippable entity).

There's no planned work to make the installed C# or C++ targets available on NuGet, which doesn't change based on this--they do not use MSBuild Sdks so Sdk acquisition doesn't affect those scenarios.

jnm2:
Thank you!

> There's no planned work to make the installed C# or C++ targets available on NuGet, which doesn't change based on this--they do not use MSBuild Sdks so Sdk acquisition doesn't affect those scenarios.

Who do I talk to?

PR #2852, opened 2018-01-10
Perf: Windows cmd.exe /D to avoid registry check

erikma:
When spawning Windows cmd.exe shell commands, add the /D parameter to avoid the overhead of checking the autorun configuration in the registry


KirillOsenkov:
Could we add this under an escape hatch perhaps? So if someone's build breaks they can turn it back off?

erikma:
RE: cmd line switch to turn off: Not enough of a perf win to complexify the command line IMO. I'll withdraw the PR.

KirillOsenkov:
We have other ways to do escape hatches, namely environment variables. That's what I meant, and not necessarily a command line switch.

rainersigwald:
https://github.com/dotnet/cli/issues/8331 was root-caused to a user having autoruns and would have been fixed by this, as a data point.

KirillOsenkov:
Yeah, I didn’t mean to scare this PR away ;)

erikma:
Easy to bring back, and if there is support overhead from autorun it seems worth it.

erikma:
Reopened with pushed branch. @jeffkl If we need to restore old logic I need to know which env var you'd like (or push a new commit if you'd like)

jeffkl:
I'm almost done.  I've added an escape hatch and some unit tests for good measure.  I will push after local testing is complete.

AndyGerlicher:
Looks good, except for the compile issue with the test and the conflict :)

rainersigwald:
@AndyGerlicher So you think we should take this potentially-breaking change?

jeffkl:
We'll merge this in master once we've branched for 15.6 so that this won't go out until 15.7 preview 1

jeffkl:
Thanks for the contribution @erikma!

Issue #2640, opened 2017-10-12
Provide submission API that can return back named task items and properties


viktorveis:
Today, MSbuild submission API accepts a list of targets. Please enhance it to also accept list of properties and items to build and return them back via BuildResults. This will make it easier to move MSBuild out-of-process.

rainersigwald:
Additional context for this request: it's primarily about improving performance of out-of-proc design-time builds.

Currently, a caller has the option of setting `BuildRequestDataFlags.ProvideProjectStateAfterBuild`, which will populate `BuildResult.ProjectStateAfterBuild` with a full copy of the `ProjectInstance` after the request completes, including _all_ properties and items. Unfortunately, the copy + IPC cost of getting all that data back to the calling process from the out-of-process node is high (possibly 2×), so the IDE would like to avoid doing that.

That can be accomplished today by adding a new target that explicitly returns a known item, and then calling it (that was the motivation for https://github.com/Microsoft/msbuild/pull/2623 which was implemented for 15.5 in another way), but that can be a pain if you know that some other target will populate the items/properties that you want.

So the request here is to add something like `BuildRequestDataFlags.ReturnSubsetOfStateAfterBuild`, with a list of items passed in the `BuildRequestData` that opts in to specific item returns.

An additional consideration is that the wire cost of shipping back item metadata that won't be used might also be nontrivial. Some consumers already filter away metadata that isn't on an "allow list" after getting a `BuildResult`. It would be more efficient to do that filtering on the producing end to save the wire cost.

PR #2854, opened 2018-01-11
Request explicit properties and items in an API build

rainersigwald:
First stab at a design for #2640.

TODO:

- [ ] Design approval from CPS
- [x] Tests
- [x] ~Filter `ResultsByTarget`~ (Dropping for now as a secondary goal)
- [x] ~Ensure in-proc `BuildResult` cache doesn't get polluted after filtering `ResultsByTarget`~

### DIFF
--- a/ref/net46/Microsoft.Build/Microsoft.Build.cs
+++ b/ref/net46/Microsoft.Build/Microsoft.Build.cs
@@ -987,8 +987,10 @@ namespace Microsoft.Build.Execution
         public BuildRequestData(Microsoft.Build.Execution.ProjectInstance projectInstance, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
         public BuildRequestData(Microsoft.Build.Execution.ProjectInstance projectInstance, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
         public BuildRequestData(Microsoft.Build.Execution.ProjectInstance projectInstance, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags, System.Collections.Generic.IEnumerable<string> propertiesToTransfer) { }
+        public BuildRequestData(Microsoft.Build.Execution.ProjectInstance projectInstance, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags, System.Collections.Generic.IEnumerable<string> propertiesToTransfer, Microsoft.Build.Execution.RequestedProjectState requestedProjectState) { }
         public BuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, string toolsVersion, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
         public BuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, string toolsVersion, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public BuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, string toolsVersion, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags, Microsoft.Build.Execution.RequestedProjectState requestedProjectState) { }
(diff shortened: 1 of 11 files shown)


ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -984,6 +984,8 @@ public partial class BuildRequestData
         public BuildRequestData(Microsoft.Build.Execution.ProjectInstance projectInstance, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags, System.Collections.Generic.IEnumerable<string> propertiesToTransfer) { }
         public BuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, string toolsVersion, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
         public BuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, string toolsVersion, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public BuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, string toolsVersion, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags, Microsoft.Build.Execution.BuildResultFilter filter) { }
```
jeffkl: What's the threshold of parameters where we decide to stop adding them to the constructor and just have a setter? 😁

src/Build/Instance/ProjectInstance.cs
```diff
@@ -1474,6 +1569,21 @@ public ProjectInstance DeepCopy()
...
+        /// Create an independent clone of this object, keeping ONLY the explicitly
+        /// requested project state.
+        /// </summary>
+        /// <remarks>
+        /// Useful for reducing the wire cost of IPC for out-of-proc nodes used during
+        /// design-time builds that only need to populate a known set of data.
+        /// </remarks>
+        /// <param name="filter">Project state that should be returned.</param>
+        /// <returns></returns>
+        public ProjectInstance FilteredCopy(RequestedProjectState filter)
```
jeffkl: Is there a way to only filter when we're about to send the data over the wire?  I feel that would be more straightforward where the out-of-proc node has all the data and as its writing to the named pipe its filtering out properties/items.  This seems to clone the ProjectInstance in preparation for serializing.  Do you think it would be better to do what I suggest and is it even possible with our current code?

ref/net46/Microsoft.Build/Microsoft.Build.cs
rainersigwald: Perpetually Just One More!

src/Build/Instance/ProjectInstance.cs
rainersigwald: I don't think we can intercept only at serialization time. This approach (filter in the TargetBuilder when constructing a BuildResult) has the advantage of working in-proc as well, which is nice.

Though I guess the contract could be "return _at least_ the requested state", and returning the whole shebang in-proc would actually be faster (no deep clone).

If we were to do the filtering at the Translator level, we'd have to have the filter available in ProjectInstance, so that its TranslateAllState method could apply the filter. That's possible, but it seems like odd layering: only the result really cares about the filter; why should the ProjectInstance care?

Thoughts?

rainersigwald:
Note that I dropped filtering of TargetResults--it can be added again later if it proves necessary.

ref/net46/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -1005,6 +1008,7 @@ public enum BuildRequestDataFlags
         IgnoreExistingProjectState = 4,
         None = 0,
         ProvideProjectStateAfterBuild = 2,
+        ProvideSubsetOfStateAfterBuild = 32,
```
rainersigwald: Should we require this, or infer it when you pass a `RequestedProjectState`? Right now it's not really distinct.

ref/netstandard1.3/Microsoft.Build/Microsoft.Build.cs
```diff
@@ -1337,6 +1342,12 @@ public sealed partial class ProjectTaskOutputPropertyInstance : Microsoft.Build.
         public string TaskParameter { get { throw null; } }
         public override Microsoft.Build.Construction.ElementLocation TaskParameterLocation { get { throw null; } }
     }
+    public partial class RequestedProjectState
+    {
+        public RequestedProjectState() { }
+        public System.Collections.Generic.IDictionary<string, System.Collections.Generic.List<string>> ItemFilters { get { throw null; } set { } }
+        public System.Collections.Generic.List<string> PropertyFilters { get { throw null; } set { } }
```
rainersigwald: How do we feel about this interface? Would it be preferable to have constructor arguments, or another data type? One thing of concern is that it'd be nice to enforce case-insensitivity on in-proc cases.

PR #2858, opened 2018-01-12
Fix NuGet.MSBuildSdkResolver content files in localization package

jeffkl:
Default "buildAction" is "compile" and default "copyToOutput" is "false"

:cry:


jeffkl:
Yeah my first iteration was incorrect.  You include files with `<files />` and then you specify how to handle `<contentFiles />` in another section.  I've updated this and tested it in the .NET CLI repo

PR #2859, opened 2018-01-13
Fix NuGet SDK Resolver issues on latest NuGet

AndyGerlicher:
Latest version of NuGet adds a check to ensure that spec.Name is equal to spec.RestoreMetadata.ProjectName.

https://github.com/NuGet/NuGet.Client/commit/358bcc31911e9d2687c91e3517585f82fa510960

Also resolved an issue where we fail on the first build if the .nuget folder is missing completely.


jeffkl:
I pushed a commit to fix the error message text and reverted the nuspec that always sneaks in

PR #2863, opened 2018-01-16
[osx] Cache the IsOSX check

alanmcgovern:
This ensure we won't hit the filesystem a few million times
during project evaluation.

Fixes https://github.com/mono/mono/issues/6530

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -546,14 +546,22 @@ namespace Microsoft.Build.Shared
             get { return RuntimeInformation.IsOSPlatform(OSPlatform.Windows); }
 #endif
         }
-
+        
+#if MONO
+        private static bool? _isOSX;
+#endif
         /// <summary>
         /// Gets a flag indicating if we are running under Mac OSX
         /// </summary>
         internal static bool IsOSX
         {
 #if MONO
-            get { return File.Exists("/usr/lib/libc.dylib"); }
+            get {
+                if (!_isOSX.HasValue) {
+                    _isOSX = File.Exists("/usr/lib/libc.dylib");
+                }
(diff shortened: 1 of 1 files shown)


src/Shared/NativeMethodsShared.cs
```diff
@@ -546,14 +546,21 @@ internal static bool IsWindows
             get { return RuntimeInformation.IsOSPlatform(OSPlatform.Windows); }
 #endif
         }
-
+        
+#if MONO
+        private static bool? _isOSX;
```
Therzok: Wouldn't it be better to make a `Lazy<bool>` which calls [this](https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Core/Platform.cs#L67-L84)?

Therzok: I suppose a pinvoke is faster than IO, and lazy will also prevent multiple evaluation.

alanmcgovern: I just wanted a minimal patch which can be easily/safely backported to release branches. If we want a more invasive patch it might be frowned on when we go to backport it

alanmcgovern: (also, multiple execution is not something that we need to be particularly worried about in this case, so i left it as a simple nullable bool)

alanmcgovern:
Braces have been fixed!

Thanks

rainersigwald:
Thanks!

alanmcgovern:
It's not thread safe but the only impact of that would be us hitting the filesystem a handful of times instead of just once. Either way we've reduced literally *millions* of filesystem hits to a tiny bounded number of invocations, typically just 1.

PR #2864, opened 2018-01-16
[Performance] Do not probe for file existence every time we want to check if this is a Mac

migueldeicaza:
Addresses https://github.com/mono/mono/issues/6530


Therzok:
Dupe of https://github.com/Microsoft/msbuild/pull/2863

migueldeicaza:
Alan's version is better

PR #2867, opened 2018-01-16
Skip TF-compat checks for P2Ps that don't reference output

### LINKED ISSUE #2661: P2Ps should be allowed when ReferenceOutputAssembly=false even given TFM incompatibilities
_From @AArnott on March 3, 2017 18:40_

With VS2015 projects, I can have a P2P from a portable library to a net46 library by setting metadata on the project reference:

```xml
<ProjectReference Include="..\SomeNet46Lib\lib.csproj">
  <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
</ProjectReference>
```

But with the .NET SDK projects, even with this metadata the build fails:

> C:\Program Files (x86)\Microsoft Visual Studio\2017\d15rel\MSBuild\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.Sdk.Common.targets(73,5): error : Project 'C:\git\pinvoke\src\CodeGeneration\CodeGeneration.csproj'
targets '.NETFramework,Version=v4.6'. It cannot be referenced by a project that targets '.NETPortable,Version=v0.0,Profile=Profile92'. [C:\git\pinvoke\src\CodeGeneration\CodeGeneration.csproj]

This blocks scenarios where a P2P exists merely for the sake of ensuring build ordering but without the assembly reference. In my particular scenario, the referenced project provides a binary that the build of the portable library picks up for code generation purposes.

_Copied from original issue: dotnet/sdk#939_

rainersigwald:
Allows having a `ProjectReference` pointing to a project with an incompatible TF, as long as you don't reference the output of that project.

This is common, for example, if you have a netcoreapp1.0 code generator that you want to run in a later project's build.

When we introduced multitargeting support in 15.x, we started doing complicated dances to figure out the "most compatible" TF of the referenced project, given the referencing project's current TF. But that's only relevant if there's going to be a direct reference.

This was especially annoying because adding a solution-level build dependency secretly creates `ProjectReference` items with `ReferenceOutputAssembly=false`, _even though they're not in the project file_.

Fixes #2661 / dotnet/sdk#939.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1543,9 +1543,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
       We give this treatment to .vcxproj by default since no .vcxproj can target more
       than one framework.
+
+      Likewise if the dependency is for build ordering instead of an assembly reference
+      (ReferenceOutputAssembly=false), skip the checks since we can't know what TF
+      the output would need to be compatible with.
    -->
    <ItemGroup>
-      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and '%(Extension)' == '.vcxproj'">
+      <_MSBuildProjectReferenceExistent
+        Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and
+                   ('%(Extension)' == '.vcxproj' or '%(ReferenceOutputAssembly)' == 'false')">
         <SkipGetTargetFrameworkProperties>true</SkipGetTargetFrameworkProperties>
       </_MSBuildProjectReferenceExistent>
    </ItemGroup>


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1543,9 +1543,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+      Likewise if the dependency is for build ordering instead of an assembly reference
+      (ReferenceOutputAssembly=false), skip the checks since we can't know what TF
+      the output would need to be compatible with.
    -->
    <ItemGroup>
-      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and '%(Extension)' == '.vcxproj'">
+      <_MSBuildProjectReferenceExistent
+        Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and
+                   ('%(Extension)' == '.vcxproj' or '%(ReferenceOutputAssembly)' == 'false'">
```
rainersigwald: I'm slightly worried about this change. Could there be any projects today that ReferenceOutputAssembly=false but somehow depend on the TF check (and weren't broken by 15.5's changes to it)? I can't think of any, and I think this change makes a lot of sense. But I do worry.

jeffkl: They would have had to have taken a dependency on `_MSBuildProjectReferenceExistent` to have custom logic do something with dependencies since they set `ReferenceOutputAssembly` to `false`, and we can't guarantee anything if they did that.

I think its also missing a trailing `)` right?

```diff
-('%(Extension)' == '.vcxproj' or '%(ReferenceOutputAssembly)' == 'false'
+('%(Extension)' == '.vcxproj' or '%(ReferenceOutputAssembly)' == 'false')
```

src/Tasks/ResolveProjectBase.cs
```diff
@@ -293,6 +293,7 @@ protected void AddSyntheticProjectReferences(string currentProjectAbsolutePath)
                     item.SetMetadata("ReferenceOutputAssembly", "false");
                     item.SetMetadata("LinkLibraryDependencies", "false");
                     item.SetMetadata("CopyLocal", "false");
+                    item.SetMetadata("SkipGetTargetFrameworkProperties", "false");
```
jeffkl: Why not set this to `true`?

rainersigwald: Apparently skipping caffeine this morning wasn't my best idea.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: missing trailing `)`: yup!

They wouldn't have had to take a dependency on `_MSBuildProjectReferenceExistent`, though--they could have just had

```xml
<ProjectReference Include="..\some\other.csproj">
  <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
</ProjectReference>
```

And it would have flowed through to here.

src/Tasks/ResolveProjectBase.cs
```diff
@@ -293,6 +293,7 @@ protected void AddSyntheticProjectReferences(string currentProjectAbsolutePath)
                     item.SetMetadata("ReferenceOutputAssembly", "false");
                     item.SetMetadata("LinkLibraryDependencies", "false");
                     item.SetMetadata("CopyLocal", "false");
+                    item.SetMetadata("SkipGetTargetFrameworkProperties", "true");
```
nguerrera: Is this necessary if ReferenceOutputAssembly=false now implies SkipGetTargetFrameworkProperties?

rainersigwald: Not strictly, but I can't think of a compat problem with doing this, while I can maybe conceive of one with the other approach, so I decided to do both and possibly back out the other.

rainersigwald:
I did an E2E test with a solution that has an incompatible solution dependency and an incompatible-but-marked-no-ReferenceOutputAssembly ProjectReference. 

[IncompatibleSolutionDependencies.zip](https://github.com/Microsoft/msbuild/files/1636848/IncompatibleSolutionDependencies.zip)

It seems to restore fine (even with no patch) using cli 2.1.4, and solution build succeeds with only the targets part of this patch applied.

src/Tasks/ResolveProjectBase.cs
nguerrera: I would keep the other, too. I can't think of a scenario where you want to pick the best TFM but not reference the output assembly. SkipGetTargetFrameworkProperties was intended as an escape hatch rather than something you needed to spell out in real scenarios.

rainersigwald:
Unfortunately, this was a bad solution to the problem, because it changed behavior for projects that did have a matching TF in the `ReferenceOutputAssembly=false` case. That broke several early adopters -- see #2922 and https://github.com/Microsoft/msbuild/pull/2923#issuecomment-361667146.

As a result, I'm going to revert this.

But all is not lost! https://github.com/NuGet/NuGet.Client/pull/1983 adds the ability to pass "fallback frameworks" to the `GetReferenceNearestTargetFrameworkTask`, which makes it easier to avoid the error. #2934 tracks pulling that into common targets.

PR #2869, opened 2018-01-16
Minor improvements to SDK resolver end-user experience

jeffkl:
A few minor improvements:

1. Make SdkResolverLoader easier to mock by splitting functionality into more virtual methods
2. Add a different error message if the SDK resolver assembly itself cannot be loaded
3. Fix issue where if one resolver failed to load, the remaining ones would not be loaded
4. More unit tests

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Build.BackEnd.SdkResolution
             }
             catch (Exception e)
             {
-                loggingContext.LogWarning(null, new BuildEventFileInfo(location), "CouldNotLoadSdkResolverAssembly", resolverPath, e.Message);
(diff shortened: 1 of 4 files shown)


src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -73,5 +57,58 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder)
...
+        protected virtual void LoadResolvers(string resolverPath, LoggingContext loggingContext, ElementLocation location, List<SdkResolver> resolvers)
+        {
+            Assembly assembly;
+            try
+            {
+                assembly = LoadResolverAssembly(resolverPath, loggingContext, location);
+            }
+            catch (Exception e)
+            {
+                loggingContext.LogWarning(null, new BuildEventFileInfo(location), "CouldNotLoadSdkResolverAssembly", resolverPath, e.Message);
```
AndyGerlicher: Should this be an error or warning? If one of the resolver fails to load it would probably mean your MSBuild instance is messed up. You could end up hitting a different resolver that may work.

jeffkl: It is now throwing an `InvalidProjectFileException` which stops the build with a logged error or can be caught by consumers of the API.

PR #2875, opened 2018-01-18
Update ref file to match GenAPI results

mfilippov:


mfilippov:
It's strange locally I have this change when build Full framework version.

rainersigwald:
IIRC that symbol is exported only for Debug builds. It's pretty annoying that it toggles back and forth, though.

mfilippov:
@rainersigwald Thanks for the information.

PR #2878, opened 2018-01-18
Ensure *.CopyComplete file gets removed on Clean

radical:
The @(CopyUpToDateMarker) file is not added to @(FileWrites), so it gets
left behind after a Clean.

https://bugzilla.xamarin.com/show_bug.cgi?id=58174
- This is a Xamarin.Android bug, but the same issue is reproducible with
  regular .net projects too

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -4308,7 +4308,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
          input to projects that reference this one. -->
     <Touch Files="@(CopyUpToDateMarker)"
            AlwaysCreate="true"
-           Condition="'@(ReferencesCopiedInThisBuild)' != ''" />
+           Condition="'@(ReferencesCopiedInThisBuild)' != ''">
+        <Output TaskParameter="TouchedFiles" ItemName="FileWrites" />
+    </Touch>
 
   </Target>
 


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4310,6 +4310,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
            AlwaysCreate="true"
            Condition="'@(ReferencesCopiedInThisBuild)' != ''" />
```
rainersigwald: Could you instead do this?

```xml
     <Touch Files="@(CopyUpToDateMarker)"
             AlwaysCreate="true"
             Condition="'@(ReferencesCopiedInThisBuild)' != ''">
       <Output TaskParameter="TouchedFiles" ItemName="FileWritesShareable"/>
     </Touch>
```

That is more parallel to the copy task above, and it might be infinitesimally faster to avoid the additional implied item creation below.

radical: Yes, that is definitely better. BTW, why `*Shareable`? IIUC, this file should get removed if the project's build outputs are being removed. Please correct me if I'm wrong.

rainersigwald: You're right, use `FileWrites`, not `-Shareable`. That naming convention never really sunk in for me and I find it confusing every time I have to think about it.

radical:
Updated.

AndyGerlicher:
Looks good, thanks!

PR #2880, opened 2018-01-19
Update toolset unix script

cdmihai:
This makes build.sh -build -ci -skipTests -bootstrapOnly work.

Future PRs for getting tests to work with the stage0 msbuild, and getting the bootstrapped msbuild to work at all.


cdmihai:
Checked that this does not break cibuild.cmd on Windows.

PR #2868, opened 2018-01-16
Update toolset, retarget to .NET Core 2.0 and .NET Standard 2.0

dsplaisted:
Addresses #2706

PR #2883, opened 2018-01-21
more readable

callmekohei:
It's a little hard to read init-tool.sh.

So I think that it should be more readble.

### DIFF
--- a/init-tools.sh
+++ b/init-tools.sh
@@ -1,21 +1,72 @@
 #!/usr/bin/env bash
 
+# ---------------------------------------------------------------------
+# Util functions
+# ---------------------------------------------------------------------
+
+partingLine () {
+  printf -v _hr "%*s" $(tput cols) && echo ${_hr// /${1--}}
+}
+
+
+# ---------------------------------------------------------------------
+# Global variables
+# ---------------------------------------------------------------------
+
+#   msbuild
+#   │
+#   ├── init-tools.sh                        <---- __scriptpath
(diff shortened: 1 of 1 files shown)


AndyGerlicher:
I like this, but if you look at #2868 we're about to update our toolsets and change a lot of the way we're building, I don't think these changes will transfer easily :(. If you did want to commit off of that branch (which will go to master soon), that's something we would consider. Really do appreciate you taking the time to improve things. Hope some of this work can be saved.

fbessho:
@AndyGerlicher I'm an outsider here but are you asking @callmekohei to merge his change to init-tool.sh to the `exp/update-toolset` branch?
When I check the branch, #2868 doesn't touch `init-tool.sh` so I think adding his change on this PR to the `exp/update-toolset` branch can be done easily.

I can see init-tools.*cmd* in `git diff` below but not init-tools.*sh*.
```
git diff --name-only origin/master..origin/exp/update-toolset | grep "init-tools"
init-tools.cmd
```

rainersigwald:
`init-tools.sh` should no longer be used in the `exp/update-toolset` branch.  It should have been deleted like `init-tools.cmd` was. Thanks for noticing @fbessho.

@callmekohei Sorry, but since the script is obsoleted by that PR I don't think we should take this. I really appreciate the work, though--the changes are very nice!

callmekohei:
Thank you (^_^)/

PR #2885, opened 2018-01-22
Generate binding redirects for .NET 4.7.2+

### LINKED ISSUE #2481: Generate binding redirects should be on by default
_From @jcouv on August 4, 2017 19:37_

A number of customers ran into issues with ValueTuple in .NET Standard 2.0 in unittests with EntityFramework. It turns out the problem was missing binding redirects.
@weshaggard recommended two settings for such projects:
```
    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
```

Those settings are apparently set for executable project templates, but not for unittests.

https://github.com/aspnet/EntityFramework/issues/9046#issuecomment-314187616
https://github.com/ErikEJ/EntityFramework.SqlServerCompact/issues/463#issuecomment-314188903

FYI @ericstj @terrajobst @nguerrera 

_Copied from original issue: dotnet/project-system#2692_

rainersigwald:
In The Old Days, binding redirect generation was set to off-by-default.
That has caused a great deal of confusion because it's usually the right
thing to do but it's not obvious when you need to do it.

Unfortunately, we currently think we can't change the default for all
projects, because a user could have been avoiding binding redirects
intentionally, making creating them a breaking change.

Since no projects currently target .NET Framework 4.7.2 (since it's not
released), it is not a breaking change to turn on binding redirects by
default for this (and higher) versions--a user must explicitly opt into
targeting the framework and thus into this new behavior.

A user can still specify

```xml
<AutoGenerateBindingRedirects>false</AutoGenerateBindingRedirects>
```

to explicitly opt out of binding redirects for a given project.

Closes #2481, in partial support of

https://github.com/dotnet/designs/blob/2be39b7e1bc9c4df45b1d51ebb2d1abb0689e4e0/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -324,6 +324,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <PropertyGroup>
     <GenerateBindingRedirectsOutputType Condition="'$(OutputType)'=='exe' or '$(OutputType)'=='winexe'">true</GenerateBindingRedirectsOutputType>
+    <!-- It would be a breaking change to automatically turn on binding redirects for existing projects, so turn them on only when opting into a new framework. -->
+    <AutoGenerateBindingRedirects Condition="'$(AutoGenerateBindingRedirects)' == '' and '$(TargetFrameworkIdentifier)' == '.NETFramework' and '$(TargetFrameworkVersion.TrimStart(v))' >= '4.7.2'">true</AutoGenerateBindingRedirects>
   </PropertyGroup>
   <PropertyGroup Condition="'$(AutoUnifyAssemblyReferences)' == ''">
     <AutoUnifyAssemblyReferences>true</AutoUnifyAssemblyReferences>
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -325,7 +325,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <PropertyGroup>
     <GenerateBindingRedirectsOutputType Condition="'$(OutputType)'=='exe' or '$(OutputType)'=='winexe'">true</GenerateBindingRedirectsOutputType>
     <!-- It would be a breaking change to automatically turn on binding redirects for existing projects, so turn them on only when opting into a new framework. -->
-    <AutoGenerateBindingRedirects Condition="'$(AutoGenerateBindingRedirects)' == '' and '$(TargetFrameworkIdentifier)' == '.NETFramework' and '$(TargetFrameworkVersion.TrimStart(v))' >= '4.7.2'">true</AutoGenerateBindingRedirects>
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -324,6 +324,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <PropertyGroup>
     <GenerateBindingRedirectsOutputType Condition="'$(OutputType)'=='exe' or '$(OutputType)'=='winexe'">true</GenerateBindingRedirectsOutputType>
+    <!-- It would be a breaking change to automatically turn on binding redirects for existing projects, so turn them on only when opting into a new framework. -->
+    <AutoGenerateBindingRedirects Condition="'$(AutoGenerateBindingRedirects)' == '' and '$(TargetFrameworkIdentifier)' == '.NETFramework' and '$(TargetFrameworkVersion.Substring(1))' >= '4.7.2'">true</AutoGenerateBindingRedirects>
```
jeffkl: Another reference to `TargetFrameworkVersion.Substring(1)` indicates that there's not always a leading "v".

https://github.com/Microsoft/msbuild/blob/master/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1898-L1899

Does this need similar logic?

rainersigwald: I don't know why that logic is the way it is. All the other comparisons assume it starts with a `v`. But yeah, should be cautious.

rainersigwald:
I confirmed that these worked with a private internal .NET 4.7.2 developer pack by creating a new project and referencing

```xml
    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
    <PackageReference Include="WindowsAzure.Storage" Version="7.2.1" />
```

Then removing the NuGet-altered binding redirects, removing the template binding redirect generation, and confirming that binding redirects still got created.

rainersigwald:
@dotnet-bot test this please

rainersigwald:
@dotnet-bot test this please

rainersigwald:
fyi @dsplaisted this carries a fix for #2886.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -324,6 +324,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <PropertyGroup>
     <GenerateBindingRedirectsOutputType Condition="'$(OutputType)'=='exe' or '$(OutputType)'=='winexe'">true</GenerateBindingRedirectsOutputType>
+    <!-- It would be a breaking change to automatically turn on binding redirects for existing projects, so turn them on only when opting into a new framework. -->
+    <AutoGenerateBindingRedirects Condition="'$(AutoGenerateBindingRedirects)' == '' and '$(TargetFrameworkIdentifier)' == '.NETFramework' and '$(TargetFrameworkVersion.TrimStart(vV))' >= '4.7.2'">true</AutoGenerateBindingRedirects>
```
terrajobst: What about class libraries? [We talked about not wanting to turn this feature on](https://github.com/dotnet/designs/blob/master/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md#class-libraries) for class libraries unless they are unit test projects.

PR #2895, opened 2018-01-23
Allow SDK Version in explicit project imports

AndyGerlicher:
All the code was here but the attribute was not allowed on the Import
element at project parse time for Version or MinimumVersion. This change
allows those attributes and rest of the code worked as expected. Added
tests for the scenario and a few others.

### DIFF
--- a/src/Build.OM.UnitTests/Construction/ProjectSdkImplicitImport_Tests.cs
+++ b/src/Build.OM.UnitTests/Construction/ProjectSdkImplicitImport_Tests.cs
@@ -6,11 +6,14 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Xml;
 using Microsoft.Build.Construction;
 using Microsoft.Build.Engine.UnitTests;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
+using Microsoft.Build.Framework;
+using Shouldly;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


src/Build.OM.UnitTests/Construction/ProjectSdkImplicitImport_Tests.cs
```diff
@@ -398,5 +452,11 @@ private void VerifyPropertyFromImplicitImport(Project project, string propertyNa
 
             Assert.Equal(expectedValue, property.EvaluatedValue);
         }
+
+        private SdkReference GetParsedSdk(ProjectImportElement element)
+        {
+            PropertyInfo parsedSdkInfo = typeof(ProjectImportElement).GetProperty("ParsedSdkReference", BindingFlags.Instance | BindingFlags.NonPublic);
```
rainersigwald: Isn't the point of this test assembly to _not_ have to depend on internals?

AndyGerlicher: Yeah, but for verification I thought this was cleaner than anything else. I was concerned mainly with parsing, and using this allows me to assert that the data going to the resolver from the evaluator is exactly what was intended. It might mean there's not enough disambiguation in the OM since I think some of the internals have changed since it was originally exposed. But not something that I wanted to change at this point.

AndyGerlicher:

AndyGerlicher:

PR #2898, opened 2018-01-23
Prevent OverflowException in Expander.cs from crashing MSBuild

KirillOsenkov:
Convert.ChangeType can throw OverflowException.

Catch that and return null otherwise it crashes MSBuild.
See https://github.com/Microsoft/msbuild/issues/2882.

This fixes the actual crash, but not the systemic issues uncovered by the crash. This is just to stop the bleeding. Proper fixes will follow.

This is an easy safe fix that I'd like to get into 15.6.


rainersigwald:
Can you rebase on the `vs15.6` branch so it's easier to merge either way?

KirillOsenkov:
@AndyGerlicher @rainersigwald all green - can this go in?

KirillOsenkov:
Added another commit that adds some more well-known functions to the fast path. 

This is needed so that Roslyn.sln evaluates cleanly without first-chance exceptions. Roslyn has added usage of these functions in this commit:
https://github.com/dotnet/roslyn/commit/56ec3cfa6ede7c45dc9bc1ca53c07f6243d8e016

KirillOsenkov:
Added another commit to account for this usage introduced by @rainersigwald just yesterday ;)
https://github.com/Microsoft/msbuild/commit/f388fc5025c426401e7c1af3effc13e4f60ca08c

AndyGerlicher:
This will go in tonight before I start the insertion build. It just has to stay green long enough to merge :)

KirillOsenkov:
Cool. This seems to be the last ones. Roslyn now evaluates without throwing.

KirillOsenkov:
Yay! Thanks! I'm so happy this went in!

src/Build/Evaluation/Expander.cs
```diff
@@ -3347,6 +3347,30 @@ private bool TryExecuteWellKnownFunction(out object returnVal, object objectInst
...
+                            returnVal = text.PadRight(totalWidth, paddingChar[0]);
+                            return true;
+                        }
+                    }
+                    else if (string.Equals(_methodMethodName, "TrimStart", StringComparison.OrdinalIgnoreCase))
+                    {
+                        string trimChars;
+                        if (TryGetArg(args, out trimChars) && trimChars.Length > 0)
+                        {
+                            returnVal = text.TrimStart(trimChars.ToCharArray());
```
rainersigwald: 👍 I meant to check for this since I just added a use for this overload in common targets.

PR #2899, opened 2018-01-24
Project.LastEvaluationId should be valid during logging

cdmihai:
CPS expects Project.LastEvaluationId to reflect the id of the ongoing, not yet finished evaluation.
This invariant regressed in #2748

LoggingService now creates an evaluation build event context separate from sending project started events. This allows the Evaluator to set the evaluation Id on Project before the first evaluation event gets fired.


AndyGerlicher:
FYI rebased to 15.6

PR #2903, opened 2018-01-24
Include obsolete assemblies in ExternalAPIs

rainersigwald:
This is required to switch internal VS build processes off of the (obsolete) fork of MSBuild that's still checked into the VS repo. These projects are referenced but weren't available in the ExternalAPIs package.


rainersigwald:
Moving to 15.7 since we don't need this for 15.6 but I think we will need it before we figure out the master insertion problems.

PR #2907, opened 2018-01-24
RI vs15.6 to master

jeffkl:
* Fix preprocessor when imported documents specify Sdk

Implicit imports for imported projects are shown in the preprocessor at the top and bottom of the top level project rather than the imported project.

To fix this, I've created a mapping of which documents contain the implicit imports.  Before processing a document, the imports are then added to that document rather than the top level one.

* Add unit test


jeffkl:
What kind of commit should I do?

AndyGerlicher:
Always do a merge commit for these. Otherwise it will have to resolve conflicts later.

PR #2910, opened 2018-01-25
Readme updates

rainersigwald:


rainersigwald:
It's not the destination, it's the journey.

PR #2924, opened 2018-01-30
Fix Windows build, embed interop types, generate documentation files

dsplaisted:
- Fix Windows build when not called from build script
- Embed interop types
- Generate documentation files, include them in NuGet packages

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,11 +1,7 @@
 <Project>
-  <!-- Taken from Microsoft.VisualStudio.SDK.EmbedInteropTypes NuGet package which tries to fix this issue
-       (but doesn't include this assembly in their list).  -->
-  <Target Name="LinkVSEmbeddableAssemblies" DependsOnTargets="ResolveReferences" AfterTargets="ResolveReferences">
-    <ItemGroup>
-      <ReferencePath Condition="'%(FileName)' == 'Microsoft.VisualStudio.Setup.Configuration.Interop'">
-        <EmbedInteropTypes>true</EmbedInteropTypes>
-      </ReferencePath>
-    </ItemGroup>
-  </Target>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <PackageReference Include="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.15" />
+  </ItemGroup>
+
 </Project>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -39,6 +39,10 @@
          https://github.com/Microsoft/msbuild/issues/2865-->
     <GenerateResourceMSBuildArchitecture>CurrentArchitecture</GenerateResourceMSBuildArchitecture>
     <GenerateResourceMSBuildRuntime>CurrentRuntime</GenerateResourceMSBuildRuntime>
+
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <!-- Disable warnings from XML documentation. https://github.com/Microsoft/msbuild/issues/2927 -->
+    <NoWarn>$(NoWarn);1570;1572;1573;1587</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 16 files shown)


dsplaisted:
@dotnet-bot test OSX10.13 Build for CoreCLR

Directory.Build.targets
```diff
@@ -1,11 +1,21 @@
...
+       Microsoft.VisualStudio.SDK.EmbedInteropTypes package enables by default), because otherwise
+       VtblGap methods get generated in the reference source, resulting in compilation errors:
+          CS0531: 'ISetupInstance2._VtblGap1_8()': interface members cannot have a definition
+    -->
+  <Target Name="LinkVSEmbeddableAssemblies" DependsOnTargets="ResolveReferences" AfterTargets="LinkVSSDKEmbeddableAssemblies"
+          BeforeTargets="FindReferenceAssembliesForReferences">
     <ItemGroup>
       <ReferencePath Condition="'%(FileName)' == 'Microsoft.VisualStudio.Setup.Configuration.Interop'">
-        <EmbedInteropTypes>true</EmbedInteropTypes>
+        <EmbedInteropTypes>false</EmbedInteropTypes>
```
cdmihai: Doesn't this mean we have to redistribute them?

src/Directory.Build.props
```diff
@@ -39,6 +39,9 @@
          https://github.com/Microsoft/msbuild/issues/2865-->
     <GenerateResourceMSBuildArchitecture>CurrentArchitecture</GenerateResourceMSBuildArchitecture>
     <GenerateResourceMSBuildRuntime>CurrentRuntime</GenerateResourceMSBuildRuntime>
+
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>$(NoWarn);1570;1572;1573;1587</NoWarn>
```
cdmihai: What are those warnings? Shouldn't we leave them on?

Directory.Build.targets
AndyGerlicher: Yeah this is wrong. Adding that package above should set `EmbedInteropTypes` to `true` which is what we want (passes them as `/link` rather than a reference to the compiler).

src/Shared/UnitTests/TestEnvironment.cs
```diff
@@ -74,10 +74,6 @@ private void Cleanup()
                 foreach (var item in _invariants)
                     item.AssertInvariant(_output);
             }
-            else
-            {
-                throw new InvalidOperationException("Already disposed");
```
cdmihai: Was this failing anywhere? Why?

src/Directory.Build.props
AndyGerlicher: At least document what they are and if/when they should be removed from this list.

src/Shared/UnitTests/TestAssemblyInfo.cs
```diff
@@ -91,3 +95,24 @@ public void Dispose()
...
+    public void Dispose()
+    {
+        if (!_disposed)
+        {
+            _testEnvironment.Dispose();
+
+            _disposed = true;
+        }
+    }
+}
```
AndyGerlicher: This is great! I'm actually pretty surprised the tests pass with this :)

src/Shared/UnitTests/TestEnvironment.cs
dsplaisted: Yes, when an invariant failed, then the exception would stop `GC.SuppressFinalize` from being called, and so `Cleanup` would be called again from the destructor.

Directory.Build.targets
dsplaisted: I'm not sure, I think they're in the GAC.  I had problems with the reference assembly source when this was enabled.  If we have to embed interop types for this assembly we'll have to find a way to fix the reference assembly issue.

src/Directory.Build.props
dsplaisted: These are the warnings about the XML doc comments.  There were hundreds of them.  I couldn't see that they were disabled before the toolset upgrade, but I checked a bit of the code in master, and it looked like the warnings would have been generated there too.  So I think somehow they were being ignored.

PR #2925, opened 2018-01-30
Stop using static constructor to wire up AssemblyResolve event handler in NuGetSdkResolver

jeffkl:
My original intent was to wire up the AssemblyResolve event as soon as possible.  Later I moved all NuGet type references to private classes instead which means that NuGet assemblies are not loaded until these private classes are accessed.  The private classes are not accessed unless the user specified a version.

So wiring up AssemblyResolve in a static constructor means that in Visual Studio all AssemblyResolve events are sent through the NuGet SDK resolver which is not ideal.  Instead, it can be wired up prior to resolving an SDK and then removed.

I also added a unit test to ensure that NuGet assemblies are not loaded when no version is specified.

### DIFF
--- a/src/NuGetSdkResolver.UnitTests/NuGetSdkResolver_Tests.cs
+++ b/src/NuGetSdkResolver.UnitTests/NuGetSdkResolver_Tests.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using Microsoft.Build.Engine.UnitTests;
 using NuGet.Versioning;
 using Shouldly;
 using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using Xunit;
 
 using SdkResolverContextBase = Microsoft.Build.Framework.SdkResolverContext;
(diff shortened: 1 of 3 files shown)


jeffkl:
We found a corner case where ReSharper is loading NuGet assemblies of the same version but a different public key token.  When running under the debugger, I saw my `AssemblyResolve` event handler getting called a for every missing assembly after the NuGet SDK resolver is loaded.

I think we need to add something to DDRITs to ensure that the NuGet SDK resolver is working in VS.

src/NuGetSdkResolver.UnitTests/NuGetSdkResolver_Tests.cs
```diff
@@ -101,6 +104,38 @@ public void TryGetNuGetVersionNoVersionSpecified()
...
         }
 
+#if FEATURE_APPDOMAIN
+        /// <summary>
+        /// Verifies that when an SDK version is null that no NuGet assemblies are loaded.  This helps ensure that we're not loading
+        /// extra assemblies unless they are needed.  A lot of private classes exist in the NuGetSdkResolver in order to make sure
+        /// that types are loaded until they are needed.
+        /// </summary>
+        [Fact]
+        private void TryGetNuGetVersionNullVersionShouldNotLoadNuGetAssemblies()
```
cdmihai: Is it worth writing a test that uses reflection to peek at the AssemblyResolve handlers and assert their number does not change after a nuget resolver call with a non-null version? :)

PR #2935, opened 2018-01-31
Do not avoid best-match TF checks for ReferenceOutputAssembly=false

### LINKED ISSUE #2922: Error on GetNativeManifest when P2P to a multitargeting project with ReferenceOutputAssembly=false

rainersigwald:
This reverts commit 91c86a746b312fce1aba31f8fb8540e949c11a01.

Fixes #2922 but undoes the fix for #2661.

Details as to why in https://github.com/Microsoft/msbuild/pull/2923#issuecomment-361999701 and https://github.com/Microsoft/msbuild/pull/2867#issuecomment-361998280.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1545,15 +1545,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
       We give this treatment to .vcxproj by default since no .vcxproj can target more
       than one framework.
-
-      Likewise if the dependency is for build ordering instead of an assembly reference
-      (ReferenceOutputAssembly=false), skip the checks since we can't know what TF
-      the output would need to be compatible with.
    -->
    <ItemGroup>
-      <_MSBuildProjectReferenceExistent
-        Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and
-                   ('%(Extension)' == '.vcxproj' or '%(ReferenceOutputAssembly)' == 'false')">
+      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' == '' and '%(Extension)' == '.vcxproj'">
         <SkipGetTargetFrameworkProperties>true</SkipGetTargetFrameworkProperties>
       </_MSBuildProjectReferenceExistent>
    </ItemGroup>


rainersigwald:
@cdmihai You did this in master with afccbb11e11e9d60af474821a38882156a0354b4. Why was it needed there?

cdmihai:
@rainersigwald 
Because it caused our new build logic to fail due to these lines:
- https://github.com/Microsoft/msbuild/blob/master/src/MSBuild/MSBuild.csproj#L261
- https://github.com/Microsoft/msbuild/blob/0dcae6d2841e71d9ca1031c099c6883e64583afd/src/Build.UnitTests/Microsoft.Build.Engine.UnitTests.csproj#L23-L24
- https://github.com/Microsoft/msbuild/blob/0dcae6d2841e71d9ca1031c099c6883e64583afd/src/Tasks.UnitTests/Microsoft.Build.Tasks.UnitTests.csproj#L25

In this way:

```
Build FAILED.

E:\projects\msbuild_4\artifacts\.dotnet\2.1.1-preview-007216\sdk\2.1.1-preview-007216\Sdks\Microsoft.NET.Sdk\build\Microsoft.PackageDependencyResolution.targets(167,5): error : Assets file 'E:\projects\msbuild_4\artifacts\2\Debug\obj\Po
rtableTask\project.assets.json' doesn't have a target for '.NETCoreApp,Version=v2.0'. Ensure that restore has run and that you have included 'netcoreapp2.0' in the TargetFrameworks for your project. [E:\projects\msbuild_4\src\Samples\Po
rtableTask\PortableTask.csproj]
E:\projects\msbuild_4\artifacts\.dotnet\2.1.1-preview-007216\sdk\2.1.1-preview-007216\Sdks\Microsoft.NET.Sdk\build\Microsoft.PackageDependencyResolution.targets(167,5): error : Assets file 'E:\projects\msbuild_4\artifacts\2\Debug\obj\Po
rtableTask\project.assets.json' doesn't have a target for '.NETFramework,Version=v4.6'. Ensure that restore has run and that you have included 'net46' in the TargetFrameworks for your project. [E:\projects\msbuild_4\src\Samples\Portable
Task\PortableTask.csproj]
E:\projects\msbuild_4\artifacts\.dotnet\2.1.1-preview-007216\sdk\2.1.1-preview-007216\Sdks\Microsoft.NET.Sdk\build\Microsoft.PackageDependencyResolution.targets(167,5): error : Assets file 'E:\projects\msbuild_4\artifacts\2\Debug\obj\Ta
skWithDependency\project.assets.json' doesn't have a target for '.NETCoreApp,Version=v2.0'. Ensure that restore has run and that you have included 'netcoreapp2.0' in the TargetFrameworks for your project. [E:\projects\msbuild_4\src\Samp
les\TaskWithDependency\TaskWithDependency.csproj]
E:\projects\msbuild_4\artifacts\.dotnet\2.1.1-preview-007216\sdk\2.1.1-preview-007216\Sdks\Microsoft.NET.Sdk\build\Microsoft.PackageDependencyResolution.targets(167,5): error : Assets file 'E:\projects\msbuild_4\artifacts\2\Debug\obj\Nu
Get.MSBuildSdkResolver\project.assets.json' doesn't have a target for '.NETCoreApp,Version=v2.0'. Ensure that restore has run and that you have included 'netcoreapp2.0' in the TargetFrameworks for your project. [E:\projects\msbuild_4\sr
c\NuGetSdkResolver\NuGet.MSBuildSdkResolver.csproj]
    0 Warning(s)
    4 Error(s)
```

Our repo was using `ReferenceOutputAssembly="false"` with the intent of: I do not have a compile time references to these projects, but I want them in my output directory (in various subdirectories like "sdkresolvers") and in the output directories of everyone who references me. Also, they are managed projects and would want to build them with the closest TFM to my TFM.

PR #2943, opened 2018-02-01
Enable source linking

dsplaisted:
Add back source link packages we had before the toolset upgrade.

I'm not sure how to verify this is working, we may need to get a build after merging into the repo to do so.

@tmat Do you have a better way of doing this?

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -52,6 +52,10 @@
 
   <ItemGroup>
     <PackageReference Include="Nerdbank.GitVersioning" Version="$(GitVersioningVersion)" PrivateAssets="All" />
+    <PackageReference Include="PdbGit" Version="3.0.41" PrivateAssets="All"
+                      Condition="'$(DebugType)' == 'full' And '$(IsUnitTestProject)' != 'true'" />
+    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.7.6" PrivateAssets="All"
+                      Condition="'$(DebugType)' != 'full' And '$(IsUnitTestProject)' != 'true'" />    
   </ItemGroup>
   
   <PropertyGroup Condition="$(TargetFramework.StartsWith('net4'))">


AndyGerlicher:
This is on both Core and full framework?

dsplaisted:
@AndyGerlicher

> This is on both Core and full framework?

I think this is the same logic as before.  It uses one of two packages, depending on whether the `DebugType` is `full`.  It looks like the DebugType is full on full framework, and embedded on .NET Core and Mono.

rainersigwald:
Yeah, looks good to my recollection.

tmat:
This shouldn't be necessary. SourceLink is automatically generated by the RepoToolset.

tmat:
Re `DebugType`: please use the defaults that the toolset sets, that is:
1) On Jenkins it uses embedded PDBs. This makes debugging Jenkins crash dumps super easy.
2) On developer machine it currently[*] uses embedded PDBs. Makes it consistent with Jenkins. 
3) In official CI build it will build with portable PDBs and convert them to Windows PDBs for publishing into symbol store. Note that this requires changes in MicroBuild definition. If you point me to your official build definition I can make sure it's set up correctly.

[*] Unfortunately the support for portable/embedded PDBs on .NET Framework 4.7.1 was removed due to some bugs in the stack trace logic, so we might temporarily (until .NET FX 4.7.2 is broadly available) switch back to `full` so that xunit reported stack traces have source info in them. This will be done in the toolset, so no changes will be needed in the repos (other than updating to the latest toolset).

rainersigwald:
Let's tackle this the Arcade way after the switch.

tmat:
@rainersigwald There is really no reason to wait for the switch. It'd actually by better for msbuild repo to incrementally make these changes.

rainersigwald:
@tmat I'm very confused. You said earlier "don't do this". Now you want me to reopen it?

tmat:
I did not mean this exact change. I meant general support for SourceLink. The toolset adds SourceLink package automatically. So that should just work unless there is something overridden in the repo that would prevent it from working. If you checked that SourceLink indeed works for msbuild binaries then there is nothing more to do.

PR #2944, opened 2018-02-01
Use LargeAddressAware package instead of editbin.exe

jeffkl:
Package includes a managed assembly that sets the appropriate flag

https://github.com/KirillOsenkov/LargeAddressAware

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -45,7 +45,7 @@
 
   <!-- Production Dependencies -->
   <PropertyGroup>
-    <LargeAddressAwareVersion>1.0.1</LargeAddressAwareVersion>
+    <LargeAddressAwareVersion>1.0.2</LargeAddressAwareVersion>
   </PropertyGroup>
 
   <!-- Test Dependencies -->
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -45,6 +45,7 @@
 
   <!-- Production Dependencies -->
   <PropertyGroup>
+    <LargeAddressAwareVersion>1.0.1</LargeAddressAwareVersion>
   </PropertyGroup>
 
   <!-- Test Dependencies -->
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -41,7 +41,7 @@
 
     <GenApiVersion>2.1.0-prerelease-02404-02</GenApiVersion>
     <GitVersioningVersion>2.1.16</GitVersioningVersion>
-    <LargeAddressAwareVersion>1.0.2</LargeAddressAwareVersion>
+    <LargeAddressAwareVersion>1.0.3</LargeAddressAwareVersion>
   </PropertyGroup>
 
   <!-- Production Dependencies -->
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -41,11 +41,11 @@
 
     <GenApiVersion>2.1.0-prerelease-02404-02</GenApiVersion>
     <GitVersioningVersion>2.1.16</GitVersioningVersion>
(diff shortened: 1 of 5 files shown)


jeffkl:
I verified that the tool is running during build:
image

jeffkl:
Thanks @KirillOsenkov!

KirillOsenkov:
In case you need to check, you can verify if the binary got stamped correctly by running `dumpbin /headers msbuild.exe` where dumpbin is at:
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.12.25827\bin\Hostx86\x86\dumpbin.exe
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\dumpbin.exe
```

build/Versions.props
```diff
@@ -45,6 +45,7 @@
 
   <!-- Production Dependencies -->
   <PropertyGroup>
+    <LargeAddressAwareVersion>1.0.1</LargeAddressAwareVersion>
```
rainersigwald: What's the difference between "toolset" and "production"?

KirillOsenkov:
I'm also wondering if we should just set AnyCPU, 32-bit preferred on MSBuild.exe. This implies large address aware when run in 32-bit:
https://stackoverflow.com/a/35068310/37899

build/Versions.props
jeffkl: I figured Repo Toolset was sort of a brand name?  Should I put this in that section?

KirillOsenkov:
I published a new NuGet package with @jeffkl's changes:
https://www.nuget.org/packages/LargeAddressAware/1.0.2

build/Versions.props
rainersigwald: I have no idea; I was really asking :)

jeffkl:
@rainersigwald that's weird, I saw your commit in my fork too but not in this PR.  I did `git pull` and then `git push -f`

```
D:\msbuild>git pull
remote: Counting objects: 3, done.
remote: Total 3 (delta 2), reused 3 (delta 2), pack-reused 0
Unpacking objects: 100% (3/3), done.
From https://github.com/jeffkl/msbuild
   d3346621..0e41941e  LargeAddressAware -> origin/LargeAddressAware
Updating d3346621..0e41941e
Fast-forward
 README.md | 2 --
 1 file changed, 2 deletions(-)

D:\msbuild>git push -f
Everything up-to-date
```

And now the commit shows up in the PR?  That's fascinating...

src/MSBuild/MSBuild.csproj
```diff
@@ -217,6 +217,7 @@
     <Reference Include="System.Core" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Configuration" />
+    <PackageReference Include="LargeAddressAware" Version="$(LargeAddressAwareVersion)" PrivateAssets="All" />
```
cdmihai: Shouldn't PrivateAssets be set to `build` ? https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

As far as I understand from that documentation, only any of the `contentfiles;analyzers;build` values are consumed but do not flow to the parent projects

cdmihai: Nevermind, read it wrong, `All` is good :)

cdmihai:
Shouldn't this be turned on by `<LargeAddressAware>true</LargeAddressAware>`? 
https://github.com/KirillOsenkov/LargeAddressAware#largeaddressaware

jeffkl:

PR #2956, opened 2018-02-05
WIP: Ignore, trying some stuff

jeffkl:


src/Framework/XamlTypes/ContentType.cs
```diff
@@ -103,8 +103,8 @@ public string GetMetadata(string metadataName)
                 throw new ArgumentNullException("metadataName");
             }
 
-            string value;
-            _metadata.Value.TryGetValue(metadataName, out value);
+            _metadata.Value.TryGetValue(metadataName, out string value);
```
rainersigwald: Doesn't count until we have a Mono CI working again :(

PR #2960, opened 2018-02-06
WIP: Add extensibility points for source link

tmat:
Adds a few extensibility points to common targets that enable source control and SourceLink package features.

See https://github.com/tmat/repository-info/blob/master/docs/Readme.md for details.

TODO:
- [ ] Replace temp links to Standard CI spec with final links
- [ ] Include revision id in AssemblyInformationalVersion
- [ ] Docs


tmat:
@nguerrera

tmat:
@rainersigwald Please review.

Not sure why the tests are failing with
```
System.InvalidOperationException : Task attempted to log before it was initialized. Message was: The value of SourceRoot.ContainingRoot not found in SourceRoot items: 'c:\\MyProjects\\MyProject\\'
```

They work on my box. Is log initialization different on Linux/OSX?

tmat:
@AndyGerlicher This PR adds the minimal amount of extensibility points to the common targets that are needed for linking PDBs and packages to the originating source (via SourceLink, commit sha in AssemblyInformationalAttribute, Repository Url in nuget packages, etc.)

tmat:
@jinujoseph @richlander FYI

AndyGerlicher:
@tmat not sure if I'm just missing something at the moment, haven't tried it out. But why couldn't all this all come from a package reference?

tmat:
@AndyGerlicher  Because NuGet targets, dotnet SDK and the compiler targets will have a dependency on these properties and targets.

tmat:
@AndyGerlicher I did reduce the set of targets and properties that go to common targets to minimum. I don't think there is anything more that can be reasonably removed without breaking layering or introducing hacks.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,110 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+  -->
+
+  <Target Name="InitializeSourceRoot"
+          DependsOnTargets="$(InitializeSourceRootDependsOn)">
+    <ItemGroup>
+      <_InvalidSourceRoot Include="@(SourceRoot)" Condition="!HasTrailingSlash('%(SourceRoot.Identity)')" />
+    </ItemGroup>
+
+    <Error Text="SourceRoot paths are required to end with a slash or backslash: @(_InvalidSourceRoot, ', ')"
+           Condition="'@(_InvalidSourceRoot)' != ''" />
```
nguerrera: Why can't we just add them? That is what common targets normally does for this.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,110 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      
+    If the project references a package that provides source control information (such as Microsoft.Build.Tasks.Git, Microsoft.Build.Tasks.TFVC, etc.)
+    the revision id becomes available in local build environments as well.
+    
+    Extensibility: Redefine SetSourceRevisionId target set SetSourceRevisionId.
+  -->
+
+  <Target Name="SetSourceRevisionId">
+    <PropertyGroup>
+      <SourceRevisionId>$(STANDARD_CI_SOURCE_REVISION_ID)</SourceRevisionId>
```
nguerrera: Shouldn't this allow the user to override $(SourceRevisionId) just like $(PrivateRepositoryUrl) below?

src/Tasks/Microsoft.Common.props
```diff
@@ -11,6 +11,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- 
+      ContinuousIntegrationBuild is set to true when building on a CI server. The variable is then used to enable settings that only 
+      apply to official builds, as opposed to local builds on developer machine.
+      
+      Build servers that comply with Standard CI specification set STANDARD_CI_REPOSITORY_URL and STANDARD_CI_SOURCE_REVISION_ID variables.
+      See https://github.com/dotnet/designs/blob/86f4ed0e39fc1b1ab5c4128990b17c4aead4420f/proposed/standard-ci-env-variables.md.
+    -->
+    <ContinuousIntegrationBuild Condition="'$(DesignTimeBuild)' != 'true' and '$(STANDARD_CI_REPOSITORY_URL)' != ''">true</ContinuousIntegrationBuild>
```
nguerrera: We should check if ContinuousIntegrationBuild is already set, no?

src/Tasks/MapSourceRoots.cs
```diff
@@ -0,0 +1,113 @@
...
+        ///   <term>SourceControl</term><description>Indicates name of the source control system the source root is tracked by (e.g. Git, TFVC, etc.), if any.</description>
+        ///   <term>NestedRoot</term><description>If a value is specified the source root is nested (e.g. git submodule). The value is a path to this root relative to the containing root.</description>
+        ///   <term>ContainingRoot</term><description>Identifies another source root item that this source root is nested under.</description>
+        /// </list>
+        /// </summary>
+        [Required]
+        public ITaskItem[] SourceRoots { get; set; }
+
+        [Output]
+        public ITaskItem[] MappedSourceRoots { get; set; }
```
nguerrera: private set for Output

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: Do we know which separator to add on Windows (`\` or `/`)? The source root path needs to end with exactly the same separator that the full paths of source files that flow into the compiler use. 
Is it possible that the compiler will see a source path with `/` in it on Windows?

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: Ah, yes.

src/Tasks/Microsoft.Common.props
tmat: Yes.

src/Tasks/Microsoft.Common.CurrentVersion.targets
nguerrera: There's some magic that changes backslash to slash for ToolTask invocation on Unix

Take a deep breath:

https://github.com/Microsoft/msbuild/blob/cc9cc0c54454d427b12d73255b950908894ed10e/src/Utilities/ToolTask.cs#L1725

So I think appending \ will give the compiler \ on Windows and / on Unix.

nguerrera: I think it is possible for the compiler to see / on Windows, but not \ on Unix.

nguerrera: Why can't shouldn't the compiler understand that there's no difference between / and \ on Windows?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,110 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    Extensibility: Add target to InitializeSourceRootDependsOn that populates SourceRoot item group.
+  -->
+
+  <Target Name="InitializeSourceRoot"
+          DependsOnTargets="$(InitializeSourceRootDependsOn)">
+    <ItemGroup>
+      <_InvalidSourceRoot Include="@(SourceRoot)" Condition="!HasTrailingSlash('%(SourceRoot.Identity)')" />
+    </ItemGroup>
+
+    <Error Text="SourceRoot paths are required to end with a slash or backslash: @(_InvalidSourceRoot, ', ')"
```
nguerrera: Localization?

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: Source paths in the PDB are matched to the patterns in SourceLink by simple prefix match when interpreted by the debugger.

I don't think requiring the path to end with slash is an issue. SourceRoot items will usually be generated by another build task. Writing SourceRoots manually should be very rare.

nguerrera: I don't think it's much of an issue. I asked because the append-trailing-slash-if-not-already-there idiom is pervasive in common targets. If there's a good reason to do it differently here, that's fine. Might be worth a comment.

tmat: Will add a comment.

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: Fixed. Using the same pattern as `InitializePrivateRepositoryUrl`

src/Tasks/MapSourceRoots.cs
tmat: Fixed.

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: How do I localize an error message in .targets file?

cdmihai: Given that there's other error tasks in the file and I searched the visual studio drops for all occurrences of Microsoft.Common.CurrentVersion.targets but did not find anything, it's probably safe to say that, because it's always been not localized, it can remain not localized.

tmat:
test OSX10.13 Build for CoreCLR  please

tmat:
@rainersigwald All tests fixed. Thanks for help. Does the change look good now?

tmat:
@rainersigwald ping

tmat:
test Ubuntu16.04 Build for CoreCLR  please

rainersigwald:
To be clear, I'm thinking of a package like `SourceLink.Base` that:

* Delivers the `MapSourceRoots` task
* Defines a bunch of default properties in its .props
* Defines the targets here
* Is depended upon by specific plugins, like the Git provider

Someone could still override `InitializeSourceRevisionId` (define it in a NuGet package that depends on this package, to say nothing of adding to `$(InitializeSourceRevisionIdDependsOn)`) and PrivateRepository too.

tmat:
> I still don't understand why this can't be done in a package.

The properties and targets declared in this change are intended to be used in the following targets:

1) NuGet.Build.Tasks.Pack.targets
So that Pack task can embed revision id (e.g. commit SHA for git repos) into the package nuspec.

2) Microsoft.CppBuild.targets
So that C++ compiler can use SourceRoots to generate deterministic path map.

3) Microsoft.CSharp.Core.targets, Microsoft.VisualBasic.Core.targets, Microsoft.FSharp.Core.targets
So that C#, VB and F# compilers can use SourceRoots to generate deterministic path map.

4) Microsoft.NET.GenerateAssemblyInfo.targets
So that .NET SDK can include revision id in AssemblyInformationalVersion attribute.

If the targets added by this PR were in a package than all of the above targets would need to depend on that package somehow, which is impossible as far as I can tell.

tmat:
> package than all of the above targets would need to depend on that package somehow

I guess that's not strictly true. We could alternatively copy these targets into all of the above targets files and make a few copies of the `MapSourceRoots` task. That would probably work. But then what are common targets good for if not for sharing targets that are common?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,120 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    Standard CI specification: https://github.com/dotnet/designs/blob/86f4ed0e39fc1b1ab5c4128990b17c4aead4420f/proposed/standard-ci-env-variables.md.
+      
+    If the project references a package that provides source control information (such as Microsoft.Build.Tasks.Git, Microsoft.Build.Tasks.TFVC, etc.)
+    the revision id becomes available in local build environments as well.
+    
+    Extensibility: Add target to InitializeSourceRevisionIdDependsOn to initialize SourceRevisionId.
+  -->
+
+  <Target Name="InitializeSourceRevisionId"
+          DependsOnTargets="_InitializeSourceRevisionIdFromStandardCI;$(InitializeSourceRevisionIdDependsOn)">
```
eerhardt: Should `_InitializeSourceRevisionIdFromStandardCI` go after `$(InitializeSourceRevisionIdDependsOn)`?  That way other extensions can hook into this target, and they don't have to blindly overwrite the property?

As it is coded now, if an extension wants to write to `<SourceRevisionId>`, it can't check for empty, because `_InitializeSourceRevisionIdFromStandardCI` will set it first.

Same applies to `InitializePrivateRepositoryUrl` below.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,120 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+  <!--
+    ========================
+    DeterministicSourcePaths
+    ========================
+    
+    Unless specified otherwise enable deterministic source root when building deterministically on CI server, but not for local builds.
+    In order for the debugger to find source files when debugging a locally built binary the PDB must contain original, unmapped local paths.
+  -->
+  <PropertyGroup>
+    <DeterministicSourcePaths Condition="'$(DeterministicSourcePaths)' == '' and '$(Deterministic)' == 'true' and '$(ContinuousIntegrationBuild)' != ''">true</DeterministicSourcePaths>
```
eerhardt: `'$(ContinuousIntegrationBuild)' != ''` seems strange.  What if it is set to `false`?

It seems that the condition should be `'$(ContinuousIntegrationBuild)' == 'true'`.

tmat: Yes! Good catch.

rainersigwald:
> If the targets added by this PR were in a package than all of the above targets would need to depend on that package somehow, which is impossible as far as I can tell.

Why? Why do those target have direct dependencies on this? Can't they just use properties/items? If so, the properties/items can just not be set by default, and the other targets can treat that as "don't opt into this".

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: That depends on whether STANDARD_CI_SOURCE_REVISION_ID should be the default value that should be used if available or a fallback value that should be used if there is no extension setting SourceRevisionId, right?

tmat: The extension could decide to chain their target before `InitializeSourceRevisionIdFromStandardCI` target to initialize to a different default value like so:

```
<Target Name="MyInitializeSourceRevisionIdFromStandardCI"
              BeforeTargets="InitializeSourceRevisionIdFromStandardCI">
    <PropertyGroup>
      <SourceRevisionId Condition="'$(SourceRevisionId)' == ''">MyRevisionId</SourceRevisionId>
    </PropertyGroup>
</Target>
```

tmat:
@rainersigwald How will the targets above make sure the properties are initialized before their target run?

src/Tasks/Microsoft.Common.CurrentVersion.targets
eerhardt: Think about if you were writing an extension to this target and you wanted to set `<SourceRevisionId>`.  How would you do it?

Maybe instead of using a `Target` to set this property, you should just set the property in a normal `<PropertyGroup>`.  Then the typical MSBuild ordering applies.  An extension can set it either before OR after you.  Here they only get 1 option - after.

tmat:
The properties can't be initialized during evaluation phase. Tasks need to be called to retrieve the values.

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: I have a working extension that does it: https://github.com/tmat/repository-info/blob/master/src/Microsoft.Build.Tasks.SourceControl/build/Microsoft.Build.Tasks.SourceControl.targets#L62-L67

It can't be set in normal `<PropertyGroup>` since a task has to be called to retrieve this value.

eerhardt: Why have both extension mechanisms?  Why not just use `BeforeTargets` and `AfterTargets`?  And not have `InitializeSourceRevisionIdDependsOn` at all?

rainersigwald:
It would be the responsibility of the package that delivers these targets to ensure they happen at the right time by setting the right `BeforeTargets`. I suspect that `BeforeTargets="BeforeBuild"` would do the trick though you could also mention `CoreCompile` and friends to be sure.

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: `AfterTargets` doesn't work for the consumption. If I need to use this value in another target I add `InitializeSourceRevisionId` to my target's `DependsOnTargets`. `AfterTargets` of `InitializeSourceRevisionId` won't necessarily run.

tmat: > Why have both extension mechanisms?

"... depends on whether STANDARD_CI_SOURCE_REVISION_ID should be the default value that should be used if available or a fallback value that should be used if there is no extension setting SourceRevisionId"

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,120 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    -->
+    <Error Text="SourceRoot paths are required to end with a slash or backslash: @(_InvalidSourceRoot, ', ')"
+           Condition="'@(_InvalidSourceRoot)' != ''" />
+
+    <!--
+      Assign top-level source roots a deterministic path (MappedPath metadata).
+    -->
+
+    <ItemGroup Condition="'@(_MappedSourceRoot)' != ''">
+      <_MappedSourceRoot Remove="@(_MappedSourceRoot)" />
```
rainersigwald: Is this just defensive? That's not something we generally do.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,120 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    The format of the id is specific to the source control system the project is built from. A commit hash for GIT repositories, 
+    changeset number for TFVC repositories, etc. 
+   
+    Here, the id is set to the value supplied by the CI server building the project, provided that the server follows 
+    Standard CI specification: https://github.com/dotnet/designs/blob/86f4ed0e39fc1b1ab5c4128990b17c4aead4420f/proposed/standard-ci-env-variables.md.
+      
+    If the project references a package that provides source control information (such as Microsoft.Build.Tasks.Git, Microsoft.Build.Tasks.TFVC, etc.)
+    the revision id becomes available in local build environments as well.
+    
+    Extensibility: Add target to InitializeSourceRevisionIdDependsOn to initialize SourceRevisionId.
```
rainersigwald: Why not just override this target?

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: I guess we could also have `InitializeSourceRevisionIdDependsOn` initialized to `_InitializeSourceRevisionIdFromStandardCI` and then let extensions to prepend or append their target name to this property.

tmat: @rainersigwald Is there any guidelines for how to write these extension points?

src/Tasks/MapSourceRoots.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                }
+            }
+
+            if (success)
+            {
+                MappedSourceRoots = SourceRoots;
+            }
+
+            return success;
```
rainersigwald: In our experience, the best practice for returning from `Task.Execute` is

```c#
return !Log.HasLoggedErrors;
```

That helps avoid the confusing situation where you fail a task but have no error in the log about why.

tmat:
How should that package know what are all the consumers of these properties?

src/Tasks/MapSourceRoots.cs
```diff
@@ -0,0 +1,117 @@
...
+                                root.SetMetadata(Names.MappedPath, mappedPath);
+                                i++;
+                            }
+                        }
+                    }
+                }
+            }
+
+            string EndWithSlash(string path)
+                => (path[path.Length - 1] == '/') ? path : path + '/';
```
rainersigwald: If this stays here you should have access to `FileUtilities.EnsureTrailingSlash`.

src/Tasks/MapSourceRoots.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            // assign mapped paths to process source control roots first:
```
rainersigwald: What's the intention behind this ordering?

rainersigwald:
I'd be open to the idea of adding a new stub target in common targets to act as your hook point. Then any target (not Roslyn because of downlevel compat, but everybody else) could DependsOn it directly, and the package could inject targets before it. How's that sound?

tmat:
@rainersigwald What do you mean by a stub target? Isn't that what this change is adding?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,120 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      <_InvalidSourceRoot Include="@(SourceRoot)" Condition="!HasTrailingSlash('%(SourceRoot.Identity)')" />
+    </ItemGroup>
+
+    <!--
+      The SourceRoot is required to have a trailing directory separator.
+      We do not append one implicitly as we do not know which separator to append on Windows.
+      The usage of SourceRoot might be sensitive to what kind of separator is used (e.g. in SourceLink where it needs
+      to match the corresponding separators used in paths given to the compiler).
+    -->
+    <Error Text="SourceRoot paths are required to end with a slash or backslash: @(_InvalidSourceRoot, ', ')"
```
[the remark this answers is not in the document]
tmat: The task is only invoked in deterministic build. I think it's better do the validation for all builds.

To address the localization: is there a localized version of Error task for predefined resource strings?

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: Yes, it's defensive since `<Output>` with `ItemName` adds to the list. Shouldn't we do it more generally? What if someone has a temporary item group of the same name?

rainersigwald:
No, this PR adds real targets. I'm proposing a new target like

```xml
<Target Name="GetSourceMetadata" />
```

Then things could `DependsOnTargets="GetSourceMetadata"`, and `InitializePrivateRepositoryUrl` could `BeforeTargets="GetSourceMetadata"`.

[image](https://mermaidjs.github.io/mermaid-live-editor/#/edit/Z3JhcGggUkwKCkNvcmVDb21waWxlIC0tPnxEZXBlbmRzT258IEdldFNvdXJjZU1ldGFkYXRhWyJHZXRTb3VyY2VNZXRhZGF0YSA8YnIvPihDb21tb24udGFyZ2V0cykiXQpDbENvbXBpbGUgLS0-fERlcGVuZHNPbnwgR2V0U291cmNlTWV0YWRhdGEKUGFjay0tPnxEZXBlbmRzT258IEdldFNvdXJjZU1ldGFkYXRhCgpHZW5lcmF0ZUFzc2VtYmx5SW5mby0tPnxEZXBlbmRzT258IEdldFNvdXJjZU1ldGFkYXRhCkdldFNvdXJjZU1ldGFkYXRhLS0-fEJlZm9yZVRhcmdldHN8IEluaXRpYWxpemVQcml2YXRlUmVwb3NpdG9yeVVybApHZXRTb3VyY2VNZXRhZGF0YS0tPnxCZWZvcmVUYXJnZXRzfCBJbml0aWFsaXplU291cmNlUmV2aXNpb25JZAo)

src/Tasks/Microsoft.Common.CurrentVersion.targets
tmat: That's one way to do it. The disadvantage is that the extension can't fall back to the default (like a base call in overridden method).

tmat:
Sounds good for `PrivateRepositoryUrl` and `SourceRevisionId` and initialize the properties from STANDARD_CI_XXX values in evaluation phase. 

What about `InitializeSourceRoot`? Currently this target gathers the data from the extension target, validates that the gathered data are valid (paths ending with slash) and then calculates the deterministic path mapping, so that all consumers of InitializeSourceRoot have those mapped paths available.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: No, no formal guidelines that I'm aware of. I added some comments about the structure I'd expect, which I think is closer to what Eric expected too.

In general, I'd rather avoid creating new `xDependsOn` properties. They were required for MSBuild 2, but BeforeTargets/AfterTargets is a much better extensibility mechanism.

tmat: OK. It would be nice to have these guidelines written down somewhere, so we don't need to spend time figuring this out every time.

src/Tasks/Microsoft.Common.props
```diff
@@ -11,6 +11,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 -->
 
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- 
+      ContinuousIntegrationBuild is set to true when building on a CI server. The variable is then used to enable settings that only 
+      apply to official builds, as opposed to local builds on developer machine.
```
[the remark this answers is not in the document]
tmat: Have a better name? Many repos call this variable `CIBuild`. STANDARD_CI_XXX variables include the "CI" in their name and this variable is tied to those (when available). So I think that's consistent naming.

src/Tasks/MapSourceRoots.cs
tmat: OK.

src/Tasks/MapSourceRoots.cs
tmat: In most cases there will be a single SourceRoot that corresponds to the repository root, plus some additional SourceRoots for source packages and other external source roots.

The first SourceRoot in the item group is mapped to `/_/` path prefix. Since the mapped paths are visible to the user it's better if `/_/` prefix is commonly used for all the files in the repo (as opposed to `/_N/` prefix based on the order the targets added their SourceRoots).

I'll add a comment.

rainersigwald:
In that case I'd expect:

* This package provides the `InitializeSourceRoot` target
* The extension target is in another package that has a NuGet dependency on this one
* The extension target has `BeforeTargets="InitializeSourceRoot"`

[image](https://mermaidjs.github.io/mermaid-live-editor/#/edit/Z3JhcGggVEQKCnN1YmdyYXBoIENvbW1vbi50YXJnZXRzCkdldFNvdXJjZU1ldGFkYXRhWyJHZXRTb3VyY2VNZXRhZGF0YSA8YnIvPihDb21tb24udGFyZ2V0cykiXQplbmQKCkNvcmVDb21waWxlIC0tPnxEZXBlbmRzT258IEdldFNvdXJjZU1ldGFkYXRhWyJHZXRTb3VyY2VNZXRhZGF0YSA8YnIvPihDb21tb24udGFyZ2V0cykiXQpDbENvbXBpbGUgLS0-fERlcGVuZHNPbnwgR2V0U291cmNlTWV0YWRhdGEKUGFjay0tPnxEZXBlbmRzT258IEdldFNvdXJjZU1ldGFkYXRhCkdlbmVyYXRlQXNzZW1ibHlJbmZvLS0-fERlcGVuZHNPbnwgR2V0U291cmNlTWV0YWRhdGEKCnN1YmdyYXBoIEZhbmN5TmV3LkNvbW1vbi5udXBrZwpJbml0aWFsaXplUHJpdmF0ZVJlcG9zaXRvcnlVcmxbIkluaXRpYWxpemVQcml2YXRlUmVwb3NpdG9yeVVybCJdCkluaXRpYWxpemVTb3VyY2VSZXZpc2lvbklkWyJHZXRTb3VyY2VNZXRhZGF0YSJdCkluaXRpYWxpemVTb3VyY2VSb290WyJJbml0aWFsaXplU291cmNlUm9vdCJdCmVuZAoKCkdldFNvdXJjZU1ldGFkYXRhLS0-fEJlZm9yZVRhcmdldHN8IEluaXRpYWxpemVQcml2YXRlUmVwb3NpdG9yeVVybApHZXRTb3VyY2VNZXRhZGF0YS0tPnxCZWZvcmVUYXJnZXRzfCBJbml0aWFsaXplU291cmNlUmV2aXNpb25JZApHZXRTb3VyY2VNZXRhZGF0YS0tPnxCZWZvcmVUYXJnZXRzfCBJbml0aWFsaXplU291cmNlUm9vdAoKc3ViZ3JhcGggRXh0ZW5kZXJGb3IuRmFuY3lOZXcubnVwa2cKSW5pdGlhbGl6ZVNvdXJjZVJvb3QtLT58QmVmb3JlVGFyZ2V0c3wgRXh0ZW5kU291cmNlUm9vdAplbmQ)

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: By the way, this

> `AfterTargets` doesn't work for the consumption. If I need to use this value in another target I add `InitializeSourceRevisionId` to my target's `DependsOnTargets`. `AfterTargets` of `InitializeSourceRevisionId` won't necessarily run.

Isn't true. It should be (targets should make a DAG), but it's not. The details of the traversal are [https://docs.microsoft.com/en-us/visualstudio/msbuild/target-build-order#determining-the-target-build-order](documented), and I don't see any way to change them to a DAG approach without breaking the whole world, so they can be relied upon.

tmat:
InitializeSourceRoot needs to be called from C#, VB, F# and C++ targets. These targets consume the mapped paths (the output of the MapSourceRoot task).

rainersigwald:
I don't understand how that impacts the dependency graph above. Can you elaborate?

tmat:
Currently each compiler has their own build task assembly. So to remove the MapSourceRoot from common tasks we would need 4 copies of that task.

rainersigwald:
It would ship in what I'm calling `FancyNew.Common.nupkg`, as only one copy.

tmat:
Ok, so how do I add a FancyNew.Common.nupkg dependency to C++ targets?

rainersigwald:
You'd add it at the project level, like all NuGet references. Referencing a package would serve as the opt-in for this feature set.

For C++ specifically, this might require jumping through a couple of hoops until vcxproj is taught to handle PackageReference in the project, but I think it could be done with (shudder) an `install.ps1`.

tmat:
So all projects in the world will have this NuGet reference? It shouldn't be opt-in. The intent is to make it the default.

rainersigwald:
I thought that when we discussed this a while ago we all agreed that it might be default later but must be opt-in now.

tmat:
Not determinism. Deterministic build is on by default for .NET Core projects. We agreed on making SourceLink opt-in. This is not about SourceLink.

The SourceRoot item group is used for both deterministic build and for SourceLink.

PR #2962, opened 2018-02-07
Fix NuGet packages

### LINKED ISSUE #2951: NuGet packages have incorrect cross-reference versions
The NuGet packages that come out of `build.cmd -pack` incorrectly specify dependencies between versions to a static, incorrect version number `15.1.0-dev`:

image

Fortunately, this has already been fixed with https://github.com/NuGet/NuGet.Client/pull/1915 so we just need to update the version of the SDK we're using.

### LINKED ISSUE #2953: Microsoft.Build.Runtime package from 2.0 SDK
Currently, master doesn't produce the `Microsoft.Build.Runtime` package, but it's needed for API consumers as well as CLI.

### LINKED ISSUE #2954: NuGet packages with implementations
Currently, the NuGet packages built from master using `build -pack` contain only reference assemblies (all implementations are `nop ret`). They should contain the implementation assemblies as well, and the current reference assemblies should be labeled as such (they're currently in `lib/` instead of `ref/`).

jeffkl:
This is rebased on top of Daniel's change because it only works on a newer .NET SDK.  So some unit tests will fail until that goes in.

Here are the issues I found:

1. Documentation file was being written twice because the path was incorrect.  I was not able to figure out why so I just added a hack.
2. A few common package references were not set with `PrivateAssets="All"` so they were showing up as dependencies.
3. Package icon was no longer set
4. `DebugType` was not being set correctly on `Microsoft.Build.Engine` and `Microsoft.Build.Conversion.Core` so no symbols were being produced.
5. Package include satellite assemblies by default.  I couldn't find a good way to exclude them in Pack so I added a hack to do it.
6. Some references were not conditioned on `TargetFramework` so there were extra dependencies on the package.

Other improvements I've made:

1. Move reference assembly source back to `\ref' and no longer build them as projects.
  a. We can build ref assemblies with the compiler now.
2. Configured `Microsoft.Build.Framework` to use globbing and reduce its project XML
3. Exclude `*.log` and `*.binlog` so they don't show up in Solution Explorer

Fixes #2954
Fixes #2951
Fixes #2953


src/Build/Microsoft.Build.csproj
```diff
@@ -11,30 +11,32 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
 
     <GenerateReferenceAssemblySources>true</GenerateReferenceAssemblySources>
+    <IsPackable>true</IsPackable>
+    <Description>This package contains the $(MSBuildProjectName) assembly which is used to create, edit, and evaluation MSBuild projects.</Description>
```
rainersigwald: s/evaluation/evaluate/

src/Deprecated/Engine/Microsoft.Build.Engine.csproj
```diff
@@ -4,7 +4,22 @@
     <NoWarn>$(NoWarn);618;612</NoWarn>
     <GenerateReferenceAssemblySources>true</GenerateReferenceAssemblySources>
     <EnableDefaultItems>false</EnableDefaultItems>
+    <CreateTlb>true</CreateTlb>
```
rainersigwald: This doesn't actually do anything, does it? Just been carried around from all our old projects?

src/Directory.Build.targets
```diff
@@ -66,11 +67,11 @@
     <GenAPIAssemblyName Condition="'$(GenAPIAssemblyName)' == ''">$(MSBuildProjectName)</GenAPIAssemblyName>
     <GenAPIShortFrameworkIdentifier Condition="$(TargetFramework.StartsWith('net4'))">net</GenAPIShortFrameworkIdentifier>
     <GenAPIShortFrameworkIdentifier Condition="$(TargetFramework.StartsWith('netstandard'))">netstandard</GenAPIShortFrameworkIdentifier>
-    <GenAPITargetPath>$(RepoRoot)src\Package\ReferenceAssemblies\$(GenAPIAssemblyName)\$(GenAPIShortFrameworkIdentifier)\$(GenAPIAssemblyName).cs</GenAPITargetPath>
+    <GenAPITargetPath>$(RepoRoot)ref\$(GenAPIAssemblyName)\$(GenAPIShortFrameworkIdentifier)\$(GenAPIAssemblyName).cs</GenAPITargetPath>
```
rainersigwald: I don't like having `ref/` as a top-level folder. Can it be under `src/`? It doesn't seem important enough to junk up the root and the main GitHub page.

src/Directory.Build.targets
```diff
@@ -92,6 +93,19 @@
        in VS 15.5 (which is what we use for Jenkins right now). -->
   <Target Name="GenerateSatelliteAssemblies" />
 
+  <Target Name="RemoveSatelliteDllsFromBuildOutputInPackage"
+          BeforeTargets="GenerateNuspec"
+          AfterTargets="_CalculateInputsOutputsForPack"
```
rainersigwald: `After`, or `DependsOn`? If really after, please comment why because that's a confusing relationship.

src/Utilities/Microsoft.Build.Utilities.csproj
```diff
@@ -1,201 +1,187 @@
...
-    <PackageReference Include="System.Text.Encoding.CodePages" Version="4.4.0" />
+    
     <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETStandard'">
+    <PackageReference Include="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="1.2.304-preview5" Condition="'$(TargetFrameworkIdentifier)' != '.NETStandard'" />
+
+    <Reference Include="System.Configuration" Condition="'$(TargetFrameworkIdentifier)' != '.NETStandard'" />
```
rainersigwald: Redundant conditions

src/Deprecated/Engine/Microsoft.Build.Engine.csproj
jeffkl: I don't know, I just moved the properties up from the bottom so they were set in one place

src/Directory.Build.targets
jeffkl: I wanted this to trigger before GenerateNuspec which has a `DependsOn` that ends with `_CalculateInputsOutputsForPack`.  What's the right way to say "run me before this target but after its dependson?

rainersigwald: Technically, just BeforeTargets does that, because rule 5 is after rule 4: https://docs.microsoft.com/en-us/visualstudio/msbuild/target-build-order#determining-the-target-build-order

I'd say to also say `DependsOnTargets="$(GenerateNuspecDependsOn)"` but the fact that it adds to it in its dependencies but not in its properties makes that a bit annoying.

jeffkl:

Issue #2022, opened 2017-04-28
Improper handling of slashes in solution files


Sumo-MBryant:
If a solution file contains a relative path to a project with a backslash (which Visual Studio and `dotnet sln add` both emit), and the project uses a project type GUID that is not recognised by MSBuild, any attempts to build the solution on Linux fail with the error:

```
{path_to_project}.metaproj : error MSB4025: The project file could not be loaded. Could not find file '{path_to_project}.metaproj'.
```

It appears that the code in `Microsoft.Build.Construction.ProjectInSolution.CanBeMSBuildProjectFile()` is failing on Linux.

# Reproduction Steps:

1. Create a solution file named *solution.sln* with the following contents:

```
Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio 15
VisualStudioVersion = 15.0.26403.7
MinimumVisualStudioVersion = 10.0.40219.1
Project("AAAAAAAA-1760-4266-BCC7-CA923CBCF16C") = "project", "directory\project.proj", "{1FD5F8F7-D4D3-4760-B503-03883D7CDFD7}"
EndProject
Global
	GlobalSection(SolutionConfigurationPlatforms) = preSolution
		Debug|Any CPU = Debug|Any CPU
	EndGlobalSection
	GlobalSection(ProjectConfigurationPlatforms) = postSolution
		{1FD5F8F7-D4D3-4760-B503-03883D7CDFD7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
		{1FD5F8F7-D4D3-4760-B503-03883D7CDFD7}.Debug|Any CPU.Build.0 = Debug|Any CPU
	EndGlobalSection
EndGlobal
```

2. Create a project file *project.proj* in a sub directory named *directory* with the following contents:

```xml
<Project>
  <PropertyGroup>
    <DefaultProjectTypeGuid>AAAAAAAA-1760-4266-BCC7-CA923CBCF16C</DefaultProjectTypeGuid>
  </PropertyGroup>

  <Target Name="Build">
    <Message Text="Build target for $(MSBuildProjectFile)" Importance="high" />
  </Target>
</Project>
```

3. On Linux, run `dotnet msbuild solution.sln`

## Expected Output:

```
  Build target for project.proj
```

## Actual Output:

```
project.proj.metaproj : error MSB4025: The project file could not be loaded. Could not find file '/homeproject.proj.metaproj'.
```

# Workaround

[...]

PR #2963, opened 2018-02-07
Support backslashes in solution paths

rainersigwald:
Add a test that will fail on forward-slash OSes when reading a
backslash-containing relative path from a solution file.

Regression test for #2022.


rainersigwald:
@radical Do you mind me taking the core of your change from mono@86adc96 like this? I was hoping to be able to just cherry-pick or merge it but it has mono-only tests that would make that inconvenient.

radical:
@rainersigwald go ahead!

PR #2964, opened 2018-02-07
Fix node task crash in SDK resolution which hangs MSBuild.exe

jeffkl:
In a multi-proc build when an SDK could not be found, a NullReferenceException would occur which would crash the thread on the main node responsible for responding to SDK resolution requests.  The out-of-proc node would then hang waiting for a response and MSBuild's main thread would be waiting for the out-of-proc node to finish.

The code is more resilient now by catching all exceptions and sending the response in a finally.  Any errors that occur in the request handler thread are sent back to the out-of-proc node to be logged.

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/MainNodeSdkResolverService.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/MainNodeSdkResolverService.cs
@@ -239,7 +239,10 @@ namespace Microsoft.Build.BackEnd.SdkResolution
                     catch (Exception e)
                     {
                         ILoggingService loggingService = Host.GetComponent(BuildComponentType.LoggingService) as ILoggingService;
-                        loggingService.LogFatalError(new EvaluationLoggingContext(loggingService, request.BuildEventContext, request.ProjectPath).BuildEventContext, e, new BuildEventFileInfo(request.ElementLocation), "CouldNotRunSdkResolver", "foo", e.Message);
+
+                        EvaluationLoggingContext loggingContext = new EvaluationLoggingContext(loggingService, request.BuildEventContext, request.ProjectPath);
+
+                        loggingService.LogFatalBuildError(loggingContext.BuildEventContext, e, new BuildEventFileInfo(request.ElementLocation));
                     }
                     finally
                     {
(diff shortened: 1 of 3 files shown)


jeffkl:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/563880

src/Build/BackEnd/Components/SdkResolution/MainNodeSdkResolverService.cs
```diff
@@ -222,21 +222,33 @@ private void ProcessRequests()
...
+                        // node to be logged.
+                        response = new SdkResolverResponse(e);
+                    }
+                    finally
+                    {
+                        // Get the node manager and send the response back to the node that requested the SDK
+                        INodeManager nodeManager = Host.GetComponent(BuildComponentType.NodeManager) as INodeManager;
 
-                    nodeManager.SendData(request.NodeId, response);
+                        nodeManager.SendData(request.NodeId, response);
```
cdmihai: So if sdk resolution errors, a packet is sent to the node, the node logs an error, which travels back as a packet to the BuildManager, 
 then the BuildManager sends a shutdown packet back to the node. I wonder if there's enough API head room for the BuildManager to shortcuit this and just fail the build directly. On the other hand I don't think this sdk failure communication happens too often, so probably it's not worth optimizing.

src/Build/BackEnd/Components/SdkResolution/OutOfProcNodeSdkResolverService.cs
```diff
@@ -68,7 +68,15 @@ public override string ResolveSdk(int submissionId, SdkReference sdk, LoggingCon
...
-            if (!SdkResolverService.IsReferenceSameVersion(sdk, response.Version))
+            if (!String.IsNullOrWhiteSpace(response.Error))
+            {
+                // Log an error that occured on the main node request handler thread which cannot log errors
+                loggingContext.LogErrorFromText(null, null, null, new BuildEventFileInfo(sdkReferenceLocation), response.Error);
+
+                return null;
+            }
+
+            if (response.Version != null && !SdkResolverService.IsReferenceSameVersion(sdk, response.Version))
```
cdmihai: Why would the version be null here?

PR #2972, opened 2018-02-08
Machine independent toolset

mfilippov:
Configuration MSBuild for portable (xcopy-install) toolsets: works on WinNT and linux/mac via Mono, isolates from machine environment: uses only tools installed with it, ignores Registry and GAC and Visual Studio installations to provide the same build experience on all machines


Directory.Build.props
```diff
@@ -66,4 +66,12 @@
     <DebugType Condition="'$(DebugType)' == ''">embedded</DebugType>
     <MonoBuild>true</MonoBuild>
   </PropertyGroup>
+  <!-- Configuration MSBuild for portable (xcopy-install) toolsets: works on WinNT and linux/mac via Mono, isolates from machine environment:
+    uses only tools installed with it, ignores Registry and GAC and Visual Studio installations to provide the same build experience on all machines -->
+  <PropertyGroup Condition="'$(Configuration)' == 'MachineIndependent'">
```
cdmihai: Rather than making the configuration space more complicated, what if you introduce a simple property called `MachineIndependent`? That way you can have Release / Debug machine independent.

mfilippov: I'm not sure about it because separate configuration allows edit project files in IDE with correctly #ifdef states. An example I found problems in compilation some registry dependent tests when I disable registry.

cdmihai: Good point, didn't think about the IDE scenario :). New config it is then.

AndyGerlicher:
I like this in general as a distribution of MSBuild. This is more similar to the .NET Core version, but likely more features. The one thing I wonder here though is if a new compilation target is the right approach? I supposed we can do this for now. But ideally this is the same code and these are run-time flags. Maybe a different `MSBuild.exe` entry-point that targets locked-down features? I realize that this is likely beyond the scope of what you're intending here, just looking for discussion and hopefully we can converge on something that is easier to maintain. And by that I mean, for example, that today we have at least VS MSBuild, .NET CLI MSBuild, Mono MSBuild, OmniSharp (or is that just Mono? @DustinCampbell ), and internally CoreXT (which is proprietary, but you can see some traces here). If we add to this list officially it should be for good reason. And then should we go further? We've had a concept of an MSBuild strict language in the past that would not let you do stuff like GAC/Registry to make it more deterministic across machines. With the ecosystem where it is (NuGet/packaging), this actually seems like a pretty good time to go further in that direction.

Also it would be good to know how this is intended to be used. Should we output this to zip file in blob storage or something? Or are you planning to rebuild and resign this anyway?

mfilippov:
@AndyGerlicher You right,  runtime flags approach is better than different compilation target. But right now MSBuild has a lot #ifdef section, and it isn't possible. We created this configuration when building cross-platform toolset independent from machine environment and use it now on our CI and developers machines with macOS, Linux or Windows. If you published artifacts as the zip archive, it would perfect, we can use it as input for our internal toolset build process.

mfilippov:
@AndyGerlicher I rebase changes to resolve the conflict.

src/Directory.BeforeCommon.targets
```diff
@@ -87,7 +87,7 @@
     <DefineConstants>$(DefineConstants);FEATURE_THREAD_CULTURE</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_THREAD_PRIORITY</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_MULTIPLE_TOOLSETS</DefineConstants>
-    <DefineConstants>$(DefineConstants);FEATURE_NODE_REUSE</DefineConstants>
+    <DefineConstants Condition="'$(MachineIndependentBuild)' != 'true'">$(DefineConstants);FEATURE_NODE_REUSE</DefineConstants>
```
rainersigwald: What's the motivation for this? Can we not support node reuse in isolation?

mfilippov: We disable it because we have a crashes with node reuse enabled on Mono.

Issue #2974, opened 2018-02-08
MSBuild leaks handles when creating worker nodes


rainersigwald:
https://twitter.com/BruceDawson0xB/status/961059017060622336

https://developercommunity.visualstudio.com/content/problem/195192/process-handle-leak-when-launching-msbuild.html

Something leaks handles to MSBuild.exe worker nodes.

Repro outside VS just using the API:

```c#
static void Main(string[] args)
{
    var instances = MSBuildLocator.QueryVisualStudioInstances().ToList();

    MSBuildLocator.RegisterInstance(instances[0]);

    for (int i = 0; i < 99; i++)
    {
        BuildProject();
    }
}

private static void BuildProject()
{
    var p = new BuildParameters
    {
        EnableNodeReuse = false,
        DisableInProcNode = true,
        MaxNodeCount = 99,
    };

    var req = new BuildRequestData(@"s:\work\AFewProjectsForNodeReuse\AFewProjectsForNodeReuse\AFewProjectsForNodeReuse.csproj", new Dictionary<string, string>(), null, new[] { "Build" }, null, BuildRequestDataFlags.None);

    BuildManager.DefaultBuildManager.Build(p, req);
}
```

Thanks @randomascii!

rainersigwald:
The problem is here:

https://github.com/Microsoft/msbuild/blob/5ec4c4cc63d844af5490690bdb64560a055c43e8/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs#L527-L559

`CreateProcess` [documents ](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx) that

> Handles in PROCESS_INFORMATION must be closed with CloseHandle when they are no longer needed.

but we're not.

PR #2976, opened 2018-02-08
Check exact match for "running msbuild.exe"

### LINKED ISSUE #2194: Unable to find 15.0 toolset in a project with MSBuild in it's name.
Repro steps:

- Create a project called "MSBuildTest" (or something with msbuild in the name) and reference the Microsoft.Build package.
- Add the following code 

```csharp
        {
            AppDomain.CurrentDomain.AssemblyResolve += (sender, eventArgs) =>
            {
                string msbuildPath = @"C:\Program Files (x86)\Microsoft Visual Studio\D15PreRel\MSBuild\15.0\Bin\";
                var targetAssembly = Path.Combine(msbuildPath, new AssemblyName(eventArgs.Name).Name + ".dll");
                return File.Exists(targetAssembly) ? Assembly.LoadFrom(targetAssembly) : null;
            };
            Run();
        }

        private static void Run()
        {
            ProjectCollection collection = new ProjectCollection();
            var project = collection.LoadProject(@"ConsoleApp.csproj");
        }
```
- Make sure there's a ConsoleApp.csproj which has ToolsVersion="15.0" in it.

Expected: The project should loaded fine.

Found: Throws 
```
Microsoft.Build.Exceptions.InvalidProjectFileException: 'The tools version "15.0" is unrecognized. Available tools versions are "2.0", "3.5", "4.0".  ConsoleApp.csproj'
```

This is because BuildEnvironmentHelper tries to ascertain the mode in which msbuild is launched and one of the checks it does is to look at the process name and sees if it contains “MSBuild”. In this case, it think the BuildMode is Standalone and doesn't read msbuild.exe.config and hence doesn't get the 15.0 toolset.

rainersigwald:
Fixes the problem where a program using the MSBuild API and
containing the string "MSBuild" in its name was incorrectly detected as
running in MSBuild.exe in standalone mode.

Fixes #2194.

Note that the concerns in https://github.com/Microsoft/msbuild/issues/2194#issuecomment-307151488 about test detection no longer applies after b8d01363.

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -327,11 +327,15 @@ namespace Microsoft.Build.Shared
         /// <returns></returns>
         private static bool IsProcessInList(string processName, string[] processList)
         {
-            return
-                processList.Any(
-                    s =>
-                        Path.GetFileNameWithoutExtension(processName)?
-                            .IndexOf(s, StringComparison.OrdinalIgnoreCase) >= 0);
+            var processFileName = Path.GetFileNameWithoutExtension(processName);
+
+            if (string.IsNullOrEmpty(processFileName))
+            {
+                return false;
+            }
+
+            return processList.Any(s =>
+                processFileName.Equals(s, StringComparison.OrdinalIgnoreCase));
         }
 
         private static string GetProcessFromRunningProcess()


rainersigwald:
Hit this again testing #2974 and accidentally nuked my whole machine by infinitely spawning my test program instead of `MSBuild.exe`. Way past time to fix this annoyance.

PR #2978, opened 2018-02-08
Reset current directory on Core

### LINKED ISSUE #2977: MSBuild/Core holds a handle to project directories after build completion
Discovered by @mikeharder and reported offline. Thanks!

> It appears that each of the “dotnet.exe MSBuild.dll” processes keeps a file handle open to the last directory it built, which prevents my automation from deleting the temp directory after the build:
> handles
> I can workaround this by killing all the dotnet.exe processes, but I don’t hit this issue with MSBuild on Framework plus NodeReuse, so I wanted to let you know in case it’s a product issue.

rainersigwald:
Full-framework MSBuild was resetting the current directory at the end of build requests and at node idle, which helps avoid holding handles to project directories that the user may wish to delete while a worker node is still alive in the background. That wasn't done on .NET Core because at the time of the original port, there wasn't an `Environment.SystemDirectory` to use.

That property is now available, but returns the empty string on Linux, which fails `Directory.SetCurrentDirectory()`. So I'm avoiding it by using the current MSBuild tools directory, which should exist as long as the MSBuild process keeps running.

Fixes #2977.

### DIFF
--- a/src/Build/BackEnd/Node/OutOfProcNode.cs
+++ b/src/Build/BackEnd/Node/OutOfProcNode.cs
@@ -503,10 +503,9 @@ namespace Microsoft.Build.Execution
             // Shutdown any Out Of Proc Nodes Created
             _taskHostNodeManager.ShutdownConnectedNodes(_shutdownReason == NodeEngineShutdownReason.BuildCompleteReuse);
 
-#if FEATURE_ENVIRONMENT_SYSTEMDIRECTORY
-            // Restore the original current directory.
-            NativeMethodsShared.SetCurrentDirectory(Environment.SystemDirectory);
-#endif
+            // On Windows, a process holds a handle to the current directory,
+            // so reset it away from a user-requested folder that may get deleted.
+            NativeMethodsShared.SetCurrentDirectory(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory);
 
             // Restore the original environment.
             // If the node was never configured, this will be null.
@@ -710,10 +709,8 @@ namespace Microsoft.Build.Execution
             }
             catch (DirectoryNotFoundException)
             {
-#if FEATURE_ENVIRONMENT_SYSTEMDIRECTORY
                 // Somehow the startup directory vanished. This can happen if build was started from a USB Key and it was removed.
                 NativeMethodsShared.SetCurrentDirectory(Environment.SystemDirectory);
-#endif
             }
 
             // Replicate the environment.  First, unset any environment variables set by the previous configuration.
--- a/src/Build/BackEnd/Node/OutOfProcNode.cs
+++ b/src/Build/BackEnd/Node/OutOfProcNode.cs
@@ -710,7 +710,7 @@ namespace Microsoft.Build.Execution
             catch (DirectoryNotFoundException)
             {
                 // Somehow the startup directory vanished. This can happen if build was started from a USB Key and it was removed.
-                NativeMethodsShared.SetCurrentDirectory(Environment.SystemDirectory);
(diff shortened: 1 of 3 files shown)


src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -503,10 +503,9 @@ private NodeEngineShutdownReason HandleShutdown(out Exception exception)
             // Shutdown any Out Of Proc Nodes Created
             _taskHostNodeManager.ShutdownConnectedNodes(_shutdownReason == NodeEngineShutdownReason.BuildCompleteReuse);
 
-#if FEATURE_ENVIRONMENT_SYSTEMDIRECTORY
-            // Restore the original current directory.
-            NativeMethodsShared.SetCurrentDirectory(Environment.SystemDirectory);
-#endif
+            // On Windows, a process holds a handle to the current directory,
+            // so reset it away from a user-requested folder that may get deleted.
+            NativeMethodsShared.SetCurrentDirectory(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory);
```
rainersigwald: Please check my logic about the new default value for this!

I can also do an if-windows conditional.

rainersigwald:
macOS: Possible issue connecting to GitHub or cloning the repository. This could be caused by a lack of network connectivity or a lack of disk space. Please email netciadmin.

rainersigwald:
The feature flag is sadly still used, in a confusing place for bad reasons.

https://github.com/Microsoft/msbuild/blob/0dcae6d2841e71d9ca1031c099c6883e64583afd/src/Shared/UnitTests/FileUtilities_Tests.cs#L558-L746

Those tests still need to be disabled, but I couldn't come up with a better name for the flag that describes how they're broken. Totally open to suggestions!

jeffkl:
Hmm, perhaps the tests should just not use `Environment.SystemDirectory`?  it looks like they could use any directory

src/Build/BackEnd/Node/OutOfProcNode.cs
cdmihai: Can setting cwd to a bogus value have any external behaviour? Files written, node failure logs, etc? If yes, do an if-windows conditional. If no, it doesn't really matter.

src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -710,10 +709,8 @@ private void HandleNodeConfiguration(NodeConfiguration configuration)
             }
             catch (DirectoryNotFoundException)
             {
-#if FEATURE_ENVIRONMENT_SYSTEMDIRECTORY
                 // Somehow the startup directory vanished. This can happen if build was started from a USB Key and it was removed.
                 NativeMethodsShared.SetCurrentDirectory(Environment.SystemDirectory);
-#endif
```
cdmihai: Why the inconsistency with the other changes? Not that it matters, right?

rainersigwald:
@dotnet-bot test this please

(don't trust stale results from February, but want this in)

src/Build/BackEnd/Node/OutOfProcNode.cs
rainersigwald: AFAIK none of that depends on CWD . . .

src/Build/BackEnd/Node/OutOfProcNode.cs
rainersigwald: No good reason.

rainersigwald:

An odd hang:

```
MSBuild Failure: From calculating based on the memorystream, about to read n = -21. length = 135784, rawPosition = 135805, readLength = 21, stringLength = 108, currPos = 87.
Waiting for debugger to attach to process: 75712
```

PR #2985, opened 2018-02-12
Upgrade to netcoreapp2.1

AndyGerlicher:
This resolves issues debugging when using types that are available only in 2.1 runtime. Regardless of this change, CLI will be on 2.1 so this change should be safe.


AndyGerlicher:
Note this does not move our DLLs to netstandard2.1, just our entry-point.

PR #2991, opened 2018-02-13
Do not throw an InvalidProjectFileException if ProjectLoadSettings.IgnoreMissingImports is specified

jeffkl:
InvalidProjectFileExceptions cause modal dialogs in Visual Studio which so we should respect `ProjectLoadSettings`.

Related to https://github.com/dotnet/project-system/issues/3240

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -2499,6 +2499,28 @@ namespace Microsoft.Build.Evaluation
 
                 if (string.IsNullOrEmpty(sdkRootPath))
                 {
+                    if (_loadSettings.HasFlag(ProjectLoadSettings.IgnoreMissingImports))
+                    {
+                        ProjectImportedEventArgs eventArgs = new ProjectImportedEventArgs(
+                            importElement.Location.Line,
+                            importElement.Location.Column,
+                            ResourceUtilities.GetResourceString("CouldNotResolveSdk"),
+                            importElement.ParsedSdkReference.ToString())
+                        {
+                            BuildEventContext = _evaluationLoggingContext.BuildEventContext,
+                            UnexpandedProject = importElement.Project,
+                            ProjectFile = importElement.ContainingProject.FullPath,
+                            ImportedProjectFile = null,
+                            ImportIgnored = true,
+                        };
+
+                        _evaluationLoggingContext.LogBuildEvent(eventArgs);
+
+                        projects = new List<ProjectRootElement>();
+
+                        return;
+                    }
+
                     ProjectErrorUtilities.ThrowInvalidProject(importElement.SdkLocation, "CouldNotResolveSdk", importElement.ParsedSdkReference.ToString());
                 }
 
--- a/src/Build.OM.UnitTests/Construction/ProjectSdkImplicitImport_Tests.cs
+++ b/src/Build.OM.UnitTests/Construction/ProjectSdkImplicitImport_Tests.cs
@@ -448,15 +448,32 @@ namespace Microsoft.Build.UnitTests.OM.Construction
   </PropertyGroup>
 </Project>
 ";
+            MockLogger logger = new MockLogger();
+            ProjectCollection projectCollection = new ProjectCollection();
+            projectCollection.RegisterLogger(logger);
+
             ProjectRootElement rootElement = ObjectModelHelpers.CreateInMemoryProjectRootElement(projectContents);
 
             Project project = new Project(rootElement,
                 globalProperties: null,
                 toolsVersion: null,
-                projectCollection: new ProjectCollection(),
+                projectCollection: projectCollection,
                 loadSettings: ProjectLoadSettings.IgnoreMissingImports);
 
             project.GetPropertyValue("Success").ShouldBe("true");
+            
+            ProjectImportedEventArgs[] events = logger.BuildMessageEvents.OfType<ProjectImportedEventArgs>().ToArray();
+
+            // There are two implicit imports so there should be two logged ProjectImportedEventArgs
+            events.Length.ShouldBe(2);
+
+            events[0].Message.ShouldStartWith("MSB4236");
+            events[0].ImportIgnored.ShouldBeTrue();
+            events[0].ImportedProjectFile.ShouldBeNull();
+
+            events[1].Message.ShouldStartWith("MSB4236");
+            events[1].ImportIgnored.ShouldBeTrue();
+            events[1].ImportedProjectFile.ShouldBeNull();
         }
 
         public void Dispose()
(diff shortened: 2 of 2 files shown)


jeffkl:
fyi @jviau

jeffkl:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems/edit/567235

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2499,6 +2499,28 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                            BuildEventContext = _evaluationLoggingContext.BuildEventContext,
+                            UnexpandedProject = importElement.Project,
+                            ProjectFile = importElement.ContainingProject.FullPath,
+                            ImportedProjectFile = null,
+                            ImportIgnored = true,
+                        };
+
+                        _evaluationLoggingContext.LogBuildEvent(eventArgs);
+
+                        projects = new List<ProjectRootElement>();
```
rainersigwald: This line confuses me, but seems correct from context.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2499,6 +2499,28 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                        ProjectImportedEventArgs eventArgs = new ProjectImportedEventArgs(
+                            importElement.Location.Line,
+                            importElement.Location.Column,
+                            ResourceUtilities.GetResourceString("CouldNotResolveSdk"),
+                            importElement.ParsedSdkReference.ToString())
+                        {
+                            BuildEventContext = _evaluationLoggingContext.BuildEventContext,
+                            UnexpandedProject = importElement.Project,
+                            ProjectFile = importElement.ContainingProject.FullPath,
+                            ImportedProjectFile = null,
```
jviau: Currently I use `ImportedProjectFile` to track imports that were ignored, so I can still attempt reloads as they change (in-case they become valid). Can we get 2 separate events, one for sdk.props and one for sdk.targets?

jeffkl: The event will be logged twice but since no project was imported then `ImportedProjectFile` is null.  `UnexandedProject` is `Sdk.props` and `Sdk.targets`.  Will that work?

jeffkl:

src/Build/Evaluation/Evaluator.cs
jeffkl: Yeah, `projects` is an `out` parameter so it must be set before returning...

src/Build/Evaluation/Evaluator.cs
jeffkl: Tehavior we saw when trying things out was that Visual Studio unloads a project and when you try to reload it you get an error.  An SDK not found with `IgnoreMissingImports` would evaluate to this:

```xml
<Project>
  
  <PropertyGroup>
    <TargetFramework>netstandard1.3</TargetFramework>
  </PropertyGroup>

</Project>
```

And you get this when you save it, VS unloads it, and you try to reload it.

image

Are you saying you would leave the project loaded and then attempt to reload it if the user fixed it?

jviau: When I added `IgnoreInvalidImport` I refined the contract for `ImportedProjectFile`. It should only be null if the condition is false or it is a glob with no matches. This is a case I didn't consider - import is ignored, but the full path is still unknown (because any resolver could eventually supply it). I will have to make a change to CPS to handle this.

Unless you want to set the full path was what the default SDK resolver would supply - the SDK directory under msbuild. Not sure if that is the correct behavior though.

jeffkl: The default SDK resolver only returns the path if it exists so the `Evaluator` would not know the value.  The only options are:

1. Use the value in the Project attribute, `Sdk.props` and `Sdk.targets`
2. Use `null` since a path could not be determined

jeffkl:
Hmm Jenkins seems to be using the netci.groovy from master instead of my branch so the build on Mac is failing.

rainersigwald:
I don't understand why it's running both flavors of macOS CI job. But since the correct one passed I'm ok with moving forward (pending shiproom approval).

src/Build/Evaluation/Evaluator.cs
jeffkl: @jviau what's your decision on the options?

jviau: Trying to understand your comment - above. With this change, when you tried to load a project with a broken SDK, that dialog popped up instead?

To answer your question, CPS will try to watch your invalid imports and reload them and reevaluate when they change. In this case though, I agree with you in that the path will have to be `null`. I don't think we will be able to gracefully handle SDK being changed, may have to force a full reload when that happens.

jeffkl: > With this change, when you tried to load a project with a broken SDK, that dialog popped up instead?

With my change, I simulated what would happen if Visual Studio ignored the invalid import.  What happens is Visual Studio unloads the project and when you attempt to reload the project you get the dialog above.

jviau: Hmm okay. I wonder if this now meets the bar for 15.6 since we will have to make a considerable change to CPS as well.

dasMulli: Maybe there's a difference between a "core" SDK like m.net.sdk(.web) and a custom one not being resolved? As long as you don't mess around with that core sdk and only have a broken reference to a NuGet one, does it still show that error dialog?

jeffkl: @dasMulli correct, if another non "core" SDK is invalid and ignored, you will not get a dialog.  You still get a build error which is what we want but VS does not unload the project

jviau: I am not sure CPS will be able to tell if a "core" SDK (or really any SDK) has broken at the time of reload. It is far too expensive to re-evaluate the project during the reload. Comparing if any `Sdk` elements or attributes changed won't be enough either, as they can be evaluated - ie: `<Sdk Name="$(MySdk)" Version="$(SdkVersion)" />`

dasMulli: > I wonder if this now meets the bar for 15.6 since we will have to make a considerable change to CPS as well.

The experience when breaking the core SDK isn't very good in 15.5 either, but this fix would help a lot when tinkering with custom SDKs, which is a feature in 15.6 that encourages users to modify SDK imports. If some form of a fix that doesn't make me click through hundreds (=6 ^^) of dialogs when I break something in the custom import (or on the NuGet feed) makes it into 15.6.*, that would be awesome!!

For reference, this happens when I break the core SDK in 15.5 (after which the project is just unloaded):

screen shot 2018-02-14 at 19 45 26
screen shot 2018-02-14 at 19 44 49

jviau: I agree that experience is awful. To get a good experience we need to do a lot of validation of _what_ has changed in the project during a reload, so that we never reload into a broken "core" SDK. This is very expensive to do, especially when a common import is changed (we have to validate every project that imports it). I am thinking we should have to do something like track every property that is impacting the final SDK result, and if any of them change, we submit an error to the error list asking the user to eventually reload the solution.

jviau:
When is this going to be inserted to 15.6? CPS will need changes as well, the `null` is going to cause us to throw with our current code.

jeffkl:
@AndyGerlicher was working on an insertion today

PR #2998, opened 2018-02-15
Fix package versions

jeffkl:
Official build with no prerelease in version.json:
  `Microsoft.Build.15.7.47`

Official build with a prerelease in version.json:
  `Microsoft.Build.15.7.0-preview-000047`

Local builds:
  `Microsoft.Build.15.7.0-preview-000047-private`


AndyGerlicher:
Note this drops the `-random_number` at the end that we used to have. It didn't work as is and we couldn't think of a reason to keep it. It was probably in case a build step failed after package publish and we needed to rebuild (which isn't the case anymore).

PR #3009, opened 2018-02-20
Only write the generated app.config if needed

### LINKED ISSUE #3008: GenerateBindingRedirects always writes to the intermediate app.config
### Steps to reproduce

Consider the following project file:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.1</TargetFramework>
    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
</PropertyGroup>

<Target AfterTargets="CopyFilesToOutputDirectory" Name="CheckAppConfig" Inputs="@(AppConfigWithTargetPath)" Outputs="$(MSBuildProjectDirectory)/mydir/%(AppConfigWithTargetPath.TargetPath)">
    <Message Text="AppConfig Written!" />
    <Copy SourceFiles="@(AppConfigWithTargetPath)" DestinationFiles="$(MSBuildProjectDirectory)/mydir/%(AppConfigWithTargetPath.TargetPath)" />
</Target>

</Project>
```

### Expected  behavior

When ran twice in a row, the `CheckAppConfig` target should only run the first time

### Actual behavior

`CheckAppConfig` runs every time msbuild is ran.

### Environment data

This occurs regardless of environment.

### Notes

This occurs because the `GenerateBindingRedirects` task writes to the intermediate app.config file even if no updates to the file are needed.

This issue is essentially a duplicate of #2613 which was closed because it doesn't cause recompilation, while this is true any targets or tasks that use the app.config as an input and write it to another location will always be triggered.

mletterle:
We compare what we're going to write to what is already there if
present. If the contents are already there, then don't bother.

Note: This is rather naive, as any difference will cause the file to be
written. For example if, for some reason, the order of the binding
redirects changes.

This fixes #3008

### DIFF
--- a/src/Tasks.UnitTests/GenerateBindingRedirects_Tests.cs
+++ b/src/Tasks.UnitTests/GenerateBindingRedirects_Tests.cs
@@ -12,12 +12,28 @@ using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 using Microsoft.Build.UnitTests;
 using Microsoft.Build.Utilities;
+using Shouldly;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Microsoft.Build.Tasks.Unittest
 {
-    public class GenerateBindingRedirectsTests
+    public class GenerateBindingRedirectsTests : IDisposable
     {
+        private readonly ITestOutputHelper _output;
+        private readonly TestEnvironment _env;
+
+        public GenerateBindingRedirectsTests(ITestOutputHelper output)
+        {
+            _output = output;
+            _env = TestEnvironment.Create(output);
+        }
+
+        public void Dispose()
+        {
+            _env.Dispose();
+        }
+
         /// <summary>
         /// In this case,
         /// - A valid redirect information is provided for <see cref="GenerateBindingRedirects"/> task.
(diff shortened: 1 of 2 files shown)


rainersigwald:
Thanks!

This looks great (except for the compile error . . .) but we need to make sure it doesn't break Visual Studio's Fast-Up-To-Date check before we can go forward with it (because we have layers of timestamp-based incrementality and it's unfortunate).

mletterle:
Gosh darnit, I somehow missed committing the using statement. :man_shrugging:

AndyGerlicher:
@panopticoncentral Any idea if this could break the fast up-to-date check? I'd like to take this for 15.8.

panopticoncentral:
I don't *think* so, since you're just not writing the file if it didn't change. I think this should be OK, but it should obviously be tested.

AndyGerlicher:
Added a quick unit test for the scenario. This should be good to go. Thanks a lot @mletterle for the contribution!
@rainersigwald

rainersigwald:
My simple tests suggest that the .exe.config isn't considered an output in FUTD so this should be ok. If we ever enlighten FUTD about that, we'll have an unnecessary-rebuild problem, but we can cross that when we come to it.

src/UnitTests.Shared/ShouldlyAssertionExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+            var upper = expected.Add(range);
+            var lower = expected.Subtract(range);
+
+            if (actual > upper)
+            {
+                actual.ShouldBeLessThan(upper);
+            }
+            else if (actual < lower)
+            {
+                actual.ShouldBeLessThan(lower);
```
rainersigwald: Replace these two with `actual.ShouldBeInRange(lower, upper)`?

src/UnitTests.Shared/ShouldlyAssertionExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Shouldly;
+
+namespace Microsoft.Build.UnitTests.Shared
+{
+    public static class ShouldlyAssertionExtensions
+    {
+        public static void ShouldBeCloseTo(this DateTime actual, DateTime expected)
+        {
+            ShouldBeCloseTo(actual, expected, TimeSpan.FromSeconds(30));
```
rainersigwald: 30 seconds seems . . . generous. Maybe 2 or 4?

PR #3012, opened 2018-02-20
Ensure GetNuGetPackageVersionEx runs on Pack P2P

AndyGerlicher:
The Pack target calls into the P2P reference target _GetProjectVersion which calls GetBuildVersion. Setting our GetNuGetPackageVersionEx fixup target to run after GetBuildVersion to ensure that it always runs. This was causing out dependency versions to be different than our actual versions.


src/Directory.Build.targets
```diff
@@ -105,7 +105,7 @@
     </ItemGroup>
   </Target>
   
-  <Target Name="GetNuGetPackageVersionEx" BeforeTargets="GenerateNuspec" DependsOnTargets="GetBuildVersion">
+  <Target Name="GetNuGetPackageVersionEx" BeforeTargets="GenerateNuspec" AfterTargets="GetBuildVersion">
```
rainersigwald: I think I'd rather see `BeforeTargets="GenerateNuspec" AfterTargets="GetBuildVersion" DependsOnTargets="GetBuildVersion"` but this is ok

AndyGerlicher: Yup, you were right! What I had didn't work for a few cases. And I ended up with what you have here :)

PR #3017, opened 2018-02-22
Evaluation Context

cdmihai:
It doesn't do anything right now. In the future, it will cache SDK resolution results, IO operations, etc.

### DIFF
--- a/src/Build/Evaluation/Context/EvaluationContextBase.cs
+++ b/src/Build/Evaluation/Context/EvaluationContextBase.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+// </copyright>
+
+using PublicEvaluationContext = Microsoft.Build.Framework.EvaluationContext.EvaluationContext;
+
+namespace Microsoft.Build.Evaluation.Context
+{
+    internal abstract class EvaluationContextBase : PublicEvaluationContext
+    {
+    }
+}
--- a/src/Build/Resources/Strings.resx
+++ b/src/Build/Resources/Strings.resx
@@ -1467,6 +1467,9 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
   <!-- #################################################################################################-->
   <!-- The OM_ prefix indicates at the point of use that there is no error code in the message;
        these should not be used in an InvalidProjectFileException -->
+  <data name="OM_IncorrectEvaluationContextType">
+    <value>The provided evaluation context of type {0} was not created using the API factory {1}.</value>
+  </data>
   <data name="OM_IncorrectObjectAssociation">
     <value>The "{0}" object specified does not belong to the correct "{1}" object.</value>
   </data>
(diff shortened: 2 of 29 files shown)


AndyGerlicher:
Looks good! Only thing I wonder is if we want to get something using it as a POC before we commit? Since there are public APIs that could change if something was missed? But it looks designed well for non-breaking changes.

@jeffkl would like your input if you can get the SDK resolution to use this to cache for VS sessions.

src/Framework/EvaluationContext/EvaluationContext.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+
+namespace Microsoft.Build.Framework.EvaluationContext
+{
+    /// <summary>
+    /// An object used by the caller to extend the lifespan of evaluation caches (by passing the object on to other evaluations).
+    /// The caller should throw away the context when the environment changes (IO, environment variables, SDK resolution inputs, etc).
+    /// </summary>
+    public abstract class EvaluationContext
```
jeffkl: Should this be named `ProjectEvaluationContext`?  I don't think we'll have another evaluation context so it might not be worth it but it caught my eye

src/Framework/EvaluationContext/EvaluationContextFactory.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+// </copyright>
+
+namespace Microsoft.Build.Framework.EvaluationContext
+{
+    /// <summary>
+    /// Factory for <see cref="EvaluationContext"/>
+    /// </summary>
+    public abstract class EvaluationContextFactory
```
jeffkl: Do we need a factory?  Can there be a static method on `EvaluationContext` so the pattern is aligned with stuff like `ProjectRootElement.Create()`

src/Build/Resources/Strings.resx
```diff
@@ -1467,6 +1467,9 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
   <!-- #################################################################################################-->
   <!-- The OM_ prefix indicates at the point of use that there is no error code in the message;
        these should not be used in an InvalidProjectFileException -->
+  <data name="OM_IncorrectEvaluationContextType">
+    <value>The provided evaluation context of type {0} was not created using the API factory {1}.</value>
```
jeffkl: I'm interested to know when this will happen?  Can't we ensure that the provided evaluation context is always valid?

src/Framework/EvaluationContext/EvaluationContext.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+// </copyright>
+
+using System;
+
+namespace Microsoft.Build.Framework.EvaluationContext
```
jeffkl: The namespace here should probably just be `Microsoft.Build.Framework`

src/Build/Evaluation/Context/EvaluationContextBase.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+// </copyright>
+
+using PublicEvaluationContext = Microsoft.Build.Framework.EvaluationContext.EvaluationContext;
```
jeffkl: Changing the namespace of `EvaluationContext` would obviously remove the need for this

cdmihai:
Replaced by https://github.com/Microsoft/msbuild/pull/3075

PR #3022, opened 2018-02-23
Update README.md

Matthew-Bonner:
Change cibuild.cmd to build.cmd since there is no cibuild.cmd and build.cmd appears to the the correct command to run.


Matthew-Bonner:
Closing for now. See https://github.com/Microsoft/msbuild/pull/3022/commits/2cf5d09bb8c772fb9607e5031696e3c11fe612cd for why.

rainersigwald:
You're confused because this is super confusing. We recently changed the entry point for our builds. Now, CI servers start at `build/cibuild.cmd`, but humans should be directed to `build.cmd` in the root, so this change is good. I'm going to reopen and accept it.

PR #3032, opened 2018-02-26
Remove cibuild.* and fix bash shebang

kasper3:


kasper3:
@dotnet-bot test ci please

kasper3:
@rainersigwald, can you please ask @dotnet-bot to test CI. I changed netci.groovy to call build.cmd/build.sh scripts, and now cibuild.cmd/sh are not required.

AndyGerlicher:
Agreed with what Rainer said. Things could be clarified, but the cibuild is designed to specify arguments to the main build script and isn't redundant.

kasper3:
I understand the sentiment. However, in this case it is literally 12 out of 13 redundant lines of code and the only line that is different has two extra arguments. This patch was a try to consolidate the implementations, so to dogfood the script in CI. Maintaining two scripts vs. one, mitigating the chances of non-ci one getting staled. With that being said, I respectfully disagree.

PR #3033, opened 2018-02-26
Add test for reproduce Filename metadata problem on Unix.

mfilippov:
I found a problem with slashes in metadata evaluation. I created repro test for this problem. But I'm not sure where I should fix it. Please help me find the right place.

### DIFF
--- a/src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
@@ -15,7 +15,9 @@ using Microsoft.Build.Construction;
 using Microsoft.Build.Engine.UnitTests;
 using Microsoft.Build.Engine.UnitTests.Globbing;
 using Microsoft.Build.Evaluation;
+using Microsoft.Build.Execution;
 using Microsoft.Build.Shared;
+using Shouldly;
 using InvalidProjectFileException = Microsoft.Build.Exceptions.InvalidProjectFileException;
 using Xunit;
 
@@ -2712,6 +2714,30 @@ namespace Microsoft.Build.UnitTests.OM.Definition
 
             Assert.Equal(1, items.Count);
             return items[0];
+        }
+
+        /// <summary>
+        /// Item metadata "Filename" should not depends on platform specific slashes.
+        /// </summary>
+        [Fact]
+        public void FileNameMetadataEvaluationShouldNotDependsFromPlatformSpecificSlashes()
+        {
+            using (var env = TestEnvironment.Create())
+            using (var projectCollection = new ProjectCollection())
+            {
+                var testFiles = env.CreateTestProjectWithFiles(@"<?xml version=`1.0` encoding=`utf-8`?>
+                    <Project ToolsVersion=`msbuilddefaulttoolsversion` DefaultTargets=`Validate` xmlns=`msbuildnamespace`>
+                      <ItemGroup>
+                        <A Include=`A\B\C\D.cs` />
+                        <B Include=`@(A->'%(Filename)_test.ext')` />
+                      </ItemGroup>
+                    </Project>");
+                var project = new Project(testFiles.ProjectFile, new Dictionary<string, string>(), null, projectCollection);
+                var buildManager = BuildManager.DefaultBuildManager;
+                var projectInstance = buildManager.GetProjectInstanceForBuild(project);
+                var itemB = projectInstance.Items.Single(i => i.ItemType == "B").EvaluatedInclude;
+                itemB.ShouldBe("D_test.ext");
+            }
         }
     }
 }
--- a/src/Shared/Modifiers.cs
+++ b/src/Shared/Modifiers.cs
@@ -440,7 +440,8 @@ namespace Microsoft.Build.Shared
                         }
                         else
                         {
-                            modifiedItemSpec = Path.GetFileNameWithoutExtension(itemSpec);
+                            // Fix path to avoid problem with Path.GetFileNameWithoutExtension when backslashes in itemSpec on Unix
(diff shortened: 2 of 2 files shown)


mfilippov:
@cdmihai Could you advise where I can search the problem?

cdmihai:
Some places come to mind:

- entry point in expander, which interprets the itemspec string: https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Expander.cs#L1651
- entry point into well known metadata computation: https://github.com/Microsoft/msbuild/blob/master/src/Shared/Modifiers.cs#L390

Regarding the test placement, since this is an evaluation bug, it might better go to Microsoft.Build.UnitTests.OM.Definition.ProjectItem_Tests

mfilippov:
@cdmihai Thank you for the information I try to fix it.

mfilippov:
@cdmihai I fixed the issue. Failure looks like not connected with my changes.

mfilippov:
@cdmihai I did not replace slashes in this place https://github.com/Microsoft/msbuild/blob/e8b480c373eadd9daa975c6e96e4dbbabcd9a4fe/src/Shared/Modifiers.cs#L435 because null can be returned on Unix only when pass / as dir name. It case looks like not relevant.

mfilippov:
@dotnet-bot test RHEL7.2 Build for CoreCLR  please

mfilippov:
@cdmihai PR ready to review.

src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
```diff
@@ -2713,5 +2715,40 @@ private static ProjectItem GetOneItem(string content)
             Assert.Equal(1, items.Count);
             return items[0];
         }
+
+        /// <summary>
+        /// Item metadata "Filename" should not depends on platform specific slashes.
+        /// </summary>
+        [Fact]
+        public void FileNameMetadataEvaluationShouldNotDependsFromPlatformSpecificSlashes()
```
cdmihai: Rather than manually handling files and the project collection, you could use Microsoft.Build.UnitTests.TestEnvironment.CreateTestProjectWithFiles to handle everything for you.

src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
```diff
@@ -2713,5 +2715,40 @@ private static ProjectItem GetOneItem(string content)
...
+
+            try
+            {
+                File.WriteAllText(projectA, ObjectModelHelpers.CleanupFileContents(contentsA));
+                var projectCollection = new ProjectCollection();
+                var project = new Project(projectA, new Dictionary<string, string>(), null, projectCollection);
+                var buildManager = BuildManager.DefaultBuildManager;
+                var projectInstance = buildManager.GetProjectInstanceForBuild(project);
+                var itemB = projectInstance.Items.Single(i => i.ItemType == "B").EvaluatedInclude.Replace("/", "\\");
+                itemB.ShouldNotContain("A\\B\\C\\A\\B\\C");
```
cdmihai: Please assert the exact expected value, not what it shouldn't contain, since there's only one correct value, but many incorrect ones.

cdmihai:
The RHEL job's busted, don't worry about it :)

mfilippov:
@cdmihai I fixed review problems.

src/Build.OM.UnitTests/Definition/ProjectItem_Tests.cs
```diff
@@ -2713,5 +2715,29 @@ private static ProjectItem GetOneItem(string content)
...
+                    <Project ToolsVersion=`msbuilddefaulttoolsversion` DefaultTargets=`Validate` xmlns=`msbuildnamespace`>
+                      <ItemGroup>
+                        <A Include=`A\B\C\D.cs` />
+                        <B Include=`@(A->'%(Filename)_test.ext')` />
+                      </ItemGroup>
+                    </Project>");
+                var project = new Project(testFiles.ProjectFile, new Dictionary<string, string>(), null, projectCollection);
+                var buildManager = BuildManager.DefaultBuildManager;
+                var projectInstance = buildManager.GetProjectInstanceForBuild(project);
+                var itemB = projectInstance.Items.Single(i => i.ItemType == "B").EvaluatedInclude.Replace("/", "\\");
```
rainersigwald: Is this `Replace` needed? There shouldn't be any slashes in the include now, right?

src/Shared/Modifiers.cs
```diff
@@ -440,7 +440,8 @@ internal static string GetItemSpecModifier(string currentDirectory, string itemS
                         }
                         else
                         {
-                            modifiedItemSpec = Path.GetFileNameWithoutExtension(itemSpec);
+                            // Fix path to avoid problem with Path.GetDirectoryName when backslashes in itemSpec on Unix
```
rainersigwald: Haven't you already called `GetDirectoryName`? Should the comment say `GetFileNameWithoutExtension` or should the fixup be applied up a level?

mfilippov: I'm not sure because GetDirectoryName handles a problem when called it on the root folder. As see in corefx implementation on Unix it happened only on "/". For this context "empty string", it is correct behavior for the filename. I can't find any way when it may be a problem. I try to make the minimum perf impact on my fix.

mfilippov:
@rainersigwald Remove replace it stayed from my previous version.

src/Shared/Modifiers.cs
cdmihai: I think @rainersigwald is saying that there might be a typo in the comment: `Path.GetFileNameWithoutExtension` instead of `Path.GetDirectoryName`. Is this correct?

mfilippov: @cdmihai Sure, sorry I didn't catch up it.

PR #3035, opened 2018-02-27
Fix some content files in Microsoft.Build.Runtime

jeffkl:
Not all files in the package were marked as "Content" because they are in a different item group.

1. Disable AddAppConfigToBuildOutputs otherwise we end up with an app.config in the netstandard2.0 folder
2. Copy _BuildOutputInPackage to _PackageFiles with metadata so that all files in the package are content files
3. Clear _BuildOutputInPackage since we don't want any files in the package to NOT be content


jeffkl:

PR #3038, opened 2018-02-28
Set MSBuildAllProjects automatically to the project and all imports

### LINKED ISSUE #1299: Make MSBuildAllProjects magically append on import

jeffkl:
* Set MSBuildAllProjects after first pass to the project and all imports, sorted descending by last modified 
* Log a low importance message that the property was ignored and set automatically
* Add escape hatch to disable the functionality

Fixes #1299

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -201,9 +201,9 @@ namespace Microsoft.Build.Evaluation
         private readonly EvaluationProfiler _evaluationProfiler;
 
         /// <summary>
-        /// Keeps track of the newest project to become the value of MSBuildAllProjects.
+        /// Keeps track of all imported projects with the newest first.
         /// </summary>
-        private ProjectRootElement _newestProject;
+        private SortedList<DateTime, ProjectRootElement> _allProjects;
 
         /// <summary>
         /// Stores a value indicating if the project set MSBuildAllProjects so that a message can be logged that the property was ignored.
@@ -690,8 +690,16 @@ namespace Microsoft.Build.Evaluation
         DataCollection.CommentMarkProfile(8816, endPass0);
 #endif
 
-                // By default, the "newest" project is the project itself.  As projects are imported, they will replace this pointer if they are newer than the project.
-                _newestProject = _projectRootElement;
+                if (!Traits.Instance.EscapeHatches.EnableLegacyMSBuildAllProjects)
+                {
+                    _allProjects = new SortedList<DateTime, ProjectRootElement>(new ComparerDescendingAllowDuplicates<DateTime>());
+
+                    if (_projectRootElement.FullPath != null)
+                    {
+                        // By default, the "newest" project is the project itself.  As projects are imported, they will replace this pointer if they are newer than the project.
+                        _allProjects.Add(_projectRootElement.LastWriteTimeWhenRead, _projectRootElement);
+                    }
+                }
 
                 // Pass1: evaluate properties, load imports, and gather everything else
                 using (_evaluationProfiler.TrackPass(EvaluationPass.Properties))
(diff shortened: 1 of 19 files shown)


jeffkl:
I'll write some unit tests if we agree on the implementation

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2316,6 +2350,12 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
                         {
                             imports.Add(importedProjectElement);
 
+                            // Store a pointer to the newest project to be set later as MSBuildAllProjects
+                            if (importedProjectElement.TimeLastChanged > _newestProject.TimeLastChanged)
```
jeffkl: I'm not sure why but this doesn't work as expected.  TimeLastChanged is not the last write time of the project?

jeffkl:
This is on hold, VCXPROJ has a dependency on MSBuildAllProjects.

jeffkl:
This is now ready for review!  It now sets MSBuildAllProjects to the project and all imports with the newest project being first in the list

jeffkl:
@dotnet-bot 
test Windows_NT Build for Full please
test OSX10.13 Build for CoreCLR please

jeffkl:
@Microsoft/msbuild-maintainers please review

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1383,6 +1418,15 @@ private void EvaluatePropertyElement(ProjectPropertyElement propertyElement)
         {
             using (_evaluationProfiler.TrackElement(propertyElement))
             {
+                if (propertyElement.Name.Equals(ReservedPropertyNames.allProjects, StringComparison.OrdinalIgnoreCase)
+                    && !Traits.Instance.EscapeHatches.EnableLegacyMSBuildAllProjects)
+                {
+                    // Ignore evaluating MSBuildAllProjects unless the escape hatch is enabled.  Also keep track of whether or not it was set
+                    // so a message can be logged indicating it was ignored.
+                    _msbuildAllProjectsWasSet = true;
```
rainersigwald: I'm still a bit leery of this--I'd rather see it also check for the usual pattern `$(MSBuildAllProjects);$(MSBuildThisFileFullPath)` for safety.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -200,6 +200,16 @@ internal class Evaluator<P, I, M, D>
 
         private readonly EvaluationProfiler _evaluationProfiler;
 
+        /// <summary>
+        /// Keeps track of all imported projects with the newest first.
```
cdmihai: nit: newest as in latest found by the evaluator, or newest as in most recently written to?

cdmihai: Though I guess the DateTime in the type says it all :)

---
In reply to: [174532847](https://github.com/Microsoft/msbuild/pull/3038#discussion_r174532847) [](ancestors = 174532847)

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1383,6 +1418,15 @@ private void EvaluatePropertyElement(ProjectPropertyElement propertyElement)
         {
             using (_evaluationProfiler.TrackElement(propertyElement))
             {
+                if (propertyElement.Name.Equals(ReservedPropertyNames.allProjects, StringComparison.OrdinalIgnoreCase)
```
cdmihai: This is not the hottest of paths, but still hot, so it might show up on perf. I'll run a perf test and see how it looks.

Worst case, we need to reuse the pattern from other string matchings in hot paths, like this one: [https://github.com/Microsoft/msbuild/blob/master/src/Shared/Modifiers.cs#L93-L111](https://github.com/Microsoft/msbuild/blob/master/src/Shared/Modifiers.cs#L93-L111)

cdmihai:
If a project is left out on purpose from the manual pattern, it will be included in the automated one. But I guess that it's not a common case.

jeffkl:
We've decided we can't touch MSBuildAllProjects and we'll have to introduce a new property instead.

PR #3041, opened 2018-03-02
Workaround asp service crash

cdmihai:
After a recent change in MSBuild (https://github.com/Microsoft/msbuild/commit/b8d01363d9781b38e202a692b07914924eb16df5#diff-32568b659f3471e9ecdeb620a363af3c), the ASP form intellisense stopped working. There was no visible error in VS, it just reverted back to the html autocomplete, and thus failing some DDRITS.
It happens because the ASP process is not being able to find Microsoft.Build.Framework when referenced from Microsoft.Build.Utiliites.Core. It probably needs to set some assembly resolvers?

VSTS bug: https://devdiv.visualstudio.com/DevDiv/MSBuild/_workitems/edit/574986

The temporary workaround is to ifdef out the code that depends on framework. In this case this is simple to do, since the code is just detecting whether to run tests or not.

Another better workaround is to just teach MSBuild the name of the new process.

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -309,12 +309,17 @@ namespace Microsoft.Build.Shared
         {
             return FileUtilities.CombinePaths(visualStudioRoot, "MSBuild", CurrentToolsVersion, "Bin", "MSBuild.exe");
         }
-
+#if !WORKAROUND_NOT_RUNNING_TESTS
         private static bool? _runningTests;
         private static readonly object _runningTestsLock = new object();
+#endif
 
         private static bool CheckIfRunningTests()
         {
+        //VS ASP intellisense server fails without this flag. Remove when issue fixed: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/574986
+#if WORKAROUND_NOT_RUNNING_TESTS
+            return false;
+#else
             if (_runningTests != null)
             {
                 return _runningTests.Value;
@@ -337,6 +342,7 @@ namespace Microsoft.Build.Shared
 
                 return _runningTests.Value;
             }
+#endif
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


cdmihai:

cdmihai:

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -309,14 +311,33 @@ private static string GetMSBuildExeFromVsRoot(string visualStudioRoot)
...
+                //  See the comments on the TestInfo class for an explanation of why it works this way.
+                var frameworkAssembly = typeof(Framework.ITask).Assembly;
+                var testInfoType = frameworkAssembly.GetType("Microsoft.Build.Framework.TestInfo");
+                var runningTestsField = testInfoType.GetField("s_runningTests", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static);
+
+                _runningTests = (bool)runningTestsField.GetValue(null);
+
+                return _runningTests.Value;
+            }
         }
```
AndyGerlicher: Nit: Some redundant cases here that could fall through a bit cleaner:
```C#
if (_runningTests == null)
{
    lock (_runningTestsLock)
    {
        if (_runningTests == null)
        {
            //  Check if running tests via the TestInfo class in Microsoft.Build.Framework.
            //  See the comments on the TestInfo class for an explanation of why it works this way.
            var frameworkAssembly = typeof(Framework.ITask).Assembly;
            var testInfoType = frameworkAssembly.GetType("Microsoft.Build.Framework.TestInfo");
            var runningTestsField = testInfoType.GetField("s_runningTests", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static);

            _runningTests = (bool)runningTestsField.GetValue(null);
        }
    }
}

return _runningTests.Value;
```

PR #3048, opened 2018-03-05
Update to RepoToolset 1.0.0-beta-62705-01

tmat:
The new toolset includes fixes required by Orchestrated and Source builds.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -39,6 +39,14 @@
     <UpdateXlfOnBuild>true</UpdateXlfOnBuild>
     <ErrorOnOutOfDateXlf>true</ErrorOnOutOfDateXlf>
 
+    <UpdateXlfOnBuild Condition="'$(CIBuild)' != 'true'">true</UpdateXlfOnBuild>
+
+    <!-- Workaround that can be removed when we update machines to 15.6.
+         Use Satellite assembly generation task from Microsoft.NET.Sdk even when building with
+         full Framework MSBuild.  This will support public signing, is deterministic, and always
+         generates them as AnyCPU. -->
+    <GenerateSatelliteAssembliesForCore>true</GenerateSatelliteAssembliesForCore>
+
     <!-- Work around issue where bootstrapped TaskHostTask seems to be loading the wrong assemblies.
          https://github.com/Microsoft/msbuild/issues/2865-->
     <GenerateResourceMSBuildArchitecture>CurrentArchitecture</GenerateResourceMSBuildArchitecture>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -39,8 +39,6 @@
     <UpdateXlfOnBuild>true</UpdateXlfOnBuild>
     <ErrorOnOutOfDateXlf>true</ErrorOnOutOfDateXlf>
 
-    <UpdateXlfOnBuild Condition="'$(CIBuild)' != 'true'">true</UpdateXlfOnBuild>
-
     <!-- Workaround that can be removed when we update machines to 15.6.
          Use Satellite assembly generation task from Microsoft.NET.Sdk even when building with
          full Framework MSBuild.  This will support public signing, is deterministic, and always
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -36,7 +36,7 @@
   <!-- Toolset Dependencies -->
   <PropertyGroup>
     <DotNetCliVersion>2.1.300-preview2-008046</DotNetCliVersion>
-    <RoslynToolsRepoToolsetVersion>1.0.0-beta-62512-02</RoslynToolsRepoToolsetVersion>
+    <RoslynToolsRepoToolsetVersion>1.0.0-beta-62705-01</RoslynToolsRepoToolsetVersion>
     <VSWhereVersion>2.2.7</VSWhereVersion>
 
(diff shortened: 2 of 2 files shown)


tmat:
OSX failure is due to network failure

tmat:
@nguerrera @dsplaisted @rainersigwald Please review

src/Directory.Build.props
```diff
@@ -39,6 +39,14 @@
     <UpdateXlfOnBuild>true</UpdateXlfOnBuild>
     <ErrorOnOutOfDateXlf>true</ErrorOnOutOfDateXlf>
 
+    <UpdateXlfOnBuild Condition="'$(CIBuild)' != 'true'">true</UpdateXlfOnBuild>
+
+    <!-- Workaround that can be removed when we update machines to 15.6.
```
nguerrera: This can only be removed if repo toolset sets this to true again. Is that the plan?

src/Directory.Build.props
```diff
@@ -39,6 +39,14 @@
     <UpdateXlfOnBuild>true</UpdateXlfOnBuild>
     <ErrorOnOutOfDateXlf>true</ErrorOnOutOfDateXlf>
 
+    <UpdateXlfOnBuild Condition="'$(CIBuild)' != 'true'">true</UpdateXlfOnBuild>
```
nguerrera: Why is this coming in now?

src/Directory.Build.props
```diff
@@ -39,6 +39,14 @@
     <UpdateXlfOnBuild>true</UpdateXlfOnBuild>
```
nguerrera: @dsplaisted Why is this hard-coded to true? It should be false in CI.

src/Directory.Build.props
nguerrera: I think this won't have the desired effect because of https://github.com/Microsoft/msbuild/pull/3048/files/85cd086eefce5854d3f591ea67372aad6990d35d#r172350635

src/Directory.Build.props
nguerrera: cc @cdmihai I see it's from https://github.com/tmat/msbuild/commit/677197e28c8f999d5d4f10a72dbd59fd0d49c8e1

You're never going to get an out-of-date error like this and the build machine will be editing source files, which is bad.

If you want to do the updating manually before each hand-off, you can do ErrorOnOutOfDateXlf=false. And UpdateXlfOnBuild=false always.

src/Directory.Build.props
tmat: Yes. Once on 15.6 i'll clean all this up.

src/Directory.Build.props
tmat: Copy & paste error

src/Utilities.UnitTests/xlf/strings.pl.xlf
```diff
@@ -0,0 +1,22 @@
...
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
+  <file datatype="xml" source-language="en" target-language="pl" original="../strings.resx">
+    <body>
+      <trans-unit id="MyErrorResource">
+        <source>Oops I wiped your harddrive {0}</source>
+        <target state="new">Oops I wiped your harddrive {0}</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="MyWarningResource">
+        <source>Be nice or I wipe your harddrive {0}</source>
```
nguerrera: Tests are getting xlf generated now.

src/Directory.Build.props
cdmihai: Is there any issue with these settings?

With these settings, people would include the xlf diffs with the PR that also changes the resx, so there should be no source changes on the build machine. Besides, our CI job is set to [(hopefully) fail](https://github.com/Microsoft/msbuild/blob/master/build/build.ps1#L276-L285) if there is any output out of git status.
I like `UpdateXlfOnBuild=true` because stuff just happens. Otherwise, the build would fail saying the xlf is out of sync, and then people would have to understand what that error means and how to sync the xlf. `ErrorOnOutOfDateXlf=true` is just a sanity check :)

nguerrera: The recommendation for this workflow is to set UpdateXlfOnBuild to true but false on CI. This is how all of our other repos work. If you have a separate git status check, it's equivalent. But I would still prefer that you not set it unconditionally so that nobody without the git status check copies from you incorrectly.

tmat:
@nguerrera I left the removal of Xliff package as is for now. I though I already fixed it in the toolset but turns out I didn't. Next update will be able to remove that.

tmat:
@rainersigwald Could you please merge? We need the fix for orchestrated build.

rainersigwald:
What in orchestrated build is consuming our master branch? I thought nothing was (I am pleasantly surprised that something is).

tmat:
I'm not sure about the branch. Perhaps this needs to be merged to other branches as well?

rainersigwald:
Right now our 15.7 branch is in the past, not using repotoolset. Now when we merge to it it'll get this. That's on our to-do list.

PR #3049, opened 2018-03-05
 Fix race adding configuration in BuildManager

AndyGerlicher:
We have crash reports of a race condition in the AddNewConfiguration
method. Adding a lock to the only entry to that method currently without
a lock.

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/548385

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildManager.cs
+++ b/src/Build/BackEnd/BuildManager/BuildManager.cs
@@ -49,12 +49,12 @@ namespace Microsoft.Build.Execution
         /// <summary>
         /// The singleton instance for the BuildManager.
         /// </summary>
-        static private BuildManager s_singletonInstance;
+        private static BuildManager s_singletonInstance;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


rainersigwald:
I'm very confused by the access/locking rules here. Do we need to guard all write accesses to `_configCache` in `BuildManager`? Several remain unguarded, including the `BuildManager.AddNewConfiguration` that's throwing in the bug, presumably because a configuration was inserted between `GetMatchingConfiguration` and `AddNewConfiguration`. I think this large lock is sufficient, since it's the public entry point I can find for `AddNewConfiguration`, but how will we know if it's too broad?

cdmihai:
@rainersigwald I think the lock is not too broad. There's writes to shared state here and there all the way to the dataflow post method, which should be quick since it's async.

cdmihai:
@dotnet-bot test this please

AndyGerlicher:

AndyGerlicher:

Issue #166, opened 2015-08-26
Multi-core Node Process Priority


AndyGerlicher:
From a customer bug:
When running msbuild with parameter /m for multicore usage with low priority, (e.g. `start "" /LOW /B /WAIT msbuild /m [target]`) only the first msbuild.exe node is running with low priority. The spawned nodes run with normal priority.

Expected behavior: All spawned msbuild nodes run with main msbuild priority, in this example LOW

Ultimately we would like to do this, or even add a /priority:low switch to MSBuild itself for scenarios when CPU usage is very high on long builds and people are still wanting to use their machine.

One "gotcha" is node reuse. If the node is already around and "Normal", this will likely complicate things (or make the scenario not work).

informatorius:
As alternative (also to fix the node reuse gotcha), a registry or environment variable could control default msbuild process priority. 
The registry setting CpuPriorityClass=5 in
HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\msbuild.exe\PerfOptions
do not work for spawned msbuild nodes.

rainersigwald:
From https://github.com/Microsoft/msbuild/issues/1055#issuecomment-248671465:

> Looks like the bug is on line 368 of src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs. Instead of setting the value to NORMAL_PRIORITY_CLASS, just set it to zero.

Thanks for finding that @phkelly.

Fragmachine:
This issue has been causing us problems for a while when running tests on our local machines. Since it's such a simple fix I've created a custom "FrankenMSBuild" to get around it. To make it work I had to copy files such as the Tracker.exe, the numbered folders and the MSBuild.exe.config file (where I then remove the 'Microsoft.Build' entry) from the official MSBuild folder, a bit sketchy to say the least. Having an official fix would thus be ideal. What's preventing this from happening? Would it help if I offered to contribute this one line change? Seems a bit silly but if that helps to move forward on this issue I can certainly do that.

rainersigwald:
@Fragmachine Feel free to open up a PR, especially if you've tested it locally for a while.

I don't think I'm worried enough about the `/nodereuse:true` case to block the straightforward fix. If we get complaints, we can consider later adding process priority to the node handshake.

Fragmachine:
Great! However, I will have to wait for legal authorization from EA before I can do the PR. That might take a while ...

[...]

PR #3052, opened 2018-03-06
Removed the forced normal priority when creating a new node.

Fragmachine:
The child process will instead inherit the setting from its parent.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -465,7 +465,7 @@ namespace Microsoft.Build.BackEnd
 
             // Null out the process handles so that the parent process does not wait for the child process
             // to exit before it can exit.
-            uint creationFlags = BackendNativeMethods.NORMALPRIORITYCLASS;
+            uint creationFlags = 0;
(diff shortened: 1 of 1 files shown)


cdmihai:
The documentation suggests that the priority is only inherited when it's low, otherwise it defaults to normal: _If none of the priority class flags is specified, the priority class defaults to NORMAL_PRIORITY_CLASS unless the priority class of the creating process is IDLE_PRIORITY_CLASS or BELOW_NORMAL_PRIORITY_CLASS. In this case, the child process receives the default priority class of the calling process._

https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx

What's the reason behind your change? Do you want to lower the msbuild node priority?

rainersigwald:
@cdmihai Yes, this fixes #166 which makes it impossible to nice a multiproc build.

Fragmachine:
I'll just add that I've been running with this for a while now and it makes a huge difference when running builds in the background on the local machine. Instead of a forced coffee break, due to the machine effectively being hijacked with the mouse cursor locking up etc., one can continue to work as normal.

cdmihai:

rainersigwald:
My vote in https://github.com/Microsoft/msbuild/issues/166#issuecomment-369294512 was "wait for someone to complain".

Fragmachine:
Thanks guys, really appreciate it!

PR #3063, opened 2018-03-07
Add source control information extensibility point

tmat:
Adds a few extensibility points to common targets that enable source control and SourceLink package features and fully deterministic compilation.

See https://github.com/tmat/repository-info/blob/master/docs/Readme.md for details.

TODO:
- [x] Replace temp links to Standard CI spec with final links
- [ ] Follow up change in dotnet SDK to include revision id in AssemblyInformationalVersion (https://github.com/dotnet/sdk/pull/2028)
- [ ] Follow up change in C#, VB and F# targets files to use source roots to generate PathMap (https://github.com/dotnet/roslyn/pull/25325)
- [ ] Follow up change in C++ compiler to implement deterministic path mapping and use source roots to generate it
- [ ] Follow up change in NuGet targets to populate RepositoryUrl from PrivateRepositoryUrl (https://github.com/NuGet/NuGet.Client/pull/2081)
- [ ] Create source control and source link packages using these extensibility points

Supersedes https://github.com/Microsoft/msbuild/pull/2960

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -721,6 +721,34 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </ItemDefinitionGroup>
 
   <!--
+    Target that allows targets consuming source control confirmation to establish a dependency on targets producing this information.
+
+    Any target that reads SourceRevisionId, PrivateRepositoryUrl, SourceRoot, and other source control properties and items 
+    should depend on this target and be conditioned on '$(SourceControlInformationFeatureSupported)' == 'true'.
+
+    SourceRevisionId property uniquely identifies the source control revision of the repository the project belongs to.
+    For Git repositories this id is a commit hash, for TFVC repositories it's the changeset number, etc.
+
+    PrivateRepositoryUrl property stores the URL of the repository supplied by the CI server or retrieved from source control manager.
+    Targets consuming this property shall not publish its value implicitly as it might inadvertently reveal an internal URL.
+    Instead, they shall only do so if the project sets PublishRepositoryUrl property to true. For example, the NuGet Pack target
+    may include the repository URL in the nuspec file generated for NuGet package produced by the project if PublishRepositoryUrl is true.
+
+    SourceRoot item group lists all source roots that the project source files reside under and their mapping to source control server URLs,
+    if available. This includes both source files under source control as well as source files in source packages. SourceRoot items are 
+    used by compilers to determine path map in deterministic build and by SourceLink provider, which maps local paths to URLs of source files
+    stored on the source control server.
+
+    Source control information provider that sets these properties and items shall execute before this target (by including 
+    InitializeSourceControlInformation in its BeforeTargets) and set source control properties and items that haven't been initialized yet.
+  -->
+  <Target Name="InitializeSourceControlInformation" />
+
+  <PropertyGroup>
+    <SourceControlInformationFeatureSupported>true</SourceControlInformationFeatureSupported>
+  </PropertyGroup>
+
+  <!--
     ***********************************************************************************************
     ***********************************************************************************************
                                                                 Build Section


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
   </ItemDefinitionGroup>
 
+  <!--
+    Target that allows targets consuming source control confirmation to establish a dependency
+    on the targets producing this information.
+
+    Any target that reads SourceRevisionId, PrivateRepositoryUrl, and other source control related
+    properties and items should depend on this target.
+
+    Source control information provider that sets these properties and items shall override this target
```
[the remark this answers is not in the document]
eerhardt: The issue is there can be only one `override`, so if 2 things want to set these properties/items...

tmat:
@richlander FYI

src/Tasks/Microsoft.Common.CurrentVersion.targets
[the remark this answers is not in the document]
tmat: I'll change it to `BeforeTargets`. There might potentially be scenario when one target fills in some information and another target fills in more info.

tmat:
@eerhardt Updated the comments. Let me know if it's clearer now.

src/Tasks/Microsoft.Common.props
```diff
@@ -11,6 +11,41 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      Any target that reads SourceRevisionId shall depend on InitializeSourceControlInformation target.
+      A source control information provider that sets SourceRevisionId shall do so in a target that executes before InitializeSourceControlInformation target.
+    -->
+    <SourceRevisionId Condition="'$(SourceRevisionId)' == ''">$(STANDARD_CI_SOURCE_REVISION_ID)</SourceRevisionId>
+
+    <!--
+      PrivateRepositoryUrl is the URL of the repository supplied by the CI server or retrieved from source control manager.
+
+      Targets consuming this property shall not publish its value implicitly as it might inadvertently reveal an internal URL.
+      Instead, they shall only do so if the project sets PublishRepositoryUrl property to true. For example, the NuGet Pack target
```
eerhardt: > they shall only do so if the project sets PublishRepositoryUrl property to true

Ok - this helped me understand what is going on.  Thanks!

tmat:
@rainersigwald I'd like to get this into VS 15.7 and the next dotnet SDK (whatever the version number is). Should I retarget the PR to vs15.7 branch?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,19 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    on the targets producing this information.
+
+    Any target that reads SourceRevisionId, PrivateRepositoryUrl, and other source control related
+    properties and items should depend on this target.
+
+    Source control information provider that sets these properties and items shall execute before
+    this target (by including InitializeSourceControlInformation in its BeforeTargets) and set
+    source control properties and items that haven't been initialized yet.
+  -->
+  <Target Name="InitializeSourceControlInformation" />
```
rainersigwald: Do we need to insert the DependsOn property we talked about somewhere so folks can inject into this but still have a package compatible with VS2015?

tmat: Yes, i'll add the property. 

I also have one more piece of feedback coming from NuGet team to look into before we merge.

tmat:
@AndyGerlicher @rainersigwald Based on feedback from NuGet team I have conditioned reading the `STANDARD_CI` variables based on a new `UseStandardCIVariables` property. We want to avoid potential (build or test) breaks that initializing these properties from the Standard CI variables might cause. 

For example, if `Deterministic` is on the compiler emits deterministic paths when running on CI build (`ContinuousIntegrationBuild` is true). This might affect unit tests running on the CI that read paths via `CallerFilePathAttribute`.

`UseStandardCIVariables` property will be switched on by the SourceLink package, so that customers who use that package and build deterministically (the default in .NET SDK projects) get all the deterministic goodness as well. Customers not using SourceLink can switch it on themselves.

tmat:
@nguerrera  FYI

tmat:
On second thought, since we need to make the usage of Standard CI opt-in and thus they are not defaults anymore it seems it would be better to look at Standard CI as another source information provider and move the properties to a separate package (`Microsoft.Build.StandardCI`). Dropping in this package opts into reading source control info from Standard CI env variables. SourceLink package would then depend on this package.

tmat:
@rainersigwald Could you please merge?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -720,6 +720,34 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+    SourceRoot item group lists all source roots that the project source files reside under and their mapping to source control server URLs,
+    if available. This includes both source files under source control as well as source files in source packages. SourceRoot items are 
+    used by compilers to determine path map in deterministic build and by SourceLink provider, which maps local paths to URLs of source files
+    stored on the source control server.
+
+    Source control information provider that sets these properties and items shall execute before this target (by including 
+    InitializeSourceControlInformation in its BeforeTargets) and set source control properties and items that haven't been initialized yet.
+  -->
+  <Target Name="InitializeSourceControlInformation" />
```
rainersigwald: Did you decide that we didn't need a DependsOn property here? I thought we wanted that for some packages that wanted to support downlevel.

tmat: I have added `SourceControlInformationFeatureSupported` so that they can condition on that. Let me double check that this property is actually reasonable to use.

tmat: It looks good:
https://github.com/dotnet/sdk/pull/2028/files#diff-6746d5c3e16c263867ca48a7b3a962bfR52

tmat:
Thanks everyone!

PR #3070, opened 2018-03-08
Update FileVersionHash used in node communications

AndyGerlicher:
The current handshake for task host communication uses only the assembly
major version portion. This causes issue if the protocol is changed for
any '15' release. Changed the FileVersionHash to hash the full assembly
informational version (which includes the commit hash, e.g.
'15.7.79-preview+g60af12df92').


AndyGerlicher:

PR #3074, opened 2018-03-09
Honor MSBUILDDEBUGPATH when dumping exception information to file

natemcmaster:
Currently trying to figure out why the build is failing inside an Alpine Linux docker container, but I can't get the log files because the container file system is automatically cleaned up.

This supports setting MSBUILDDEBUGPATH to control where MSBuild dump files are written.

cref https://github.com/Microsoft/msbuild/issues/1038

cc @cdmihai

### DIFF
--- a/src/Shared/ExceptionHandling.cs
+++ b/src/Shared/ExceptionHandling.cs
@@ -301,22 +301,26 @@ namespace Microsoft.Build.Shared
                 if (s_dumpFileName == null)
                 {
                     Guid guid = Guid.NewGuid();
-                    string tempPath = Path.GetTempPath();
+                    string dumpDirEnv = Environment.GetEnvironmentVariable("MSBUILDDEBUGPATH");
+                    string dumpDir = !string.IsNullOrEmpty(dumpDirEnv)
+                            ? dumpDirEnv
+                            : Path.GetTempPath();
 
                     // For some reason we get Watson buckets because GetTempPath gives us a folder here that doesn't exist.
                     // Either because %TMP% is misdefined, or because they deleted the temp folder during the build.
-                    if (!Directory.Exists(tempPath))
+                    if (!Directory.Exists(dumpDir))
                     {
                         // If this throws, no sense catching it, we can't log it now, and we're here
                         // because we're a child node with no console to log to, so die
-                        Directory.CreateDirectory(tempPath);
+                        Directory.CreateDirectory(dumpDir);
                     }
 
-                    s_dumpFileName = Path.Combine(tempPath, "MSBuild_" + guid.ToString() + ".failure.txt");
+                    var pid = Process.GetCurrentProcess().Id;
+                    s_dumpFileName = Path.Combine(dumpDir, $"MSBuild_pid-{pid}_{guid:n}.failure.txt");
 
                     using (StreamWriter writer = FileUtilities.OpenWrite(s_dumpFileName, append: true))
                     {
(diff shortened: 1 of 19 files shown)


src/Shared/ExceptionHandling.cs
```diff
@@ -301,22 +301,26 @@ internal static void DumpExceptionToFile(Exception ex)
                 if (s_dumpFileName == null)
                 {
                     Guid guid = Guid.NewGuid();
-                    string tempPath = Path.GetTempPath();
+                    string dumpDirEnv = Environment.GetEnvironmentVariable("MSBUILDDEBUGPATH");
```
jeffkl: Can you just make this:

```C#
string dumpDir = Environment.GetEnvironmentVariable("MSBUILDDEBUGPATH") ?? Path.GetTempPath();
```

And get rid of the other variable?  Or am I missing something?

src/Shared/ExceptionHandling.cs
```diff
@@ -301,22 +301,26 @@ internal static void DumpExceptionToFile(Exception ex)
...
+                    if (!Directory.Exists(dumpDir))
                     {
                         // If this throws, no sense catching it, we can't log it now, and we're here
                         // because we're a child node with no console to log to, so die
-                        Directory.CreateDirectory(tempPath);
+                        Directory.CreateDirectory(dumpDir);
                     }
 
-                    s_dumpFileName = Path.Combine(tempPath, "MSBuild_" + guid.ToString() + ".failure.txt");
+                    var pid = Process.GetCurrentProcess().Id;
```
jeffkl: Why did you introduce a new variable here?

natemcmaster: I wanted to put PID into the filename. Makes it just a little easier to correlate the log file to other parts of the logging output without having to open each file to find the PID.

src/Shared/ExceptionHandling.cs
natemcmaster: Just being safe in case MSBUILDDEBUGPATH is set but empty.

natemcmaster: You could do this too, but calling .GetEnvironmentVariable twice seemed uglier.

```
string dumpDir = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBUILDDEBUGPATH"))
                            ? Environment.GetEnvironmentVariable("MSBUILDDEBUGPATH")
                            : Path.GetTempPath();
```

natemcmaster:
I've updated to include mention of MSBUILDDEBUGPATH  in the MSB4166  error message

natemcmaster:
Yeah...I wasn't sure about adding that b/c it's mostly noise, but this project's editorconfig has set trim_trailing_whitespace = true, so :shrug: sorry for the noise

natemcmaster:
Ironic. OSX failed with `error MSB4166: Child node "6" exited prematurely. Shutting down. Diagnostic information may be found in files in "/tmp/" and will be named MSBuild_*.failure.txt. This location can be changed by setting the MSBUILDDEBUGPATH environment variable to a different directory.`

jeffkl:
We broke our Mac build in 4037f0694eee49321d61f4453fade907f71e6a9a, @AndyGerlicher is investigating...

natemcmaster:
Thanks for the review everyone. I don't have write access to this repo, so someone will have to merge for me.

PR #3075, opened 2018-03-10
Cache sdk resolution in evaluation context

cdmihai:


src/Build/BackEnd/Components/SdkResolution/SdkResolverCachingWrapper.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.Build.Collections;
+using Microsoft.Build.Construction;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Shared;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.Build.BackEnd.SdkResolution
+{
+
+    internal sealed class SdkResolverCachingWrapper: ISdkResolverService
```
jeffkl: Do we want the evaluation context to also be reused for builds?  For example I create the evaluation service, evaluate some projects and then build but the SDK resolver cache won't be re-used because the submission ID will be set.  If so, this wrapper could just use a const for submission ID.

jeffkl: I'm not in love with the need for a wrapper, is there any way to just turn caching on for a particular instance of SdkResolverService?

cdmihai: Builds based on Project objects do not reevaluate. Instead, they execute their targets based on the prior evaluated state from the Project object. However, msbuild task invocations from p2ps (like `GetTargetPath` ) do get evaluated from scratch. So rather than transferring sdk resolution results from Project evaluations to builds, the bigger problem to solve here is how to pass a pool of evaluated Projects to a submission build, such that the p2ps do not get reevaluated but instead reuse the prior in-memory evaluations.

cdmihai: True, in this particular case there is not much of the wrapper. My initial hope was to also remove the cache from the out of proc sdk service, but that one turns out be a different cache. As discussed, I'll push the cache into a subclass of SdkResolverService.

cdmihai:
@jviau @lifengl

cdmihai:
@dotnet-bot test this please

PR #3077, opened 2018-03-12
Explicitly set MSBUILDDEBUGPATH during tests

cdmihai:
Inspired by #3074

### DIFF
--- a/src/Shared/UnitTests/TestAssemblyInfo.cs
+++ b/src/Shared/UnitTests/TestAssemblyInfo.cs
@@ -47,6 +47,9 @@ public class MSBuildTestAssemblyFixture : IDisposable
         _testEnvironment.CreateFolder(newTempPath);
 
         _testEnvironment.SetTempPath(newTempPath);
+
+        // Most places that use this variable fallback to temp, but setting it here makes it explicit.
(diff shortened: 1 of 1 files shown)


src/Shared/UnitTests/TestAssemblyInfo.cs
```diff
@@ -47,6 +47,9 @@ public MSBuildTestAssemblyFixture()
         _testEnvironment.CreateFolder(newTempPath);
 
         _testEnvironment.SetTempPath(newTempPath);
+
+        // Most places that use this variable fallback to temp, but setting it here makes it explicit.
+        _testEnvironment.SetEnvironmentVariable("MSBUILDDEBUGPATH", newTempPath);
```
rainersigwald: Should it bet temp or a new directory with new cleanup?

The current test invariant assumes temp and should probably be updated:

https://github.com/cdmihai/msbuild/blob/3cf7ee2f6ef932293e1b8505e7eb9d00eae3860e/src/Shared/UnitTests/TestEnvironment.cs#L326-L328

cdmihai: Ah, good catch. I'll move my change to #3074 and also extract out the path computation in a common place that can be accessed by the TestEnvironment.

cdmihai:
Moving changes to #3074

PR #3080, opened 2018-03-12
Create COM type libraries for VS

rainersigwald:
This fixes a few things that were missed long ago when MSBuild's code was pulled out of the internal VS repo and into this one.

* ComVisible is set to false by default in VS, but wasn't here. Setting it back to false + explicit enabling for members.
* COM tlb files were produced for some MSBuild assemblies, and consumed in VS native projects. Turned that on manually here.
* Added Microsoft.Build.Engine.dll and Microsoft.Build.Conversion.dll to the ExternalAPIs package so they can be referenced in the VS repo, even though they're obsolete and we wish they were unused.
* Undid #1910, which constituted a breaking change for VS because projects there still use the obsolete APIs and have warnaserror enabled.


rainersigwald:
I built an ExternalAPIs package containing these changes and consumed it in an internal VS PR: https://devdiv.visualstudio.com/DevDiv/MSBuild/_git/VS/pullrequest/108517

After many iterations of changing more projects to reference it instead of the stale internal projects, it has passed a build, so I think these changes are good.

cc @tmeschter

rainersigwald:
ah, cross-platform. Sigh, will fix that.

Issue #3082, opened 2018-03-13
Revitalize Tasks.UnitTests2 tests


rainersigwald:
There's a folder https://github.com/Microsoft/msbuild/tree/e6e619d4e28e27a127f5efb7c7bfbde0849c68b2/src/Tasks.UnitTests2 that hasn't ever built in open-source MSBuild.

We should either delete it or bring it back. Unfortunately, a quick look didn't reveal how valuable the tests were--there's a lot of boilerplate but also some actual tests.

YuliiaKovalova:
Close as not relevant anymore.

PR #3083, opened 2018-03-13
Delete stale projects

### LINKED ISSUE #2727: _NTDRIVE and _NTROOT
There are numerous references in the code base to _NTDRIVE and _NTROOT. This is from our legacy razzle build system in Visual Studio.

At first I thought this was just hold over from that time and that we had shimmed the values in some way. But looking through the code I can't see anywhere we shim the values. Instead the values seem to be always unset and hence resulting in a lot of invalid imports in the code. Example:

``` xml
 <Import Project="$(_NTDRIVE)$(_NTROOT)\tools\Microsoft.DevDiv.Settings.targets"/>
```

Is this code just dead and hanging around for no reason? Or do we still build this code via razzle in some configuration?

rainersigwald:
These projects have never built in the open-source MSBuild (they still
refer to internal tooling), so they must not be that important.

Fixes #2727, except for some things I'm leaving for #3082.

There's actual code in ApexTests, but I couldn't find anything that uses it even in the VS repo.


src/Build/Microsoft.Build.csproj
```diff
@@ -786,9 +786,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Remove="UnitTests\**;UnitTestsPublicOM\**;ApexTests\**;Collections\RetrievableEntryHashSet\Originals\*" />
-    <EmbeddedResource Remove="UnitTests\**;UnitTestsPublicOM\**;ApexTests\**" />
-    <None Remove="UnitTests\**;UnitTestsPublicOM\**;ApexTests\**" />
+    <Compile Remove="UnitTests\**;UnitTestsPublicOM\**;Collections\RetrievableEntryHashSet\Originals\*" />
```
AndyGerlicher: The UnitTests and UnitTestsPublicOM folders shouldn't exist anymore under Build, right?

AndyGerlicher:

rainersigwald:
Filed a bug to think about bringing them back (our successors can close it won't fix in 20 years).

PR #3098, opened 2018-03-16
Fix parsing and cleanup in binary logger

cdmihai:


src/Build.UnitTests/BinaryLogger_Tests.cs
```diff
@@ -24,6 +25,14 @@ public class BinaryLoggerTests
                <Exec Command='echo a'/>
             </Target>
          </Project>";
+        private TestEnvironment _env;
```
kasper3: Maybe `private readonly`?

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -80,6 +82,9 @@ public enum ProjectImportsCollectionMode
         /// </summary>
         public void Initialize(IEventSource eventSource)
         {
+            _intialTargetOutputLogging = Environment.GetEnvironmentVariable("MSBUILDTARGETOUTPUTLOGGING");
+            _initialLogImports = Environment.GetEnvironmentVariable("MSBUILDLOGIMPORTS");
+
```
AndyGerlicher: If you're going to get these values anyway, should we add a check if they're explicitly turned off not to turn them on? So if they're set to `false` or `0` or whatever honor that, otherwise set them normally?

src/Build.UnitTests/BinaryLogger_Tests.cs
```diff
@@ -24,67 +26,72 @@ public class BinaryLoggerTests
...
-            {
-                ObjectModelHelpers.BuildProjectExpectSuccess(s_testProject, binaryLogger);
-                Assert.True(Directory.Exists(directory));
-                Assert.True(File.Exists(log));
-            }
-            finally
-            {
-                ObjectModelHelpers.DeleteDirectory(directory);
-            }
+            Assert.False(File.Exists(_logFile));
```
AndyGerlicher: You could use `_env.ExpectFile(".binlog")` which will automatically assert that the file exists at the end (and doesn't create it).

src/Build/Logging/BinaryLogger/BinaryLogger.cs
cdmihai: We could do this for `MSBUILDLOGIMPORTS` because the enabling code checks for a specific value. However, can't be done for `MSBUILDTARGETOUTPUTLOGGING` because the enabling code checks for not null or empty. We'd need to introduce a new environment variable with an actual value and then keep using the old one for back compat. So I'd say not worth doing.

@KirillOsenkov Would there be any user interest to add the feature to turn off import and target output logging for the binary logger? Such as perf? If so, then it may be worth introducing a new env var which can be turned off.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -34,6 +34,8 @@ public sealed class BinaryLogger : ILogger
         private BinaryWriter binaryWriter;
         private BuildEventArgsWriter eventArgsWriter;
         private ProjectImportsCollector projectImportsCollector;
+        private string _intialTargetOutputLogging;
```
KirillOsenkov: >_intial [](start = 23, length = 7)

Typo: intial should be initial

src/Build/Logging/BinaryLogger/BinaryLogger.cs
KirillOsenkov: Perhaps, not sure if worth doing it now.

src/Build.UnitTests/BinaryLogger_Tests.cs
```diff
@@ -24,67 +24,54 @@ public class BinaryLoggerTests
...
-            }
-            finally
-            {
-                Environment.SetEnvironmentVariable(MSBUILDTARGETOUTPUTLOGGING, originalTargetOutputLogging);
-            }
+            Assert.Equal(mockLogger1.FullLog, mockLogger2.FullLog);
         }
 
         [Fact]
-        public void NonExistingDirectory()
```
KirillOsenkov: Are you removing the NonExistingDirectory() test?

KirillOsenkov:
Looks good, just fix the typo (intial) and make sure the removing of a test is intentional.

Just curious, what were the problems here? That the test would pollute the environment and other tests inherited the new env. variables? And that it incorrectly stripped the "?

src/Build.UnitTests/BinaryLogger_Tests.cs
cdmihai: Its asserts are being replaced by the new TestEnvironments checks, which makes the test obsolete.

cdmihai:
@KirillOsenkov 
The first commit fixes a bug where the `LogFile=<path.binlog>` does not work because of a typo in that one line. The receiver of the trim method was wrong :)
The polluted environment broke an invariant in my perf tests where each test iteration should produce the same evaluation, but the extra env vars caused extra global props to appear.

PR #3105, opened 2018-03-19
Use a task in _CheckForInvalidConfigurationAndPlatform

jeffkl:
This gives us the ability to localize messages that are logged in this target.

Related to #3059


src/Tasks/CheckConfigurationAndPlatform.cs
```diff
@@ -0,0 +1,98 @@
+﻿using Microsoft.Build.Shared;
+using System;
+
+namespace Microsoft.Build.Tasks
+{
+    /// <summary>
+    /// Verifies some critical properties like OutputPath are set.  This runs as part of the _CheckForInvalidConfigurationAndPlatform target.
+    /// </summary>
+    public class CheckConfigurationAndPlatform : TaskExtension
```
cdmihai: Rather than make a specialized task, why not make general localized Error and Warning tasks (maybe subclass the existing ones if that fits) that take in a resource name?

jeffkl: I started with that but couldn’t figure out a good way to specify the assembly and resource. You would need one Error task per assembly that contained resources right?  And if you specify an assembly in XML, it would need to be a path. And then the way you call it with format args would be interesting. Any ideas?

rainersigwald: Maybe just always have it be a resource in the Tasks resources, and just pass in a name? It'd only work for us, but it'd be a step in the right direction, I think.

Passing in things for the format string is definitely interesting. Maybe pass in an item and have metadata like Field1 Field2? That's not exactly elegant :(

dsplaisted: See https://github.com/dotnet/sdk/blob/master/src/Tasks/Common/MessageBase.cs and https://github.com/dotnet/sdk/blob/master/src/Tasks/Common/NETSdkError.cs for how we handle this in dotnet/sdk

AndyGerlicher:
@jeffkl do you want to update this for what Daniel suggested or close the issue? I'd rather not take this as-is but rather introduce a generic task (generic for us to use).

jeffkl:
@AndyGerlicher I barely had time to code this up as it was.  The only thing missing is unifying our `HasTrailingSlash` logic.

jeffkl:
I'd rather just fix HasTrailingSlash, if we want a whole new task that accepts a resource name and arguments that'll be a different PR.

PR #3092, opened 2018-03-14
WIP: Validate handle inheritance fix

rainersigwald:

Issue #3091, opened 2018-03-14
Core MSBuild nodes on Windows inherit process handles, keeping parent process alive


rainersigwald:
### Steps to reproduce

Easiest to observe in 66fd093ddda7 when running tests: a `dotnet.exe` process lives for a long time, causing the msbuild instance that invoked tests to hang.

### Expected  behavior

The process that invoked msbuild exits when the build is complete.

### Actual behavior

The process hangs until its child worker node idle-exits.

PR #3106, opened 2018-03-19
.NET Core node reuse redux

rainersigwald:
Enable node reuse for .NET Core (runtime 2.1+).

Interesting features:
* Attempts to connect to _every_ `dotnet` process, with no attempt to find only ones that are running MSBuild.
* There's an absolute length limit on the named pipe path of 108 chars on Linux. Tests setting custom temp dirs in deep paths ran afoul of this, so I changed the temp-path allocation strategy, which also required test-infrastructure changes that Mihai helped me figure out.
* I'm using the new `PipeOptions.CurrentUserOnly` as the only security mechanism on 2.1 and not running the old Windows-only validation code even on Windows.
* Long-lived worker nodes exposed #3091, which also manifested as a test hang. I fixed this in two ways:
  * Avoid `/nodereuse:true` in tests by explicitly setting it to `false` everywhere.
  * Changed process spawning on Windows to use the old `CreateProcess` codepath that explicitly stops inheriting stdout/stdin/stderr. See #3092 for proof that this works (Win tests don't hang) _without_ the test changes above.

For that last change, I highly recommend turning ignore-whitespace mode on for the diff; it's pretty clear when you do and baffling when you don't.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Build.BackEnd
     internal abstract class NodeProviderOutOfProcBase
     {
         /// <summary>
-        /// The maximum number of bytes to write 
+        /// The maximum number of bytes to write
         /// </summary>
         private const int MaxPacketWriteSize = 1048576;
 
@@ -64,7 +64,7 @@ namespace Microsoft.Build.BackEnd
         internal delegate void NodeContextTerminateDelegate(int nodeId);
 
         /// <summary>
-        /// The build component host. 
+        /// The build component host.
         /// </summary>
         protected IBuildComponentHost ComponentHost
         {
@@ -209,7 +209,7 @@ namespace Microsoft.Build.BackEnd
                 Stream nodeStream = TryConnectToProcess(nodeProcess.Id, 0 /* poll, don't wait for connections */, hostHandshake, clientHandshake);
                 if (nodeStream != null)
                 {
-                    // Connection successful, use this node.   
+                    // Connection successful, use this node.
                     CommunicationsUtilities.Trace("Successfully connected to existed node {0} which is PID {1}", nodeId, nodeProcess.Id);
                     return new NodeContext(nodeId, nodeProcess.Id, nodeStream, factory, terminateNode);
                 }
@@ -311,6 +311,8 @@ namespace Microsoft.Build.BackEnd
         }
 
 #if FEATURE_NAMED_PIPES_FULL_DUPLEX
+
+#if !FEATURE_PIPEOPTIONS_CURRENTUSERONLY
         //  This code needs to be in a separate method so that we don't try (and fail) to load the Windows-only APIs when JIT-ing the code
         //  on non-Windows operating systems
         private void ValidateRemotePipeSecurityOnWindows(NamedPipeClientStream nodeStream)
@@ -329,7 +331,7 @@ namespace Microsoft.Build.BackEnd
                 throw new UnauthorizedAccessException();
             }
         }
-
+#endif
 
         /// <summary>
         /// Attempts to connect to the specified process.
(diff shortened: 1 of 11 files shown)


rainersigwald:
```
Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.ForceSomeOutOfDate
Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.ForceOutOfDate
Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.ForceSomeOutOfDate
Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.ForceOutOfDate
```

These seem to be flaky.

rainersigwald:
@Microsoft/msbuild-maintainers I think this is ready to go. Please review when you get a chance.

rainersigwald:
fyi @pjanotti

src/Build/Microsoft.Build.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(LibraryTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(RuntimeOutputTargetFrameworks)</TargetFrameworks>
```
cdmihai: Why does this require going from netstandard to netcoreapp?

cdmihai: Ah, found the reason in the commit message :)

---
In reply to: [176602885](https://github.com/Microsoft/msbuild/pull/3106#discussion_r176602885) [](ancestors = 176602885)

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -290,16 +290,15 @@ private List<Process> GetPossibleRunningNodes(out string msbuildName, ref string
                 msbuildLocation = "MSBuild.exe";
             }
 
-
             msbuildName = Path.GetFileNameWithoutExtension(msbuildLocation);
 
-            List<Process> nodeProcesses = new List<Process>(Process.GetProcessesByName(msbuildName));
+            var expectedProcessName = Path.GetFileNameWithoutExtension(GetCurrentHost()) ?? msbuildName;
+
+            List<Process> nodeProcesses = new List<Process>(Process.GetProcessesByName(expectedProcessName));
```
cdmihai: In case there's a host, is it worth probing it's commandline to further filter out the non msbuild ones, to avoid the time wasted trying to connect to other processes, like the vbcscompiler? :)

src/Shared/UnitTests/TestAssemblyInfo.cs
```diff
@@ -44,11 +44,33 @@ public MSBuildTestAssemblyFixture()
 
         //  Use a project-specific temporary path
         //  This is so multiple test projects can be run in parallel without sharing the same temp directory
-        string newTempPath = Path.Combine(AppContext.BaseDirectory, "TestTemp");
+        var subdirectory = Path.GetRandomFileName();
 
-        _testEnvironment.CreateFolder(newTempPath);
+        string newTempPath = Path.Combine(Path.GetTempPath(), subdirectory);
+        var assemblyTempFolder = _testEnvironment.CreateFolder(newTempPath);
```
cdmihai: Hopefully the file name constraints will always be a superset of the directory name constraints.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -487,88 +491,119 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
...
-            processStartInfo.CreateNoWindow = (creationFlags | BackendNativeMethods.CREATENOWINDOW) == BackendNativeMethods.CREATENOWINDOW;
-            processStartInfo.UseShellExecute = false;
-
-            Process process;
-            try
+            if (NativeMethodsShared.IsWindows)
             {
-                process = Process.Start(processStartInfo);
+                // Repeat the executable name _again_ because Core MSBuild expects it
+                commandLineArgs = exeName + " " + commandLineArgs;
```
cdmihai: commandLineArgs is written to 3 times in this method. Maybe it would be more readable to extract its construction out

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
rainersigwald: I think so -- I'd like to pursue this in a future version. Unfortunately, there's no corefx API for getting another process's command line, so it'll have to be OS-specific.

rainersigwald: Filed #3125 to follow up on this.

rainersigwald:
@riezebosch that sounds bad. Can you please file a new issue with instructions about how to reproduce it?

PR #3107, opened 2018-03-19
 Fix serialization issue with AssemblyNameExtension

AndyGerlicher:
There were some subtle issues with AssemblyNameExtension serialization:
* Added hasAN and hasCI to indicate whether an AssemblyName or CultureInfo was serialized (and used them correctly).
* Switched from CultureInfo Name -> LCID. The default invariant culture has no Name, but has an LCID.
* Serialized the remappedFrom HashSet field.
* Added unit tests to verify serialization is correct.

It wasn't trivial to repro (originally) so I verified a lot of this manually and did some cleanup on the way. Verified it fixes the issue in the repro. I removed `FEATURE_ASSEMBLYNAME_CULTUREINFO` since it was only used in AssemblyNameExtension and no longer needed for `netstandard2.0`

### DIFF
--- a/src/Shared/AssemblyNameExtension.cs
+++ b/src/Shared/AssemblyNameExtension.cs
@@ -350,15 +350,7 @@ namespace Microsoft.Build.Shared
             {
                 // Is there a string?
                 CreateAssemblyName();
-#if FEATURE_ASSEMBLYNAME_CULTUREINFO
                 return asAssemblyName.CultureInfo;
-#else
-                if (asAssemblyName.CultureName == null)
-                {
-                    return null;
-                }
-                return new CultureInfo(asAssemblyName.CultureName);
-#endif
             }
         }
 
@@ -752,7 +744,6 @@ namespace Microsoft.Build.Shared
         internal static bool CompareCultures(AssemblyName a, AssemblyName b)
         {
             // Do the Cultures match?
-#if FEATURE_ASSEMBLYNAME_CULTUREINFO
             CultureInfo aCulture = a.CultureInfo;
             CultureInfo bCulture = b.CultureInfo;
             if (aCulture == null)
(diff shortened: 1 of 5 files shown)


src/Shared/AssemblyNameExtension.cs
```diff
@@ -140,7 +141,12 @@ private AssemblyNameExtension(SerializationInfo info, StreamingContext context)
                 var flags = (AssemblyNameFlags) info.GetInt32("flags");
                 var processorArchitecture = (ProcessorArchitecture) info.GetInt32("cpuarch");
 
-                var cultureInfoName = info.GetString("ciName");
+                CultureInfo cultireInfo = null;
+                var hasCultureInfo = info.GetBoolean("hasCI");
+                if (hasCultureInfo)
+                {
+                    cultireInfo = new CultureInfo(info.GetInt32("ci"));
```
rainersigwald: typo: cultire

src/Shared/UnitTests/AssemblyNameEx_Tests.cs
```diff
@@ -676,6 +679,69 @@ public void VerifyReverseVersionComparer()
...
+
+            using (MemoryStream ms = new MemoryStream(bytes))
+            {
+                BinaryFormatter formatter = new BinaryFormatter();
+                assemblyNameDeserialized = (AssemblyNameExtension)formatter.Deserialize(ms);
+            }
+
+            assemblyNameDeserialized.Equals(assemblyNameOriginal).ShouldBeTrue();
+            assemblyNameDeserialized.RemappedFromEnumerator.Count().ShouldBe(1);
+            assemblyNameDeserialized.RemappedFromEnumerator.First().Equals(assemblyRemappedFrom).ShouldBeTrue();
```
rainersigwald: Why not just

```c#
assemblyNameDeserialized.RemappedFromEnumerator.First().ShouldBe(assemblyRemappedFrom);
```

src/Shared/UnitTests/AssemblyNameEx_Tests.cs
```diff
@@ -676,6 +679,69 @@ public void VerifyReverseVersionComparer()
             Assert.True(assemblies[1].Equals(z));
             Assert.True(assemblies[2].Equals(y));
         }
+
+        [Theory]
+        [InlineData("System.Xml")]
+        [InlineData("System.XML, Version=2.0.0.0")]
+        [InlineData("System.Xml, Culture=de-DE")]
+        [InlineData("System.Xml, Version=10.0.0.0, Culture=en, PublicKeyToken=b03f5f7f11d50a3a, Retargetable=Yes")]
+        [InlineData("System.Drawing, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a")]
```
rainersigwald: Which of these would have failed before?

AndyGerlicher: That's on my list to do :)

AndyGerlicher:

PR #3112, opened 2018-03-20
Expand logFile at the end of the script

AndyGerlicher:
If the file didn't exist Get-ChildItem (what was using to expand the
full path) fail and logFile would be null. Moving that to the end.


jeffkl:
I think if you put "skip CI please" in the PR you won't have to wait for a build

PR #3113, opened 2018-03-20
⬆️ Update to CLI 2.1.300-preview3-008387

rainersigwald:
This enables SuppressNETCoreSdkPreviewMessage.


cdmihai:
Huh, I would expect core to fail in case of corefx breaking changes, but not full framework :)

rainersigwald:
Let's just go to preview2 RTM once it's available.

PR #3114, opened 2018-03-20
⬆️ Shouldly 3.0.0

rainersigwald:


rainersigwald:
Yes, definitely.

AndyGerlicher:
Does this fix or improve anything? Just curious.

rainersigwald:
@AndyGerlicher Not that I know of. I saw it was out when I went looking to see if they'd added support for embedded PDBs, which every assertion failure on .NET Core now complains about.

I'm happy to wait on this past the 15.8 fork point, just thought I'd propose it while thinking of it.

PR #3115, opened 2018-03-21
Fix XmlDocument reader to allow opening files in read-only mode

stan-sz:
XmlDocumentWithLocation class has the ability to openXML documents in the read-only mode (comments and whitespace is replaced with an empty string). Currently this functionaly is not available, because the XML file location is not passed to the XML reader constructor and thus not available for the XmlDocumentWithLocation for further verification.
Additionally, the IsReadOnly property is now overriden, to correctly inform callers about the read mode of the document.
Added unit tests that cover these scenarios.

### DIFF
--- a/src/Build.UnitTests/Construction/ElementLocation_Tests.cs
+++ b/src/Build.UnitTests/Construction/ElementLocation_Tests.cs
@@ -363,6 +363,7 @@ namespace Microsoft.Build.UnitTests.Construction
             {
                 var doc = new XmlDocumentWithLocation(loadAsReadOnly: true);
                 doc.Load(_pathToCommonTargets);
+                Assert.True(doc.IsReadOnly);
                 doc.Save(FileUtilities.GetTemporaryFile());
             }
            );
@@ -389,6 +390,7 @@ namespace Microsoft.Build.UnitTests.Construction
                 doc.Load(xmlReader);
             }
 #endif
+            Assert.True(doc.IsReadOnly);
             Assert.Throws<InvalidOperationException>(() => {
                 doc.Save(new MemoryStream());
             });
@@ -414,6 +416,7 @@ namespace Microsoft.Build.UnitTests.Construction
                 doc.Load(xmlReader);
             }
 #endif
+            Assert.True(doc.IsReadOnly);
             Assert.Throws<InvalidOperationException>(() =>
             {
                 doc.Save(new StringWriter());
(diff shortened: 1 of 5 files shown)


src/Build.UnitTests/Construction/ElementLocation_Tests.cs
```diff
@@ -363,6 +363,7 @@ public void SaveReadOnly1()
             {
                 var doc = new XmlDocumentWithLocation(loadAsReadOnly: true);
                 doc.Load(_pathToCommonTargets);
+                Assert.True(doc.IsReadOnly);
```
jeffkl: This is testing that we get an InvalidOperationException on save, you should add a new unit test that just tests the IsReadOnly property and not do these checks in the SaveReadOnly tests.

src/Build.UnitTests/Construction/ProjectRootElement_Tests.cs
```diff
@@ -52,5 +51,59 @@ public void IsEmptyXmlFileReturnsFalse(string contents, bool useByteOrderMark)
...
+                Assert.Equal(1, children.Count);
+                Assert.Equal("Project", children[0].Name);
+                Assert.Equal(2, children[0].ChildNodes.Count);
+                Assert.Equal("Initial Comment", children[0].ChildNodes[0].Value);
+                Assert.Equal("Ending Comment", children[0].ChildNodes[1].Value);
+            }
+        }
+
+        [Fact]
+        public void ProjectLoadedStrippingCommentsAndWhiteSpaceIsReadOnly()
```
jeffkl: 🥇

rainersigwald:
This looks related:

```
System.ArgumentNullException: Value cannot be null.
Parameter name: value
   at System.String.StartsWith(String value, StringComparison comparisonType)
   at Microsoft.Build.Construction.XmlDocumentWithLocation.DetermineWhetherToLoadReadOnly(String fullPath) in /_/src/Build/ElementLocation/XmlDocumentWithLocation.cs:line 368
   at Microsoft.Build.Construction.XmlDocumentWithLocation.Load(XmlReader reader) in /_/src/Build/ElementLocation/XmlDocumentWithLocation.cs:line 146
   at Microsoft.Build.Construction.ProjectRootElement.LoadDocument(String fullPath, Boolean preserveFormatting) in /_/src/Build/Construction/ProjectRootElement.cs:line 2114
   at Microsoft.Build.Construction.ProjectRootElement..ctor(String path, ProjectRootElementCache projectRootElementCache, Boolean preserveFormatting) in /_/src/Build/Construction/ProjectRootElement.cs:line 235
```

src/Build.UnitTests/Construction/ElementLocation_Tests.cs
stan-sz: This PR adds dedicated tests for the IsReadOnly property. The additional asserts I added hare are asserting that these test cases are really operating on a read-only object.

PR #3126, opened 2018-03-23
Microsoft.Build: netstandard2.0 instead of netcoreapp2.0

rainersigwald:
It's harder for projects to depend on us if we're netcoreapp2.0, because
they must themselves be .NET Core specific rather than netstandard.

We only need to be netcoreapp2.1, though--in 2.0 mode we can be
netstandard.

### DIFF
--- a/src/Build/Microsoft.Build.csproj
+++ b/src/Build/Microsoft.Build.csproj
@@ -38,11 +38,11 @@
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
     <PackageReference Include="System.Diagnostics.TraceSource" Version="4.0.0" />
-    <PackageReference Include="System.IO.Pipes.AccessControl" Version="4.3.0" />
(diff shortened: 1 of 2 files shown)


rainersigwald:
That cannot be related

src/Build/Microsoft.Build.csproj
```diff
@@ -1,6 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(RuntimeOutputTargetFrameworks)</TargetFrameworks>
+    <!-- Node reuse requires an API new to .NET Core 2.1 not yet available
+         in .NETStandard. -->
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
```
cdmihai: How does this not break our netcoreapp2.0 target? I assume that one's also getting tested, right?

rainersigwald: The netcoreapp2.0 target uses the netstandard2.0 library from here.

PR #3130, opened 2018-03-24
Escape hatch: share stdout with node children

rainersigwald:
The VSTest runner emits log messages to its stdout. So far, in .NET
Core, that has resulted in them getting piped all the way up through
the worker MSBuild that launched the task, the entry-point MSBuild
coordinating the build, and the `dotnet test` invocation that started
it all.

The introduction of node reuse for .NET Core makes this untenable:
worker nodes should not share stdout with the process that happened to
launch them, because they will be long-lived and may do entirely
unrelated builds.

But for now, this breaks the `dotnet test` scenario. This commit creates
an escape hatch environment variable
`MSBUILDENSURESTDOUTFORTASKPROCESSES` that can be used in combination
with `/nodereuse:false` to create a cone of MSBuild processes that all
share std handles.

Enables workaround for Microsoft/vstest#1503.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -23,6 +23,7 @@ using Microsoft.Build.Internal;
 
 using BackendNativeMethods = Microsoft.Build.BackEnd.NativeMethods;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


rainersigwald:
cc @AndyGerlicher @dsplaisted @livarcocc

AndyGerlicher:
Looks good, but I want #3131 before we do a build.

PR #3136, opened 2018-03-26
Get PGO data from the vs15.7 release branch

rainersigwald:
This is much closer to our insertion point than the VS master branch, so it has many fewer "I don't know how to correlate this old trace with these new codepaths" problems.

It looks very good. Starred is a build from the experimental branch here, everything else is current builds of d15.7stg:

image


rainersigwald:

PR #3148, opened 2018-03-28
Bump version to 15.8

rainersigwald:


src/Shared/Constants.cs
```diff
@@ -87,7 +87,7 @@ internal static class MSBuildConstants
         internal const string CurrentAssemblyVersion = Microsoft.VisualStudio.Internal.BrandNames.VSGeneralAssemblyVersion;
 #endif
 
-        internal const string CurrentAssemblyFileVersion = "15.7.0.0";
+        internal const string CurrentAssemblyFileVersion = "15.8.0.0";
```
AndyGerlicher: Do we actually use this anymore? Since GitVersioning works on Core I would hope we don't?

rainersigwald: I wondered that too; let me poke around a bit.

Issue #1467, opened 2016-12-10
Null reference exception when shutting down worker nodes


BillHiebert:
The exception is caught, but it seems like it is some kind of bug.
1. From VS 2017 create a new ASP.NET Core web application (empty is fine)
2. Attach a debugger and enable it to break on NullReferenceExceptions
3. Close VS

Null reference exception with this stack:

>	Microsoft.Build.dll!Microsoft.Build.BackEnd.NodeProviderOutOfProc.ShutdownAllNodes() Line 180	C#
 	Microsoft.Build.dll!Microsoft.Build.BackEnd.NodeManager.ShutdownAllNodes() Line 185	C#
 	Microsoft.Build.dll!Microsoft.Build.Execution.BuildManager.ShutdownAllNodes() Line 718	C#
 	Microsoft.VisualStudio.CommonIDE.dll!Microsoft.VisualStudio.CommonIDE.BuildManager.BuildManagerAccessor.Shutdown()	Unknown
 	Microsoft.VisualStudio.CommonIDE.dll!Microsoft.VisualStudio.CommonIDE.BuildManager.BuildManagerPackage.ShutdownEventHandlers()	Unknown
 	Microsoft.VisualStudio.CommonIDE.dll!Microsoft.VisualStudio.CommonIDE.BuildManager.BuildManagerPackage.Dispose(bool disposing)	Unknown
 	Microsoft.VisualStudio.Shell.15.0.dll!Microsoft.VisualStudio.Shell.Package.Microsoft.VisualStudio.Shell.Interop.IVsPackage.Close()	Unknown

rainersigwald:
This is because we create a new `NodeManager` (with no `BuildParameters`) on shutdown:

```
 1      Microsoft.Build.dll!Microsoft.Build.BackEnd.NodeProviderOutOfProc.InitializeComponent(Microsoft.Build.BackEnd.IBuildComponentHost host = {Microsoft.Build.Execution.BuildManager})
 2      Microsoft.Build.dll!Microsoft.Build.BackEnd.BuildComponentFactoryCollection.BuildComponentEntry.GetInstance(Microsoft.Build.BackEnd.IBuildComponentHost host)
 3      Microsoft.Build.dll!Microsoft.Build.BackEnd.BuildComponentFactoryCollection.GetComponent(Microsoft.Build.BackEnd.BuildComponentType type)
[...]

PR #3151, opened 2018-03-29
Support ShutdownAllNodes on .NET Core

rainersigwald:
There were two problems blocking shutdown of idle worker nodes on .NET core:

* #1467
* We only looked for nodes with process name "MSBuild.exe", which didn't catch those running in the dotnet host

This fixes both.

I recommend going by commit, since they're individually small transforms.

cc @peterhuene


src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -198,12 +198,8 @@ public void VerifyEnvironmentSavedBetweenCalls()
...
         /// the system before this method was called.
         /// </summary>
-#if RUNTIME_TYPE_NETCORE
-        [Theory(Skip = "https://github.com/Microsoft/msbuild/issues/1975")]
-#elif MONO
-        [Theory(Skip = "https://github.com/Microsoft/msbuild/issues/1240")]
-#else
-        [Theory(Skip = "https://github.com/Microsoft/msbuild/issues/2057")]
+#if FEATURE_NODE_REUSE
+        [Theory]
```
rainersigwald: @AndyGerlicher So I didn't really do any new robustness improvements for this test but I hate to leave it off . . . thoughts? I could enable the test only in master so as not to torpedo potential future 15.7 changes?

rainersigwald: yup, saw the same problem on my Linux machine just now. Pulled that change out of here.

PR #3158, opened 2018-04-02
Don't steal nodes when not reusing nodes

### LINKED ISSUE #3140: MSBuild should allow CI tools to isolate invocations
To be robust CI tools need to keep build/test processes from interfering with each other:
* Cleanup left over processes from old runs
* Kill process trees after a timeout

The way I have implemented this is to (pseudo code):
```
// Launch process with CREATE_SUSPENDED
...

// Setup sandbox job object
HANDLE SandboxJob = CreateJobObjectW(nullptr, JobName);
AssignProcessToJobObject(SandboxJob, pi.hProcess);

JOBOBJECT_EXTENDED_LIMIT_INFORMATION LimitInfo;
LimitInfo.BasicLimitInformation.LimitFlags |= JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
SetInformationJobObject(Job, JobObjectExtendedLimitInformation, &LimitInfo, sizeof(LimitInfo));

ResumeThread(pi.hThread);
// Let process run until exit
...

// Cleanup any processes left in job
TerminateJobObject(SandboxJob, 255);
```

The problem you run into with MSBuild is that even if you specify `/nodeReuse:false` it will still share nodes between MSBuild.exe invocations.

### Current  behavior
```
Process             Job Object
MSBuild.exe         Build 1
    MSBuild.exe     Build 1
    MSBuild.exe     Build 1
    MSBuild.exe     Build 1
    MSBuild.exe     Build 1
MSBuild.exe         Build 2   < Uses nodes created in Build 1
```
After the first MSBuild.exe exits all nodes will also be killed and the second build will fail with:
`MSBUILD : error MSB4166: Child node "7" exited prematurely`

### Proposed behavior
To workaround this I previously injected code into the build process to make the MSBuild communication pipes unique. This has worked well, but the solution is problematic and prone to break as APIs and software evolve.

A relatively simple fix for this would be to allow the CI tool to make the communication pipes unique. See: https://github.com/Malterlib/msbuild/commit/079b2e1abf5a0fdb058e0d287bf6a86730d49438
```
Process             Job Object
MSBuild.exe         Build 1    < CI tool have set a unique MSBUILD_PIPE_NAME for this job object
    MSBuild.exe     Build 1
    MSBuild.exe     Build 1
    MSBuild.exe     Build 1
    MSBuild.exe     Build 1
MSBuild.exe         Build 2    < CI tool have set a unique MSBUILD_PIPE_NAME for this job object
    MSBuild.exe     Build 2
    MSBuild.exe     Build 2
    MSBuild.exe     Build 2
    MSBuild.exe     Build 2
```
Another option could be to exclude node reuse if you detect that the other node is inside another job object.

rainersigwald:
Previously, node reuse was part of the handshake between nodes but two
MSBuild process cones started at roughly the same time could steal each
other's newly-created processes:

 * MSBuild A would start a process with `/nr:false`
 * MSBuild A' would start and wait for a connection
 * MSBuild B would look for live nodes that accepted its handshake
 * MSBuild B would connect to and handshake with MSBuild A'
 * MSBuild A would try to connect to MSBuild A', fail, and start another
   new process.

Prevent this by only attempting to connect to processes we started when
using `/nr:false`. Fixes #3140.

The only change is wrapping the formerly `#if`ed connect-to-idle-node code with an additional runtime check. Whitespace makes it look bigger.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -172,33 +172,37 @@ namespace Microsoft.Build.BackEnd
             }
 
 #if FEATURE_NODE_REUSE
-            var candidateProcesses = GetPossibleRunningNodes(msbuildLocation);
-
-            CommunicationsUtilities.Trace("Attempting to connect to each existing msbuild.exe process in turn to establish node {0}...", nodeId);
-            foreach (Process nodeProcess in candidateProcesses)
+            // Try to connect to idle nodes if node reuse is enabled.
+            if (_componentHost.BuildParameters.EnableNodeReuse)
             {
-                if (nodeProcess.Id == Process.GetCurrentProcess().Id)
-                {
-                    continue;
-                }
+                var candidateProcesses = GetPossibleRunningNodes(msbuildLocation);
 
-                // Get the full context of this inspection so that we can always skip this process when we have the same taskhost context
-                string nodeLookupKey = GetProcessesToIgnoreKey(hostHandshake, clientHandshake, nodeProcess.Id);
-                if (_processesToIgnore.Contains(nodeLookupKey))
+                CommunicationsUtilities.Trace("Attempting to connect to each existing msbuild.exe process in turn to establish node {0}...", nodeId);
+                foreach (Process nodeProcess in candidateProcesses)
                 {
-                    continue;
-                }
+                    if (nodeProcess.Id == Process.GetCurrentProcess().Id)
+                    {
+                        continue;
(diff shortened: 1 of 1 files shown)


rainersigwald:
@ryanbrandenburg @natemcmaster FYI, I put privates built off this commit into `~/.dotnet/sdk` in https://github.com/aspnet/BuildTools/pull/627 and the tests don't hang (since you're already using `/nr:false`).

natemcmaster:
Thanks @rainersigwald! Once we have an SDK with this fix, let me know and I can work on getting it into aspnet.

PR #3165, opened 2018-04-05
Changes expression evaluation so that static property functions can be used in transformation

cclauson:
As discussed on this issue:
https://github.com/Microsoft/msbuild/issues/3163

Based on having run into the same problem described on this StackOverflow:
https://stackoverflow.com/questions/5103026/in-msbuild-can-i-use-the-string-replace-function-on-a-metadata-item?lq=1

According to the discussion there, this worked before MSBuild 12.0.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -261,11 +261,14 @@ namespace Microsoft.Build.Evaluation
 
             ErrorUtilities.VerifyThrowInternalNull(elementLocation, "elementLocation");
 
-            string result = MetadataExpander.ExpandMetadataLeaveEscaped(expression, _metadata, options);
-            result = PropertyExpander<P>.ExpandPropertiesLeaveEscaped(result, _properties, options, elementLocation, _usedUninitializedProperties);
-            result = ItemExpander.ExpandItemVectorsIntoString<I>(this, result, _items, options, elementLocation);
-            result = FileUtilities.MaybeAdjustFilePath(result);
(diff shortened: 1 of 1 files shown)


cclauson:
The failures on windows say:
"20:37:40 [xUnit] [INFO] - [xUnit.Net-v2 (default)] - No test report file(s) were found with the pattern 'artifacts/**/TestResults/*.xml' relative to 'D:\j\w\innerloop_Win---325d1729' for the testing framework 'xUnit.Net-v2 (default)'.  Did you enter a pattern relative to the correct directory?  Did you generate the result report(s) for 'xUnit.Net-v2 (default)'?"

Is this an infrastructure problem?

cclauson:
Modified PR so that it made no changes to source, the same checks still failed, suggesting that this is an infrastructure problem.

rainersigwald:
@cclauson Other PRs and CI on the master branch are happily passing, so I do suspect this change. I suspect that the changes cause a difference in evaluation that causes our tests to not be built/run correctly.

Do all tests pass on your machine? I tried building your changes with `build -bootstraponly` (which uses MSBuild from `PATH` to build and directly tests the output, bypassing any possible evaluation changes for our own repo) and I see many failed tests that look related to this change.

cclauson:
@rainersigwald Thanks for the response!  I was having trouble getting tests to run locally before, I will take another pass at this and try to identify the problem.

cclauson:
Yup, not an example of great follow through on my part here.  Was eventually able to repro broken tests but didn't find the time to debug them.

Issue #3161, opened 2018-04-03
MSBuild can hang indefinitely on non-Windows platforms


rainersigwald:
### Steps to reproduce

Run `msbuild` (or more likely `dotnet build` or `dotnet msbuild` inside an `<Exec>`  task in a project, or run unit tests that exec msbuild inside a build, or run two multiproc builds concurrently.

### Expected  behavior

MSBuild does not connect to nodes that are busy doing a build, and doesn't hang.

### Actual behavior

MSBuild can hang when attempting to connect to an MSBuild process that is busy doing another build. In a multiple-builds situation, this can make a subsequent build wait for the results of a previously-started one, even though they share no data. In the nested-msbuild situation, it can mean an infinite hang while trying to talk to nodes above it in the process tree.

rainersigwald:
This is because of https://github.com/dotnet/corefx/issues/28791, which may not be fixable.

It was the cause of hangs in aspnet/BuildTools#627, which was worked around by #3158.

rainersigwald:
Ok, so with a connection-side timeout (https://github.com/rainersigwald/msbuild/commit/45837e199c4657a36e281bd9abd28c220a59dc82) the child MSBuild can start up and connect to its own children, but the outer MSBuild now hangs; I think because it's waiting on the inner MSBuild's idle worker processes.

I'm going to submit a PR to turn node reuse off by default on !Windows for preview 2.

rainersigwald:
Suspicion confirmed because the build completed after 15 minutes of idle time. But I don't know what caused the wait yet; it doesn't look like Linux normally has any awareness of grandchild processes.

rainersigwald:
Simple repro of the infinite-hang case: https://github.com/rainersigwald/SampleMSBuildProjects/tree/msbuild-inside-msbuild.

PR #3172, opened 2018-04-06
Avoid hangs with node reuse on non-Windows

rainersigwald:
There were two problems leading to #3161. I reverted the only-on-Windows mitigation for that and fixed the problems.

First, dotnet/corefx#28791 means that the named pipe connection behavior we'd been relying on on Windows didn't work on UNIXy platforms, because the pipe could be connected and it would just go nowhere. I fixed that by wrapping a timeout around the wait-for-handshake-reply code to give up if it isn't speedy.

Then, tests (in other repos) that ran `MSBuild /m /nr:true` (by default) were hanging for a very long time, because they were waiting for any newly-spawned worker nodes to hit their idle timeout. See 
the commit message of e604f5f for details.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -475,6 +475,9 @@ namespace Microsoft.Build.BackEnd
             {
                 if (!Traits.Instance.EscapeHatches.EnsureStdOutForChildNodesIsPrimaryStdout)
                 {
+                    // Redirect the streams of worker nodes so that this MSBuild.exe's
+                    // parent doesn't wait on idle worker nodes to close streams
+                    // after the build is complete.
                     startInfo.hStdError = BackendNativeMethods.InvalidHandle;
                     startInfo.hStdInput = BackendNativeMethods.InvalidHandle;
                     startInfo.hStdOutput = BackendNativeMethods.InvalidHandle;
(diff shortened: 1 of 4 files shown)


src/MSBuild/XMake.cs
```diff
@@ -545,7 +545,7 @@ string commandLine
 #endif
                 int cpuCount = 1;
 #if FEATURE_NODE_REUSE
-                bool enableNodeReuse = NativeMethodsShared.IsWindows;
```
cdmihai: nit: Shame this has to be repeated twice. It would be nice if BuildParameters had a Default singleton to access default values. But then it isn't immutable, so the Default can get mutated :(

src/Shared/CommunicationsUtilities.cs
```diff
@@ -368,6 +368,24 @@ internal static long ReadLongForHandshake(this PipeStream stream, byte[] leading
...
+            var readTask = stream.ReadAsync(bytes, 0, bytes.Length);
+            const int HandshakeReadTimeout = 30;
+
+            // Manual timeout here because the timeout passed to Connect() just before
+            // calling this method does not apply on UNIX domain socket-based
+            // implementations of PipeStream.
+            // https://github.com/dotnet/corefx/issues/28791
+            if (!readTask.Wait(HandshakeReadTimeout))
+            {
+                throw new IOException(string.Format(CultureInfo.InvariantCulture, "Did not recieve return handshake in {0}ms", HandshakeReadTimeout));
```
cdmihai: Typo: "recieve"

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -510,6 +510,9 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
                 processStartInfo.Arguments = commandLineArgs;
                 if (!Traits.Instance.EscapeHatches.EnsureStdOutForChildNodesIsPrimaryStdout)
                 {
+                    processStartInfo.RedirectStandardInput = true;
```
cdmihai: nit: I'd move the commit commit into the code here.

src/MSBuild/XMake.cs
rainersigwald: Yeah, I felt the same way when I was making the original temporary change I'm undoing here. I think some of them could have been avoided but was making a minimum change.

AndyGerlicher:

PR #3185, opened 2018-04-11
⬆️ Update to CLI 2.1.300-preview2

rainersigwald:
This enables `SuppressNETCoreSdkPreviewMessage` and should get us in a stable place for prodcon, but I'm still working on the latter.

A bunch of test disablements here, because of breaking changes in .NET Core 2.1 path handling.

### DIFF
--- a/build/SignToolData.json
+++ b/build/SignToolData.json
@@ -40,7 +40,7 @@
             ]
         },
         {
-            "certificate": null,
+            "certificate": "NuGet",
             "strongName": null,
             "values": [
                 "packages/*.nupkg"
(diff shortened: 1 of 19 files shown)


rainersigwald:
Since the Mac CI lab is pokey, I'm testing on my laptop.

rainersigwald:
Passed on my Mac.

rainersigwald:
Per offline discussion with @AndyGerlicher, retargeting to `vs15.7` since we've shipped .NET Core SDK 2.1.300-preview2, which was why we created the -preview4 branch. We plan to ship preview4 and higher out of vs15.7 as we have done before.

I'll just let this commit merge in all of the existing preview4 commits, too.

rainersigwald:
Let's just take everything at once via #3188.

PR #3190, opened 2018-04-12
Log a telemetry event on SDK resolution

AndyGerlicher:
Log the SDK name, requested version, and the actual version returned
from the resolver service for an SDK. Note this will fire once per SDK
per build request or in VS once per new evaluation context.


cdmihai:
Until VS picks up the Evaluation context, it will fire twice per evaluated project. In addition, I don't think anyone's listening to telemetry events in VS. Which actually brings an interesting idea: should VS filter these out in their logging forwarders to avoid needless event shuffling?

@lifengl

PR #3194, opened 2018-04-13
RI vs15.7

rainersigwald:


rainersigwald:
Closed because will require manual merge (without touching the stabilization branch), which can't be done on GH.

PR #3199, opened 2018-04-16
GetReferenceAssemblyPaths needs GetFrameworkPaths

### LINKED ISSUE #3196: GetReferenceAssemblyPaths target should depend on GetFrameworkPaths
See https://github.com/dotnet/sdk/issues/1730.  Normally `GetFrameworkPaths` will run before `GetReferenceAssemblyPaths`, but if you have a target that runs early and only depends on the latter, then `GetFrameworkPaths` may not run and you will get an error such as:

> C:\Program Files\dotnet\sdk\2.1.1-preview-007128\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.ConflictResolution.targets(58,5): error : Framework list file path 'v4.0.30319\RedistList\FrameworkList.xml' is not rooted. Only full paths are supported. [D:\Temp\repro_pack\repro_pack.csproj]

I've submitted a [PR](https://github.com/dotnet/sdk/pull/2145) to fix the target dependencies in the SDK, but it would be good for the target dependency to be updated in MSBuild itself.

rainersigwald:
GetReferenceAssemblyPaths references
`@(_CombinedTargetFrameworkDirectoriesItem)` and other properties/items
that get defined (for .NET projects) in `GetFrameworkPaths`, but did
not depend on it. Dependencies were enforced by listing
GetFrameworkPaths first whenever dependencies existed, but that is hard
to enforce, and when missed caused dotnet/sdk#1730.

I'm adding this explicitly rather than appending to the already-existent
but undefined `$(GetReferenceAssemblyPathsDependsOn)` because a GitHub
search found a few projects (mostly Mono-related) that _override_ rather
than append to GetReferenceAssemblyPathsDependsOn. That was safe because
no released MSBuild ever defined anything in that property, but could
introduce problems if we started defining it.

Fixes #3196.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1168,7 +1168,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     -->
   <Target
       Name="GetReferenceAssemblyPaths"
-      DependsOnTargets="$(GetReferenceAssemblyPathsDependsOn)">
+      DependsOnTargets="$(GetReferenceAssemblyPathsDependsOn);GetFrameworkPaths">
 
     <!-- if TargetFrameworkDirectory doesn't have a custom value, clear it out; that way we can get reference paths and target framework directories all in the right order -->
     <PropertyGroup>


rainersigwald:
Another possiblity would be appending to `$(GetReferenceAssemblyPathsDependsOn)` in `Microsoft.NETFramework.CurrentVersion.targets` where `GetFrameworkPaths` is defined. That wouldn't be a comprehensive fix--it'd still be susceptible to complete-overrides like the Mono things I found above--but it'd have somewhat better scoping.

And since almost everything has the double-dependencies set up (how it's been working for a long time) the possiblity of overriding to lose this new link doesn't seem totally catastrophic.

Preferences on approach?

cc @dsplaisted

rainersigwald:

Issue #3210, opened 2018-04-18
Reduce GC pressure in InterningBinaryReader.ReadString()


lifengl:
From trace when we open a .Net Core solution, we noticed that more than 10% virtual allocation was triggered by creating the StringBuilder in Microsoft.Build.InteringBinaryReader.ReadString()

It appears a part of msbuild code has already tried to share stringBuffer.  Can we do the same thing here?  It could reduce GC press in the design time build phase right after loading a solution.

If we count the phase right after loading solution, this code path contributes 26.8% of virtual allocations.  Although the memory allocation is temporary, it adds lot of GC pressure during that phase.

lifengl:
Traces shows this stack

\|     +   microsoft.build.ni!InterningBinaryReader.ReadString
--
\|      +   microsoft.build.ni!Microsoft.Build.BackEnd.NodePacketTranslator+NodePacketReadTranslator.Translate(System.String   ByRef)

alone with a few other stacks:

Name
--
\|    \|  +   mscorlib.ni!BinaryWriter.Write
\|    \|     +   microsoft.build.ni!Microsoft.Build.BackEnd.NodePacketTranslator+NodePacketWriteTranslator.Translate(System.String   ByRef)


\|    + microsoft.build.ni!Microsoft.Build.BackEnd.NodeProviderOutOfProcBase+NodeContext.HeaderReadComplete(System.IAsyncResult)
--


Name
--
\|     + microsoft.build.ni!InterningBinaryReader.ReadString
\|      + microsoft.build.framework.ni!ProjectStartedEventArgs.CreateFromStream
\|      + microsoft.build.ni!Microsoft.Build.BackEnd.NodePacketTranslator+NodePacketReadTranslator.Translate(System.String ByRef)

Are the major contributor to the LOH heap during the DT build time after loading a large solution.  That happens when the size of the solution is bigger to the point, that the solution configuration xml string becomes very bigger, so all build packet buffers are all beyond the threshold and pushed to the LOH.  There are also large amount of DT build for large solutions, and maybe multiple build nodes also contributes to this.  The code path above quickly allocates lots of memory, and due to LOH, they are not recycled until Gen2 GC.  This creates a huge memory pressure (in one dump, the LOH grow to 600M), and combine with other allocations, it causes VS to run out of memory.

rainersigwald:
Looks like our `ReusableStringBuilder` isn't a great solution for this for a couple of reasons:

[...]

PR #3213, opened 2018-04-19
WIP: Cache reads in InterningBinaryReader

rainersigwald:
This passes tests but I don't know if it helps with #3210 at all; that needs to be tested.


rainersigwald:
Let's revisit this as part of the work we plan to do around memory reductions and `ReusableStringBuilder`, @BenVillalobos.

Issue #3214, opened 2018-04-20
MSBUILD : error MSB4223: A node of the required type OutOfProc could not be created.


natemcmaster:
Our CI builds started failing with a bunch of errors that say.

```
MSBUILD : error MSB4223: A node of the required type OutOfProc could not be created.
```

### Steps to reproduce
```
git clone https://github.com/aspnet/Universe
git checkout 983eb37cd61bee4f97f9bc44914575a9cad150b2
build.cmd /p:SkipTests=true /p:CI=true
```

### Expected  behavior
Builds should pass. Only [difference in this commit](https://github.com/aspnet/Universe/compare/983eb37cd61bee4f97f9bc44914575a9cad150b2~1...983eb37cd61bee4f97f9bc44914575a9cad150b2) is that we updated to MSBuild version 15.7.169.51914 (via .NET Core CLI 2.1.300-preview3-008646). The previous commit builds with MSBuild  version 15.7.145.53551  (via .NET Core CLI 2.1.300-preview2-008530)

### Actual behavior
About 15-20 minutes into the build, MSBuild begins to fail with 
```
MSBUILD : error MSB4223: A node of the required type OutOfProc could not be created.
```

Internal build logs available here: http://aspnetci/viewLog.html?buildId=452289&tab=buildResultsDiv&buildTypeId=Releases_21Public_UpdateUniverse and again http://aspnetci/viewLog.html?buildId=452160&tab=buildResultsDiv&buildTypeId=Releases_21Public_UniverseCoherence.

We reverted to the 2.1.300-preview2 CLI to workaround for now.

### Environment data
`msbuild /version` output:
```
MSBuild version 15.7.169.51914 for .NET Core
```

OS info: Windows Server 2012 R2

If applicable, version of the tool that invokes MSBuild (Visual Studio, dotnet CLI, etc): dotnet CLI

cc @rainersigwald possibly related to node reuse changes?

rainersigwald:
All on Windows, right? https://github.com/Microsoft/msbuild/commit/979ef17199fb4a0ee7d9c4a92305a6ea2bac8914 added a timeout on Windows as well as *nix but I expected it to have no effect on Windows since we would have failed to connect before the timeout had time to happen. Maybe that was wrong.

[...]

PR #3215, opened 2018-04-20
Avoid handshake timeout on Windows

rainersigwald:
The old version of this code didn't have a timeout when reading the
handshake packet, and appeared to be reliable. The timeout was required
to work around a cross-platform design limitation: the pipe always
connects on *nix systems, even when nothing is listening to it.

This caused failures on Windows (#3214), because the timeout was
exceeded there. For maximum robustness, avoid the timeout entirely on
Windows.

### DIFF
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -355,34 +355,51 @@ namespace Microsoft.Build.Internal
         /// <summary>
         /// Extension method to read a series of bytes from a stream
         /// </summary>
-        internal static long ReadLongForHandshake(this PipeStream stream)
+        internal static long ReadLongForHandshake(this PipeStream stream
+#if NETCOREAPP2_1
+            , int handshakeReadTimeout
+#endif
+            )
         {
-            return stream.ReadLongForHandshake((byte[])null, 0);
+            return stream.ReadLongForHandshake((byte[])null, 0
+#if NETCOREAPP2_1
+                , handshakeReadTimeout
+#endif
+                );
         }
 
         /// <summary>
         /// Extension method to read a series of bytes from a stream.
         /// If specified, leading byte matches one in the supplied array if any, returns rejection byte and throws IOException.
         /// </summary>
-        internal static long ReadLongForHandshake(this PipeStream stream, byte[] leadingBytesToReject, byte rejectionByteToReturn)
+        internal static long ReadLongForHandshake(this PipeStream stream, byte[] leadingBytesToReject,
+            byte rejectionByteToReturn
+#if NETCOREAPP2_1
+            , int timeout
+#endif
+            )
         {
             byte[] bytes = new byte[8];
 
 #if NETCOREAPP2_1
             if (!NativeMethodsShared.IsWindows)
             {
+                // Enforce a minimum timeout because the Windows code can pass
+                // a timeout of 0 for the connection, but that doesn't work for
+                // the actual timeout here.
+                timeout = Math.Max(timeout, 50);
+
                 // A legacy MSBuild.exe won't try to connect to MSBuild running
                 // in a dotnet host process, so we can read the bytes simply.
                 var readTask = stream.ReadAsync(bytes, 0, bytes.Length);
-                const int HandshakeReadTimeout = 30;
 
                 // Manual timeout here because the timeout passed to Connect() just before
                 // calling this method does not apply on UNIX domain socket-based
                 // implementations of PipeStream.
                 // https://github.com/dotnet/corefx/issues/28791
-                if (!readTask.Wait(HandshakeReadTimeout))
+                if (!readTask.Wait(timeout))
                 {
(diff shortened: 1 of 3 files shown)


rainersigwald:
Note: still need testing on *nix to see if we need to up the timeout or abandon node reuse by default there again.

nguerrera:
> or abandon node reuse by default there again.

😢

rainersigwald:
Yeah, I'd like to avoid that, but I don't want to make builds flaky. Since we accidentally found a great stress test, @natemcmaster is running builds on Ubuntu to see if they fail in a similar way on their machines, or if it's somehow Windows-specific.

rainersigwald:
They did fail on Ubuntu the same way. Next push will have a compromise solution:

* Extend handshake timeout on connect-to-just-launched-node case to 30s, matching the Windows connect-to-pipe timeout.
* Do not use handshake timeouts on Windows where they aren't necessary.
* Extend handshake timeout _from the worker-node side_ to 60s, using an old constant there too. Since this timeout starts only after getting a pipe connection from another node, which immediately sends the packet, this should never come close to timing out.

rainersigwald:
On my Ubuntu 17.10 machine (8 logical procs), I built using 2.1.300-preview3-008646, and got overall failure, with **128** instances of `Did not receive return handshake in 30ms` in the comm log.

I built at aed1374 and see only **4** instances of that in the logging, so this change looks good. However, the build still failed with

```
The process cannot access the file '/home/raines/src/aspnet/Universe/modules/Razor/src/Microsoft.AspNetCore.Razor.Language/bin/Release/netstandard2.0/Microsoft.AspNetCore.Razor.Language.dll' because it is being used by another process.
```

Which I assume is unrelated. Retrying.

rainersigwald:
Nope, something's still busted; I'm seeing 50ms timeouts even on just-created processes that should be 30000ms. Looking . . .

rainersigwald:

```
Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.ForceSomeOutOfDate
```

src/Shared/CommunicationsUtilities.cs
```diff
@@ -355,64 +355,86 @@ internal static void WriteLongForHandshake(this PipeStream stream, long value)
         /// <summary>
         /// Extension method to read a series of bytes from a stream
         /// </summary>
-        internal static long ReadLongForHandshake(this PipeStream stream)
+        internal static long ReadLongForHandshake(this PipeStream stream
+#if NETCOREAPP2_1
+            , int handshakeReadTimeout
```
cdmihai: What if instead of ifdefs you leave the timeout in and pass 0 on non-netcore?

src/Shared/NodeEndpointOutOfProcBase.cs
```diff
@@ -422,7 +422,11 @@ private void PacketPumpProc()
                     // it will cause the host to reject us; new hosts expect 00 and old hosts expect F5 or 06).
                     try
                     {
-                        long handshake = localReadPipe.ReadLongForHandshake(/* reject these leads */ new byte[] { 0x5F, 0x60 }, 0xFF /* this will disconnect the host; it expects leading 00 or F5 or 06 */);
+                        long handshake = localReadPipe.ReadLongForHandshake(/* reject these leads */ new byte[] { 0x5F, 0x60 }, 0xFF /* this will disconnect the host; it expects leading 00 or F5 or 06 */
+#if NETCOREAPP2_1
+                            , ClientConnectTimeout /* wait a long time for the handshake from this side */
```
cdmihai: Was this a dead field before? Curious when / how it was used.

src/Shared/CommunicationsUtilities.cs
rainersigwald: The intention for the ifdefs is to avoid touching the output of the full framework build since we don't _have_ to and it's in escrow. It's definitely not nice clean code :(

src/Shared/NodeEndpointOutOfProcBase.cs
rainersigwald: I couldn't find any use of it before.

rainersigwald: (But it was basically exactly what I wanted, so I used it. There's no principled reason for it to be double the other timeout, though.)

rainersigwald:
Just discussed with shiproom: approved for 2.1.300-RC.1.

LanPodder:
Using both node and dotnet core from binaries, setting env variables and running dotnet run on a dotnet angular application causes "node --version failed with exit code 1" error

nguerrera:
@LanPodder, please open a separate issue and include full steps to reproduce the problem. Which env variables, what are the contents of project file, etc. Also, the issue should be filed at dotnet/core or dotnet/sdk.

“Node” here refers to msbuild nodes, not node.js, so I don’t think it is related to what you’re seeing.

Issue #1558, opened 2017-01-13
MSBuild is logging a bad toolset version if you don't have one:


davkean:
If you build a project without a Toolset, you get the following which seems to indicate something wrong:

```
1>Project file contains ToolsVersion="". This toolset may be unknown or missing, in which case you may be able to resolve this by installing the appropriate version of MSBuild, or the build may have been forced to a particular ToolsVersion for policy reasons. Treating the project as if it had ToolsVersion="15.0". For more information, please see http://go.microsoft.com/fwlink/?LinkId=293424.

```

Can we remove this message?

PR #3221, opened 2018-04-24
No toolsversion message when one is not mentioned in the project

cdmihai:
A log message is generated when the toolsversion specified in the project differs from the one msbuild chooses. Now that toolsversion is not required anymore, suppress this message when there is no explicit toolsversion reference in the project file.

image

### DIFF
--- a/src/Build/Instance/ProjectInstance.cs
+++ b/src/Build/Instance/ProjectInstance.cs
@@ -2082,7 +2082,12 @@ namespace Microsoft.Build.Execution
                         toolsVersion = visualStudioVersion.ToString(CultureInfo.InvariantCulture) + ".0";
                     }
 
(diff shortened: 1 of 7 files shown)


src/Build/Instance/ProjectInstance.cs
```diff
@@ -2082,7 +2082,12 @@ internal static ProjectInstance[] LoadSolutionForBuild(string projectFile, Prope
                         toolsVersion = visualStudioVersion.ToString(CultureInfo.InvariantCulture) + ".0";
                     }
 
-                    string toolsVersionToUse = Utilities.GenerateToolsVersionToUse(explicitToolsVersion: null, toolsVersionFromProject: toolsVersion, getToolset: buildParameters.GetToolset, defaultToolsVersion: Constants.defaultSolutionWrapperProjectToolsVersion);
+                    string toolsVersionToUse = Utilities.GenerateToolsVersionToUse(
+                        explicitToolsVersion: null,
```
cdmihai: `explicitToolsVersion` is a horrible name and threw me off initially. It does not refer to the toolsversion being explicit in the file, but to the toolsversion being specified via the APIs. I suggest renaming it everywhere to to `explicitToolsVersionViaApi`. Sounds good?

AndyGerlicher:
Mention closes #1558 in the commit please.

PR #3239, opened 2018-04-26
WIP: Add support for an SDK Resolver manifest file

### LINKED ISSUE #3176: Allow SDK Resolver discovery via config file
For assembly loading logic to be easier, allow for a config file to be dropped in the `SdkResolvers` folder rather than the full closure of DLLs (or run-time AssemblyResolve events).

Example:
`C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\SdkResolvers\Microsoft.Build.NuGetSdkResolver.config`
```XML
<SdkResolver Source="..\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\Microsoft.Build.NuGetSdkResolver.dll"
             Type="Microsoft.Build.NuGetSdkResolver" />
```

This is in addition to the existing pattern of loading `SdkResolvers\<DirectoryName>\<DirectoryName>.dll`

AndyGerlicher:
Allow for an xml file to be read instead of a resolver dll. This allows
for easier assembly loading and deployment for resolvers that have
complex dependencies (NuGetSdkResolver).

File format should be:
```XML
<SdkResolver>
  <Path>...</Path>
</SdkResolver>
```
Closes #3176
See: https://github.com/NuGet/Home/issues/6799


src/Build/BackEnd/Components/SdkResolution/SdkResolverManifest.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.IO;
+using System.Runtime.Serialization;
+using System.Xml;
+
+namespace Microsoft.Build.BackEnd.SdkResolution
+{
+    /// <summary>
+    /// Serialization contract for an SDK Resolver manifest
+    /// </summary>
+    [DataContract(Name = "SdkResolver", Namespace = "")]
```
rainersigwald: Why this kind of XML serialization rather than just reading the XML manually, as we do elsewhere?

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -22,12 +25,14 @@ namespace Microsoft.Build.Engine.UnitTests.BackEnd
 {
     public class SdkResolverLoader_Tests
     {
+        private readonly ITestOutputHelper _output;
         private readonly MockLogger _logger;
         private readonly LoggingContext _loggingContext;
 
-        public SdkResolverLoader_Tests()
+        public SdkResolverLoader_Tests(ITestOutputHelper output)
```
rainersigwald: ❤️

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -181,6 +183,54 @@ public void VerifyWarningLoggedWhenResolverAssemblyCannotBeLoaded()
             _logger.ErrorCount.ShouldBe(0);
         }
 
+        [Fact]
+        public void SdkReoslverLoaderReadsManifestFile()
```
rainersigwald: Typo: `resoslver`

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -181,6 +183,54 @@ public void VerifyWarningLoggedWhenResolverAssemblyCannotBeLoaded()
...
+                SdkResolverLoader loader = new SdkResolverLoader();
+                var resolversFound = loader.FindPotentialSdkResolvers(root);
+
+                resolversFound.Count.ShouldBe(1);
+                resolversFound.First().ShouldBe(assemblyToLoad);
+            }
+        }
+
+        [Fact]
+        public void SdkReoslverLoaderErrorsWithInvalidManifestFile()
```
rainersigwald: typo

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -181,6 +183,54 @@ public void VerifyWarningLoggedWhenResolverAssemblyCannotBeLoaded()
...
+
+                Directory.CreateDirectory(resolverPath);
+                File.WriteAllText(resolverManifest, $@"
+                    <SdkResolver2>
+                      <Path>{assemblyToLoad}</Path>
+                    </SdkResolver2>");
+
+                SdkResolverLoader loader = new SdkResolverLoader();
+
+                Should.Throw<SerializationException>(() => loader.FindPotentialSdkResolvers(root));
```
rainersigwald: How does this failure manifest to the user?

src/Build/BackEnd/Components/SdkResolution/SdkResolverManifest.cs
AndyGerlicher: This is used for the AssemblyFolder config stuff. No real reason other than simplicity. I'm not opposed to just reading the xml manually.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -47,15 +48,80 @@ internal class SdkResolverLoader
...
+                if (manifest == null || string.IsNullOrEmpty(manifest.Path))
+                {
+                    ProjectFileErrorUtilities.ThrowInvalidProjectFile(new BuildEventFileInfo(location), "SdkResolverDllInManifestMissing", pathToManifest, string.Empty);
+                }
+
+                path = manifest.Path;
+            }
+            catch (SerializationException e)
+            {
+                ProjectFileErrorUtilities.ThrowInvalidProjectFile(new BuildEventFileInfo(location), e, "SdkResolverManifestInvalid", pathToManifest, e.Message);
```
rainersigwald: Isn't it usually better to pass `e.ToString()` rather than `Message`? IIRC that tends to have more information.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -47,15 +48,80 @@ internal class SdkResolverLoader
...
+            }
+
+            if (!Path.IsPathRooted(path))
+            {
+                path = Path.Combine(manifestFolder, path);
```
rainersigwald: No need to check if it's rooted first; if the rightmost argument to `Path.Combine()` is rooted it wins.

rainersigwald: Or is that just to save a `GetFullPath()` call?

src/Build/Microsoft.Build.csproj
```diff
@@ -234,6 +233,7 @@
     <Compile Include="BackEnd\Components\SdkResolution\ISdkResolverService.cs" />
     <Compile Include="BackEnd\Components\SdkResolution\OutOfProcNodeSdkResolverService.cs" />
     <Compile Include="BackEnd\Components\SdkResolution\OutOfProcNodeSdkResolverServiceFactory.cs" />
+    <Compile Include="BackEnd\Components\SdkResolution\SdkResolverManifest.cs" />
```
rainersigwald: extreme nit: move this down so it's alphabetical? But alphabetical order is a mess in this file anyway so meh.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -47,15 +48,80 @@ internal class SdkResolverLoader
...
             if (string.IsNullOrEmpty(rootFolder) || !FileUtilities.DirectoryExistsNoThrow(rootFolder))
-                return new List<string>();
+            {
+                return assembliesList;
+            }
+
+            foreach (var subfolder in new DirectoryInfo(rootFolder).GetDirectories())
+            {
+                var assembly = Path.Combine(subfolder.FullName, $"{subfolder.Name}.dll");
+                var manifest = Path.Combine(subfolder.FullName, $"{subfolder.Name}.xml");
```
cdmihai: What if both a dll and a manifest exist?

cdmihai: What if two manifests point to the same dll? Would that lead to loading errors?

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -47,15 +48,80 @@ internal class SdkResolverLoader
...
+                }
+
+                path = manifest.Path;
+            }
+            catch (SerializationException e)
+            {
+                ProjectFileErrorUtilities.ThrowInvalidProjectFile(new BuildEventFileInfo(location), e, "SdkResolverManifestInvalid", pathToManifest, e.Message);
+            }
+
+            if (!Path.IsPathRooted(path))
```
cdmihai: What are the paths relative to? The current working directory? Does that mean relative paths would differ for VS, CLI, vscode, resharper, etc?

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
AndyGerlicher: This will lose stacktraces, but they're not very useful. The message in this case is really specific (invalid character in the xml on line X column Y, etc.), The stack trace would just be in the deserialize method.

rainersigwald: Good enough for me. Maybe comment that here?

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
AndyGerlicher: That is why I did that. GetFullPath was nice for when it had `..\..\..\..\..\..\` in it. :) but not necessary otherwise.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
AndyGerlicher: Relative to the resolver folder which is always MSBuildBin\SdkResolver\SomeResolverName\.

The config will likely have to be different for different deployments (VS and CLI).

PR #3240, opened 2018-04-26
[mono][test] Fix Exec timeout test

radical:
It is working around a difference in behavior in .netcore/HighSierra,
but isn't required for Mono.


akoeplinger:
@rainersigwald did you file a corefx bug for the change? the change is indeed unexpected.

PR #3250, opened 2018-04-30
Update NuGet to 4.8.0-preview1.5158

AndyGerlicher:


AndyGerlicher:
@rainersigwald Assembly version changed for 15.8. We need to insert or the versions will driff (until we get the resolver moved).

PR #3252, opened 2018-05-01
Preserve ProjectStarted GlobalProperties in binary log

rainersigwald:
The lack of this meant that some operations done by a logger (like figuring out why a project was built multiple times) weren't possible with only the binary log, but worked fine directly attached to a real build.


src/Build.UnitTests/BuildEventArgsSerialization_Tests.cs
```diff
@@ -66,7 +66,7 @@ public void RoundtripProjectStartedEventArgs()
                 properties: new List<DictionaryEntry>() { new DictionaryEntry("Key", "Value") },
                 items: new List<DictionaryEntry>() { new DictionaryEntry("Key", new MyTaskItem() { ItemSpec = "TestItemSpec" }) },
                 parentBuildEventContext: new BuildEventContext(7, 8, 9, 10, 11, 12),
-                globalProperties: null, // we do not serialize GlobalProperties
```
rainersigwald: @KirillOsenkov Do you recall if this was deliberate for some reason? I couldn't think of one and assumed this was a bug but since it's explicitly mentioned here I thought I'd ask.

KirillOsenkov: I'm guessing either to avoid noise or to avoid data bloat? I'm not sure myself.

PR #3253, opened 2018-05-01
Diagnostic/evaluation logging

panopticoncentral:
This PR proposes to address several logging issues:

* There are several situations in Visual Studio where loggers need to specify precisely the events that they want to receive so that efficient forwarding loggers can be created. Without more fine-grained event lists, forwarding logger creators have to rely on verbosity, which may include many events that the loggers don't need but will get forwarded.

* Evaluation profiling is now governed by the project instance creator, but this means that if a logger wishes to receive evaluation profiling it has to have a way to talk to the project instance creator. This PR proposes an alternate method whereby a logger can specify that it wants evaluation profiling without having to state that up front in the build parameters.

In adding a test for the evaluation profiling, I took the liberty of doing a little syntax cleanup in the MockLogger. If that's not OK, or if it looks wrong, I can back it out, no problem. Also, totally open to any suggestions for a better way to do this.


cdmihai:
@dotnet-bot test 
Windows_NT Build for Full please

src/Framework/IDiagnosticLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        /// <summary>
+        /// Log task inputs
+        /// </summary>
+        TaskInputs = 0x100000,
+
+        /// <summary>
+        /// Log evaluation profiles
+        /// </summary>
+        EvaluationProfile = 0x200000
```
cdmihai: total nit: I prefer `1 << 2`, `1 << 3` `1 << 4`, etc for what I think is better readability, but it's completely a personal preference, feel free to ignore :)

src/Build/Evaluation/Evaluator.cs
```diff
@@ -250,7 +252,8 @@ internal class Evaluator<P, I, M, D>
             _sdkResolverService = sdkResolverService;
             _submissionId = submissionId;
             _evaluationContext = evaluationContext;
-            _evaluationProfiler = new EvaluationProfiler((loadSettings & ProjectLoadSettings.ProfileEvaluation) != 0);
+            _profileEvaluation = profileEvaluation;
+            _evaluationProfiler = new EvaluationProfiler(_profileEvaluation);
```
cdmihai: Since the Evaluator.EvaluationProfiler field is always instantiated and it already knows whether it's on or off, the Evaluator._profileEvaluation field does not seem like it's needed. You could query the EvaluationProfiler instead.

cdmihai:
Another alternative would be to turn profiling on / off via a new Project API. I guess it depends on what you have more easily access to, logger objects, or project objects.

cdmihai:
As it is now, it is possible for API users to create new IDiagnosticLoggers, but then the default forwarding loggers will ignore them, which would be unexpected. This means whenever users create IDiagnosticLoggers, they need to also pass in a forwarding logger which filters. Is too cumbersome? If it is too cumbersome, then some solutions might be:
- ship a new DiagnosticForwardingLogger which knows how to filter based on the DiagnosticInformation enum. Users would have to use one of these whenever they use a diagnostic logger. Or maybe we could add one automatically?
- Replace the current CentralForwardingLogger (forwards everything, used when no distributed loggers are specified) with a DiagnosticForwardingLogger in the cases where all central loggers are IDiagnostic ones (using the flag union)

src/Framework/IDiagnosticLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// <summary>
+    /// A logger that specifies at a specific level what information it wants to be logged. This is
+    /// useful to specify to a forwarding logger exactly what events are to be forwarded, and also
+    /// allows turning on task input logging and evaluation profiling.
+    /// </summary>
+    public interface IDiagnosticLogger : ILogger
```
cdmihai: I don't like the name Diagnostic logger. The intent is to filter events based on various traits (class type, importance, event classes (profiling, telemetry, evaluation, etc)), via a whitelist enum. Some options:
- FilteringLogger
- ConfigurableLogger. Clashes with the existing ConfigurableForwardingLogger

src/Framework/IDiagnosticLogger.cs
```diff
@@ -0,0 +1,139 @@
...
+
+using System;
+
+namespace Microsoft.Build.Framework
+{
+    /// <summary>
+    /// Diagnostic information that a build logger cares about.
+    /// </summary>
+    [Flags]
+    public enum DiagnosticInformation
```
cdmihai: How comprehensive is this supposed to be?

What about telemetry events? [https://github.com/Microsoft/msbuild/blob/master/src/Framework/TelemetryEventArgs.cs](https://github.com/Microsoft/msbuild/blob/master/src/Framework/TelemetryEventArgs.cs)

cdmihai:
We've talked a bit about this:

Requirements
- MSBuild does need a new type of logger that specifies a whitelist of capabilities it is interested in. These capabilities need to be queriable, and then the engine would need to activate only the logging capabilities which are specified by the loggers.
- these capabilities could be of two major kinds: filters (what subsets of events the logger is interested in) and features (log task inputs, log target outputs)
- some capabilities make sense for specific loggers but not all. For example, performance summary is specific to the console and file loggers, but not ILogger in general. So it does not make sense to have everything in one big bucket?
- there needs to be some form of backwards compatibility, where old loggers can swim with new loggers.

Implementation considerations
- As a way to go forward, it's probably best to implement these requirements in incremental steps instead of all at once. But then we need to make sure the initial types are compatible with the long term needs
  - rather than specify all capabilities up front now in this PR, only include the ones that are actually listened to
- Is a flag enum the best way to represent a set of capabilities? Since there are many classes and subclasses of capabilities, and since some capabilities are specific only to some logger implementations, having a huge single enum for everything might become unmanageable.
- Backward compatibility
  - Ideally, it would be nice if all loggers become this new type of configurable, queriable, capability based loggers, and the engine does a set union of capabilities. I doubt we could use reflection to look at which events are subscribed to, because the loggers could have custom code to further filter events.
  - one option would be to translate old tasks to a set of capabilities, which probably means enabling all event types
  - another option would be to turn everything on if old style loggers are detected, and maybe issue a diagnostic level warning saying something better exists?

src/Framework/IEventSource3.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Build.Framework
+{
+    /// <summary>
+    /// This interface defines the events raised by the build engine.
+    /// Loggers use this interface to subscribe to the events they
+    /// are interested in receiving.
+    /// </summary>
+    public interface IEventSource3 : IEventSource
```
rainersigwald: Should `3` inherit from `IEventSource2`?

src/Build.UnitTests/Evaluation/EvaluationLogging_Tests.cs
```diff
@@ -173,5 +174,19 @@ public void ProjectShouldHaveValidEvaluationIdDuringEvaluation()
                     }
                 });
         }
+
+        [Fact]
+        public void TurnOnProfileEvaluationFromLogger()
```
cdmihai: Can you also add a test where logging task inputs is turned on via a logger?

cdmihai:
I don't think this works with out of proc nodes, because there's no mechanism to transmit the IEventSource3 settings to the out of proc nodes' LoggingService.

Out of proc nodes [get the forwarding loggers](https://github.com/Microsoft/msbuild/blob/master/src/Build/BackEnd/Node/OutOfProcNode.cs#L773) via their LoggerDescriptions, [as sent from the BuildManager](https://github.com/Microsoft/msbuild/blob/master/src/Build/BackEnd/BuildManager/BuildManager.cs#L1645-L1665) when a node is created.

One way to do it is to put the serializable state of LoggingService (profile evaluation and log task inputs for now) in a separate data class which implements INodePacketTranslator and make it a field of NodeConfiguration. That way the OutOfProcNode can use it to instantiate its logging service.

panopticoncentral:
@cdmihai You're correct, this doesn't work for out of proc. It seems to me that the simplest way to make it work is to mirror what happens for everything else -- the ConfigurableForwardingLogger takes a property that controls whether or not it sets the properties, just like it takes properties that control whether or not events are forwarded.

panopticoncentral:
@cdmihai Let me know what you think about the changes to the forwarding logger. Also, I'm happy to add a test for the task inputs, is there a good model test that I can use as a base?

cdmihai:
Regarding the log task inputs test, as far as I can tell, we have no test for it :(. You can use the [BuildManager tests](https://github.com/Microsoft/msbuild/blob/master/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs#L3517-L3522) as a test example to force out of proc nodes, and then use the mock logger to assert task inputs were logged. This would be useful to test for event profiling as well. As for the test class, you can create a new one.

Regarding out of proc logging, I don't think that persisting the settings just in the ConfigurableForwardingLogger is sufficient, because not all logging scenarios end up using the configurable logger:
1. Degenerate case: when no loggers are subscribed, MSBuild still creates a ConfigurableForwardingLogger to get ProjectStarted / Finished events from the out of proc nodes, which are needed by the scheduler in the central node.
2. When all loggers are central loggers (no IForwardingLoggers), then there is no ConfigurableForwardingLogger used. Rather, in this case MSBuild creates a CentralForwardingLogger which forwards all events back. For IEventSource3, maybe it would be better to just delete the central logger and replace it with a configurable one which has everything turned on.
3. When all loggers are distributed loggers (pairs of central and forwarding loggers), then MSBuild uses the user forwarding loggers (which can be of any user defined type) on top of which it adds a minimal configurable logger as in case 1. (because it can't be sure the forwarding loggers would forward project started / finished). This means that if there's distributed loggers setting state in their IEventSource3, there's no mechanism in MSBuild to serialize that state. This means that every user with custom forwarding loggers needs to duplicate the changes you did in ConfigurableForwardingLogger. I'd rather we make a mechanism where MSBuild handles this.
4. When there's both central loggers and distributed loggers, then MSBuild mixes 2. and 3. which is the most inefficient thing one can do, as it leads to rampant event multiplication over the pipes.

I think there's two main ways to go:
1. Loggers are responsible to serialize IEventSource3 state: In both the central and out of proc nodes, the logging service aggregates it's state from the loggers. Not sure how to achieve this with user custom loggers, as they only have an interface to implement, and sadly interfaces don't have default implementations. Maybe the BuildManager can create yet another configurable logger to capture and aggregate the state from all user distributed loggers? But this leads to event multiplication and worse perf.
2. LoggingService is responsible for serializing IEventSource3 state. Since the LoggingService knows the final state after querying all loggers, it can transfer that state over the pipe to the out of proc nodes' LoggingServices. The downside here is that the out proc nodes wouldn't be able to tell which logger asked for something, they'd only get the aggregated result. But I don't think we'll ever need that level of resolution.

panopticoncentral:
@cdmihai I updated the PR based on our conversation, can you take a look? Also, should this PR be targeted against master or vs15.8?

src/Build/BackEnd/Components/Logging/LoggingService.cs
```diff
@@ -404,96 +404,81 @@ public int NextTaskId
...
+            get;
+            set;
+        } = null;
+
+        /// <summary>
+        /// Should evaluation events include profiling information?
+        /// </summary>
+        public bool IncludeEvaluationProfile
+        {
+            get => _includeEvaluationProfile || _eventSinkDictionary.Values.OfType<EventSourceSink>().Any(sink => sink.IncludeEvaluationProfiles);
```
cdmihai: These linqs will execute on each evaluation start, which might get picked up in small project restore perf tests (.net core perf is pretty critical about this scenario). Could you hide them behind a Lazy?

panopticoncentral:
@dotnet-bot test OSX10.13 Build for Mono please

panopticoncentral:

rainersigwald:
@panopticoncentral Mono builds aren't passing at the moment and won't block your merge.

panopticoncentral:
@rainersigwald What about the CoreCLR for Ubuntu14.04? It says it's triggered but nothing happening...?

cdmihai:

PR #3256, opened 2018-05-02
DebuggerDisplay parsed conditional expressions

rainersigwald:
When trying to explain how the parser worked for dotnet/project-system#3510 I was annoyed at what it took to step through and understand the parse.

With this change, the debugger now shows pretty values.


rainersigwald:
Example:

image

davkean:
Nice!

PR #3259, opened 2018-05-02
Remove package dependency on Microsoft.NETCore.App (vs15.7)

AndyGerlicher:
Prepping this in case we need it.

This dependency should not leak into our NuGet packages. This was added as the result of a bug (see https://github.com/dotnet/sdk/issues/2204) and causes issues like the ones listed in #3228.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -63,7 +63,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppVersion)" Condition="'$(TargetFramework)' == 'netcoreapp2.1'" />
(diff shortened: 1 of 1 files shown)


src/Directory.Build.props
```diff
@@ -63,7 +63,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppVersion)" Condition="'$(TargetFramework)' == 'netcoreapp2.1'" />
+    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppVersion)" Condition="'$(TargetFramework)' == 'netcoreapp2.1'" PrivateAssets="All"/>
```
cdmihai: Should it be conditioned on `DisableImplicitFrameworkReferences` which is the flag that also tells the SDK to not add an implicit one? https://github.com/AndyGerlicher/msbuild/commit/e735b93886f1ac64b82eec2684cedcdd6892ef5f#diff-9002cc0a94d7643cb073436b973a63ebR143

rainersigwald: No, I don't think so: this should only be added when that's set.

PR #3267, opened 2018-05-04
Expose sdk resolution in Project

### LINKED ISSUE #3203: Add Project.Sdks property to retrieve SDKs needed to evaluate a project
VS would like to display a list of SDKs (with version) used in a given project.

The `Project` class has properties for `Imports` and other things but hasn't been augmented to report Sdks. It should do so, with a signature like

```c#
public IList<SdkReference> Sdks
```

An SDK can import another SDK, which could suggest a hierarchical model for this. But the complexity of that is probably too high for the initial implementation of this feature: since Microsoft.NET.Sdk imports common.targets imports Directory.Build.props, any SDK-enabled import in a D.B.props/targets would appear to be referenced from Microsoft.NET.Sdk, which isn't accurate. Any hierarchy would have to account for that.

cdmihai:
- `ProjectImportElement.OriginalElement` points to the original `ProjectElement` which generated implicit imports (`ProjectRootElement` or `ProjectSdkElement`).
- `ResolvedImport.SdkResult` points to sdk resolution results.

`Project.Imports.Where(i -> i.SdkResult != null)` retrieves all `ResolvedImports` that came from sdks, with pointers to the `SdkResult` and the import element. This should allow VS to display the sdks a Project imports, and to navigate to the real elements which generated those imports.

Closes #3203

### DIFF
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -451,7 +451,9 @@ namespace Microsoft.Build.Framework
     public abstract partial class SdkResult
     {
         protected SdkResult() { }
-        public bool Success { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]protected set { } }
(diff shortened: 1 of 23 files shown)


cdmihai:
@pilchie

cdmihai:
A potential perf caveat is that [Project.Imports](https://github.com/Microsoft/msbuild/blob/master/src/Build/Definition/Project.cs#L862) clones the underlying list on each access. Would this get called sufficiently many times from VS to be a perf issue? If yes, then I'll look into reducing it.

Pilchie:
I don't think the newly added code will call this very often, so I'm not worried about it.

PR #3272, opened 2018-05-04
Initialize repotoolset so source-build can supply a built version.

dseefeld:
For source-build 2.1 RC this was a patch for msbuild to enable source-build to supply the version of repotoolset that was source-built.  This moves the changes into msbuild so the patch can be removed.

cc @cdmihai

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -278,19 +278,36 @@ function InstallDotNetCli {
 }
 
 function InstallRepoToolset {
+  CreateDirectory "$RepoToolsetDir"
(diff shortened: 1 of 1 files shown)


cdmihai:
@rainersigwald wrote at https://github.com/Microsoft/msbuild/pull/3287#discussion_r187116715
_So the logic is "if repotools appears to have already been restored (has a path written to a text file) then just return that path, otherwise write a project and have it write the path, "restoring" the SDK automatically during that"?_

Appears so. This will make it different from how the windows build gets repotoolset, but since the windows version is simpler, I'd keep them separate for now.

PR #3277, opened 2018-05-07
Add support for fallback search paths for Target frameworks

radical:
When running on Mono, target frameworks are searched for in
$mono_prefix/lib/mono/xbuild-frameworks, which becomes
/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild-frameworks.

a new /Library/Frameworks/Mono.framework/Versions/@VERSION@ directory
"lost" (For example, installed by Xamarin.Android). The file is still there, but msbuild
will no longer find it.

To address this upgrade scenario, we can add a fallback path:

`/Library/Frameworks/Mono.framework/External/xbuild-frameworks`

Implementation:

We add a new `$(TargetFrameworkFallbackSearchPaths)` property which
can have the list of fallback search paths and gets used by the `GetReferenceAssemblyPaths`.

For Mono/OSX case, we point that to /Library/Frameworks/Mono.framework/External/xbuild-frameworks

So, effectively, target frameworks are looked up, in order:

  1. $(TargetFrameworkRootPath) or the default location, if
     $(TargetFrameworkRootPath) is ''

  2. If not found in (1), then try to find it in the fallback search
     paths specified in $(TargetFrameworkFallbackSearchPaths)

Fallback search paths also allow overriding the fallback paths to test local in-tree builds of
custom target frameworks. Simply changing `$(TargetFrameworkRootPath)` would
mean that the in-tree builds of the custom frameworks would work but the
default (.net) ones would not be found!

### DIFF
--- a/src/Tasks/GetReferenceAssemblyPaths.cs
+++ b/src/Tasks/GetReferenceAssemblyPaths.cs
@@ -121,6 +121,18 @@ namespace Microsoft.Build.Tasks
         [Output]
         public string TargetFrameworkMonikerDisplayName { get; set; }
 
+        /// <summary>
+        /// Target frameworks are looked up in @RootPath. If it cannot be found
+        /// there, then paths in @TargetFrameworkFallbackSearchPaths
+        /// are used for the lookup, in order. This can have multiple paths, separated
+        /// by ';'
+        /// </summary>
+        public string TargetFrameworkFallbackSearchPaths
+        {
+            get;
+            set;
+        }
+
         #endregion
 
         #region ITask Members
@@ -181,7 +193,7 @@ namespace Microsoft.Build.Tasks
 
             try
             {
-                _tfmPaths = GetPaths(RootPath, moniker);
+                _tfmPaths = GetPaths(RootPath, TargetFrameworkFallbackSearchPaths, moniker);
 
                 if (_tfmPaths != null && _tfmPaths.Count > 0)
                 {
@@ -192,7 +204,7 @@ namespace Microsoft.Build.Tasks
                 // There is no point in generating the full framework paths if profile path could not be found.
                 if (targetingProfile && _tfmPaths != null)
                 {
-                    _tfmPathsNoProfile = GetPaths(RootPath, monikerWithNoProfile);
+                    _tfmPathsNoProfile = GetPaths(RootPath, TargetFrameworkFallbackSearchPaths, monikerWithNoProfile);
                 }
 
                 // The path with out the profile is just the reference assembly paths.
@@ -222,11 +234,14 @@ namespace Microsoft.Build.Tasks
         /// <summary>
         /// Generate the set of chained reference assembly paths
         /// </summary>
-        private IList<String> GetPaths(string rootPath, FrameworkNameVersioning frameworkmoniker)
+        private IList<String> GetPaths(string rootPath, string targetFrameworkFallbackSearchPaths, FrameworkNameVersioning frameworkmoniker)
         {
-            IList<string> pathsToReturn = String.IsNullOrEmpty(rootPath) ?
-                ToolLocationHelper.GetPathToReferenceAssemblies(frameworkmoniker) :
-                ToolLocationHelper.GetPathToReferenceAssemblies(rootPath, frameworkmoniker);
+            IList<String> pathsToReturn = ToolLocationHelper.GetPathToReferenceAssemblies(
+                                                frameworkmoniker.Identifier,
+                                                frameworkmoniker.Version.ToString(),
+                                                frameworkmoniker.Profile,
+                                                rootPath,
+                                                targetFrameworkFallbackSearchPaths);
 
             if (!SuppressNotFoundError)
             {
--- a/src/Utilities.UnitTests/ToolLocationHelper_Tests.cs
+++ b/src/Utilities.UnitTests/ToolLocationHelper_Tests.cs
@@ -2613,30 +2613,177 @@ namespace Microsoft.Build.UnitTests
         }
 
         [Fact]
-        public void GetPathToReferenceAssembliesWithCustomTargetFrameworkRoot()
+        public void GetPathToStandardLibrariesWithCustomTargetFrameworkInFallbackSearchPathAndNullRoot()
         {
             using (var env = TestEnvironment.Create())
             {
                 string frameworkName = "Foo Framework";
                 string frameworkVersion = "v0.1";
-                string rootDir = Path.Combine(env.DefaultTestDirectory.FolderPath, "framework-root");
+                string customFrameworkRootPath = Path.Combine(env.DefaultTestDirectory.FolderPath, "framework-root");
 
-                string asmPath = CreateNewFrameworkAndGetAssembliesPath(env, frameworkName, frameworkVersion, rootDir);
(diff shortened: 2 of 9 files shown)


radical:
This effectively revives the older PR #1389. 

Andy had commented https://github.com/Microsoft/msbuild/pull/1389#issuecomment-270253519 .

To answer his main point:

`1. I prefer not to have fallback. It seems un-natural here. You're passing in a string that says look exactly here, but if that fails go ahead and look in all these other places too. It seems cleaner to just drop the fallback and pass in a list of places to look. Ideally typed as a list and if you pass in a string (legacy purposes) it would just pass that along as new List<string> { singlePath } to the new method that does the search that only takes a list.`

- The `targetFrameworkRootPath` argument is deliberately kept separate from the fallback paths. The reason (from
the commit message):
```
    - The reason for using a `targetFrameworkRootPath` parameter separate from
    the `targetFrameworkFallbackSearchPaths` is that we need to
    automatically fallback to a default path, which is being determined in
    code (`FrameworkLocationHelper.programFilesReferenceAssemblyLocation`).

    - Also, the parameter `targetFrameworkRootPath` roughly corresponds to the
    `$(TargetFrameworkRootPath)` property and changing that to have multiple
    paths would change it's meaning.
```

Also, this behavior would be consistent with what we have for `MSBuildExtensionsPath`,
which is to have the lookup order as (from PR description)

```
... target frameworks are looked up, in order:

1.  $(TargetFrameworkRootPath) or the default location, if
     $(TargetFrameworkRootPath) is ''

2.  If not found in (1), then try to find it in the fallback search
     paths specified in $(TargetFrameworkFallbackSearchPaths)
```

mono/ExtensionsPath-ToolsVersion/Microsoft.Common.targets/ImportBefore/Microsoft.Common.Mono.Before.targets
```diff
@@ -0,0 +1,24 @@
+﻿<!--
+***********************************************************************************************
```
AndyGerlicher: This needs to be moved somewhere else not off the root. Maybe in `src/Tasks/Mono/Microsoft.Common.Mono.Before.targets`?

src/Utilities/ToolLocationHelper.cs
```diff
@@ -1827,6 +1833,22 @@ public static string GetPathToStandardLibraries(string targetFrameworkIdentifier
...
+        /// <param name="targetFrameworkVersion">Version being targeted</param>
+        /// <param name="targetFrameworkProfile">Profile being targeted</param>
+        /// <param name="platformTarget">What is the targeted platform, this is used to determine where we should look for the standard libraries. Note, this parameter is only used for .net frameworks less than 4.0</param>
+        /// <param name="targetFrameworkRootPath">Root directory where the target framework will be looked for. Uses default path if this is null</param>
+        /// <param name="targetFrameworkFallbackSearchPaths">';' separated list of paths that are looked up if the the framework cannot be found in @targetFrameworkRootPath</param>
+        /// <exception cref="ArgumentNullException">When the frameworkName is null</exception>
+        /// <returns>Collection of reference assembly locations.</returns>
+        public static string GetPathToStandardLibraries(string targetFrameworkIdentifier, string targetFrameworkVersion, string targetFrameworkProfile, string platformTarget, string targetFrameworkRootPath, string targetFrameworkFallbackSearchPaths = null)
         {
```
AndyGerlicher: Remove the default value, it's hidden by the overload w/o this parameter.

src/Utilities/ToolLocationHelper.cs
```diff
@@ -2184,6 +2226,38 @@ private static IList<string> HandleLegacy20(VersionToPath PathToDotNetFramework)
...
+        /// <param name="frameworkName">A frameworkName class which represents a TargetFrameworkMoniker. This cannot be null.</param>
+        /// <returns>Collection of reference assembly locations.</returns>
+        public static IList<String> GetPathToReferenceAssemblies(string targetFrameworkRootPath, string targetFrameworkFallbackSearchPaths, FrameworkNameVersioning frameworkName)
+        {
+            IList<string> pathsList = String.IsNullOrEmpty(targetFrameworkRootPath)
+                                        ? GetPathToReferenceAssemblies(frameworkName)
+                                        : GetPathToReferenceAssemblies(targetFrameworkRootPath, frameworkName);
+
+            if (pathsList?.Count > 0)
+                return pathsList;
```
AndyGerlicher: Surround in `{  }` block please. Same for the copy/paste a few lines down as well.

src/Tasks/GetReferenceAssemblyPaths.cs
```diff
@@ -278,18 +290,15 @@ public override bool Execute()
         /// <summary>
         /// Generate the set of chained reference assembly paths
         /// </summary>
-        private IList<String> GetPaths(string rootPath, FrameworkNameVersioning frameworkmoniker)
+        /// FIXME: do we really need the new arg? or should we just use the property?
```
AndyGerlicher: Don't think it matters, this method can't be made static as is anyway.

mono/ExtensionsPath-ToolsVersion/Microsoft.Common.targets/ImportBefore/Microsoft.Common.Mono.Before.targets
radical: I wanted to keep the mono-specific files[1] under `mono/`, like https://github.com/mono/msbuild/tree/xplat-master/mono . I will move the other mono-specific targets to this kinda folder and use `ImportBefore` (and similar) hooks to import then, instead of hardcoding the import in the main targets files.

1. mono-specific files like any additional targets, tasks, build scripts etc, to keep them cleanly away from rest of the msbuild code

src/Utilities.UnitTests/ToolLocationHelper_Tests.cs
```diff
@@ -2614,33 +2613,178 @@ public void GetPathToStandardLibrariesWithNullTargetFrameworkRootPath()
...
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                string customFrameworkDir = Path.Combine(env.CreateFolder().FolderPath, "framework-root");
+
+                CheckGetPathToReferenceAssemblies(env, customFrameworkDir, null,
+                                                    (string frameworkName, string frameworkVersion, string frameworkProfile, string _customFrameworkDir, string _fallbackSearchPaths)
+                                                         => ToolLocationHelper.GetPathToReferenceAssemblies(frameworkName, "v" + frameworkVersion, frameworkProfile, _customFrameworkDir));
+            }
+        }
```
AndyGerlicher: Please don't prefix _ in the variable name for lambda inputs.

radical:
Thanks for the review! Updated to address review comments.

radical:
> Only thing left is that file. Per the conversation today can we remove it altogether?

Done!

AndyGerlicher:
@radical Looks like a small conflict that needs to be fixed :(

radical:
Updated.

AndyGerlicher:

radical:
Will this go into `vs15.8` ?

radical:
nvm, https://github.com/Microsoft/msbuild/pull/3424 answers my question :)

PR #3278, opened 2018-05-08
Ignore missing imports during restore

jeffkl:
Some shipped SDKs have imports that are property based but are missing conditions.  This makes it impossible to retrofit them with a NuGet package since NuGet restore won't succeed because an import is missing.  This change makes it so that during Restore, projects are loaded with the ProjectLoadSettings that ignore missing, empty, and invalid imports.  This is what Visual Studio uses so that it can load projects that haven't been restored yet.  The build would still fail if there are missing imports.

Treat `/Target:Restore` the same as `/Restore` without a build

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
@@ -1155,6 +1155,14 @@ namespace Microsoft.Build.BackEnd
             // Get the hosted ISdkResolverService.  This returns either the MainNodeSdkResolverService or the OutOfProcNodeSdkResolverService depending on who created the current RequestBuilder
             ISdkResolverService sdkResolverService = _componentHost.GetComponent(BuildComponentType.SdkResolverService) as ISdkResolverService;
 
+            // Use different project load settings if the build request indicates to do so
+            ProjectLoadSettings projectLoadSettings = _componentHost.BuildParameters.ProjectLoadSettings;
+
+            if (_requestEntry.Request.BuildRequestDataFlags.HasFlag(BuildRequestDataFlags.IgnoreMissingEmptyAndInvalidImports))
+            {
+                projectLoadSettings |= ProjectLoadSettings.IgnoreMissingImports | ProjectLoadSettings.IgnoreInvalidImports | ProjectLoadSettings.IgnoreEmptyImports;
+            }
+
             return new ProjectInstance(
                 _requestEntry.RequestConfiguration.ProjectFullPath,
                 globalProperties,
@@ -1169,8 +1177,9 @@ namespace Microsoft.Build.BackEnd
                     BuildEventContext.InvalidProjectContextId,
                     BuildEventContext.InvalidTargetId,
                     BuildEventContext.InvalidTaskId),
-                    sdkResolverService,
-                    _requestEntry.Request.SubmissionId);
+                sdkResolverService,
+                _requestEntry.Request.SubmissionId,
+                projectLoadSettings);
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


jeffkl:
For example, an SDK might have a template like this:

``` xml

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>
  
  <Import Project="$(SomeSdkDir)\Settings.props" />

</Project>
```

And `SomeSdkDir` is set by installing something on your machine.  We can't retrofit this with a NuGet package because restore fails with an `InvalidProjectFileException`.

jeffkl:

src/MSBuild/XMake.cs
```diff
@@ -1117,7 +1117,10 @@ bool enableProfiler
                     {
                         try
                         {
-                            if (enableRestore)
+                            // Determine if the user specified /Target:Restore which means we should only execute a restore in the fancy way that /restore is executed
+                            bool restoreOnly = String.Equals(request.TargetNames.SingleOrDefault(), MSBuildConstants.RestoreTargetName, StringComparison.OrdinalIgnoreCase);
```
nguerrera: This feels especially magical. Are there any other examples of /t:Foo behaving differently than /t:Bar?

jeffkl: This is only for the case when you do `MSBuild /T:Restore` on a CI machine or script where you want to restore.  If the functionality to ignore missing imports only works for `MSBuild.exe /restore`, I thought it would be a bad user experience.  We're already treating `Restore` in such a special way, it seemed like a good idea to do this.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1155,6 +1155,14 @@ private ProjectInstance LoadProjectFromFile()
             // Get the hosted ISdkResolverService.  This returns either the MainNodeSdkResolverService or the OutOfProcNodeSdkResolverService depending on who created the current RequestBuilder
             ISdkResolverService sdkResolverService = _componentHost.GetComponent(BuildComponentType.SdkResolverService) as ISdkResolverService;
 
+            // Use different project load settings if the build request indicates to do so
+            ProjectLoadSettings projectLoadSettings = _componentHost.BuildParameters.ProjectLoadSettings;
+
+            if (_requestEntry.Request.BuildRequestDataFlags.HasFlag(BuildRequestDataFlags.IgnoreMissingEmptyAndInvalidImports))
```
cdmihai: Since there is already a BuildParameters.ProjectLoadSettings, why bother with an additional BuildParameters.IgnoreMissingEmptyAndInvalidImports bool which then has to be translated back to ProjectLoadSettings? Instead, in Xmake.cs, you could compose the ProjectLoadSettings directly.

jeffkl: Because you only get one BuildParameters between `BeginBuild` and `EndBuild` and we're queuing two build requests (Restore, Build).

cdmihai: Mkay, so this is to allow finer granularity within a single session.
Regarding restore and build, I don't remember, why aren't they running in separate begin / end build sessions? Seems like we'd want to keep them as separate as possible.

AndyGerlicher:
I'm good with this as is, but would like to see some a test at some level. Not the easiest change to test, but see if you can add something?

jeffkl:
@AndyGerlicher I'll definitely add some tests before merging.  I'm going to demo it next Monday at the CLI sync.  If everyone signs off, I'll add the tests then.

PR #3279, opened 2018-05-08
Add a ReferenceCopyLocalOutputGroup

### LINKED ISSUE #3069: MSBuild should define an output group with copy local content
I've noticed in 4 separate occasions folks use BuiltProjectOutputGroupDependencies to mean "copy local dependencies" but this is incorrect.  This output group represents dependencies passed to the compiler (`ReferencePath`).  MSBuild doesn't copy these to the output, instead `ReferenceCopyLocalPaths` are copied (as determined by RAR).  I think MSBuild should create a different output group for this.

The 4 places I've noticed that got this wrong are:
1. ClickOnce design time /cc @tmeschter 
2. [Azure SDK](https://github.com/MicrosoftDocs/azure-cloud-services-files/blob/master/Azure%20Targets%20SDK%202.9/Microsoft.WindowsAzure.targets#L1793-L1819)
3. Web projects (fixed in vs 15.4)
4. Service Fabric (fixed https://github.com/Azure/service-fabric-issues/issues/54)

The one place I noticed that got it right was `CopyLocalFilesOutputGroup` in Microsoft.AppXPackage.targets, but they created their own outputgroup.

It'd be nice if common.targets could define a single hook for getting this information and we could have everyone use it.

tmeschter:
Fixes #3069.

There are a number of places where we need to get the set of items produced by a project's references and needed for packaging purposes or just running the code on the target platform. These may well be different from the reference passed to the compiler (e.g., a NuGet package may have a single `ref` assembly for compiling but multiple `lib` assemblies for different runtimes) and may not be code at all (e.g. PDBs or resource assemblies). As noted in #3069 a number of individual project types have implemented their own targets to collect this information, with varying levels of correctness.

This change adds a new `ReferenceCopyLocalPathsOutputGroup` to Microsoft.Common.CurrentVersion.targets to make this functionality centrally available to anyone who needs it. It also does some refactoring to make it a little easier to extend the other output groups.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -5817,9 +5817,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     This target performs population of the Built project output group dependencies.
     ============================================================
     -->
+  <PropertyGroup>
+    <BuiltProjectOutputGroupDependenciesDependsOn>
+      $(BuiltProjectOutputGroupDependenciesDependsOn);
+      $(CommonOutputGroupsDependsOn)
+    </BuiltProjectOutputGroupDependenciesDependsOn>
+  </PropertyGroup>
+
   <Target
       Name="BuiltProjectOutputGroupDependencies"
-      DependsOnTargets="$(CommonOutputGroupsDependsOn)"
+      DependsOnTargets="$(BuiltProjectOutputGroupDependenciesDependsOn)"
       Returns="@(BuiltProjectOutputGroupDependency)">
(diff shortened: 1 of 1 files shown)


nguerrera:
LGTM.

tmeschter:
@rainersigwald @cdmihai Do you know if bits with this change have made it into the VS repo?

rainersigwald:
It should be in lab/d15.8stg, I think. If not, https://devdiv.visualstudio.com/DevDiv/MSBuild/_git/VS/pullrequest/123194 should get it into rel/d15.8 if I'm reading commit history right.

PR #3284, opened 2018-05-09
Add threading hint that RequestHandlerPumpProc is long running

### LINKED ISSUE #3281: RequestHandlerPumpProc leads into thread pool exhaustion issue
The code starts a normal task and call into this method, which doesn't exit.  It will take one thread out of the thread pool, and when the overall number goes lower, it can lead into thread pool exhaustion issue.

The fix is easy, you need pass a flag to tell CLR that this is a long run task, so it starts it with a new thread.  (Or you can create your own thread to run this proc.)

jeffkl:
Fixes #3281


jeffkl:
@lifengl does this look right?

PR #3287, opened 2018-05-09
Update build.sh to uptake source-built repotoolset

cdmihai:


build/build.sh
```diff
@@ -306,19 +306,36 @@ function InstallDotNetCli {
...
 
 function InstallRepoToolset {
+  CreateDirectory "$RepoToolsetDir"
   ReadJson "$RepoRoot/global.json" "RoslynTools.RepoToolset"
   RepoToolsetVersion=$readjsonvalue
-  RepoToolsetDir="$NuGetPackageRoot/roslyntools.repotoolset/$RepoToolsetVersion/tools"
-  RepoToolsetBuildProj="$RepoToolsetDir/Build.proj"
+  RepoToolsetLocationFile="$RepoToolsetDir/$RepoToolsetVersion.txt"
+
+  if [[ -a "$RepoToolsetLocationFile" ]]; then
```
rainersigwald: for my edification,  is there a difference between `-a` and `-e`? I didn't know what this meant having only used `-e` but I'm a bash-scripting novice.

build/build.sh
```diff
@@ -306,19 +306,36 @@ function InstallDotNetCli {
...
+    if [[ -a "$path" ]]; then
+      RepoToolsetBuildProj=$path
+      return
+    fi
+  fi
+  if [[ "$restore" != true ]]; then
+    ExitIfError 2 "Toolset version $RepoToolsetVersion has not been restored."
+  fi
 
   echo "Using repo tools version: $readjsonvalue"
```
rainersigwald: Shouldn't this line be just after the `ReadJson` call so it's printed on early-return too?

build/build.sh
```diff
@@ -306,19 +306,36 @@ function InstallDotNetCli {
...
   ReadJson "$RepoRoot/global.json" "RoslynTools.RepoToolset"
   RepoToolsetVersion=$readjsonvalue
-  RepoToolsetDir="$NuGetPackageRoot/roslyntools.repotoolset/$RepoToolsetVersion/tools"
-  RepoToolsetBuildProj="$RepoToolsetDir/Build.proj"
+  RepoToolsetLocationFile="$RepoToolsetDir/$RepoToolsetVersion.txt"
+
+  if [[ -a "$RepoToolsetLocationFile" ]]; then
+    local path=`cat $RepoToolsetLocationFile`
+    if [[ -a "$path" ]]; then
+      RepoToolsetBuildProj=$path
```
rainersigwald: So the logic is "if repotools appears to have already been restored (has a path written to a text file) then just return that path, otherwise write a project and have it write the path, "restoring" the SDK automatically during that"?

cdmihai:
Didn't notice #3272. Closing this as duplicate

PR #3291, opened 2018-05-09
Add ZipDirectory and Unzip tasks

### LINKED ISSUE #1781: Create a task to zip / unzip files
Producing zip archives is a very common build step. Would be nice to have a built-in msbuild task that does this.

jeffkl:
`ZipDirectory` task zips up a whole directory.  It will fail if the destination file already exists.  Zipping individual files is a lot harder because you have to calculated a base path for all files and make zip entries relative to that.  The code in the CLR that does this is very efficient by reusing string buffers and I didn't want to have to replicate the code.  So for now there's only `ZipDirectory` instead of a more generic `Zip`.

`Unzip` unzips files from an archive to a directory.
* Defaults to skip unzipping files that are already up-to-date
* Logs every file that was unzipped
* Supports cancellation
* Supports overwriting read-only files
* Supports unzipping multiple files to the same directory

Fixes #1781

### DIFF
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -1160,7 +1160,7 @@ namespace Microsoft.Build.Tasks
         public Microsoft.Build.Framework.ITaskItem[] TypeLibFiles { get { throw null; } set { } }
         public override bool Execute() { throw null; }
     }
-    public partial class Unzip : Microsoft.Build.Tasks.TaskExtension, Microsoft.Build.Framework.ICancelableTask, Microsoft.Build.Framework.ITask
+    public sealed partial class Unzip : Microsoft.Build.Tasks.TaskExtension, Microsoft.Build.Framework.ICancelableTask, Microsoft.Build.Framework.ITask
     {
         public Unzip() { }
         [Microsoft.Build.Framework.OutputAttribute]
@@ -1294,6 +1294,15 @@ namespace Microsoft.Build.Tasks
         public Microsoft.Build.Framework.ITaskItem XslCompiledDllPath { get { throw null; } set { } }
         public string XslContent { get { throw null; } set { } }
         public Microsoft.Build.Framework.ITaskItem XslInputPath { get { throw null; } set { } }
+        public override bool Execute() { throw null; }
+    }
+    public sealed partial class ZipDirectory : Microsoft.Build.Tasks.TaskExtension
+    {
+        public ZipDirectory() { }
+        [Microsoft.Build.Framework.RequiredAttribute]
+        public Microsoft.Build.Framework.ITaskItem DestinationFile { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        [Microsoft.Build.Framework.RequiredAttribute]
+        public Microsoft.Build.Framework.ITaskItem SourceDirectory { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public override bool Execute() { throw null; }
     }
 }
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -1163,16 +1163,12 @@ namespace Microsoft.Build.Tasks
     public sealed partial class Unzip : Microsoft.Build.Tasks.TaskExtension, Microsoft.Build.Framework.ICancelableTask, Microsoft.Build.Framework.ITask
     {
         public Unzip() { }
-        [Microsoft.Build.Framework.OutputAttribute]
-        public Microsoft.Build.Framework.ITaskItem[] DestinationFiles { get { throw null; } }
         [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem DestinationFolder { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool OverwriteReadOnlyFiles { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool SkipUnchangedFiles { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem[] SourceFiles { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
-        [Microsoft.Build.Framework.OutputAttribute]
-        public Microsoft.Build.Framework.ITaskItem[] UnzippedFiles { get { throw null; } }
         public void Cancel() { }
         public override bool Execute() { throw null; }
     }
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -1165,6 +1165,7 @@ namespace Microsoft.Build.Tasks
         public Unzip() { }
         [Microsoft.Build.Framework.OutputAttribute]
         public Microsoft.Build.Framework.ITaskItem[] DestinationFiles { get { throw null; } }
+        [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem DestinationFolder { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool OverwriteReadOnlyFiles { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool SkipUnchangedFiles { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -1298,6 +1298,7 @@ namespace Microsoft.Build.Tasks
         public ZipDirectory() { }
         [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem DestinationFile { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool Overwrite { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem SourceDirectory { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public override bool Execute() { throw null; }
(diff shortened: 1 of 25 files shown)


src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,211 @@
...
+                    // for logging and cancellation, we need to check for it ourselves.
+                    Log.LogErrorFromResources("Unzip.ErrorExtractingResultsInFilesOutsideDestination", destinationPath.FullName, destinationDirectory.FullName);
+                    continue;
+                }
+                
+                TaskItem taskItem = new TaskItem(EscapingUtilities.Escape(destinationPath.FullName));
+
+                sourceTaskItem.CopyMetadataTo(taskItem);
+
+                _destinationFiles.Add(taskItem);
```
AndyGerlicher: Do we always want this? Seems like it could be a giant list in some cases and use up a lot of memory and may or may not be needed? The output folder might be enough?

jeffkl: Its what `<Copy />` does so I though I'd ... :expressionless: ... 😎 _copy_ it.

Technically you could copy a ton of files as well 

```xml
<ItemGroup>
  <Files Include="Folder\**" />
</ItemGroup>

<Copy
  SourceFiles="@(Files)"
  DestinationFolder="Foo"
  />
```

And we'll waste all of the memory.  What should I do?

CZEMacLeod: @jeffkl Would it save memory to make _unzippedFiles and _destinationFiles a collection of tuples of sourceTaskItem and destinationPath.FullName? That way if you access the output properties DestinationFiles or UnzippedFiles you could build the full list of TaskItems on demand?

jeffkl: We decided to just remove it.  If necessary, we can add the functionality later.

ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
```diff
@@ -1281,6 +1293,15 @@ public partial class XslTransformation : Microsoft.Build.Tasks.TaskExtension
         public Microsoft.Build.Framework.ITaskItem XslInputPath { get { throw null; } set { } }
         public override bool Execute() { throw null; }
     }
+    public sealed partial class ZipDirectory : Microsoft.Build.Tasks.TaskExtension
+    {
+        public ZipDirectory() { }
+        [Microsoft.Build.Framework.RequiredAttribute]
+        public Microsoft.Build.Framework.ITaskItem DestinationFile { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        [Microsoft.Build.Framework.RequiredAttribute]
+        public Microsoft.Build.Framework.ITaskItem SourceDirectory { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
```
rainersigwald: I'm a bit worried about this interface. It's pretty much impossible to get correct incremental behavior out of this. You'd have to:

* Enumerate the contents of the folder _just before_ the target that invokes `ZipDirectory`.
* Additionally compute a hash of those items to avoid the removed-an-input problem.
* Use the new stuff in the target that calls this.

Accepting `Item[]` would be more complex in implementation but easier to make correct.

src/Tasks/ZipDirectory.cs
```diff
@@ -0,0 +1,47 @@
...
+        public ITaskItem DestinationFile { get; set; }
+
+        [Required]
+        public ITaskItem SourceDirectory { get; set; }
+
+        public override bool Execute()
+        {
+            if (!Directory.Exists(SourceDirectory.ItemSpec))
+            {
+                Log.LogErrorFromResources("ZipDirectory.ErrorDirectoryDoesNotExist", SourceDirectory.ItemSpec);
```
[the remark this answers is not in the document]
jeffkl: This is the directory to zip up, so it must exist

src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,187 @@
...
+                }
+            }
+
+            return !_canceling && !Log.HasLoggedErrors;
+        }
+
+        /// <summary>
+        /// Extracts all files to the specified directory.
+        /// </summary>
+        /// <param name="sourceTaskItem">The original <see cref="ITaskItem"/> containing details about the source file.</param>
```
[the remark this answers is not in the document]
jeffkl: I was at one point, good catch

ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
jeffkl: I discussed this with @AndyGerlicher and we're concerned more about the complexity required to make it incremental out-of-the-box.  So we're going to leave it as-is.

AndyGerlicher: I think if we wrote a target that used this we would need to make it incremental. But given that it zips a directory, it doesn't seem worth the complexity to implement at the task level. It might be worth writing a sample in the docs page that is a target that zips your output folder that would be incremental as an example.

jeffkl:

src/MSBuild/Microsoft.Build.CommonTypes.xsd
```diff
@@ -5114,6 +5114,21 @@ elementFormDefault="qualified">
         </xs:complexType>
     </xs:element>
 
+    <xs:element name="Unzip" substitutionGroup="msb:Task">
```
rainersigwald: Don't forget to manually port these to the internal VS repo after this merges.

src/Tasks.UnitTests/Unzip_Tests.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                unzip.Execute().ShouldBeTrue(() => _mockEngine.Log);
+
+                _mockEngine.Log.ShouldContain(Path.Combine(destination.FolderPath, "BE78A17D30144B549D21F71D5C633F7D.txt"), () => _mockEngine.Log);
+                _mockEngine.Log.ShouldContain(Path.Combine(destination.FolderPath, "A04FF4B88DF14860B7C73A8E75A4FB76.txt"), () => _mockEngine.Log);
+            }
+        }
+
+#if  RUNTIME_TYPE_NETCORE
+        [Fact(Skip = "Can't figure out how to make CreateDirectory throw on non-Windows")]
```
rainersigwald: Use

```c#
[PlatformSpecific(TestPlatforms.Windows)]
```

instead.

src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            BuildEngine3.Yield();
+
+            foreach (ITaskItem sourceFile in SourceFiles.TakeWhile(i => !_cancellationToken.IsCancellationRequested))
+            {
+                if (!File.Exists(sourceFile.ItemSpec))
+                {
+                    Log.LogErrorFromResources("Unzip.ErrorFileDoesNotExist", sourceFile.ItemSpec);
+                    continue;
```
cdmihai: Why not eagerly exit to reduce the time it takes to finish the build on error?

cdmihai: Ah, to show all errors at once rather than one at a time. The alternative is to complicate things and check everything first, and do the extraction second.

---
In reply to: [188788554](https://github.com/Microsoft/msbuild/pull/3291#discussion_r188788554) [](ancestors = 188788554)

src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether files should be skipped if the destination is unchanged.
+        /// </summary>
+        public bool SkipUnchangedFiles { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets an array of <see cref="ITaskItem"/> objects containing the paths to .zip archive files to unzip.
+        /// </summary>
+        [Required]
+        public ITaskItem[] SourceFiles { get; set; }
```
cdmihai: What happens if there's files in multiple archives with the same path? Should there be an option to warn or error about this?

src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,193 @@
...
+                {
+                    Log.LogErrorWithCodeFromResources("Unzip.ErrorCouldNotCreateDestinationDirectory", destinationPath.DirectoryName, e.Message);
+                    continue;
+                }
+
+                if (OverwriteReadOnlyFiles && destinationPath.Exists && destinationPath.IsReadOnly)
+                {
+                    try
+                    {
+                        destinationPath.IsReadOnly = false;
```
cdmihai: Should it be made readonly after the write to preserve user data? :)

src/Tasks/ZipDirectory.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (destinationFile.Exists)
+            {
+                if(!Overwrite)
+                {
+                    Log.LogErrorFromResources("ZipDirectory.ErrorFileExists", destinationFile.FullName);
+
+                    return false;
+                }
+
+                File.Delete(destinationFile.FullName);
```
cdmihai: The other task wraps each IO in a try catch. Should this one be wrapped too?

src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,193 @@
...
+                }
+
+                try
+                {
+                    Log.LogMessageFromResources(MessageImportance.Normal, "Unzip.FileComment", zipArchiveEntry.FullName, destinationPath.FullName);
+
+                    using (Stream destination = File.Open(destinationPath.FullName, FileMode.Create, FileAccess.Write, FileShare.None))
+                    using (Stream stream = zipArchiveEntry.Open())
+                    {
+                        stream.CopyToAsync(destination, 81920, _cancellationToken.Token);
```
cdmihai: Shouldn't the returned task be awaited before the entire task returns?

cdmihai: Curious what happens if another file exists in the destination and is longer than the archive entry. Does this override or does it just copy over the source stream, leaving some trailing bytes from the destination? The CopyTo documentation is a bit unclear on this.

cdmihai: Oh, did not notice the FileMode.Create. Nevermind.

---
In reply to: [188796551](https://github.com/Microsoft/msbuild/pull/3291#discussion_r188796551) [](ancestors = 188796551)

src/Tasks/Unzip.cs
jeffkl: I think its okay to just ignore it, do you want me to log something?

src/Tasks/Unzip.cs
jeffkl: I'm not sure.  I'd assume that if the file is readonly in the archive it should be readonly after extraction but I don't see attributes like that in the `ZipArchiveEntry` class...  It seems a little weird to be setting the readonly attribute based on what the file _used_ to have.  Thoughts?

rainersigwald:
`Copy` doesn’t restore the RO bit, so I’d say don’t do it here either.

jeffkl:

jeffkl:
@cdmihai can you please review my changes to address your comments?  I think this is ready to merge but wanted to make sure you were happy.

src/Tasks/ZipDirectory.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        public bool Overwrite { get; set; }
+
+        /// <summary>
+        /// Gets or sets a <see cref="ITaskItem"/> containing the full path to the source directory to create a zip archive from.
+        /// </summary>
+        [Required]
+        public ITaskItem SourceDirectory { get; set; }
+
+        public override bool Execute()
```
cdmihai: Can you please make this yieldable as well?

jeffkl:

src/Tasks/Unzip.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <summary>
+        /// Extracts all files to the specified directory.
+        /// </summary>
+        /// <param name="sourceArchive">The <see cref="ZipArchive"/> containing the files to extract.</param>
+        /// <param name="destinationDirectory">The <see cref="DirectoryInfo"/> to extract files to.</param>
+        private void Extract(ZipArchive sourceArchive, DirectoryInfo destinationDirectory)
+        {
+            foreach (ZipArchiveEntry zipArchiveEntry in sourceArchive.Entries.TakeWhile(i => !_cancellationToken.IsCancellationRequested))
+            {
+                FileInfo destinationPath = new FileInfo(Path.Combine(destinationDirectory.FullName, zipArchiveEntry.FullName));
```
natemcmaster: Sorry to comment after you've just merged, but would you consider normalizing directory separators? One of the most common errors I've encountered is that zips created on Windows using `\` as a directory separator will not extract correctly on Linux/macOS. Instead `\` becomes part of the filename.
image

This common error is something we worked around in the unzip task used in the .NET Core build system: see https://github.com/dotnet/arcade/blob/0d117c0b3649565a6d9aacf9f435e29ab67c3c0d/src/Microsoft.DotNet.Build.Tasks.IO/src/UnzipArchive.cs#L64-L76

jeffkl:
@natemcmaster I'll double check but I _think_ my unit tests cover that scenario.

src/Tasks/ZipDirectory.cs
```diff
@@ -0,0 +1,85 @@
...
+                        Log.LogErrorFromResources("ZipDirectory.ErrorFailed", sourceDirectory.FullName, destinationFile.FullName, e.Message);
+
+                        return false;
+                    }
+                }
+
+                try
+                {
+                    Log.LogMessageFromResources(MessageImportance.High, "ZipDirectory.Comment", sourceDirectory.FullName, destinationFile.FullName);
+                    ZipFile.CreateFromDirectory(sourceDirectory.FullName, destinationFile.FullName);
```
Flithor, line 70: I think need make the "includeBaseDirectory" argment as an option

Issue #3220, opened 2018-04-24
Proposal: DependsOnTask for project items Include


kasper3:
For `Problem statement` and `Part 1`, see https://github.com/Microsoft/msbuild/issues/3219

Part 2:

```xml
<Compile
  Include="$(MSBuildProjectDirectory)\SharedDTOs\RemoteFile.vb"
  DependsOnTask="DownloadMySourceFile"
  Condition="!Exist($(MSBuildProjectDirectory)\SharedDTOs\RemoteFile.vb)" />
```

this will indicate the design-time or first-evaluation-time to run the task before evaluating this line, ignore the task result and continue evaluation once task is finished. If task fails (due to insufficient credentials / no internet connection / wrong URL / any other reason), a reason warning will be issued and subsequent steps will fail as they fail today:

> CSC : error CS2001: Source file 'C:\testproj\nonexisting.cs' could not be found.

for `<Compile>` and for `<Content>`

> error : File not found: 'C:\testproj\mockimage.png'

etc.

rainersigwald:
I'm not so sure about this. Would it apply to evaluation-time (as opposed to created-during-execution) items? That is, is it legal to have

```xml
<Project>
  <ItemGroup>
    <I Include="foo.bar" DependsOnTask="CreateFoo" />
  </ItemGroup>
</Project>
```

? What about

```xml
<Project InitialTargets="IncludeFoo">
  <Target Name="IncludeFoo">
    <ItemGroup>
      <I Include="foo.bar" DependsOnTask="CreateFoo" />
    </ItemGroup>
  </Target>
</Project>
```

? How does the execution differ?

Do you want to call a task or a target with this? If a task, how would you pass arguments to the task, or would it be a requirement that the task not take any arguments?

CZEMacLeod:
I think you might need some kind of metabase to store the etag, last modified, cache control settings of the downloaded file. Also control over when the file is updated - some kind of max query rate control that can be applied by the consumer. If this is just to restore missing files, that is one thing, but if it is to download and keep a file up to date that is another...
Is it applicable to have DownloadFiles as an ItemGroup and maybe make it visible separately in the UI (e.g. Visual Studio) with a property page to set the metadata.
I think the downloaded files would be stored in the IntermediateOutputFiles path or similar...

```
<ItemGroup>
	<DownloadFile Include="https://my.company.com/logo.png">
		<TargetGroup>Content</TargetGroup>
[...]

PR #3297, opened 2018-05-11
Add DownloadFile task

### LINKED ISSUE #3219: Proposal: DownloadFile task
Problem statement:

The mechanics require to accomplish:

* Download a file before build
* Include as Content, Compiler or None in the build / package

are quite cumbersome. Some folks accomplish this by downloading the content before invoking msbuild (isolated mode). Others use `<Exec>` and utilities like Invoke-WebRequest or curl-wget etc. Then there are people like me who are using [Roslyn]CodeTaskFactory and HttpClient in UsingTask to avoid the cross-platform disparities. But then reevaluation of package/compile targets makes it harder if you want to include the file in the build.

Part 1:

If we could get a decent download task that takes some basic parameters like URL, Destination, Credentials and use HttpClient behind the scenes from msbuild, that will mitigate the pain.

```xml
<DownloadFile
  TaskName="DownloadMySourceFile"
  Url="https://raw.githubusercontent.com/andreburto/FtpThing/d0e133c/FtpThing/RemoteFile.vb"
  Destination="$(MSBuildProjectDirectory)\SharedDTOs\RemoteFile.vb" />
```

Part 2 is moved to #3220

If you think it is common enough to be considered out of the box, please share your thoughts.

jeffkl:
* Supports cancellation
* Supports retries (off by default)
* Supports skipping download if file is up-to-date (on by default)
* Supports getting file name from response headers or URL if possible
* Has output parameter with downloaded path

Fixes #3219 

Follow-up items:

- [ ] Documentation for DownloadFile task


rainersigwald:
Mention "closes #3219" please. fyi @kasper3

jeffkl:
> Mention "closes #3219" please.

#3219 sounds like it wants to have a way to download a file _before_ evaluation.  In my opinion we shouldn't do that so I just said this PR is related.

src/Tasks/DownloadFile.cs
```diff
@@ -0,0 +1,280 @@
...
+        /// </summary>
+        /// <param name="uri">The <see cref="Uri"/>.</param>
+        /// <param name="filename">Receives the name of the file.</param>
+        /// <returns><code>true</code> if a file name could be determined, otherwise <code>false</code>.</returns>
+        private bool TryGetFileName(Uri uri, out string filename)
+        {
+            // Not all URIs contain a file name so users will have to specify one
+            // Example: http://www.download.com/file/1/
+
+            filename = !String.IsNullOrWhiteSpace(DestinationFileName?.ItemSpec)
```
kasper3: Could we call TrygetFileName after we have recieved the headers?

Precedence (descending order):

* `DestinationFileName` (user-specified)
* `client.ResponseHeaders["content-disposition"]`, if `DestinationFileName` is empty
* `url.LocalPath`, if `DestinationFileName` is empty and `client.ResponseHeaders["content-disposition"]` is non-empty and meaningful (`<file-name>.<extension>` format)

so we can call urls like https://lorempizza.com/380/240, we get this kind of filename from response header: `imagedoc-darknoise.png` (instead of 240)

kasper3:
@jeffkl, first of all kudos on the good HttpClient synchronous implementation with the disposables in correct place! 👍 

I have left one comment about content disposition, which @CZEMacLeod mentioned in #3220.

Secondly, this utility is good as it is, we can safely say that it addresses #3219.

> #3219 sounds like it wants to have a way to download a file *before* evaluation. In my opinion we shouldn't do that so I just said this PR is related.

True, that is #3220, we can discuss it separately to make `DependsOnTask` semantics for design-time evaluation friendlier. In VS's Solution Explorer, the delayed inclusions (`<ItemGroup> <Compile Include="/some/other/root/**/*.cs" /> </ItemGroup>` inside the target, does not get showed up, but MSBuild does compile), so the issue is to have some friendly/implicit semantics for consumer to reevaluate the design time target, that aids the VS Solution Explorer. Hence the idea of `DependsOnTask` to prevent consumer from entering the domain of "doing too much" in their project files. If that is technically impossible, then I accept that. :)

AndyGerlicher:
Only concern I had here is the `SkipUnchangedFiles` flag just checking the size and defaulting to `true`. I could easily see this getting used to download a version file and ending up in scenarios that are hard to debug  when the file doesn't always change lengths. For binaries it's probably ok though?

One thought that's not a great solution but could help is always download when under a certain size or content type?

CZEMacLeod:
@AndyGerlicher I don't know how you would store the information, but similarly to the Content-Disposition, perhaps the caching headers could get used, and/or store the ETag? Could they get stored in alternate streams on the file, or is that too platform specific? Some kind of metadata cache file would allow for this. If the file exists, you could use the last updated date and add the If-Modified-Since header to prevent downloading the file again... Or use the info about when it last downloaded from the aforementioned metadata.

jeffkl:
I found a `LastModified` property ([here](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpcontentheaders.lastmodified?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev15.query%3FappId%3DDev15IDEF1%26l%3DEN-US%26k%3Dk(System.Net.Http.Headers.HttpContentHeaders.LastModified);k(DevLang-csharp)%26rd%3Dtrue&view=netframework-4.7.2)) that I'm going to use.

If the property is not set, then it will download the file again.

magol:
Does it support download via other protocols than http(s), such as FTP or SMB/CIFS?

jeffkl:
@magol 

> Does it support download via other protocols than http(s), such as FTP or SMB/CIFS?

Not at the moment, it only supports HTTP endpoints

magol:
@jeffkl 
> Not at the moment, it only supports HTTP endpoints

But it is possible that it is added later?

jeffkl:
@magol possibly.  This currently does a simple non-authenticated HTTP GET.  If you need more functionality, please feel free to contribute.

PR #3301, opened 2018-05-12
Add support for overriding MSBuildExtensionsPath search paths via a property

radical:
Add property allowing users to override fallback paths for MSBuildExtensionsPath.

For OSX, this essentially sets the fallback paths to:

`<property name="MSBuildExtensionsPath" value="$(MSBuildExtensionsPathFallbackPathsOverride);/Library/Frameworks/Mono.framework/External/xbuild/" />`

.. whichs allows the default fallback paths to be overridden by setting
a property - `$(MSBuildExtensionsPathFallbackPathsOverride)`.

One of the useful use cases for this is to be able to run tests with
in-tree builds of assemblies/targets meant for
`$(MSBuildExtensionsPath)/`. Simply changing `$(MSBuildExtensionsPath)` would
mean that the in-tree builds of the custom frameworks would work but the
default ones would not be found!

### DIFF
--- a/src/Build/Definition/ProjectImportPathMatch.cs
+++ b/src/Build/Definition/ProjectImportPathMatch.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Build.BackEnd;
 using Microsoft.Build.Shared;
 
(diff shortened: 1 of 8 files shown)


src/Build/Definition/ProjectImportPathMatch.cs
```diff
@@ -46,15 +64,44 @@ public ProjectImportPathMatch(INodePacketTranslator translator)
...
+            /// did not have any property reference and so we just return the list we prepared during
+            /// construction.
+            return SearchPathsWithNoExpansionRequired ?? SplitSearchPaths(expandPropertyReferences(PropertyValue));
+        }
+
+        /// <summary>
+        /// Splits @fullString on @_separatorForExtensionsPathSearchPaths
+        /// </summary>
+        /// FIXME: handle ; in path on Unix
+        private List<string> SplitSearchPaths(string fullString) => fullString
```
marek-safar: Could be static

src/Build/Definition/ProjectImportPathMatch.cs
```diff
@@ -46,15 +64,44 @@ public ProjectImportPathMatch(INodePacketTranslator translator)
...
+                                                                        .Split(new[] {_separatorForExtensionsPathSearchPaths}, StringSplitOptions.RemoveEmptyEntries)
+                                                                        .Distinct().ToList();
+
+        /// <summary>
+        /// Returns true if @value might have a property reference
+        /// </summary>
+        private bool HasPropertyReference(string value) => value.Contains("$(");
```
marek-safar: Could be static

AndyGerlicher:
Sorry for the delay getting back. I wanted to review this carefully to make sure I understood all the implications.

In general, I don't think "fallback" is a good feature. It served a purpose and it's implemented in a simple way and hasn't caused too many issues. I don't regret adding it and I would still make the same decision, but I don't think expanding on it is a good idea. Having the engine do more like this is asking to make things harder to debug. And I think it's one of those that effort should be spent avoiding usage of the feature entirely (`MSBuildExtensionsPath` should be a deprecated concept). But I get that it's not easy so I'm not unwilling to budge here. But I want to make sure we're not missing another solution.

Just trying to understand this feature, here's a scenario I came up with:
```XML
<Project>
  <Import Project="$(TestWithFallback)\foo.props" />
  <Import Project="$(TestWithFallback)\foo.props" />
  <Target Name="Build" >
    <Message Importance="High" Text="TestWithFallback: $(TestWithFallback)" />
    <Message Importance="High" Text="FooProp: $(FooProp)" />
  </Target>
</Project>
```

If `foo.props` redefines `$(TestFallbackPath)` then that isn't a duplicate import. But importing `<Import Project="$(TestWithFallback)\*.props" />` will not import both files as it would have previously and changes the result of this build. And with all this `TestWithFallback: $(TestWithFallback)` comes back with the original value which I made invalid in my case just to make it harder to debug. It becomes a very point-in-time concept with no real way to debug or trace at that point in time.


Also as a side note, this allocates on every usage of fallback and would likely set off perf regression tests. These are code paths where that need to be de-linqified and reduce any allocation possible.

radical:
Closing this. We are using custom tools `xabuild` and `xibuild` instead now which generate a custom `app.config` with the extra fallback paths.

PR #3306, opened 2018-05-14
Display informational version

### LINKED ISSUE #918: Dev versions of MSBuild should put the version with commit ID in the logo
We get banners like

```
Microsoft (R) Build Engine version 15.1.188.30335
```

But it'd be really nice (for like 8 people) if that had the full form of the version that includes the git commit ID for easy checkout.

jeffkl:
Closes #918

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -3476,7 +3476,7 @@ namespace Microsoft.Build.CommandLine
             const string frameworkName = ".NET Framework";
 #endif
 
-            Console.WriteLine(ResourceUtilities.FormatResourceString("CopyrightMessage", ProjectCollection.Version.ToString(), frameworkName));
+            Console.WriteLine(ResourceUtilities.FormatResourceString("CopyrightMessage", ThisAssembly.AssemblyInformationalVersion, frameworkName));
         }
 
         /// <summary>
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -3476,7 +3476,11 @@ namespace Microsoft.Build.CommandLine
             const string frameworkName = ".NET Framework";
 #endif
 
+#if THISASSEMBLY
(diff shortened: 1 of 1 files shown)


jeffkl:
Sample:

```
Microsoft (R) Build Engine version 15.8.60-preview+gb69bce195d for .NET Framework

15.8.60-preview+gb69bce195d
```

src/MSBuild/XMake.cs
```diff
@@ -3476,7 +3476,7 @@ private static void DisplayCopyrightMessage()
             const string frameworkName = ".NET Framework";
 #endif
 
-            Console.WriteLine(ResourceUtilities.FormatResourceString("CopyrightMessage", ProjectCollection.Version.ToString(), frameworkName));
+            Console.WriteLine(ResourceUtilities.FormatResourceString("CopyrightMessage", ThisAssembly.AssemblyInformationalVersion, frameworkName));
```
rainersigwald: 🎉

src/MSBuild/XMake.cs
```diff
@@ -3539,7 +3539,7 @@ private static void ShowHelpPrompt()
         /// </summary>
         private static void ShowVersion()
         {
-            Console.Write(ProjectCollection.Version.ToString());
+            Console.WriteLine(ThisAssembly.AssemblyInformationalVersion);
```
rainersigwald: This one I'm more worried about. Think people are parsing the output of `msbuild /nologo /version` and will be broken by a non-`System.Version` compatible output? I suspect they might :(

jeffkl: agreed

src/MSBuild/XMake.cs
cdmihai: `ThisAssembly` is not available during sourdebuild builds (as git versioning is turned off). There's a `THISASSEMBLY` symbol you can ifdef on :(

jeffkl: Added

PR #3309, opened 2018-05-14
Don't race on multitargeted project -> skipped compat check

### LINKED ISSUE #2366: Build failure related to dependencies in a mixed solution
After converting a subset of projects from legacy project system to new in https://github.com/antlr/antlrcs/pull/74/commits/4b0ae50a1fcc01e1f149c0d79272fb69da0a423b, I found that the build failed due to a file access error until I turned off parallel builds (removed `/m`) in https://github.com/antlr/antlrcs/pull/74/commits/407a344523fa4bd1bf9492827156bc2dfd9b2982. I'm guessing this is caused by a failure to properly track dependencies when the build contains both legacy and new projects.

rainersigwald:
A multitargeted project referencing a non-SDK project had a race
condition. The inner builds each had `TargetFramework` set as a global
property, and each passed that global property along when building
referenced non-SDK projects. Since those projects aren't aware of TF,
they raced each other.

Fixes #2366 for vcxproj references by ensuring that referenced projects
which skipped TF checks also don't inherit TF/RID.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1688,9 +1688,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
       <!--
          Remove the items we've touched from _MSBuildProjectReferenceExistent. This will leave all projects where
-         SkipGetTargetFrameworkProperties was set. Then add all AnnotatedProjects back.
+         SkipGetTargetFrameworkProperties was set.
       -->
       <_MSBuildProjectReferenceExistent Remove="@(_MSBuildProjectReferenceExistent)" Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' != 'true'" />
+
+      <!-- If we skipped querying the other project, ensure we don't pollute it with the TargetFramework/RuntimeIdentifier
+           that may be set for this project as a global property (if we're in an inner build). -->
+      <_MSBuildProjectReferenceExistent>
+          <UndefineProperties>%(_MSBuildProjectReferenceExistent.UndefineProperties);TargetFramework;RuntimeIdentifier</UndefineProperties>
+      </_MSBuildProjectReferenceExistent>
+
+      <!-- Then add all AnnotatedProjects back. -->
       <_MSBuildProjectReferenceExistent Include="@(AnnotatedProjects)" />
     </ItemGroup>
   </Target>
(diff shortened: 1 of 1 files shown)


rainersigwald:
cc @nguerrera

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1688,9 +1688,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+      <!-- If we skipped querying the other project, ensure we don't pollute it with the TargetFramework/RuntimeIdentifier
+           that may be set for this project as a global property (if we're in an inner build).
+
+           But allow setting TF explicitly. -->
+      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SetTargetFramework)' == ''">
+          <UndefineProperties>%(_MSBuildProjectReferenceExistent.UndefineProperties);TargetFramework</UndefineProperties>
+      </_MSBuildProjectReferenceExistent>
+      <_MSBuildProjectReferenceExistent>
+          <UndefineProperties>%(_MSBuildProjectReferenceExistent.UndefineProperties);RuntimeIdentifier</UndefineProperties>
```
rainersigwald: @nguerrera Do you think we should clear RID when traversing to a project that had `SkipGetTargetFrameworkProperties=true`? We weren't before. Maybe I should just leave it out here too?

nguerrera: We need to run this through the SDK tests as we did for the original double eval fix. I forget all of the subtleties of RID cases that @TheRealPiotrP worked through.

nguerrera: cc @dsplaisted His recent work might make running those tests against an arbitrary msbuild  (or at least CLI) easier.

rainersigwald: That'd be exciting! It would be nice to see if we could get an optional Jenkins test to run SDK tests on new MSBuild bits (long term . . .)

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1688,9 +1688,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
       <_MSBuildProjectReferenceExistent Remove="@(_MSBuildProjectReferenceExistent)" Condition="'%(_MSBuildProjectReferenceExistent.SkipGetTargetFrameworkProperties)' != 'true'" />
+
+      <!-- If we skipped querying the other project, ensure we don't pollute it with the TargetFramework/RuntimeIdentifier
+           that may be set for this project as a global property (if we're in an inner build).
+
+           But allow setting TF explicitly. -->
+      <_MSBuildProjectReferenceExistent Condition="'%(_MSBuildProjectReferenceExistent.SetTargetFramework)' == ''">
+          <UndefineProperties>%(_MSBuildProjectReferenceExistent.UndefineProperties);TargetFramework</UndefineProperties>
+      </_MSBuildProjectReferenceExistent>
+      <_MSBuildProjectReferenceExistent>
```
jeffkl: Can we add a comment to this item as well?

rainersigwald:
Closing this due to age.

PR #3313, opened 2018-05-15
Add RoslynCodeTaskFactory and DownloadFile samples

jeffkl:
DownloadFile task was added by https://github.com/Microsoft/msbuild/pull/3297

RoslynCodeTaskFactory was added by https://github.com/Microsoft/msbuild/pull/3175

Delete empty `src/Samples/Directory.Build.targets`


src/Samples/DownloadFile/DownloadFile.targets
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
rainersigwald: I'd say to not specify the xml header or the namespace in samples, since they're no longer needed.

src/Samples/DownloadFile/DownloadFile.targets
```diff
@@ -0,0 +1,18 @@
...
+          AfterTargets="PrepareForBuild">
+
+    <!-- Download a file -->
+    <DownloadFile
+      SourceUrl="https://raw.githubusercontent.com/Microsoft/msbuild/master/LICENSE"
+      DestinationFolder="$(MSBuildProjectDirectory)">
+      <Output TaskParameter="DownloadedFile" ItemName="Content" />
+    </DownloadFile>
+
+    <!-- Include the f-->
```
rainersigwald: the f?

src/Samples/DownloadFile/DownloadFile.targets
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="DownloadLicense"
+          AfterTargets="PrepareForBuild">
+
+    <!-- Download a file -->
+    <DownloadFile
+      SourceUrl="https://raw.githubusercontent.com/Microsoft/msbuild/master/LICENSE"
+      DestinationFolder="$(MSBuildProjectDirectory)">
```
rainersigwald: Can you specify a path or something so this can be named `DownloadedFile` instead of `LICENSE`? I found the `.gitignore` confusing.

jeffkl:
> I think I'd rather see all of this in the Docs repo rather than here. . . Do you think it needs to be here?

I wanted more test coverage to ensure that we don't break at least RoslynCodeTaskFactory.

kasper3:
> I wanted more test coverage

👍 also it is bit strange when we search and find zero usages of built-in task in the repo.

PR #3205, opened 2018-04-17
FallbackTargetFrameworks in P2P refs

rainersigwald:
Allows compatibility between a ProjectReference that doesn't build for
a directly-compatible TargetFramework but is compatible through
NuGet's AssetTargetFallback rules.

Closes #2934.

PR #3315, opened 2018-05-15
Skip tf-compat checks for solution dependencies

### LINKED ISSUE #1915: Solution-Level Project Dependencies require framework compatibility between projects in metaproj
If you build a VS solution that has project dependencies outside of project references from the command line with MSBuild, the metaproj requires that the two projects have compatible frameworks. Sometimes this is not desirable (i.e. building a tool used to codegen before building projects that use the generated code).  I have 
[an example](https://github.com/Microsoft/msbuild/files/873625/MetaProjectSolutionDependencies.zip) that has a .NET Core project that has a Project Dependency on a .NET 4.5 project. It builds successfully in VS, but command line MSBuild fails with the following:

```
"C:\Users\jkori\OneDrive\Documents\Visual Studio 2017\Projects\MetaProjectSolutionDependencies\MetaProjectSolutionDepen
dencies.sln" (default target) (1) ->
"C:\Users\jkori\OneDrive\Documents\Visual Studio 2017\Projects\MetaProjectSolutionDependencies\NetCore\NetCore.csproj.m
etaproj" (default target) (2) ->
"C:\Users\jkori\OneDrive\Documents\Visual Studio 2017\Projects\MetaProjectSolutionDependencies\NetCore\NetCore.csproj"
(default target) (4) ->
"C:\Users\jkori\OneDrive\Documents\Visual Studio 2017\Projects\MetaProjectSolutionDependencies\NetFramework\NetFramewor
k.csproj" (GetTargetFrameworkProperties target) (3:2) ->
(GetTargetFrameworkProperties target) ->
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.Sdk.
Common.targets(73,5): error : Project 'C:\Users\jkori\OneDrive\Documents\Visual Studio 2017\Projects\MetaProjectSolutio
nDependencies\NetFramework\NetFramework.csproj' targets '.NETFramework,Version=v4.5'. It cannot be referenced by a proj
ect that targets '.NETCoreApp,Version=v1.1'. [C:\Users\jkori\OneDrive\Documents\Visual Studio 2017\Projects\MetaProject
SolutionDependencies\NetFramework\NetFramework.csproj]

    0 Warning(s)
    1 Error(s)
```

rainersigwald:
Explicitly opts synthetic project references injected by solution
dependencies out of the TargetFramework-compatibility/find-best-match
dance instituted in 15.x for .NET Standard support.

Fixes #1915 and the most common cause of #2661.

Note: I proposed this in https://github.com/Microsoft/msbuild/pull/2867, but we decided not to take it, as it seemed redundant with the change to ignore TF compat when `ReferenceOutputAssembly=false`. But that had to be rolled back, because several repos depended on TF checks in that case. This is more scoped: it affects only solution build order dependencies. Hopefully that means that it only unblocks people and doesn't create new breaks.

### DIFF
--- a/src/Tasks/ResolveProjectBase.cs
+++ b/src/Tasks/ResolveProjectBase.cs
@@ -293,6 +293,7 @@ namespace Microsoft.Build.Tasks
                     item.SetMetadata("ReferenceOutputAssembly", "false");
                     item.SetMetadata("LinkLibraryDependencies", "false");
                     item.SetMetadata("CopyLocal", "false");
+                    item.SetMetadata("SkipGetTargetFrameworkProperties", "true");
 
                     updatedProjectReferenceList.Add(item);
                 }


rainersigwald:
The longer-term fix for #2661 is #3205, but that requires changing project files (or using an import) in a way that we can't do automatically while building a solution. I think it's worth doing this to avoid that for users.

PR #3321, opened 2018-05-17
R# for CopyTask, copy tests

erikma:


VENESSAMARTINE03:
Please stop emailing me

src/Tasks.UnitTests/Copy_Tests.cs
```diff
@@ -381,13 +387,25 @@ public void CopyOverReadOnlyFileParameterIsSetWithDestinationFolder()
                 Directory.CreateDirectory(destinationFolder);
 
                 using (StreamWriter sw = FileUtilities.OpenWrite(source1, true))
+                {
                     sw.Write("This is a source file1.");
+                }
+
                 using (StreamWriter sw = FileUtilities.OpenWrite(source2, true))
+                {
+
```
rainersigwald: nit: unnecessary blank line

src/Shared/UnitTests/MockEngine.cs
```diff
@@ -198,14 +154,10 @@ internal string Log
 
                 _log.Clear();
             }
-            get { return _log.ToString(); }
+            get => _log.ToString();
```
jeffkl: It seems weird that the getter is after the setter, do you mind fixing that?  Its like that in this whole file for some reason

src/Tasks.UnitTests/Copy_Tests.cs
erikma: Fixed.

erikma: RAR reverted in next push

src/Shared/UnitTests/MockEngine.cs
erikma: Good catch. Reversed in whole file

PR #3324, opened 2018-05-17
Add source build switches in order to be able to build source-build on windows

safern:
I'm not sure if this should go against master.

The windows build script was missing the DotNetBuildFromSource and DotNetCoreSdkDir parameters in order to build msbuild on source-build as we do for non-Windows.

cc: @rainersigwald @weshaggard

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -14,6 +14,8 @@ Param(
   [switch] $bootstrapOnly,
   [string] $verbosity = "minimal",
   [string] $hostType,
+  [switch] $DotNetBuildFromSource,
+  [string] $DotNetCoreSdkDir = "",
   [Parameter(ValueFromRemainingArguments=$true)][String[]]$properties
 )
 
@@ -179,7 +181,13 @@ function KillProcessesFromRepo {
 }
 
 function Build {
-  InstallDotNetCli
+  if (![string]::IsNullOrEmpty($DotNetCoreSdkDir) -and (Test-Path -Path $DotNetCoreSdkDir)) {
+    $env:DOTNET_INSTALL_DIR = $DotNetCoreSdkDir
+  }
+  else {
+    InstallDotNetCli
+  }
+
   $env:DOTNET_HOST_PATH = Join-Path $env:DOTNET_INSTALL_DIR "dotnet.exe"
 
   if ($prepareMachine) {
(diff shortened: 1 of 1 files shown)


build/build.ps1
```diff
@@ -210,9 +218,18 @@ function Build {
 
   echo "Repo toolset used from: $RepoToolsetBuildProj"
 
-  $solution = Join-Path $RepoRoot "MSBuild.sln"
+  $createTlb = $true
```
rainersigwald: Don't explicitly set this to `true`; that'll try to create TLBs for projects that don't need them. Set it to false or leave it unset.

safern: Ok, sounds good.

cdmihai:
Why the need to support sourcebuild on Windows? Wasn't it needed only for Redhat?

build/build.ps1
```diff
@@ -210,9 +218,18 @@ function Build {
...
+    $solution = Join-Path $RepoRoot "MSBuild.SourceBuild.sln"
+    $createTlb = $false
+  }
+  else
+  {
+    $solution = Join-Path $RepoRoot "MSBuild.sln"
+  }
 
-  $commonMSBuildArgs = "/m", "/clp:Summary", "/v:$verbosity", "/p:Configuration=$configuration", "/p:Projects=$solution", "/p:CIBuild=$ci", "/p:RepoRoot=$reporoot"
+  $commonMSBuildArgs = "/m", "/clp:Summary", "/v:$verbosity", "/p:Configuration=$configuration", "/p:Projects=$solution", "/p:CIBuild=$ci", "/p:RepoRoot=$reporoot", "/p:DisableNerdbankVersioning=$DotNetBuildFromSource", "/p:CreateTlb=$createTlb"
```
cdmihai: The NerdbankVersioning restriction comes from Redhat. Is this still valid for Windows?

safern:
> Why the need to support sourcebuild on Windows? Wasn't it needed only for Redhat?
 
We want our developers to be able to produce an SDK or a runtime with their changes using source build infrastructure to simplify this process instead of having them to build all the repos on the stack.

weshaggard:
>Why the need to support sourcebuild on Windows? Wasn't it needed only for Redhat

Redhat is one of the first targets but not only one. We need to make the source-build work on all the OS's. For our devs working on the full product on windows it is still very much interesting to build the entire stack on windows.

build/build.ps1
safern: I've deleted this as we're already passing this msbuild parameter through source-build: 

https://github.com/dotnet/source-build/blob/dev/release/2.1/repos/msbuild.proj#L9

build/build.ps1
safern: Fixed.

PR #3340, opened 2018-05-24
Fix master source-build

safern:
In 15.7_sourcebuild we set explicitly the MicrosoftNETCoreAppVersion and add the packagereference to Microsoft.NETCore.App to assure we get the right version of reference assemblies to build against. My theory here is that we use the full msbuild to build the *.sln and not dotnet msbuild to build and that is causing to take VS's sdk instead of the bootstraped cli.

Just for reference this is what we have in 15.7_sourcebuild:

https://github.com/Microsoft/msbuild/blob/vs15.7_sourcebuild/src/Directory.Build.props#L63

https://github.com/Microsoft/msbuild/blob/vs15.7_sourcebuild/build/Versions.props#L50

cc: @rainersigwald @cdmihai

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -47,6 +47,7 @@
   <!-- Production Dependencies -->
   <PropertyGroup>
     <VisualStudioSetupInteropVersion>1.16.30</VisualStudioSetupInteropVersion>
+    <MicrosoftNETCoreAppVersion Condition="'$(DotNetBuildFromSource)' == 'true'">2.1.0-rc1</MicrosoftNETCoreAppVersion>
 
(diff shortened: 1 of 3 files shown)


rainersigwald:
If we need to do this, we should probably do it as an undo of https://github.com/Microsoft/msbuild/pull/3263 instead. Specifically, I think it'd need the `DisableImplicitFrameworkReferences` part.

But this shouldn't be necessary. It sounds like this build configuration isn't picking up the private .NET Core SDK that gets downloaded (/pointed to the sourcebuilt location). Can we chase that instead?

PR #3375, opened 2018-06-06
Specify ngenApplication and avoid ngening satellite assemblies

tmat:
Without `vs.file.ngenApplication` VS uses configuration of devenv process for NGEN, which might have different binding redirects than `msbuild.exe.config`.

### DIFF
--- a/src/Package/MSBuild.VSSetup/files.swr
+++ b/src/Package/MSBuild.VSSetup/files.swr
@@ -11,20 +11,20 @@ folder InstallDir:\MSBuild\15.0
   file source=$(ThirdPartyNotice)
 
 folder InstallDir:\MSBuild\15.0\Bin
-  file source=$(MSBuildConversionBinPath)Microsoft.Build.Conversion.Core.dll vs.file.ngenArchitecture=all
-  file source=$(X86BinPath)Microsoft.Build.dll vs.file.ngenArchitecture=all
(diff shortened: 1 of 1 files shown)


AndyGerlicher:
@tmat I'm not sure this is what we want though. Ultimately perf in Visual Studio more important (or higher impact/urgency at least) than command-line. We need to be very mindful to keep our binding redirects in sync with devenv.exe because of this, but ultimately for sln load and other perf scenarios I believe using devenv.exe to optimize is what we want?

tmat:
Are these binaries loaded to devenv process?

AndyGerlicher:
Design time builds happen **on the UI thread**, so yeah :).

Not sure about ngen and satellite assemblies though. I would assume they should be ngen and then yes same rules apply, VS will need resources that come from those.

tmat:
Re satellite assemblies - they have no code, so there is no point in NGENing them.

tmat:
Also, why don't we run design time builds in msbuild.exe process and send back to devenv only the output values the project system needs? @davkean

AndyGerlicher:
Yeah, updating immutable is a huge nightmare. Everyone in VS will have to do it in a couple places. Us being one of them. Has Roslyn updated already?

tmat:
Closing for now. Updating SCI and SRM instead: https://github.com/Microsoft/msbuild/pull/3376

PR #3380, opened 2018-06-07
LOC CHECKIN | Microsoft/msbuild vs15.8 | 20180607

v-zbsail:


AndyGerlicher:
We're about to do an RI to 15.8 so we'll close this to avoid conflicts.

PR #61, opened 2015-04-06
add AppVeyor build, compile and run unit tests

ctaggart:
This allows this project and forks of this project to easily build and run all the unit tests on AppVeyor. Example test output:
https://ci.appveyor.com/project/ctaggart/msbuild/build/12/tests

PR #3384, opened 2018-06-07
[mono] Fix VSMac build with .netcore 2.1 installed (#61)

radical:
For the mono build we have a dependency on
`Microsoft.Extensions.DependencyModel`, and so we get the assembly in
the binDir.

When building a netstandard project, the MSBuild sdk resolver resolves
to the installed dotnet (2.1 RC in this case).
`Microsoft.NET.Build.Tasks` is loaded from there but it's dependency is
available in msbuild's(entry assembly) binDir itself, so it gets loaded
from there.  And because the one from dotnet has a new
class `Microsoft.Extensions.DependencyModel.RuntimeFile`,
this fails at runtime with:

`System.TypeLoadException: Could not resolve type with token 01000030 from typeref (expected class 'Microsoft.Extensions.DependencyModel.RuntimeFile' in assembly 'Microsoft.Extensions.DependencyModel, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60')`

This assembly isn't required for mono msbuild because it is meant for
reading `.deps` files. And this does not affect building netcoreapps,
because in that case the sdk resolver uses the installed `dotnet`. And
if it isn't installed, then we just use our bundled SDK which already
has this assembly.

Fixes bug: https://github.com/mono/mono/issues/8715

### DIFF
--- a/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj
+++ b/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj
@@ -27,9 +27,6 @@
 
     <!-- Use the NuGet.targets that we bundle, since that is what we will install so we should be testing against that -->
     <Content Include="$(RepoRoot)nuget-support\tasks-targets\*" CopyToOutputDirectory="PreserveNewest" />
-
-    <!-- Include DependencyModel libraries. -->
-    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelVersion)" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(MonoBuild)' != 'true'">


AndyGerlicher:

AndyGerlicher:
@dsplaisted any idea on this? Looks like you added it. Not sure what this is for :(. But given this is removing just for mono I think we should be able to take this if it helps? And it won't actually effect the produced binaries.

dsplaisted:
I don't remember the details, but I think I probably just included this for Mono for parity with .NET Core.  At the time I don't think Mono was working in the repo, so it was probably never tested.

radical:
I had added it when getting the mono build to work and I did that for parity with the !mono builds.

PR #3387, opened 2018-06-07
[tests][mono] Fix mono/osx test - Use Path.Combine to construct paths

radical:
Microsoft.Build.UnitTests.BackEnd.IntrinsicTask_Tests.RemoveComplexMidlExample


radical:
Looking at the failures

radical:
This was a result of a mono specific patch, and should be looked at when that patch has to be upstreamed. Closing for now. Sorry about the noise!

PR #3393, opened 2018-06-08
Catch a possible exception when trying to resolve from HKLM (#4)

radical:
On Mono, RegistryKey.OpenSubKey will attempt to create the registry hive's on-disk folder, if it doesn't exist. For HKLM, that folder is in /etc/mono/registry - but only root has write access to /etc, so the constructor throws an exception. Let's avoid looking in the registry at all, on non-Windows.


cdmihai:

cdmihai:

PR #3398, opened 2018-06-11
Add GetFileHash and VerifyFileHash tasks

natemcmaster:
Add tasks for computing the file hash and verifying an expected file hash using SHA256, SHA384, or SHA512. These tasks were originally part of aspnet/buildtools, then moved to dotnet/arcade. I've polished them up and added better tests/logging. I'm not set on the API names so let me know if there is a better MSBuild convention for the task and property names.

Resolves https://github.com/Microsoft/msbuild/issues/3369

cc @jeffkl

### DIFF
--- a/src/Tasks/FileIO/GetFileHash.cs
+++ b/src/Tasks/FileIO/GetFileHash.cs
@@ -15,7 +15,9 @@ namespace Microsoft.Build.Tasks
     /// </summary>
     public sealed class GetFileHash : TaskExtension
     {
-        internal const string DefaultFileHashAlgorithm = "SHA256";
+        internal const string _defaultFileHashAlgorithm = "SHA256";
+        internal const string _hashEncodingHex = "hex";
+        internal const string _hashEncodingBase64 = "base64";
 
         /// <summary>
         /// The files to be hashed.
@@ -26,29 +28,23 @@ namespace Microsoft.Build.Tasks
         /// <summary>
         /// The algorithm. Allowed values: SHA256, SHA384, SHA512. Default = SHA256.
         /// </summary>
-        public string Algorithm { get; set; } = DefaultFileHashAlgorithm;
+        public string Algorithm { get; set; } = _defaultFileHashAlgorithm;
 
         /// <summary>
-        /// The metadata name where the hash is store in each item. File hash is in hex. Defaults to "FileHash".
+        /// The metadata name where the hash is stored in each item. Defaults to "FileHash".
         /// </summary>
         public string MetadataName { get; set; } = "FileHash";
 
         /// <summary>
-        /// The metadata name where the base64 encoded hash is store in each item. File hash is in hex. Defaults to "FileHashBase64".
+        /// The encoding to use for generated hashs. Defaults to "hex". Allowed values = "hex", "base64"
         /// </summary>
-        public string MetadataNameBase64 { get; set; } = "FileHashBase64";
+        public string HashEncoding { get; set; } = _hashEncodingHex;
 
         /// <summary>
-        /// The hash of the file in hex. This is only set if there was one item group passed in.
+        /// The hash of the file. This is only set if there was one item group passed in.
         /// </summary>
         [Output]
         public string Hash { get; set; }
-
-        /// <summary>
-        /// The hash of the file base64 encoded. This is only set if there was one item group passed in.
-        /// </summary>
-        [Output]
-        public string HashBase64 { get; set; }
 
         /// <summary>
         /// The input files with additional metadata set to include the file hash.
@@ -64,18 +60,30 @@ namespace Microsoft.Build.Tasks
                 return false;
             }
 
+            if (!TryParseHashEncoding(HashEncoding, out var encoding))
+            {
+                Log.LogErrorFromResources("FileHash.UnrecognizedHashEncoding", HashEncoding);
+                return false;
+            }
+
             foreach (var file in Files)
             {
                 if (!File.Exists(file.ItemSpec))
                 {
                     Log.LogErrorFromResources("FileHash.FileNotFound", file.ItemSpec);
-                    continue;
                 }
+            }
 
+            if (Log.HasLoggedErrors)
+            {
+                return false;
+            }
+
+            foreach (var file in Files)
+            {
                 var hash = ComputeHash(Algorithm, file.ItemSpec);
                 file.SetMetadata("FileHashAlgoritm", Algorithm);
-                file.SetMetadata(MetadataName, ConversionUtilities.ConvertByteArrayToHex(hash));
-                file.SetMetadata(MetadataNameBase64, Convert.ToBase64String(hash));
+                file.SetMetadata(MetadataName, EncodeHash(encoding, hash));
             }
 
             Items = Files;
@@ -83,10 +91,44 @@ namespace Microsoft.Build.Tasks
             if (Files.Length == 1)
             {
                 Hash = Files[0].GetMetadata(MetadataName);
-                HashBase64 = Files[0].GetMetadata(MetadataNameBase64);
             }
 
             return !Log.HasLoggedErrors;
+        }
+
+        internal string EncodeHash(HashEncoding encoding, byte[] hash)
+        {
+            switch (encoding)
+            {
+                case Tasks.HashEncoding.Hex:
+                    return ConversionUtilities.ConvertByteArrayToHex(hash);
+                case Tasks.HashEncoding.Base64:
+                    return Convert.ToBase64String(hash);
+                default:
+                    throw new NotImplementedException();
+            }
+        }
+
+        internal static bool TryParseHashEncoding(string value, out HashEncoding encoding)
+        {
+            encoding = default(HashEncoding);
+
+            if (string.IsNullOrEmpty(value))
+            {
+                return false;
+            }
+
+            switch (value.ToLowerInvariant())
+            {
+                case _hashEncodingHex:
+                    encoding = Tasks.HashEncoding.Hex;
+                    return true;
+                case _hashEncodingBase64:
+                    encoding = Tasks.HashEncoding.Base64;
+                    return true;
+            }
+
+            return false;
         }
 
         internal static bool SupportsAlgorithm(string algorithmName) => _supportedAlgorithms.Contains(algorithmName);
--- a/src/Tasks/FileIO/GetFileHash.cs
+++ b/src/Tasks/FileIO/GetFileHash.cs
@@ -1,0 +1,126 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Security.Cryptography;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Shared;
+
+namespace Microsoft.Build.Tasks
+{
+    /// <summary>
+    /// Computes the checksum for a single file.
+    /// </summary>
+    public sealed class GetFileHash : TaskExtension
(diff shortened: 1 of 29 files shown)


src/Tasks/FileIO/HashHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Security.Cryptography;
+
+namespace Microsoft.Build.Tasks
+{
+    internal class HashHelper
```
jeffkl: I think you should just move `HashHelper`'s stuff to `GetFileHash` and call it from `VerifyFileHash`.

src/Tasks/FileIO/HashHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private static Dictionary<string, Func<HashAlgorithm>> _supportedAlgorithms
+            = new Dictionary<string, Func<HashAlgorithm>>(StringComparer.OrdinalIgnoreCase)
+            {
+                ["SHA256"] = () => new SHA256Managed(),
```
jeffkl: Although I doubt this will be instantiated a lot, I think I'd prefer to see these be some sort of Lazy so that we don't end up with N instances of these classes when you might only be using one.  `Lazy<T>` still allocates, perhaps a factory method with a `switch`?  Is there a benefit of having them as static and would it be better to just instantiate them when needed?

src/Tasks/FileIO/VerifyFileHash.cs
```diff
@@ -0,0 +1,81 @@
...
+                byte[] expectedHash = Convert.FromBase64String(HashBase64);
+                if (!expectedHash.SequenceEqual(hash))
+                {
+                    var actualHash = Convert.ToBase64String(hash);
+                    Log.LogErrorFromResources("VerifyFileHash.HashMismatch", File, Algorithm, HashBase64, actualHash);
+                    return false;
+                }
+            }
+
+            return true;
```
jeffkl: Best practice is to `return !Log.HasLoggedErrors;`

src/Tasks/Resources/Strings.resx
```diff
@@ -2745,6 +2745,27 @@
...
+  <data name="VerifyFileHash.HashMismatch" xml:space="preserve">
+    <value>MSB3952: File hash mismatch. Expected {0} to have a {1} file hash of {2}, but it was {3}.</value>
+    <comment>{StrBegin="MSB3952: "}</comment>
+  </data>
+  <data name="FileHash.UnrecognizedHashAlgorithm" xml:space="preserve">
+    <value>MSB3953: Unrecognized hash algorithm name '{0}'.</value>
+    <comment>{StrBegin="MSB3953: "}</comment>
+  </data>
+  <data name="FileHash.FileNotFound" xml:space="preserve">
+    <value>MSB3954: failed to compute file hash '{0}'. File not found.</value>
```
jeffkl:
```diff
-MSB3954: failed to compute file hash '{0}'. File not found.
+MSB3954: Failed to compute hash for file '{0}' because it does not exist or is inaccessible.
```

natemcmaster:
Updated with PR feedback.

natemcmaster:
> Microsoft.Build.UnitTests.GetSDKReferenceFiles_Tests.GetSDKReferenceFilesTestFixture.VerifyReferencesLoggedX64

Test failure seems unrelated to my change. https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_Full_prtest/597/testReport/junit/Microsoft.Build.UnitTests.GetSDKReferenceFiles_Tests/GetSDKReferenceFilesTestFixture/VerifyReferencesLoggedX64/

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -0,0 +1,126 @@
...
+        [Required]
+        public ITaskItem[] Files { get; set; }
+
+        /// <summary>
+        /// The algorithm. Allowed values: SHA256, SHA384, SHA512. Default = SHA256.
+        /// </summary>
+        public string Algorithm { get; set; } = DefaultFileHashAlgorithm;
+
+        /// <summary>
+        /// The metadata name where the hash is store in each item. File hash is in hex. Defaults to "FileHash".
```
cdmihai: typo: "where the hash is stored in each item"

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -0,0 +1,126 @@
...
+                Log.LogErrorFromResources("FileHash.UnrecognizedHashAlgorithm", Algorithm);
+                return false;
+            }
+
+            foreach (var file in Files)
+            {
+                if (!File.Exists(file.ItemSpec))
+                {
+                    Log.LogErrorFromResources("FileHash.FileNotFound", file.ItemSpec);
+                    continue;
```
cdmihai: Is it worth computing any partial hashes if a file does not exist?

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -0,0 +1,126 @@
...
+                }
+
+                var hash = ComputeHash(Algorithm, file.ItemSpec);
+                file.SetMetadata("FileHashAlgoritm", Algorithm);
+                file.SetMetadata(MetadataName, ConversionUtilities.ConvertByteArrayToHex(hash));
```
cdmihai: Should the user be able to choose which of the two representations to use? I assume that they would only use one of them, so the other becomes wasted memory.

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -0,0 +1,126 @@
...
+                "SHA384",
+                "SHA512",
+            };
+
+        private static HashAlgorithm CreateAlgorithm(string algorithmName)
+        {
+            switch (algorithmName.ToUpperInvariant())
+            {
+                case "SHA256":
+                    return SHA256.Create();
```
cdmihai: Can these objects be reused for the entire execution or do they carry state and need to be created for each ComputeHash call?

jeffkl:
> Are these tasks expensive enough to have them yield and reacquire the engine?

I'd imagine it depends on the size of the file.  I don't think it would be used to hash huge files, but perhaps it could check the file size and then yield?

src/Tasks/FileIO/GetFileHash.cs
jeffkl: Aren't task instances recreated?

cdmihai: Yes, but hash algorithm objects are created for each file to hash. Probably not a problem unless thousands of files get hashed.

natemcmaster: IIUC HashAlgorithm objects carry state so need to be recreated for each file they are hashing.

src/Tasks/FileIO/GetFileHash.cs
natemcmaster: Would it be better if we changed the tasks to support a property such as `UseBase64Encoding="true"` instead? That would allow us to operating in hex or base64 mode and avoid wasting memory.

src/Tasks/FileIO/GetFileHash.cs
natemcmaster: no, good point. We could run a separate loop first to check File.Exists before ever computing a hash.

src/Tasks/FileIO/GetFileHash.cs
cdmihai: Maybe an input property called "HashEncoding" with two valid options, `hex` and `base64`? And then single metadata / output would be either one or the other.

src/Tasks/FileIO/GetFileHash.cs
cdmihai: 👍 
Two loops might regress spinning disks a bit, but hopefully it's an ever decreasing problem. :)

cdmihai:
> Do we already have a good way to run benchmarks?

It's not easy yet to add arbitrary tests. Probably easiest way would be to create github repos where hashing / checking hook up in the design time build target graph. The you could create a couple of repos with increasing file size / number of files hashed.

cdmihai:
But maybe it's only worth doing this when we actually observe slow run times. :)

natemcmaster:
I've updated with PR feedback.

src/Tasks/FileIO/VerifyFileHash.cs
```diff
@@ -54,25 +49,35 @@ public override bool Execute()
...
+            if (!hashesMatch)
             {
-                byte[] expectedHash = Convert.FromBase64String(HashBase64);
-                if (!expectedHash.SequenceEqual(hash))
-                {
-                    var actualHash = Convert.ToBase64String(hash);
-                    Log.LogErrorFromResources("VerifyFileHash.HashMismatch", File, Algorithm, HashBase64, actualHash);
-                    return false;
-                }
+                var actualHash = encoding == Tasks.HashEncoding.Hex
```
CZEMacLeod: This should probably use the same EncodeHash function from GetFileHash. Perhaps it could be an extension method of HashEncoding? It would mean that if another encoding method were added it would be less places to change.

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -64,31 +60,77 @@ public override bool Execute()
...
+                case Tasks.HashEncoding.Hex:
+                    return ConversionUtilities.ConvertByteArrayToHex(hash);
+                case Tasks.HashEncoding.Base64:
+                    return Convert.ToBase64String(hash);
+                default:
+                    throw new NotImplementedException();
+            }
+        }
+
+        internal static bool TryParseHashEncoding(string value, out HashEncoding encoding)
```
CZEMacLeod: Is there a particular reason you are using switch with all the hardcoded text strings here rather than just `return Enum.TryParse<HashEncoding>(value, true, out encoding);`

I am thinking about ensuring that if new encoding methods are added in the future, that it doesn't require changes in lots of places.

src/Tasks/FileIO/VerifyFileHash.cs
natemcmaster: Updated.

src/Tasks/FileIO/GetFileHash.cs
natemcmaster: No particular reason. Enum.TryParse is better so switched to that.

natemcmaster:
Linux failures appear to be unrelated to my changes. Is this a known issue?
```
"/mnt/j/w/Microsoft_msbuild/master/innerloop_Ubuntu16.04_CoreCLR_prtest/artifacts/Debug/bootstrap/netcoreapp2.1/MSBuild/MSBuild.dll". Beginning retry 1 in 1000ms. The process cannot access the file '/mnt/j/w/Microsoft_msbuild/master/innerloop_Ubuntu16.04_CoreCLR_prtest/artifacts/Debug/bin/MSBuild.Bootstrap/publish/MSBuild.dll' because it is being used by another process. [/mnt/j/w/Microsoft_msbuild/master/innerloop_Ubuntu16.04_CoreCLR_prtest/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj]
```

jeffkl:
> Linux failures appear to be unrelated to my changes. Is this a known issue?

I'll take a look if any fail on your latest commit

natemcmaster:
It must have just been flakiness. Re-running Ubuntu passed. But if you want to investigate that failure more, there's a link: https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Ubuntu16.04_CoreCLR_prtest/619/

Issue #53, opened 2015-03-30
MSBuild should handle paths longer than MAX_PATH


ariccio:
As seen in [Fix 260 character file name length limitation](http://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/2156195-fix-260-character-file-name-length-limitation), there's quite a bit of support for better _longer-than-`MAX_PATH`-filename_ handling.

Will Buik's response w/regard to fixing it was:

> We understand that this can be a frustrating issue, however, fixing it requires a large and complicated architectural change across different products and features including Visual Studio, TFS, MSBuild and the .NET Framework. Dedicating resources to this work item would come at the expense of many other features and innovation.

I wondered, how big can those architectural changes be (for MSBuild)?

According to [Naming Files, Paths, and Namespaces: Maximum Path Length Limitation](https://msdn.microsoft.com/en-us/library/aa365247.aspx#maxpath), (which I have mirrored [here](https://github.com/ariccio/altWinDirStat/raw/master/filesystem-docs-n-stuff/Naming%20Files%2C%20Paths%2C%20and%20Namespaces%20%28Windows%29.pdf) ):

> The Windows API has many functions that also have Unicode versions to permit an extended-length path for a maximum total path length of 32,767 characters. This type of path is composed of components separated by backslashes, each up to the value returned in the `lpMaximumComponentLength` parameter of the GetVolumeInformation function (this value is commonly 255 characters). To specify an extended-length path, use the `"\\?\"` prefix. For example, `"\\?\D:\very long path"`.
> _[other stuff]_
> 
> Because you cannot use the `"\\?\"` prefix with a relative path, **relative paths are always limited to a total of MAX_PATH characters**.

_(emphasis mine)_

...Which means that wherever MSBuild uses full paths, we ~~can~~ should be able to just prepend `"\\?\"` to ask for a long path. 

Of course, we'd need to rework the existing path-length-workaround hack. 

This won't fix the whole ecosystem, but it'll get us just one step closer.

I'd like to fix this myself (it seems simple enough), so in accordance with:

> - Contributions must be discussed with the team first, or they will likely be declined. As our process matures and our experience grows, the team expects to take larger contributions.
> - Only contributions referencing an approved Issue will be accepted.

...this is the suggested issue.

I'm a native developer at heart, not an experienced C# developer, but this shouldn't require anything crazy.

Direct references to MAX_PATH:
[...]

PR #3406, opened 2018-06-12
WIP: Improve performance and PathTooLong handling for native directory/path methods

ccastanedaucf:
So far builds long path projects on latest Windows after opting in via group policy and inserting
```
<AppContextSwitchOverrides value="Switch.System.IO.UseLegacyPathHandling=false;Switch.System.IO.BlockLongPaths=false" />
```
in MSBuild.exe.config after building. Need to test other platforms.


src/Shared/FileUtilities.cs
```diff
@@ -251,6 +247,70 @@ internal static String GetDirectoryNameOfFullPath(String fullPath)
...
+        {
+            ErrorUtilities.VerifyThrowArgumentLength(path, "path");
+            string fullPath = GetFullPath(path);
+#if FEATURE_LEGACY_GETFULLPATH
+            if (NativeMethodsShared.IsWindows)
+            {
+                return fullPath;
+            }
+#endif
+            return FixFilePath(fullPath);
```
ccastanedaucf: Preserved the behavior of the original function here, not sure I understand why FixFilePath is specifically not called on net46 Windows, yet called on mono and netcore Windows

src/Shared/FileUtilities.cs
```diff
@@ -251,6 +247,70 @@ internal static String GetDirectoryNameOfFullPath(String fullPath)
...
+        }
+
+#if !CLR2COMPATIBILITY
+        internal static string NormalizePath(params string[] paths)
+        {
+            return NormalizePath(Path.Combine(paths));
+        }
+#endif
+
+        private unsafe static string GetFullPath(string path)
```
ccastanedaucf: After factoring this out of NormalizePath(), seems like it belongs in NativeMethodsShared since functionally it's very similar to GetCurrentDirectory() in that it's a wrapper around an external call

AndyGerlicher:
Be sure to mention #53 in your commit message to get the association.

src/MSBuildTaskHost/MSBuildTaskHost.csproj
```diff
@@ -205,4 +205,8 @@
     <PackageReference Include="SourceLink.Create.CommandLine" Version="2.1.2" />
     <PackageReference Include="LargeAddressAware" Version="$(LargeAddressAwareVersion)" PrivateAssets="All" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
+    <!-- Need Win32 API on .NET Core to ping registry to determine long path support -->
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.3.0" Condition="'$(OsEnvironment)'=='windows'" />
```
rainersigwald: Actually, looks like this is in enough places to just reference it everywhere by putting it in [`src/Directory.Build.props`](https://github.com/Microsoft/msbuild/blob/7f95dc1575d19b9fac4da14b6f754aefe9d73a12/src/Directory.Build.props).

src/Shared/FileUtilities.cs
rainersigwald: Sounds reasonable to me

src/Shared/FileUtilities.cs
```diff
@@ -251,6 +253,75 @@ internal static String GetDirectoryNameOfFullPath(String fullPath)
...
```
rainersigwald: Can you scope the `unsafe` to a block within the method just to reduce the surface area?

src/Shared/FileUtilities.cs
```diff
@@ -276,140 +347,40 @@ private unsafe static bool AreStringsEqual(char* buffer, int len, string s)
...
-        /// Gets the canonicalized full path of the provided path.
-        /// Path.GetFullPath The pre .Net 4.6.2 implementation of Path.GetFullPath is slow and creates strings in its work.
-        /// Therefore MSBuild has its own implementation on full framework.
-        /// Guidance for use: call this on all paths accepted through public entry
-        /// points that need normalization. After that point, only verify the path
-        /// is rooted, using ErrorUtilities.VerifyThrowPathRooted.
-        /// ASSUMES INPUT IS ALREADY UNESCAPED.
-        /// </summary>
-        internal static string NormalizePath(string path)
+        private static bool IsUNCPath(string path)
```
rainersigwald: Does this need to do anything on non-Windows? Can we just short-circuit it there?

src/Shared/FileUtilities.cs
```diff
@@ -893,33 +799,29 @@ internal static bool FileExistsNoThrow(string fullPath)
...
-                }
-                else
-                {
-#endif
-                    return NativeMethodsShared.FileExists(fullPath);
-#if !CLR2COMPATIBILITY
-                }
+            if (NativeMethodsShared.IsWindows && Traits.Instance.CacheFileExistence)
+            {
+                fullPath = AttemptToShortenPath(fullPath);
```
rainersigwald: Is this necessary if `GetFileAttributes()` always does it too?

src/Shared/NativeMethodsShared.cs
```diff
@@ -44,9 +45,14 @@ internal static class NativeMethodsShared
         internal const int FILE_ATTRIBUTE_DIRECTORY = 0x00000010;
         internal const int FILE_ATTRIBUTE_REPARSE_POINT = 0x00000400;
 
+        private const int MAX_PATH_LEGACY_WINDOWS = 260;
```
rainersigwald: I'd leave this named `MAX_PATH` for uniformity. We can use it in a way that makes it optional.

src/Shared/NativeMethodsShared.cs
```diff
@@ -1181,19 +1236,37 @@ internal static int GetParentProcessId(int processId)
...
+                char* buffer = stackalloc char[bufferSize];
+                int pathLength = GetCurrentDirectoryAndAssertSuccess(bufferSize, buffer);
+                return new string(buffer, startIndex: 0, length: pathLength);
             }
-
+#endif
             return Directory.GetCurrentDirectory();
         }
 
+        private unsafe static int GetCurrentDirectoryAndAssertSuccess(int nBufferLength, char* lpBuffer)
```
rainersigwald: I don't know if you need to explicitly say `AndAssertSuccess` in the name, since it's the default .NET error-handling strategy to throw on failure when in a method that doesn't start with `Try...`.

src/Shared/NativeMethodsShared.cs
```diff
@@ -1181,19 +1236,37 @@ internal static int GetParentProcessId(int processId)
...
+        {
+            int pathLength = GetCurrentDirectory(nBufferLength, lpBuffer);
+            AssertNativeMethodSuccess(pathLength);
+            return pathLength;
+        }
+
+        internal static void AssertNativeMethodSuccess(int result)
```
rainersigwald: Following our [ErrorUtilities convention](https://github.com/Microsoft/msbuild/blob/7f95dc1575d19b9fac4da14b6f754aefe9d73a12/src/Shared/ErrorUtilities.cs#L617), I'd probably call this `VerifyThrowWin32Result(int result)`

src/Shared/NativeMethodsShared.cs
```diff
@@ -460,6 +473,48 @@ public SystemInformationData()
         /// </remarks>
         internal static int MAX_PATH = 260;
 
+        /// <summary>
+        /// Gets an enum for the max path restrictions of the current OS.
+        /// </summary>
+        internal static MaxPathLimits OSMaxPathLimit
```
rainersigwald: Does this being in `shared` mean that we'll actually do this registry call once per assembly? That seems suboptimal.

src/Build.OM.UnitTests/Microsoft.Build.Engine.OM.UnitTests.csproj
```diff
@@ -14,6 +14,11 @@
     <PackageReference Include="Shouldly" Version="$(ShouldlyVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
+    <!-- Need Win32 API on .NET Core to ping registry to determine long path support -->
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.3.0" Condition="'$(OsEnvironment)'=='windows'" />
```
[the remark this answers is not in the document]
ccastanedaucf: Just added that change since the runtime checks were already there and this caused non-Windows builds to fail

src/Shared/NativeMethodsShared.cs
ccastanedaucf: I was wondering if this belonged here since the runtime OS checks seem to all be here. Is there a better spot to put this?

src/Shared/NativeMethodsShared.cs
ccastanedaucf: Right now it would conflict with the static int MAX_PATH already declared which can't be used as a const for the enum. But I think it would be safe to just change that to a const. It has a comment saying it's intended for unit tests to change the value, but no unit tests actually change it,.

src/Shared/FileUtilities.cs
ccastanedaucf: I can extract that piece outside the method. Seems like if I block scope it then variable scoping ends up making the unsafe block still cover most of the method.

src/Shared/FileUtilities.cs
ccastanedaucf: I assumed ```FileExistenceCache``` was also being called elsewhere but looks like this is the only entry point. ```AttemptToShortenPath()``` also sneaks in ```FixFilePath()``` so it could effect the ```GetOrAdd()``` key if one caller fixes the file path and another doesn't, but so far I don't see that problem.

src/Deprecated/Engine/Microsoft.Build.Engine.csproj
```diff
@@ -18,10 +18,14 @@
   
   <ItemGroup>
     <Reference Include="System.Configuration" />
-
     <ProjectReference Include="..\..\Framework\Microsoft.Build.Framework.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
+    <!-- Need Win32 API on .NET Core to ping registry to determine long path support -->
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.3.0" Condition="'$(OsEnvironment)'=='windows'" />
```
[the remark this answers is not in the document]
ccastanedaucf: Don't think so, didn't see this was in the deprecated folder

src/Shared/FileUtilities.cs
ccastanedaucf: Nope, so far it's only being used under the ```isWindows``` check above but that probably belongs here too.

src/Shared/NativeMethodsShared.cs
```diff
@@ -460,6 +473,48 @@ public SystemInformationData()
...
+        }
+
+        private static bool IsMaxPathLimitLegacyWindows()
+        {
+            if (!IsWindows)
+            {
+                return false;
+            }
+            RegistryKey fileSystemKey = Registry.LocalMachine.OpenSubKey(WINDOWS_FILE_SYSTEM_REGISTRY_KEY);
+            object longPathsEnabledValue = fileSystemKey.GetValue(WINDOWS_LONG_PATHS_ENABLED_VALUE_NAME, 0);
```
ccastanedaucf: Still unsure how to prevent doing the registry call per assembly

AndyGerlicher:
@rainersigwald tried closing and reopening... Should we have seen the WIP bot?

rainersigwald:
@AndyGerlicher I expected that too, but it's appearing on new PRs so 🤷‍♂️

src/Shared/NativeMethodsShared.cs
cdmihai: I guess we all know who's MAX_PATH this is, so no point calling it `WindowsLegacyMaxPath` :p

ccastanedaucf:
Closing since this PR was split between #3442 , #3443, #3454, and #3507 ages ago

PR #3414, opened 2018-06-14
Avoid allocations when deserializing build event args

jeffkl:
Use our singleton `BuildEventContext.Invalid` instead.


jeffkl:

PR #3419, opened 2018-06-18
shorten sgen reference parameter line

GromHSCR:
shorten sgen reference parameter line to avoid too long string on GenerateSerializationAssemblies

### DIFF
--- a/src/Tasks/SGen.cs
+++ b/src/Tasks/SGen.cs
@@ -253,9 +253,10 @@ namespace Microsoft.Build.Tasks
                 // Append the references, if any.
                 if (References != null)
                 {
-                    foreach (string reference in References)
+                    if (References.Length > 0)
                     {
-                        commandLineBuilder.AppendSwitchIfNotNull("/reference:", reference);
+                        //Appending references if 
+                        commandLineBuilder.AppendSwitchIfNotNull("/r:", string.Join(",", References));
                     }
                 }
 
--- a/src/Tasks/SGen.cs
+++ b/src/Tasks/SGen.cs
@@ -255,7 +255,10 @@ namespace Microsoft.Build.Tasks
                 {
                     if (References.Length > 0)
                     {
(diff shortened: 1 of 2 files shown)


GromHSCR:
Met some problem with GenerateSerializationAssemblies, when command line is too long and exceed 32000 symbols. With max symbols need research but I can shorten references line by split reference lines by "," as described in SGEN.exe, and replace /reference with /r and it works. For me it shortened 34500 symbols line with 30100 line.

src/Tasks/SGen.cs
```diff
@@ -253,9 +253,10 @@ protected override string GenerateCommandLineCommands()
                 // Append the references, if any.
                 if (References != null)
                 {
-                    foreach (string reference in References)
+                    if (References.Length > 0)
                     {
-                        commandLineBuilder.AppendSwitchIfNotNull("/reference:", reference);
+                        //Appending references if
```
rainersigwald: Please complete comment

src/Tasks/SGen.cs
```diff
@@ -253,9 +253,10 @@ protected override string GenerateCommandLineCommands()
                 // Append the references, if any.
                 if (References != null)
                 {
-                    foreach (string reference in References)
+                    if (References.Length > 0)
                     {
-                        commandLineBuilder.AppendSwitchIfNotNull("/reference:", reference);
+                        //Appending references if 
+                        commandLineBuilder.AppendSwitchIfNotNull("/r:", string.Join(",", References));
```
rainersigwald: I would prefer to avoid shortening `/reference:`; it's generally good style to use long-form command line arguments for readability and maintainability, and the real win here is not having to repeat it per `Reference`.

GromHSCR: I can leave full "/reference". Should I change this with the new commit within this request or create a new branch with one commit and recreate request?

Issue #3320, opened 2018-05-16
Perf, exception contents, and style: Use newer C# features; run code refactoring


erikma:
Discussed offline with members of the MSBuild team: This item covers several PRs to be sent to update MSBuild code style using ReSharper. R# also finds performance-affecting cases like setting methods to static and specifying readonly on variables only set by a constructor. While using it on the codebase it also pointed out some actual code bugs that I'll open separate items for.

There are some widespread naming bugs for method parameters where a string is used for a variable name in e.g. ArgumentNullException but has drifted away from changes to the underlying variable name. Use nameof() to fix these cases and improve exception contents.

PR #3421, opened 2018-06-18
More automated code refactoring and .NET 1 data structure reduction

erikma:
src\Build\ - partial tree
For #3320


src/Build.OM.UnitTests/Definition/ProjectCollection_Tests.cs
```diff
@@ -1269,7 +1266,6 @@ public void ProjectCollectionChangedEvent2()
             project.SetProperty("a", "1");
             collection.SetGlobalProperty("a", "1");
             VerifyProjectCollectionEvents(collection, true, "2");
-            project.GetPropertyValue("a").Equals("2", StringComparison.OrdinalIgnoreCase);
```
AndyGerlicher: Rather than delete this change to `project.GetPropertyValue("a").ShouldBe("2")`? I suppose it could be testing that it didn't throw, but more likely a mstest -> nunit -> xunit conversion issue.

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
```diff
@@ -1326,15 +1292,14 @@ private void IssueBuildRequest(BuildRequestBlocker blocker)
         /// <param name="isLastTask"><code>true</code> if this is the last task for this queue, otherwise <code>false</code>.</param>
         /// <returns>True if the task was scheduled, false otherwise.</returns>
         /// <remarks>This method will return false if an attempt is made to schedule an action after the queue has been shut down.</remarks>
-        private bool QueueAction(Action action, bool isLastTask)
```
AndyGerlicher: I assume this was because the value was never used? Should delete the xml comment as well please.

src/Build.OM.UnitTests/Definition/ProjectCollection_Tests.cs
erikma: I found after enabling this and the others like it in Assert.IsTrue() lines that they were in fact wrong - because there is no full path on the project XML provided to the ProjCollection, the project is not added to the collection, and so changes to the collection global properties were not propagating. After debugging for awhile I figured I'd dump the lines instead of asserting false, since they're useless.

LMK if you disagree, I can restore under a false assertion.

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
erikma: Correct on unused. Fixed in next push.

src/Build.OM.UnitTests/Definition/ProjectCollection_Tests.cs
rainersigwald: That sounds concerning. Should we fix the full-path thing instead?

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEntry.cs
```diff
@@ -171,68 +144,33 @@ internal BuildRequestEntry(BuildRequest request, BuildRequestConfiguration reque
...
         /// <summary>
         /// Returns the current state of the build request.
         /// </summary>
-        public BuildRequestEntryState State
-        {
-            [DebuggerStepThrough]
-            get
-            { return _state; }
-        }
+        public BuildRequestEntryState State { [DebuggerStepThrough] get; private set; }
```
rainersigwald: Does `[DebuggerStepThrough]` work for autoprops? I thought it would be unnecessary.

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -264,33 +254,19 @@ private BuildRequestConfiguration(INodePacketTranslator translator)
...
         /// Flag indicating whether or not the configuration is actually building.
         /// </summary>
-        public bool IsActivelyBuilding
-        {
-            get
-            {
-                return (_activelyBuildingTargets != null) && (_activelyBuildingTargets.Count > 0);
-            }
-        }
+        public bool IsActivelyBuilding => (_activelyBuildingTargets != null) && (_activelyBuildingTargets.Count > 0);
```
rainersigwald: Further simplification to

```c#
_activelyBuildingTargets?.Count > 0
```

?

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -557,61 +526,35 @@ public Lookup BaseLookup
...
                 ErrorUtilities.VerifyThrow(!IsCached, "Configuration is cached, we shouldn't be accessing the lookup.");
 
-                if (null == _baseLookup)
-                {
-                    _baseLookup = new Lookup(Project.ItemsToBuildWith, Project.PropertiesToBuildWith);
-                }
-
-                return _baseLookup;
+                return _baseLookup ??
+                       (_baseLookup = new Lookup(Project.ItemsToBuildWith, Project.PropertiesToBuildWith));
```
rainersigwald: Does this pattern produce better codegen? I think the old way is more readable.

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
```diff
@@ -452,13 +389,13 @@ bool ITaskExecutionHost.SetTaskParameters(IDictionary<string, Tuple<string, Elem
         /// <returns>True of the outputs were gathered successfully, false otherwise.</returns>
         bool ITaskExecutionHost.GatherTaskOutputs(string parameterName, ElementLocation parameterLocation, bool outputTargetIsItem, string outputTargetName)
         {
-            ErrorUtilities.VerifyThrow(_taskFactoryWrapper != null, "Need a taskFactoryWrapper to retrieve outputs from.");
+            ErrorUtilities.VerifyThrow(_UNITTESTONLY_TaskFactoryWrapper != null, "Need a taskFactoryWrapper to retrieve outputs from.");
```
rainersigwald: Going to an autoprop makes this look very wrong here. It's _used_ everywhere, but should be _set_ only in unit tests. I think I'd rather see this reverted.

src/Build/Collections/ItemDictionary.cs
```diff
@@ -407,32 +396,14 @@ internal Enumerator(IEnumerable<ICollection<T>> listEnumerable)
             /// <summary>
             /// Get the current item
             /// </summary>
-            public T Current
-            {
-                get
-                {
-                    // Undefined if enumerator is before or after collection: we return null
```
rainersigwald: Preserve comment in a `/// <remarks>` ?

src/Build/Collections/RetrievableEntryHashSet/HashSet.cs
```diff
@@ -78,12 +77,12 @@ namespace Microsoft.Build.Collections
     /// the same time. 
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    [DebuggerTypeProxy(typeof(Microsoft.Build.Collections.HashSetDebugView<>))]
+    [DebuggerTypeProxy(typeof(HashSetDebugView<>))]
```
rainersigwald: Please revert this file. It's copy-🍝, and these changes will make it hard to pull improvements to the base type in here.

src/Build/Collections/Triple.cs
```diff
@@ -18,53 +15,30 @@ namespace Microsoft.Build.Collections
     /// <typeparam name="C">Type of third object</typeparam>
     internal struct Triple<A, B, C>
```
rainersigwald: This type appears unused. Delete it entirely?

src/Build/Collections/WeakReference.cs
```diff
@@ -39,7 +34,7 @@ internal class WeakReference<T>
         /// </summary>
         internal WeakReference(T target)
```
rainersigwald: There's been a [Framework `WeakReference<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.weakreference-1?view=netframework-4.7.1#applies-to) since 4.5. Can we switch to that and just delete this?

src/Build/Construction/ProjectSdkElement.cs
```diff
@@ -1,4 +1,11 @@
-﻿using Microsoft.Build.Internal;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//-----------------------------------------------------------------------
+// </copyright>
+// <summary>Definition of ProjectSdkElement class.</summary>
+//-----------------------------------------------------------------------
```
rainersigwald: Can you get rid of

```diff
+// </copyright>
+// <summary>Definition of ProjectSdkElement class.</summary>
+//-----------------------------------------------------------------------
```

?

src/Build/Definition/ProjectCollection.cs
```diff
@@ -881,20 +861,8 @@ internal int MaxNodeCount
...
-                return _projectRootElementCache;
-            }
-
-            private set
-            {
-                // no locks required because this field is only set in the constructor.
-                _projectRootElementCache = value;
-            }
-        }
+        // No locks required because this field is only set in the constructor.
```
rainersigwald: Since it's an autoprop with no setter, you can delete this comment

src/Build.OM.UnitTests/Definition/ProjectCollection_Tests.cs
erikma: The doc comment on ProjectCollection.OnAfterRenameLoadedProject() says no:

        /// Stores the project in the list of loaded projects if it has a name.
        /// Does not store the project if it has no name because it has not been saved to disk yet.
        /// If the project previously had a name, but was not in the collection already, throws InvalidOperationException.
        /// If the project was not previously in the collection, sets the collection's global properties on it.

src/Build.OM.UnitTests/Definition/ProjectCollection_Tests.cs
```diff
@@ -353,23 +369,19 @@ public void ErrorTwoProjectsEquivalentOneCollection()
...
             Assert.Throws<InvalidOperationException>(() =>
             {
-                Project project = new Project();
-                project.FullPath = "c:\\x";
-
-                Project project2 = new Project();
-                project2.FullPath = "c:\\x";
-            }
-           );
+                _ = new Project { FullPath = "c:\\x" };
```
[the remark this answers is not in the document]
erikma: Done in next push

src/Build.OM.UnitTests/Definition/ProjectCollection_Tests.cs
```diff
@@ -569,9 +576,9 @@ public void ChangingGlobalPropertiesUpdatesCollection()
         [Fact]
         public void SettingGlobalPropertiesOnCollectionUpdatesProjects()
         {
-            ProjectCollection collection = new ProjectCollection();
-            Project project1 = new Project(collection);
-            project1.FullPath = "c:\\y"; // load into collection
+            var collection = new ProjectCollection();
+            var project1 = new Project(collection) { FullPath = "c:\\y" };
+            // load into collection
```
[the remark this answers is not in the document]
erikma: Removed

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
dfederm: That reduces to `int? > 0` which I don't think works. You'd need to do something like `(_activelyBuildingTargets?.Count ?? 0) > 0`, but IMHO that's worse than what's there now.

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEntry.cs
erikma: Didn't find a strong opinion when looking around for this. I'll proceed to remove these from the auto props - some here and more in later PRs.

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
rainersigwald: `int? {comparison} int` is [well-defined](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types#operators): it returns `false`. So I still think I want this change.

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
erikma: R# likes one-liners (well, 2 in this case) over 4-liners. Reverted.

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
erikma: You're right, I thought I reverted this already. Fixed.

src/Build/Collections/ItemDictionary.cs
erikma: Done

src/Build/Collections/RetrievableEntryHashSet/HashSet.cs
erikma: Point. Need to fix the original source then pull back. :) Kept one line of change (removal) pasted below since it's a perf win in a critical path:

 _slots[i].hashCode = hashCode;

src/Build/Collections/Triple.cs
erikma: Good catch

src/Build/Construction/ProjectSdkElement.cs
erikma: done

src/Build/Definition/ProjectCollection.cs
erikma: Done

src/Build/Collections/WeakReference.cs
erikma: Removed. Audited all WeakReference usage in the codebase. OpportunisticIntern.cs could really benefit from WeakReference<T> but because it's in Shared it can't use it until the last of the .NET 3.5 code back compat is removed. I did update OpportunisticIntern.cs in the usual ways; please review after the next push.

src/Build/Collections/WeakValueDictionary.cs
```diff
@@ -175,15 +174,13 @@ public bool TryGetValue(K key, out V value)
...
-            value = wrappedValue.Target;
-
-            if (value == null)
+            if (!wrappedValue.TryGetTarget(out value))
             {
                 _dictionary.Remove(key);
                 return false;
             }
 
-            return result;
```
rainersigwald: Could get rid of `result` entirely, right?

```c#
             if (!_dictionary.TryGetValue(key, out WeakReference<V> wrappedValue))
             {
                 value = null;
                 return false;
             }
```

erikma: Needed near the top of the method as a return from a TryGetValue()

erikma: Oh wait duh I get it. Done in next push

src/Build/BackEnd/Shared/BuildResult.cs
```diff
@@ -496,7 +475,7 @@ internal bool ResultBelongsToRootRequest
         /// <returns>The results for the specified target</returns>
         /// <exception>KeyNotFoundException is returned if the specified target doesn't exist when reading this property.</exception>
         /// <exception>ArgumentException is returned if the specified target already has results.</exception>
-        public ITargetResult this[string target]
+        public TargetResult this[string target]
```
uniqueiniquity: Was this intended to be a breaking change? This seems to have broken our usage of the indexer. @AndyGerlicher

PR #3431, opened 2018-06-20
[mono] Fix Release-MONO configuration mappings in MSBuild.sln

radical:
`Release-MONO` configurations were incorrectly mapped to `Release`,
which breaks mono build.


radical:
This should be harmless, but I didn't try to load it in VS!

PR #3437, opened 2018-06-21
Fix issue where a previously used isolated context gets reused again

cdmihai:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/635449

### DIFF
--- a/src/Build/Evaluation/Context/EvaluationContext.cs
+++ b/src/Build/Evaluation/Context/EvaluationContext.cs
@@ -58,6 +58,14 @@ namespace Microsoft.Build.Evaluation.Context
             return context;
         }
 
+        private EvaluationContext CreateUsedIsolatedContext()
+        {
+            var context = Create(SharingPolicy.Isolated);
+            context._used = 1;
+
+            return context;
+        }
+
         internal EvaluationContext ContextForNewProject()
         {
             // Projects using isolated contexts need to get a new context instance 
@@ -66,11 +74,11 @@ namespace Microsoft.Build.Evaluation.Context
                 case SharingPolicy.Shared:
                     return this;
                 case SharingPolicy.Isolated:
-                    // use the first isolated context
-                    var used = Interlocked.CompareExchange(ref _used, 1, 0);
-                    return used == 0
+                    // reuse the first isolated context, as it has not seen an evaluation yet.
+                    var previousValueWasUsed = Interlocked.CompareExchange(ref _used, 1, 0);
+                    return previousValueWasUsed == 0
                         ? this
-                        : Create(Policy);
+                        : CreateUsedIsolatedContext();
                 default:
                     ErrorUtilities.ThrowInternalErrorUnreachable();
                     return null;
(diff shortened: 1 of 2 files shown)


cdmihai:

AndyGerlicher:

AndyGerlicher:
Few test failures, don't seem related: https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_Full_prtest/640/

src/Build/Evaluation/Context/EvaluationContext.cs
```diff
@@ -66,11 +74,11 @@ internal EvaluationContext ContextForNewProject()
                 case SharingPolicy.Shared:
                     return this;
                 case SharingPolicy.Isolated:
-                    // use the first isolated context
-                    var used = Interlocked.CompareExchange(ref _used, 1, 0);
-                    return used == 0
+                    // reuse the first isolated context, as it has not seen an evaluation yet.
```
rainersigwald: _if_ it has not seen an evaluation yet?

src/Build/Evaluation/Context/EvaluationContext.cs
```diff
@@ -66,11 +74,11 @@ internal EvaluationContext ContextForNewProject()
                 case SharingPolicy.Shared:
                     return this;
                 case SharingPolicy.Isolated:
-                    // use the first isolated context
-                    var used = Interlocked.CompareExchange(ref _used, 1, 0);
-                    return used == 0
+                    // reuse the first isolated context if it has not seen an evaluation yet.
+                    var previousValueWasUsed = Interlocked.CompareExchange(ref _used, 1, 0);
+                    return previousValueWasUsed == 0
```
adrianvmsft: Shouldn't this condition be flipped?

cdmihai: The problem actually was that the newly spawned isolated instance forgot that its coming from a line that has been previously used. This lead to every other isolated context to get erroneously reused.

Intended behaviour
1st isolated instance, first evaluation. Reuse the instance because it has not seen any instance
1st isolated instance, second evaluation. Spawn and use 2nd isolated instance
2nd isolated instance. Spawn and use 3rd isolated instance

Buggy behaviour
1st isolated instance, first evaluation. Reuse the instance because it has not seen any instance
1st isolated instance, second evaluation. Spawn and use 2nd isolated instance. But this instance forgot that it needs respawning.
2nd isolated instance. Since it forgot that it needs respawning, it gets reused

src/Build/Evaluation/Context/EvaluationContext.cs
```diff
@@ -58,6 +58,14 @@ public static EvaluationContext Create(SharingPolicy policy)
             return context;
         }
 
+        private EvaluationContext CreateUsedIsolatedContext()
+        {
+            var context = Create(SharingPolicy.Isolated);
+            context._used = 1;
```
adrianvmsft: nit: this may not be needed as it is already handled by the Interlocked.CompareExchange below.
- if previous value was 0, it gets switched to 1 (and method doesn't even get called)
- if previous value was 1, it stays as 1 - that's also when CreatedIsolatedContext gets called

adrianvmsft: Nevermind, I didn't realize that is on a new object.

PR #3441, opened 2018-06-21
Use Central Package Versioning SDK

AndyGerlicher:
Uses SDK from https://github.com/Microsoft/MSBuildSdks

Versions for `PackageReference` can only be specified in `build/Packages.props` file. If you add a `Version=` attribute to any other `PackageReference` in our tree you get an error that you must specify it in the Packages.props file (it gives the path to it in the error message).

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="roslyn-tools" value="https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/build/Packages.props
+++ b/build/Packages.props
@@ -1,0 +1,68 @@
+<Project>
+
+  <PropertyGroup>
+      <!-- Prodcon-overrideable, using repo-toolset naming convention
+      <MicrosoftNetCompilersVersion Condition="'$(MicrosoftNetCompilersVersion)' == ''">2.8.2</MicrosoftNetCompilersVersion>
+      <MicrosoftNETCoreCompilersVersion Condition="'$(MicrosoftNETCoreCompilersVersion)' == ''">2.8.0</MicrosoftNETCoreCompilersVersion>-->
+
+      <!--<NuGetPackageVersion Condition="'$(NuGetPackageVersion)' == ''">4.8.0-preview1.5185</NuGetPackageVersion>
+          <NuGetBuildTasksVersion Condition="'$(NuGetBuildTasksVersion)' == ''">$(NuGetPackageVersion)</NuGetBuildTasksVersion>
+          <NuGetCommandsVersion Condition="'$(NuGetCommandsVersion)' == ''">$(NuGetPackageVersion)</NuGetCommandsVersion>
+          <NuGetProtocolVersion Condition="'$(NuGetProtocolVersion)' == ''">$(NuGetPackageVersion)</NuGetProtocolVersion>
+      -->
+      <NuGetPackageVersion>4.8.0-preview1.5185</NuGetPackageVersion>
+      <NuGetBuildTasksVersion>$(NuGetPackageVersion)</NuGetBuildTasksVersion>
+      <NuGetCommandsVersion>$(NuGetPackageVersion)</NuGetCommandsVersion>
+      <NuGetProtocolVersion>$(NuGetPackageVersion)</NuGetProtocolVersion>
+      
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Update="LargeAddressAware" Version="1.0.3" />
+    <PackageReference Update="Microsoft.Build.NuGetSdkResolver" Version="$(NuGetBuildTasksVersion)" />
+    <PackageReference Update="Microsoft.CodeAnalysis.Build.Tasks" Version="3.0.0-beta1-61516-01" />
+    <PackageReference Update="Microsoft.DotNet.BuildTools.GenAPI" Version="2.1.0-prerelease-02404-02" />
+    <PackageReference Update="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
+    <PackageReference Update="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" />
+    <PackageReference Update="Microsoft.NETCore.Compilers" Version="$(MicrosoftNetCoreCompilersVersion)" />
+    <PackageReference Update="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.15" />
+    <PackageReference Update="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="1.16.30" />
+    <PackageReference Update="Microsoft.Win32.Registry" Version="4.3.0" />
+    <PackageReference Update="NuGet.Build.Tasks" Version="$(NuGetBuildTasksVersion)" />
+    <PackageReference Update="PdbGit" Version="3.0.41" />
+    <PackageReference Update="Shouldly" Version="3.0.0" />
+    <PackageReference Update="SourceLink.Create.CommandLine" Version="2.1.2" />
+    <PackageReference Update="System.CodeDom" Version="4.4.0" />
+    <PackageReference Update="System.Collections.Concurrent" Version="4.3.0" />
+    <PackageReference Update="System.Collections.Immutable" Version="1.5.0" />
+    <PackageReference Update="System.Diagnostics.TraceSource" Version="4.0.0" />
+    <PackageReference Update="System.IO.Compression" Version="4.3.0" />
+    <PackageReference Update="System.IO.FileSystem.Primitives" Version="4.3.0" />
+    <PackageReference Update="System.Linq.Parallel" Version="4.0.1" />
+    <PackageReference Update="System.Net.Http" Version="4.3.0" />
+    <PackageReference Update="System.Reflection.Metadata" Version="1.6.0" />
+    <PackageReference Update="System.Reflection.TypeExtensions" Version="4.1.0" />
+    <PackageReference Update="System.Resources.Writer" Version="4.0.0" />
+    <PackageReference Update="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+    <PackageReference Update="System.Runtime.Loader" Version="4.0.0" />
+    <PackageReference Update="System.Runtime.Serialization.Primitives" Version="4.1.1" />
+    <PackageReference Update="System.Security.Principal.Windows" Version="4.3.0" />
+    <PackageReference Update="System.Text.Encoding.CodePages" Version="4.0.1" />
+    
+    <PackageReference Update="System.Threading.Tasks.Dataflow" Version="4.5.24.0" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'" />
+    <PackageReference Update="System.Threading.Tasks.Dataflow" Version="4.6.0" Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'"/>
+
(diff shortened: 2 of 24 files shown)


AndyGerlicher:
FYI @cdmihai and I verified all references are identical before and after the build. Dataflow was the only one that slightly differed between Core and Full Framework we had to add a Condition for in the Packages.props.

build/Packages.props
```diff
@@ -0,0 +1,61 @@
...
+    <!-- Prodcon-overrideable, using repo-toolset naming convention -->
+    <NuGetBuildTasksVersion>$(NuGetPackageVersion)</NuGetBuildTasksVersion>
+    <NuGetCommandsVersion>$(NuGetPackageVersion)</NuGetCommandsVersion>
+    <NuGetProtocolVersion>$(NuGetPackageVersion)</NuGetProtocolVersion>
+</PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Update="LargeAddressAware" Version="1.0.3" />
+    <PackageReference Update="Microsoft.Build.Build" Version="15.5.180" />
+    <PackageReference Update="Microsoft.Build.Framework" Version="15.5.180" />
```
rainersigwald: I don't think we should centrally control these; I think we should let the samples directory free.

src/Directory.Build.props
```diff
@@ -1,4 +1,10 @@
 ﻿<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <!-- Add IsImplicitlyDefined to any packages that were 'magically' referenced (RepoTools!). This
+         should be removed if RepoTools adds the metadata for their package references. -->
+    <PackageReference Update="@(PackageReference)" IsImplicitlyDefined="true" />
```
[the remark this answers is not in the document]
tmat: What does `IsImplicitlyDefined` mean?

build/Packages.props
```diff
@@ -0,0 +1,61 @@
...
+    <PackageReference Update="System.Threading.Thread" Version="4.0.0" />
+    <PackageReference Update="System.Xml.XPath" Version="4.3.0" />
+    <PackageReference Update="xunit.assert" Version="$(XUnitVersion)" />
+    <PackageReference Update="xunit.console" Version="$(XUnitVersion)" />
+    <PackageReference Update="xunit.core" Version="$(XUnitVersion)" />    
+  </ItemGroup>
+
+  <ItemGroup>
+    <GlobalPackageReference Include="Nerdbank.GitVersioning" Version="2.1.23" Condition="'$(DisableNerdbankVersioning)' != 'true'" />
+    <GlobalPackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0-alpha-003" PrivateAssets="all"/>
```
tmat: What's the difference between `PackageReference` and `GlobalPackageReference`?

build/Packages.props
```diff
@@ -0,0 +1,61 @@
+<Project>
+
+<PropertyGroup>
```
tmat: Nit: Indentation is off.

build/Packages.props
```diff
@@ -0,0 +1,61 @@
...
+    <PackageReference Update="LargeAddressAware" Version="1.0.3" />
+    <PackageReference Update="Microsoft.Build.Build" Version="15.5.180" />
+    <PackageReference Update="Microsoft.Build.Framework" Version="15.5.180" />
+    <PackageReference Update="Microsoft.Build.NuGetSdkResolver" Version="$(NuGetBuildTasksVersion)" />
+    <PackageReference Update="Microsoft.Build.Utilities.Core" Version="15.5.180" />
+    <PackageReference Update="Microsoft.CodeAnalysis.Build.Tasks" Version="3.0.0-beta1-61516-01" />
+    <PackageReference Update="Microsoft.DotNet.BuildTools.GenAPI" Version="2.1.0-prerelease-02404-02" />
+    <PackageReference Update="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
+    <PackageReference Update="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" />
+    <PackageReference Update="Microsoft.NETCore.Compilers" Version="$(MicrosoftNetCoreCompilersVersion)" />
```
tmat: Do you use Microsoft.Net.Compiler and Microsoft.NETCore.Compilers directly in the repo? The toolset adds these automatically.

build/Packages.props
```diff
@@ -0,0 +1,61 @@
...
+<PropertyGroup>
+    <NuGetPackageVersion>4.8.0-preview1.5185</NuGetPackageVersion>
+
+    <!-- Prodcon-overrideable, using repo-toolset naming convention -->
+    <NuGetBuildTasksVersion>$(NuGetPackageVersion)</NuGetBuildTasksVersion>
+    <NuGetCommandsVersion>$(NuGetPackageVersion)</NuGetCommandsVersion>
+    <NuGetProtocolVersion>$(NuGetPackageVersion)</NuGetProtocolVersion>
+</PropertyGroup>
+
+  <ItemGroup>
```
tmat: Specifying versions explicitly, not using `-Version` variables will break Orchestrated Build dependency uptake. Not sure if msbuild currently needs the uptake, but if it does you'll need to move to use `-Version` variables and make the PackageReference updates after the variables have been pulled from Orchestrated Build.

tmat:
We can do this change but I'd recommend starting with moving your build definitions to YAML, then updating to the latest RepoToolset and after that we can make this change in such a way that's compatible with Maestro and Orchestrated Build.

rainersigwald:
@tmat what does this have to do with YAML?

build/Packages.props
AndyGerlicher: Global applies to all projects. The equivalent of putting a PackageReference in Directory.Build.props

build/Packages.props
AndyGerlicher: I just moved it, didn't add it new. So not sure.

build/Packages.props
AndyGerlicher: I tried to leave anything that would be supplied externally. So `$(NuGetPackageVersion)` could still be provided same as before.

build/Packages.props
rainersigwald: IIRC we use them to deploy binaries into our bootstrap folder. But maybe that stuff is now copied from the LKG instead.

tmat:
@rainersigwald The latest and future versions of RepoToolset will require some changes to the build definition. They are much easier to do with YAML build definitions.

AndyGerlicher:
That sounds good of course, but not sure what it has to do with this? This is about PackageReference items not having versions in them and using a single file to specify package versions. They can still flow upstream like before and all that.

src/Directory.Build.props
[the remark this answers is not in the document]
AndyGerlicher: https://docs.microsoft.com/en-us/nuget/tools/package-manager-ui
It means that a package is implicitly defined from an SDK. If set VS doesn't prompt to update it (since you can't) and Central Package Versioning won't error for having a version defined.

build/Packages.props
AndyGerlicher: Agreed, esp. given which packages those were. I disabled it entirely for that tree.

NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="roslyn-tools" value="https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
rainersigwald: Does this interfere with the byzantine repo-toolset approach to defining these things?

src/Directory.Build.props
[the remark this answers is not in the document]
tmat: Thanks.

src/Samples/Directory.Build.targets
```diff
@@ -1,5 +1,9 @@
 ﻿<Project>
-  
+
+  <ItemGroup>
+    <PackageReference Include="Nerdbank.GitVersioning" Version="2.1.23" Condition="'$(DisableNerdbankVersioning)' != 'true'" />
```
rainersigwald: Honestly not sure we need this since we don't ship them, but it's not terrible.

tmat:
The problem is with deleting the version variables here: https://github.com/Microsoft/msbuild/pull/3441/files#diff-9260067da6a2db86b9a5db30fa8ee2aeL41

The toolset imports `Versions.props` first. After that it imports version overrides supplied by Orchestrated Build, so that they overwrite the values of the `-Version` properties that should flow from Orchestrated Build. 

Also, you said you want to update RepoToolset to add the IsImplicitlyDefinted metadata. We should do that but then you'll need to use the latest RepoToolset to be able to remove your workaround.

build/Packages.props
```diff
@@ -0,0 +1,59 @@
+<Project>
+
+  <PropertyGroup>
+      <NuGetPackageVersion>4.8.0-preview1.5185</NuGetPackageVersion>
+
+      <!-- Prodcon-overrideable, using repo-toolset naming convention -->
+      <NuGetBuildTasksVersion>$(NuGetPackageVersion)</NuGetBuildTasksVersion>
+      <NuGetCommandsVersion>$(NuGetPackageVersion)</NuGetCommandsVersion>
+      <NuGetProtocolVersion>$(NuGetPackageVersion)</NuGetProtocolVersion>
```
AndyGerlicher: FYI this is where I would expect to add any dependencies from upstream providers to show up. So if ProdCon specifies a version it would be picked up here and specified below using that property. Should work exactly as it did before.

tmat: Oh, I see. I get it now.

tmat: Should be good then.

NuGet.Config
tmat: Not sure what you mean. This is needed due to https://github.com/Microsoft/msbuild/issues/2982

src/Samples/Directory.Build.targets
AndyGerlicher: We kinda do. We have targets elsewhere that run as after targets for GitVersioning. So we'd have to refactor that for these to build without it. Easy enough but might as well just version them the same.

build/Packages.props
```diff
@@ -0,0 +1,58 @@
...
+    <PackageReference Update="Microsoft.DotNet.BuildTools.GenAPI" Version="2.1.0-prerelease-02404-02" />
+    <PackageReference Update="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
+    <PackageReference Update="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" />
+    <PackageReference Update="Microsoft.NETCore.Compilers" Version="$(MicrosoftNetCoreCompilersVersion)" />
+    <PackageReference Update="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.15" />
+    <PackageReference Update="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="1.16.30" />
+    <PackageReference Update="Microsoft.Win32.Registry" Version="4.3.0" />
+    <PackageReference Update="NuGet.Build.Tasks" Version="$(NuGetBuildTasksVersion)" />
+    <PackageReference Update="PdbGit" Version="3.0.41" />
+    <PackageReference Update="Shouldly" Version="3.0.0" />
```
cdmihai: Should the test dependencies appear under a test only conditioned ItemGroup?

AndyGerlicher: No. Packages.props defines the list of allowed versions, it does not define that anything will use it. And the central package versioning SDK adds the enforcement of if you specify a version in your csproj (even a `$()`) it will error. The leaf node projects decide if they use a PackageReference by including the item. `GlobalPackageReference` later in the `Packages.props` are items that are applied to all projects. So we could add Shoudly there gated on `$(IsTestProject) == true` if we wanted.

PR #3445, opened 2018-06-22
Use XmlTextReader everywhere

### LINKED ISSUE #1401: NewLine normalization on .NET Core
Related to #1340

In Full Framework, we do not normalize any line endings in a user's project file. To do this we had to use `XmlTextReader`. This behavior is legacy in all versions of MSBuild and should be preserved.

In .NET Core, `XmlTextReader` is not available and there does not seem to be a way to *not* normalize EOL characters. This means doing any project file manipulation via the .NET Core version of MSBuild will result in all `\r\n` characters normalized to `\n` on Save.

In the absence of built-in functionality to do this (it was taken out in .NET Core) the only way I can think of is trying to determine the 'right' EOL character on Load and replacing on Save. This could be:
1. If any `\n` characters appear by itself, replace all NewLine characters with `\n` (do nothing, this is all .NET Core supports).
2. If all EOL characters are `\r\n` on the way in, `string.Replace()` on Save.

Thoughts? We could try to be fancier and count the number of `\r\n` and if they appear more often than `\n` use that? And of course if we do this for .NET Core, should we do the same in Full Framework MSBuild? I think not to preserve behavior, and I don't think there's a good solution here unfortunately.

rainersigwald:
This API wasn't in .NET Core when we originally ported but is now available in
all of our .NET Core targets. Using it everywhere improves programmatic editing
of MSBuild files.

Closes #1401.

### DIFF
--- a/src/Build.OM.UnitTests/Construction/WhiteSpacePreservation_Tests.cs
+++ b/src/Build.OM.UnitTests/Construction/WhiteSpacePreservation_Tests.cs
@@ -472,9 +472,7 @@ multi-line comment here
 
             VerifyAssertLineByLine(expected, actual);
 
-#if FEATURE_XMLTEXTREADER
             VerifyLineEndings(actual);
-#endif
         }
 
         private void VerifyAssertLineByLine(string expected, string actual)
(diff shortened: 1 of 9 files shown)


AndyGerlicher:
Not sure why it failed. Seems like Tasks unit tests failed but I didn't see a failure in the log. https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_CoreCLR_prtest/658/

AndyGerlicher:

PR #3447, opened 2018-06-22
Finding MSBuild via environment variable supports Visual Studio detection

cdmihai:
If the user points MSBUILD_EXE_PATH to an msbuild inside Visual Studio, then the BuildEnvironment should be configured to the Visual Studio file layout.

### DIFF
--- a/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
+++ b/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
@@ -59,15 +59,11 @@ namespace Microsoft.Build.Engine.UnitTests
             }
         }
 
-#if RUNTIME_TYPE_NETCORE
-        [Theory(Skip = "NA on .net core")]
-#else
         [Theory]
-#endif
         [InlineData(true)]
         [InlineData(false)]
-        [Trait("Category", "nonlinuxtests")]
-        [Trait("Category", "nonosxtests")]
+        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp, "No Visual Studio install for netcore")]
+        [PlatformSpecific(TestPlatforms.Windows)]
         public void FindVisualStudioEnvironmentByEnvironmentVariable(bool is64BitMSbuild)
         {
             using (var env = new EmptyVSEnviroment())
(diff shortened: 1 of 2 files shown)


cdmihai:
@dfederm

src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -59,6 +59,41 @@ public void FindBuildEnvironmentByEnvironmentVariable()
...
 
+#if RUNTIME_TYPE_NETCORE
+        [Theory(Skip = "NA on .net core")]
+#else
+        [Theory]
+#endif
+        [InlineData(true)]
+        [InlineData(false)]
+        [Trait("Category", "nonlinuxtests")]
+        [Trait("Category", "nonosxtests")]
```
rainersigwald:
```c#
        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp, "No Visual Studio install for netcore")]
        [PlatformSpecific(TestPlatforms.Windows)]
```

As a simplification?

src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
```diff
@@ -59,6 +59,41 @@ public void FindBuildEnvironmentByEnvironmentVariable()
...
+
+                env.WithEnvironment("MSBUILD_EXE_PATH", msBuildPath);
+                BuildEnvironmentHelper.ResetInstance_ForUnitTestsOnly(ReturnNull, ReturnNull, ReturnNull, env.VsInstanceMock, env.EnvironmentMock, () => false);
+
+                BuildEnvironmentHelper.Instance.Mode.ShouldBe(BuildEnvironmentMode.VisualStudio);
+                BuildEnvironmentHelper.Instance.MSBuildExtensionsPath.ShouldBe(vsMSBuildDirectory);
+                BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory.ShouldBe(msbuildBinDirectory);
+                BuildEnvironmentHelper.Instance.CurrentMSBuildExePath.ShouldBe(msBuildPath);
+                BuildEnvironmentHelper.Instance.CurrentMSBuildConfigurationFile.ShouldBe(msBuildConfig);
+                BuildEnvironmentHelper.Instance.RunningInVisualStudio.ShouldBeFalse();
```
rainersigwald: Possibly worth a comment because this confused me. We're not _running_ in VS at this point, but it's a VS instance.

src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
cdmihai: Yes. I didn't remember what they were so I copied them from other tests. Replaced throughout the file. Also took out the other tests which were expecting the VS mode under .net core. They were only passing due to how the environment was setup, and would have failed in real life anyway.

PR #3472, opened 2018-07-03
WIP: VSTS PR build test

rainersigwald:


chcosta:
VSTS YAML now supports `Build.Reason` in the template evaluation context which means that you can share the same yml file for both CI and internal builds if you want to.

For example Arcade uses the same yml file in its builds

https://github.com/dotnet/arcade/blob/master/.vsts-ci.yml#L6

.vsts-dotnet-ci.yml
```diff
@@ -3,6 +3,7 @@ trigger:
 
 phases:
 - phase: FullOnWindows
+  displayName: "Build on Windows using full MSBuild"
   queue: 'Hosted VS2017'
```
chcosta: Specify `timeoutInMinutes` - https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted-schema.md#queue

.vsts-dotnet.yml
```diff
@@ -55,7 +55,7 @@ steps:
   displayName: Run build.cmd
   inputs:
     filename: '$(comspec)'
-    arguments: '/c "call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat" & $(Build.Repository.LocalPath)\build.cmd -skiptests -bootstraponly -pack -sign -configuration Release -properties /p:SignType=$(SignType)"'
+    arguments: '/c "call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat" & $(Build.Repository.LocalPath)\build.cmd -pack -sign -configuration Release -properties /p:SignType=$(SignType)"'
```
chcosta: I think `VS140COMNTOOLS` might be a usable variable instead of  `C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools`

rainersigwald: `VS140COMNTOOLS` is dev14/VS 2015, so it's not a good option for us (we need the 2017/dev15 environment). The _right_ thing to do here is use `vswhere.exe` to find VS and call the script from there. We should do that but it'll keep for now, I think.

rainersigwald:
Just talked this over with @chcosta and @AndyGerlicher. We're going to try to light this up in parallel with the existing Jenkins CI infrastructure for a while. I'm going to push one more change to this PR and unless something goes wrong merge to the experimental branch, then prep a squashed payload for PR to master.

.vsts-dotnet-ci.yml
```diff
@@ -142,6 +152,7 @@ phases:
 
 #- phase: MonoOnMac
 #  queue: 'Hosted macOS Preview'
+#  timeoutInMinutes: 60
```
chcosta: nit: I know it's just a comment, but the spacing would be wrong if this were uncommented.  `timeoutInMinutes` should be a child of `queue`

rainersigwald: Thanks! I'm deleting the commented-out ones in my PR to master; there were several things like this that had drifted.

rainersigwald:
image
 

Yay! I'm going to close this in favor of getting #3530 working.

PR #3475, opened 2018-07-03
Fix how warnings are mutated

### LINKED ISSUE #3474: Warnings are not mutated if a project is built multiple times
### Steps to reproduce

Directory.Build.props
```xml
  <PropertyGroup>
    <MSBuildWarningsAsMessages>123</MSBuildWarningsAsMessages>
  </PropertyGroup>
  <Target Name="Repro" BeforeTargets="Build">
    <Warning Code="123" Text="Testing" />
  </Target>
</Project>
```

ClassLibrary1\ClassLibrary1.csproj
```xml
  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>
</Project>
```

ClassLibrary2\ClassLibrary2.csproj
```xml
  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="..\ClassLibrary1\ClassLibrary1.csproj" />
  </ItemGroup>
</Project>
```

ClassLibrary3\ClassLibrary3.csproj
```xml
  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="..\ClassLibrary1\ClassLibrary1.csproj" />
    <ProjectReference Include="..\ClassLibrary2\ClassLibrary2.csproj" />
  </ItemGroup>
</Project>
```

Command line
```
msbuild /m
```
### Expected  behavior
Warnings are suppressed

### Actual behavior
Warning is emitted for ClassLibrary1 or ClassLibrary2

### Root Cause
The list of warnings to treat as a message or error are keyed off the project instance.  As projects finish, the codes are [removed from the list](https://github.com/Microsoft/msbuild/blob/master/src/Build/BackEnd/Components/Logging/EventSourceSink.cs#L265-L266).  However, projects can start and finish multiple times.  This means the key used, a single `int`, will not suffice. The key must contain both the project instance ID and the project context ID.

jeffkl:
The list of warnings to mutate are stored according to the project instance ID.  When a project is finished, the warnings are removed.  However, a project starts and finishes several times during a build.  This leads to warnings failing to elevated or suppressed.

This fix moves the warning mutation logic out of EventSourceSink into the LoggingService itself instead.  This simplifies the implementation a little.

It also changes the key to the project instance ID and project context ID.  When projects start and finish, the key will work properly.  I moved the unit tests as well.

Fixes #3474


src/Build/BackEnd/Components/Logging/LoggingService.cs
```diff
@@ -1103,6 +1113,16 @@ internal void InjectNonSerializedData(LogMessagePacket loggingPacket)
         #endregion
 
         #region Private Methods
+        private static string GetWarningsAsErrorOrMessageKey(BuildEventContext buildEventContext)
+        {
+            return $"{buildEventContext.ProjectInstanceId}-{buildEventContext.ProjectContextId}";
```
cdmihai: Computing an int instead would avoid creating 3 string objects for each call.

src/Build/BackEnd/Components/Logging/LoggingService.cs
```diff
@@ -1258,18 +1276,84 @@ private void LoggingEventProcessor(object loggingEvent)
...
+                        BuildEventContext = warningEvent.BuildEventContext,
+                        ProjectFile = warningEvent.ProjectFile,
+                    };
+                }
+            }
+
+            if (loggingEvent is BuildErrorEventArgs errorEvent)
+            {
+                // Keep track of build submissions that have logged errors.  If there is no build context, add BuildEventContext.InvalidSubmissionId.
+                _buildSubmissionIdsThatHaveLoggedErrors.Add(errorEvent.BuildEventContext?.SubmissionId ?? BuildEventContext.InvalidSubmissionId);
```
cdmihai: >_buildSubmissionIdsThatHaveLoggedErrors [](start = 16, length = 39)

I realize now that `_buildSubmissionIdsThatHaveLoggedErrors` is per node, so hopefully its value is not required in the main node. This problem existed with the old implementation too.

src/Build/BackEnd/Components/Logging/LoggingService.cs
```diff
@@ -1258,18 +1276,84 @@ private void LoggingEventProcessor(object loggingEvent)
...
+            }
+
+            if (loggingEvent is ProjectFinishedEventArgs projectFinishedEvent && projectFinishedEvent.BuildEventContext != null)
+            {
+                _warningsAsErrorsByProject?.Remove(GetWarningsAsErrorOrMessageKey(projectFinishedEvent));
```
cdmihai: These calls support the idea that the key method should return an int.

cdmihai:
Can you add a unit test for the failing scenario in #3474?

src/Build/BackEnd/Components/Logging/LoggingService.cs
jeffkl: Its set by all nodes but only read by the main node.

PR #3478, opened 2018-07-04
fix issues when accessing files in windows system32 directory

kingofthebongo2008:
32 bit apps do not have access to system32 directory on windows.
however some 64 bit tools like link.exe access data here which is recored in tlog files.
when tlogs are read these files are not accessed and msbuild issues rebuilds.
example files required by link.exe are
C:\WINDOWS\SYSTEM32\C_1252.NLS
C:\WINDOWS\SYSTEM32\TZRES.DLL

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -503,6 +503,16 @@ namespace Microsoft.Build.Shared
 #endif
         }
 
+        private static readonly bool s_is64BitProcess = Marshal.SizeOf(NullIntPtr) == 8;
+
+        /// <summary>
+        /// Gets a flag indicating if we are running as a 64 bit process
+        /// </summary>
+        internal static bool Is64BitProcess
+        {
+            get { return s_is64BitProcess; }
+        }
+
         private static readonly object IsMonoLock = new object();
 
         private static bool? _isMono;
@@ -744,6 +754,18 @@ namespace Microsoft.Build.Shared
                 WIN32_FILE_ATTRIBUTE_DATA data = new WIN32_FILE_ATTRIBUTE_DATA();
                 bool success = false;
 
+                if (!Is64BitProcess)
+                {
+                    //32 bit processes do not have access to system32. to get time for files there we must remap to a special
+                    string windowsDirectory = Environment.SystemDirectory.ToLowerInvariant();
+                    string fullPathUpper = fullPath.ToLowerInvariant();
+
+                    if (fullPathUpper.StartsWith(windowsDirectory))
+                    {
+                        fullPath = fullPathUpper.Replace("system32", "sysnative");
+                    }
+                }
+
                 success = GetFileAttributesEx(fullPath, 0, ref data);
                 if (success)
                 {
@@ -885,6 +907,18 @@ namespace Microsoft.Build.Shared
 
                 WIN32_FILE_ATTRIBUTE_DATA data = new WIN32_FILE_ATTRIBUTE_DATA();
                 bool success = false;
+
+                if (!Is64BitProcess)
+                {
+                    //32 bit processes do not have access to system32. to get time for files there we must remap to a special
+                    string windowsDirectory = Environment.SystemDirectory.ToLowerInvariant();
+                    string fullPathUpper = fullPath.ToLowerInvariant();
+
+                    if (fullPathUpper.StartsWith(windowsDirectory))
+                    {
+                        fullPath = fullPathUpper.Replace("system32", "sysnative");
+                    }
+                }
 
                 success = NativeMethodsShared.GetFileAttributesEx(fullPath, 0, ref data);
 


kingofthebongo2008:
This is a sample file produced from tensorflow cmake buildsystem on windows 10 bulgarian

[link.read.1.tlog.zip](https://github.com/Microsoft/msbuild/files/2164138/link.read.1.tlog.zip)

src/Shared/NativeMethodsShared.cs
```diff
@@ -886,6 +908,18 @@ internal static DateTime GetLastWriteFileUtcTime(string fullPath)
                 WIN32_FILE_ATTRIBUTE_DATA data = new WIN32_FILE_ATTRIBUTE_DATA();
                 bool success = false;
 
+                if (!Is64BitProcess)
+                {
+                    //32 bit processes do not have access to system32. to get time for files there we must remap to a special
+                    string windowsDirectory = Environment.SystemDirectory.ToLowerInvariant();
+                    string fullPathUpper = fullPath.ToLowerInvariant();
+
+                    if (fullPathUpper.StartsWith(windowsDirectory))
```
cdmihai: You can use `StartsWith(String, StringComparison)` to avoid the extra ToLower string allocations.

cdmihai:
Why not use 64 bit msbuild instead?

src/Shared/NativeMethodsShared.cs
```diff
@@ -503,6 +503,16 @@ internal static bool IsBSD
 #endif
         }
 
+        private static readonly bool s_is64BitProcess = Marshal.SizeOf(NullIntPtr) == 8;
```
Therzok: IntPtr.Size == 8?

kingofthebongo2008: This is a copy of another file that checks the 64bit process.
If you have a better practice will replace it.

src/Shared/NativeMethodsShared.cs
kingofthebongo2008: ok, will change

src/Shared/NativeMethodsShared.cs
kingofthebongo2008: Tensorflow flow by default builds with 32 bit version of msbuild with the 32 bit c++ toolchain and fails.
Moving to 64 bit toolchain fixes the issue, but we get this, which is a problem.
Here is the link.
https://support.microsoft.com/en-us/help/942589/a-32-bit-application-cannot-access-the-system32-folder-on-a-computer-t

rainersigwald:
Unfortunately, I don't think it'll be this easy.

_Always_ swapping `sysnative` in for `system32` isn't correct, because it depends on the bitness of both MSBuild and the child process:

* For 32-bit MSBuild and 32-bit child, the paths match and shouldn't be transformed.
* For 32-bit MSBuild and 64-bit child, we should apply this transformation: map `system32` to `sysnative`. We should also map `syswow64` to `system32`.
* For 64-bit MSBuild and 32-bit child, we should apply an inverse transformation: accesses to `sysnative` should be mapped to `system32` and accesses to `system32` should be mapped to `syswow64`
* For 64-bit MSBuild and 64-bit child, the paths match and shouldn't be transformed.

To know which set of transformations to apply, we'd have to understand both bitnesses. It'd be hard to plumb that information through to this method in all cases, because it's used very generically.

**Can you elaborate on the problem you're observing, please?** While MSBuild is definitely looking at the wrong files for these paths, I would not expect it to cause spurious rebuilds, since the OS files will almost always be older than the build outputs. Maybe we can fix the observed problem without tackling the more complex path mapping.

kingofthebongo2008:
I am building tensorflow on windows with cmake.

Tensorflow on average builds around 4 - 5 hours on my machine.
One of the projects during the linking process spits out the link tlog file, that i have attached.
On a second build, msbuild cannot access these files in system32 and decides that a rebuild is needed.
The project which needs rebuilding is relatively high in the tree and this triggers a rebuild of the whole tensorflow project.

Basically it looks like: Touch one line of code, build for 4 hours, touch another line, another 4 hours of your life gone.

With the current fix this does not happen anymore and the targets are checked correctly.

What are these children processes? I do not think anything is needed to remap, every process is for itself.

We may remap only lines in the tlog files.

But this is a general issue with msbuild, it is just not realiable this way.

kingofthebongo2008:
We may remap in one of these functions, where the problem begins.

 	Microsoft.Build.Utilities.Core.dll!Microsoft.Build.Shared.NativeMethodsShared.GetLastWriteFileUtcTime(string fullPath) Line 919	C#
>	Microsoft.Build.Utilities.Core.dll!Microsoft.Build.Utilities.CanonicalTrackedInputFiles.IsUpToDate(Microsoft.Build.Framework.ITaskItem sourceFile) Line 564	C#
 	Microsoft.Build.Utilities.Core.dll!Microsoft.Build.Utilities.CanonicalTrackedInputFiles.CheckIfSourceNeedsCompilation(System.Collections.Concurrent.ConcurrentQueue<Microsoft.Build.Framework.ITaskItem> sourcesNeedingCompilationList, bool allOutputFilesExist, Microsoft.Build.Framework.ITaskItem source) Line 328	C#
 	Microsoft.Build.Utilities.Core.dll!Microsoft.Build.Utilities.CanonicalTrackedInputFiles.ComputeSourcesNeedingCompilationFromPrimaryFiles.AnonymousMethod__0(int index) Line 264	C#

rainersigwald:
This change would introduce the same problem in reverse when using a 32-bit `link.exe`: it would attempt to get the timestamp of, for example, `c:\windows\sysnative\tzres.dll` when the process actually read `c:\windows\system32\tzres.dll`.

> On a second build, msbuild cannot access these files in system32

This is the part I find confusing. Why? On my system at least there are equivalent files 

```
c:\Windows\System32\C_1252.NLS
c:\Windows\SysWOW64\C_1252.NLS
```

> We may remap in one of these functions

That helps scope the problem down but doesn't resolve the fundamental issue.

kingofthebongo2008:
I have only this:
c:\Windows\System32\C_1252.NLS
I do not have this
c:\Windows\SysWOW64\C_1252.NLS

But i have 
c:\windows\system32\tzres.dll
and
c:\windows\syswow64\tzres.dll

kingofthebongo2008:
srrms to me that msbuild should know the bitness of the tool somehow to be able to track the files in the tlog files, makes more sense this way

rainersigwald:
> msbuild should know the bitness of the tool somehow to be able to track the files in the tlog files

I agree that that's a reasonable request. Would you mind filing an issue to track it?

> I do not have this
> ```
> c:\Windows\SysWOW64\C_1252.NLS
> ```

What version of windows are you running? I see that file in both places on two Windows 10 version 1803 machines and an old Win7 VM I have installed. I wonder if something's gone slightly wrong with your machine, leading you to find this issue.

kingofthebongo2008:
I am on
Windows 10 Pro Insider Preview
Build 17692.rs_prerelease.180609-1317
User interface is Bulgarian.
I think this is why link accesses this C_1252.NLS, this is cyrillic,
though i build other projects here with visual studio and have no issues,
but i think i am using the 32 bit linker.

kingofthebongo2008:
ok, i filed an issue

rainersigwald:
Closing in favor of a more comprehensive fix for #3482, since the approach here doesn't work for all scenarios.

PR #3485, opened 2018-07-05
Update wiki clone

rainersigwald:


rainersigwald:
Note: I also updated the wiki pages themselves to point back to here, so this should be the last time this is necessary.

README.md
```diff
@@ -58,8 +58,8 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 
 Before you contribute, please read through the contributing and developer guides to get an idea of what kinds of pull requests we will or won't accept.
 
-* [Contributing Guide](https://github.com/Microsoft/msbuild/wiki/Contributing-Code)
-* [Developer Guide](https://github.com/Microsoft/msbuild/wiki/Building-Testing-and-Debugging)
+* [Contributing Guide](documentation/wiki/Contributing-Code.md)
```
cdmihai: Was this a search and replace all occurrences?

rainersigwald: I did it manually in this document (searched for `wiki`, fixed links). There may be some stragglers, but we have the "redirects" in place so that shouldn't be fatal.

PR #3459, opened 2018-06-28
[WIP] Capture metaproj in binlog (#3402)

ccastanedaucf:
Currently succeeds in collecting ```.metaproj``` files into ```.binlog``` when ```MSBUILDEMITSOLUTION=1```. Example below shows opening the ```.binlog``` from MSBuildStructuredLogViewer where```ConsoleApp1``` has a build dependency on ```ClassLib1```.

metaproj

PR #3488, opened 2018-07-06
Fix silent failure when attempting to build web projects outside of .NET Framework

ccastanedaucf:
Found this while debugging tests on #3459 

When generating the metaproject for a web project, if targeting a framework version >= 4.0 and aspnet_compiler.exe 4.0 is not found, ```ToolPath``` is silently set to null and ```ProjectInstance``` will end up failing elsewhere with a ```NullReferenceException```. Not a problem if targeting < 4.0 since the method already used a ```VerifyThrow``` check.

### DIFF
--- a/src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
+++ b/src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
@@ -2023,8 +2023,13 @@ EndGlobal
                 ProjectCollection collection = new ProjectCollection();
                 collection.RegisterLogger(logger);
 
-                ProjectInstance[] instances = SolutionProjectGenerator.Generate(solution, globalProperties, null, BuildEventContext.Invalid, collection.LoggingService);
-
+#if !FEATURE_INSTALLED_MSBUILD
+                Assert.Throws<InvalidProjectFileException>(() => {
+#endif
+                    ProjectInstance[] instances = SolutionProjectGenerator.Generate(solution, globalProperties, null, BuildEventContext.Invalid, collection.LoggingService);
+#if !FEATURE_INSTALLED_MSBUILD
+                });
+#endif
                 Version ver = new Version("4.34");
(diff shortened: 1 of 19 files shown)


rainersigwald:
Can you elaborate on how you hit the problem here? Since `aspnet_compiler.exe` is part of the .NET Framework and we need .NET Framework 4.6+ to run, I wouldn't expect to ever go down this new error path. It's not like 2.0, where we might be running on a machine that doesn't have that installed. Based on the assert you put in, maybe we should just change the solution parser code when we're running on .NET Core to just error out and not even try to build this old style of project?

ccastanedaucf:
Sure, as seen [here](https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_Full_prtest/730/testReport/junit/Microsoft.Build.UnitTests.Construction/SolutionProjectGenerator_Tests/TestTargetFrameworkVersionGreaterThan4/) ```toProjectRootElement()``` hits a ```NullReferenceException``` when attempting to copy over the key/value pairs in each task after failing to find the compiler. Noticed it would strangely pass on my machine only on full framework then fail everywhere on the CI tests.

ccastanedaucf:
So this looks fishy. This snippet can still execute on 4.6 even though the compiler should be there.
https://github.com/Microsoft/msbuild/blob/d136a42365dba0e6fe5f96dbd0ee4ba6ca77aa8b/src/Shared/FrameworkLocationHelper.cs#L1307-L1310

The conditions for ```FEATURE_INSTALLED_MSBUILD``` are ```MachineIndependentBuild != true``` and ```TargetFramework.StartsWith('net4')```. Not sure what triggers the first one, but if the compiler is always there on 4.6 would this be incorrect?

rainersigwald:
Hmm, I think we may not have granular-enough controls. `MachineIndependentBuild` is intended to disable some of our features that entangle with Visual Studio and the registry. I'm surprised to see it flow into things like `GetPathToDotNetFramework`. I suspect that happened because `FEATURE_INSTALLED_MSBUILD` was originally only unset for .NET Core and `MachineIndependentBuild` was bolted on later.

PR #3494, opened 2018-07-09
WIP: Drop -preview for 15.8 release

AndyGerlicher:


rainersigwald:
Do we need to wait until p5 ships before we can do this? We'll need to publish a myget package for CLI to consume for p4, right?

PR #3496, opened 2018-07-09
Restore private enum value ResourceType.Prerequisite

erikma:
The value is produced via enum string parsing even though it is not used in the MSBuild codebase directly


erikma:
@rainersigwald Does retargeting also result in auto cherry-pick back to master? Manual cherry-pick?

rainersigwald:
No, but we (I) merge release branches back to master on a semi-regular basis.

PR #3504, opened 2018-07-11
Specify explicit MicroBuildPluginsSwixBuildVersion

AndyGerlicher:
This was causing an issue with the internal build signing our vsix package.


AndyGerlicher:
@tmat

Issue #3505, opened 2018-07-11
Copy task should log absolute file paths


KirillOsenkov:
Currently Copy task logs relative file paths, and sometimes this is ambiguous. To properly reason about double-writes we'd need the Copy task to call Path.GetFullPath() on all paths for source and destination before logging them.

See https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/180

rainersigwald:
Seems reasonable to me. It's rare to change a project's CWD, so it's _almost_ possible to disambiguate by combining with the project path folder. But it's possible to change working directory.

geeeyetee:
In addition to ensuring absolute path in the log, I think it makes sense to do the same in the `Identity` of `CopiedFiles` output items. If that presents too much risk of breaking existing scripts, a new metadata key (`AbsolutePath`?) could be added to `CopiedFiles` output items.

rainersigwald:
@geeeyetee There's already an automatic `%(FullPath)` metadatum; it's computed on demand rather than added to the item directly.

I'm definitely leery of changing `%(Identity)` of the output; those are usually relative paths and it's very hard to predict what would break on such a change.

geeeyetee:
@rainersigwald How silly of me. I had been staring myopically at output using _MSBuild Structured Log Viewer_. Your explanation of the on-demand nature of `%(FullPath)` explains why a log viewer would not be able to display it alongside other metadata.

Agreed that changing `%(Identity)` of the output would likely break stuff-- and for no real gain since during build `%(FullPath)` is available.

PR #3506, opened 2018-07-11
Log absolute source and destination paths in copy task (#3505)

ccastanedaucf:


src/Tasks/Copy.cs
```diff
@@ -262,7 +262,9 @@ private void LogDiagnostic(string message, params object[] messageArgs)
             if (!linkCreated)
             {
                 // Do not log a fake command line as well, as it's superfluous, and also potentially expensive
-                Log.LogMessageFromResources(MessageImportance.Normal, "Copy.FileComment", sourceFileState.Name, destinationFileState.Name);
+                string sourceFilePath = Path.GetFullPath(sourceFileState.Name);
+                string destinationFilePath = Path.GetFullPath(destinationFileState.Name);
```
rainersigwald: Should these instead use `FileUtilities.GetFullPathNoThrow`? I think it's less allocatey on Windows and it's used in `GenerateResource` so it should be available in this assembly already.

ccastanedaucf: ```FileUtilities.GetFullPathNoThrow``` is a little misleading since it calls ```FileUtilities.NormalizePath``` which replaces the directory separators on non-Windows. Not sure if its callers depend on that behavior.

rainersigwald: Is that a problem in this situation though?

ccastanedaucf: Path would be semantically wrong in the log but that's really it

ccastanedaucf: Oops completely misread what ```FixFilePath``` does, it shouldn't be a problem

KirillOsenkov:
Wow! Thank you so much!

PR #3507, opened 2018-07-11
Add long path switches to config

ccastanedaucf:
Should be safe now that we're on Dev16.

### DIFF
--- a/src/MSBuild/MSBuild.exe.manifest
+++ b/src/MSBuild/MSBuild.exe.manifest
@@ -1,7 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
   <assemblyIdentity version="15.1.0.0" name="MSBuild" />
-
+  
+  <application xmlns="urn:schemas-microsoft-com:asm.v3">
+      <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
+          <ws2:longPathAware>true</ws2:longPathAware>
+      </windowsSettings>
+  </application>
+  
   <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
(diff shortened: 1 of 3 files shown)


ccastanedaucf:

ccastanedaucf:

ccastanedaucf:
I remember spending a while trying to figure that one out, some unit tests would use the ```app.config``` and others (I believe everything under Tasks?) would completely ignore it. I think I had to set this config file for the unit tests: https://github.com/Microsoft/msbuild/blob/master/src/Shared/UnitTests/App.config

ccastanedaucf:
I remember spending a while trying to figure that one out, some unit tests would use the ```app.config``` and others (I believe everything under Tasks?) would completely ignore it.

ccastanedaucf:
I'm double checking that one since I'd have a better idea of what's going wrong now vs 5 weeks ago

ccastanedaucf:
Completely figured it out. So .NET Framework tests can't use the two ```App.config``` switches since the tests run on 4.6.1, and .NET Core tests will already have them enabled by default. Remember the switches are so we can target 4.6.1 but still use the new long path behavior when running on > 4.6.2.

What makes it seem like ```App.config``` is being read: a bunch of tests on .NET Framework fail since they expect a ```PathTooLong``` exception that doesn't get thrown. But since most of the path-related calls on Windows net46 use native calls, it behaves as if the switches were set for many tests.

Which also means... if the remaining managed path-related calls went through ```FileUtilities```, ```FileSystems```, ect. where the conditional switches for native/managed calls are, long paths would actually work even when running on 4.6.1. And at that point the ```app.config``` switches wouldn't even be necessary. Would also make running tests with long paths enabled on 4.6.1 really easy since the behavior would be uniform, and not a mix of things that throw and don't throw ```LongPathException```. But that's a whole other issue.

hickford:
So MSBuild 16 and later (Visual Studio 2017 and later) can handle paths? Is this documented anywhere?

PR #3516, opened 2018-07-15
Unqualified metadata case sensitivity fix

### LINKED ISSUE #2870: MSBuild treats unqualified metadata case sensitive
### Steps to reproduce

Project file
```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
  <ItemGroup>
    <Foo Include="Foo">
      <SENSITIVE>X</SENSITIVE>
    </Foo>
  </ItemGroup>
  <Target Name="Build">
    
    <Message Importance="high" Text="%(Foo.FileName)=%(Foo.sensitive)"/>
    <!-- Output: Foo=X -->
    
    <Message Importance="high" Text="%(Foo.FileName)=%(sensitive)"/>
    <!-- error MSB4096: The item "Foo" in item list "Foo" does not define a value for metadata "sensitive".  
                        In order to use this metadata, either qualify it by specifying %(Foo.sensitive), or ensure that
                        all items in this list define a value for this metadata. -->
  </Target>
</Project>
```

Command line
```
msbuild
```
### Expected  behavior
Outputs `Foo=X` twice

### Actual behavior
Fails with MSB4096

### Environment data
`msbuild /version` output: 15.5.180.51428. Also tested with MSBuild from .NET 4.0.

OS info: Windows 10 1703

MeikTranel:
_(This is my first PR in a microsoft repository (or any popular repository for that matter)... so any feedback would be greatly appreciated.)_

This PR attempts to fix #2870 

Note: I wasn't sure were exactly to put the tests. I hope i haven't missed it quite as much 😄

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -30,7 +30,6 @@ using ProjectItemInstanceFactory = Microsoft.Build.Execution.ProjectItemInstance
 
 using Xunit;
 using Microsoft.Build.BackEnd;
-using Microsoft.Build.Engine.UnitTests;
 using Microsoft.Build.Shared.FileSystem;
 using Shouldly;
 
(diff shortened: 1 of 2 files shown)


MeikTranel:
@rainersigwald do you know what's happening with osx build? From what I can gather it seems to be unrelated to my changes.

src/Build/Collections/CopyOnWritePropertyDictionary.cs
```diff
@@ -299,7 +299,8 @@ public bool Equals(CopyOnWritePropertyDictionary<T> other)
         /// </summary>
         bool IDictionary<string, T>.ContainsKey(string key)
         {
-            return PropertyNames.Contains(key);
+            T discard;
+            return ((IDictionary<string, T>)this).TryGetValue(key, out discard);
```
rainersigwald: Did you consider

```c#
return _properties.ContainsKey(key);
```

? That seems a bit more straightforward, and avoids memory allocations for the unneeded `T`.

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -650,5 +652,100 @@ public void MetadataPropertyFunctionBug()
...
+                        <SensitiVE>X</SensitiVE>
+                    </Foo>
+                </ItemGroup>
+                <Target Name=`AccessItemMetadata`>
+                    <CreateProperty Value=`%(Foo.FileName)=%(Foo.sensitive)`>  
+                        <Output TaskParameter=`Value` PropertyName=`Qualified` />  
+                    </CreateProperty>          
+                    <CreateProperty Value=`%(Foo.FileName)=%(sensitive)`>  
+                        <Output TaskParameter=`Value` PropertyName=`Unqualified` />  
+                    </CreateProperty>
```
rainersigwald: Can you change this to use more-modern syntax like

```xml
<PropertyGroup>
 <Qualified>%(Foo.FileName)=%(Foo.sensitive)</Qualified>
 <Unqualified>%(Foo.FileName)=%(sensitive)</Unqualified>
</PropertyGroup>
```

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -650,5 +652,100 @@ public void MetadataPropertyFunctionBug()
...
+            ProjectInstance project = new Project(
+                XmlReader.Create(
+                    new StringReader(projectContents)
+                )
+            ).CreateProjectInstance();
+
+            project.Build();
+
+            Assert.NotNull(project.GetProperty("Qualified"));
+            Assert.NotNull(project.GetProperty("Unqualified"));
```
rainersigwald: Can you change these asserts to additionally check the content of the properties for good measure?

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
MeikTranel: This one is funny. Structured log viewer got me so focused on always using tasks over shorthand property/itemgroups, because CreateProperty etc. give better binary logs. Maybe i should raise an issue over at https://github.com/KirillOsenkov/MSBuildStructuredLog :smile:

cdmihai:
@MeikTranel @rainersigwald 

Moved the tests closer to the place of failure, the Expander. Also switched them to what some of the other Expander tests are doing, which is do a build and assert output.

cdmihai:
@MeikTranel I removed the merge commit in favor of rebase to have history show better, and I had to force push to your branch.

src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -1138,6 +1137,85 @@ public void HasMetadata()
...
+                    <Message Importance=`high` Text=`UnqualifiedNotMatchCase %(Foo.FileName)=%(sensitive)`/>
+                    <Message Importance=`high` Text=`UnqualifiedMatchCase %(Foo.FileName)=%(SENSITIVE)`/>
+                  </Target>
+                </Project>
+                ");
+
+            logger.AssertLogContains("QualifiedNotMatchCase Foo=X");
+            logger.AssertLogContains("QualifiedMatchCase Foo=X");
+            logger.AssertLogContains("UnqualifiedNotMatchCase Foo=X");
+            logger.AssertLogContains("UnqualifiedMatchCase Foo=X");
```
MeikTranel: I was trying to avoid using the log assertions, because i thought they would introduce lots of overhead and asserting via lookup over large strings didn't feel as precise to me? Is there big advantage of these i am missing?

src/Build/Collections/CopyOnWritePropertyDictionary.cs
```diff
@@ -299,8 +299,7 @@ public bool Equals(CopyOnWritePropertyDictionary<T> other)
         /// </summary>
         bool IDictionary<string, T>.ContainsKey(string key)
         {
-            T discard;
-            return ((IDictionary<string, T>)this).TryGetValue(key, out discard);
+            return ((IDictionary<string, T>)this).TryGetValue(key, out _);
```
MeikTranel: @cdmihai although it is kind of pointless because the hint @rainersigwald was totally right and improves this further than just the avoided allocation of discard...
what happens with the discard underscore in terms of IL? does it just completely avoid the allocation?

src/Build/Collections/CopyOnWritePropertyDictionary.cs
MeikTranel: I stepped through this to compare what's happening. Aside from the allocation this is better in other ways. the sheer decrease in branch depth is worth it from my perspective. 
the old solution went through tons of functions before eventually landing where the case insensitive lookup was happening; by going through the wrapped dictionary's containskey that path is way more direct. 

Good call

src/Build/Collections/CopyOnWritePropertyDictionary.cs
rainersigwald: I think it's just cleaner to look at and produces almost-identical IL (except that the local variable isn't named). Example IL from https://dotnetfiddle.net/zOkQjl

```
    .locals init ([0] class DataClass V_0)
...
    IL_0001:  ldc.i4.3
    IL_0002:  ldloca.s   V_0
    IL_0004:  call       void Program::Method(int32,
                                              class DataClass&)
```

So it still has to allocate the local and pass it to the method with the `out` param, but it's easier to tell at the call site that we don't care about the answer. I don't know if the JIT does anything clever; I don't see anything about that in [the docs](https://docs.microsoft.com/en-us/dotnet/csharp/discards#calls-to-methods-with-out-parameters).

src/Build.UnitTests/Evaluation/Expander_Tests.cs
rainersigwald: IMO the advantage is consistency with the other similar tests, which outweighs the efficiency problems.

MeikTranel: That's true. I can see that.

cdmihai:
@MeikTranel 
Squashed and merged. Thank you for your contribution!

PR #3518, opened 2018-07-16
Support ctrl-c cancellation when replaying a binary log

rainersigwald:
I was looking at an issue where I issued an incorrect replay command and tried to cancel it, but it didn't work.

So . . . now it works.

I think we could improve some of the existing cancellation logic by plumbing the same CTS down, but I didn't do that here.

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -3391,7 +3391,7 @@ namespace Microsoft.Build.CommandLine
 
             try
             {
-                replayEventSource.Replay(binaryLogFilePath);
+                replayEventSource.Replay(binaryLogFilePath, s_cts.Token);
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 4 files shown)


src/MSBuild/XMake.cs
```diff
@@ -109,10 +109,9 @@ public enum ExitType
         private static ManualResetEvent s_cancelComplete = new ManualResetEvent(true);
 
         /// <summary>
-        /// Set to 1 when the cancel method has been invoked.
-        /// Never reset to false: subsequent hits of Ctrl-C should do nothing
+        /// Cancel when handling Ctrl-C
         /// </summary>
-        private static int s_receivedCancel;
+        private static CancellationTokenSource s_cts = new CancellationTokenSource();
```
cdmihai: nit: Rename to s_cancellationToken

rainersigwald: #3521

PR #3519, opened 2018-07-16
Merge Vs15.9stg -> master

cdmihai:


src/Directory.BeforeCommon.targets
```diff
@@ -136,7 +136,7 @@
 
   <PropertyGroup>
     <IBCMergeSubPath>x86/MSBuild</IBCMergeSubPath>
-    <IBCMergeBranch>lab/d15.8stg</IBCMergeBranch>
+    <IBCMergeBranch>rel/d15.8</IBCMergeBranch>
```
AndyGerlicher: This is wrong. But it's not worse than the current value, we just need to remember to change it to master at some point.

PR #3525, opened 2018-07-18
Exclude CompilerGeneratedAttribute from ref sources

AndyGerlicher:
This seems to be a common refactor and is completely useless to review as part of our public API. Since we missed an actual breaking change I thought this would be useful. Also figured since the tool supported it I'd throw a license header in there as well.


ref/ExcludeAttributeList.txt
```diff
@@ -0,0 +1 @@
+T:System.Runtime.CompilerServices.CompilerGeneratedAttribute
```
rainersigwald: Can you add `System.Diagnostics.DebuggerStepThroughAttribute` while we're at it? That one is also not interesting to API consumers and can change depending on our whims.

AndyGerlicher:
@rainersigwald updates. Now excluding:
```C#
T:System.Diagnostics.DebuggerBrowsableAttribute
T:System.Diagnostics.DebuggerDisplayAttribute
T:System.Diagnostics.DebuggerStepThroughAttribute
T:System.Runtime.CompilerServices.CompilerGeneratedAttribute
```

PR #3527, opened 2018-07-19
WIP: Update GenAPI version

AndyGerlicher:
@weshaggard any idea what's up with this? Adding some obvious garbage to the source. Also seems to change spacing which I guess is fine.

### DIFF
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -4,7 +4,9 @@ namespace Microsoft.Build.Framework
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public partial struct BuildEngineResult
     {
-        public BuildEngineResult(bool result, System.Collections.Generic.List<System.Collections.Generic.IDictionary<string, Microsoft.Build.Framework.ITaskItem[]>> targetOutputsPerProject) { throw null;}
+        private object _dummy;
+        private int _dummyPrimitive;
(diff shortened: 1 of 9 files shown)


ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -4,7 +4,9 @@ namespace Microsoft.Build.Framework
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
     public partial struct BuildEngineResult
     {
-        public BuildEngineResult(bool result, System.Collections.Generic.List<System.Collections.Generic.IDictionary<string, Microsoft.Build.Framework.ITaskItem[]>> targetOutputsPerProject) { throw null;}
+        private object _dummy;
```
weshaggard: >Adding some obvious garbage to the source.

Is this the garbage you were referring to? If so that is by design see https://github.com/dotnet/corefx/issues/6185.

AndyGerlicher:
Thanks Wes for the description of the change. I think maybe the name threw me off thinking it was a mistake.

Closing this for now, we're not having any issues and doesn't really benefit us to have the additional private fields.

weshaggard:
>Closing this for now, we're not having any issues and doesn't really benefit us to have the additional private fields.

Adding these fields can sometimes cause source breaking changes and potentially incorrect code being written so I'd suggest getting updated as soon as you can to help reduce your exposure.

AndyGerlicher:
We actually just use this to review in PRs. Would you still recommend having this in that case? We were on the fence about just taking this but since we don't compile against these it didn't seem important.

weshaggard:
If it is for just review then no don't worry about taking it, unless you update it often from the tool and you want to eliminate diff noise. 

Do you guys plan to ship refs in your packages at any point?

PR #3528, opened 2018-07-19
Remove netcoreap20

cdmihai:


cdmihai:
Here's me thinking that deleting the tfm will just work. :)

cdmihai:
@crummel, does this break any sourcebuild scenarios?

crummel:
@cdmihai This should be fine for source-build.  We should be on at least netcoreapp2.1 for everything consuming MSBuild.  I'll try it out but no need to hold it up on our account.

joeltankam:
@AndyGerlicher can you please explain why it was necessary to remove .NET Standard 2.0 targeting (d351690) ? 
Now libraries (not applications) referencing `Microsoft.Build` can no longer target .NET Standard, which is inconvenient.

AndyGerlicher:
I don't really remember the details, but I'm pretty sure that Microsoft.Build.dll uses APIs that are not available in all implementations of netstandard2.0. I think we're in agreement here that it is inconvenient, but I don't believe that it can be resolved without breaking changes.

dsplaisted:
@joeltankam The reason was that it's not generally possible to create working .NET Standard build tasks.  There are lots of APIs (System.Reflection.Metadata, System.Collections.Immutable, System.Text.Json) where build tasks that use those APIs need to include the implementations, and the implementations are different for .NET Core and .NET Framework.  So compiling an MSBuild task for .NET Standard doesn't help if you need to ship different dependencies for .NET Core and .NET Framework.

joeltankam:
@AndyGerlicher @dsplaisted Thank you.

PR #3530, opened 2018-07-20
Adopt VSTS CI builds (in parallel to existing Jenkins)

rainersigwald:
I think the versioning changes I had to make here break the official build, so I need to run an official build before this is ready for review.

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -1,0 +1,146 @@
+trigger:
+- master
+- exp/*
+
+phases:
+- phase: FullOnWindows
+  displayName: "Build on Windows using full MSBuild"
+  queue: 'Hosted VS2017'
+  steps:
+  - script: set
+    displayName: Command Line Script
+  - task: BatchScript@1
+    displayName: VsDevCmd
+    inputs:
+      filename: 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat'
+      modifyEnvironment: true
+  - task: BatchScript@1
+    displayName: cibuild.cmd
+    timeoutInMinutes: 60
+    inputs:
+      filename: 'build/cibuild.cmd'
+  - task: PublishTestResults@2
+    displayName: Publish .NET Framework Test Results
+    inputs:
+      testRunTitle: 'Windows-on-full Full Framework'
+      testRunner: XUnit
+      testResultsFiles: 'artifacts/**/*UnitTests_net46*.xml'
+      publishRunAttachments: true
+      mergeTestResults: true
+    condition: always()
+  - task: PublishTestResults@2
+    displayName: Publish .NET Core 2.0 Test Results
+    inputs:
+      testRunTitle: 'Windows-on-full .NET Core 2.0'
+      testRunner: XUnit
+      testResultsFiles: 'artifacts/**/*UnitTests_netcoreapp2.0*.xml'
+      publishRunAttachments: true
+      mergeTestResults: true
+    condition: always()
+  - task: PublishTestResults@2
+    displayName: Publish .NET Core 2.1 Test Results
+    inputs:
+      testRunTitle: 'Windows-on-full .NET Core 2.1'
+      testRunner: XUnit
+      testResultsFiles: 'artifacts/**/*UnitTests_netcoreapp2.1*.xml'
+      publishRunAttachments: true
+      mergeTestResults: true
+    condition: always()
+
+- phase: CoreOnWindows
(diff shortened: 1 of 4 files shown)


chcosta:
Not a blocker at all to this PR, just capturing my thoughts..

It would be nice if we could get the CI yml to be something uniformly applicable to almost any "RepoToolset" repo.  Modulo some command-line arguments / config info which are the only things necessary to be described in the repo.  

Arcade, today, is somewhat close to that except it's obviously not quite universal enough.  It doesn't do any Mac builds, it has some explicit command-line arguments that could be moved to configuration properties, it uses a single file for both official and internal builds.  Pulling out the official stuff would greatly simplify the yml

I suppose, that there are a couple of ways that this could be reasonably brought up.  Arcade was written under the assumption that everything under `eng/common` folder would be ubiquitously available via Darc / Maestro.  That hasn't happened yet and it's not clear what the timeline is, but we could make use of template [reuse from other repositories](https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted-reusefromrepo.md) and keep the templates for this in a branch of Arcade.


FYI @tmat @mmitche

rainersigwald:
I think this is ready to go. The packages it produces are consumable

* Full internal build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=1887197&view=results
* Corresponding insertion with VS val build: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/133812?_a=overview

And the CI build badges correctly (see above) and should be passing: https://dotnet.visualstudio.com/public/_build/results?buildId=7523

.vsts-dotnet-ci.yml
```diff
@@ -0,0 +1,146 @@
+trigger:
+- master
+- exp/*
+
+phases:
+- phase: FullOnWindows
+  displayName: "Build on Windows using full MSBuild"
+  queue: 'Hosted VS2017'
+  steps:
+  - script: set
```
chcosta: Is this step needed?

.vsts-dotnet-ci.yml
```diff
@@ -0,0 +1,146 @@
...
+    displayName: Command Line Script
+  - task: BatchScript@1
+    displayName: VsDevCmd
+    inputs:
+      filename: 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat'
+      modifyEnvironment: true
+  - task: BatchScript@1
+    displayName: cibuild.cmd
+    timeoutInMinutes: 60
```
chcosta: I think the default timeout for steps is infinite.

The default for the phase is 60, not sure if you need to explicitly set this here - https://github.com/Microsoft/vsts-agent/blob/master/docs/preview/yamlgettingstarted-job.md#timeout

.vsts-dotnet-ci.yml
rainersigwald: Not strictly, but it's been very useful in our official build to debug what's going on in the environment when things are really confusing. I'll remove it.

.vsts-dotnet-ci.yml
rainersigwald: I think I couldn't figure out how to set it explicitly on phase so I set it on the part that could hang. Removed now; we can adjust later.

rainersigwald:
Ok, so: PR builds are failing now for mysterious reasons--I'm talking with VSTS folks about improving diagnosability at least (and while we're at it they can tell me what went wrong).

Before that failed, some tests were failing: https://dotnet.visualstudio.com/public/_build/results?buildId=7553&view=ms.vss-test-web.test-result-details

```
Shouldly.ShouldAssertException : \"Environment at start of build:\r\nPath = D:\\a\\1\\s\\artifacts\\Debug\\bootstrap\\...\"\n should contain (case insensitive comparison)\n\"BUILD_SOURCEVERSIONMESSAGE = Set build version only in SetMicrobuildVersion\n\nVSTS PR jobs don't have permission to set the version of the build job,\nso move the attempt to do so into an explicit place, and call into\nthat place before establishing signing and before the build completes.\"\n but did not
```

I _suspect_ but can't yet prove this is because of newline normalization (and I guess "latest commit message" isn't a Jenkins environment variable so we haven't seen this).

.vsts-dotnet-ci.yml
chcosta: Yeah, the experience for PR builds, where you can't publish build artifacts (yet) and capture binlogs, is pretty bad. They (VSTS) have an active issue to improve this.

rainersigwald:
screenshot from 2018-07-24 15-54-40
let's do it. @Microsoft/msbuild-maintainers objections? somebody want to sign off?

PR #3535, opened 2018-07-23
WIP: Ensure DefaultProjectTypeGuid is set when cross-targeting.

peterhuene:
For C# and VB projects that are cross-targeted, the cross-targeting targets are
evaluated from the root project and not the "current version" targets for the
languages (this is currently where `DefaultProjectTypeGuid` is set for each
language).  The "current version" targets are evaluated only from the "inner"
builds perform for each framework target; thus the `DefaultProjectTypeGuid` is
never set from the evaluation of a cross-targeted project file.

This means that when `dotnet` evaluates the project to determine the default
project type GUID to use for a command like `dotnet sln add`, it fails to do
and displays an error message to the user.

See related CLI issue dotnet/cli#9477.

The fix is to duplicate setting `DefaultProjectTypeGuid` in the cross-targeting
targets.  This allows the property to be visible when evaluating the root
project file.

### DIFF
--- a/src/Tasks/Microsoft.CSharp.CrossTargeting.targets
+++ b/src/Tasks/Microsoft.CSharp.CrossTargeting.targets
@@ -16,6 +16,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <!-- Import design time targets before the common crosstargeting targets, which import targets from Nuget. -->
   <PropertyGroup>
      <CSharpDesignTimeTargetsPath Condition="'$(CSharpDesignTimeTargetsPath)'==''">$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed\Microsoft.CSharp.DesignTime.targets</CSharpDesignTimeTargetsPath>
+     <DefaultProjectTypeGuid Condition=" '$(DefaultProjectTypeGuid)' == '' ">{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</DefaultProjectTypeGuid>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


peterhuene:
Note: this has been approved by shiproom for "2.1.4xx".  However, it's unclear *which* version of the SDK it should go in currently. 

As such, I'm marking this "WIP".

rainersigwald:
This looks fine, but the VS shiproom bar is high and clearing the 2.1.4xx shiproom bar may not be sufficient for them; we'd need to check that out carefully.

Pilchie:
To bring to VS shiproom, we'd need a VSO bug with the ask mode template filled out for QB approval.  Questions like:

1. Is this a regression?
2. How many customers does it affect? 
3. What workarounds do they have, and how discoverable are they?
4. What risks are associated with the change, and how can we mitigate them?

Will be key here.

peterhuene:
@Pilchie https://devdiv.visualstudio.com/DevDiv/_workitems/edit/653834

Pilchie:
Is there an equivalent file for FSharp that should also be updated?

peterhuene:
F# is unaffected because `DefaultProjectTypeGuid` is set directly in [the SDK props](https://github.com/dotnet/netcorecli-fsc/blob/0cb690ec4c29205f1b8471b8d197324fdc3f73e9/src/FSharp.NET.Sdk/build/FSharp.NET.Sdk.props#L22) and thus always has effect.  I've confirmed that manually too.

peterhuene:
Closing this PR for now.  I'm going to open an MSBuild issue to address this going forward, but in the meantime we'll have a workaround for this in the 2.1.4xx .NET Core SDK.

peterhuene:
Opened #3558 to track fixing this in MSBuild in the future.

PR #3537, opened 2018-07-24
Fix a race condition in ToolLocationHelper

### LINKED ISSUE #3036: ToolLocationHelper.GetPathToReferenceAssemblies not thread safe
*Sadly no clear repro on this one but I think a fix can be found by just looking at the code (which I did briefly).*

I'm evaluating in // many .csproj from a huge solution (hundreds of projects) to get their output path. Sometimes on some build machines I get:
```
The expression "[Microsoft.Build.Utilities.ToolLocationHelper]::GetPathToStandardLibraries(.NETFramework, v4.6.1, '', AnyCPU)" cannot be evaluated. 
Collection was modified; enumeration operation may not execute. 
C:\Program Files (x86)\MSBuild\14.0\bin\amd64\Microsoft.Common.CurrentVersion.targets
```

I'm using Microsoft.Build@15.5.180 and my .sln is a VS2015 one.

When looking at the code:
https://github.com/Microsoft/msbuild/blob/75e5117ab8be40faad4601904600aceae6bf9ee1/src/Utilities/ToolLocationHelper.cs#L1779
There is some lock around some static variables but I guess some might be missing when iterating on the returned list. Maybe use ConcurrentDictionnary ?

AntonLapounov:
Fixes #3036.  Modifying the `dotNetFrameworkReferenceAssemblies` collection after publishing it bumps its version and breaks the `foreach` loop in the `GetPathToStandardLibraries` method.

### DIFF
--- a/src/Utilities/ToolLocationHelper.cs
+++ b/src/Utilities/ToolLocationHelper.cs
@@ -2108,11 +2108,6 @@ namespace Microsoft.Build.Utilities
                 }
             }
 
-            lock (s_locker)
-            {
-                s_cachedReferenceAssemblyPaths[referenceAssemblyCacheKey] = dotNetFrameworkReferenceAssemblies;
-            }
-
             for (int i = 0; i < dotNetFrameworkReferenceAssemblies.Count; i++)
             {
                 if (
@@ -2120,10 +2115,13 @@ namespace Microsoft.Build.Utilities
                         Path.DirectorySeparatorChar.ToString(),
                         StringComparison.Ordinal))
                 {
-                    dotNetFrameworkReferenceAssemblies[i] = string.Concat(
-                        dotNetFrameworkReferenceAssemblies[i],
-                        Path.DirectorySeparatorChar);
+                    dotNetFrameworkReferenceAssemblies[i] += Path.DirectorySeparatorChar;
                 }
+            }
+
+            lock (s_locker)
+            {
+                s_cachedReferenceAssemblyPaths[referenceAssemblyCacheKey] = dotNetFrameworkReferenceAssemblies;
             }
 
             return dotNetFrameworkReferenceAssemblies;


src/Utilities/ToolLocationHelper.cs
```diff
@@ -2108,24 +2108,22 @@ public static IList<string> GetPathToReferenceAssemblies(string targetFrameworkR
...
-            lock (s_locker)
-            {
-                s_cachedReferenceAssemblyPaths[referenceAssemblyCacheKey] = dotNetFrameworkReferenceAssemblies;
-            }
-
             for (int i = 0; i < dotNetFrameworkReferenceAssemblies.Count; i++)
             {
                 if (
                     !dotNetFrameworkReferenceAssemblies[i].EndsWith(
                         Path.DirectorySeparatorChar.ToString(),
```
danmosemsft: There is an unnecessary allocation of a string here. The fix is 
` !dotNetFrameworkReferenceAssemblies[i].EndsWith(Path.DirectorySeparatorChar)`
Which is an ordinal comparison anyway.

AntonLapounov: The `char` overload was added in .NET Core 2.0; it won't compile for the desktop .NET Framework.

danmosemsft: Oh yes, I see I signed off 😃 

Well it can just do the same thing - check length then use indexer.

rainersigwald: We have a centralized mechanism--`FileUtilities.EndsWithSlash()`.

danmosemsft:
@rainiersigwald hard to answer without knowing the bug bar. I guess its good to fix race conditions when they are found because they are unpredictable in their impact.in this case the risk seems very low.

src/Utilities/ToolLocationHelper.cs
AntonLapounov: @rainersigwald `FileUtilities.EnsureTrailingSlash` is even more suitable here, however, both of them slightly change the semantics since they also check for `Path.AltDirectorySeparatorChar` at the end of the path.  Any semantic changes should be done separately.

AntonLapounov:
Thank you for reviewing this so quickly!  Anyone with write access — please merge this fix.  Our compiler check-in system has been intermittently hitting this race condition for many months now.

rainersigwald:
Oops, I should have put an explicit waiting on this--I think we should have taken it for 15.9. Oh well, can always backport.

cdmihai:
Going to cherry pick it into 15.9stg

cdmihai:
https://github.com/Microsoft/msbuild/pull/3543

src/Utilities/ToolLocationHelper.cs
danmosemsft: https://github.com/Microsoft/msbuild/issues/3544

PR #3540, opened 2018-07-25
Normalize job names

cdmihai:
Noticed these were different.


cdmihai:
> so much copypasta here

Maybe we should put the commandline as the name :)

PR #3550, opened 2018-07-26
Report TFM via GetTargetFrameworks if TF undefined

### LINKED ISSUE #3546: msbuild should pass TFM for classic csproj project to GetReferenceNearestTargetFrameworkTask
after talking with @rainersigwald, we realized that when msbuild invokes NuGet's `GetReferenceNearestTargetFrameworkTask`, it does not pass the tfm from a referenced project if it is a classic csproj.

code in question- https://github.com/Microsoft/msbuild/blob/51f53ebf530612beede45d13c535152785e23bdd/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1700-L1713

rainersigwald:
Prior to this change, ProjectReferences to non-SDK projects did not
report a `TargetFrameworks` list to the referencing project because
prior to the .NET Core SDK, the TFM for a project was defined in
`$(TargetFrameworkMoniker)` as the long form of the moniker.

Fixes #3546 by falling back to the legacy TFM in the included-in-all-
projects copy of `GetTargetFrameworks`.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1570,6 +1570,25 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <!--
     ====================================================================================
+                                        _ComputeReferringTargetFramework
+
+    Select the moniker to use for computing project compatibility if not already set. NugetTargetMoniker (NTM) is preferred by default over
+    TargetFrameworkMoniker (TFM) because it is required to disambiguate the UWP case where TFM is fixed at .NETCore,Version=v5.0 and
+    has floating NTM=UAP,Version=vX.Y.Z. However, in other cases (classic PCLs), NTM contains multiple values and that will cause the MSBuild
+    invocation below to fail by passing invalid properties. Therefore we do not use the NTM if it contains a semicolon.
+
+    `ReferringTargetFrameworkForProjectReferences` is no longer a good name, but may be consumed by other projects so should be preserved.
+    ======================================================================================
+  -->
+  <Target Name="_ComputeReferringTargetFramework">
+    <PropertyGroup Condition="'$(ReferringTargetFrameworkForProjectReferences)' == ''">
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1703,6 +1703,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <_ThisProjectBuildMetadata Include="$(MSBuildProjectFullPath)">
         <TargetFrameworks Condition="'$(TargetFrameworks)' != ''">$(TargetFrameworks)</TargetFrameworks>
         <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">$(TargetFramework)</TargetFrameworks>
+        <TargetFrameworks Condition="'$(TargetFrameworks)' == '' and '$(TargetFramework)' == ''">$(TargetFrameworkMoniker)</TargetFrameworks>
```
nguerrera: This probably needs to respect NuGetTargetMoniker for UWP at least where it is distinct from TargetFrameworkMoniker.  But see weirdness with old school PCL in ReferrringTargetFrameworkMonikerForProjectReferences (not sure if I got the exact mouthful right from memory).

rainersigwald: Nice catch.

https://github.com/Microsoft/msbuild/blob/d2e8794ee2d6c2002a248d70ee315fc8380d82e5/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1583-L1593

nguerrera:
I forget why we aren't doing this...

rainersigwald:
Oh, I didn't mean to close this, it auto closed when I deleted the target branch. Looks like I can't reopen it, so I'll have to open a new PR.

Current status is that I'm pretty sure this breaks legacy VS projects, but need to double-check.

nguerrera:
OK, I remember you said it might not be compatible, but I didn't remember a final conclusion. Makes sense now.

rainersigwald:
Opened #3620 so I don't forget about this.

Issue #3552, opened 2018-07-27
OutOfProcProjectInstanceBasedBuildDoesNotReloadFromDisk intermittently fails


rainersigwald:
```
Assert.Contains() Failure
Not found: The target \"Foo\" does not exist in the project
In value: Build started.
Evaluation started (\"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\")
Importing project \"/tmp/c5jpucpx.hx4/Temporarybcb690679dc64c70b13f818c69f99224/import\" into project \"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\" at (5,3).
Evaluation finished (\"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\")
Project \"tmpUhzOjt.tmp\" (Foo target(s)):
Building with tools version \"15.0\".
Target \"Foo\" in file \"/tmp/c5jpucpx.hx4/Temporarybcb690679dc64c70b13f818c69f99224/import\" from project \"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\" (entry point):
Using \"Message\" task from assembly \"Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a\".
Task \"Message\"
Bar
Done executing task \"Message\".
Done building target \"Foo\" in project \"tmpUhzOjt.tmp\".
Done building project \"tmpUhzOjt.tmp\".
Build succeeded.
Build started.
Evaluation started (\"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\")
Project \"/tmp/c5jpucpx.hx4/Temporarybcb690679dc64c70b13f818c69f99224/import\" was not imported by \"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\" at (5,3), due to false condition; ('$(ImportIt)' == 'true') was evaluated as ('false' == 'true').
Evaluation finished (\"/tmp/c5jpucpx.hx4/tmpUhzOjt.tmp\")
Project \"tmpUhzOjt.tmp\" (Foo target(s)):
Building with tools version \"15.0\".
```

From, for example, https://dotnet.visualstudio.com/public/_build/results?buildId=8373&view=ms.vss-test-web.test-result-details

I'm going to disable for now since this seems to be our flakiest test.

rainersigwald:
[...]

PR #3553, opened 2018-07-27
WIP: Disable OutOfProcProjectInstanceBasedBuildDoesNotReloadFromDisk

rainersigwald:
It fails fairly often in VSTS CI builds. Works around #3552 until we can root-cause it.


rainersigwald:
I think #3554 may actually fix the bug and not need this. Running tests on it a few more times.

rainersigwald:
The attempted real fix looks good enough and I'm feeling optimistic today so let's do that.

PR #3566, opened 2018-07-31
Expand help text for restore command line parameter

### LINKED ISSUE #3561: Expand help text for /restore to include cache clearing
The current help text for `/restore` reads:

```
  /restore[:True|False]
                     Runs a target named Restore prior to building
                     other targets.  This is useful when your project
                     tree requires packages to be restored before they
                     can be built. Specifying /restore is the same as
                     specifying /restore:True.  Use the parameter to
                     override a value that comes from a response file.
                     (Short form: /r)
```

This help text doesn't make it clear that `/r` with a `Restore` target should be used over constructs like `DependsOnTargets="Restore"`.

Please find a few words that make users use this switch instead. It doesn't have to go into details of the cache mechanisms, just something like:
`Use this switch to ensure the build uses the output generated by the Restore target`

dasMulli:
Fixes #3561

### DIFF
--- a/src/MSBuild/Resources/Strings.resx
+++ b/src/MSBuild/Resources/Strings.resx
@@ -668,11 +668,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <data name="HelpMessage_31_RestoreSwitch" UESanitized="false" Visibility="Public">
     <value>  /restore[:True|False]
                      Runs a target named Restore prior to building
-                     other targets.  This is useful when your project
-                     tree requires packages to be restored before they
-                     can be built. Specifying /restore is the same as
-                     specifying /restore:True.  Use the parameter to
-                     override a value that comes from a response file.
+                     other targets and ensures the build for these
+                     targets uses the latest restored build logic.
+                     This is useful when your project tree requires
+                     packages to be restored before it can be built.
+                     Specifying /restore is the same as specifying
+                     /restore:True. Use the parameter to override
+                     a value that comes from a response file.
                      (Short form: /r)
     </value>
     <comment>
--- a/src/MSBuild/Resources/xlf/Strings.cs.xlf
+++ b/src/MSBuild/Resources/xlf/Strings.cs.xlf
@@ -1353,14 +1353,16 @@ Copyright (C) Microsoft Corporation. Všechna práva vyhrazena.
       <trans-unit id="HelpMessage_31_RestoreSwitch">
         <source>  /restore[:True|False]
                      Runs a target named Restore prior to building
-                     other targets.  This is useful when your project
-                     tree requires packages to be restored before they
-                     can be built. Specifying /restore is the same as
-                     specifying /restore:True.  Use the parameter to
-                     override a value that comes from a response file.
+                     other targets and ensures the build for these
+                     targets uses the latest restored build logic.
+                     This is useful when your project tree requires
+                     packages to be restored before it can be built.
+                     Specifying /restore is the same as specifying
+                     /restore:True. Use the parameter to override
+                     a value that comes from a response file.
                      (Short form: /r)
     </source>
-        <target state="translated">  /restore[:True|False]
+        <target state="needs-review-translation">  /restore[:True|False]
                      Před sestavením jiných cílů spustí cíl s názvem
                      Restore. Je to užitečné, když váš strom
                      projektů vyžaduje obnovení balíčků, aby je bylo možné
--- a/src/MSBuild/Resources/xlf/Strings.de.xlf
+++ b/src/MSBuild/Resources/xlf/Strings.de.xlf
@@ -1348,14 +1348,16 @@ Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.
       <trans-unit id="HelpMessage_31_RestoreSwitch">
         <source>  /restore[:True|False]
                      Runs a target named Restore prior to building
-                     other targets.  This is useful when your project
-                     tree requires packages to be restored before they
-                     can be built. Specifying /restore is the same as
-                     specifying /restore:True.  Use the parameter to
-                     override a value that comes from a response file.
+                     other targets and ensures the build for these
+                     targets uses the latest restored build logic.
+                     This is useful when your project tree requires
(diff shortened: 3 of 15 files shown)


src/MSBuild/Resources/xlf/Strings.cs.xlf
```diff
@@ -1393,13 +1395,13 @@ Copyright (C) Microsoft Corporation. Všechna práva vyhrazena.
     </note>
       </trans-unit>
       <trans-unit id="HelpMessage_32_ProfilerSwitch">
-        <source>  /profileevaluation:&lt;file&gt;    
-                     Profiles MSBuild evaluation and writes the result
```
dasMulli: This whitespace-only change happens during the build when it updates the XLF files.. I suppose I should keep it in bc it needs to be done at some point?

rainersigwald: It looks like your editor stripped some trailing whitespace when you edited the `.resx` from which these are updated, and the tooling faithfully copied that whitespace difference here. I think that's fine--we shouldn't have the trailing whitespace, and you're generating real diffs for these files anyway.

@cdmihai Do you think we should avoid this? I assume the translation team will see the whitespace change and re-mark these as "translated", but maybe we should do that for them?

src/MSBuild/Resources/Strings.resx
```diff
@@ -668,11 +668,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <data name="HelpMessage_31_RestoreSwitch" UESanitized="false" Visibility="Public">
     <value>  /restore[:True|False]
                      Runs a target named Restore prior to building
-                     other targets.  This is useful when your project
-                     tree requires packages to be restored before they
-                     can be built. Specifying /restore is the same as
-                     specifying /restore:True.  Use the parameter to
-                     override a value that comes from a response file.
+                     other targets and ensures the build uses the latest
+                     restore output instead of possibly cached output.
```
rainersigwald: Maybe 

> ensures the build uses the latest restored build logic.

? But I'm not sure that's actually clearer.

src/MSBuild/Resources/xlf/Strings.cs.xlf
dasMulli: I'll try to be more careful on the next iteration since I assume there are going to be phrasing changes.

src/MSBuild/Resources/Strings.resx
dasMulli: I think it's more specific ("correct") but maybe not more helpful..

jeffkl: Maybe something like

```
  /restore[:True|False]
                     Runs a target named Restore, resets the context,
                     and then runs other targets.  This is useful when
                     your project tree requires packages to be
                     restored before they can be built. Specifying
                     /restore is the same as specifying /restore:True.
                     Use the parameter to override a value that comes
                     from a response file.
                     (Short form: /r)
```

dasMulli: > `resets the context`

I agree that this is more "correct".
However, my goal was to explicitly tell users that they should use it to get correct restore+build behavior.

Any ideas how to address both concerns?

```

```

Btw I also changed `… before they can be built.` => `... before it can be built.` in this PR since `project tree` is singular.

Mikejo5000: LG2M

dasMulli:
Take two.
@rainersigwald @jeffkl @Mikejo5000 what do you think?

Mikejo5000:
@dasMulli There's a typo: "thes".
"ensures the build for thes" 
-->
"ensures that the build for these..."

dasMulli:
@Mikejo5000 thanks!! updated

rainersigwald:
VSTS PR job failed with this on the macOS leg

```
2018-08-01T20:24:54.1813720Z . 'build/cibuild.sh'
2018-08-01T20:24:54.1941120Z [command]/bin/bash --noprofile --norc /Users/vsts/agent/2.136.1/work/_temp/1ac41118-1a00-4a64-87b9-0a76303593de.sh
2018-08-01T20:24:54.1956860Z Executed from /Users/vsts/agent/2.136.1/work/_temp/1ac41118-1a00-4a64-87b9-0a76303593de.sh with arguments: -build -ci -prepareMachine
2018-08-01T20:24:54.5589490Z dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300/dotnet-sdk-2.1.300-osx-x64.tar.gz
2018-08-01T20:24:56.9263070Z curl: (56) LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 54
2018-08-01T20:24:56.9523080Z dotnet-install: Cannot download: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300/dotnet-sdk-2.1.300-osx-x64.tar.gz
2018-08-01T20:24:56.9573210Z dotnet-install: Downloading legacy link: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300/dotnet-dev-osx-x64.2.1.300.tar.gz
2018-08-01T20:24:57.3566620Z curl: (22) The requested URL returned error: 404 
2018-08-01T20:24:57.3581010Z dotnet_install: Error: Could not download .NET Core SDK version 2.1.300
2018-08-01T20:24:57.3594580Z Failed to install stage0
2018-08-01T20:24:57.3608240Z Using dotnet from: /Users/vsts/agent/2.136.1/work/1/s/build/../artifacts/.dotnet/2.1.300
2018-08-01T20:24:57.3650190Z /Users/vsts/agent/2.136.1/work/1/s/build/build.sh: line 355: /Users/vsts/agent/2.136.1/work/1/s/build/../artifacts/.dotnet/2.1.300/dotnet: No such file or directory
2018-08-01T20:24:57.3664160Z Failed to clear NuGet cache
2018-08-01T20:24:57.3800430Z ##[error]Bash exited with code '127'.
```

That doesn't look related to this change and Jenkins/macOS passed, so I'm going to merge anyway.

rainersigwald:
Thanks, @dasMulli!

PR #3568, opened 2018-08-01
Normalize some resx and xlf files for editing in VS Code

dasMulli:
cc @rainersigwald @jeffkl


dasMulli:
This may require updating strings in the tooling used to process xlf files

dasMulli:
No, only these. but afaik the trailing newlines are added in the loc PRs, but it didn't occur to me that they are also added back by vs code - provided you open and edit an xlf file (if you change the translation state or want to contribute a translation).

rainersigwald:
Out of an abundance of caution, let's wait for #3573 before merging this (I'm pretty sure there are no conflicts but just in case).

PR #3576, opened 2018-08-02
Log full exception string during ToolTask exceptions

### LINKED ISSUE #911: Exceptions thrown by ToolTask aren't diagnosable
We got a report of an internal build that failed with this error:

```
Microsoft.CppCommon.targets(356,5): error MSB6003: The specified task executable "CL.EXE" could not be run. Access to the path 'F:\REDACTED\x86\CL.read.1.tlog' is denied.
```

That's thrown from [`ToolTask`](https://github.com/Microsoft/msbuild/blob/204cfd215fdc1c92322b0b22165cc8c4c3259e02/src/Utilities/ToolTask.cs#L1676), which uses this method:

``` C#
/// <summary>
/// This method takes in an exception and if MSBuildDiagnostics is set then it will display the stack trace
/// if it is not set only the message will be displayed, this is to fix the problem where the user was getting
/// stack trace when a shorter message was better
/// </summary>
/// <returns>exception message</returns>
private string GetErrorMessageWithDiagnosticsCheck(Exception e)
{
    // If MSBuildDiagnostics is set show stack trace information
    if (Environment.GetEnvironmentVariable("MSBuildDiagnostics") != null)
    {
        // Includes stack trace
        return e.ToString();
    }
    else
    {
        // does not include stack trace
        return e.Message;
    }
}
```

This was clearly intentionally added (it looks like it was in the Orcas/VS2008 time frame), but it seems like a bad idea--it makes the error above basically undiagnosable, because you have no idea what's complaining about permissions.

MSBuild should at least log the stack to the higher logging verbosities, though it might be better to just roll back to including the whole stack trace all the time.

jeffkl:
Fixes #911

### DIFF
--- a/src/Utilities/ToolTask.cs
+++ b/src/Utilities/ToolTask.cs
@@ -1495,31 +1495,15 @@ namespace Microsoft.Build.Utilities
             {
                 if (!_terminatedTool)
                 {
-                    LogPrivate.LogErrorWithCodeFromResources("General.InvalidToolSwitch", ToolExe, GetErrorMessageWithDiagnosticsCheck(e));
+                    LogPrivate.LogErrorWithCodeFromResources("General.InvalidToolSwitch", ToolExe, e.ToString());
                 }
                 return false;
             }
-            catch (Win32Exception e)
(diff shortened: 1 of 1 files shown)


AndyGerlicher:

PR #3579, opened 2018-08-03
Always throttle log message handling

### LINKED ISSUE #3577: OOM when generating binlog for boost.hana
It reproes using 15.8 preview 6 (I didn't try older versions).
    Microsoft (R) Build Engine version 15.8.166+gd4e8d81a88 for .NET Framework

1. Clone https://github.com/xiangfan-ms/hana
2. Create build folder and enter build folder
2. Generate the solution (-DBOOST_ROOT may be optional to repro OOM)
```
cmake .. -DBOOST_ROOT=[path_to_boost_167] -D"CMAKE_CXX_FLAGS=/EHsc /Zc:ternary /bigobj /we4667" -DBOOST_HANA_ENABLE_STRING_UDL=false
```
4. Build the solution
```
msbuild /noconlog /t:Rebuild /p:Configuration=Debug /p:Platform="Win32" /m /bl:hana.binlog;ProjectImports=None Boost.Hana.sln /flp:v=n;logfile=hana.log

Unhandled Exception: OutOfMemoryException.
```

It doesn't OOM if I don't generate binlog. According to the log file, the OOM happens at the very early stage of the build.

rainersigwald:
Fixes #3577 by applying the throttling policy to the processing of all
logging events, not just those from other nodes.

While this will slow the build down, it will keep the memory usage of
log events in the to-be-processed queue bounded, preventing the OOM
reported in the bug.

### DIFF
--- a/src/Build.UnitTests/BackEnd/LoggingServicesLogMethod_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/LoggingServicesLogMethod_Tests.cs
@@ -1772,7 +1772,7 @@ namespace Microsoft.Build.UnitTests.Logging
             /// Override the method to log which event was processed so it can be verified in a test
             /// </summary>
             /// <param name="buildEvent">Build event which was asked to be processed</param>
-            internal override void ProcessLoggingEvent(object buildEvent, bool allowThrottling = false)
+            internal override void ProcessLoggingEvent(object buildEvent)
             {
                 if (buildEvent is BuildEventArgs)
                 {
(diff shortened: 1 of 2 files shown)


rainersigwald:
I don't know why yet, but this is incomplete. So far, it looks like the queue fills and then we hit a deadlock of some sort.

rainersigwald:
This is sufficiently broken by https://github.com/Microsoft/msbuild/issues/3577#issuecomment-410805588 that I'm just going to close it and rethink.

PR #3593, opened 2018-08-08
Introduce MSBuildLastModifiedProject

### LINKED ISSUE #3591: Automatically prepend path of the project file with the latest timestamp
The current idiom of putting all project files in `$(MSBuildAllFiles)` has [perf implications](https://github.com/Microsoft/msbuild/issues/3004). Since it's used for incrementality checks, we can provide a reserved property like `MSBuildLatestProject` which avoids appending.

An alternative would be to [compute](https://github.com/Microsoft/msbuild/issues/1299) `MSBuildAllFiles`. This has complications though because the c++ sdks appends more files to it than just msbuild project files.

jeffkl:
Represents the project in the import graph with the most recent last modified time.  This is a step towards deprecating `MSBuildAllProjects`.

If the project is an in-memory project but has no imports, the property is not set.  If the project is an in-memory project and has imports, the property does not include the project itself.

Closes #3591

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -200,6 +200,11 @@ namespace Microsoft.Build.Evaluation
         private readonly EvaluationProfiler _evaluationProfiler;
 
         /// <summary>
+        /// Keeps track of the project that is newest between the project and all imports.
+        /// </summary>
+        private ProjectRootElement _newestProject;
+
+        /// <summary>
         /// Private constructor called by the static Evaluate method.
         /// </summary>
         private Evaluator(
@@ -245,6 +250,13 @@ namespace Microsoft.Build.Evaluation
             _sdkResolverService = sdkResolverService;
             _submissionId = submissionId;
             _evaluationProfiler = new EvaluationProfiler(profileEvaluation);
+
+            // Only track the newest project for on disk projects.  There's no way to guarantee the MSBuildNewestProject property
+            // is accurate for in-memory projects
+            if (projectRootElement.FullPath != null)
+            {
+                _newestProject = projectRootElement;
+            }
         }
 
         /// <summary>
@@ -715,6 +727,11 @@ namespace Microsoft.Build.Evaluation
                 using (_evaluationProfiler.TrackPass(EvaluationPass.Properties))
                 {
                     PerformDepthFirstPass(_projectRootElement);
+                }
+
+                if (_newestProject != null)
+                {
+                    _data.SetProperty(ReservedPropertyNames.newestProject, _newestProject.FullPath, isGlobalProperty: false, mayBeReserved: true);
                 }
 
                 List<string> initialTargets = new List<string>(_initialTargetsList.Count);
@@ -2351,6 +2368,11 @@ namespace Microsoft.Build.Evaluation
                         else
                         {
                             imports.Add(importedProjectElement);
+
+                            if (_newestProject != null && importedProjectElement.LastWriteTimeWhenRead > _newestProject.LastWriteTimeWhenRead)
+                            {
+                                _newestProject = importedProjectElement;
+                            }
 
                             if (_logProjectImportedEvents)
                             {
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -200,9 +200,9 @@ namespace Microsoft.Build.Evaluation
         private readonly EvaluationProfiler _evaluationProfiler;
 
         /// <summary>
-        /// Keeps track of the project that is newest between the project and all imports.
+        /// Keeps track of the project that is last modified of the project and all imports.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2217,7 +2217,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     ====================================================================================================
   -->
   <Target Name="GenerateBindingRedirects"
-    Inputs="$(MSBuildAllProjects);@(AppConfigFile);$(ResolveAssemblyReferencesStateFile);$(IntermediateOutputPath);@(SuggestedBindingRedirects)"
+    Inputs="$(MSBuildNewestProject);$(MSBuildAllProjects);@(AppConfigFile);$(ResolveAssemblyReferencesStateFile);$(IntermediateOutputPath);@(SuggestedBindingRedirects)"
```
rainersigwald: For targets we control, do we need to pass `$(MSBuildNewestProject);$(MSBuildAllProjects)`? That's what I'd expect for someone who wanted to support MSBuild 15 and MSBuild 16, but in our targets where we'll only run on 16+ I think we should switch instead of passing an additional file. That way we get the benefit of not having to `stat` all the project files.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -245,6 +250,13 @@ internal class Evaluator<P, I, M, D>
             _sdkResolverService = sdkResolverService;
             _submissionId = submissionId;
             _evaluationProfiler = new EvaluationProfiler(profileEvaluation);
+
+            // Only track the newest project for on disk projects.  There's no way to guarantee the MSBuildNewestProject property
+            // is accurate for in-memory projects
```
rainersigwald: I don't think this is right. It would be more compatible for the implementation to be defined as "newest imported file that appears on disk", since that's what you'd get out of `$(MSBuildAllProjects)`.

AndyGerlicher:
How about `MSBuildLatestProject` instead of Newest?

rainersigwald:
`MSBuildLastModifiedImport`?

src/Tasks/Microsoft.Common.CurrentVersion.targets
AndyGerlicher: I agree, we should never use `MSBuildAllProjects` as an Input in our targets that ship with MSBuild again. Setting it is a different story, but shouldn't consume it.

radical:
What is this useful for? Explanation of the reasoning for this in the commit message would be helpful. (I did try to follow the issues, but I'm not very clear. Explaining here makes sense, IMHO).

jeffkl:
@radical please see https://github.com/Microsoft/msbuild/issue_comments#issuecomment-411782588

AndyGerlicher:
`MSBuildLastModifiedProject`? Keep project in the name since it's used in the existing and other areas?

radical:
@jeffkl Thanks! IMHO, that or similar explanation should be included in the commit messages.

cdmihai:
Can you add a test that imports from several files and asserts the property points to the correct one?

src/Tasks/Microsoft.Common.CurrentVersion.targets
jeffkl: After discussing this, it might better to not set it and leave it in the `Inputs`.  That way if other people are adding to it we won't break them.  They would need to remove the instance of setting the `MSBuildAllProjects` property if its just an import but if its not an import the behavior would stay the same.

jeffkl:
We need to decide what to do with existing MSBuildAllProjects:

1. Remove all places where we set it and remove it from Inputs for our targets
  Helps move us deprecate it, could cause a lot of breaks
2. Leave our places where we set it, remove it from Inputs for our targets.
  Helps 3rd party targets until we update them, does not address the memory usage
3. Remove all places where we set it, leave it as an Input for our targets
  Helps reduce memory use, helps with back compat, doesn't deprecate the property

cdmihai:
I'd go with option 1 :)

Users that add additional files to it, and have targets that consume MSBuildAllProjects for incrementality, will continue to work if they only care about their files only.

To help with back compat, if MSBuildAllProjects is not set, we could set it to MSBuildLastModifiedProject.

Nirmal4G:
It could contain multiple target files, if that's the case then the name should be plural `MSBuildLastModifiedProjects`.

dasMulli:
@Nirmal4G no, the idea is to no longer contain multiple projects in there.
`MSBuildAllProjects` is used by most targets to track incremental compile input - to be able to detect project modifications and to some extent scope changes (e.g. after a NuGet restore that changes the imported obj\*.nuget.props|targets files).
The idea is to contain a single target in the "last modified" property implemented here instead so MSBuild doesn't have to check for all the files contain there, and most importantly remove the allocations involved in creating and making that property available.

jeffkl:
We're going a different route, so closing this PR and opening a new one

PR #3598, opened 2018-08-09
Target net472

cdmihai:


cdmihai:
@crummel, look ok for sourcebuild?

cdmihai:
@mmitche, are there any pools with >dev15.7 and net472? We'll need net472 starting with this PR.

crummel:
Looks good for source-build, thanks for pinging me!

netci.groovy
```diff
@@ -107,7 +107,7 @@ def CreateJob(script, runtime, osName, isPR, machineAffinityOverride = null, sho
                     }
 
                     // this agent has VS 15.7 on it, which is a min requirement for our repo now.
-                    machineAffinityOverride = 'Windows.10.Amd64.ClientRS3.DevEx.Open'
+                    machineAffinityOverride = 'windows.10.amd64.clientrs4.devex.15.8.open '
```
rainersigwald: We could also just drop Jenkins now, unless anyone would like to complain about VSTS?

src/MSBuild/app.amd64.config
```diff
@@ -46,18 +46,6 @@
           <codeBase version="15.1.0.0" href="..\Microsoft.Build.Conversion.Core.dll"/>
         </dependentAssembly>
 
-        <!-- Redirects for facade assemblies -->
```
rainersigwald: Do we really not need these any more? I think I expected them to remain (maybe with an updated version) with no `codeBase`.

src/Package/MSBuild.Engine.Corext/MsBuild.Engine.Corext.nuspec
```diff
@@ -29,8 +29,6 @@
     <file src="$X86BinPath$/Microsoft.Build.Tasks.Core.dll" target="v15.0/bin" />
     <file src="$X86BinPath$/Microsoft.Build.Utilities.Core.dll" target="v15.0/bin" />
     <file src="$X86BinPath$/System.Collections.Immutable.dll" target="v15.0/bin" />
-    <file src="$X86BinPath$/System.IO.Compression.dll" target="v15.0/bin" />
-    <file src="$X86BinPath$/System.Runtime.InteropServices.RuntimeInformation.dll" target="v15.0/bin" />
```
rainersigwald: This has implications for internal consumers, right? Has the internal cloud builder image moved to 4.7.2?

cdmihai:
@dotnet-bot test this please

netci.groovy
cdmihai: I'd be in favor of that. The rhel 7.2 job is the only one left that needs porting. Do you know what queue to use for rhel 7.2 machines?

src/Package/MSBuild.Engine.Corext/MsBuild.Engine.Corext.nuspec
cdmihai: Last time we talked they seemed OK with it.

src/MSBuild/app.amd64.config
cdmihai: Very good question. Those types moved in-box, so I figured that if VS also moves to 472, they'll also start using the in-box types, in which case the redirects are not needed anymore. Is this a valid assumption?

src/Directory.Build.props
```diff
@@ -23,7 +23,8 @@
     <Platforms>AnyCPU;x64</Platforms>
     
     <!-- Defaults for target frameworks and architecture -->
-    <LibraryTargetFrameworks>$(FullFrameworkTFM);netstandard2.0</LibraryTargetFrameworks>
```
cdmihai: @crummel is it OK if msbuild does not build full framework on !windows?

crummel: No, I think that will cause problems - SDK builds on the full framework to support using VS Code with a source-built SDK (issues https://github.com/dotnet/source-build/issues/125 and https://github.com/dotnet/sdk/issues/2249) and has a dependency on MSBuild, so we'll need the full-framework here too.  Would something like an `[En|Dis]ableFullFrameworkBuild` property that defaults to what you need and we can override make sense here?

cdmihai: OK, added it back :)

cdmihai:
@dotnet-bot test Windows_NT Build for CoreCLR
Previous: https://ci2.dot.net/job/Microsoft_msbuild/job/master/job/innerloop_Windows_NT_CoreCLR_prtest/895/

src/Directory.Build.props
crummel: Thanks!

PR #3605, opened 2018-08-14
Prepend the last modified project to MSBuildAllProjects

### LINKED ISSUE #1299: Make MSBuildAllProjects magically append on import

jeffkl:
During evaluation, keep track of the last modified project in the import graph.
Prepend the path to the last modified project to `MSBuildAllProjects`
Stop setting `MSBuildAllProjects` in our targets
Add unit tests

Closes #1299

### DIFF
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4444,10 +4444,14 @@ namespace Microsoft.Build.UnitTests.Evaluation
                 File.SetLastWriteTime(project2.ProjectFile, DateTime.Now);
                 File.SetLastWriteTime(primaryProject.ProjectFile, DateTime.Now.AddHours(-1));
 
-
                 Project project = new Project(primaryProject.ProjectFile, null, null);
 
-                project.GetPropertyValue(Constants.MSBuildAllProjectsPropertyName).ShouldStartWith(project2.ProjectFile);
+                string propertyValue = project.GetPropertyValue(Constants.MSBuildAllProjectsPropertyName);
+
+                propertyValue.ShouldStartWith(project2.ProjectFile);
+
+                propertyValue.ShouldNotContain(primaryProject.ProjectFile, Case.Insensitive);
+                propertyValue.ShouldNotContain(project1.ProjectFile, Case.Insensitive);
             }
         }
 
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -18,6 +18,7 @@ using Microsoft.Build.Framework;
 using Microsoft.Build.Internal;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Shared.FileSystem;
+using Shouldly;
 using Xunit;
 
 using InvalidProjectFileException = Microsoft.Build.Exceptions.InvalidProjectFileException;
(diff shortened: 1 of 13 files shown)


src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -1997,7 +1998,7 @@ public void AllEvaluatedPropertiesAndImports()
                 // the whole list.
                 // We have to dump it into a dictionary because AllEvaluatedProperties contains duplicates, but we're preparing to Properties,
                 // which doesn't, so we need to make sure that the final value in AllEvaluatedProperties is the one that matches.
-                foreach (ProjectProperty property in project.AllEvaluatedProperties.TakeWhile(property => property.Xml == null))
+                foreach (ProjectProperty property in project.AllEvaluatedProperties.Where(property => property.Xml == null))
```
rainersigwald: Why this change? If it's needed, please update the comment to remove the "don't need to scan the whole list" bit.

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4425,6 +4426,56 @@ public void ConditionWithShortCircuitAndErrorDoesNotFailBuild(string projectInne
...
+<Import Project=""{project1.ProjectFile}"" />
+<Import Project=""{project2.ProjectFile}"" />
+</Project>");
+
+                // Project1 and primary project last modified an hour ago, project2 is the newest
+                File.SetLastWriteTime(project1.ProjectFile, DateTime.Now.AddHours(-1));
+                File.SetLastWriteTime(project2.ProjectFile, DateTime.Now);
+                File.SetLastWriteTime(primaryProject.ProjectFile, DateTime.Now.AddHours(-1));
+
+
```
rainersigwald: nit: extra blank line

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4425,6 +4426,56 @@ public void ConditionWithShortCircuitAndErrorDoesNotFailBuild(string projectInne
...
+                Project project = new Project(primaryProject.ProjectFile, null, null);
+
+                project.GetPropertyValue(Constants.MSBuildAllProjectsPropertyName).ShouldStartWith(project2.ProjectFile);
```
rainersigwald: Validate no project1 path?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2725,6 +2743,25 @@ private static string StringifyList(IList<string> strings)
...
+        {
+            if (_lastModifiedProject != null)
+            {
+                P oldValue = _data.GetProperty(Constants.MSBuildAllProjectsPropertyName);
+
+                P newValue = _data.SetProperty(
+                    Constants.MSBuildAllProjectsPropertyName,
+                    oldValue == null
+                        ? _lastModifiedProject.FullPath
+                        : String.Format(CultureInfo.CurrentCulture, "{0};{1}", _lastModifiedProject.FullPath, oldValue.EvaluatedValue), isGlobalProperty: false, mayBeReserved: false);
```
rainersigwald: Do we want `CurrentCulture`? Shouldn't it be invariant?

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
jeffkl: This code assumes that all reserved properties are at the beginning of the list.  Originally I had added a new reserved property but it was at the end so this test broke.  Should I update the comment or revert this change?

src/Build/Evaluation/Evaluator.cs
jeffkl: All other places in this file use `CurrentCulture`...

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
rainersigwald: I'd just update the comment. It's test code; the perf benefit of an early stop doesn't strike me as worth much.

src/Build/Evaluation/Evaluator.cs
AndyGerlicher: I'd change it. And just use string interpolation rather than `string.Format`.

Issue #3586, opened 2018-08-07
Consider walking directories in up-to-date checks


rainersigwald:
@ccastanedaucf has observed (https://github.com/Microsoft/msbuild/pull/3547#issuecomment-410354029) significant time spent in up-to-date checks in https://github.com/Microsoft/msbuild/blob/680224d8ecf9a7da82c001d95db336fd9cfd12d8/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs#L937

Git for Windows got a big speedup by implementing some caching around checks for individual file timestamps (with the `core.fscache` setting). See https://github.com/git-for-windows/git/blob/master/compat/win32/fscache.c and its history.

The gist of the idea is that it's faster to enumerate all files in a directory and get every file's modification timestamp than it is to get the timestamps on an individual basis. Since there's probably a fairly high degree of overlap in the directories of files we're checking for timestamps, I suspect we could get some benefit from this technique too.

AndyGerlicher:
@ccastanedaucf can you take a look at implementing this? Might help get Windows closer to Linux after #3547 is in.

cdmihai:
Would be nice if you could implement it over the existing IFileSystem :)

Also, this might be a very good use case for the fast allocation free directory enumeration introduced by @JeremyKuhne  https://blogs.msdn.microsoft.com/jeremykuhne/2018/03/09/custom-directory-enumeration-in-net-core-2-1/

4creators:
The above problems are a huge burden in some dotnet repos see #2843

rainersigwald:
@4creators This will probably help with that, but I'd be shocked if it was a very high percentage of the overall incremental build time in those repos.

ccastanedaucf:
@cdmihai small bench comparing methods of getting last write time for 10 files on Windows. Right now we use the equivalent of ```File.GetLastWriteTimeUtc```.
```
                       Method |      Mean |     Error |    StdDev |
----------------------------- |----------:|----------:|----------:|
     File_GetLastWriteTimeUtc | 257.60 us | 0.7359 us | 0.5745 us |
 DirectoryInfo_EnumerateFiles |  53.60 us | 1.1738 us | 1.4416 us |
         FileSystemEnumerable |  30.44 us | 0.1749 us | 0.1550 us |
```

JeremyKuhne:
> small bench comparing methods of getting last write time for 10 files on Windows

[...]

Issue #2699, opened 2017-11-03
ResolveAssemblyReference may be improved by caching IO in "system" directories


cdmihai:
By changing the RAR IO cache to permanently cache certain files without any subsequent staleness checks improves RAR time by ~10% on a Roslyn build.
In my experiment, I permanently cached IO queries under `c:\Windows`, `c:\Program Files`, and `c:\Users\<username>\.nuget`. This reduced RAR time from 21.5 seconds to 19.6 seconds.

Next steps:
- [ ] instrument RAR to do a tally of all files it looks at, how many are shared between projects, how many come from directories that we know won't change during a build, etc

After figuring out the roots that won't change, the question is how to get this information into RAR. Probably an extra task argument.

The big risk here is if files under a permanent root do change over the lifetime of an MSBuild node with NodeReuse on. Highest chance of this happening is with the nuget cache. In this case, maybe nuget can write a hash at the root of its cache and update it whenever anything changed in the cache. Then RAR could define a couple of different cache policies per each root (e.g. `NeverCheck` for stuff under Windows or Program Files, `CompareRootHash` for the nuget cache, etc). But continuously checking that hash will reduce some of the perf gains.

danmosemsft:
Maybe something should be added to ITask_n_ that would allow tasks to subscribe to an event indicating the build completed. Assuming it would be sufficient to cache for the lifetime of a build.

cdmihai:
Can you please provide more details on the ITask build finished usage and benefits? The system file cache needs to be at least process wide to be useful, so I don't see how the information that other builds have finished under the current process benefits caching.

danmosemsft:
I was assuming that your cache would be in a static so it could be reused by all RAR invocations on the node during the build, then cleared at the end of the build (so the next build reusing the node would see any updates to whatever the folders are). If that's the case then there needs to be a mechanism to do that clearing. If there was an event on ITask that a task could subscribe to to know that the build was ending and the node was going dormant for reuse, it could use that opportunity to clear out that static. If any of the assumptions above are not correct then my comment isn't relevant :)

cdmihai:
Oh, I see, you're suggesting a way to clear the cache between builds when node reuse is on. Yes, it is actually a good solution. :) Tasks could register 'finalizers' to clear out state. I like it!

danmosemsft:
There may already be existing static fields in task classes that should be cleaned up in this manner.

rokonec:
[...]

PR #3608, opened 2018-08-14
Walk directories in up-to-date checks on Windows

ccastanedaucf:
Started with Git for Windows style caching as discussed (listed in the table as FS cache). Pros, very simple to implement and only needed to hook in a couple lines in the up-to-date method. Cons, doesn't perform as well due to hitting huge directories such as the temp folder in many targets, then proceeding to load up the cache with potentially thousands of unnecessary files. Since the cache is just scoped to the method, this ends up happening many, many times. Would perform much better if we had a cache persistent across up-to-date checks, but there's no good way to do invalidation yet. The test project ends up loading around 10k directories, but many of them are duplicates of previous target input/outputs.

Instead, grouping inputs and outputs by directory first, then walking each directory while checking if each walked file is in the set works much better. Still have to create a ton of strings to check hashtables, split then reconstruct paths, and walk unneeded files, but it allows short circuiting the walk when we've found everything in a directory and keeps the dictionary from blowing up in size. Lots of extra code right now since it doesn't fit well at all in the current  ```IsAnyOutOfDate()``` due to the different flow and Enumeration API, but shouldn't be difficult to just create an enumerator from the native calls on .NET Framework now that the logic is all there.

Also have memory usage concerns. Since Dictionary/HashSet can't use ```Span<Char>``` or ```Memory<Char>``` as keys, in both cases there's a good number of extra strings being created compared to before.

.NET Core build times below, averaged over 6 runs each:

| WebLarge Build | GetFileAttributesEx  | System.IO.Enumeration w/ local-scope FS cache | System.IO.Enumeration w/ local-scope input/output cache| 
| ------------- | ------------- | ------------- | ------------- |
| Incremental no-op  | 17527 ms | 16407 ms | 15546 ms |
| Incremental no-op parallel  | 6601 ms | 6199 ms | 5739 ms |

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -1098,9 +1098,17 @@ namespace Microsoft.Build.BackEnd
             // PERF -- we could change this to ensure that we walk the shortest list first (because we walk that one entirely): 
             //         possibly the outputs list isn't actually the shortest list. However it always is the shortest
             //         in the cases I've seen, and adding this optimization would make the code hard to read.
-            (string filePath, DateTime fileTime) oldestOutput = GetOldestFile(projectDirectory, outputs);
-            (string filePath, DateTime fileTime) newerInput = GetNewerFile(projectDirectory, inputs, oldestOutput.fileTime);
-            dependencyAnalysisDetailEntry = CompareInputAndOutput(newerInput, oldestOutput);
+            FileDateInfo candidateNewerInput = GetArbitraryFile(projectDirectory, inputs);
+            FileDateInfo candidateOlderOutput = GetFileWithTimeCondition(projectDirectory, outputs, candidateNewerInput.Time, FileTimeCondition.FindOlder);
+            dependencyAnalysisDetailEntry = CompareInputAndOutput(candidateNewerInput, candidateOlderOutput);
+            bool shouldSearchAllInputs = dependencyAnalysisDetailEntry == null;
+
+            if (shouldSearchAllInputs)
+            {
+                candidateNewerInput = GetFileWithTimeCondition(projectDirectory, inputs, candidateOlderOutput.Time, FileTimeCondition.FindNewer);
+                dependencyAnalysisDetailEntry = CompareInputAndOutput(candidateNewerInput, candidateOlderOutput);
+
+            }
             return dependencyAnalysisDetailEntry != null;
         }
 
@@ -1112,11 +1120,11 @@ namespace Microsoft.Build.BackEnd
             }
         }
 
-        private static DependencyAnalysisLogDetail CompareInputAndOutput((string filePath, DateTime fileTime) newerInput, (string filePath, DateTime fileTime) oldestOutput)
+        private static DependencyAnalysisLogDetail CompareInputAndOutput(FileDateInfo newerInput, FileDateInfo olderOutput)
         {
-            bool isMissingOutput = IsInvalidFileTime(oldestOutput.fileTime);
-            bool isMissingInput = IsInvalidFileTime(newerInput.fileTime);
-            bool isTargetUpToDate = !isMissingOutput && !isMissingInput && newerInput.fileTime <= oldestOutput.fileTime;
+            bool isMissingOutput = IsInvalidFileTime(olderOutput.Time);
+            bool isMissingInput = IsInvalidFileTime(newerInput.Time);
+            bool isTargetUpToDate = !isMissingOutput && !isMissingInput && newerInput.Time <= olderOutput.Time;
 
             // All exist and no inputs are newer than any outputs; up to date
             if (isTargetUpToDate)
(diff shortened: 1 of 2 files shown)


src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
```diff
@@ -1075,6 +1085,175 @@ internal static bool IsAnyOutOfDate<T>(out DependencyAnalysisLogDetail dependenc
             return false;
         }
 
+#if FEATURE_ENUMERATION
+        internal static bool IsAnyOutOfDateWip<T>(out DependencyAnalysisLogDetail dependencyAnalysisDetailEntry, string projectDirectory, IList<T> inputs, IList<T> outputs)
```
AndyGerlicher: These two (get oldest and get newer) are pretty similar. One way to perhaps short circuit and make these more generic is instead of finding the oldest output then checking to see if there's a newer input, pick an input at random (first) and use that as a candidate for the first iteration.

```C#
  var candidateNewest = intputs[0].DateTime;
  var newestOutput = FindFileOlderThan(outputs, candidateNewest, out bool done);
  if (!done)
  {
    FindFileNewerThan(inputs, newestOutput);
  }
```
Then `FindFileNewerThan` and `FindFileOlderThan` become the same method just with a flipped check. Plus there's a possibility that you happen to get an input that was really new (git branch change) and you stop super early.

src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
```diff
@@ -1075,6 +1085,175 @@ internal static bool IsAnyOutOfDate<T>(out DependencyAnalysisLogDetail dependenc
...
+                    }
+                    if (fileNamesToFind.Count == 0)
+                    {
+                        break;
+                    }
+                }
+
+                bool isMissingFile = fileNamesToFind.Count > 0;
+
+                if (isMissingFile)
```
AndyGerlicher: Not sure I follow what this is for. It's a file you were supposed to find but the enumeration missed it? Couldn't there be more than one of these?

src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
```diff
@@ -960,6 +963,13 @@ private static List<string> GetItemSpecsFromItemVectors(ItemVectorPartitionColle
         /// <returns>true, if any "input" is newer than any "output", or if any input or output does not exist.</returns>
         internal static bool IsAnyOutOfDate<T>(out DependencyAnalysisLogDetail dependencyAnalysisDetailEntry, string projectDirectory, IList<T> inputs, IList<T> outputs)
         {
+#if FEATURE_ENUMERATION
```
cdmihai: You could make this a Trait to make it easier to turn on via an environment variable

src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
```diff
@@ -1075,6 +1085,175 @@ internal static bool IsAnyOutOfDate<T>(out DependencyAnalysisLogDetail dependenc
...
+            ValidateTargetItemsType(outputs);
+
+            // Algorithm: walk through all the outputs to find the oldest output
+            //            walk through the inputs as far as we need to until we find one that's newer (if any)
+
+            // PERF -- we could change this to ensure that we walk the shortest list first (because we walk that one entirely): 
+            //         possibly the outputs list isn't actually the shortest list. However it always is the shortest
+            //         in the cases I've seen, and adding this optimization would make the code hard to read.
+            (string filePath, DateTime fileTime) oldestOutput = GetOldestFile(projectDirectory, outputs);
+            (string filePath, DateTime fileTime) newerInput = GetNewerFile(projectDirectory, inputs, oldestOutput.fileTime);
```
cdmihai: If there's potential overlap between input and output files (when they share the same directory), then it would be worth it to reuse data between the two methods

src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
```diff
@@ -1075,6 +1085,175 @@ internal static bool IsAnyOutOfDate<T>(out DependencyAnalysisLogDetail dependenc
...
+            IDictionary<string, ISet<string>> directoryToFileNames = MapDirectoryToFileNames(escapedFilePaths);
+
+            foreach (string directoryName in directoryToFileNames.Keys)
+            {
+                ISet<string> fileNamesToFind = directoryToFileNames[directoryName];
+                IEnumerable<(string fileName, DateTime fileTime)> files = GetFilesInDirectory(projectDirectory, directoryName);
+
+                foreach ((string fileName, DateTime fileTime) in files)
+                {
+                    bool isOlderFile = fileNamesToFind.Remove(fileName) && fileTime < oldestFileTime;
```
cdmihai: I think checking the set if the filename is there is cheaper than removing the file from the set

cdmihai: Ignore that, apparently Remove is o(1) :)

---
In reply to: [210367450](https://github.com/Microsoft/msbuild/pull/3608#discussion_r210367450) [](ancestors = 210367450)

4creators:
How about trying to devirtualize some calls? Perhaps using concrete implementations instead of interfaces as call sites could be an acceptable design trade off for increased performance - interesting how this would impact benchmarks.

4creators:
How about trying to devirtualize some calls? Perhaps using concrete implementations instead of interfaces as call sites could be an acceptable design trade off for increased performance - interesting how this would impact benchmarks.

ccastanedaucf:
I saw #2699 and was wondering if files under ```C:\Users\<username>\.nuget``` and ```C:\Program Files\dotnet\sdk\``` can change during a build, and if not then I believe we can partially cache IO checks here between targets. They respectively make up 45% and 4-18% (depending on .NET Core or .NET Framework) of filepaths that pass through ```IsAnyOutOfDate``` when building WebLargeCore, so it'd be another huge boost.

rainersigwald:
Alas, they can: the `Restore` target can definitely add new files under `.nuget` (though the regular build phase _shouldn't_ do so, it also can).

rainersigwald:
Closing since we've let this sit for way too long without making a pro/con decision. Let's revisit it if #3586 comes up in profiles again.

PR #3611, opened 2018-08-15
Stop IBCMerge for VSSetup project

rainersigwald:
src\Package\MSBuild.VSSetup\MSBuild.VSSetup.csproj has an assembly name
that matches a shipped assembly, so the IBCMerge targets attempt to
apply PGO data to it, but it does not have any matching methods because
it's not a real project.

Respect the preexisting `$(MicroBuild_EnablePGO)` when deciding whether
to include the IBCMerge targets.


rainersigwald:
Note target is 15.9, since this is a build-infra change that reduces warnings.

PR #3613, opened 2018-08-15
Target472

cdmihai:
Clone of #3598

### DIFF
--- a/src/Tasks/Microsoft.Build.Tasks.csproj
+++ b/src/Tasks/Microsoft.Build.Tasks.csproj
@@ -1000,6 +1000,7 @@
     <Content Update="@(Content)" Pack="false" />
   </ItemGroup>
 
+  <!-- The RoslynCodeTaskFactory needs ref assemblies to compile against. -->
   <Target Name="CopyMscorlib" AfterTargets="Build">
     <Copy SourceFiles="$(NuGetPackageRoot)\netstandard.library\2.0.3\build\netstandard2.0\ref\netstandard.dll" DestinationFolder="$(ArtifactsBinDir)/MSBuild/$(TargetFramework)/ref" />
     <Copy SourceFiles="$(NuGetPackageRoot)\netstandard.library\2.0.3\build\netstandard2.0\ref\mscorlib.dll" DestinationFolder="$(ArtifactsBinDir)/MSBuild/$(TargetFramework)/ref" />
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -22,7 +22,7 @@ phases:
     inputs:
       testRunTitle: 'Windows-on-full Full Framework'
       testRunner: XUnit
-      testResultsFiles: 'artifacts/**/*UnitTests_net46*.xml'
+      testResultsFiles: 'artifacts/**/*UnitTests_net472*.xml'
       publishRunAttachments: true
       mergeTestResults: true
     condition: always()
@@ -55,7 +55,7 @@ phases:
     inputs:
       testRunTitle: 'Windows-on-core Full Framework'
       testRunner: XUnit
-      testResultsFiles: 'artifacts/**/*UnitTests_net46*.xml'
+      testResultsFiles: 'artifacts/**/*UnitTests_net472*.xml'
       publishRunAttachments: true
       mergeTestResults: true
     condition: always()
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,6 +6,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <FullFrameworkTFM>net472</FullFrameworkTFM>
+  </PropertyGroup>
+
+  <PropertyGroup>
     <OsEnvironment Condition="$([MSBuild]::IsOSPlatform('windows'))">windows</OsEnvironment>
     <OsEnvironment Condition="$([MSBuild]::IsOSPlatform('linux'))">linux</OsEnvironment>
     <OsEnvironment Condition="$([MSBuild]::IsOSPlatform('osx'))">osx</OsEnvironment>
--- a/build/Packages.props
+++ b/build/Packages.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-      <NuGetPackageVersion Condition="'$(NuGetPackageVersion)' == ''">4.8.0-preview1.5185</NuGetPackageVersion>
+      <NuGetPackageVersion Condition="'$(NuGetPackageVersion)' == ''">4.8.0-rtm.5362</NuGetPackageVersion>
       <NuGetBuildTasksVersion Condition="'$(NuGetBuildTasksVersion)' == ''">$(NuGetPackageVersion)</NuGetBuildTasksVersion>
       <NuGetCommandsVersion Condition="'$(NuGetCommandsVersion)' == ''">$(NuGetPackageVersion)</NuGetCommandsVersion>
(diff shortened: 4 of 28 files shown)


cdmihai:
@crummel any idea why crossplat builds fail to find the net472 ref assemblies? https://dotnet.visualstudio.com/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_build/results?buildId=13243

crummel:
I know I was specifically pointed to the `003` version but I don't know if that's related, I'm trying it out now.

crummel:
Ah, I think this is the issue:
`Installing Microsoft.NETFramework.ReferenceAssemblies.net471 1.0.0-alpha-003.`

Looking into why that's happening instead of net472 now.

crummel:
`004` actually works properly for me but fails later:
```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---> System.DllNotFoundException: Unable to load shared library '/home/chris/.nuget/packages/nerdbank.gitversioning/2.1.23/build/MSBuildFull/lib/linux/x86_64/libgit2-1196807.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libssl.so.1.0.0: cannot open shared object file: No such file or directory
   at System.Runtime.Loader.AssemblyLoadContext.InternalLoadUnmanagedDllFromPath(String unmanagedDllPath)
   at System.Runtime.Loader.AssemblyLoadContext.LoadUnmanagedDllFromPath(String unmanagedDllPath)
   at MSBuildExtensionTask.ContextAwareTask.CustomAssemblyLoader.LoadUnmanagedDll(String unmanagedDllName)
   at System.Runtime.Loader.AssemblyLoadContext.ResolveUnmanagedDll(String unmanagedDllName, IntPtr gchManagedAssemblyLoadContext)
   at LibGit2Sharp.Core.NativeMethods.git_libgit2_init()
   at LibGit2Sharp.Core.NativeMethods.LibraryLifetimeObject..ctor()
   at LibGit2Sharp.Core.NativeMethods..cctor()
   --- End of inner exception stack trace ---
   at LibGit2Sharp.Core.NativeMethods.RemoveHandle()
   at LibGit2Sharp.Core.NativeMethods.LibraryLifetimeObject.Finalize()
Failed to run MSBuild: "/home/chris/Microsoft_msbuild/build/../artifacts/.dotnet/2.1.401-preview-009229/dotnet" msbuild "/home/chris/.nuget/packages/roslyntools.repotoolset/1.0.0-beta2-62901-01/tools/Build.proj" /m /clp:Summary /v:minimal /p:Configuration=Debug /p:RepoRoot="/home/chris/Microsoft_msbuild/build/.." /p:Projects="/home/chris/Microsoft_msbuild/build/../MSBuild.sln" /p:CIBuild=false /p:DisableNerdbankVersioning=false /bl:"/home/chris/Microsoft_msbuild/build/../artifacts/Debug/log/Build.binlog" /p:Restore=true /p:Build=true /p:Rebuild=false /p:Test=false /p:Sign=false /p:Pack=false /p:CreateBootstrap=true
```

Looks like my setup might not be right to build?

rainersigwald:
@crummel What distro/version are you seeing that on? IIRC that's a failure we see in versioning on newer Linuxes.

cdmihai:
@crummel had the exact libgit2sharp failure but fixed it by installing libunwind8 and libcurl3

cdmihai:
Tried reverting back to 003 but I still see ` Installing Microsoft.NETFramework.ReferenceAssemblies.net471 1.0.0-alpha-003.`

crummel:
I don't know if this helps, but it looks like in the working version, `GetReferenceAssemblyPaths` is being called with `RootPath`:
```
GetReferenceAssemblyPaths: (TargetId:292)
                     Task Parameter:RootPath=/root/.nuget/packages/microsoft.netframework.referenceassemblies.net46/1.0.0-alpha-003/build/ (TaskId:160)
````
but in a broken build it's only ever called with the TFM:
```
GetReferenceAssemblyPaths: (TargetId:319)
                     Task Parameter:TargetFrameworkMoniker=.NETFramework,Version=v4.7.2 (TaskId:233)
```

This difference in the logs happens right before the reference error, so I'm going to continue to look into it.

**Comment by crummel** [comment]
It appears `TargetFrameworkIdentifier` is never `.NETFramework` in the broken build, so the `TargetFrameworkRootPath` is never set.  Continuing to investigate...

**Comment by crummel** [comment]
@cdmihai Looks like the problem is here: https://github.com/Microsoft/msbuild/blob/6f7b5381e1b91e32734636a2128d0465e6f1e55f/src/Samples/Directory.Build.targets#L5
This is causing some kind of conflict.  I upgraded this to the `004` version as well and the build succeeded through the bootstrap process, but failed shortly after that with a lot of copies of this error:
```
/msbuild/artifacts/Debug/bootstrap/netcoreapp2.1/MSBuild/Microsoft.CSharp.CurrentVersion.targets(321,5): error MSB4019: The imported project "/msbuild/artifacts/Debug/bootstrap/netcoreapp2.1/MSBuild/Roslyn/Microsoft.CSharp.Core.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
````

**Comment by crummel** [comment]
The second issue appears to be related to the `'$(MSBuildRuntimeType)' == 'Core'` condition introduced recently here:
https://github.com/Microsoft/msbuild/blob/6f7b5381e1b91e32734636a2128d0465e6f1e55f/src/Tasks/Microsoft.Build.Tasks.csproj#L962

Removing that for both of those PackageReferences succeeds all the way through the build.

**Comment by cdmihai** [comment]
D'oh, I completely forgot about the samples.

The compiler issue is not related to this, working on that separately: #3615

**Comment by cdmihai** [comment]
Whenever this passes, I am going to manually push it as a new `exp/net472` branch in our main repo. That way we can review other work based on 472.

File: src/Tasks/Microsoft.Build.Tasks.csproj [hunk 1]
```diff
@@ -999,9 +999,4 @@
     -->
     <Content Update="@(Content)" Pack="false" />
   </ItemGroup>
-
-  <Target Name="CopyMscorlib" AfterTargets="Build">
```
cdmihai: @jeffkl Is it OK to remove these? Does RoslynCodeTaskFactory need them or something?

src/Tasks/Microsoft.Build.Tasks.csproj
```diff
@@ -999,9 +999,4 @@
...
-    <Copy SourceFiles="$(NuGetPackageRoot)\netstandard.library\2.0.3\build\netstandard2.0\ref\netstandard.dll" DestinationFolder="$(ArtifactsBinDir)/MSBuild/$(TargetFramework)/ref" />
-    <Copy SourceFiles="$(NuGetPackageRoot)\netstandard.library\2.0.3\build\netstandard2.0\ref\mscorlib.dll" DestinationFolder="$(ArtifactsBinDir)/MSBuild/$(TargetFramework)/ref" />
```
rainersigwald: This worries me -- @jeffkl don't we still need these for RoslynCodeTaskFactory?

src/Tasks/Microsoft.Build.Tasks.csproj
jeffkl: Yes we need them so that RoslynCodeTaskFactory can compile assemblies.  They're just ref assemblies

cdmihai:
Pushed to https://github.com/Microsoft/msbuild/tree/exp/net472
Will merge into master when VS moves to 472.

PR #3615, opened 2018-08-15
Refer to Compilers package unconditionally

rainersigwald:
MSBuild's unit tests depend on a functional MSBuild deployment,
including Roslyn tasks and targets. These weren't getting copied on the
initial build because the NuGet package that contained them (for full
framework) wasn't guaranteed to be restored. Relax the conditions on
the PackageReference that attempted to ensure that.

### DIFF
--- a/src/Tasks/Microsoft.Build.Tasks.csproj
+++ b/src/Tasks/Microsoft.Build.Tasks.csproj
@@ -955,14 +955,21 @@
     <PackageReference Include="System.Collections.Immutable" />
   </ItemGroup>
 
+  <!-- Tasks need to mimic redistributing the compilers, so add references to both full framework and .net core -->
+  <ItemGroup>
+    <!-- reference both when repotoolset is not adding compiler references -->
+    <PackageReference Include="Microsoft.Net.Compilers" ExcludeAssets="all" Condition="'$(UsingToolMicrosoftNetCompilers)' == 'false'" />
+    <PackageReference Include="Microsoft.NETCore.Compilers" ExcludeAssets="all" Condition="'$(UsingToolMicrosoftNetCompilers)' == 'false'" />
+
+    <!-- reference the missing package when repotoolset is adding compiler references based on MSBuildRuntimeType https://github.com/dotnet/roslyn-tools/blob/master/src/RepoToolset/tools/Compiler.props -->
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.Build.Tasks.csproj
```diff
@@ -955,20 +955,11 @@
     <PackageReference Include="System.Collections.Immutable" />
   </ItemGroup>
 
-  <ItemGroup>
-    <!-- Repotools will add a PackageReference for the compilers package based on the current runtime.
```
rainersigwald: @AndyGerlicher This didn't appear to be the case in my testing; how did you figure out that it was?

AndyGerlicher: This file adds the PackageReference:
https://github.com/dotnet/roslyn-tools/blob/master/src/RepoToolset/tools/Compiler.props

And it gets imported here:
https://github.com/dotnet/roslyn-tools/blob/8e0ee77c949d0c8f42ecbaa3ccd6d382111519c4/src/RepoToolset/tools/Settings.props#L18

But only based on `$(UsingToolMicrosoftNetCompilers)`. Which we set to `false` in our repo. And that's ignored and set to `true` here:
https://github.com/dotnet/roslyn-tools/blob/145e3b3d376ce385f40dfc1ef24dbefaec6b3b6e/src/RepoToolset/tools/DefaultVersions.props#L34

But only when `$(CIBuild)` is `true`, which is set here:
https://github.com/Microsoft/msbuild/blob/master/build/build.ps1#L239

When you pass -ci to the build.ps1. Which is why this PR worked on your machine but will likely fail in some weird way if VSTS eventually picks it up.

It's so straightforward it really almost explains itself.

AndyGerlicher: In all seriousness though I would love to disable that entirely and do it explicitly when it's needed like you've got it. This was a nightmare.

Oh and I remembered the reason this worked before is because it didn't matter that we were adding a PR twice (once in repo tools once here). It matters now because CLI emits a warning if there are any two package references that are `IsImplicitlyDefined=true`.

cdmihai:
@rainersigwald how does one restart the VSTS CI?

Issue #2095, opened 2017-05-12
Allow sdk resolvers to query arbitrary msbuild properties


nguerrera:
Today the context available to sdk resolvers is fixed by https://github.com/Microsoft/msbuild/blob/master/src/Framework/Sdk/SdkResolverContext.cs#L9

Seeing that [`SolutionFilePath` comes from `$(SolutionPath)`](https://github.com/Microsoft/msbuild/blob/0e8c695f7af207a6281c76de2369f91f9067da13/src/Build/Evaluation/Evaluator.cs#L2374),

I wonder if we can  just have an extra method on SdkResolverContext that lets resolvers query any propety:

``` C#
public abstract class SdkResolverContext {
   // ...
   public abstract string GetProperty(string name);
}
```

This would allow resolvers to document arguments that are passed as plain msbuild properties. 

Without this, we're currently using environment variables  to override some of our resolvers behavior:

https://github.com/dotnet/cli/blob/09dd14bfe467e1cd264740af6ed4a8a243ccb53a/src/Microsoft.DotNet.MSBuildSdkResolver/MSBuildSdkResolver.cs#L20-L24
https://github.com/dotnet/cli/blob/09dd14bfe467e1cd264740af6ed4a8a243ccb53a/src/Microsoft.DotNet.MSBuildSdkResolver/MSBuildSdkResolver.cs#L94-L98

Our immediate use case was to have a test hook, but there are production scenarios too. The main one is pulling down the .NET Core SDK to an arbitrary, non-global location as part of the build and signaling to the resolver to choose the appropriate sdk from there and not from program files. (Now, that actually overlaps with another feature that's evolving and we will likely land in a place where you can edit global.json to get this behavior without setting any environment variables, but the mechanism here applies more generally to arbitrary resolvers with arbitrary input.)

cc @jaredpar @AndyGerlicher

jaredpar:
Very much in favor of this approach.  Today virtually anything about the MSBuild experience can be customized via properties.  True there are a few cases where you need to be careful about ordering (`<UsingTask>` for example).  But overall you can completely configure the build via properties.

Having SDK deviate from that would be a bit painful.

rainersigwald:
I'm strongly opposed to breaking the encapsulation of the resolver in this way. Plus, it would be very, very confusing: only global properties would be available in the normal case (`<Project Sdk="" />`), because at the time of SDK resolution none of the project file has been parsed.

jaredpar:
[...]

PR #3617, opened 2018-08-16
Add global properties to SdkResolverContext

jeffkl:
This will allow the NuGet SDK resolver to read the global property `NuGetInteractive` so it can prompt the user.

### DIFF
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -453,6 +453,7 @@ namespace Microsoft.Build.Framework
     public abstract partial class SdkResolverContext
     {
         protected SdkResolverContext() { }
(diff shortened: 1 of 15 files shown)


jeffkl:
Please review ASAP to get this in so I can update NuGet's SDK resolver.

AndyGerlicher:
@nguerrera any thoughts or concerns with this?

rainersigwald:
I'm not a fan of this approach, as I mentioned in https://github.com/Microsoft/msbuild/issues/2095#issuecomment-301152970.

nguerrera:
I've asked for it a few times, See #2095.

rainersigwald:
Though restricting to global properties helps, I'm not sure it's enough since you could have global properties set in a p2p reference.

nguerrera:
At the end of #2095, I asked if global properties alone would be OK. Seems we have varying answers still. :)

nguerrera:
I have no objections and you can close #2095 if you decide to take it. I don't have a pressing need for this either.

jeffkl:
Nevermind...

PR #3621, opened 2018-08-16
Add Interactive property to SdkResolverContext

jeffkl:
Defaults to false.  Currently only set to `true` if the global property `NuGetInteractive` is set to "true".

In 16.0, we should add an `/interactive` command-line argument which sets this instead.


jeffkl:
Replacement of #3617

jeffkl:
Please review at your earliest convenience!

PR #3625, opened 2018-08-17
Add ContainsItem item intrinsic function

cdmihai:
Realized this is really cumbersome to do during evaluation time.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -2554,6 +2554,30 @@ namespace Microsoft.Build.Evaluation
                 }
 
                 /// <summary>
+                /// Intrinsic function that returns a boolean to indicate if any of the items have the given value
+                /// Using a case insensitive comparison
+                /// </summary>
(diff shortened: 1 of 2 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -2553,6 +2553,30 @@ internal static ItemTransformFunction GetItemTransformFunction(IElementLocation
...
+                    foreach (var item in itemsOfType)
+                    {
+                        if (item.Key.Equals(searchedItem, StringComparison.OrdinalIgnoreCase))
+                        {
+                            hasItem = true;
+                            break;
+                        }
+                    }
+
+                    yield return new Pair<string, S>(hasItem ? "true" : "false", null /* no base item */);
```
rainersigwald: This is pretty different from the return pattern in `AnyHaveMetadataValue` which is otherwise similar:

https://github.com/Microsoft/msbuild/blob/4919c286d3bbf5cf1740860eefaea2b8c4e21b94/src/Build/Evaluation/Expander.cs#L2615-L2627

Do you know what changes downstream if we don't return the first matching item?

cdmihai:
Closing as I didn't think of `AnyHaveMetadataValue('Identity', whatever)`. No use adding multiple ways to do the same think, and the metadata function is not that bad.

PR #3631, opened 2018-08-18
Enable max path detection

ccastanedaucf:
Also might be useful to have the ability to revert to old behavior if necessary. Should this be controlled via environment variable?


ccastanedaucf:
Well that's a new one. Test ```ForceSomeOutOfDate``` kept failing on .NET Core both locally and on CI for a few hours while I couldn't figure out what could have possibly changed in behavior. Then it started passing locally and on CI again.

rainersigwald:
I looked at the test and the failing assertion and I don't think I can explain that 😕

PR #3632, opened 2018-08-19
Log debug dumps

danmosemsft:
Fix https://github.com/Microsoft/msbuild/issues/1038

When terminating due to a child node crash, we invite the user to look for diagnostic files in %temp%. This is problematic on a CI system, because unless configured especially to understand MSBuild dump files, the user will likely not be able to access them. (Occasionally we encounter this problem in the CoreFX CI for one reason or another.)

* Dump the contents of those dump files to the console, if we are rudely terminating.
* Only do this for files that were written to since the BuildManager was instantiated - so we don't show any stale ones. We could further filter by PID's of our child nodes but wiring this through is messy and does not seem to add anything useful.
* Correct parameter misordering I noticed

I tested this by adding bugs to the child nodes to make them crash.

Old output:
```
     0>MSBUILD : error MSB4166: Child node "2" exited prematurely. Shutting down. Diagnostic information may be found in files in "C:\Users\danmo\AppData\Local\Temp\" and will be named MSBuild_*.failure.txt. This location can be changed by setting the MSBUILDDEBUGPATH environment variable to a different directory.
     1>Done Building Project "C:\dotnet\1\1.sln" (rebuild target(s)) -- FAILED.
```
New output:
```
     0>MSBUILD : error MSB4166: Child node "2" exited prematurely. Shutting down. Diagnostic information may be found in files in "C:\Users\danmo\AppData\Local\Temp\" and will be named MSBuild_*.failure.txt. This location can be changed by setting the MSBUILDDEBUGPATH environment variable to a different directory.
MSBUILD : error MSB4166: C:\Users\danmo\AppData\Local\Temp\MSBuild_pid-71208_af99fbd50dfb4bf7b6a0dbeff520cd46.failure.txt:
MSBUILD : error MSB4166: UNHANDLED EXCEPTIONS FROM PROCESS 71208:
MSBUILD : error MSB4166: =====================
MSBUILD : error MSB4166: 8/18/2018 4:52:22 PM
MSBUILD : error MSB4166: System.AggregateException: One or more errors occurred. (Parameter "packet" cannot be null.) ---> System.ArgumentNullException: Parameter "packet" cannot be null.
MSBUILD : error MSB4166:    at Microsoft.Build.Shared.ErrorUtilities.VerifyThrowArgumentNull(Object parameter, String parameterName, String resourceName) in C:\git\MSBuild\src\Shared\ErrorUtilities.cs:line 778
...
MSBUILD : error MSB4166:    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
MSBUILD : error MSB4166:    at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)<---
MSBUILD : error MSB4166:
MSBUILD : error MSB4166: ===================
MSBUILD : error MSB4166:
MSBUILD : error MSB4166:
     1>Done Building Project "C:\dotnet\1\1.sln" (rebuild target(s)) -- FAILED.
```

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildManager.cs
+++ b/src/Build/BackEnd/BuildManager/BuildManager.cs
@@ -208,6 +208,11 @@ namespace Microsoft.Build.Execution
         /// </summary>
         private bool _disposed;
 
+        /// <summary>
+        /// When the BuildManager was created.
+        /// </summary>
+        private DateTime _instantiationTimeUtc;
+
 #if DEBUG
         /// <summary>
         /// <code>true</code> to wait for a debugger to be attached, otherwise <code>false</code>.
@@ -250,6 +255,7 @@ namespace Microsoft.Build.Execution
             _projectFinishedEventHandler = OnProjectFinished;
             _loggingThreadExceptionEventHandler = OnThreadException;
             _legacyThreadingData = new LegacyThreadingData();
+            _instantiationTimeUtc = DateTime.UtcNow;
         }
 
         /// <summary>
@@ -1409,7 +1415,8 @@ namespace Microsoft.Build.Execution
                     foreach (BuildSubmission submission in _buildSubmissions.Values)
                     {
                         BuildEventContext buildEventContext = new BuildEventContext(submission.SubmissionId, BuildEventContext.InvalidNodeId, BuildEventContext.InvalidProjectInstanceId, BuildEventContext.InvalidProjectContextId, BuildEventContext.InvalidTargetId, BuildEventContext.InvalidTaskId);
-                        loggingService.LogError(buildEventContext, new BuildEventFileInfo(String.Empty) /* no project file */, "ChildExitedPrematurely", node, ExceptionHandling.DebugDumpPath);
+                        string exception = ExceptionHandling.ReadAnyExceptionFromFile(_instantiationTimeUtc);
+                        loggingService.LogError(buildEventContext, new BuildEventFileInfo(String.Empty) /* no project file */, "ChildExitedPrematurely", node, ExceptionHandling.DebugDumpPath, exception);
                     }
                 }
                 else if (shutdownPacket.Reason == NodeShutdownReason.Error && _buildSubmissions.Values.Count == 0)
@@ -1418,7 +1425,7 @@ namespace Microsoft.Build.Execution
                     if (shutdownPacket.Exception != null)
                     {
                         ILoggingService loggingService = ((IBuildComponentHost)this).GetComponent(BuildComponentType.LoggingService) as ILoggingService;
-                        loggingService.LogError(BuildEventContext.Invalid, new BuildEventFileInfo(String.Empty) /* no project file */, "ChildExitedPrematurely", shutdownPacket.Exception.ToString(), ExceptionHandling.DebugDumpPath);
+                        loggingService.LogError(BuildEventContext.Invalid, new BuildEventFileInfo(String.Empty) /* no project file */, "ChildExitedPrematurely", node, ExceptionHandling.DebugDumpPath, shutdownPacket.Exception.ToString());
                         OnThreadException(shutdownPacket.Exception);
                     }
                 }
--- a/src/Shared/ExceptionHandling.cs
+++ b/src/Shared/ExceptionHandling.cs
@@ -5,6 +5,7 @@
 namespace Microsoft.Build.AppxPackage.Shared
 #else
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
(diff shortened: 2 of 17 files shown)


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -208,6 +208,11 @@ public class BuildManager : INodePacketHandler, IBuildComponentHost, IDisposable
         /// </summary>
         private bool _disposed;
 
+        /// <summary>
+        /// When the BuildManager was created.
+        /// </summary>
+        private DateTime _instantiationTimeUtc;
```
rainersigwald: Does the `Utc` suffix carry information? I thought that was embedded in the `Kind` property.

src/Shared/ExceptionHandling.cs
```diff
@@ -351,6 +353,28 @@ internal static void DumpExceptionToFile(Exception ex)
...
+        /// since the provided time, otherwise returns an empty string.
+        /// </summary>
+        internal static string ReadAnyExceptionFromFile(DateTime fromTimeUtc)
+        {
+            string result = "";
+            IEnumerable<string> files = FileSystems.Default.EnumerateFiles(DebugDumpPath, "MSBuild*failure.txt");
+
+            foreach (string file in files)
+            {
+                if (File.GetLastWriteTimeUtc(file) >= fromTimeUtc)
```
rainersigwald: Clever!

src/Shared/ExceptionHandling.cs
```diff
@@ -351,6 +353,28 @@ internal static void DumpExceptionToFile(Exception ex)
...
             }
         }
+
+        /// <summary>
+        /// Returns the content of any exception dump files modified
+        /// since the provided time, otherwise returns an empty string.
+        /// </summary>
+        internal static string ReadAnyExceptionFromFile(DateTime fromTimeUtc)
+        {
+            string result = "";
```
rainersigwald: I assume using a StringBuilder here isn't worth much since this is an error-handling about-to-shut-down path and we don't care about GC? Only counterpoint I can think of is "what if we're already under memory pressure and that's why something crashed".

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1418,7 +1425,7 @@ private void HandleNodeShutdown(int node, NodeShutdown shutdownPacket)
                     if (shutdownPacket.Exception != null)
                     {
                         ILoggingService loggingService = ((IBuildComponentHost)this).GetComponent(BuildComponentType.LoggingService) as ILoggingService;
-                        loggingService.LogError(BuildEventContext.Invalid, new BuildEventFileInfo(String.Empty) /* no project file */, "ChildExitedPrematurely", shutdownPacket.Exception.ToString(), ExceptionHandling.DebugDumpPath);
+                        loggingService.LogError(BuildEventContext.Invalid, new BuildEventFileInfo(String.Empty) /* no project file */, "ChildExitedPrematurely", node, ExceptionHandling.DebugDumpPath, shutdownPacket.Exception.ToString());
```
rainersigwald: Was this just horribly wrong before? It sure looks better now . . .

src/Build/BackEnd/BuildManager/BuildManager.cs
danmosemsft: You're right I believe (although if you stray from UTC, it doesn't track anything but "not UTC"). However I am not sure that comparisons of DateTime flag when you are comparing against non-UTC. The concept of Kind was bolted on later. I considered using DateTimeOffset, but the pattern seems to be to use DateTime and the Utc suffix to help visually spot mistakes, eg.,

```c#
        // The newest file what we have seen
        private DateTime _newestFileTimeUtc = DateTime.MinValue;
```
```c#
                    destinationPath.LastWriteTimeUtc = zipArchiveEntry.LastWriteTime.UtcDateTime;
```

src/Build/BackEnd/BuildManager/BuildManager.cs
danmosemsft: Yes, but the other code path was generally or always what happened so maybe nobody noticed.

src/Shared/ExceptionHandling.cs
danmosemsft: I don't think there's a clear winner for a small number of appends if done only once, as StringBuilder resizes without knowledge of what is to come, so it could do several resizes, and those need to be GC'd too. Fair enough though, I changed.

danmosemsft:
Updated.

src/Shared/ExceptionHandling.cs
```diff
@@ -351,6 +354,31 @@ internal static void DumpExceptionToFile(Exception ex)
...
             }
         }
+
+        /// <summary>
+        /// Returns the content of any exception dump files modified
+        /// since the provided time, otherwise returns an empty string.
+        /// </summary>
+        internal static string ReadAnyExceptionFromFile(DateTime fromTimeUtc)
+        {
+            var builder = new StringBuilder();
```
cdmihai: Nit: You could use the ReusableStringBuilder here. I wouldn't care too much since this abnormal termination. Though VS does have a metric which counts some form of memory at VS exit (though it probably does not include node crashes)

danmosemsft: After this point, VS presumably won't work any more anyway.. at least I'm assuming the BuildManager can't be reused, who knows what state everything is in.

cdmihai:
Thank you @danmosemsft!

PR #3648, opened 2018-08-23
Add Microsoft.IO.Redist and System.Memory package

cdmihai:
Extracted common changes from 2 branches that I'm working on.
Tested the changes by aliasing the IO types in FileUtilities. (actual change in separate branch)

### DIFF
--- a/build/Packages.props
+++ b/build/Packages.props
@@ -13,6 +13,7 @@
     <PackageReference Update="Microsoft.CodeAnalysis.Build.Tasks" Version="3.0.0-beta1-61516-01" />
     <PackageReference Update="Microsoft.DotNet.BuildTools.GenAPI" Version="2.1.0-prerelease-02404-02" />
     <PackageReference Update="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
+    <PackageReference Update="Microsoft.IO.Redist" Version="4.6.0-preview1-26822-04" />
     <PackageReference Update="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" />
     <PackageReference Update="Microsoft.NETCore.Compilers" Version="$(MicrosoftNetCoreCompilersVersion)" />
     <PackageReference Update="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.15" />
@@ -29,6 +30,7 @@
     <PackageReference Update="System.IO.Compression" Version="4.3.0" />
     <PackageReference Update="System.IO.FileSystem.Primitives" Version="4.3.0" />
     <PackageReference Update="System.Linq.Parallel" Version="4.0.1" />
+    <PackageReference Update="System.Memory" Version="4.5.1" />
     <PackageReference Update="System.Net.Http" Version="4.3.0" />
     <PackageReference Update="System.Reflection.Metadata" Version="1.6.0" />
     <PackageReference Update="System.Reflection.TypeExtensions" Version="4.1.0" />
@@ -38,7 +40,7 @@
     <PackageReference Update="System.Runtime.Serialization.Primitives" Version="4.1.1" />
     <PackageReference Update="System.Security.Principal.Windows" Version="4.3.0" />
     <PackageReference Update="System.Text.Encoding.CodePages" Version="4.0.1" />
-    
+
     <PackageReference Update="System.Threading.Tasks.Dataflow" Version="4.5.24.0" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'" />
     <PackageReference Update="System.Threading.Tasks.Dataflow" Version="4.6.0" Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'"/>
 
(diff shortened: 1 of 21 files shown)


cdmihai:
@ccastanedaucf You can use `exp/net472` after this gets merged in to experiment with spans and stuff.

build/Packages.props
```diff
@@ -38,7 +39,9 @@
     <PackageReference Update="System.Runtime.Serialization.Primitives" Version="4.1.1" />
     <PackageReference Update="System.Security.Principal.Windows" Version="4.3.0" />
     <PackageReference Update="System.Text.Encoding.CodePages" Version="4.0.1" />
-    
+
+    <PackageReference Update="Microsoft.IO.Redist" Version="4.6.0-preview1-26822-04" />
```
rainersigwald: nit: I feel like this should be sorted alphabetically.

build/build.ps1
```diff
@@ -39,6 +40,7 @@ function Print-Usage() {
     Write-Host "  -skipTests              Don't run tests"
     Write-Host "  -test                   Run tests. Ignores skipTests"
     Write-Host "  -bootstrapOnly          Don't run build again with bootstrapped MSBuild"
+    Write-Host "  -rebuildWithBootstrap   Run build again with bootstrapped MSBuild. Ignores bootstrapOnly"
```
rainersigwald: In my head, these should be

```
-noBootstrap
-bootstrap
```

That'd be quicker to type on the command line. But I'll adjust based on consensus.

src/Framework.UnitTests/Microsoft.Build.Framework.UnitTests.csproj
```diff
@@ -12,13 +12,18 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.VisualStudio.Setup.Configuration.Interop" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'" />
```
rainersigwald: nit: Can you move this into the new ItemGroup you created with this condition?

src/Build/Evaluation/ProjectRootElementCache.cs
```diff
@@ -208,7 +208,7 @@ internal ProjectRootElementCache(bool autoReloadFromDisk)
 
                 if (projectRootElement != null && _autoReloadFromDisk)
                 {
-                    FileInfo fileInfo = FileUtilities.GetFileInfoNoThrow(projectFile);
+                    var fileInfo = FileUtilities.GetFileInfoNoThrow(projectFile);
```
rainersigwald: I'm a bit skeptical of this change; isn't our style generally "use `var` only when the type is unambiguous, usually because it's a constructor on the right-hand side"? Or will this help with a substitution later?

src/Utilities/Microsoft.Build.Utilities.csproj
```diff
@@ -31,6 +32,10 @@
     <PackageReference Include="System.Text.Encoding.CodePages" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <PackageReference Include="Microsoft.IO.Redist" />
```
rainersigwald: These references have setup impact, right? Can you add them to the VSIX?

cdmihai: Good point.

build/build.ps1
rainersigwald: In my head `-bootstrap` means "build with the bootstrapped MSBuild". I totally get that that's not 100% logical.

src/Build/Evaluation/ProjectRootElementCache.cs
rainersigwald: Hmm. On the one hand this is nice and easy. On the other, if it were harder to do that, it might be easier to see the impact of changes in PR? I'm fine leaving the change in.

akoeplinger:
I assume this impacts the Mono build too right?

/cc @radical

cdmihai:
@akoeplinger @radical 
Not quite sure how this impacts Mono. The [redist package ](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.IO.Redist) targets 4.7.2, which should be OK with Mono? Maybe it would require updating Mono for our repo?

akoeplinger:
@cdmihai it depends on what the package does (haven't looked), if it just p/invokes into Win32 APIs that won't work on Mono :)

cdmihai:
@rainersigwald, ready to merge?

build/cibuild.cmd
```diff
@@ -1,3 +1,3 @@
 @echo off
-powershell -NoLogo -NoProfile -ExecutionPolicy ByPass "& '%~dp0build.ps1'" -build -pack -sign -ci -prepareMachine %*
```
rainersigwald: Should we really drop `-sign`? IIRC it only validates that the official build would sign things, when in CI mode.

cdmihai: Would we catch missing entries in `signtooldata.json` if we take it out?

rainersigwald: No, I don't think we would--that's why I'm asking if it should be put back in.

cdmihai: Oh, I only now realize I actually deleted `-sign`. I'll go ahead and add it back :)

PR #3652, opened 2018-08-24
Optimize RegularExpressionFromFileSpec (#2380)

### LINKED ISSUE #2380: 2.5% of evaluation time is spent converting file spec into a regex string
Investigating long evaluation times in Roslyn and customer traces, and 2.5% of evaluation time (about 1% of build time) is spent just converting a file spec into a regex string:

image

Looks to be most time spent in Replace - is there a way we can be smarter about this and avoid all the replaces?

ccastanedaucf:
Fixes #2380

Got rid of all ```StringBuilder.Replace()``` calls by just iteratively building the regex without the tagging format and with no allocations. Execution time is so low it just gets absorbed into ```GetFileSpecInfo``` in the profiler, the 7ms that shows up is all from ```SplitFileSpec()```.

Perf profile on master:
beforeperf

Perf profile on this PR:
afterperf

### DIFF
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -1122,16 +1122,10 @@ namespace Microsoft.Build.Shared
                 + FileSpecRegexParts.BeginningOfLine.Length + FileSpecRegexParts.FixedDirWildcardDirSeparator.Length
                 + FileSpecRegexParts.WildcardDirFilenameSeparator.Length + FileSpecRegexParts.EndOfLine.Length
             );
-            matchFileExpression.Append(FileSpecRegexParts.BeginningOfLine);
 
             AppendRegularExpressionFromFixedDirectory(matchFileExpression, fixedDirectoryPart);
-            matchFileExpression.Append(FileSpecRegexParts.FixedDirWildcardDirSeparator);
-
             AppendRegularExpressionFromWildcardDirectory(matchFileExpression, wildcardDirectoryPart);
-            matchFileExpression.Append(FileSpecRegexParts.WildcardDirFilenameSeparator);
-
             AppendRegularExpressionFromFilename(matchFileExpression, filenamePart);
-            matchFileExpression.Append(FileSpecRegexParts.EndOfLine);
 
             return matchFileExpression.ToString();
         }
@@ -1180,6 +1174,8 @@ namespace Microsoft.Build.Shared
 
         private static void AppendRegularExpressionFromFixedDirectory(StringBuilder regex, string fixedDir)
         {
+            regex.Append(FileSpecRegexParts.BeginningOfLine);
+
             /*
              * Capture the leading \\ in UNC paths, so that the doubled slash isn't
              * reduced in a later step.
@@ -1200,7 +1196,10 @@ namespace Microsoft.Build.Shared
 
         private static void AppendRegularExpressionFromWildcardDirectory(StringBuilder regex, string wildcardDir)
         {
+            regex.Append(FileSpecRegexParts.FixedDirWildcardDirSeparator);
+
             // fixed-directory + **\
+
             bool hasRecursiveOperatorAtStart = wildcardDir.Length > 2 && wildcardDir[0] == '*' && wildcardDir[1] == '*';
 
             if (hasRecursiveOperatorAtStart)
@@ -1223,6 +1222,8 @@ namespace Microsoft.Build.Shared
                     AppendRegularExpressionFromChar(regex, ch);
                 }
             }
+
+            regex.Append(FileSpecRegexParts.WildcardDirFilenameSeparator);
         }
 
         private static void AppendRegularExpressionFromFilename(StringBuilder regex, string filename)
@@ -1261,6 +1262,8 @@ namespace Microsoft.Build.Shared
                     i += 2;
                 }
             }
+
+            regex.Append(FileSpecRegexParts.EndOfLine);
         }
 
         /*
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -1223,25 +1223,25 @@ namespace Microsoft.Build.Shared
         private static void AppendRegularExpressionFromWildcardDirectory(StringBuilder regex, string wildcardDir)
         {
             // fixed-directory + **\
-            bool hasRecursionOperatorAtStart =
+            bool hasRecursiveOperatorAtStart =
                 wildcardDir.Length > 2 && wildcardDir[0] == '*' && wildcardDir[1] == '*'
                 && FileUtilities.IsAnySlash(wildcardDir[2]);
 
-            if (hasRecursionOperatorAtStart)
+            if (hasRecursiveOperatorAtStart)
             {
                 regex.Append(FileSpecRegexParts.LeftDirs);
             }
-            int startIndex = hasRecursionOperatorAtStart ? IndexOfNextNonCollapsibleChar(wildcardDir, 3) : 0;
+            int startIndex = hasRecursiveOperatorAtStart ? IndexOfNextNonCollapsibleChar(wildcardDir, 3) : 0;
 
             for (int i = startIndex; i < wildcardDir.Length; i = IndexOfNextNonCollapsibleChar(wildcardDir, i + 1))
             {
                 char ch = wildcardDir[i];
 
-                bool isRecursionOperator =
+                bool isRecursiveOperator =
                     i < wildcardDir.Length - 3 && FileUtilities.IsAnySlash(ch) && wildcardDir[i + 1] == '*'
                     && wildcardDir[i + 2] == '*' && FileUtilities.IsAnySlash(wildcardDir[i + 3]);
 
-                if (isRecursionOperator)
+                if (isRecursiveOperator)
                 {
                     regex.Append(FileSpecRegexParts.MiddleDirs);
                     i += 3;
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -1098,8 +1098,10 @@ namespace Microsoft.Build.Shared
         }
 
         /// <summary>
-        /// Given a file spec, create a regular expression that will match that
-        /// file spec.
+        /// Given a split file spec consisting of a directory without wildcard characters,
+        /// a sub-directory containing wildcard characters,
+        /// and a filename which may contain wildcard characters,
+        /// create a regular expression that will match that file spec.
         /// 
         /// PERF WARNING: this method is called in performance-critical
         /// scenarios, so keep it fast and cheap
@@ -1140,12 +1142,16 @@ namespace Microsoft.Build.Shared
             return StringBuilderCache.GetStringAndRelease(matchFileExpression);
         }
 
-        /*
-         * It is not legal for there to be a ".." after a wildcard.
-         *
-         * By definition, "**" must appear alone between directory slashes. If there is any remaining "**" then this is not
-         * a valid filespec.
-         */
+
+        /// <summary>
+        /// Determine if the filespec is legal according to the following conditions:
+        /// 
+        /// (1) It is not legal for there to be a ".." after a wildcard.
+        /// 
+        /// (2) By definition, "**" must appear alone between directory slashes.If there is any remaining "**" then this is not
+        ///     a valid filespec.
+        /// </summary>
+        /// <returns>True if both parts meet all conditions for a legal filespec.</returns>
         private static bool IsLegalFileSpec(string wildcardDirectoryPart, string filenamePart) =>
             !HasDotDot(wildcardDirectoryPart)
             && !HasMisplacedRecursiveOperator(wildcardDirectoryPart)
@@ -1181,7 +1187,14 @@ namespace Microsoft.Build.Shared
             return false;
         }
 
-        private static void AppendRegularExpressionFromFixedDirectory(StringBuilder regex, string fixedDir)
+
+        /// <summary>
+        /// Append the regex equivalents for character sequences in the fixed directory part of a filespec:
+        ///
+        /// (1) The leading \\ in UNC paths, so that the doubled slash isn't reduced in the last step
+        /// 
+        /// (2) Common filespec characters
+        /// </summary>
         {
             regex.Append(FileSpecRegexParts.BeginningOfLine);
 
@@ -1203,7 +1216,15 @@ namespace Microsoft.Build.Shared
             }
         }
 
-        private static void AppendRegularExpressionFromWildcardDirectory(StringBuilder regex, string wildcardDir)
+        /// <summary>
+        /// Append the regex equivalents for character sequences in the wildcard directory part of a filespec:
+        ///
+        /// (1) The leading **\ if existing
+        ///
+        /// (2) Each occurence of recursive wildcard \**\
+        /// 
+        /// (3) Common filespec characters
+        /// </summary>
         {
             regex.Append(FileSpecRegexParts.FixedDirWildcardDirSeparator);
 
@@ -1235,7 +1256,21 @@ namespace Microsoft.Build.Shared
             regex.Append(FileSpecRegexParts.WildcardDirFilenameSeparator);
         }
 
-        private static void AppendRegularExpressionFromFilename(StringBuilder regex, string filename)
+        /// <summary>
+        /// Append the regex equivalents for character sequences in the filename part of a filespec:
+        ///
+        /// (1) Trailing dots in file names have to be treated specially.
+        ///     We want:
+        ///
+        ///         *. to match foo
+        ///
+        ///     but 'foo' doesn't have a trailing '.' so we need to handle this while still being careful
+        ///     not to match 'foo.txt' by modifying the generated regex for wildcard characters * and ?
+        /// 
+        /// (2) Common filespec characters
+        ///
+        /// (3) Ignore the .* portion of any *.* sequence
+        /// </summary>
         {
             /* 
              * Trailing dots in file names have to be treated specially.
@@ -1275,10 +1310,9 @@ namespace Microsoft.Build.Shared
             regex.Append(FileSpecRegexParts.EndOfLine);
         }
 
-        /*
-         *  Replace characters with their regex equivalents.
-         */
-        private static void AppendRegularExpressionFromChar(StringBuilder regex, char ch)
+        /// <summary>
+        /// Append the regex equivalents for characters common to all filespec parts.
+        /// </summary>
         {
             if (ch == '*')
             {
@@ -1307,29 +1341,34 @@ namespace Microsoft.Build.Shared
             ch == '\\' || ch == '$' || ch == '(' || ch == ')' || ch == '+'
             || ch == '.' || ch == '[' || ch == '^' || ch == '{' || ch == '|';
 
-        /*
-         * Iteratively skip three cases involving directory separators
-         * 
-         *  (1) \.\ -> \
-         *        This is an identity, so for example, these two are equivalent,
-         * 
-         *            dir1\.\dir2 == dir1\dir2
-         * 
-         *    (2) \\ -> \
-         *      Double directory separators are treated as a single directory separator,
-         *      so, for example, this is an identity:
-         * 
-         *          f:\dir1\\dir2 == f:\dir1\dir2
-         * 
-         *      The single exemption is for UNC path names, like this:
-         * 
-         *          \\server\share != \server\share
-         * 
-         *      This case is handled by isUncPath in
-         *      a prior step.
-         *
-         *  (3) **\** -> **
-         */
+        /// <summary>
+        /// Given an index directly after a directory separator,
+        /// iteratively skip three cases involving directory separators:
+        ///
+        ///  (1) \.\ -> \
+        ///     This is an identity, so for example, these two are equivalent,
+        ///
+        ///         dir1\.\dir2 == dir1\dir2
+        /// 
+        ///     (2) \\ -> \
+        ///         Double directory separators are treated as a single directory separator,
+        ///         so, for example, this is an identity:
+        ///
+        ///             f:\dir1\\dir2 == f:\dir1\dir2
+        ///
+        ///         The single exemption is for UNC path names, like this:
+        ///
+        ///             \\server\share != \server\share
+        /// 
+        ///         This case is handled by isUncPath in
+        ///         a prior step.
+        ///
+        /// (3) \**\**\ -> \**\
+        ///              This is an identity, so for example, these two are equivalent,
+        ///
+        ///                 dir1\**\**\ == dir1\**\
+        /// </summary>
+        /// <returns>The first index of a non-collapsible character.</returns>
         private static int IndexOfNextNonCollapsibleChar(string str, int startIndex)
         {
             if (startIndex > 0 && !FileUtilities.IsAnySlash(str[startIndex - 1]))
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -54,6 +54,23 @@ namespace Microsoft.Build.Shared
 
         private readonly GetFileSystemEntries _getFileSystemEntries;
 
+        private static class FileSpecRegexParts
+        {
+            internal const string BeginningOfLine = "^(?<FIXEDDIR>";
+            internal const string FixedDirWildcardDirSeparator = ")(?<WILDCARDDIR>";
+            internal const string WildcardDirFilenameSeparator = ")(?<FILENAME>";
+            internal const string EndOfLine = ")$";
+
+            internal const string AnyNonSeparator = @"[^/\\]*";
+            internal const string AnySingleCharacterButDot = @"[^\.].";
+            internal const string AnythingButDot = @"[^\.]*";
+            internal const string DirSeparator = @"[/\\]+";
+            internal const string LeftDirs = @"((.*/)|(.*\\)|())";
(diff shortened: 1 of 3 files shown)


src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
+         */
+        private static bool IsLegalFileSpec(string fixedDirectoryPart, string wildcardDirectoryPart, string filenamePart) =>
+            !HasDotDot(wildcardDirectoryPart)
+            && !HasMisplacedRecursiveOperator(wildcardDirectoryPart)
+            && !HasMisplacedRecursiveOperator(filenamePart)
+            && !HasTags(fixedDirectoryPart)
+            && !HasTags(wildcardDirectoryPart)
+            && !HasTags(filenamePart);
+
+        private static bool HasDotDot(string str)
```
ccastanedaucf: Probably overkill but ```IndexOf(string)``` would come up as an extra 10-20 ms in the profiler

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
-            foreach (var separator in directorySeparatorStrings)
+                if (isRecursiveOperator && !isSurroundedBySlashes)
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        private static bool HasTags(string str)
```
ccastanedaucf: Assuming this check isn't needed anymore but left it in since it'll break some tests expecting an invalid itemspec

AndyGerlicher:
Ran the perf test suite on this, looks great!

### DesignTimeBuild Time (ms) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :red_circle: | yes | 36.1101 -> 36.3213 (0.585%)
DotnetWebProject | :ok_hand: | no | 229.4107 -> 231.434 (0.882%)
DotnetMvcProject | :ok_hand: | no | 234.3295 -> 234.5659 (0.101%)
Picasso | :white_check_mark: | yes | 1526.6705 -> 1512.6367 (-0.919%)
SmallP2POldCsproj | ::ok_hand: | no | 67.1613 -> 67.0383 (-0.183%)
SmallP2PNewCsproj | :white_check_mark: | yes | 643.7473 -> 617.3086 (-4.107%)
LargeP2POldCsproj | :white_check_mark: | yes | 10979.2714 -> 10670.8274 (-2.809%)
OrchardCore | :white_check_mark: | yes | 47947.0214 -> 45360.8638 (-5.394%)

### DesignTimeBuild Memory (bytes) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | ::ok_hand: | no | 6247020 -> 6246936 (-0.001%)
DotnetWebProject | :ok_hand: | no | 21732236 -> 22759441 (4.727%)
DotnetMvcProject | :ok_hand: | no | 20162385 -> 20860295 (3.461%)
Picasso | ::ok_hand: | no | 174440870 -> 174123575 (-0.182%)
SmallP2POldCsproj | :ok_hand: | no | 8441062 -> 8448840 (0.092%)
SmallP2PNewCsproj | :white_check_mark: | yes | 92625486 -> 90286415 (-2.525%)
LargeP2POldCsproj | :ok_hand: | no | 1052420434 -> 1066704640 (1.357%)
OrchardCore | :ok_hand: | no | 2430658337 -> 2458709300 (1.154%)

### SerialEvaluationIsolatedContext Time (ms) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :white_check_mark: | yes | 35.1401 -> 33.7489 (-3.959%)
DotnetWebProject | :white_check_mark: | yes | 48.8903 -> 43.7387 (-10.537%)
DotnetMvcProject | :white_check_mark: | yes | 56.6676 -> 49.1987 (-13.18%)
Picasso | :white_check_mark: | yes | 297.6602 -> 292.9195 (-1.593%)
SmallP2POldCsproj | :white_check_mark: | yes | 50.9343 -> 50.013 (-1.809%)
SmallP2PNewCsproj | :white_check_mark: | yes | 204.5492 -> 187.2366 (-8.464%)
LargeP2POldCsproj | :white_check_mark: | yes | 868.5078 -> 843.8699 (-2.837%)
OrchardCore | :white_check_mark: | yes | 3137.9839 -> 2882.398 (-8.145%)

### SerialEvaluationIsolatedContext Memory (bytes) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :white_check_mark: | yes | 5653767 -> 5621361 (-0.573%)
DotnetWebProject | :white_check_mark: | yes | 7626169 -> 7528520 (-1.28%)
DotnetMvcProject | :white_check_mark: | yes | 8248635 -> 8131113 (-1.425%)
Picasso | :ok_hand: | no | 10284682 -> 11156381 (8.476%)
SmallP2POldCsproj | :white_check_mark: | yes | 6886863 -> 6877999 (-0.129%)
SmallP2PNewCsproj | ::ok_hand: | no | 13197814 -> 12985334 (-1.61%)
LargeP2POldCsproj | :white_check_mark: | yes | 34802172 -> 34710996 (-0.262%)
OrchardCore | :ok_hand: | no | 72407100 -> 73282998 (1.21%)

### SerialEvaluationSharedContextSecondRun Time (ms) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :ok_hand: | no | 29.8731 -> 29.9022 (0.097%)
DotnetWebProject | :red_circle: | yes | 37.0155 -> 37.5264 (1.38%)
DotnetMvcProject | :white_check_mark: | yes | 39.3031 -> 39.1849 (-0.301%)
Picasso | :white_check_mark: | yes | 216.6768 -> 215.2372 (-0.664%)
SmallP2POldCsproj | ::ok_hand: | no | 41.3459 -> 41.2791 (-0.162%)
SmallP2PNewCsproj | :white_check_mark: | yes | 132.0347 -> 130.2436 (-1.357%)
LargeP2POldCsproj | :red_circle: | yes | 642.5098 -> 646.3364 (0.596%)
Generated_100_100_v150 | :ok_hand: | no | 1135.5954 -> 1137.4028 (0.159%)
OrchardCore | :white_check_mark: | yes | 1565.6139 -> 1554.1346 (-0.733%)
Roslyn | :ok_hand: | no | 2593.743 -> 2595.423 (0.065%)
WebLargeCore | :red_circle: | yes | 1797.4158 -> 1831.1872 (1.879%)

### SerialEvaluationSharedContextSecondRun Memory (bytes) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :ok_hand: | no | 5518280 -> 5518300 (0%)
DotnetWebProject | :ok_hand: | no | 6833885 -> 6833912 (0%)
DotnetMvcProject | :white_check_mark: | yes | 7267657 -> 7259947 (-0.106%)
Picasso | ::ok_hand: | no | 34843396 -> 34119966 (-2.076%)
SmallP2POldCsproj | ::ok_hand: | no | 6836973 -> 6836915 (-0.001%)
SmallP2PNewCsproj | :white_check_mark: | yes | 22398116 -> 22382209 (-0.071%)
LargeP2POldCsproj | :red_circle: | yes | 30665857 -> 31686302 (3.328%)
Generated_100_100_v150 | :ok_hand: | no | 42871034 -> 42929289 (0.136%)
OrchardCore | :white_check_mark: | yes | 77939970 -> 71980056 (-7.647%)
Roslyn | :white_check_mark: | yes | 82273267 -> 79369584 (-3.529%)
WebLargeCore | :white_check_mark: | yes | 65633344 -> 64620796 (-1.543%)

### SerialEvaluationSharedContext Time (ms) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :white_check_mark: | yes | 34.268 -> 32.7927 (-4.305%)
DotnetWebProject | :white_check_mark: | yes | 47.511 -> 42.9607 (-9.577%)
DotnetMvcProject | :white_check_mark: | yes | 53.2443 -> 48.1227 (-9.619%)
Picasso | :white_check_mark: | yes | 246.9161 -> 241.9223 (-2.022%)
SmallP2POldCsproj | :white_check_mark: | yes | 45.1364 -> 44.5008 (-1.408%)
SmallP2PNewCsproj | :white_check_mark: | yes | 169.7934 -> 157.7443 (-7.096%)
LargeP2POldCsproj | :white_check_mark: | yes | 697.9884 -> 686.8751 (-1.592%)
Generated_100_100_v150 | ::ok_hand: | no | 1148.9547 -> 1147.1287 (-0.159%)
OrchardCore | :white_check_mark: | yes | 2127.2808 -> 1887.7685 (-11.259%)
Roslyn | :white_check_mark: | yes | 3122.5227 -> 2939.8678 (-5.85%)
WebLargeCore | :white_check_mark: | yes | 2241.0077 -> 2105.3806 (-6.052%)

### SerialEvaluationSharedContext Memory (bytes) (src\msbuild.A vs src\msbuild.B)
:--- | :--- | :--- | :---
DotnetConsoleProject | :white_check_mark: | yes | 5723313 -> 5690329 (-0.576%)
DotnetWebProject | :white_check_mark: | yes | 7686122 -> 7576491 (-1.426%)
DotnetMvcProject | :white_check_mark: | yes | 8296917 -> 8178579 (-1.426%)
Picasso | ::ok_hand: | no | 10648360 -> 9614129 (-9.713%)
SmallP2POldCsproj | :white_check_mark: | yes | 6886687 -> 6886148 (-0.008%)
SmallP2PNewCsproj | :ok_hand: | no | 11986035 -> 13037796 (8.775%)
LargeP2POldCsproj | :white_check_mark: | yes | 32241482 -> 32047966 (-0.6%)
Generated_100_100_v150 | :white_check_mark: | yes | 43695407 -> 43508549 (-0.428%)
OrchardCore | ::ok_hand: | no | 64591140 -> 62002704 (-4.007%)
Roslyn | :ok_hand: | no | 82565309 -> 84794605 (2.7%)
WebLargeCore | :ok_hand: | no | 66186877 -> 67674628 (2.248%)

src/Shared/FileMatcher.cs
ccastanedaucf: It's a behavior change since <: and :> would be now allowed in an the filespec, I think that seems reasonable

rainersigwald: Yeah, I guess they're illegal now because they're illegal win32 FS characters so we'd fall back anyway? But when you remove most of the checks things get super confusing?

src/Shared/FileMatcher.cs
```diff
@@ -54,6 +54,23 @@ internal class FileMatcher
...
 
+        private static class FileSpecRegexParts
+        {
+            internal const string BeginningOfLine = "^(?<FIXEDDIR>";
+            internal const string FixedDirWildcardDirSeparator = ")(?<WILDCARDDIR>";
+            internal const string WildcardDirFilenameSeparator = ")(?<FILENAME>";
+            internal const string EndOfLine = ")$";
+
+            internal const string AnyNonSeparator = @"[^/\\]*";
+            internal const string AnySingleCharacterButDot = @"[^\.].";
```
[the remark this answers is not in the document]
ccastanedaucf: That's my thought too, first is a negation group and the second is a wildcard

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
+                + FileSpecRegexParts.BeginningOfLine.Length + FileSpecRegexParts.FixedDirWildcardDirSeparator.Length
+                + FileSpecRegexParts.WildcardDirFilenameSeparator.Length + FileSpecRegexParts.EndOfLine.Length
+            );
+            matchFileExpression.Append(FileSpecRegexParts.BeginningOfLine);
+
+            AppendRegularExpressionFromFixedDirectory(matchFileExpression, fixedDirectoryPart);
+            matchFileExpression.Append(FileSpecRegexParts.FixedDirWildcardDirSeparator);
+
+            AppendRegularExpressionFromWildcardDirectory(matchFileExpression, wildcardDirectoryPart);
+            matchFileExpression.Append(FileSpecRegexParts.WildcardDirFilenameSeparator);
```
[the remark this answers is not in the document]
ccastanedaucf: I just wanted to separate the dynamically generated regex parts from the static parts, but the sequence is always the same so it would be fine sticking those in the respective methods.

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
+         * filespec comes in with either "<:" or ":>", return false to
+         * prevent intrusion into the special processing.
+         */
+        private static bool IsLegalFileSpec(string fixedDirectoryPart, string wildcardDirectoryPart, string filenamePart) =>
+            !HasDotDot(wildcardDirectoryPart)
+            && !HasMisplacedRecursiveOperator(wildcardDirectoryPart)
+            && !HasMisplacedRecursiveOperator(filenamePart)
+            && !HasTags(fixedDirectoryPart)
+            && !HasTags(wildcardDirectoryPart)
+            && !HasTags(filenamePart);
```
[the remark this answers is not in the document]
ccastanedaucf: It's small enough not to get picked up by the profiler at all, the only thing that pops up occasionally is ``StringBuilder.ToString()```

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
+        private static void AppendRegularExpressionFromFixedDirectory(StringBuilder regex, string fixedDir)
+        {
             /*
              * Capture the leading \\ in UNC paths, so that the doubled slash isn't
              * reduced in a later step.
              */
-            matchFileExpression.Replace("<:fixeddir:><:dirseparator:><:dirseparator:>", "<:fixeddir:><:uncslashslash:>");
+            bool isUncPath = fixedDir.Length > 1
+                             && FileUtilities.IsAnySlash(fixedDir[0])
+                             && FileUtilities.IsAnySlash(fixedDir[1]);
```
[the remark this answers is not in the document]
ccastanedaucf: I kept it here for sake of behavior consistency with ```<:fixeddir:><:dirseparator:><:dirseparator:>", "<:fixeddir:><:uncslashslash:>```. It's definitely wrong outside of Windows, I'll have to see if any tests break on Linux.

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
+         *      This case is handled by isUncPath in
+         *      a prior step.
+         *
+         *  (3) **\** -> **
+         */
+        private static int IndexOfNextNonCollapsibleChar(string str, int startIndex)
+        {
+            int i = startIndex;
+            bool isFound = false;
+            bool isPrevRecursiveOperator = i > 2 && str[i - 3] == '*' && str[i - 2] == '*';
```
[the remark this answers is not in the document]
ccastanedaucf: Correct since ```\**\``` or ```**\``` will converted to the regex equivalent, and skipping over the consumed characters lands you after the ```\``` when calling this method with the offset and ```i + 1```. Although I just thought of possibly negating the ```(i + 1) and starting at the first wildcard.

src/Shared/FileMatcher.cs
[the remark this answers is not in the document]
ccastanedaucf: Oh I see, ```//``` can't be used to start a UNC path. So the condition should really be, Unix always false, Windows true if starting with '''\\'''.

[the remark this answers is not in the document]
rainersigwald: Yeah, that sounds right to me.

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1111,280 @@ private static string RegularExpressionFromFileSpec
...
             }
 
-            /*
-             * Its not legal for there to be a ".." after a wildcard.
-             */
-            if (wildcardDirectoryPart.Contains(dotdot))
+            var matchFileExpression = new StringBuilder(
+                fixedDirectoryPart.Length + wildcardDirectoryPart.Length + filenamePart.Length
+                + FileSpecRegexParts.BeginningOfLine.Length + FileSpecRegexParts.FixedDirWildcardDirSeparator.Length
+                + FileSpecRegexParts.WildcardDirFilenameSeparator.Length + FileSpecRegexParts.EndOfLine.Length
```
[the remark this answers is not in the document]
ccastanedaucf: What should be under the ```#if DEBUG``` check? All it's uses I see are split between throwing exceptions, Debug.Assert, or writing to console.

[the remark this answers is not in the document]
rainersigwald: I was thinking something like

```c#
#if DEBUG
   ErrorUtilities.VerifyThrow(PrecomputedConstant == FileSpecRegexParts.BeginningOfLine.Length + FileSpecRegexParts.FixedDirWildcardDirSeparator.Length
                + FileSpecRegexParts.WildcardDirFilenameSeparator.Length + FileSpecRegexParts.EndOfLine.Length, "Checked-in length of known regex components differs from computed length. Update checked-in constant.");
#endif
```

But with a better name than `PrecomputedConstant`.

AndyGerlicher:
Looks great. I would like to know if there are any tests you could add to increase coverage or at minimum I'd like you to add some `///` comments on the new methods while you have context on what they do.

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1117,245 @@ private static string RegularExpressionFromFileSpec
...
-                isLegalFileSpec = false;
-                return String.Empty;
+                return string.Empty;
             }
 
-            /*
-             * Its not legal for there to be a ".." after a wildcard.
-             */
-            if (wildcardDirectoryPart.Contains(dotdot))
+            var matchFileExpression = StringBuilderCache.Acquire(FileSpecRegexInitialBufferSize);
```
cdmihai: You could use our own ReuseableStringBuilder? That way when we replace it we replace it everywhere consistently :)

cdmihai:
`<:` and `:>` are valid filespec substrings now right? You can add a test for it. Linux for now :)

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1117,245 @@ private static string RegularExpressionFromFileSpec
...
+        {
+            regex.Append(FileSpecRegexParts.BeginningOfLine);
 
             /*
              * Capture the leading \\ in UNC paths, so that the doubled slash isn't
              * reduced in a later step.
              */
-            matchFileExpression.Replace("<:fixeddir:><:dirseparator:><:dirseparator:>", "<:fixeddir:><:uncslashslash:>");
+            bool isUncPath = NativeMethodsShared.IsWindows && fixedDir.Length > 1
+                             && fixedDir[0] == '\\' && fixedDir[1] == '\\';
```
cdmihai: Why not use Microsoft.Build.Shared.FileUtilities.IsUNCPath

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1117,245 @@ private static string RegularExpressionFromFileSpec
...
-            matchFileExpression.Replace(".", "\\.");
-            matchFileExpression.Replace("[", "\\[");
-            matchFileExpression.Replace("?", "\\?");
-            matchFileExpression.Replace("^", "\\^");
-            matchFileExpression.Replace("{", "\\{");
-            matchFileExpression.Replace("|", "\\|");
+            while (!isNonCollapsibleCharFound && i < str.Length)
+            {
+                bool isSeparator = FileUtilities.IsAnySlash(str[i]);
+                bool isRelativeSeparator = i < str.Length - 1 && str[i] == '.' && FileUtilities.IsAnySlash(str[i + 1]);
```
cdmihai: Shouldn't this check for `/./` instead of `./`? What if the path contains a directory ending in `.`?

src/Shared/FileMatcher.cs
ccastanedaucf: That one has different behavior. It returns false if any other ```\``` is in the string which seems like the wrong behavior for this looking at ```FileMatcher_Tests```, and also seems like the wrong behavior for determining if something is a UNC path but it's been part of ```NormalizePath()``` forever.

src/Shared/FileMatcher.cs
ccastanedaucf: Prev slash is implicit thanks to the first check in the method

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1119,281 @@ private static string RegularExpressionFromFileSpec
...
-             *      A "\." at the end of a line is equivalent to nothing, so:
-             * 
-             *          dir1\dir2\. == dir1\dir2             *
-             */
-            int sizeBefore;
-            do
+            for (int i = startIndex; i < fixedDir.Length; i = IndexOfNextNonCollapsibleChar(fixedDir, i + 1))
             {
-                sizeBefore = matchFileExpression.Length;
+                AppendRegularExpressionFromChar(regex, fixedDir[i]);
```
cdmihai: This method also checks for characters that should never appear in the fixed directory party, which could make it confusing to understand. How about composing `AppendRegularExpressionFromChar` from 2 methods, one that operators on non-wildcard characters and one that operators on wildcard characters?

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1119,281 @@ private static string RegularExpressionFromFileSpec
...
+        ///
+        /// (3) \**\**\ -> \**\
+        ///              This is an identity, so for example, these two are equivalent,
+        ///
+        ///                 dir1\**\**\ == dir1\**\
+        /// </summary>
+        /// <returns>The first index of a non-collapsible character.</returns>
+        private static int IndexOfNextNonCollapsibleChar(string str, int startIndex)
+        {
+            if (startIndex > 0 && !FileUtilities.IsAnySlash(str[startIndex - 1]))
```
cdmihai: Can `startIndex-1` never be `.` or `*`? I'd add a precondition to document this.

src/Shared/FileMatcher.cs
```diff
@@ -54,6 +54,29 @@ internal class FileMatcher
 
         private readonly GetFileSystemEntries _getFileSystemEntries;
 
+        private static class FileSpecRegexParts
+        {
+            internal const string BeginningOfLine = "^(?<FIXEDDIR>";
+            internal const string FixedDirWildcardDirSeparator = ")(?<WILDCARDDIR>";
```
cdmihai: Instead of appending `)(`, why not have the previous phases add the closing parenthesis themselves?

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1119,281 @@ private static string RegularExpressionFromFileSpec
...
+            }
+            int startIndex = hasRecursiveOperatorAtStart ? IndexOfNextNonCollapsibleChar(wildcardDir, 3) : 0;
 
-                ErrorUtilities.VerifyThrow(matchFileExpression.Length <= sizeBefore,
-                    "Expression reductions cannot increase the length of the expression.");
-            } while (matchFileExpression.Length < sizeBefore);
+            for (int i = startIndex; i < wildcardDir.Length; i = IndexOfNextNonCollapsibleChar(wildcardDir, i + 1))
+            {
+                char ch = wildcardDir[i];
+                bool isRecursiveOperator = i < wildcardDir.Length - 2 && wildcardDir[i + 1] == '*' && wildcardDir[i + 2] == '*';
```
cdmihai: nit: seems simpler to check `i` and `i + 1` rather than `i + 1` and `i + 2`

src/Shared/FileMatcher.cs
```diff
@@ -1094,215 +1119,281 @@ private static string RegularExpressionFromFileSpec
...
-             *        fixed-directory + **\
-             *        \**\
-             *        **\**
-             * 
-             */
-            do
+                if (isRecursiveOperator)
+                {
+                    regex.Append(FileSpecRegexParts.MiddleDirs);
+                    i += 3;
```
cdmihai: Asserts that the recursive chars are surrounded by file separators would help with documentation.

src/Shared/FileMatcher.cs
ccastanedaucf: It turns into 3 methods since slashes also never appear in filenames, so it'd really be a separate method for each part only sharing the ```sb.Append()``` of escaped and normal characters, and at that point I'd just stick it in the parent call. So code duplication vs clarity?

ccastanedaucf:
I'm gonna see about getting better unit tests in since while there's hundreds of tests indirectly cover this method, there's nothing directly testing it. I think ```GetFileSpecInfo``` is the appropriate level since its the first parent, already internal, and has no test coverage.

src/Shared/FileMatcher.cs
cdmihai: Yea, no code duplication :)
One option would be to make it configurable from the caller. Either a bool parameter for each special character, or an array, or an array of structs containing the input character to be matched and it's corresponding regex to be appended.

src/Shared/FileMatcher.cs
cdmihai: Yup, what if the starting string is `./` or `**`, then the method wouldn't move the cursor over them because startindex is not a slash.

src/Shared/FileMatcher.cs
cdmihai: Yes, looks more clear. Since the stringbuilder is hopefully presized, shouldn't be that much of a hit. You could also call `sb.EnsureCapacity(sb.length + WildCardGroupStart.Length + GroupEnd.Start)` to control eventual resizes.

ccastanedaucf:
Caught a bug involving trailing ```.``` and bigger one where in a recursive sequence such as ```/.//**/``` with more than a single leading slash gets treated as a directory separator instead of a recursive operator. Surprised there wasn't a single test that used a similar case.

davkean:
@ccastanedaucf Awesome work on this!

Issue #3653, opened 2018-08-25
Cannot compile satellite resource assemblies for Windows' pseudo-locales


martincostello:
### Steps to reproduce

To replicate view the https://github.com/martincostello/msbuild-pseudo-locale-repro GitHub repo and follow the instructions in the `README`.

### Expected  behavior

The compilation output includes the following files:

`./PseudoLocales/bin/Debug/netcoreapp2.1/PseudoLocales.dll`.
`./PseudoLocales/bin/Debug/netcoreapp2.1/PseudoLocales.pdb`.
`./PseudoLocales/bin/Debug/netcoreapp2.1/fr/PseudoLocales.resources.dll`.
`./PseudoLocales/bin/Debug/netcoreapp2.1/qps-ploc/PseudoLocales.resources.dll`.

### Actual behavior

The compilation output includes the following files:

`./PseudoLocales/bin/Debug/netcoreapp2.1/PseudoLocales.dll`.
`./PseudoLocales/bin/Debug/netcoreapp2.1/PseudoLocales.pdb`.
`./PseudoLocales/bin/Debug/netcoreapp2.1/fr/PseudoLocales.resources.dll`.

The following file is not emitted:

`./PseudoLocales/bin/Debug/netcoreapp2.1/qps-ploc/PseudoLocales.resources.dll`.

### Environment data
`msbuild /version` output:

```
Microsoft (R) Build Engine version 15.8.166+gd4e8d81a88 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

15.8.166.59604
```

OS info:

```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.401
 Commit:    91b1c13032

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.401\

Host (useful for support):
  Version: 2.1.3
  Commit:  124038c13e
```

martincostello:
Raised PR with suggested fix as #3654.

PR #3654, opened 2018-08-25
Support Windows' pseudo-locales (#3653)

martincostello:
As described in #3653, MSBuild does not produce satellite resource assemblies for Windows' [pseudo-locales](https://docs.microsoft.com/en-gb/windows/desktop/Intl/pseudo-locales) (at least on systems running Windows 10, Version 1803 and later).

This is because they are not returned by the OS as supported cultures. From [the documentation](https://docs.microsoft.com/en-gb/windows/desktop/Intl/using-pseudo-locales-for-localization-testing):

> For Windows 10, version 1803, editing the Windows Registry like this has no effect. But you can still call the non-enumerating NLS APIs with the names of the pseudo-locales (see the code examples above) to populate your user interface (UI).

The ability to do this is useful where developers of an application wish to use the built-in Windows pseudo-locales to test their own applications' localisation support.

This PR special-cases the 4 pseudo-locales in `CultureInfoCache` so that they are considered to be valid cultures. This then allows the `AssignCulture` task to populate the `Culture` and `WithCulture` metadata for any `EmbeddedResource` item for these cultures so that a satellite resource assembly is compiled for the application to use.

The relevant bits of code this change affects are shown below:

https://github.com/Microsoft/msbuild/blob/7125b9d66637d2babc88087b28cd4272133bd793/src/Tasks/AssignCulture.cs#L139-L140

https://github.com/Microsoft/msbuild/blob/d42d3504057ef2b88dd4f68c4bfc5591371bd6fe/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2912-L2917

https://github.com/Microsoft/msbuild/blob/d42d3504057ef2b88dd4f68c4bfc5591371bd6fe/src/Tasks/Microsoft.Common.CurrentVersion.targets#L3592

This PR also updates the values in the `HardcodedCultureNames` property of `CultureInfoCache` with the 4 pseudo-locales and an updated list of values generated by my local Windows 10 OS.

The original source of the list has been removed (aspnet/Localization#323) as well as the tool that generated it (aspnet/Localization#130), so I re-generated it using that code locally in a throw-away project. I've only added _new_ entries generated by the list. Any values that weren't generated I've left in place.

Compiling [this repro application](https://github.com/martincostello/msbuild-pseudo-locale-repro) with the changes made to MSBuild by this PR, the `qps-ploc/PseudoLocales.resources.dll` satellite resource assembly is correctly compiled and used by the application when run for that culture.


src/Tasks/CultureInfoCache.cs
```diff
@@ -35,6 +35,20 @@ static CultureInfoCache()
...
+            // These pseudo-locales are available in versions of Windows from Vista and later.
+            // However, from Windows 10, version 1803, they are not returned when enumerating the
+            // installed cultures, even if the registry keys are set. Therefore, add them to the list manually.
+            var pseudoLocales = new[] { "qps-ploc", "qps-ploca", "qps-plocm", "qps-Latn-x-sh" };
+
+            foreach (string pseudoLocale in pseudoLocales)
+            {
+                if (!ValidCultureNames.Contains(pseudoLocale))
+                {
+                    ValidCultureNames.Add(pseudoLocale);
```
rainersigwald: Since it's a `HashSet`, could you just do this unconditionally?

Actually, you could probably just do `ValidCultureNames.UnionWith(pseudoLocales)`. Though there's a `foreach` above so this being implemented the same way is nice.

src/Tasks/CultureInfoCache.cs
```diff
@@ -46,120 +60,286 @@ internal static bool IsValidCultureString(string name)
...
-        
+
 #if !FEATURE_CULTUREINFO_GETCULTURES
-        // copied from https://github.com/aspnet/Localization/blob/5e1fb16071affd15f15b9c732833f3ae2ac46e10/src/Microsoft.Framework.Globalization.CultureInfoCache/CultureInfoList.cs
-        // removed the empty string from the list
+        // Copied from https://github.com/aspnet/Localization/blob/5e1fb16071affd15f15b9c732833f3ae2ac46e10/src/Microsoft.Framework.Globalization.CultureInfoCache/CultureInfoList.cs
+        // Regenerated using the tool (removed by https://github.com/aspnet/Localization/pull/130)
+        //   * Removed the empty string from the list
+        //   * Any cultures not present when regenerated were retained
+        //   * Added the Windows pseudo-locales
```
rainersigwald: Should these be added to this list when they're going to be added above to handle the dynamic list-of-cultures case?

src/Tasks.UnitTests/AssignCulture_Tests.cs
```diff
@@ -192,6 +192,32 @@ public void Regress283991()
             Assert.Equal(0, t.AssignedFilesWithCulture.Length);
             Assert.Equal(1, t.AssignedFilesWithNoCulture.Length);
         }
+
+        /*
+        * Method:   PseudoLocalization
```
rainersigwald: Nit: can you make this a doc comment like the others in this file?

src/Tasks/CultureInfoCache.cs
martincostello: I was just being conservative in case a different (older) version of Windows to the one I'm running did return the pseudo locales and then threw an exception if it was already in there when it was added. Happy to amend if adding when already present would be safe.

Do you want me to retain the `foreach` or use the `UnionWith()`?

src/Tasks/CultureInfoCache.cs
martincostello: When `HardcodedCultureNames` is used the method returns early on line 30. I can either rework the `#if !FEATURE_CULTUREINFO_GETCULTURES` to not return early and remove them from the hard-coded list, or leave it as-is.

rainersigwald: Ah, so it is. This is great, then.

src/Tasks.UnitTests/AssignCulture_Tests.cs
martincostello: On inspection of the file, they're all `/* */` format except the one immediately before this new test. Happy to update to `///` format if you prefer though.

src/Tasks/CultureInfoCache.cs
rainersigwald: `HashSet<T>.Add` can be called for the same item many times. Its [return value](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=netframework-4.6) indicates whether it was actually added or already present, but we don't care in this case, so it's fine to ignore it. I prefer this because it's easier to read and reason about, and reduces indenting.

Let's just keep the `foreach`. It's consistent and evidently that's all `UnionWith` does anyway: https://github.com/dotnet/corefx/blob/58a310782b805df19a3939b12544c39296db5ed3/src/System.Collections/src/System/Collections/Generic/HashSet.cs#L507-L518

src/Tasks.UnitTests/AssignCulture_Tests.cs
rainersigwald: No, consistency with the majority is the right call. Naturally the one I looked at was the oddball 🤷‍♂️

src/Tasks/CultureInfoCache.cs
martincostello: Sorry, not quite sure from the comment. So do you just want me to drop the `Contains()`?

rainersigwald:
At the moment, we have `vs15.9` and `master` is for 16.0 and .NET Core 3.0. We don't have a fancy auto-merger at the moment, but we live without. We'll see what @AndyGerlicher thinks about 15.9 vs 16; I think we still have the option but we might not want to take the risk.

src/Tasks/CultureInfoCache.cs
rainersigwald: Correct.

martincostello: Fixed in ce49c38a1caf238fda6a3e7ea14b112b8cf4836f.

AndyGerlicher:
@martincostello This looks great, thanks for the contribution! Can you rebase to vs15.9 branch? That way we can get this in 15.9 / .NET Core 2.2 release.

martincostello:
I've updated the base branch for the PR to `vs15.9`, but looks like I need to tidy it up a bit more first.

PR #3669, opened 2018-08-29
Enumeration key disambiguates out of cone relative globs

### LINKED ISSUE #3644: MsBuildCacheFileEnumerations can produce incorrect results
### Steps to reproduce

Zipped repro: [repro.zip](https://github.com/Microsoft/msbuild/files/2308369/repro.zip)

Command line
```
msbuild /restore & bin\Debug\net46\repro.exe
```

Notably, two projects have the following:
```
  <ItemGroup>
    <Foo Include="..\Schema Objects\**\*.txt" />
  </ItemGroup>
```

And `MsBuildCacheFileEnumerations` is set.

Interestingly, If I change the structure so that the "Schema Objects" dir is adjacent to the project, it seems to work as expected.

### Expected  behavior
```
Project A: E:\tmp\MSBuildGlobRepro\a\Schema Objects\Bar\Baz\a.txt
Project B: E:\tmp\MSBuildGlobRepro\b\Schema Objects\Bar\Baz\b.txt
```

### Actual behavior
```
Project A: E:\tmp\MSBuildGlobRepro\a\Schema Objects\Bar\Baz\a.txt
Project B: E:\tmp\MSBuildGlobRepro\b\Schema Objects\Bar\Baz\a.txt
```

Note that project A ends up with project A's inputs.

### Environment data
MSBuild 15.8.166
OS info: Windows

cdmihai:
Fixes #3644

### DIFF
--- a/src/Utilities/Microsoft.Build.Utilities.csproj
+++ b/src/Utilities/Microsoft.Build.Utilities.csproj
@@ -137,6 +137,9 @@
     <Compile Include="..\Shared\ReadOnlyEmptyCollection.cs">
       <Link>Shared\ReadOnlyEmptyCollection.cs</Link>
     </Compile>
+    <Compile Include="..\Shared\ReuseableStringBuilder.cs" >
+      <Link>Shared\ReuseableStringBuilder.cs</Link>
+    </Compile>
     <Compile Include="..\Shared\ReadOnlyEmptyDictionary.cs">
       <Link>Shared\Collections\ReadOnlyEmptyDictionary.cs</Link>
     </Compile>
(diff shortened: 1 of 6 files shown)


src/Utilities/Microsoft.Build.Utilities.csproj
```diff
@@ -137,6 +137,9 @@
     <Compile Include="..\Shared\ReadOnlyEmptyCollection.cs">
       <Link>Shared\ReadOnlyEmptyCollection.cs</Link>
     </Compile>
+    <Compile Include="..\Shared\ReuseableStringBuilder.cs" >
```
cdmihai: @ccastanedaucf, I also had to add the reference :)

AndyGerlicher:
This should target `vs15.9` branch.

cdmihai:
Ah, the failure happens because Path.Combine throws for illegal characters. I guess using the IO.Redist types would have helped here. I'll ... do a manual combine instead :(

cdmihai:
@rainersigwald @AndyGerlicher 
Fixed the test failures. I'll remove the try catch once we move to 472 and IO.Redist.

PR #3673, opened 2018-08-29
Remove scatter files from manifest dependencies for ClickOnce

sidiesen:
[This previous PR](https://github.com/Microsoft/msbuild/pull/3382) introduced a change to use ReferenceCopyLocalPaths files as dependencies to generate the ClickOnce manifest. 
However, if the project contains scatter files that are native assemblies, we incorrectly recognize those files as managed dependencies which causes ClickOnce publishing to fail. (See comments in linked PR.)

This changes removes the scatter files from the _DeploymentManifestDependencies ItemGroup which should fix that issue.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -3989,11 +3989,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         SatelliteAssemblies="@(IntermediateSatelliteAssembliesWithTargetPath);@(ReferenceSatellitePaths)"
         TargetCulture="$(TargetCulture)">
 
-      <Output TaskParameter="OutputAssemblies" ItemName="_DeploymentManifestDependencies"/>
+      <Output TaskParameter="OutputAssemblies" ItemName="_UnfilteredDeploymentManifestDependencies"/>
       <Output TaskParameter="OutputFiles" ItemName="_DeploymentManifestFiles"/>
       <Output TaskParameter="OutputEntryPoint" ItemName="_DeploymentResolvedManifestEntryPoint"/>
 
     </ResolveManifestFiles>
+    
+    <!-- _DeploymentManifestDependencies contains every file that has to be copied locally, we have to filter out scatter files, 
+    <!-- since those aren't actually depencies for the ClickOnce manifest. -->
+    <ItemGroup>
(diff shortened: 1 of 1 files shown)


AndyGerlicher:
@Jinhuafei does this look ok to you?

Jinhuafei:
@AndyGerlicher do you know what is supposed to be included in ReferenceCopyLocalPaths and _ReferenceScatterPaths? To me, the behavior of ReferenceCopyLocalPaths item group is not consistent when resolving content type item.

sidiesen:
@Jinhuafei based on our offline conversation about the inconsistent behavior yesterday I took a closer look and I think the reason why the behavior is different is because of the difference in how ReferenceCopyLocalPaths is computed in the different scenarios we were looking at:
- if the assemblies are added to the project or included as a regular reference to the project, their occurrence in ReferenceCopyLocalPaths is computed by the task ResolveAssemblyReference, which alongside the copy local paths also computes lists of scatter files/satellite assemblies/referenced files etc.
- if the assemblies are added via a nuget package, their occurrence in ReferenceCopyLocalPaths is computed by the task ResolveNuGetPackageAssets, which adds them to the list of copy local files, but does not compute the extra information of which of the files are scatter files/satellite assemblies/etc, which is why we are not able to exclude those in this case.

This seems to be the reason why we could confirm that this fix is still failing with projects that have such native assemblies included via nuget packages.

Looking at the task ResolveNuGetPackageAssets it does have an output parameter named "_NuGetContentItems" which seems to be the closest we can get here in this scenario, however, this output is empty in the repro on my machine, while the native assemblies in the nuget package are clearly defined as content files. @AndyGerlicher can you shed some light as to what "content items" means in regards to the ResolveNuGetPackageAssets task?

xwlantian:
Any reason why not keep use `ReferencePath` instead of `ReferenceCopyLocalPaths` for `_DeploymentReferencePaths`?

xwlantian:
Everything works as expected if reverts the PR change.

Jinhuafei:
@xwlantian the original change is to address [this issue](https://github.com/Microsoft/msbuild/issues/3069). Without the change, you will see missing runtime files in the publish artifacts when .net standard packages are used in the project.

xwlantian:
@Jinhuafei Maybe you should address the missing runtime files in `ResolveNativeReferences` target, I think they should be in either `NativeReferenceFile` or _`DeploymentNativePrerequisite`  which are outputs of `ResolveNativeReferences` target.

PR #3689, opened 2018-09-01
filter CopyLocal files before ClickOnce manifest generation

sidiesen:
With the [recent change](https://github.com/Microsoft/msbuild/pull/3382) for ClickOnce manifest file generation to use copy local files instead of reference files we are pulling in unwanted files into manifest generation that can have side effects. (Satellite assemblies, scatter files and content files.)

This change should filter out most of these and adapt the ResolveManifestFiles so it can handle a corner case where we might accidentally pull in native assemblies from nuget packages that it can't handle.


cdmihai:
@sidiesen this PR seems like it includes the changes from your other PR #3675. Please close the one you don't want.

sidiesen:
@cdmihai sry, forgot to close the other one. 
Please don't merge this PR yet, we identified some additional issues with ClickOnce manifests and there are likely more changes incoming, we're currently awaiting test results for their fixes.

PR #3697, opened 2018-09-04
Add /interactive switch

jeffkl:
Sets built-in property `MSBuildInteractive` and is passed to SDK resolvers
Keeps back compat with existing `NuGetInteractive` global property

### DIFF
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -204,6 +204,8 @@ namespace Microsoft.Build.Evaluation
         /// </summary>
         private ProjectRootElement _lastModifiedProject;
 
+        private readonly bool _interactive;
+
         /// <summary>
         /// Private constructor called by the static Evaluate method.
         /// </summary>
@@ -219,7 +221,8 @@ namespace Microsoft.Build.Evaluation
             ISdkResolverService sdkResolverService,
             int submissionId,
             EvaluationContext evaluationContext,
-            bool profileEvaluation)
+            bool profileEvaluation,
+            bool interactive)
         {
             ErrorUtilities.VerifyThrowInternalNull(data, nameof(data));
             ErrorUtilities.VerifyThrowInternalNull(projectRootElementCache, nameof(projectRootElementCache));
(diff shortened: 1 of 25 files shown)


jeffkl:
I will add some unit tests before merging

AndyGerlicher:
Looks good pending the unit tests.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -251,6 +254,10 @@ internal class Evaluator<P, I, M, D>
             _submissionId = submissionId;
             _evaluationProfiler = new EvaluationProfiler(profileEvaluation);
 
+            // In 15.9 we add support for the global property "NuGetInteractive" to allow SDK resolvers to be interactive.
```
rainersigwald: typo: add**ed**

src/MSBuild/Resources/Strings.resx
```diff
@@ -707,6 +707,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_34_InteractiveSwitch" UESanitized="false" Visibility="Public">
+    <value>  /interactive[:True|False]
+                     Indicates that actions in the build are allowed to
+                     interact with the user.  When enabled, the build can
+                     block indefinitely so it should only be specified
+                     in an environment where a user is present.
```
rainersigwald: Should we invert this guidance to something like

> It should never be specified in a fully automated scenario.

? Feels like a warning on when you shouldn't use it might be better/scarier than guidance on when it should?

jeffkl:
@Microsoft/msbuild-maintainers this is ready for review!

src/MSBuild/Resources/Strings.resx
jeffkl: I really struggled on the wording here.  Should I change it to:

```
Indicates that actions in the build are allowed to
interact with the user.  Do not use this argument
in a fully automated scenario.
```

rainersigwald: Yeah, I think so. But I'm also struggling with this one.

jeffkl: @AndyGerlicher you must decide on the usage text 😄

AndyGerlicher:
```
Indicates that actions in the build are allowed to
interact with the user.  Do not use this argument
in an automated scenario where interactivity is
not expected.
```

That was my stab at the last part, but I don't know... any of these are fine really :). I would probably get rid of the rest though, it's not more helpful.

jeffkl: I like your wording, I've updated it accordingly

Issue #3680, opened 2018-08-31
Initial ProjectGraph interfaces/stubs


dfederm:
Add initial interfaces/stubs to code against for ProjectGraph.

From the design, something similar to:
```
	namespace Microsoft.Build.Graph
	{
	    public class ProjectGraph
	    {
	        // Creates a graph starting at the given project file.
	        public ProjectGraph(string projectFile) { }
	
	        // Creates a graph starting at the given project files, for example all projects in a solution.
	        public ProjectGraph(IEnumerable<string> projectFiles) { }
	
	        // All project nodes in the graph.
	        IReadOnlyCollection<ProjectGraphNode> ProjectNodes { get; }
	    }
	
	    public class ProjectGraphNode
	    {
	        // No public creation.
	        internal ProjectGraphNode() { }
	
	        // Projects which this project references.
	        IReadOnlyCollection<ProjectGraphNode> ReferencedProjects { get; }
	
	        // Projects which reference this project.
	        IReadOnlyCollection<ProjectGraphNode> ReferencingProjects { get; }
	
	        // The evaluated project
	        Project Project { get; }
	    }
}
```

dasMulli:
curious: What's the goal (/ scenarios) of the project graph effort?

dfederm:
@dasMulli Apologies, I've just post the motivation and design here: #3696

PR #3701, opened 2018-09-04
Initial ProjectGraph interfaces/stubs

dfederm:
Closes: #3680

### DIFF
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Build.Exceptions;
+
+namespace Microsoft.Build.Graph
+{
+    /// <summary>
+    /// Represents a graph of evaluated projects.
+    /// </summary>
+    internal sealed class ProjectGraph
+    {
+        /// <summary>
+        /// Constructs a graph starting from the given project file.
+        /// </summary>
+        /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
+        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
+        public ProjectGraph(string entryProjectFile) { }
+
+        /// <summary>
+        /// Constructs a graph starting from the given project files.
+        /// </summary>
+        /// <param name="entryProjectFiles">The project files to use as the entry points in constructing the graph</param>
+        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
+        public ProjectGraph(IEnumerable<string> entryProjectFiles) { }
+
+        /// <summary>
+        /// Get an unordered collection of all project nodes in the graph.
+        /// </summary>
+        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => Array.Empty<ProjectGraphNode>();
+    }
+}
--- a/src/Build/Graph/ProjectGraphNode.cs
+++ b/src/Build/Graph/ProjectGraphNode.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Build.Evaluation;
+
+namespace Microsoft.Build.Graph
+{
+    /// <summary>
+    /// Represents the node for a particular project in a project graph.
+    /// </summary>
+    internal sealed class ProjectGraphNode
+    {
+        // No public creation.
+        internal ProjectGraphNode() { }
+
+        /// <summary>
+        /// Gets an unordered collection of graph nodes for projects which this project references.
+        /// </summary>
+        public IReadOnlyCollection<ProjectGraphNode> ProjectReferences => Array.Empty<ProjectGraphNode>();
+
+        /// <summary>
+        /// Gets the evaluated project represented by this node in the graph.
+        /// </summary>
+        public Project Project => new Project();
+    }
+}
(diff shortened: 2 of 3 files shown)


src/Build/Graph/ProjectGraph.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Build.Exceptions;
+
+namespace Microsoft.Build.Graph
+{
+    /// <summary>
+    /// Represents a graph of evaluated projects.
+    /// </summary>
+    internal class ProjectGraph
```
dfederm: The class is internal for now to avoid anything leaking into the public API prematurely. It wouldn't be the end of the world if it did, but it might be easier to make breaking changes early if it's all internal.

rainersigwald: Sounds good as long as we don't have to do too many additional `InternalsVisibleTo` tricks.

dfederm: Any preference to class vs interface? Interfaces may make for better mocking for UTs, but comes at the cost of another layer of abstraction. Realistically there would probably not be multiple impls of this interface.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Constructs a graph starting from the given project file.
+        /// </summary>
+        /// <param name="projectFile">The project file</param>
+        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
+        public ProjectGraph(string projectFile) { }
```
rainersigwald: I think I'd prefer the name `entryProject` for this parameter. Especially since in VS land it'll likely often be a solution.

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    internal class ProjectGraphNode
+    {
+        // No public creation.
+        internal ProjectGraphNode() { }
+
+        /// <summary>
+        /// Gets an unordered collection of graph nodes for projects which this project references.
+        /// </summary>
+        public IReadOnlyCollection<ProjectGraphNode> ReferencedProjects => Array.Empty<ProjectGraphNode>();
```
rainersigwald: Ok, so.

We spend a lot of time overpronouncing referenc**ed** and referenc**ing** and other similar words. For this formal construction, can we come up with two words that don't sound similar and use them consistently? It could save a lot of confusion in the future.

rainersigwald: Maybe `ProjectReferences` (parallel to what goes into the project file) and . . . ?

Or `DependsOnProjects` and `BeforeProjects` by (bad) extension of the target-graph operators? Possibly more confusing though since edges in the target graph have only one anchor node.

src/Build/Graph/ProjectGraph.cs
dfederm: Agreed. I think once we start needing things to be public (without adding shenanigans / hacks), we'll make them public. I anticipate at least a few changes before that happens though :)

src/Build/Graph/ProjectGraphNode.cs
dfederm: I anticipated naming being difficult which is why this PR is just that by itself :).

In theory we don't *need* a collection referenc**ing** projects since we have the full graph, but I figured it'd just make for easier traversal. If we dropped it though this discussion becomes moot. Personally I think we should keep it though and just figure out a good name somehow.

And to add more options, we could also list these as `Dependencies` instead as a possibly more generic term. Although I like your `ProjectReferences` since it's very clear about how these were constructed. It might box us in in case there some theoretical future way to construct a graph, but I think that would be fairly low probability considering how prevalent project reference items are.

Regarding referenc**ed** vs referenc**ing** (or dependencies vs dependents), I'm completely open to suggestion. I really like your idea of `ProjectReferences` for referenc**ed**, but I'm not sold on what the name for referenc**ing** should be...

QuickBuild uses parent/children (children depend on their parents), but that's at odds with how people tend to visualize these graphs with the "root" at the top and doesn't really align with anything currently in MSBuild terminology-wise.

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Build.Evaluation;
+
+namespace Microsoft.Build.Graph
+{
+    /// <summary>
+    /// Represents the node for a particular project in a project graph.
+    /// </summary>
+    internal class ProjectGraphNode
```
jeffkl: I would vote for `ProjectGraphItem` instead of `ProjectGraphNode`...

rainersigwald: I prefer Node (or Vertex) in keeping with most graph-algo standards I've seen.

src/Build/Graph/ProjectGraphNode.cs
rainersigwald: Every time you say parent/child I get confused and have to think for a second so I'd like to vote against that one :)

Unfortunately, I also don't have a great idea for incoming edges.

* ProjectsThatDependOnThis
* ReverseReferences
* BackReferences

src/Build/Graph/ProjectGraph.cs
dfederm: I like `entryProjectFile` (`entryProject` maybe for an overload that takes a `Project`), but I was hoping that for VS they'd just use the overload that takes multiple projects and not feed us a sln file here. They already know all the projects; no need to force MSBuild to parse the sln all over again.

rainersigwald: I agree that inside VS it makes sense for them to hand us the list of projects. I was thinking mostly of command-line builds of solutions (as opposed to traversal-project-driven builds). Those are super common outside Microsoft.

dfederm: Poop, you're right. Although potentially we could do the metaproj thing and feed that to here instead of the sln itself, but that's an impl detail we can get to when the time comes :)

src/Build/Graph/ProjectGraphNode.cs
dfederm: According to Wikipedia they're usually called "vertices, nodes, or points". I feel like I usually see "node". I usually see "item" or "element" in more of a list collection (gut feel).

PR #3705, opened 2018-09-04
Tarball-build versioning

rainersigwald:
A couple of things came up when looking into the latest Mono tarball build with @directhex. I think these patches make sense in mainline.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -38,7 +38,7 @@
     <XunitOptions Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">$(XunitOptions) -notrait category=nonnetcoreapptests</XunitOptions>
 
     <XUnitRunnerAdditionalArguments>$(XunitOptions)</XUnitRunnerAdditionalArguments>
-    
+
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


src/Directory.Build.targets
```diff
@@ -111,6 +111,18 @@
...
   <Import Project="$([System.IO.Path]::Combine('$(RepoRoot)', 'build', 'GetBuildVersionStub.proj'))" Condition="'$(DisableNerdbankVersioning)' == 'true'" />
 
+  <!-- In circumstances where we aren't using GitVersioning, set
+       versions manually to something vaguely correct. -->
+  <Target Name="OverrideRepoToolsetVersions"
+          AfterTargets="_InitializeAssemblyVersion"
+          Condition="'$(DisableNerdbankVersioning)' == 'true'">
+    <PropertyGroup>
+      <AssemblyVersion>15.1.0.0</AssemblyVersion>
+      <FileVersion>15.8.0.0</FileVersion>
```
cdmihai: Should these be populated from https://github.com/Microsoft/msbuild/blob/master/version.json

rainersigwald:
Obsoleted by #4318

PR #3706, opened 2018-09-04
Log process path, command line and current directory at start of build

yash256:


yash256:
Closes issue https://github.com/Microsoft/msbuild/issues/2810

src/MSBuild/XMake.cs
```diff
@@ -2092,10 +2092,13 @@ out enableProfiler
                         Console.WriteLine(ResourceUtilities.FormatResourceStringStripCodeAndKeyword("PickedUpSwitchesFromAutoResponse", autoResponseFileName));
                     }
 
-                    if (verbosity == LoggerVerbosity.Diagnostic)
+                    if (verbosity >= LoggerVerbosity.Detailed)
                     {
+                        string currentWorikingDirectory = Directory.GetCurrentDirectory();
                         string equivalentCommandLine = commandLineSwitches.GetEquivalentCommandLineExceptProjectFile();
```
AndyGerlicher: It might be useful here to also log the path to dotnet.exe
```C#
#if RUNTIME_TYPE_NETCORE
    var hostProcess = Process.GetCurrentProcess()?.MainModule?.FileName;
    Console.WriteLine($"Dotnet host path: {hostProcess}");
#endif
```

AndyGerlicher:
So this looks good matching the pattern that's already here, but I don't think it will solve the requirement. The log messages will appear on the console, not in the log file which is what Kirill is asking for (to show in the viewer). And this ends up probably a more difficult change. To do it right we have to generate the messages, but then not actually log them until the logging infrastructure is initialized, and I don't think anything else in xmake.cs does this. And the command-line arguments are likely lost after code moves out of xmake.cs. So it might have to be something like passing in deferred messages that get logged first thing when the logger initializes? @rainersigwald am I over complicating this?

rainersigwald:
Unfortunately, I don't think that's overcomplicating it :(

KirillOsenkov:
See my attempt at: https://github.com/microsoft/msbuild/pull/5026

PR #3707, opened 2018-09-04
Add binding redirects for redistributed dlls

cdmihai:
Used the dll versions from what gets into our vsix.


cdmihai:
@tmat

cdmihai:
> when we insert with Microsoft.IO.Redist do we need to add a binding redirect to VS?

Not until something in VS actually takes a dependency on it. AFAIU we need redirects for all our redistributed dlls indifferent of VS to make it easy for tasks sharing these dependencies. But when VS also takes a dependency to one of these we need to then follow VS.

One risk is when we have a redirect to a version, and then VS comes along and adds a dependency and redirect to an older version. This is when we'd have to follow VS and downgrade, which is a small breaking change for tasks depending on a version between VS and ours. But I think it's worth the risk as opposed to prematurely adding our redirects to devenv and thus eagerly increasing the surface of things we have to be manually aware of.

KirillOsenkov:
@tmat pushed the change: https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/2e277ec3dd6ad55d1d6a2a026f943a99895748c5

Will publish an update this week (remind me if I forget).

PR #3711, opened 2018-09-06
Add /isolateProjects argument

dfederm:
Closes: #3687

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/Build/BackEnd/BuildManager/BuildParameters.cs
@@ -209,6 +209,8 @@ namespace Microsoft.Build.Execution
 
         private bool _interactive;
 
+        private bool _isolateProjects;
+
         /// <summary>
         /// Constructor for those who intend to set all properties themselves.
         /// </summary>
@@ -286,6 +288,7 @@ namespace Microsoft.Build.Execution
             WarningsAsMessages = other.WarningsAsMessages == null ? null : new HashSet<string>(other.WarningsAsMessages, StringComparer.OrdinalIgnoreCase);
             _projectLoadSettings = other._projectLoadSettings;
             _interactive = other._interactive;
+            _isolateProjects = other._isolateProjects;
         }
 
 #if FEATURE_THREAD_PRIORITY
@@ -725,6 +728,14 @@ namespace Microsoft.Build.Execution
             set => _interactive = value;
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether projects should build in isolation.
+        /// </summary>
+        public bool IsolateProjects
+        {
+            get => _isolateProjects;
+            set => _isolateProjects = value;
+        }
 
         /// <summary>
         /// Retrieves a toolset.
@@ -775,6 +786,7 @@ namespace Microsoft.Build.Execution
             translator.Translate(ref _logInitialPropertiesAndItems);
             translator.TranslateEnum(ref _projectLoadSettings, (int) _projectLoadSettings);
             translator.Translate(ref _interactive);
+            translator.Translate(ref _isolateProjects);
 
             // ProjectRootElementCache is not transmitted.
             // ResetCaches is not transmitted.
--- a/src/MSBuild/CommandLineSwitches.cs
+++ b/src/MSBuild/CommandLineSwitches.cs
@@ -104,6 +104,7 @@ namespace Microsoft.Build.CommandLine
             ProfileEvaluation,
             RestoreProperty,
             Interactive,
+            IsolateProjects,
             NumberOfParameterizedSwitches,
         }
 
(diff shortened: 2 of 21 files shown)


src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -725,6 +728,11 @@ public bool Interactive
             set => _interactive = value;
         }
 
+        public bool EnableGraphBuild
```
dfederm: I had to make it `public` since `XMake.cs` is in a different dll...

This is the flag future work will hook off of though.

Potentially though if graph builds "wrap" the build manager, then this flag should be renamed something more like "isolated project build" or something since it'll just enable the export enforcement stuff and the scheduling would all be done "outside" of this. If OTOH we have `BuildManager` take a `ProjectGraph`, then this flag is fine here, or perhaps should even not be settable and just be implied when a `ProjectGraph` is supplied.

I expect this to change a little as we continue development, which is why I'm not super happy this needs to be public...

src/MSBuild/CommandLineSwitches.cs
```diff
@@ -237,6 +238,7 @@ bool emptyParametersAllowed
 #if DEBUG
             new ParameterlessSwitchInfo(  new string[] { "waitfordebugger", "wfd" },        ParameterlessSwitch.WaitForDebugger,       null , null             ),
 #endif
+            new ParameterlessSwitchInfo(  new string[] { "graphbuild", "gb" },              ParameterlessSwitch.GraphBuild,            null , null             ),
```
dfederm: Completely open to suggestion for the switch name.

Also, I made it a parameterless switch rather than a parameterized one which takes true|false. I feel like it makes more sense, but I'm open to suggestion here as well.

src/MSBuild/XMake.cs
```diff
@@ -2071,6 +2077,11 @@ bool recursing
                         interactive = ProcessBooleanSwitch(commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.Interactive], defaultValue: true, resourceName: "InvalidInteractiveValue");
                     }
 
+                    if (commandLineSwitches[CommandLineSwitches.ParameterlessSwitch.GraphBuild])
+                    {
+                        enableGraphBuild = true;
```
jeffkl: Make this a parameterized switch and call `ProcessBooleanSwitch()` so that users can have the feature enabled/disabled in a response file but override the value at the command-line

src/MSBuild/CommandLineSwitches.cs
rainersigwald: I think I lean toward something more like `/isolateProjects`. If we really want to expose the graph concept I'd probably go with just `/graph`.

src/MSBuild/Resources/Strings.resx
```diff
@@ -722,6 +722,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_35_GraphBuildSwitch" UESanitized="false" Visibility="Public">
```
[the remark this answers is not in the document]
rainersigwald: This is a reasonable stake in the ground. We should register and point to an `aka.ms` link in the help text that points to the docs page when it's ready, in addition to a succinct description here.

src/MSBuild/Resources/Strings.resx
```diff
@@ -722,6 +722,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <value>/graphBuild
+                     Causes MSBuild to construct a project graph and build it
+                     from the bottom up as opposed to discovering and building
```
rainersigwald: I don't like the phrase "bottom up", but we have plenty of time to tweak this :)

src/MSBuild/Resources/Strings.resx
```diff
@@ -722,6 +722,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_35_GraphBuildSwitch" UESanitized="false" Visibility="Public">
+    <value>/graphBuild
```
rainersigwald: Tests are failing because this didn't copy the two space characters from the other messages. Nice that someone actually added tests for indentation consistency . . .

src/Build/BackEnd/BuildManager/BuildParameters.cs
rainersigwald: I think we can manage the API changes; it's definitely not bad enough to add friction to making forward progress here. Maybe we should consider adding a diff-API-against-last-release test to CI.

src/MSBuild/CommandLineSwitches.cs
dfederm: I'm fine with `/isolateProjects`. It does kind of separate the whole "isolated builds" feature, including exported targets and all that jazz, from the project graph feature, which is probably a good thing but may also cause a little confusion since the features are somewhat related.

Specifically, isolated project builds would use the graph for scheduling purposes when building a set of projects. However there's also the scenario where we're doing a build of 1 project and are somehow importing the exported target results from the project references (eg. the QuickBuild scenario, although I imagine it being used to super fast incremental builds in VS or even command-line as well), which doesn't necessarily have anything to do with the project graph stuff.

src/MSBuild/Resources/Strings.resx
dfederm: Yea, I should run tests before embarrassingly sending out PRs with failing tests :)

Issue #3712, opened 2018-09-07
Multi-appdomain msbuild instances may fail to manage cached configuration files


jeromelaban:
### Steps to reproduce

Create an application that instantiates the build engine in separate app domains, and make lots of build requests.

### Expected  behavior

Builds end properly in each app domain.

### Actual behavior

Builds can randomly fail with this exception: 
```
UNHANDLED EXCEPTIONS FROM PROCESS 15024:
=====================
9/6/2018 4:09:56 PM
System.IO.FileNotFoundException: Could not find file 'C:\Users\build-svc-defpool\AppData\Local\Temp\MSBuild15024\Configuration191.cache'.
File name: 'C:\Users\build-svc-defpool\AppData\Local\Temp\MSBuild15024\Configuration191.cache'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at Microsoft.Build.BackEnd.BuildRequestConfiguration.GetConfigurationTranslator(TranslationDirection direction)
   at Microsoft.Build.BackEnd.BuildRequestConfiguration.RetrieveFromCache()
   at Microsoft.Build.BackEnd.BuildRequestEngine.ActivateBuildRequest(BuildRequestEntry entry)
   at Microsoft.Build.BackEnd.BuildRequestEngine.<>c__DisplayClass39_0.<SubmitBuildRequest>b__0()
   at Microsoft.Build.BackEnd.BuildRequestEngine.<>c__DisplayClass67_0.<QueueAction>b__0()
```

### Environment data
```
C:\s\github\msbuild>msbuild /version
Microsoft (R) Build Engine version 15.8.168+ga8fba1ebd7 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

15.8.168.64424
```

OS info:
msbuild.exe cli, not dotnet CLI, which does not allow for app domains.

PR #3713, opened 2018-09-07
Update FileUtilities to include the current app domain id

jeromelaban:
This avoids race conditions such as #3712, where the field https://github.com/Microsoft/msbuild/blob/0591c15d6c638cad38091fbe625dde968f86748d/src/Shared/FileUtilities.cs#L44 can be set to the same value depending on the current app domain.

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Build.Shared
         {
             if (cacheDirectory == null)
             {
-                cacheDirectory = Path.Combine(Path.GetTempPath(), String.Format(CultureInfo.CurrentUICulture, "MSBuild{0}", Process.GetCurrentProcess().Id));
+                cacheDirectory = Path.Combine(Path.GetTempPath(), String.Format(CultureInfo.CurrentUICulture, "MSBuild{0}-{1}", Process.GetCurrentProcess().Id, AppDomain.CurrentDomain.Id));
             }
 
             return cacheDirectory;


rainersigwald:
Looks like `net46\Microsoft.Build.Engine.UnitTests.dll` timed out on Windows, which I don't think I've seen before. Retrying since this seems unlikely to have caused that.

jeromelaban:
Thanks @rainersigwald !

So far, it's the only issue we've encountered using [Uno.SourceGeneration](https://github.com/nventive/Uno.SourceGeneration) with the many build hours the past 4 years done using this multi-AppDomain technique, and it only appears when the CI servers are under heavy stress. 

But I understand this is a corner case use, which will eventually disappear because .NET Core does not support AppDomains, and I'll have to change the segregation model to support it.

src/Shared/FileUtilities.cs
```diff
@@ -105,7 +105,7 @@ internal static string GetCacheDirectory()
         {
             if (cacheDirectory == null)
             {
-                cacheDirectory = Path.Combine(Path.GetTempPath(), String.Format(CultureInfo.CurrentUICulture, "MSBuild{0}", Process.GetCurrentProcess().Id));
+                cacheDirectory = Path.Combine(Path.GetTempPath(), String.Format(CultureInfo.CurrentUICulture, "MSBuild{0}-{1}", Process.GetCurrentProcess().Id, AppDomain.CurrentDomain.Id));
```
cdmihai: Not related to this PR, but I hope this write isn't a race :)
Theoretically it's not, as a process only builds one project at a time.

PR #3715, opened 2018-09-07
Use reference assemblies

rainersigwald:
We didn't turn these on for a while because we needed our bootstrap compilers to support them, but we should be long past that now. This will speed up incremental builds in the repo.


cdmihai:
SignTool failed. The ref assembly needs signed as well I guess?

rainersigwald:
It shouldn't need to be signed but maybe something thinks it does anyway? I'll look at it in a bit.

rainersigwald:
Arcade does this for us for free.

PR #3716, opened 2018-09-07
Export targets

### LINKED ISSUE #3685: Schedule exported targets in execution
From the design, item defines exported targets: `@(MSBuildExportedTargets)`

Add exported targets into the execution stack such that they run last.

Likely the implementation of this will simply be to push the export targets onto the target execution stack in reverse order before the default and initial targets.

cdmihai:
Resolves #3685


src/Build/BackEnd/BuildManager/BuildRequestData.cs
```diff
@@ -75,6 +75,11 @@ public enum BuildRequestDataFlags
         /// This is especially useful during a restore since some imports might come from packages that haven't been restored yet.
         /// </summary>
         IgnoreMissingEmptyAndInvalidImports = 1 << 6,
+
+        /// <summary>
+        /// Turns on logic for graph based isolated project builds
+        /// </summary>
+        IsolatedProjectBuild = 1 << 7
```
cdmihai: Should this be `ExecuteExportedTargets` instead? I bet we're going to need `IsolatedProjectBuild` as we do more things differently for isolated project builds.

cdmihai: Or instead of using a master switch we use n fine grained switches which get turned on by the Scheduler?

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -657,15 +667,27 @@ public List<string> GetTargetsUsedToBuildRequest(BuildRequest request)
             ErrorUtilities.VerifyThrow(request.ConfigurationId == ConfigurationId, "Request does not match configuration.");
             ErrorUtilities.VerifyThrow(_projectInitialTargets != null, "Initial targets have not been set.");
             ErrorUtilities.VerifyThrow(_projectDefaultTargets != null, "Default targets have not been set.");
+            ErrorUtilities.VerifyThrow(_projectExportTargets != null,  "Export targets have not been set.");
 
             List<string> initialTargets = _projectInitialTargets;
             List<string> nonInitialTargets = (request.Targets.Count == 0) ? _projectDefaultTargets : request.Targets;
 
-            var allTargets = new List<string>(initialTargets.Count + nonInitialTargets.Count);
+            var isSingleProjectBuild = request.IsIsolatedProjectBuild();
```
dfederm: >SingleProject [](start = 18, length = 13)

"isIsolatedBuild" perhaps

src/Build/Definition/Project.cs
```diff
@@ -3052,6 +3052,7 @@ internal Data(Project project, PropertyDictionary<ProjectPropertyInstance> globa
             public ItemDictionary<ProjectItem> Items { get; private set; }
 
             public List<ProjectItemElement> EvaluatedItemElements { get; private set; }
+            public IReadOnlyCollection<string> ExportTargets { get; set; }
```
dfederm: This should probably have comments since it's public?

src/Build/Instance/ProjectInstance.cs
```diff
@@ -709,6 +714,12 @@ public List<ProjectItemElement> EvaluatedItemElements
             private set;
         }
 
+        public IReadOnlyCollection<string> ExportTargets
```
dfederm: Comments here too

src/Build/BackEnd/Shared/BuildResult.cs
```diff
@@ -70,6 +71,8 @@ public class BuildResult : INodePacket, IBuildResults
         /// </summary>
         private List<string> _defaultTargets;
 
+        private IReadOnlyCollection<string> _exportTargets;
```
dfederm: Ideally initial and default targets would also become read only collections. Not necessarily in scope for this PR though.

src/Build/BackEnd/BuildManager/BuildRequestData.cs
dfederm: We'll eventually need it for enforcement. Would this be where we'd read it from?

dfederm: I don't see where this is ever set. Is that intended to be in this PR, or a future PR which wires up the command-line flag to this?

cdmihai: I guess there's two ways of toggling these for builds:
- globally via build parameters. This means that the entire build (all builds submitted in one begin / end build session) should go into `isolated build project` mode, or `execute export targets` mode
- locally for each project build request. This means that we can control these things per each build.

Would we ever want to mix these things? If not, then maybe build parameters would be better. In which case it should be internal so we don't clutter the public APIs.

src/Build/Definition/Project.cs
cdmihai: It's in the Data nested class which is not visible.

src/Build/BackEnd/Shared/BuildResult.cs
cdmihai: ProjectInstance exposes the target lists, so maybe the intention there was to be able to mutate them. But I don't see any good coming from mutating ExportTargets.

src/Build/BackEnd/BuildManager/BuildRequestData.cs
cdmihai: It's set in a test :)
Future PRs will pipe the end to end scenario.

dfederm: We'll never need to mix and match. The build is either an isolated build or not.

dfederm: (So BuildParameters is probably the right thing to use here if possible)

cdmihai:
@dfederm Switched to build parameters.

src/Build/BackEnd/Shared/BuildResult.cs
dfederm: Good point, it's part of the public interface =/

PR #3721, opened 2018-09-08
Project Graph creation to support containing a single project

yash256:


src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -0,0 +1,39 @@
...
+                    <Message Text='Building test project'/>
+                  </Target>
+                </Project>
+                ";
+            string testDirectory = Path.GetTempPath();
+            string projectPath = Path.Combine(testDirectory, "ProjectGraphTest.proj");
+            File.WriteAllText(projectPath, projectContents);
+            var projectGraph = new ProjectGraph(projectPath);
+            
+            Assert.True(projectGraph.ProjectNodes.Count == 1);
```
dfederm: You should use `Assert.Equals` instead for better error messaging. When it fails you'll see something like "Expected 1 but got 2" instead of "Expected true to be false"

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -23,6 +23,6 @@ internal sealed class ProjectGraphNode
         /// <summary>
         /// Gets the evaluated project represented by this node in the graph.
         /// </summary>
-        public Project Project => new Project();
+        public Project Project { get; protected set; }
```
dfederm: Should be a private setter (may need to pass the `Project` in the ctor)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -12,12 +13,19 @@ namespace Microsoft.Build.Graph
...
         /// <summary>
         /// Constructs a graph starting from the given project file.
         /// </summary>
         /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
-        public ProjectGraph(string entryProjectFile) { }
+        public ProjectGraph(string entryProjectFile)
+        {
+            var graphNode =
+                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(entryProjectFile));
```
dfederm: Eventually we'll need to probably take a project collection (optionally), but for now I think this is just fine to get things moving. We can worry about all the "right" overloads to have after the basic functionality takes shape.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -0,0 +1,39 @@
...
+                  <Target Name='Build'>
+                    <Message Text='Building test project'/>
+                  </Target>
+                </Project>
+                ";
+            string testDirectory = Path.GetTempPath();
+            string projectPath = Path.Combine(testDirectory, "ProjectGraphTest.proj");
+            File.WriteAllText(projectPath, projectContents);
+            var projectGraph = new ProjectGraph(projectPath);
+
```
rainersigwald: Nit: whitespace

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -0,0 +1,39 @@
...
+                  </PropertyGroup>
+
+                  <Target Name='Build'>
+                    <Message Text='Building test project'/>
+                  </Target>
+                </Project>
+                ";
+            string testDirectory = Path.GetTempPath();
+            string projectPath = Path.Combine(testDirectory, "ProjectGraphTest.proj");
+            File.WriteAllText(projectPath, projectContents);
```
rainersigwald: We have some nice methods to ensure cleanup of this type of test state. It's [`TestEnvironment`](https://github.com/Microsoft/msbuild/blob/master/src/Shared/UnitTests/TestEnvironment.cs).  Usage will probably be something like

```c#
using (var env = TestEnvironment.Create())
{
    TransientTestProjectWithFiles testProject = env.CreateTestProjectWithFiles(projectContents , Array.Empty<string>());
    var projectGraph = new ProjectGraph(testProject.ProjectFile);
}

```

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -0,0 +1,39 @@
...
+            var projectGraph = new ProjectGraph(projectPath);
+            
+            Assert.Equal(1, projectGraph.ProjectNodes.Count);
+            Project projectNode = projectGraph.ProjectNodes.First().Project;
+            Assert.Equal(projectPath, projectNode.FullPath);
```
rainersigwald: We're moving toward adopting [Shouldly](https://shouldly.readthedocs.io/en/latest/) for test assertions. Add

```c#
using Shouldly;
```

and change these to be like

```c#
projectGraph.ProjectNodes.Count.ShouldBe(1);
projectNode.FullPath.ShouldBe(testProject.ProjectFile);
```

src/Build/Graph/ProjectGraph.cs
```diff
@@ -29,6 +37,6 @@ internal sealed class ProjectGraph
         /// <summary>
         /// Get an unordered collection of all project nodes in the graph.
         /// </summary>
-        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => Array.Empty<ProjectGraphNode>();
+        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => _projectNodes;
```
rainersigwald: I'm starting to wonder if we need this API. It feels a bit weird to expose "here's a list of nodes; don't ever `foreach` over it" instead of graph traversal operations and maybe things like `NodesWithNoProjectReferences`. But maybe those really do belong in the caller? No need to change for this PR but let's keep thinking about this.

dfederm: I do think it's quite common still to get a collection of all unique projects. `foreach` is fine as long as it's not `foreach` -> build

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.UnitTests;
+using Xunit;
+using Shouldly;
+
+namespace Microsoft.Build.Graph.UnitTests
+{
+    public class ProjectgraphTests
```
rainersigwald: Can you change this back to PascalCase with capital G?

src/Build/Graph/ProjectGraph.cs
cdmihai: Another potential overload is one accepting an EvaluationContext. If none is given, we should probably instantiate and use our own.

Speaking of that, should we load the project via the project collection or via the `Project.FromFile` static factory? The latter provides more configuration via [ProjectOptions](https://github.com/Microsoft/msbuild/blob/master/src/Build/Definition/ProjectOptions.cs).

But for now, this way is probably good.

src/Build/Graph/ProjectGraph.cs
cdmihai: I guess people will use the nodes for whatever reason, including hand made builds. Unless we add explicit checks, it will happen :)

Issue #2282, opened 2017-07-11
CreateCSharpManifestResourceName.cs vs CreateVisualBasicManifestResourceName.cs


tmat:
Diffing src\Tasks\CreateCSharpManifestResourceName.cs vs src\Tasks\CreateVisualBasicManifestResourceName.cs shows suspicious differences between the languages. It seems that some bug fixes made to C# has not been applied to VB. 

Ideally, these tasks would share code instead of having separate implementations that diverge over time.

tmat:
cc @nguerrera

AndyGerlicher:
Team Triage:
This sounds great. Marking up for grabs.

isiahto:
Hello,
I have gone through both C# and VB ManifestResourceName class, it seems like they both contain an `internal static string CreateManifestNameImpl` method to generate the same output (according to the unit tests)

Is this issue aimed to extract the impl from the 2 classes?
I would like to work on this issue, please advise!

rainersigwald:
@isiahto Yes, that's the idea. Based on tmat's comments, it sounds like you should start by extracting the C# implementation and then adopt that in the VB one.

We'd love to take a PR that addressed this. Please target the `master` branch, and let us know here if you have any questions. I'll remove the `up-for-grabs` label from this if you want to work on it.

isiahto:
Yes please, thank you.

rainersigwald:
Internally no clear idea whether it diverges intentionally or not; commit message on fix for J# and C# is:

```
   Beta1 bug
    
       CreateJSharpManifestResourceName.java - [309027]
       CreateCSharpManifestResourceName.cs - [309027]
       CreateCSharpManifestResourceName_Tests.cs - [309207]
```

And doesn't say why/why not VB too.

BenVillalobos:
Unassigning myself and adding some context: I recall that VB files had to be considered "special" because of the fact that we manually parse c# files to find the namespace. Talking with Kathleen a while back, I remember it having to do with the differences in namespaces. See `GetFirstClassNameFullyQualified`.

Things may change down the road, but I don't think we can merge these today.

YuliiaKovalova:
Close it as not planned.

PR #3725, opened 2018-09-09
WIP: Extract `CreateManifestNameImpl` logic to refactor duplicate code

isiahto:
This is a WIP PR to address #2282.

#### Work done:
I have extracted the logic into a new static class, `ManifestNameCreator.cs` under the `Microsoft.Build.Tasks` namespace. Built and ran the test, works as intended in C#, but not quite in the VB one.

I have some questions and wonder if I am on the right track:
1. __Should the helper class be placed under `Microsoft.Build.Tasks` namespace?__ I tried placing the new static helper class under `ManifestUtils` (actual: `Deployment.ManifestUtilities`), however I notice the classes within there only build for `net46` but not for `netstandard2.0`, that's why I moved it out to _Build.Tasks_.
1. __Should the new parameter `Func<Stream, ExtractedClassName> getFirstClassNameFullyQualified` be added as an delegate?__ I saw the `CreateFileStream` parameter in `CreateManifestResourceName::Execute()` and found out there's a list of delegates in `delegate.cs`. Is `delegate` preferred over `Func<>`?
1. For the original C# impl, the logic captures "sub folder" as part of the output, but not for the VB impl, and I got the following test result:
```
Test Name:  Microsoft.Build.UnitTests.CreateVisualBasicManifestResourceName_Tests.CulturedBitmapWithRootNamespaceNoDirectoryPrefix
Test FullName:  Microsoft.Build.UnitTests.CreateVisualBasicManifestResourceName_Tests.CulturedBitmapWithRootNamespaceNoDirectoryPrefix
Test Source:  D:\Proj\msbuild\src\Tasks.UnitTests\CreateVisualBasicManifestResourceName_Tests.cs : line 310
Test Outcome: Failed
Test Duration:  0:00:00.001

Result StackTrace:  at Microsoft.Build.UnitTests.CreateVisualBasicManifestResourceName_Tests.CulturedBitmapWithRootNamespaceNoDirectoryPrefix() in D:\Proj\msbuild\src\Tasks.UnitTests\CreateVisualBasicManifestResourceName_Tests.cs:line 323
Result Message: 
Assert.Equal() Failure
                         ↓ (pos 15)
Expected: RootNamespace.SplashScreen.bmp
Actual:   RootNamespace.SubFolder.SplashScreen.bmp
                         ↑ (pos 15)
```
__What should I do here?__ Rainer mentioned that I should extract the C# impl and adopt to the VB call site.

I guess some of the questions I have is due to lack of understanding of how the component works, could someone kindly explain a bit of how this specific build task is used? And any example would be greatly appreciated!

Thanks


isiahto:
Hi @AndyGerlicher, @rainersigwald,

I have extracted the function into a new static class, i spotted 2 differences between the C# and VB name creating functions,
- C# Manifest Resource Name: __Does not include__ Root Namespace, but __includes__ subfolder
- VB Manifest Resource Name: __Includes__ Root Namespace, but __does not include__ subfolder

So I extracted the function with 2 extra params (bool flags for in/excluding those two values). Also created two wrappers to maintain current param count and hide the flag settings from callee as its part of the name creation logic.

The tests for those 2 classes are tightly related with the original static functions within them, I am not sure if I should create another set of tests for this new helper class(?)

Thanks!

PR #3732, opened 2018-09-11
filter references before manifest generation

sidiesen:
With the recent change of using ReferenceCopyLocalPaths for manifest creation there are a few edge cases that broke:
- if there is a native dll referenced as a Content file in a nuget package it was incorrectly assumed it was a managed reference, which could cause build breaks.
- the ClickOnce application files window was not always properly populated.
- files that were set as "Include" in the application files window but had Copy Local set to false were not correctly published
- .NET Core reference libraries were incorrectly published with ClickOnce

This change also introduces an item group "_ExcludedAssembliesFromManifestGeneration" that allows users to manually specify files that should explicitly be excluded from the list of dependencies for the manifest generation, in case there are further cases emerging of files in nuget packages getting incorrectly flagged as a reference. This allows the user to unblock themselves by explicitly excluding the faulty files.


Jinhuafei:
LGTM

PR #3737, opened 2018-09-13
Update bindingRedirect versions of Workflow build tasks dlls

XinxingLiu:
To adapt assemblies version changes in Dev16. These two assemblies are dropped to msbuild folder when installing VS, so they have the same version with VS.


XinxingLiu:
To adapt assemblies version changes in Dev16. These two assemblies are dropped to msbuild folder when installing VS, so they have the same version with VS.

XinxingLiu:
@AndyGerlicher Could you help to check in this change? It seems I could not do it as I am not the owner of this repo. Thanks!

PR #3739, opened 2018-09-13
Another keyword to check with rebuild issues

Dandielo:
This may happen when creating a custom build rule and improperly writing a 'TLog Read' file. (ex. writing relative paths to it)


rainersigwald:
Thanks! Can you give an example of how to produce this error? I don't see an obvious place for it in our codebase and it's not quite clear to me from the description. I'd like to see if there's a longer, more unique phrase we could mention here, since the word `missing` appears in several unrelated warnings and might produce false positives.

Dandielo:
Well, i did had the problem when working with a custom .targets file trying do add support for reading and writing dependencies from the source file to a {project}.read.1.tlog file. The issue was i've been writing the path prefixed with the '^' character as a relative path ex. ../../some/directory/input_file.sth which would be always missing, the relative path was valid from the vcxproj prespective

PR #3744, opened 2018-09-14
Revert "Merge pull request #3716 from cdmihai/exportTargets"

cdmihai:
Export targets got replaced by an explicit project reference protocol: #3743


cdmihai:
@dfederm :)

PR #3747, opened 2018-09-14
Load project references recursively into project

yash256:


yash256:
Closes https://github.com/Microsoft/msbuild/issues/3683

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +28,69 @@ internal sealed class ProjectGraph
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
         public ProjectGraph(string entryProjectFile)
         {
-            var graphNode =
-                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(entryProjectFile));
-            _projectNodes.Add(graphNode);
+            AddNewNode(entryProjectFile);
```
dfederm: This doesn't recurse. Seems like it should call the other ctor overload, or the `LoadGraph` directly.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +28,69 @@ internal sealed class ProjectGraph
...
+            var graphNode =
+                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(projectFilePath));
+            _allParsedProjects.Add(projectFilePath, graphNode);
+            return graphNode;
+        }
+
+        private List<ProjectGraphNode> LoadGraph(Queue<string> projectsToParse)
+        {
+            var parsedProjects = new List<ProjectGraphNode>();
+            while (projectsToParse.Count != 0)
```
dfederm: Nit: I think there's an `IsEmpty` property on Queues?

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -21,11 +23,16 @@ internal ProjectGraphNode(Project project)
...
         /// </summary>
-        public IReadOnlyCollection<ProjectGraphNode> ProjectReferences => Array.Empty<ProjectGraphNode>();
+        public IReadOnlyCollection<ProjectGraphNode> ProjectReferences => _projectReferences;
 
         /// <summary>
         /// Gets the evaluated project represented by this node in the graph.
         /// </summary>
         public Project Project { get; }
+
+        internal void AddProjectReferences(List<ProjectGraphNode> projectGraphNodes)
```
dfederm: Instead of an add call, can this be part of the ctor? It's always better to have it fully constructed when possible. Or I could also see an `AddProjectReference` if it were added immediately when discovered.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +28,69 @@ internal sealed class ProjectGraph
...
+        {
+            var graphNode =
+                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(projectFilePath));
+            _allParsedProjects.Add(projectFilePath, graphNode);
+            return graphNode;
+        }
+
+        private List<ProjectGraphNode> LoadGraph(Queue<string> projectsToParse)
+        {
+            var parsedProjects = new List<ProjectGraphNode>();
```
dfederm: This seems to end up with the transitive closure, not just the direct project references.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +28,69 @@ internal sealed class ProjectGraph
...
+
+        private ProjectGraphNode AddNewNode(string projectFilePath)
+        {
+            var graphNode =
+                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(projectFilePath));
+            _allParsedProjects.Add(projectFilePath, graphNode);
+            return graphNode;
+        }
+
+        private List<ProjectGraphNode> LoadGraph(Queue<string> projectsToParse)
```
dfederm: I'm not sure I understand the pattern of both using a queue and recursing. I think that leads to a whole bunch of duplication.

I think either the approach should be:
a) use a queue, no recursion. This may be harder to associate project references though, but is simpler to walk
b) recursion, no queue. This is easier to associate project references, but you end up with potentially really deep recursion.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +28,69 @@ internal sealed class ProjectGraph
...
+                string projectToParse = projectsToParse.Dequeue();
+                if (_allParsedProjects.TryGetValue(projectToParse, out ProjectGraphNode parsedProject))
+                {
+                    parsedProjects.Add(parsedProject);
+                }
+                else
+                {
+                    parsedProject = AddNewNode(projectToParse);
+                    parsedProjects.Add(parsedProject);
+                    ProjectInstance projectInstance = parsedProject.Project.CreateProjectInstance();
```
dfederm: Can't you query items from a `Project` as well?

src/Build/Graph/ProjectGraph.cs
rainersigwald: I lean toward using the queue-only approach. It's closer to what we'll want when walking through, discovering the graph in parallel.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +28,69 @@ internal sealed class ProjectGraph
...
         /// </summary>
         /// <param name="entryProjectFiles">The project files to use as the entry points in constructing the graph</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
-        public ProjectGraph(IEnumerable<string> entryProjectFiles) { }
+        public ProjectGraph(IEnumerable<string> entryProjectFiles)
+        {
+            var projectsToParse = new Queue<string>();
+            foreach (var entryProjectFile in entryProjectFiles)
+            {
+                projectsToParse.Enqueue(entryProjectFile);
```
rainersigwald: There's a `Queue(IEnumerable)` constructor so you can just do

```c#
var projectsToParse = new Queue<string>(entryProjectFiles);
```

src/Build/Graph/ProjectGraphNode.cs
rainersigwald: I agree; I'd go so far as to move the reach-into-the-project-and-extract references code into this class and call it from the constructor.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -37,6 +38,256 @@ public void TestGraphWithSingleNode()
...
+                    env.CreateTestProjectWithFiles(projectBContents, Array.Empty<string>(), "..\\B");
+                TransientTestProjectWithFiles projectC =
+                    env.CreateTestProjectWithFiles(projectCContents, Array.Empty<string>(), "..\\C");
+                List<string> projectsToParse = new List<string>();
+                projectsToParse.Add(projectA.ProjectFile);
+                projectsToParse.Add(projectB.ProjectFile);
+                ProjectGraph graph = new ProjectGraph(projectsToParse);
+
+                graph.ProjectNodes.Count.ShouldBe(3);
+                ProjectGraphNode projectNodeA = graph.ProjectNodes.First(node => node.Project.FullPath.Equals(projectA.ProjectFile));
```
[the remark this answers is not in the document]
dfederm: We'll need to think about exactly what the final public graph API is at some point. Should nodes have a "refereng**ing**" field (if we could ever come up with a good name there)? Or a lookup by `Project`/`ProjectInstance`/project file name?

This may become clear as we try to start actually using it and also when integrating this in with QuickBuild.

src/Build/Graph/ProjectGraph.cs
yash256: Count seems to be the only option

src/Build/Graph/ProjectGraphNode.cs
yash256: I have added an AddProjectReference call to update the parent node as soon as a child node is evaluated

cdmihai: I wouldn't explore references in the constructor to make it easier to stub out the graph.

src/Build/Graph/ProjectGraph.cs
cdmihai: +1 for queue only :)

src/Build/Graph/ProjectGraph.cs
cdmihai: What's the benefit of having `parsedProjects` and `_allParsedProjects`? Can't we use just `_allParsedProjects`?

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -21,11 +23,16 @@ internal ProjectGraphNode(Project project)
...
+        public IReadOnlyCollection<ProjectGraphNode> ProjectReferences => _projectReferences;
 
         /// <summary>
         /// Gets the evaluated project represented by this node in the graph.
         /// </summary>
         public Project Project { get; }
+
+        internal void AddProjectReference(ProjectGraphNode projectGraphNode)
+        {
+            _projectReferences.Add(projectGraphNode);
```
dfederm: Nit: expression body since this method is so siple:
```
internal void AddProjectReference(ProjectGraphNode projectGraphNode) => _projectReferences.Add(projectGraphNode);
```

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -37,6 +46,220 @@ public void TestGraphWithSingleNode()
...
+        /// / \ 
+        ///B   C
+        /// </summary>
+        [Fact]
+        public void TestGraphWithThreeNodes()
+        {
+            string projectAContents = @"
+                <Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003'>
+                  <PropertyGroup>
+                    <TestProperty>value</TestProperty>
```
dfederm: These `TestProperty` properties don't seem necessary in any of these projects. It's OK for some of these projects to literally just be `<Project></Project>`

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -37,6 +46,220 @@ public void TestGraphWithSingleNode()
...
+        /// <summary>
+        ///  A  
+        /// / \ 
+        ///B   C
+        /// </summary>
+        [Fact]
+        public void TestGraphWithThreeNodes()
+        {
+            string projectAContents = @"
+                <Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003'>
```
dfederm: The `xmlns` isn't really necessary.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +26,56 @@ internal sealed class ProjectGraph
...
         /// </summary>
-        /// <param name="entryProjectFiles">The project files to use as the entry points in constructing the graph</param>
-        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
-        public ProjectGraph(IEnumerable<string> entryProjectFiles) { }
+        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => _allParsedProjects.Values;
+
+        private ProjectGraphNode CreateNewNode(string projectFilePath)
+        {
+            var graphNode =
+                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(projectFilePath));
```
cdmihai: Should we use Project or ProjectInstance? Command line msbuild uses ProjectInstance. Only VS uses Project, which is a bit more expensive, and in certain cases might actually lead to different results (not super sure that's the case though)

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -13,6 +14,14 @@ namespace Microsoft.Build.Graph.UnitTests
 {
     public class ProjectGraphTests
     {
+        private const string ProjA = "A.proj";
```
dfederm: Hmm, these are hardcoded in the project content, but used as consts here.

Here's what I was planning for the next change I'm working on: https://github.com/dfederm/msbuild/blob/dfederm/graph-target-lists/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs#L96

Basically just work with project numbers and have the helper methods add the "proj" suffixes. Also, I used numbers for projects and letters for targets.

I'm fine with changing to your style though if you find a way to reduce the duplication.

Or we can just leave this and do a UT refactor after the fact :)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +26,56 @@ internal sealed class ProjectGraph
...
+            // entry project node has no parent
+            projectsToEvaluate.Enqueue(new KeyValuePair<string, ProjectGraphNode>(entryProjectFile, null));
+            while (projectsToEvaluate.Count != 0)
+            {
+                KeyValuePair<string, ProjectGraphNode> projectToEvaluateAndParentNode = projectsToEvaluate.Dequeue();
+                string projectToEvaluate = projectToEvaluateAndParentNode.Key;
+                if (!_allParsedProjects.TryGetValue(projectToEvaluate, out ProjectGraphNode parsedProject))
+                {
+                    parsedProject = CreateNewNode(projectToEvaluate);
+                    IEnumerable<ProjectItem> projectReferenceItems =
```
cdmihai: I thought a node is a tuple of project path and global properties. Is the plan to add global properties later on?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +26,56 @@ internal sealed class ProjectGraph
...
-        /// Get an unordered collection of all project nodes in the graph.
+        /// Load a graph with root node at entryProjectFile
+        /// Maintain a queue of projects to be processed- each queue item is a key value pair of the project to be evaluated and its parent
+        /// Once the project has been evaluated, add a project reference to this evaluated target from the parent node
         /// </summary>
-        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => _projectNodes;
+        /// <param name="entryProjectFile"></param>
+        private void LoadGraph(string entryProjectFile)
+        {
+            var projectsToEvaluate = new Queue<KeyValuePair<string, ProjectGraphNode>>();
```
rainersigwald: This might read a bit more naturally if these were tuples rather than `KeyValuePair` but I don't feel strongly about it.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -22,21 +26,56 @@ internal sealed class ProjectGraph
...
+                string projectToEvaluate = projectToEvaluateAndParentNode.Key;
+                if (!_allParsedProjects.TryGetValue(projectToEvaluate, out ProjectGraphNode parsedProject))
+                {
+                    parsedProject = CreateNewNode(projectToEvaluate);
+                    IEnumerable<ProjectItem> projectReferenceItems =
+                        parsedProject.Project.GetItems(ProjectReferenceString);
+                    foreach (var projectReferenceToParse in projectReferenceItems)
+                    {
+                        string projectReferencePath = projectReferenceToParse.GetMetadataValue(FullPathString);
+                        projectsToEvaluate.Enqueue(new KeyValuePair<string, ProjectGraphNode>(projectReferencePath, parsedProject));
```
rainersigwald: We just talked about this in person, but . . .

Another option here is to only enqueue new projects, moving the if-already-known check to here, and if the referenced project is already known, add the reference directly. If it's not, enqueue it.

But I don't think we must change to that.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -37,6 +46,220 @@ public void TestGraphWithSingleNode()
...
+                ProjectGraphNode projectNodeB = graph.ProjectNodes.First(node => node.Project.FullPath.Contains(ProjB));
+                ProjectGraphNode projectNodeC = graph.ProjectNodes.First(node => node.Project.FullPath.Contains(ProjC));
+                projectNodeA.ProjectReferences.Count.ShouldBe(2);
+                projectNodeB.ProjectReferences.Count.ShouldBe(0);
+                projectNodeC.ProjectReferences.Count.ShouldBe(0);
+            }
+        }
+
+        /// <summary>
+        /// Test the following graph with entry project B
```
rainersigwald: In case it helps anyone else:

image

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -37,6 +46,220 @@ public void TestGraphWithSingleNode()
...
+                  </ItemGroup>
+                </Project>
+                ";
+            using (var env = TestEnvironment.Create())
+            {
+                TransientTestProjectWithFiles projectA =
+                    env.CreateTestProjectWithFiles(ProjA, projectAContents, new []{ProjB, ProjC});
+                File.WriteAllText(Path.Combine(projectA.TestRoot, ProjB), projectBContents);
+                File.WriteAllText(Path.Combine(projectA.TestRoot, ProjC), projectCContents);
+                ProjectGraph graph = new ProjectGraph(projectA.ProjectFile);
```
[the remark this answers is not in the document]
yash256: The purpose of this test is to ensure that we don't are not stuck in an infinite loop of resolving project references. https://github.com/Microsoft/msbuild/issues/3757 will address throwing errors when there is a cycle.

src/Build/Graph/ProjectGraph.cs
yash256: parsedProjects was a list that was used in a previous iteration. _allParsedProjects is the only thing in use now

src/Build/Graph/ProjectGraph.cs
```diff
@@ -11,32 +11,108 @@ namespace Microsoft.Build.Graph
...
-        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => _projectNodes;
+        public IReadOnlyCollection<ProjectGraphNode> ProjectNodes { get; }
+
+        private ProjectGraphNode CreateNewNode(string projectFilePath)
+        {
+            var graphNode =
+                new ProjectGraphNode(ProjectCollection.GlobalProjectCollection.LoadProject(projectFilePath));
+            _allParsedProjects.Add(projectFilePath, graphNode);
+            return graphNode;
+        }
```
yash256: this will be removed in the next iteration

PR #3748, opened 2018-09-15
Build a graph containing a single node

### LINKED ISSUE #3682: Build a graph containing a single node
Given a ProjectGraph with a single node, build it

dfederm:
Closes #3682

### DIFF
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
@@ -11,9 +10,11 @@ namespace Microsoft.Build.Graph
     /// <summary>
     /// Represents a graph of evaluated projects.
     /// </summary>
-    internal sealed class ProjectGraph
+    public sealed class ProjectGraph
     {
         private readonly List<ProjectGraphNode> _projectNodes = new List<ProjectGraphNode>();
+
+        // TODO: We probably want to mirror all relevant constructor overloads from Project
 
         /// <summary>
         /// Constructs a graph starting from the given project file.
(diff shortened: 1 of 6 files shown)


dfederm:
I need to add a unit test in `XMake_Tests` which just makes sure a single project can build when `/isolate` is passed in.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -11,28 +10,45 @@ namespace Microsoft.Build.Graph
...
-        /// Constructs a graph starting from the given project files.
+        /// Constructs a graph starting from the given project file.
         /// </summary>
-        /// <param name="entryProjectFiles">The project files to use as the entry points in constructing the graph</param>
+        /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
+        /// <param name="globalProperties">The global properties to use for all projects. May be null, in which case the containing project collection's global properties will be used.</param>
+        /// <param name="projectCollection">The collection with which all projects in the graph should be associated. May not be null.</param>
+        /// <param name="toolsVersion">The tools version. May be null.</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
-        public ProjectGraph(IEnumerable<string> entryProjectFiles) { }
```
dfederm: I removed this multi-project ctor as it was more of a theoretical scenario not based in a specific concrete use case, so I'm removing it for now to simplify implementation.

Anyway a metaproj type thing could solve this multi-project entry point anyway if it's really needed.

src/MSBuild/XMake.cs
```diff
@@ -1240,6 +1248,35 @@ private static BuildResult ExecuteBuild(BuildManager buildManager, BuildRequestD
...
+        {
+            lock (s_buildLock)
+            {
+                var projectGraph = new ProjectGraph(projectFile, projectCollection, globalProperties, toolsVersion);
+
+                // TODO: Do a full graph traversal
+                var projectInstance = projectGraph.EntryProjectNode.Project.CreateProjectInstance();
+                var request = new BuildRequestData(projectInstance, targets);
+
+                s_activeBuild = buildManager.PendBuildRequest(request);
```
dfederm: This is extremely rudimentary for now. I think using the build manager as a black box is an OK initial approach, but I think a better eventual solution is to just spam the build manager with requests for each project providing the correct "blocker" data to describe the dependency graph correctly. The build manager already has a JIT dependency graph effectively for who is blocked on who, but unfortunately as it stands, it's "which node is blocked on which request", while we need more of "which request is blocked on which request". Anyway, something to explore more when we get there.

src/MSBuild/XMake.cs
```diff
@@ -1142,7 +1143,14 @@ bool isolateProjects
 
                             if (!restoreOnly)
                             {
-                                results = ExecuteBuild(buildManager, request);
+                                if (isolateProjects)
```
dfederm: Hmm, this is a little strange. I still think we'll eventually need 2 separate flags. One for isolated builds and one for graph builds. Isolated means without graph means 1 single project and import the build result state from its references somehow, and a graph build could either be isolated or not (basically use graph scheduling with or without the isolation enforcement). For now I think using this one flag for everything is OK until we start needing to disambiguate the scenarios (we'll start with just graph + isolated both on)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -11,28 +10,45 @@ namespace Microsoft.Build.Graph
...
         }
 
         /// <summary>
-        /// Constructs a graph starting from the given project files.
+        /// Constructs a graph starting from the given project file.
         /// </summary>
-        /// <param name="entryProjectFiles">The project files to use as the entry points in constructing the graph</param>
+        /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
+        /// <param name="globalProperties">The global properties to use for all projects. May be null, in which case the containing project collection's global properties will be used.</param>
+        /// <param name="projectCollection">The collection with which all projects in the graph should be associated. May not be null.</param>
```
rainersigwald: nit: order here doesn't match order of args

src/MSBuild/XMake.cs
rainersigwald: I agree this is fine for now. I'm not sure what the right thing will be in the end.

cdmihai: Options David and I discussed:
- for quick prototyping, use the BuildManager as a black box. The graph builder instantiates a build manager, walks the projects bottom up and keeps pending builds on the BuildManager.
- scheduler based
  - the Scheduler already constructs a JIT dependency graph. If we can adapt the static graph to Scheduler graph, it might just work as expected. I like this option because it reuses the Scheduler mostly as is.
  - the Scheduler implements IScheduler. We can add another implementation. This is nice if we think we can do better than the Scheduler. It's also nice that we'd hopefully make something more readable / understandable. Downside is that we'll need understand more details about current scheduling.

dfederm: I'll open a work item to investigate proper scheduling.

cdmihai:
@yash256 
Should #3747 wait until this goes in? Looks like this PR causes #3747 to implement the new members on the graph.

PR #3749, opened 2018-09-15
 Use bool to check for dummy CopyOnWriteDictionary

ccastanedaucf:
Noticed a decent chunk of time in ```CopyOnWriteDictionary``` is spent just comparing references with the static dummy. Probably negligible perf gain but super easy fix. Right side is with changes. Ran this on top of the ```HybridDictionary``` removal since that was originally dominating all the runtime in here.

dictionarydummy


ccastanedaucf:
On building WebLargeCore I see around 750k ```CopyOnWriteDictionary``` instances created, and ```isDummy``` is checked around 8 million times since every operation on the dictionary checks the value.

PR #3763, opened 2018-09-18
Bump VisualStudioVersion to 16.0

rainersigwald:
This will allow MSBuild to find Visual Studio 16 instances and is part
of the larger versioning approach for MSBuild 16.

### DIFF
--- a/src/Shared/Constants.cs
+++ b/src/Shared/Constants.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Build.Shared
         /// <summary>
         /// The most current ToolsVersion known to this version of MSBuild.
         /// </summary>
-        internal const string CurrentToolsVersion = CurrentVisualStudioVersion;
+        internal const string CurrentToolsVersion = "15.0";
 
         // if you change the key also change the following clones
         // Microsoft.Build.OpportunisticIntern.BucketedPrioritizedStringList.TryIntern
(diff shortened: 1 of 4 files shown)


src/Shared/Constants.cs
```diff
@@ -58,12 +58,12 @@ internal static class MSBuildConstants
...
         /// The most current Visual Studio Version known to this version of MSBuild.
         /// </summary>
-        internal const string CurrentVisualStudioVersion = "15.0";
+        internal const string CurrentVisualStudioVersion = "16.0";
 
         /// <summary>
         /// The most current ToolsVersion known to this version of MSBuild.
         /// </summary>
-        internal const string CurrentToolsVersion = CurrentVisualStudioVersion;
+        internal const string CurrentToolsVersion = "15.0";
```
rainersigwald: This was the line I had missed that caused it to fail a bajillion tests. The fix was . . . not hard.

cdmihai: Why is it 15 and not 16?

cdmihai: Oh, toolsversion, nvm :)

src/Shared/Constants.cs
```diff
@@ -82,13 +82,7 @@ internal static class MSBuildConstants
...
         /// Current version of this MSBuild Engine assembly in the form, e.g, "12.0"
         /// </summary>
-        internal static string CurrentProductVersion
-        {
-            get
-            {
-                return "15.0";
-            }
-        }
+        internal static string CurrentProductVersion => "16.0";
```
cdmihai: Would it be better if this got read from the dll informational version major part?

rainersigwald:
@dotnet-bot test RHEL7.2 Build for CoreCLR please

rainersigwald:
Superseded by #3779.

Issue #3752, opened 2018-09-15
Traverse project graph to construct target list for each project


dfederm:
Traverse the graph reading the `ProjectReferenceTargets` items for each project to determine the target list to use for each project.

Simplified rules:
* User-provided targets is used for the entry project (or if not provided, evaluate the default targets for that project)
* For each entry target in a project, check for a `ProjectReferenceTargets` item with that target name. If one exists, the Targets metadata should be added to the target list of all project references
* If a project is referenced by multiple projects, the target lists are concatenated in an arbitrary, but deterministic way. Target lists maintain order within the list as well.
* For entry targets without a mapping, assume they don't propagate to project references in any way
* Expand the special `<default>` target as the default targets for a project (likely need to decide on a better syntax) for prior to looking up `ProjectReferenceTargets` items

See spec for details.

Issue #3757, opened 2018-09-15
Detect and error on project graph cycles


dfederm:
We should detect cycles on graph construction and error appropriately

rainersigwald:
I’d really like this error to be better than the existing cycle errors which often just say “there is a cycle involving X”. I’d like to see every project in the full cycle in the error, ideally with full paths and maybe line numbers.

PR #3769, opened 2018-09-18
Traverse project graph to construct target list for each project

dfederm:
Closes: #3752

### DIFF
--- a/ref/Microsoft.Build/net/Microsoft.Build.cs
+++ b/ref/Microsoft.Build/net/Microsoft.Build.cs
@@ -1413,11 +1413,12 @@ namespace Microsoft.Build.Graph
         public ProjectGraph(string entryProjectFile, Microsoft.Build.Evaluation.ProjectCollection projectCollection, System.Collections.Generic.Dictionary<string, string> globalProperties, string toolsVersion) { }
         public Microsoft.Build.Graph.ProjectGraphNode EntryProjectNode { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> ProjectNodes { get { throw null; } }
+        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, System.Collections.Immutable.ImmutableList<string>> GetTargetLists(string[] entryProjectTargets) { throw null; }
     }
     public sealed partial class ProjectGraphNode
     {
         internal ProjectGraphNode() { }
-        public Microsoft.Build.Evaluation.Project Project { get { throw null; } }
+        public Microsoft.Build.Execution.ProjectInstance Project { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> ProjectReferences { get { throw null; } }
     }
 }
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -3,8 +3,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
+using Microsoft.Build.Execution;
 
 namespace Microsoft.Build.Graph
 {
@@ -14,7 +17,10 @@ namespace Microsoft.Build.Graph
     public sealed class ProjectGraph
     {
         private const string ProjectReferenceItemName = "ProjectReference";
-        private const string FullPathString = "FullPath";
+        private const string FullPathMetadataName = "FullPath";
+        private const string ProjectReferenceTargetsItemType = "ProjectReferenceTargets";
+        private const string ProjectReferenceTargetsMetadataName = "Targets";
+        private const string DefaultTargetsMarker = ".default";
 
         private readonly Dictionary<string, ProjectGraphNode> _allParsedProjects =
             new Dictionary<string, ProjectGraphNode>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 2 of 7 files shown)


src/Build/Graph/ProjectGraph.cs
```diff
@@ -54,5 +59,139 @@ public ProjectGraph(string entryProjectFile)
...
+                {
+                    continue;
+                }
+
+                // Based on the entry points of this project, determine which targets to propagate down to project references.
+                var targetsToAppendToChildren = new List<string>();
+                ICollection<ProjectItem> projectReferenceTargets = node.Project.GetItems(ProjectReferenceTargetsItemType);
+                foreach (var target in targets)
+                {
+                    // TODO: Normalize targets? Default targets needs expansion. Flip the loop maybe
```
dfederm: Oops, forgot to do this part.

src/MSBuild/XMake.cs
```diff
@@ -1259,10 +1259,12 @@ private static BuildResult ExecuteBuild(BuildManager buildManager, BuildRequestD
             lock (s_buildLock)
             {
                 var projectGraph = new ProjectGraph(projectFile, projectCollection, globalProperties, toolsVersion);
+                var targetLists = projectGraph.GetTargetLists(targets);
 
                 // TODO: Do a full graph traversal
                 var projectInstance = projectGraph.EntryProjectNode.Project.CreateProjectInstance();
-                var request = new BuildRequestData(projectInstance, targets);
+                var targetList = targetLists[projectGraph.EntryProjectNode];
```
dfederm: Yes, I know `targetList == targets`. This ensures that it actually works though, and once we implement the full graph traversal for build, it'll look more like this anyway.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -54,5 +59,139 @@ public ProjectGraph(string entryProjectFile)
...
+            Dictionary<ProjectGraphNode, List<string>> targetLists = _projectNodes.ToDictionary(node => node, node => new List<string>());
+
+            // Initial state of the graph traversal.
+            var entryEdge = new ProjectGraphNodeWithTargets(EntryProjectNode, entryProjectTargets);
+            var visitedEdges = new HashSet<ProjectGraphNodeWithTargets>();
+            visitedEdges.Add(entryEdge);
+            var edgesToVisit = new Queue<ProjectGraphNodeWithTargets>();
+            edgesToVisit.Enqueue(entryEdge);
+
+            // Traverse the entire graph, visiting each node once.
```
dfederm: I guess this is actually "visiting each edge once"

src/MSBuild/XMake.cs
cdmihai: I know this was in another PR, but what if there are multiple entry project nodes?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -12,6 +14,9 @@ namespace Microsoft.Build.Graph
     /// </summary>
     public sealed class ProjectGraph
     {
+        private const string ProjectReferenceTargetsItemType = "ProjectReferenceTargets";
+        private const string ProjectReferenceTargetsMetadataName = "Targets";
```
cdmihai: We've been putting special element names in MSBuildConstants

src/Build/Graph/ProjectGraph.cs
```diff
@@ -54,5 +59,139 @@ public ProjectGraph(string entryProjectFile)
         /// Get an unordered collection of all project nodes in the graph.
         /// </summary>
         public IReadOnlyCollection<ProjectGraphNode> ProjectNodes => _projectNodes;
+
+        /// <summary>
+        /// Gets the target list for every node, given a particular target list for the entry project.
+        /// </summary>
+        /// <param name="entryProjectTargets">The target list for the entry project. May be null or empty, in which case the project's default targets will be used.</param>
+        /// <returns>A dictionary containing the target list for each node.</returns>
+        public IReadOnlyDictionary<ProjectGraphNode, string[]> GetTargetLists(string[] entryProjectTargets)
```
cdmihai: Should the target list be a readonly collection instead of mutable array?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -54,5 +59,139 @@ public ProjectGraph(string entryProjectFile)
...
+        {
+            // Seed the dictionary with empty lists for every node.
+            Dictionary<ProjectGraphNode, List<string>> targetLists = _projectNodes.ToDictionary(node => node, node => new List<string>());
```
cdmihai: Worth using a singleton for empty lists?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -54,5 +59,139 @@ public ProjectGraph(string entryProjectFile)
...
+                        targetList.RemoveAt(i);
+                    }
+                }
+            }
+
+            // Massage the final dictionary into string arrays instead of lists.
+            return targetLists.ToDictionary(pair => pair.Key, pair => pair.Value.ToArray());
+        }
+
+        private struct ProjectGraphNodeWithTargets : IEquatable<ProjectGraphNodeWithTargets>
```
cdmihai: This actually represents an edge, so should it be called something like `IncomingTargetsEdge` instead? It would make it fit in better with the `edge` named local variables.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -54,5 +59,139 @@ public ProjectGraph(string entryProjectFile)
...
+                            targetsToAppendToChildren.AddRange(ExpressionShredder.SplitSemiColonSeparatedList(targetsMetadataValue));
+                        }
+                    }
+                }
+
+                // Queue the project references for visitation, if the edge hasn't already been traversed.
+                foreach (var projectReference in node.ProjectReferences)
+                {
+                    var projectReferenceEdge = new ProjectGraphNodeWithTargets(projectReference, targetsToAppendToChildren);
+                    if (visitedEdges.Add(projectReferenceEdge))
```
cdmihai: Else branch means cycle right? If cycle, this should error.

cdmihai: If cycles would be valid, we'd need to switch to a `while(new changes){PropagateEdgeInfo}` visitation to account for feedback.

---
In reply to: [218885297](https://github.com/Microsoft/msbuild/pull/3769#discussion_r218885297) [](ancestors = 218885297)

src/Build/Graph/ProjectGraph.cs
cdmihai: Nvm, it gets mutated later on :)

---
In reply to: [218880779](https://github.com/Microsoft/msbuild/pull/3769#discussion_r218880779) [](ancestors = 218880779)

src/Build/Graph/ProjectGraph.cs
dfederm: Sure I can do that if that's the pattern, but it's a bit of a weird pattern that a constant used in literally only 1 class would be in some central location. "internal const" I get, but "private const" seems fine to be like this.

src/Build/Graph/ProjectGraph.cs
dfederm: It's an array since the `BuildRequestData` requires an array. Since this is a public API though, perhaps you're right and the conversion to an array should happen when constructing that `BuildRequestData`.

src/Build/Graph/ProjectGraph.cs
dfederm: Cycles are not valid. We will want to detect cycles on graph creation though (see #3757), so at this point it should be safe to assume no cycles.

The `else` here isn't necessarily a cycle though. Edges are defined by the incoming target list and the project, but *not* the referencing project. So if two projects share a project reference, those edges will be deduped by this. So these aren't distinct edges really, just incoming edges from the POV of a given project. It's an optimization so that projects who many many other projects depend on don't need to have a massive target list prior to target list deduplication below.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -14,7 +16,10 @@ namespace Microsoft.Build.Graph
     public sealed class ProjectGraph
     {
         private const string ProjectReferenceItemName = "ProjectReference";
-        private const string FullPathString = "FullPath";
+        private const string FullPathMetadataName = "FullPath";
+        private const string ProjectReferenceTargetsItemType = "ProjectReferenceTargets";
+        private const string ProjectReferenceTargetsMetadataName = "Targets";
+        private const string DefaultTargetsMarker = ".default";
```
dfederm: I decided to go with ".default" as its not a valid target name and doesn't "look like" a property or item access.

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -28,7 +27,7 @@ internal ProjectGraphNode(Project project)
         /// <summary>
         /// Gets the evaluated project represented by this node in the graph.
         /// </summary>
-        public Project Project { get; }
+        public ProjectInstance Project { get; }
```
dfederm: After speaking offline, we decided ProjectInstance would be better, at least for now since DefaultTargets isn't exposed on Project.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -106,5 +199,89 @@ private void LoadGraph(string entryProjectFile, ProjectCollection projectCollect
...
+                        targets = new List<string>(targets);
+                        isMutated = true;
+                    }
+
+                    targets.RemoveAt(i);
+
+                    var defaultTargets = project.DefaultTargets;
+                    for (int j = 0; j < defaultTargets.Count; j++)
+                    {
+                        targets.Insert(i++, defaultTargets[j]);
```
cdmihai: List has InsertRange

src/Build/Graph/ProjectGraph.cs
```diff
@@ -60,14 +65,103 @@ public ProjectGraph(string entryProjectFile)
...
+        /// Gets the target list to be executed for every project in the graph, given a particular target list for the entry project.
+        /// This method uses the ProjectReferenceTargets items to determine the targets to run per node. The results can then be used
+        /// to start building each project individually, assuming a given project is built after its references.
+        /// </summary>
+        /// <param name="entryProjectTargets">The target list for the entry project. May be null or empty, in which case the project's default targets will be used.</param>
+        /// <returns>A dictionary containing the target list for each node.</returns>
+        public IReadOnlyDictionary<ProjectGraphNode, string[]> GetTargetLists(string[] entryProjectTargets)
+        {
+            List<string> entryTargets = entryProjectTargets == null || entryProjectTargets.Length == 0
+                ? HydrateDefaultTargets(EntryProjectNode.Project, new List<string> { DefaultTargetsMarker })
```
cdmihai: Rather than hydrate, you can just add the default targets directly from the project instance.

cdmihai: And clone it.

---
In reply to: [219355209](https://github.com/Microsoft/msbuild/pull/3769#discussion_r219355209) [](ancestors = 219355209)

dfederm: Good call.

ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -1413,11 +1413,12 @@ public sealed partial class ProjectGraph
         public ProjectGraph(string entryProjectFile, Microsoft.Build.Evaluation.ProjectCollection projectCollection, System.Collections.Generic.Dictionary<string, string> globalProperties, string toolsVersion) { }
         public Microsoft.Build.Graph.ProjectGraphNode EntryProjectNode { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> ProjectNodes { get { throw null; } }
+        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, string[]> GetTargetLists(string[] entryProjectTargets) { throw null; }
```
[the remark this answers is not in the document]
dfederm: It needed to be public since it's called from XMake.cs, which is in another assembly.

Depending on how the scheduling bits take shape though, we can potentially remove this once we stop treating the scheduler as a black-box.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -60,14 +65,103 @@ public ProjectGraph(string entryProjectFile)
...
+                        edgesToVisit.Enqueue(projectReferenceEdge);
+                    }
+                }
+            }
+
+            // Dedupe target lists.
+            foreach (KeyValuePair<ProjectGraphNode, List<string>> pair in targetLists)
+            {
+                List<string> targetList = pair.Value;
+                HashSet<string> seenTargets = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
dfederm: Will take your work on that one :)

src/Build/Graph/ProjectGraph.cs
dfederm: Good catch. This was originally an `IList` so I did this, but forgot about `InsertRange` when I decided to make it a concrete `List`

src/Build/Graph/ProjectGraph.cs
```diff
@@ -106,5 +199,89 @@ private void LoadGraph(string entryProjectFile, ProjectCollection projectCollect
             }
         }
 
+        private static List<string> HydrateDefaultTargets(ProjectInstance project, List<string> targets)
+        {
+            // Avoid allocating until we have to
+            bool isMutated = false;
```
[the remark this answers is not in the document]
dfederm: `ImmutableList` causes a bunch of extra allocations though.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -60,14 +65,103 @@ public ProjectGraph(string entryProjectFile)
...
+                        i++;
+                    }
+                    else
+                    {
+                        targetList.RemoveAt(i);
+                    }
+                }
+            }
+
+            // Massage the final dictionary into string arrays instead of lists.
```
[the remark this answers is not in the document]
dfederm: Yea, this kind of sucks.

`BuildRequestData` only takes an array for the target list. What's more is that it defensively makes a copy of it since arrays are mutable and it could be user input when used programatically.

Furthermore, we can't add an internal ctor for this case since `XMake` is in a difference assembly than `BuildRequestData`.

We *could* add a new overload (well, a bunch of overloads) which take an immutable list since that guarantees that it can't be mutated (read only collection isn't enough; the concrete is still mutable so the caller might still do that). But that would be a significant API change.

So in the end I just decided to do this :(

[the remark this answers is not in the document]
dfederm: If we made this return an `ImmutableList<string>` instead, then the caller would need to do the array conversion anyway. However, maybe that's OK since at least it pushes the allocation out rather than always doing it. It's also just better to have immutability in a public API.

Potentially as we progress this feature a bit more, it may not even end up being public anyway, which means we could potentially further optimize then.

Will change the API.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -106,5 +203,95 @@ private void LoadGraph(string entryProjectFile, ProjectCollection projectCollect
             }
         }
 
+        private static ImmutableList<string> DetermineTargetsToPropagate(ProjectGraphNode node, ImmutableList<string> targets)
```
rainersigwald: nit: is input `entryTargets` or `targetsInCurrentProject` or something?

PR #3773, opened 2018-09-19
Add AspNetCore specific properties for csproj intellisense

jkotalik:
AspNetCore is adding a new feature to support inprocess hosting with IIS. We have a few csproj elements that can be specified that we would like to add for intellisense. There doesn't seem to be any extensibility points, so I went ahead and added them to the CommonTypes file.

cc/ @AndyGerlicher @shirhatti

### DIFF
--- a/src/MSBuild/Microsoft.Build.CommonTypes.xsd
+++ b/src/MSBuild/Microsoft.Build.CommonTypes.xsd
@@ -1100,6 +1100,16 @@ elementFormDefault="qualified">
         </xs:annotation>
     </xs:element>
     <xs:element name="AspNetConfiguration" type="msb:StringPropertyType" substitutionGroup="msb:Property"/>
+    <xs:element name="AspNetCoreHostingModel" type="msb:StringPropertyType" substitutionGroup="msb:Property">
+        <xs:annotation>
+            <xs:documentation><!-- _locID_textPackageId _locComment="" -->Indicates whether to run an ASP.NET Core application using IIS in-process or out-of-process.</xs:documentation>
+        </xs:annotation>
+    </xs:element>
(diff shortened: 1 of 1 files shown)


rainersigwald:
@jkotalik Do you want this for 15.9, or is 16.0 ok?

Also, this is the Right Place for this checkin, but we do localization only on the internal copy, so we'll need to dual-check-in to make this actually ship with VS.

rainersigwald:
Bouncing PR to retest.

jkotalik:
15.9

rainersigwald:
Can you please rebase onto the `vs15.9` branch, then? We're already in 16.0 for `master`.

jkotalik:
done.

src/MSBuild/Microsoft.Build.CommonTypes.xsd
```diff
@@ -1100,6 +1100,16 @@ elementFormDefault="qualified">
         </xs:annotation>
     </xs:element>
     <xs:element name="AspNetConfiguration" type="msb:StringPropertyType" substitutionGroup="msb:Property"/>
+    <xs:element name="AspNetCoreHostingModel" type="msb:StringPropertyType" substitutionGroup="msb:Property">
+        <xs:annotation>
+            <xs:documentation><!-- _locID_textPackageId _locComment="" -->Indicates whether to run an ASP.NET Core application using IIS in-process or out-of-process.</xs:documentation>
+        </xs:annotation>
+    </xs:element>
+    <xs:element name="AspNetCoreHostingName" type="msb:StringPropertyType" substitutionGroup="msb:Property">
```
shirhatti: Did you mean `AspNetCoreModuleName`?

jkotalik:
How should I handle merging this into VS? Should I make a separate PR or is there another process?

PR #3776, opened 2018-09-20
Unconditionally enable FEATURE_NAMED_PIPES_FULL_DUPLEX

rainersigwald:
We needed the condition before .NET Core supported duplex named pipes,
but we no longer support runtimes before that, so simplify the code by
committing to it in all cases.


radical:
@rainersigwald Nope, please go ahead.

PR #3779, opened 2018-09-20
Update ToolsVersion to `Current`

AndyGerlicher:
See #3778

### DIFF
--- a/src/Build/Evaluation/ToolsetProvider.cs
+++ b/src/Build/Evaluation/ToolsetProvider.cs
@@ -3,15 +3,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-
 using Microsoft.Build.Collections;
 using Microsoft.Build.Execution;
 using Microsoft.Build.Shared;
 using Microsoft.Build.BackEnd;
-
-using Constants = Microsoft.Build.Internal.Constants;
 
 namespace Microsoft.Build.Evaluation
 {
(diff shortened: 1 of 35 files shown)


src/Build/Utilities/Utilities.cs
```diff
@@ -323,137 +284,45 @@ internal static string CreateToolsVersionListString(IEnumerable<Toolset> toolset
...
+        /// <param name="usingDifferentToolsVersionFromProjectFile">true if the project file specifies an explicit ToolsVersion but a different one is chosen</param>
         /// <returns>The ToolsVersion we should use to build this project.  Should never be null.</returns>
-        internal static string GenerateToolsVersionToUse(string explicitToolsVersion, string toolsVersionFromProject, GetToolset getToolset, string defaultToolsVersion, out bool usingDifferentToolsVersionFromProjectFile)
+        internal static string GenerateToolsVersionToUse(string explicitToolsVersion, string toolsVersionFromProject,
+            GetToolset getToolset, string defaultToolsVersion, out bool usingDifferentToolsVersionFromProjectFile)
         {
-            string toolsVersionToUse = explicitToolsVersion;
-
-            // hosts may need to treat toolsversions later than the current one as the current one ... or may just 
-            // want to treat all toolsversions as though they're the current one, so give them that ability
```
rainersigwald: I grepped through VS and CPS and didn't see any uses of these environment variables.

AndyGerlicher:
FYI I added a commit that gets Roslyn from `..\..\15.0\Bin\Roslyn` so this can be inserted independently of a Roslyn install location change. This should be reverted as soon as possible though!

src/Build/Evaluation/ToolsetProvider.cs
```diff
@@ -31,7 +26,7 @@ internal class ToolsetProvider : IToolsetProvider, INodePacketTranslatable
         /// </summary>
         public ToolsetProvider(string defaultToolsVersion, PropertyDictionary<ProjectPropertyInstance> environmentProperties, PropertyDictionary<ProjectPropertyInstance> globalProperties, ToolsetDefinitionLocations toolsetDefinitionLocations)
```
jeffkl: can you remove the `defaultToolsVersion` input parameter?

Issue #3775, opened 2018-09-20
ProjectGraph should take GlobalProperties into account


dfederm:
A node should be a project configuration (project + global props) instead of just a project.

PR #3790, opened 2018-09-25
Account for global properties in graph nodes

dfederm:
Closes: #3775

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/MSBuild.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/MSBuild.cs
@@ -243,6 +243,7 @@ namespace Microsoft.Build.BackEnd
             }
 
             // Parse the global properties into a hashtable.
+            // The behavior of parsing global properties to define and undefine (below) combined with the behavior in Microsoft.Common.CurrentVersion.targets should match the logic in ProjectGraph.
             if (!PropertyParser.GetTableWithEscaping(Log, ResourceUtilities.GetResourceString("General.GlobalProperties"), "Properties", Properties, out Dictionary<string, string> propertiesTable))
             {
                 return false;
--- a/src/Build/BackEnd/Shared/ConfigurationMetadata.cs
+++ b/src/Build/BackEnd/Shared/ConfigurationMetadata.cs
@@ -44,6 +44,19 @@ namespace Microsoft.Build.BackEnd
         }
 
         /// <summary>
+        /// Constructor taking individual arguments.
+        /// </summary>
+        public ConfigurationMetadata(string projectFullPath, PropertyDictionary<ProjectPropertyInstance> globalProperties)
+        {
+            ErrorUtilities.VerifyThrowArgumentLength(projectFullPath, nameof(projectFullPath));
+            ErrorUtilities.VerifyThrowArgumentNull(globalProperties, nameof(globalProperties));
+
+            ProjectFullPath = projectFullPath;
+            ToolsVersion = MSBuildConstants.CurrentToolsVersion;
+            GlobalProperties = globalProperties;
+        }
+
+        /// <summary>
         /// The full path to the project to build.
         /// </summary>
         public string ProjectFullPath { get; }
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -4,10 +4,14 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Globalization;
 using System.Linq;
+using Microsoft.Build.BackEnd;
+using Microsoft.Build.Collections;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.Execution;
+using Microsoft.Build.Shared;
 
 namespace Microsoft.Build.Graph
 {
@@ -18,12 +22,22 @@ namespace Microsoft.Build.Graph
     {
         private const string ProjectReferenceItemName = "ProjectReference";
         private const string FullPathMetadataName = "FullPath";
+        private const string ToolsVersionMetadataName = "ToolsVersion";
+        private const string PropertiesMetadataName = "Properties";
+        private const string UndefinePropertiesMetadataName = "UndefineProperties";
+        private const string AdditionalPropertiesMetadataName = "AdditionalProperties";
+        private const string SetConfigurationMetadataName = "SetConfiguration";
+        private const string SetPlatformMetadataName = "SetPlatform";
+        private const string SetTargetFrameworkMetadataName = "SetTargetFramework";
+        private const string GlobalPropertiesToRemoveMetadataName = "GlobalPropertiesToRemove";
         private const string ProjectReferenceTargetsItemType = "ProjectReferenceTargets";
         private const string ProjectReferenceTargetsMetadataName = "Targets";
         private const string DefaultTargetsMarker = ".default";
 
-        private readonly Dictionary<string, ProjectGraphNode> _allParsedProjects =
-            new Dictionary<string, ProjectGraphNode>(StringComparer.OrdinalIgnoreCase);
+        private static readonly char[] PropertySeparator = { ';' };
+
+        private readonly Dictionary<ConfigurationMetadata, ProjectGraphNode> _allParsedProjects =
+            new Dictionary<ConfigurationMetadata, ProjectGraphNode>();
 
         // TODO: We probably want to mirror all relevant constructor overloads from Project
 
@@ -33,7 +47,7 @@ namespace Microsoft.Build.Graph
         /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
         public ProjectGraph(string entryProjectFile)
-            : this(entryProjectFile, ProjectCollection.GlobalProjectCollection, null, null)
+            : this(entryProjectFile, ProjectCollection.GlobalProjectCollection, null)
         {
         }
 
@@ -43,16 +57,33 @@ namespace Microsoft.Build.Graph
         /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
         /// <param name="projectCollection">The collection with which all projects in the graph should be associated. May not be null.</param>
         /// <param name="globalProperties">The global properties to use for all projects. May be null, in which case the containing project collection's global properties will be used.</param>
-        /// <param name="toolsVersion">The tools version. May be null.</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
         public ProjectGraph(
             string entryProjectFile,
             ProjectCollection projectCollection,
-            Dictionary<string, string> globalProperties,
-            string toolsVersion)
+            Dictionary<string, string> globalProperties)
         {
-            LoadGraph(entryProjectFile, projectCollection, globalProperties, toolsVersion);
-            EntryProjectNode = _allParsedProjects[entryProjectFile];
+            ErrorUtilities.VerifyThrowArgumentNull(projectCollection, nameof(projectCollection));
+
+            PropertyDictionary<ProjectPropertyInstance> globalPropertyDictionary;
+            if (globalProperties == null)
+            {
+                globalPropertyDictionary = new PropertyDictionary<ProjectPropertyInstance>(0);
+            }
+            else
+            {
+                globalPropertyDictionary = new PropertyDictionary<ProjectPropertyInstance>(globalProperties.Count);
+                foreach (KeyValuePair<string, string> entry in globalProperties)
+                {
+                    globalPropertyDictionary[entry.Key] = ProjectPropertyInstance.Create(entry.Key, entry.Value);
+                }
+            }
+
+            LoadGraph(
+                FileUtilities.NormalizePath(entryProjectFile),
+                projectCollection,
+                globalPropertyDictionary);
+            EntryProjectNode = _allParsedProjects[new ConfigurationMetadata(entryProjectFile, globalPropertyDictionary)];
             ProjectNodes = _allParsedProjects.Values;
         }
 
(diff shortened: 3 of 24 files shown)


src/Build/Graph/ProjectGraph.cs
```diff
@@ -51,8 +65,34 @@ public ProjectGraph(string entryProjectFile)
...
+                globalPropertyDictionary = new PropertyDictionary<ProjectPropertyInstance>(globalProperties.Count);
+                foreach (KeyValuePair<string, string> entry in globalProperties)
+                {
+                    globalPropertyDictionary[entry.Key] = ProjectPropertyInstance.Create(entry.Key, entry.Value);
+                }
+            }
+
+            if (string.IsNullOrEmpty(toolsVersion))
+            {
+                // TODO: This should actually be the project's tools version, but we haven't actually loaded it yet.
```
dfederm: We should perhaps discuss what we actually want to happen with this tools version thing all up before trying to solve this.

To solve this properly, we'll need to load the XML pretty early instead of passing around just the file name.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -159,14 +199,21 @@ public ProjectGraph(string entryProjectFile)
...
-            ProjectCollection projectCollection,
-            Dictionary<string, string> globalProperties,
-            string toolsVersion)
+            ConfigurationMetadata configurationMetadata,
+            ProjectCollection projectCollection)
         {
-            var project = new ProjectInstance(projectFilePath, globalProperties, toolsVersion, projectCollection);
-            var graphNode = new ProjectGraphNode(project);
-            _allParsedProjects.Add(projectFilePath, graphNode);
+            // TODO: ProjectInstance just converts the dictionary back to a PropertyDictionary, so find a way to directly provide it.
```
dfederm: Mainly just a perf TODO. IMHO, not something that needs to be addressed in this PR. Likely the solution is an internal ctor that only this code path uses, which is kinda meh.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -175,31 +222,38 @@ public ProjectGraph(string entryProjectFile)
...
                     foreach (var projectReferenceToParse in projectReferenceItems)
                     {
-                        string projectReferencePath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
-                        projectsToEvaluate.Enqueue(new KeyValuePair<string, ProjectGraphNode>(projectReferencePath, parsedProject));
+                        string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+                        string projectReferenceToolsVersion = projectReferenceToParse.GetMetadataValue(ToolsVersionMetadataName);
+                        if (string.IsNullOrEmpty(projectReferenceToolsVersion))
+                        {
+                            // TODO: This should actually be the project reference's tools version, but we haven't actually laoded it yet
+                            projectReferenceToolsVersion = toolsVersion;
```
rainersigwald: I'd lean more toward "throw if they tried to set ToolsVersion".

src/Build/Graph/ProjectGraph.cs
```diff
@@ -243,6 +297,92 @@ private static ImmutableList<string> ExpandDefaultTargets(ProjectInstance projec
...
+                    out Dictionary<string, string> propertiesTable))
+                {
+                    foreach (KeyValuePair<string, string> pair in propertiesTable)
+                    {
+                        properties[pair.Key] = ProjectPropertyInstance.Create(pair.Key, pair.Value);
+                    }
+                }
+                else
+                {
+                    throw new InvalidProjectFileException(string.Format(CultureInfo.InvariantCulture, ResourceUtilities.GetResourceString("General.InvalidPropertyError"), syntaxName, propertyNameAndValuesString));
```
dfederm: This isn't quite right. This error is logged to the task logger (if provided) in `PropertyParser.GetTableWithEscaping`, but the error is for a specific property/name pair and not the entire string. So the error string is a bit off here.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -243,6 +297,92 @@ private static ImmutableList<string> ExpandDefaultTargets(ProjectInstance projec
             return targets;
         }
 
+        /// <summary>
+        /// Gets the effective global properties for a project reference.
+        /// </summary>
+        /// <remarks>
+        /// The behavior of this method should match the logic in Microsoft.Common.CurrentVersion.targets and the MSBuild task.
```
rainersigwald: Can you add the inverse comment in the intrinsic MSBuild task?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -243,6 +297,92 @@ private static ImmutableList<string> ExpandDefaultTargets(ProjectInstance projec
...
+        /// </remarks>
+        private static PropertyDictionary<ProjectPropertyInstance> GetProjectReferenceGlobalProperties(ProjectItemInstance projectReference, PropertyDictionary<ProjectPropertyInstance> requesterGlobalProperties)
+        {
+            string propertiesString = projectReference.GetMetadataValue(PropertiesMetadataName);
+            string additionalPropertiesString = projectReference.GetMetadataValue(AdditionalPropertiesMetadataName);
+            string undefinePropertiesString = projectReference.GetMetadataValue(UndefinePropertiesMetadataName);
+            string globalPropertiesToRemoveString = projectReference.GetMetadataValue(GlobalPropertiesToRemoveMetadataName);
+
+            // TODO: Mimic AssignProjectConfiguration's behavior for determining the values for these.
```
rainersigwald: Bug for this? Or just a rollup "solutions in graph" one?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -243,6 +297,92 @@ private static ImmutableList<string> ExpandDefaultTargets(ProjectInstance projec
...
+            string setConfigurationString = projectReference.GetMetadataValue(SetConfigurationMetadataName);
+            string setPlatformString = projectReference.GetMetadataValue(SetPlatformMetadataName);
+            string setTargetFrameworkString = projectReference.GetMetadataValue(SetTargetFrameworkMetadataName);
+
+            // The properties on the project reference supercede the ones from the MSBuild task instad of appending.
```
rainersigwald: typo: `supersede`

rainersigwald: Also: really? That is confusing and sad.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -175,31 +222,38 @@ public ProjectGraph(string entryProjectFile)
...
                     IEnumerable<ProjectItemInstance> projectReferenceItems = parsedProject.Project.GetItems(ProjectReferenceItemName);
                     foreach (var projectReferenceToParse in projectReferenceItems)
                     {
-                        string projectReferencePath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
-                        projectsToEvaluate.Enqueue(new KeyValuePair<string, ProjectGraphNode>(projectReferencePath, parsedProject));
+                        string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+                        string projectReferenceToolsVersion = projectReferenceToParse.GetMetadataValue(ToolsVersionMetadataName);
+                        if (string.IsNullOrEmpty(projectReferenceToolsVersion))
+                        {
+                            // TODO: This should actually be the project reference's tools version, but we haven't actually laoded it yet
```
yash256: >laoded [](start = 124, length = 6)

typo

src/Build/Graph/ProjectGraph.cs
```diff
@@ -243,6 +297,92 @@ private static ImmutableList<string> ExpandDefaultTargets(ProjectInstance projec
...
+            // Append and remove properties as specified by the various metadata
+            MergeIntoPropertyDictionary(globalProperties, propertiesString, PropertiesMetadataName);
+            MergeIntoPropertyDictionary(globalProperties, additionalPropertiesString, AdditionalPropertiesMetadataName);
+            RemoveFromPropertyDictionary(globalProperties, globalPropertiesToRemoveString);
+            RemoveFromPropertyDictionary(globalProperties, undefinePropertiesString);
+
+            return globalProperties;
+        }
+
+        private static void MergeIntoPropertyDictionary(
```
rainersigwald: Should this be a method (or extension method) on `PropertyDictionary`? I feel like the mutations would read better that way.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -18,12 +22,22 @@ public sealed class ProjectGraph
...
+        private const string GlobalPropertiesToRemoveMetadataName = "GlobalPropertiesToRemove";
         private const string ProjectReferenceTargetsItemType = "ProjectReferenceTargets";
         private const string ProjectReferenceTargetsMetadataName = "Targets";
         private const string DefaultTargetsMarker = ".default";
 
-        private readonly Dictionary<string, ProjectGraphNode> _allParsedProjects =
-            new Dictionary<string, ProjectGraphNode>(StringComparer.OrdinalIgnoreCase);
+        private static readonly char[] PropertySeparator = { ';' };
+
+        private readonly Dictionary<ConfigurationMetadata, ProjectGraphNode> _allParsedProjects =
```
yash256: >_allParsedProjects [](start = 77, length = 18)

I don't understand the significance of having ConfigurationMetadata as the the key? Do we expect to have multiple entries for the same project path with different tools version and properties?

dfederm: Yes, that's exactly it. Different projects may depend on a project with different sets of global properties. This causes that project to build multiple times with those different sets of properties.

src/Build/Graph/ProjectGraph.cs
rainersigwald: Yeah, I think this can wait.

src/Build/Graph/ProjectGraph.cs
rainersigwald: I'm not sure I understand, can you paste the result of an error?

src/Build/Graph/ProjectGraph.cs
dfederm: Sure. In that case we can just hardcode something in `ConfigurationMetadata` (or just always use the ProjectCollection's default I guess) and just assume that at execution time the one specified in the project is the one that's actually used caching-wise.

src/Build/Graph/ProjectGraph.cs
dfederm: Sure, I'll open a new issue for it and throw it in some future Project

src/Build/Graph/ProjectGraph.cs
dfederm: Unless I'm reading the MSBuild task incorrectly, yea :(. I suspect that's why `AdditionalProperties` even exists. `UndefineProperties` seems to behave in the expected way.

src/Build/Graph/ProjectGraph.cs
dfederm: Unfortunately `PropertyDictionary` is actually `PropertyDictionary<T>` where `T` in this case is `ProjectPropertyInstance`. So a method like that would need to know how to convert stuff to `T` generically.

There is actually already a `ImportProperties<T>`, so I could actually just convert the string to an `IEnumerable<ProjectPropertyInstance>`, but really at that point we're not saving much.

I guess I could create an extension method for `PropertyDictionary<ProjectPropertyInstance>` specifically. That would allow shared code between the MSBuild task and here, but there's still the question about the weird-ish way I'm using `PropertyParser.GetTableWithEscaping` here.

src/Build/Graph/ProjectGraph.cs
dfederm: Yea, that was super unclear :)

So `PropertyParser.GetTableWithEscaping` takes a task log helper and proceeds to process each split one by one. So like "A=B;C=D" would get processed one by one "A=B" and "C=D". It does have to process them all at once though in the special case when it has to merge entries, like "A=B;C" would get split to "A=B" and "C", but then merged back together such that "A" would get set to "B;C". When it does see something invalid though, it errors on that specific entry, not the entire input. So something like "A;B=C" would split to "A" and "B=C" and then there would be an error about "A" (not "A;B=C"). Something like "A=B;=;=;=" I would guess would be similarly invalid, but the error just mention "=" as the bad string.

Here's the actual resource string in case it helps:
> MSB3100: Syntax for "{0}" parameter is not valid ({1}). Correct syntax is {0}="&lt;name&gt;=&lt;value&gt;".

TL;DR, the error message is for a specific entry, not the entire string.

src/Build/Graph/ProjectGraph.cs
dfederm: Actually, not really any shareable code between this and the MSBuild task, as it uses Dictionary<string, string> while we use `PropertyDictionary<ProjectPropertyInstance>`. So besides a nicer syntax, there's not really much benefit here to having an extension method vs just a static method in this class.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/MSBuild.cs
```diff
@@ -243,6 +243,7 @@ public async Task<bool> ExecuteInternal()
             }
 
             // Parse the global properties into a hashtable.
+            // The behavior of parsing global properties to define and undefine (below) combined with the behavior in Microsoft.Common.CurrentVersion.targets should match the logic in and ProjectGraph.
```
rainersigwald: `in and` -> `in`?

src/Build/BackEnd/Shared/ConfigurationMetadata.cs
```diff
@@ -43,6 +43,20 @@ public ConfigurationMetadata(Project project)
             ProjectFullPath = FileUtilities.NormalizePath(project.FullPath);
         }
 
+        /// <summary>
+        /// Constructor taking individual arguments.
+        /// </summary>
+        public ConfigurationMetadata(string projectFullPath, string toolsVersion, PropertyDictionary<ProjectPropertyInstance> globalProperties)
```
AndyGerlicher: This class is internal, but can we remove tools version? Just use `Constants.CurrentToolsVersion`.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -43,16 +57,41 @@ public ProjectGraph(string entryProjectFile)
...
         /// <param name="projectCollection">The collection with which all projects in the graph should be associated. May not be null.</param>
         /// <param name="globalProperties">The global properties to use for all projects. May be null, in which case the containing project collection's global properties will be used.</param>
-        /// <param name="toolsVersion">The tools version. May be null.</param>
+        /// <param name="toolsVersion">The tools version to use for all projects in the graph. May be null, in which case the project collection's default tools version will be used.</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
         public ProjectGraph(
             string entryProjectFile,
             ProjectCollection projectCollection,
             Dictionary<string, string> globalProperties,
             string toolsVersion)
```
AndyGerlicher: I really think toolsVersion should be hidden from the API entirely. **When** we deprecate it, it just means we have to change the api here as well.

src/Build/BackEnd/Shared/ConfigurationMetadata.cs
dfederm: I'll remove it from this ctor overload. I don't think I can remove it from the entire class without breaking things :)

Issue #3751, opened 2018-09-15
Define p2p contracts for core targets


dfederm:
Need to define properties and `ProjectReferenceTargets` items for core targets (Build, Clean, Rebuild, maybe Restore?)

Eg.
```
<PropertyGroup>
  <ProjectReferenceTargetsForClean>GetTargetFrameworks;Clean</ProjectReferenceTargetsForClean>
</PropertyGroup>

<ItemGroup>
  <ProjectReferenceTargets Include="Clean" Targets="$(ProjectReferenceTargetsForClean)"/>
</ItemGroup>
```

PR #3800, opened 2018-09-26
Define ProjectReferenceTargets for Build, Clean, and Rebuild

dfederm:
Closes: #3751

### DIFF
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -79,8 +79,9 @@ namespace Microsoft.Build.Graph
                 }
             }
 
+            entryProjectFile = FileUtilities.NormalizePath(entryProjectFile);
             LoadGraph(
-                FileUtilities.NormalizePath(entryProjectFile),
+                entryProjectFile,
                 projectCollection,
                 globalPropertyDictionary);
             EntryProjectNode = _allParsedProjects[new ConfigurationMetadata(entryProjectFile, globalPropertyDictionary)];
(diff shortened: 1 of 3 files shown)


src/Build/Graph/ProjectGraph.cs
```diff
@@ -79,8 +79,9 @@ public ProjectGraph(string entryProjectFile)
                 }
             }
 
+            entryProjectFile = FileUtilities.NormalizePath(entryProjectFile);
```
dfederm: Minor unrelated bug I found while testing this. Basically from the command line, the path given isn't absolute usually so the lookup below on line 87 was failing. Wasn't caught in the UT we have for this since the UT uses absolute paths when invoking MSBuild.exe.

src/Tasks/Microsoft.Common.props
```diff
@@ -148,6 +148,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.VisualStudioVersion.v*.Common.props" Condition="'$(VisualStudioVersion)' == ''" />
 
+  <!--
+    Properties for extension of ProjectReferenceTargets.
+    Append any current value which may have been provided in a Directory.Build.props since the intent was likely to append, not prepend.
```
dfederm: The exact location of this seems important.

If appended to in Directly.Build.props, this logic will append what the user did instead of append our stuff. The assumption being that usually, people would want to extend stuff after build, not before (???)

Also, if appending to `ProjectReferenceTargetsForBuild` (or the Clean one), this *will* affect Rebuild if done early enough, but not if done later. If done later, they'd need to append to the Rebuild one as well. This I guess would only be for folks who did it directly in their project, before the common targets.

The project extensions (like nuget.g.props) happens before this so has the same behavior as doing it in Directory.Build.props.

If extending these in directory.build.props, I *think* it's too late.

AndyGerlicher: This seems right to me. @rainersigwald ?

rainersigwald: Filed https://github.com/Microsoft/msbuild/issues/3808 to keep track of things like this that will eventually need to be documented.

Issue #3755, opened 2018-09-15
Add command line argument for graph builds


dfederm:
We added a command line argument for isolated builds, but we'll likely need one for graph builds as well.

PR #3806, opened 2018-09-27
Add /graphBuild parameter

dfederm:
The reasoning for yet another param (in addition to `/isolateProjects`) is based on these various scenarios:

1. `/graphBuild` - Construct and build a graph without isolation enforcement ("unexpected" calls into project references are permitted)
2. `/graphBuild /isolateProjects` - Construct and build a graph with isolation enforcement.
3. `/isolateProjects` - Build one single project with isolation enforcement. This scenario will require deserializing results from dependencies in some way (yet to be finalized).

Closes: #3755

### DIFF
--- a/src/MSBuild/CommandLineSwitches.cs
+++ b/src/MSBuild/CommandLineSwitches.cs
@@ -101,6 +101,7 @@ namespace Microsoft.Build.CommandLine
             RestoreProperty,
             Interactive,
             IsolateProjects,
+            GraphBuild,
             NumberOfParameterizedSwitches,
         }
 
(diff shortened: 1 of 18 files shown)


src/MSBuild/CommandLineSwitches.cs
```diff
@@ -277,6 +278,7 @@ bool emptyParametersAllowed
             new ParameterizedSwitchInfo(  new string[] { "restoreproperty", "rp" },             ParameterizedSwitch.RestoreProperty,            null,                           true,           "MissingRestorePropertyError",         true,   false  ),
             new ParameterizedSwitchInfo(  new string[] { "interactive" },                       ParameterizedSwitch.Interactive,                null,                           false,          null,                                  true,   false  ),
             new ParameterizedSwitchInfo(  new string[] { "isolateprojects", "isolate" },        ParameterizedSwitch.IsolateProjects,            null ,                          false,          null,                                  true,   false  ),
+            new ParameterizedSwitchInfo(  new string[] { "graphbuild", "graph" },               ParameterizedSwitch.GraphBuild,                 null ,                          false,          null,                                  true,   false  ),
```
rainersigwald: We should accept `msbuild /📈` as a shortcut for this! Or `/🕸️`.

😇

src/MSBuild/Resources/Strings.resx
```diff
@@ -730,12 +730,30 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
   </data>
+  <data name="HelpMessage_36_GraphBuildSwitch" UESanitized="false" Visibility="Public">
+    <value>  /graphBuild[:True|False]
+                     Causes MSBuild to construct and build a project graph.
+
+                     Constructing a graph involves identifying project
+                     references to form dependencies. Building that graph
+                     involves attempting to build project references prior
+                     to the projects that reference them, differing from
+                     traditional MSBuild scheduling.
```
rainersigwald: This is good for now. I suspect eventually we'll probably want something more like "this imposes requirements on your project, see aka.ms/msbuild/static for details".

dfederm: At some point prior to v16 release we'll probably want to do a full pass of all the user-facing strings we're adding with these features, potentially with a technical writer or something.

Issue #3756, opened 2018-09-15
Add project graph convenience methods and overloads


dfederm:
Probably just look at the `Project` ctors and copy all relevant overloads for `ProjectGraph`.

PR #3807, opened 2018-09-27
Add convenience constructors to ProjectGraph

dfederm:
Closes: #3756

Since we decided to not worry about any `toolsVersion`-related stuff, and since adding support for directly receiving a `ProjectRootElement` would be exceedingly complicated, there was really only need for one additional constructor.

This will balloon a little bit once we support multiple entry points though.


src/Build/Graph/ProjectGraph.cs
```diff
@@ -39,29 +39,40 @@ public sealed class ProjectGraph
...
-        /// <param name="projectCollection">The collection with which all projects in the graph should be associated. May not be null.</param>
         /// <param name="globalProperties">The global properties to use for all projects. May be null, in which case the containing project collection's global properties will be used.</param>
+        /// <param name="projectCollection">The collection with which all projects in the graph should be associated. May not be null.</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
         public ProjectGraph(
             string entryProjectFile,
-            ProjectCollection projectCollection,
-            Dictionary<string, string> globalProperties)
+            IDictionary<string, string> globalProperties,
+            ProjectCollection projectCollection)
```
dfederm: I reversed the param ordering here to better match `Project` and `ProjectInstance` ctors.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -39,29 +39,40 @@ public sealed class ProjectGraph
         private readonly Dictionary<ConfigurationMetadata, ProjectGraphNode> _allParsedProjects =
             new Dictionary<ConfigurationMetadata, ProjectGraphNode>();
 
-        // TODO: We probably want to mirror all relevant constructor overloads from Project
-
         /// <summary>
-        /// Constructs a graph starting from the given project file.
+        /// Constructs a graph starting from the given project file, evaluating with the global project collection's global properties.
```
rainersigwald: it's not the project collection's global properties is it? Isn't it just global PC and no specified properties?

dfederm: Oops, that's actually a bug on our end then. `Project` and `ProjectInstance` seem to use the PC's props as a base and applies anything else provided in addition. Currently `ProjectGraph` doesn't use the PC's props at all.

So the documentation says the correct intent, but the code doesn't follow that. Will fix that here.

dfederm: Actually, I'm wrong again. `Project` uses PC as a base, while `ProjectInstance` does not. I'll just correct the comment then and we can think about what that "right" behavior is separately.

PR #3810, opened 2018-09-28
Detect cycles in a graph

yash256:
Detect a cycle during graph creation and print all the projects that are involved in a cycle.

### DIFF
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Text;
+using Microsoft.Build.BackEnd;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.UnitTests;
@@ -97,10 +98,25 @@ namespace Microsoft.Build.Graph.UnitTests
                 TransientTestFile entryProject = CreateProject(env, 1, new[] { 2 });
                 CreateProject(env, 2, new[] { 3 });
                 CreateProject(env, 3, new[] { 1 });
+                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path));
+            }
+        }
 
-                // TODO: This should eventually throw, but for now not infinite-looping is sufficient.
-                ProjectGraph graph = new ProjectGraph(entryProject.Path);
-                graph.ProjectNodes.Count.ShouldBe(3);
+        [Fact]
+        // graph with a cycle between 2->3->6->7->2
+        public void ConstructBigGraphWithCycle()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                TransientTestFile entryProject = CreateProject(env, 1, new[] {2,3,4});
+                CreateProject(env, 2, new[] {5, 6});
+                CreateProject(env, 3, new[] {2, 8});
+                CreateProject(env, 4);
+                CreateProject(env, 5);
+                CreateProject(env, 6, new[] { 7});
+                CreateProject(env, 7, new[] { 3 });
+                CreateProject(env, 8);
+                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path));
             }
         }
 
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -103,7 +103,19 @@ namespace Microsoft.Build.Graph.UnitTests
         }
 
         [Fact]
-        // graph with a cycle between 2->3->6->7->2
+        public void ConstructWithSelfLoop()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                TransientTestFile entryProject = CreateProject(env, 1, new[] { 2, 3 });
+                CreateProject(env, 2, new[] { 2 });
+                CreateProject(env, 3);
+                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path));
+            }
+        }
+
+        [Fact]
+        // graph with a cycle between 2->6->7->3->2
         public void ConstructBigGraphWithCycle()
         {
             using (var env = TestEnvironment.Create())
@@ -112,11 +124,16 @@ namespace Microsoft.Build.Graph.UnitTests
                 CreateProject(env, 2, new[] {5, 6});
                 CreateProject(env, 3, new[] {2, 8});
                 CreateProject(env, 4);
-                CreateProject(env, 5);
-                CreateProject(env, 6, new[] { 7});
-                CreateProject(env, 7, new[] { 3 });
+                CreateProject(env, 5, new []{9, 10});
+                var proj6 = CreateProject(env, 6, new[] { 7});
+                var proj7 = CreateProject(env, 7, new[] { 3 });
                 CreateProject(env, 8);
-                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path));
+                CreateProject(env, 9);
+                CreateProject(env, 10);
+                string projects = proj6.Path + Environment.NewLine + proj7.Path;
+                // the error message can contain a cycle from 2->3->6->7 or 3->6->7->2
+                // the error message should always contain 6.proj and 7.proj
+                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path)).Message.ShouldContain(projects);
             }
         }
 
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -96,9 +96,11 @@ namespace Microsoft.Build.Graph.UnitTests
             using (var env = TestEnvironment.Create())
             {
                 TransientTestFile entryProject = CreateProject(env, 1, new[] { 2 });
-                CreateProject(env, 2, new[] { 3 });
-                CreateProject(env, 3, new[] { 1 });
-                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path));
+                var proj2 = CreateProject(env, 2, new[] { 3 });
+                var proj3 = CreateProject(env, 3, new[] { 1 });
+                var projectsInCycle = new List<string>() {entryProject.Path, proj3.Path, proj2.Path, entryProject.Path};
+                string expectedErrorMessage = ProjectGraph.FormatCircularDependencyError(projectsInCycle);
+                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path)).Message.ShouldContain(expectedErrorMessage.ToString());
             }
         }
 
@@ -121,8 +123,8 @@ namespace Microsoft.Build.Graph.UnitTests
             using (var env = TestEnvironment.Create())
             {
                 TransientTestFile entryProject = CreateProject(env, 1, new[] {2,3,4});
-                CreateProject(env, 2, new[] {5, 6});
-                CreateProject(env, 3, new[] {2, 8});
+                var proj2 = CreateProject(env, 2, new[] {5, 6});
+                var proj3 = CreateProject(env, 3, new[] {2, 8});
                 CreateProject(env, 4);
                 CreateProject(env, 5, new []{9, 10});
                 var proj6 = CreateProject(env, 6, new[] { 7});
@@ -130,10 +132,9 @@ namespace Microsoft.Build.Graph.UnitTests
                 CreateProject(env, 8);
                 CreateProject(env, 9);
                 CreateProject(env, 10);
-                string projects = proj6.Path + Environment.NewLine + proj7.Path;
-                // the error message can contain a cycle from 2->3->6->7 or 3->6->7->2
-                // the error message should always contain 6.proj and 7.proj
-                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path)).Message.ShouldContain(projects);
+                var projectsInCycle = new List<string>(){proj2.Path, proj3.Path, proj7.Path, proj6.Path, proj2.Path };
+                var errorMessage = ProjectGraph.FormatCircularDependencyError(projectsInCycle);
+                Should.Throw<CircularDependencyException>(() => new ProjectGraph(entryProject.Path)).Message.ShouldContain(errorMessage.ToString());
             }
         }
 
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -93,7 +93,7 @@ namespace Microsoft.Build.Graph
 
             var entryProjectConfigurationMetadata = new ConfigurationMetadata(FileUtilities.NormalizePath(entryProjectFile), globalPropertyDictionary);
             var nodeStates = new Dictionary<ConfigurationMetadata, NodeState>();
-            LoadGraph(entryProjectConfigurationMetadata, nodeStates, projectCollection, globalPropertyDictionary);
+            ProcessNode(entryProjectConfigurationMetadata, nodeStates, projectCollection, globalPropertyDictionary);
             EntryProjectNode = _allParsedProjects[entryProjectConfigurationMetadata];
             ProjectNodes = _allParsedProjects.Values;
         }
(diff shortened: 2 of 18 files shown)


src/Build/Graph/ProjectGraph.cs
```diff
@@ -85,6 +86,7 @@ public ProjectGraph(string entryProjectFile)
                 globalPropertyDictionary);
             EntryProjectNode = _allParsedProjects[new ConfigurationMetadata(entryProjectFile, globalPropertyDictionary)];
             ProjectNodes = _allParsedProjects.Values;
+            DetectCyclesInGraph();
```
yash256: >DetectCyclesInGraph(); [](start = 12, length = 22)

This does another traversal of the graph after LoadGraph. I am working on combining the 2 so that cycle detection can be done in the same traversal in which projects are discovered

src/Build/Graph/ProjectGraph.cs
```diff
@@ -250,6 +252,84 @@ private void LoadGraph(string entryProjectFile, ProjectCollection projectCollect
             }
         }
 
+        // Using Tarjan's algorithm for detecting strongly connected components in a directed graph
+        // Reference: https://en.wikipedia.org/wiki/Strongly_connected_component
+        private int _index = 0;
```
dfederm: Does this need to be a field on the class? Could it just be passed around as a param?

If not, rename it to something less generic (`_cycleDetectionDepth`?)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -250,6 +252,84 @@ private void LoadGraph(string entryProjectFile, ProjectCollection projectCollect
...
+                    }
+                    lowlink[projectNode] = Math.Min(lowlink[projectNode], lowlink[projectReference]);
+                }
+                else if (existsOnStack.Contains(projectReference))
+                {
+                    lowlink[projectNode] = Math.Min(lowlink[projectNode], indices[projectReference]);
+                }
+            }
+
+            if (lowlink[projectNode] == indices[projectNode])
```
ccastanedaucf: I believe you'll have to add a check for self-loops since this will fail on a self-referential dependency. Consider a graph such as
```
0 -> 1
⥁
```
where we've already visited ```1``` so the state looks like this:
```
indices: [0, 1]
lowLink: [0, 1]
stack: 0
existsOnStack: [T, F]
```

When iterating ```ProjectReferences``` from the first ```ProjectGraphNode```,  ```indices.ContainsKey(0) == true```and we won't add it to the stack again. Then although ```existsOnStack.Contains(0) == true``` and we would typically record a cycle, ```Math.min(lowlink[0], indices[0])``` will set ```lowlink[0]``` to ```0```. Then ```lowlink[projectNode] == indices[projectNode]``` is true, which triggers popping SCC from the stack. Since ```0``` is only on the stack once and ```0 == 0```, ```countOfProjectsInCycle == 1``` although a cycle exists.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -250,6 +252,84 @@ private void LoadGraph(string entryProjectFile, ProjectCollection projectCollect
             }
         }
 
+        // Using Tarjan's algorithm for detecting strongly connected components in a directed graph
+        // Reference: https://en.wikipedia.org/wiki/Strongly_connected_component
```
ccastanedaucf: Tarjan's might be overkill here unless we're doing something with _all_ subgraphs of strongly connected components, but it looks like we just detect and print _some_ cycle.

Not too familiar with the static graph work so it's not an issue if this is just being called once for a build, but something like using a simple tri-colored DFS and a stack for printing the cycle would result in less allocations and overhead, and still ```O(|V| + |E|)```.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,100 @@ public ProjectGraph(string entryProjectFile)
...
+                if (nodeState.TryGetValue(projectReferenceConfigurationMetadata, out NodeState projectReferenceNodeState))
                 {
-                    parsedProject = CreateNewNode(projectToEvaluate, projectCollection);
-                    IEnumerable<ProjectItemInstance> projectReferenceItems = parsedProject.Project.GetItems(ProjectReferenceItemName);
-                    foreach (var projectReferenceToParse in projectReferenceItems)
+                    // a project reference can be in "Processed" state. If it is "InProcess" state, it is an ancestor and there is a circular dependency
+                    if (projectReferenceNodeState == NodeState.InProcess)
                     {
-                        if (!string.IsNullOrEmpty(projectReferenceToParse.GetMetadataValue(ToolsVersionMetadataName)))
+                        if (projectToEvaluate.ProjectFullPath.Equals(projectReferenceConfigurationMetadata.ProjectFullPath))
```
dfederm: `ConfigurationMetadata` is equatable, so I think you should just call `projectToEvaluate.Equals(projectReferenceConfigurationMetadata)`

src/Build/Resources/Strings.resx
```diff
@@ -1707,13 +1707,16 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
       LOCALIZATION:  Do not localize the following words: ProjectGraph, ProjectReference, ToolsVersion.
     </comment>
   </data>
+  <data name="CircularDependencyInProjectGraph" UESanitized="false" Visibility="Public">
+    <value>MSB4251: Exception occurred while creating depenendency graph: There is a circular dependency involving the following projects: {0}</value>
```
dfederm: Remove the "Exception occurred while creating depenendency graph: " part.

src/Build/Resources/Strings.resx
```diff
@@ -1707,13 +1707,16 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
...
+  </data>
```
dfederm: I think you need the `<comment>` element to explain to the Localizers what to do. Follow the pattern used in some of the other errors. The "CircularDependency" string in particular seems like a good one to base off of in this case.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,100 @@ public ProjectGraph(string entryProjectFile)
...
+                    var loadReference = LoadGraph(projectReferenceConfigurationMetadata, nodeState, projectCollection,
+                        globalProperties);
+                    if (!loadReference.success)
+                    {
+                        if (loadReference.projectsInCycle[0].Equals(parsedProject.Project.FullPath))
+                        {
+                            // we have reached the nth project in the cycle, form error message and throw
+                            loadReference.projectsInCycle.Add(projectReferenceConfigurationMetadata.ProjectFullPath);
+                            var errorMessage = new StringBuilder(500);
+                            for (int i = loadReference.projectsInCycle.Count-1; i >= 0; i--)
```
dfederm: Style: Spaces around operators (`loadReference.projectsInCycle.Count - 1`)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,100 @@ public ProjectGraph(string entryProjectFile)
...
+
         /// <summary>
         /// Load a graph with root node at entryProjectFile
-        /// Maintain a queue of projects to be processed- each queue item is a key value pair of the project to be evaluated and its parent
-        /// Once the project has been evaluated, add a project reference to this evaluated target from the parent node
+        /// Maintain the state of each node (InProcess and Processed) to detect cycles
+        /// returns false if loading the graph is not successful
         /// </summary>
-        private void LoadGraph(string entryProjectFile, ProjectCollection projectCollection, PropertyDictionary<ProjectPropertyInstance> globalProperties)
+        private (bool success, List<string> projectsInCycle) LoadGraph(ConfigurationMetadata projectToEvaluate,
```
dfederm: Since this is now a recursive algorithm, I think "ProcessNode" might be a better name for this method.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -97,10 +98,42 @@ public void ConstructWithCycle()
...
+                CreateProject(env, 4);
+                CreateProject(env, 5, new []{9, 10});
+                var proj6 = CreateProject(env, 6, new[] { 7});
+                var proj7 = CreateProject(env, 7, new[] { 3 });
+                CreateProject(env, 8);
+                CreateProject(env, 9);
+                CreateProject(env, 10);
+                string projects = proj6.Path + Environment.NewLine + proj7.Path;
+                // the error message can contain a cycle from 2->3->6->7 or 3->6->7->2
+                // the error message should always contain 6.proj and 7.proj
```
dfederm: Shouldn't it also contain 2 and 3?

Seems like the error should be something like:
```
2.proj
3.proj
7.proj
6.proj
2.proj
```

IE, the first and last line should be the same to complete the cycle.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,100 @@ public ProjectGraph(string entryProjectFile)
...
+                    if (!loadReference.success)
+                    {
+                        if (loadReference.projectsInCycle[0].Equals(parsedProject.Project.FullPath))
+                        {
+                            // we have reached the nth project in the cycle, form error message and throw
+                            loadReference.projectsInCycle.Add(projectReferenceConfigurationMetadata.ProjectFullPath);
+                            var errorMessage = new StringBuilder(500);
+                            for (int i = loadReference.projectsInCycle.Count-1; i >= 0; i--)
+                            {
+                                errorMessage.Append(loadReference.projectsInCycle[i]).AppendLine();
```
dfederm: Instead of newlines, would something like "->" make more sense (although you'd need to account for the leading/trailing arrow.

The error showing: `2.proj -> 6.proj -> 7.proj -> 3.proj -> 2.proj` seems super clear as to what the cycle is and in what direction.

dfederm: Or, if with full file paths the string gets really long, maybe prepend the lines with an error?

I'm not really feeling super strongly about this suggestion though. Maybe look at a real error message and use a judgement call.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,101 @@ public ProjectGraph(string entryProjectFile)
...
+            ProjectCollection projectCollection,
+            PropertyDictionary<ProjectPropertyInstance> globalProperties)
         {
-            var projectsToEvaluate = new Queue<KeyValuePair<ConfigurationMetadata, ProjectGraphNode>>();
-            // entry project node has no parent
-            projectsToEvaluate.Enqueue(new KeyValuePair<ConfigurationMetadata, ProjectGraphNode>(new ConfigurationMetadata(entryProjectFile, globalProperties), null));
-            while (projectsToEvaluate.Count != 0)
+            nodeState[projectToEvaluate] = NodeState.InProcess;
+            ProjectGraphNode parsedProject = CreateNewNode(projectToEvaluate, projectCollection);
+            IEnumerable<ProjectItemInstance> projectReferenceItems =
```
dfederm: What're your editor line wrap settings? It seems to be super aggressive with line wrapping.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
yash256: It will contain 2 and 3. Since 2 and 3 are at the same depth, I am not sure if 2 will always be processed first. If 3 gets processed first, then our error message will contain 3->2->6->7->3. Asserting the exact order may result in flakiness

src/Build/Graph/ProjectGraph.cs
yash256: This is how the final error message looks like:

MSB4251: There is a circular dependency involving the following projects: 
C:\Users\yaupadh\AppData\Local\Temp\csfr4ykv.2te\Temporaryb33fd5235aa84433ad2e1717bf71add3\2.proj ->
C:\Users\yaupadh\AppData\Local\Temp\csfr4ykv.2te\Temporaryb33fd5235aa84433ad2e1717bf71add3\6.proj ->
C:\Users\yaupadh\AppData\Local\Temp\csfr4ykv.2te\Temporaryb33fd5235aa84433ad2e1717bf71add3\7.proj ->
C:\Users\yaupadh\AppData\Local\Temp\csfr4ykv.2te\Temporaryb33fd5235aa84433ad2e1717bf71add3\3.proj ->
C:\Users\yaupadh\AppData\Local\Temp\csfr4ykv.2te\Temporaryb33fd5235aa84433ad2e1717bf71add3\2.proj

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
+                string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+
+                PropertyDictionary<ProjectPropertyInstance> projectReferenceGlobalProperties = GetProjectReferenceGlobalProperties(projectReferenceToParse, globalProperties);
+                var projectReferenceConfigurationMetadata = new ConfigurationMetadata(projectReferenceFullPath, projectReferenceGlobalProperties);
+                if (nodeState.TryGetValue(projectReferenceConfigurationMetadata, out NodeState projectReferenceNodeState))
                 {
-                    parsedProject = CreateNewNode(projectToEvaluate, projectCollection);
-                    IEnumerable<ProjectItemInstance> projectReferenceItems = parsedProject.Project.GetItems(ProjectReferenceItemName);
-                    foreach (var projectReferenceToParse in projectReferenceItems)
+                    // a project reference can be in "Processed" state. If it is "InProcess" state, it is an ancestor and there is a circular dependency
```
rainersigwald: Can you clarify this comment? Maybe something like

> Because this is a depth-first search, we should only encounter new nodes or nodes whose subgraph has been completely processed. If we encounter a node that is currently being processed, it must be one of our ancestors in a circular dependency.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
+                        if (projectToEvaluate.Equals(projectReferenceConfigurationMetadata))
                         {
-                            throw new InvalidOperationException(string.Format(
-                                CultureInfo.InvariantCulture,
-                                ResourceUtilities.GetResourceString("ProjectGraphDoesNotSupportProjectReferenceWithToolset"),
-                                projectReferenceToParse.EvaluatedInclude,
-                                parsedProject.Project.FullPath));
+                            // the project being evaluated has a reference on itself
+                            throw new CircularDependencyException(string.Format(
+                                ResourceUtilities.GetResourceString("CircularDependencyInProjectGraph"),
```
rainersigwald: I think this should either have a dedicated "self-referencing" error or explicitly construct the error as

```
/path/to/SelfReferencing.csproj
->
/path/to/SelfReferencing.csproj
```

Just a single entry wouldn't be obvious.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
                     {
-                        if (!string.IsNullOrEmpty(projectReferenceToParse.GetMetadataValue(ToolsVersionMetadataName)))
+                        if (projectToEvaluate.Equals(projectReferenceConfigurationMetadata))
                         {
-                            throw new InvalidOperationException(string.Format(
-                                CultureInfo.InvariantCulture,
-                                ResourceUtilities.GetResourceString("ProjectGraphDoesNotSupportProjectReferenceWithToolset"),
-                                projectReferenceToParse.EvaluatedInclude,
-                                parsedProject.Project.FullPath));
+                            // the project being evaluated has a reference on itself
```
rainersigwald: nit: _to_ itself

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
+                            // the project being evaluated has a circular dependency involving multiple projects
+                            // add this project to the list of projects involved in cycle 
+                            var projectsInCycle = new List<string> {projectReferenceConfigurationMetadata.ProjectFullPath};
+                            return (false, projectsInCycle);
+                        }
+                    }
+                }
+                else
+                {
+                    // a new project that has to be evaluated
```
rainersigwald: maybe

```c#
// Recursively walk references of newly-discovered project
```

?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
+                            loadReference.projectsInCycle.Add(projectReferenceConfigurationMetadata.ProjectFullPath);
+                            loadReference.projectsInCycle.Add(parsedProject.Project.FullPath);
+                            var errorMessage = FormatCircularDependencyError(loadReference.projectsInCycle);
+                            throw new CircularDependencyException(string.Format(
+                                ResourceUtilities.GetResourceString("CircularDependencyInProjectGraph"),
+                                errorMessage));
+                        }
+                        else
+                        {
+                            loadReference.projectsInCycle.Add(projectReferenceConfigurationMetadata.ProjectFullPath);
```
rainersigwald: comment, please.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
+                parsedProject.AddProjectReference(_allParsedProjects[projectReferenceConfigurationMetadata]);
+            }
+
+            nodeState[projectToEvaluate] = NodeState.Processed;
+            return (true, null);
+        }
+
+        internal static string FormatCircularDependencyError(List<string> projectsInCycle)
+        {
+            var errorMessage = new StringBuilder(500);
```
rainersigwald: comment magic constant? or allocate based on something like `MAX_PATH * projectsInCycle.Count`?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -219,47 +218,113 @@ public ProjectGraph(string entryProjectFile)
...
+            errorMessage.AppendLine();
+            for (int i = projectsInCycle.Count - 1; i >= 0; i--)
```
rainersigwald: It makes me so sad that they only added `StringBuilder.AppendJoin` in .NET Core 2. 😕

PR #3818, opened 2018-10-02
added missing years to copyright statement. Year dates required for M…

james-flynn-ie:
MIT license requires the year to be stated in the copyright statement. (https://opensource.org/licenses/MIT)

Added dates 2003-2018, where 2003 was first release of MSBuild.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,7 +2,7 @@ MSBuild
 
 The MIT License (MIT)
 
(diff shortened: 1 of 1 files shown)


rainersigwald:
Thanks for the contribution! Our legal department would need to be involved with any change to our license and this doesn't seem important enough to involve them (and would require annual maintenance) so I don't think we should accept this change.

PR #3819, opened 2018-10-02
Make StronglyTypedResourceGenerator work on .NET Core

wjk:
Title says it all, really. Have tested with the project attached below. The STR code-behind is generated containing both a string and non-string resource (a Bitmap, to be precise) and compiles successfully. Code is very messy, however; I would appreciate some advice on how it should be cleaned up. Thanks!

Relates to: #2221, https://github.com/dotnet/sdk/pull/1782

[msbuild_test.zip](https://github.com/Microsoft/msbuild/files/2440080/msbuild_test.zip)


wjk:
Note that `ResXFileRef` resource entries won't be handled properly with the code in this PR. The generated property will reference the type `ResXFileRef`, not `Stream`, `Bitmap`, or whatever other type is correct in this case. A pointer on how to fix this will also be appreciated.

wjk:
I *think* I've fixed ResXFileRef support, but I'm not certain.

rainersigwald:
@nguerrera and @tanya-solyanik, do you have thoughts about this approach of reimplementing some of the resx reading instead of (I assume until) we can wait to get `System.Windows.Forms` and a full `ResXResourceReader`? It seems like a fine approach to me.

src/Tasks/system.design/stronglytypedresourcebuilder.cs
```diff
@@ -497,34 +538,18 @@ private static bool DefineResourceFetchingProperty(String propertyName, String r
...
-            // available strongly typed resource classes, we must return the 
-            // internal type.  For simplicity, we'll do that for internal strongly 
-            // typed resource classes as well.  Ideally we'd also like to check
-            // for interfaces like IList, but I don't know how to do that without
-            // special casing collection interfaces & ignoring serialization 
-            // interfaces or IDisposable.
-            while (!type.IsPublic)
-            {
-                type = type.BaseType;
-            }
```
nguerrera: It's not clear to me why this can be deleted. Won't it break existing uses?

nguerrera:
@rainersigwald I don't have a strong opinion. If the amount of code required is small (seems so here), I think it's reasonable to duplicate. Otherwise, I think we'd want to push the resx stuff down from System.Windows.Forms to base shared framework somewhere.

src/Tasks/system.design/stronglytypedresourcebuilder.cs
wjk: It had to be deleted because I'm no longer operating on types. I am dealing with the full names of types as strings, because MSBuild might not have access to the types (such as Bitmap) that the user is placing in their resx file.

wjk:
@nguerrera @rainersigwald In all honesty, this is still a very, very rough first draft. Any and all advice on how to improve this PR will be appreciated.

src/Tasks/system.design/stronglytypedresourcebuilder.cs
nguerrera: I see that. I think we would have to leave the prior behavior on full-framework msbuild with more conditional compilation. :(

rainersigwald: yeah, I'm afraid so :(

Tanya-Solyanik: @Dreddy-work

CZEMacLeod: Just my 2 cents, but I feel it would be really useful to have implementations of `System.Resources.IResourceReader` and `System.Resources.IResourceWriter` available to MSBuild and cross-platform (hopefully wrapped as a .netstandard package so it can fall through to the full framework or use a custom implementation etc.)
I believe that EF6 will need this as part of its porting.

wjk:
I think I've figured out the failing unit test. The problem is that the value [returned by the `IResourceReader` here](https://github.com/Microsoft/msbuild/pull/3819/files#diff-1d2aacce8c88cda3235c1659ea67e164R3772) is `null`. If the value is `null`, then there is by definition no way to get its type. Unfortunately, as I understand it [this resource](https://github.com/Microsoft/msbuild/blob/50dd97b58e94ee8e11e2be1aeafb71f9d0e25e88/src/Tasks.UnitTests/GenerateResource_Tests.cs#L2811-L2813) should not be null in the first place. I don't know what's going on here.

@rainersigwald @nguerrera Could you (or whomever else has an idea) please chime in with some thoughts? I don't see a way to fix this test, but I may be missing something obvious. Thanks!

src/Tasks/GenerateResource.cs
```diff
@@ -3054,8 +3052,26 @@ private void ReadResources(String filename, bool shouldUseSourcePath, String out
                             foreach (XElement dataElem in doc.Element("root").Elements("data"))
                             {
                                 string name = dataElem.Attribute("name").Value;
+                                string typeName = dataElem.Attribute("type")?.Value;
                                 string value = dataElem.Element("value").Value;
-                                AddResource(reader, name, value, filename);
+
+                                if (typeName != null)
+                                {
+                                    int pos = typeName.IndexOf(",", StringComparison.Ordinal);
```
iSazonov: Minor comment - I still wonder why you use "," _string_ if you could use ',' char.
Below too.

wjk: Because it doesn't build if I use the char. 😉

I found this out after I applied your two suggested fixes from earlier. I had to commit 8f448bc to fix the problem we inadvertently introduced.

iSazonov: Sorry, there is an overload for char `IndexOf(char ch)`.

wjk: But not if you specify the StringComparison value as well, which is what you suggested above. I admit I couldn't quite decide between keeping the char or using the StringComparison; I chose the StringComparison. I guess I could change it back, especially since the char being tested isn't an alphabetic character, so locale/case-sensitivity would not apply.

iSazonov: Sorry again, I think it is ok to ignore a locale/case-sensitivity and use `IndexOf(char ch)`.

src/Tasks/system.design/stronglytypedresourcebuilder.cs
```diff
@@ -106,8 +121,15 @@ internal static CodeCompileUnit Create(IDictionary resourceList, String baseName
             var resourceTypes = new Dictionary<String, ResourceData>(StringComparer.InvariantCultureIgnoreCase);
             foreach (DictionaryEntry de in resourceList)
             {
+                ResourceData data = de.Value as ResourceData;
+                if (data != null)
```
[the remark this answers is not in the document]
wjk: That won't work here, as we may reassign the `data` variable in the `#if` block just below.

wjk:
Could someone please help me out with the failing CI? I keep getting ~70 unit test failures, even after reverting the change I made in e45c285. All of these tests pass when I run them locally on my dev box. I am at my wits' end here, and I want to get this PR merged sooner rather than later so it can be ready for the Core 3.0 official preview.

rainersigwald:
I can reproduce some of the failures with `build.cmd -test` on my machine:

On full framework,

```
 Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.References.ReferencedAssemblySpecifiedUsingRelativePath
System.NullReferenceException : Object reference not set to an instance of an object.
   at Microsoft.Build.Tasks.ProcessResourceFiles.AddResource(ReaderInfo reader, String name, Object value, String inputFileName, Int32 lineNumber, Int32 linePosition, String customTypeName) in /_/src/Tasks/GenerateResource.cs:line 3774
   at Microsoft.Build.Tasks.ProcessResourceFiles.AddResource(ReaderInfo reader, String name, Object value, String inputFileName, String customTypeName) in /_/src/Tasks/GenerateResource.cs:line 3786
   at Microsoft.Build.Tasks.ProcessResourceFiles.ReadResources(ReaderInfo readerInfo, IResourceReader reader, String fileName) in /_/src/Tasks/GenerateResource.cs:line 3493
   at Microsoft.Build.Tasks.ProcessResourceFiles.ReadResources(String filename, Boolean shouldUseSourcePath, String outFileOrDir) in /_/src/Tasks/GenerateResource.cs:line 3044
   at Microsoft.Build.Tasks.ProcessResourceFiles.ProcessFile(String inFile, String outFileOrDir) in /_/src/Tasks/GenerateResource.cs:line 2625
   at Microsoft.Build.Tasks.ProcessResourceFiles.Run(TaskLoggingHelper log, ITaskItem[] assemblyFilesList, List`1 inputs, List`1 satelliteInputs, List`1 outputs, Boolean sourcePath, String language, String namespacename, String resourcesNamespace, String filename, String classname, Boolean publicClass, Boolean extractingResWFiles, String resWOutputDirectory) in /_/src/Tasks/GenerateResource.cs:line 2481
   at Microsoft.Build.Tasks.ProcessResourceFiles.Run(TaskLoggingHelper log, ITaskItem[] assemblyFilesList, List`1 inputs, List`1 satelliteInputs, List`1 outputs, Boolean sourcePath, String language, String namespacename, String resourcesNamespace, String filename, String classname, Boolean publicClass, Boolean extractingResWFiles, String resWOutputDirectory)
   at Microsoft.Build.Tasks.GenerateResource.Execute() in /_/src/Tasks/GenerateResource.cs:line 816
   at Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.References.ReferencedAssemblySpecifiedUsingRelativePath() in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 2837
Output:
resgen.exe /r:bin\debug\ClassLibrary20.dll /compile C:\Users\raines\AppData\Local\Temp\nci4bpf2.qsd\1ee678c1eead440488532fd43f334547\MyStrings.resx,C:\Users\raines\AppData\Local\Temp\nci4bpf2.qsd\1ee678c1eead440488532fd43f334547\MyStrings.resources
Creating a separate AppDomain because of resource "Image1" of type "ClassLibrary20.Class1, ClassLibrary20, version=1.0.0.0, Culture=neutral, PublicKeyToken=null" in "C:\Users\raines\AppData\Local\Temp\nci4bpf2.qsd\1ee678c1eead440488532fd43f334547\MyStrings.resx" on line 43.
Processing resource file "C:\Users\raines\AppData\Local\Temp\nci4bpf2.qsd\1ee678c1eead440488532fd43f334547\MyStrings.resx" into "C:\Users\raines\AppData\Local\Temp\nci4bpf2.qsd\1ee678c1eead440488532fd43f334547\MyStrings.resources".
```

On core:

```xml
      <test name="Microsoft.Build.UnitTests.WriteCodeFragment_Tests.InvalidLanguage" type="Microsoft.Build.UnitTests.WriteCodeFragment_Tests" method="InvalidLanguage" time="0.0014599" result="Fail">
        <traits>
          <trait name="Category" value="mono-osx-failing" />
        </traits>
        <failure exception-type="System.CodeDom.Compiler.CodeDomProvider+ConfigurationErrorsException">
          <message><![CDATA[System.CodeDom.Compiler.CodeDomProvider+ConfigurationErrorsException : There is no CodeDom provider defined for the language.]]></message>
          <stack-trace><![CDATA[   at System.CodeDom.Compiler.CodeDomProvider.GetCompilerInfo(String language)
   at System.CodeDom.Compiler.CodeDomProvider.CreateProvider(String language)
   at Microsoft.Build.Tasks.WriteCodeFragment.GenerateCode(String& extension) in /_/src/Tasks/WriteCodeFragment.cs:line 140
   at Microsoft.Build.Tasks.WriteCodeFragment.Execute() in /_/src/Tasks/WriteCodeFragment.cs:line 84
   at Microsoft.Build.UnitTests.WriteCodeFragment_Tests.InvalidLanguage() in S:\msbuild\src\Tasks.UnitTests\WriteCodeFragment_Tests.cs:line 34]]></stack-trace>
        </failure>
      </test>

      <test name="Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.STRWithResourcesNamespaceVB" type="Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations" method="STRWithResourcesNamespaceVB" time="0.0882194" result="Fail">
        <output><![CDATA[resgen.exe C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmpbedb2c14b77e4be3b605675e84a9533a.txt C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmpbedb2c14b77e4be3b605675e84a9533a.resources /str:VB,,,
Processing resource file "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmpbedb2c14b77e4be3b605675e84a9533a.txt" into "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmpbedb2c14b77e4be3b605675e84a9533a.resources".
Processing 4 resources from file "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmpbedb2c14b77e4be3b605675e84a9533a.txt".
Creating strongly typed resources class "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmpbedb2c14b77e4be3b605675e84a9533a.vb".
]]></output>
        <traits>
          <trait name="Category" value="mono-osx-failing" />
          <trait name="Category" value="mono-windows-failing" />
        </traits>
        <failure exception-type="Xunit.Sdk.TrueException">
          <message><![CDATA[Assert.True() Failure\r\nExpected: True\r\nActual:   False]]></message>
          <stack-trace><![CDATA[   at Microsoft.Build.UnitTests.GenerateResource_Tests.Utilities.AssertStateFileWasWritten(GenerateResource t) in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 3245
   at Microsoft.Build.UnitTests.GenerateResource_Tests.Utilities.STRNamespaceTestHelper(String strLanguage, String resourcesNamespace, String classNamespace, ITestOutputHelper output) in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 3631
   at Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.STRWithResourcesNamespaceVB() in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 1258]]></stack-trace>
        </failure>
      </test>

      <test name="Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.STRWithResourcesNamespaceAndSTRNamespaceCS" type="Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations" method="STRWithResourcesNamespaceAndSTRNamespaceCS" time="0.0226973" result="Fail">
        <output><![CDATA[resgen.exe C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmp0dc3042b713248e1b7681137c6c57fa7.txt C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmp0dc3042b713248e1b7681137c6c57fa7.resources /str:CSharp,MySTClassNamespace,,
Processing resource file "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmp0dc3042b713248e1b7681137c6c57fa7.txt" into "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmp0dc3042b713248e1b7681137c6c57fa7.resources".
Processing 4 resources from file "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmp0dc3042b713248e1b7681137c6c57fa7.txt".
Creating strongly typed resources class "C:\Users\raines\AppData\Local\Temp\ogfuf5ii.wzw\tmp0dc3042b713248e1b7681137c6c57fa7.cs".
]]></output>
        <traits>
          <trait name="Category" value="mono-osx-failing" />
          <trait name="Category" value="mono-windows-failing" />
        </traits>
        <failure exception-type="Xunit.Sdk.TrueException">
          <message><![CDATA[Assert.True() Failure\r\nExpected: True\r\nActual:   False]]></message>
          <stack-trace><![CDATA[   at Microsoft.Build.UnitTests.GenerateResource_Tests.Utilities.AssertStateFileWasWritten(GenerateResource t) in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 3245
   at Microsoft.Build.UnitTests.GenerateResource_Tests.Utilities.STRNamespaceTestHelper(String strLanguage, String resourcesNamespace, String classNamespace, ITestOutputHelper output) in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 3631
   at Microsoft.Build.UnitTests.GenerateResource_Tests.InProc.RequiredTransformations.STRWithResourcesNamespaceAndSTRNamespaceCS() in S:\msbuild\src\Tasks.UnitTests\GenerateResource_Tests.cs:line 1271]]></stack-trace>
        </failure>
      </test>
```

And others. How are you running the tests?

wjk:
@rainersigwald I just run them using the VS Test Explorer. I did this so I could use the debugger to investigate the cause of the failures. Is this an incorrect methodology?

wjk:
As it turns out, using the VS Test Explorer _was_ incorrect methodology. It runs the tests under the full-framework runtime, while the tests were failing only when run under the .NET Core runtime. Will have some fixes pushed up momentarily.

wjk:
@rainersigwald @nguerrera It's been over two weeks since I got the tests passing; I would appreciate PR review so we can get this merged soon. Thanks! 😄

src/Tasks/GenerateResource.cs
```diff
@@ -3444,18 +3460,23 @@ public static bool TryCreateCodeDomProvider(TaskLoggingHelper logger, string str
             {
                 provider = CodeDomProvider.CreateProvider(stronglyTypedLanguage);
             }
-#if FEATURE_SYSTEM_CONFIGURATION
```
nguerrera: Elsewhere you still have configuration #if'ed out. Does CodeDomProvider actually throw configuration exception on .NET Core? I think we should avoid adding a dependency on it if it's not needed.

src/Tasks/WriteCodeFragment.cs
```diff
@@ -139,16 +139,23 @@ private string GenerateCode(out string extension)
             {
                 provider = CodeDomProvider.CreateProvider(Language);
             }
+#if FEATURE_WINFORMS_RESX
```
nguerrera: (This is where it's still #if'ed out that I referenced in earlier comment)

src/Tasks/system.design/stronglytypedresourcebuilder.cs
```diff
@@ -296,6 +335,9 @@ internal static CodeCompileUnit Create(String resxFile, String baseName, String
             // keywords, etc.  So there's no point to duplicating the code above.
 
             return InternalCreate(resourceList, baseName, generatedCodeNamespace, resourcesNamespace, codeProvider, internalClass, out unmatchable);
+#else
+            throw new NotImplementedException();
```
nguerrera: This should probably have a better error message describing what isn't supported.

nguerrera:
In general, I still find it hard to follow the impact this will have on the full-framework case. Some work was done to retain the System.Type usage, but then in some places, there are still string comparison replacing type comparisons unconditionally in other places. I think I'd have an easier time reviewing if it were more obvious that this enables core without changing behavior on framework.

wjk:
@nguerrera In all honesty, the string comparison was *only* implemented because, at the time, `ResXResourceReader` and friends were not available on Core. If there was any other way, I would have used it.

However, with WinForms open-sourced and ported to Core, I was planning on adding a copy of them into the Tasks project (to break the dependency on the Forms DLL; we can always remove them again later), getting rid of the string-related code, and enabling `FEATURE_RESX_RESOURCE_READER` globally. Sound good? Thanks!

nguerrera:
I'm not sure because there are open questions about actively instantiating objects while reading resx. I have two things in mind: get away from that, but also don't change the .net framework case.  I'd have to look more closely at what ResxResourceReader does.

rainersigwald:
> I'd have to look more closely at what ResxResourceReader does.

This is high on my agenda for next week, for reasons closely related to this PR.

nguerrera:
> This is high on my agenda for next week, for reasons closely related to this PR.

👍  @wjk, I would hold off on doing any big changes to this PR while @rainersigwald is looking into this.

wjk:
@nguerrera @rainersigwald Any updates?

bergmeister:
This feature would really be useful, thanks to the PR creator, I hope this gets merged downstream to the 2.2 SDK as well.

bergmeister:
Any updates @nguerrera  @rainersigwald ?

src/Tasks/system.design/stronglytypedresourcebuilder.cs
danmosemsft: I don't know about MSBuild usage, but in CoreFX `NotImplementedException` is used when something is not implemented by design (such as, it needs to be overridden) and `PlatformNotSupportedException` when it's not relevant to the platform or not yet implemented there.

wjk:
This PR has since been superseded by #4420. Closing.

PR #3820, opened 2018-10-03
Static Graph: Add the ability to hook ProjectInstance creation to allow creation of Projects first

erikma:
Note name change from Project to ProjectInstance for previously existing public API. Can restore old name Project and make new addition something like 'UnevaluatedProject' if downstream consumers might be affected.

### DIFF
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -7,6 +7,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Text;
 using Microsoft.Build.BackEnd;
+using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.UnitTests;
 using Shouldly;
@@ -28,6 +29,45 @@ namespace Microsoft.Build.Graph.UnitTests
             }
         }
 
+        [Fact]
+        public void ConstructWithSingleNodeWithProjectInstanceFactory()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                TransientTestFile entryProject = CreateProject(env, 1);
+
+                bool factoryCalled = false;
+                var projectGraph = new ProjectGraph(
+                    entryProject.Path,
+                    ProjectCollection.GlobalProjectCollection,
+                    (projectPath, globalProperties, projectCollection) =>
+                    {
+                        factoryCalled = true;
+                        return ProjectGraph.DefaultProjectInstanceFactory(
+                            projectPath,
+                            globalProperties,
+                            projectCollection);
+                    });
+                projectGraph.ProjectNodes.Count.ShouldBe(1);
+                projectGraph.ProjectNodes.First().ProjectInstance.FullPath.ShouldBe(entryProject.Path);
+                factoryCalled.ShouldBeTrue();
+            }
+        }
+
+        [Fact(Skip="Disabling for now as most recent exp/net472 break exception throwing on graph API. Bug #3871")]
+        public void ConstructWithProjectInstanceFactory_FactoryReturnsNull_Throws()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                TransientTestFile entryProject = CreateProject(env, 1);
+
+                Should.Throw<InvalidOperationException>(() => new ProjectGraph(
+                    entryProject.Path,
+                    ProjectCollection.GlobalProjectCollection,
+                    (projectPath, globalProperties, projectCollection) => null));
+            }
+        }
+        
         /// <summary>
         ///   1
         ///  / \
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -402,13 +402,13 @@ namespace Microsoft.Build.Graph.UnitTests
                 var entryPointNode1 = projectGraph.EntryPointNodes.First();
                 var entryPointNode2 = projectGraph.EntryPointNodes.Last();
 
-                // The entrypoints should not be the same node, but should point to the same project
+                // The entry points should not be the same node, but should point to the same project
                 entryPointNode1.ShouldNotBe(entryPointNode2);
                 entryPointNode1.ProjectInstance.FullPath.ShouldBe(entryPointNode2.ProjectInstance.FullPath);
                 entryPointNode1.GlobalProperties["Platform"].ShouldBe("x86");
                 entryPointNode2.GlobalProperties["Platform"].ShouldBe("x64");
 
-                // The entrypoints should not not have the same project reference, but should point to the same project reference file
+                // The entry points should not have the same project reference, but should point to the same project reference file
                 entryPointNode1.ProjectReferences.Count.ShouldBe(1);
                 entryPointNode2.ProjectReferences.Count.ShouldBe(1);
                 entryPointNode1.ProjectReferences.First().ShouldNotBe(entryPointNode2.ProjectReferences.First());
@@ -441,13 +441,13 @@ namespace Microsoft.Build.Graph.UnitTests
                 var entryPointNode1 = projectGraph.EntryPointNodes.First();
                 var entryPointNode2 = projectGraph.EntryPointNodes.Last();
 
-                // The entrypoints should not be the same node, but should point to the same project
+                // The entry points should not be the same node, but should point to the same project
                 entryPointNode1.ShouldNotBe(entryPointNode2);
                 entryPointNode1.ProjectInstance.FullPath.ShouldBe(entryPointNode2.ProjectInstance.FullPath);
                 entryPointNode1.GlobalProperties["Platform"].ShouldBe("x86");
                 entryPointNode2.GlobalProperties["Platform"].ShouldBe("x64");
 
-                // The entrypoints should have the same project reference since it's platform-agnostic
+                // The entry points should have the same project reference since they're platform-agnostic
                 entryPointNode1.ProjectReferences.Count.ShouldBe(1);
                 entryPointNode2.ProjectReferences.Count.ShouldBe(1);
                 entryPointNode1.ProjectReferences.First().ShouldBe(entryPointNode2.ProjectReferences.First());
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Build.Graph
         /// <param name="entryProjectFile">The project file to use as the entry point in constructing the graph</param>
         /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
         public ProjectGraph(string entryProjectFile)
-            : this(new ProjectGraphEntryPoint(entryProjectFile).AsEnumerable(), ProjectCollection.GlobalProjectCollection)
+            : this(new ProjectGraphEntryPoint(entryProjectFile).AsEnumerable(), ProjectCollection.GlobalProjectCollection, null)
         {
         }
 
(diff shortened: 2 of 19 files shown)


src/Build/Graph/ProjectGraph.cs
```diff
@@ -73,6 +73,27 @@ public ProjectGraph(string entryProjectFile)
...
+        /// When false, only <see cref="Microsoft.Build.Execution.ProjectInstance"/> instances are exposed
+        /// from the graph. When true, the more memory and time expensive full <see cref="Microsoft.Build.Evaluation.Project"/>s
+        /// are also produced. Typically Projects are not needed for execution of a graph.
+        /// </param>
+        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
+        public ProjectGraph(
+            string entryProjectFile,
+            IDictionary<string, string> globalProperties,
+            ProjectCollection projectCollection,
+            bool exposeProjects)
```
dfederm: We probably want this as a Flags I'd guess? Maybe something like...
```
[Flags]
public enum ProjectGraphCreationSettings
{
  CreateProjectInstance = 0,
  CreateProject = 1,
}
```

src/Build/Graph/ProjectGraph.cs
```diff
@@ -203,31 +224,54 @@ public ProjectGraph(string entryProjectFile)
...
+            Project project = null;
+            ProjectInstance projectInstance;
+            if (exposeProject)
+            {
+                project = new Project(
+                    configurationMetadata.ProjectFullPath,
+                    globalProperties,
+                    configurationMetadata.ToolsVersion,
+                    projectCollection);
+                projectInstance = project.CreateProjectInstance(ProjectInstanceSettings.ImmutableWithFastItemLookup);
```
dfederm: I'm not super sure what `ImmutableWithFastItemLookup` does, but it seems like either we should use it in both cases, or we shouldn't use it at all (if it breaks the ability to actually build the project). MSBuild.exe doesn't use this so I suspect it breaks something.

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -27,10 +30,24 @@ public sealed class ProjectGraphNode
         /// </summary>
         public IReadOnlyCollection<ProjectGraphNode> ProjectReferences => _projectReferences;
 
+        /// <summary>
+        /// Gets a pre-evaluation project represented by this node in the graph.
```
dfederm: `Project` also has evaluated stuff on it though.

src/Build/Graph/ProjectGraph.cs
erikma: This setting is used in the internal build engine when flattening the Project to a ProjectInstance, for performance reasons. It has two effects in ProjectInstance.cs:

            var keepEvaluationCache = (settings & ProjectInstanceSettings.ImmutableWithFastItemLookup) == ProjectInstanceSettings.ImmutableWithFastItemLookup;

That flag leads to CreateEvaluatedIncludeSnapshotIfRequested() setting _itemsByEvaluatedInclude, which speeds up lookups in GetItemsByItemTypeAndEvaluatedInclude().

I'd prefer to stick with what is known to work in the internal build engine, but wouldn't mind having this as an additional flag in the proposed option flags noted above.

src/Build/Graph/ProjectGraph.cs
erikma: Given comment below seems like I'll need a flag related to ImmutableWithFastItemLookup anyway.

src/Build/Graph/ProjectGraph.cs
dfederm: My fear is if the flag basically makes it "parse-only, no target execution", which would be pretty broken in the general case.

The evaluation cache is probably (?) ok. The immutable thing is what I'm mostly afraid of. For example, it seems to imply to me that any property or item being set during target execution would just fail.

Someone more familiar with this setting might need to jump in and explain the details of how it effects target execution.

src/Build/Graph/ProjectGraph.cs
erikma: I don't like the useless surface area exposed by None. How about the default zero value be CreateProjectInstance, with the other flags as additional info? I'll send that in the next push.

dfederm: That's more of an Enum than a Flags then. Flags should always have a "None" (and sometimes an "All"). With enums though you'd need distinct, mutually exclusive values.

src/Build/Graph/ProjectGraph.cs
erikma: It'd be the normal case by default and opt-in to this behavior with a flag, which would be used in the internal build engine pretty much exclusively AFAICT.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -203,31 +224,54 @@ public ProjectGraph(string entryProjectFile)
...
         /// <summary>
-        /// Load a graph with root node at entryProjectFile
-        /// Maintain a queue of projects to be processed- each queue item is a key value pair of the project to be evaluated and its parent
-        /// Once the project has been evaluated, add a project reference to this evaluated target from the parent node
+        /// Load a graph with root node at entryProjectFile.
+        /// Maintain a queue of projects to be processed - each queue item is a key value pair of the project to be evaluated and its parent.
+        /// Once the project has been evaluated, add a project reference to this evaluated target from the parent node.
         /// </summary>
-        private void LoadGraph(string entryProjectFile, ProjectCollection projectCollection, PropertyDictionary<ProjectPropertyInstance> globalProperties)
+        private void LoadGraph(
```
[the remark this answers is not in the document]
erikma: #3823 covers the rename by itself

src/Build/Graph/ProjectGraph.cs
dfederm: If this *were* something we were building specifically for QuickBuild, it would not be in the MSBuild repo.

src/Build/Graph/ProjectGraph.cs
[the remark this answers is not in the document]
erikma: Hard to separate out now... the rename was automated but in the midst of other edits.

src/Build/Graph/ProjectGraph.cs
erikma: It is in the public MSBuild API but IIRC was prompted by performance needs of QuickBuild.

src/Build/Graph/ProjectGraph.cs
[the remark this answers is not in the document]
erikma: Rebased in current push as the topmost change

src/Build/Graph/ProjectGraph.cs
erikma: Iterated twice on the flags. Having ProjInstance as a separate option makes no sense, as when I did that and wrote UTs to test it the basic graph generation blew up on its need for ProjectInstance. Then pulled out the PI options and changed None to Default with comments about what defaults are.

src/Build/Graph/ProjectGraph.cs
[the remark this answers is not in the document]
erikma: Rebased to #3810 in next push

erikma:
@dotnet-bot test msbuild-pr

AndyGerlicher:
@erikma any idea why it's pulling in some extra commits?

dfederm:
I think he rebased on mater instead of exp/net472 :)

rainersigwald:
Yeah, looks like it was based on master then re-rebased back to net472. We must do some history manipulation before merge.

erikma:
Couldn't figure out a good way to rebase, so reverted the 2 master commits that made their way into history. And yes it was a mistaken rebase onto master before net472.

erikma:
Hmm. Downstream consumers need ProjectInstance anyway because of EvaluateCondition. EvalCond needs more work as well even on ProjInstance - in my backlog - since BuildPrediction is jumping hoops trying to duplicate work that MSBuild can do already but was never exposed publicly.

This API surface is as complicated as the MSBuild publi  API itself... as soon as you get into the choice of whether to expose Project (per the original one-pager which I had used for BuildPrediction initially) or ProjInstance, you end up in a conundrum, since both are really needed, conditionally.

In this PR I backed out pure 'conditionally' since there is no way to not specify ProjInstance, as ProjInstance stuff is needed in the graph API generation logic. I could re-surface that work, it's in this branch's history, but it looked difficult to get all the info needed to make just Project work either.

What I'm saying is that MSBuild has a duality here, and we have to embrace it, for good or ill.

2.

erikma:
@rainersigwald Should I abandon this PR? Is a change going into master to change to Project as noted above?

erikma:
Ugh, rebased master again instead of net472. New push pending to rebase properly and remove master commits from history

src/Build/Graph/ProjectGraph.cs
```diff
@@ -303,17 +334,63 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+        /// <see cref="ProjectInstanceSettings.Immutable"/>, the resulting ProjectGraph
+        /// nodes might not be buildable.
+        /// </remarks>
+        public delegate ProjectInstance ProjectInstanceFactoryFunc(
+            string projectPath,
+            Dictionary<string, string> globalProperties,
+            string toolsVersion,
+            ProjectCollection projectCollection);
+
+        internal static ProjectInstance DefaultProjectInstanceFactory(
```
cdmihai: Nit: Should this be public? In case callers want to do some stuff but still call the default factory.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -303,17 +334,63 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+        /// The returned ProjectInstance will be stored and provided with the ProjectGraph.
+        /// If this callback chooses to generate an immutable ProjectInstance, e.g. by
+        /// using <see cref="Project.CreateProjectInstance()"/> with the flag
+        /// <see cref="ProjectInstanceSettings.Immutable"/>, the resulting ProjectGraph
+        /// nodes might not be buildable.
+        /// </remarks>
+        public delegate ProjectInstance ProjectInstanceFactoryFunc(
+            string projectPath,
+            Dictionary<string, string> globalProperties,
+            string toolsVersion,
```
dfederm: Tools version is going away, so we should not include it. It's not part of the public API at all for `ProjectGraph`

src/Build/Graph/ProjectGraph.cs
AndyGerlicher: I would say no, nobody should need this.

src/Build/Graph/ProjectGraph.cs
AndyGerlicher: Also is there a reason to define a delegate instead of using `Func<string, Dictionary<string, string>, ProjectInstance>`? It does seem kind of long.

erikma: @AndyGerlicher I started with Func<> but the param documentation got long, so moved to a delegate declaration to expand param definitions. I recommend staying with this format as squashing it back is pretty ugly.

erikma: @dfederm Set to null? I was strictly copying the original code which sets that value to                 configurationMetadata.ToolsVersion

dfederm: `ConfigurationMetadata.ToolsVersion` is hardcoded to `MSBuildConstants.CurrentToolsVersion` in this case though: https://github.com/Microsoft/msbuild/blob/exp/net472/src/Build/BackEnd/Shared/ConfigurationMetadata.cs#L49

Null probably would work too I'd think.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -28,6 +29,45 @@ public void ConstructWithSingleNode()
...
+                            globalProperties,
+                            projectCollection);
+                    });
+                projectGraph.ProjectNodes.Count.ShouldBe(1);
+                projectGraph.ProjectNodes.First().ProjectInstance.FullPath.ShouldBe(entryProject.Path);
+                factoryCalled.ShouldBeTrue();
+            }
+        }
+
+        [Fact(Skip="Disabling for now as most recent exp/net472 break exception throwing on graph API")]
```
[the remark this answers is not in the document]
rainersigwald: which is #3871, thanks for filing @yash256

erikma:
@rainersigwald Removed most comment spelling and grammar fixes and R# stuff. Shall I issue a new PR off of net472 to clear the history, or is this sufficient?

AndyGerlicher:
No need to do a new PR that won't help. Just need to rebase correctly so we can merge the PR (not squash).

erikma:
Adjusted history to create an API update commit

Issue #3795, opened 2018-09-26
Support multiple entry points in the ProjectGraph


dfederm:
For some programmatic API scenarios, multiple entry points would be required.

Command-line still only has the ability to provide one file though, so this is purely for consumers of the ProjectGraph API.

Example scenario: "Open folder" (no sln), or constructing a graph for an entire repo which doesn't have a root-level traversal project.

A possible alternative would be to "fake" a project which simply has `<ProjectReference Include="**\*.*proj" />` as the entry point, but allowing multiple entry points instead is more flexible.

rainersigwald:
Counterpoint: MSBuild.exe has never had this, and VS mostly hasn't either ("open folder" is new and doesn't really work the same way). I think this should be cut at least for v1.

PR #3827, opened 2018-10-04
Add support for multiple entry points in the ProjectGraph

dfederm:
Closes: #3795

### DIFF
--- a/ref/Microsoft.Build/net/Microsoft.Build.cs
+++ b/ref/Microsoft.Build/net/Microsoft.Build.cs
@@ -1409,12 +1409,29 @@ namespace Microsoft.Build.Graph
 {
     public sealed partial class ProjectGraph
     {
+        public ProjectGraph(Microsoft.Build.Graph.ProjectGraphEntryPoint entryPoint) { }
+        public ProjectGraph(Microsoft.Build.Graph.ProjectGraphEntryPoint entryPoint, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> entryPoints) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> entryPoints, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles, System.Collections.Generic.IDictionary<string, string> globalProperties) { }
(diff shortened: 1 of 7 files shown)


src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -302,6 +301,27 @@ public void ConstructWithInvalidProperties()
             }
         }
 
+        [Fact]
+        public void ConstructWithMultipleEntryPoints()
```
rainersigwald: I'd also like to see same-project-different-properties as a test case.

src/Build/Graph/ProjectGraphEntryPoint.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            foreach (string entryProjectFile in entryProjectFiles)
+            {
+                yield return new ProjectGraphEntryPoint(entryProjectFile, globalProperties);
+            }
+        }
+
+        internal IEnumerable<ProjectGraphEntryPoint> ToEnumerable()
+        {
+            yield return this;
```
rainersigwald: This is a fun trick. Is it standard? I feel like I'd prefer the name `AsEnumerable` to emphasize that it's still a single item.

ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -1409,13 +1409,30 @@ namespace Microsoft.Build.Graph
...
+        public ProjectGraph(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> entryPoints) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> entryPoints, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles, System.Collections.Generic.IDictionary<string, string> globalProperties) { }
+        public ProjectGraph(System.Collections.Generic.IEnumerable<string> entryProjectFiles, System.Collections.Generic.IDictionary<string, string> globalProperties, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
         public ProjectGraph(string entryProjectFile) { }
+        public ProjectGraph(string entryProjectFile, Microsoft.Build.Evaluation.ProjectCollection projectCollection) { }
```
rainersigwald: Do we need this one, since you could always pass `null` as the properties dict?

dfederm: You could pass `null`, but really we don't "need" any of these ctors except the single one that actually does work. I figure all these are just for convenience.

It also is less clear on the calling side (unless you use named params) to just have:
```
new ProjectGraph(file, null, projectCollection);
```

"What's 'null' here?" I can see code reviewers asking.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
dfederm: Oooh, and a test with `GlobalPropertiesToRemove` causing convergence, ie the "agnostic project" scenario.

src/Build/Graph/ProjectGraphEntryPoint.cs
dfederm: No idea if it's standard :). I assume it's better than wrapping it in an array, but I don't actually know that.

PR #3828, opened 2018-10-04
Update ToolsVersion to `Current` (on master)

AndyGerlicher:
Replaces #3779  (and also #3821 which has the wrong base)
See #3778


rainersigwald:
My apologies, I horked your test run trying to rerun it.

PR #3831, opened 2018-10-05
Don't call registry on non-Windows platforms.

mfilippov:


rainersigwald:
Can you elaborate on what problems you ran into here? I suspect this is probably ok (at least for now) since we don't have MOTW detection on Mac at the moment, but I'm not _sure_ it's the right thing, and would like more details.

mfilippov:
@rainersigwald I catch this problem on macOS when try to call resgen target. I think COM call in static ctor it is not a good idea. In other places of this class Native calls wrapped by platform check. Sorry for formatting changes I revert it.

rainersigwald:
On the Mono runtime, or .NET Core?

mfilippov:
On mono.

radical:
Um this is what we have on mono - https://github.com/mono/msbuild/commit/fdf84d1ea9152b9e222ac3d7800c3d7d5c6f7368

rainersigwald:
@radical I think I prefer this approach since it's a smaller change, aligning better with the `resgen.exe` code this is copypasta of. Would you be willing to take this and back out yours?

rainersigwald:
> Though I think the commit message with the stack traces would be useful here.

Agreed; I'll pull that in when I merge this.

> Maybe add a comment for `allowMOTW` explaining it (allow/check dangerous resources?) or rename the field (self documenting :p)

This code originated in a different, closed-source internal project, and I tried to make minimal changes to it in case we need to pull a patch forward through that and up to the present. So I agree with this criticism but won't implement it, to keep the code like the internal stuff.

rainersigwald:
Thanks, @mfilippov.

PR #3841, opened 2018-10-08
Update the static graph spec based on internal architecture review

dfederm:
Mostly this is just reorganization and clarification based on the recent internal architecture review.

Notably, graph builds and isolated builds are described separately since they're really two separate (but still related) features. It also just helps for clarity to describe graph builds in full first before introducing the complications of isolated builds.

### DIFF
--- a/documentation/specs/static-graph.md
+++ b/documentation/specs/static-graph.md
@@ -1,20 +1,18 @@
 - [Static Graph](#static-graph)
   - [Overview](#overview)
     - [Motivations](#motivations)
-    - [M1 - MVP](#m1---mvp)
-    - [M1 non-goals](#m1-non-goals)
-    - [M2 - Single Project Enforcement](#m2---single-project-enforcement)
-    - [M3 - Dependency Tracking](#m3---dependency-tracking)
-    - [M4 - Ready for Caching/Distribution](#m4---ready-for-cachingdistribution)
-    - [M5 - Performance](#m5---performance)
   - [Project Graph](#project-graph)
+    - [Build dimensions](#build-dimensions)
     - [Building a project graph](#building-a-project-graph)
+    - [Inferring which targets to run for a project within the graph](#inferring-which-targets-to-run-for-a-project-within-the-graph)
+    - [Underspecified graphs](#underspecified-graphs)
     - [Public API](#public-api)
-  - [Inferring which targets to run for a project within the graph](#inferring-which-targets-to-run-for-a-project-within-the-graph)
-  - [Building a single project](#building-a-single-project)
-    - [Entry targets](#entry-targets)
-    - [Caching](#caching)
+  - [Isolated builds](#isolated-builds)
+    - [Isolated graph builds](#isolated-graph-builds)
+    - [Single project isolated builds](#single-project-builds)
   - [Distribution](#distribution)
+    - [Serialization](#serialization)
+    - [Deserialization](#deserialization)
   - [I/O Tracking](#io-tracking)
 
 # Static Graph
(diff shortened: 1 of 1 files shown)


documentation/specs/static-graph.md
```diff
@@ -1,20 +1,18 @@
 - [Static Graph](#static-graph)
   - [Overview](#overview)
     - [Motivations](#motivations)
-    - [M1 - MVP](#m1---mvp)
```
dfederm: I removed the timelines and such since they ended up not really mapping well to the actual order we were working on things and the GitHub projects were a much better way to track and plan.

documentation/specs/static-graph.md
```diff
@@ -203,20 +119,17 @@ The default target (represented in this spec's pseudo protocol representation as
 
 `Clean -> GetTargetFrameworks, Clean`
 
-`Rebuild -> Clean, Build`
+`Rebuild -> GetTargetFrameworks, Clean, <default>, GetNativeManifest, GetCopyToOutputDirectoryItems`
 
-`Rebuild` maps to two referenced protocols. `Clean` and `Build` get substituted with the contents of the `Clean` and `Build` protocols.
+`Rebuild` actually calls `Clean` and `Build`, which in turn uses the concatenation of the `Clean` and `Build` mappings. `GetTargetFrameworks` is repeates so gets deduped..
```
dfederm: "repeated"

documentation/specs/static-graph.md
```diff
@@ -227,34 +140,107 @@ We'll represent the project reference protocols as items in MSBuild. The target
...
-- Option 2 (better): Serialize the export target state
-  - Best for performance, but requires a serialization protocol and a way to specify how to import the state
-  - Possibly same mechanism as the persistent cross-build cache described above.
+To support distribution for isolated builds (eg. for the MS internal build engine), we need a solution for a project and a dependency building on different machines. To facilitate this, we need to serialize the build result cache to disk on the source machine which built the project reference, and deserialize it on the destination machine which is about to build the project referencing the project reference.
+
+In fact, this same mechanism can be used for non-distributed builds as well, for example for incremental builds. The caller would still need to ensure that the build result cache was valid (ie. the project reference project was not changed since the build result cache for it was produced), but the benefit would be that only the current project being built would need to be evaluated and only targets on that project would need to execute.
+
+### Serialization
```
dfederm: I'll be working on the Serialization, Deserialization, and I/O tracking sections this week. I wanted to get this update out of the way since it doesn't involve any behavioral changes to the spec, just some reorganization and clarification.

documentation/specs/static-graph.md
```diff
@@ -227,34 +140,107 @@ We'll represent the project reference protocols as items in MSBuild. The target
...
+    {
+        public class ProjectGraph
+        {
+            // Creates a graph starting at the given project file.
+            public ProjectGraph(string projectFile) { }
+            public ProjectGraph(string entryProjectFile, IDictionary<string, string> globalProperties) { }
+
+            // Creates a graph starting at the given project files.
+            public ProjectGraph(IEnumerable<string> projectFiles) { }
+            public ProjectGraph(IEnumerable<string> entryProjectFiles, IDictionary<string, string> globalProperties) { }
```
cdmihai: Does this mean that the provided global properties are repeated for each entry point file?

documentation/specs/static-graph.md
```diff
@@ -227,34 +140,107 @@ We'll represent the project reference protocols as items in MSBuild. The target
...
+### Isolated graph builds
+When building a graph in insolated mode, the graph is used to traverse and build the projects in the right order, but each individual project is built in isolation. The build result cache will just be in memory exactly as it is today, but on cache miss it will error. This enforces that both the graph and target mappings are complete and correct.
+
+Furthermore, running in this mode enforces that each (project, global properties) pair is executed only once and must execute all targets needed by all projects which reference that node. This gives it a concrete start and end time, which leads to some perf optimizations, like garbage collecting all project state (except the build results) once it finishes building. This can greatly reduce the memory overhead for large builds.
```
cdmihai: > and must execute all targets needed by all projects which reference that node

This also happens in non-isolated graph builds right? Each node is visited once.

dfederm: Each node is visited once, but non-isolated has the flexibility to just fallback to normal scheduler behavior for stuff that isn't properly expressed. So for isolated, this is a "must" while for non-isolated I guess you can think about it as a "should".

documentation/specs/static-graph.md
```diff
@@ -227,34 +141,109 @@ We'll represent the project reference protocols as items in MSBuild. The target
...
 
 Because referenced projects and their entry targets are guaranteed to be in the cache, they will not build again. Therefore we do not need to set `/p:BuildProjectReferences=false` or any other gesture that tells SDKs to not do recursive operations.
 
-Initially this result cache will just be in-memory, similar to the existing MSBuild target result cache, but eventually could persist across builds so that an incremental build of a particular project in the project graph would not require re-evaluation or any target execution in that project's dependencies. This may have significant performance improvements for incremental builds.
+**Open Issue:** With this approach initial targets will also get included due to how the `RequestBuilder` composes entry targets and how `TargetBuilder` reports build results. Is this an issue? Is it just a perf optimization to take them out from the cache? It seems like target mapping incompleteness at the very least, since if the project changes its initial targets, the cache hit from the referencing project would suddenly be a miss.
```
rainersigwald: This doesn't seem that worrisome but if we ever ship with `InitialTargets` in the cache, we'd have to keep it that way forever. I think that's fine.

documentation/specs/static-graph.md
```diff
@@ -227,34 +141,109 @@ We'll represent the project reference protocols as items in MSBuild. The target
...
+
+**Open Issue:** How do we differentiate between targets called on the outer build project and targets called on the inner build projects? For multitargeting builds, a project will actually reach into the inner build, but the graph expresses a dependency on the outer build.
```
rainersigwald: It could be as simple as "impose a new restriction on outer builds that they have forwarding targets for each inner build target". But that might be a high burden. Could also potentially have a binary outer/inner distinction in the targets (but I don't know how to express that).

documentation/specs/static-graph.md
```diff
@@ -227,34 +141,109 @@ We'll represent the project reference protocols as items in MSBuild. The target
...
+
+### Isolated graph builds
+When building a graph in insolated mode, the graph is used to traverse and build the projects in the right order, but each individual project is built in isolation. The build result cache will just be in memory exactly as it is today, but on cache miss it will error. This enforces that both the graph and target mappings are complete and correct.
```
rainersigwald: typo: `insolated`

documentation/specs/static-graph.md
dfederm: What I'm trying to explain here is that because we only cache the entry targets, none of the inner build targets would be available to other projects in isolated mode. So although Build with no global props would be in the cache, Build with a specific target framework would not. Well, it would be in the cache, but the isolated build logic would not allow it.

So it seems like for isolated builds, we're going to have to special-case multitargeting regardless, either by splitting the graph into distinct nodes per TFM or in the isolated build cache lookup case.

Issue #2056, opened 2017-05-04
Choose/When/Otherwise in Targets


dasMulli:
`<Choose>`/`<When>`/`<Otherwise>` is nice features that allows replacing complex conditions with what is effectively a switch-case-default but can only be used on the project level and not within targets.

Can this be changed to work within targets?
This would theoretically conflict with tasks being named `Choose`, are there any ways to deal with this? - like (and I feel bad suggesting this) optional xml namespace for `<MSBuild:Choose>` and prefer a task named `Choose` if defined.

dasMulli:
@dotnet/msbuild is this feasible at all?
I so, I'd hope to see this in 16.0. If not, go ahead and close it as "won't implement".

rainersigwald:
I don't have a strong objection to this, other than the conflict-with-task one. But it's not very high on my priority list, personally. Anybody else feel more strongly about it?

cdmihai:
Implementation wise it will be interesting, since the current design accepts only tasks within targets (properties and items in targets are actually implemented as tasks). But shouldn't be too hard, probably involves implementing a Choose [IntrinsicTask ](https://github.com/Microsoft/msbuild/blob/master/src/Deprecated/Engine/Tasks/IntrinsicTask.cs).

dasMulli:
Implementing a choose implicit task seems to be promising, I got a prototype at
https://github.com/dasMulli/msbuild/tree/feature/choose-in-targets ([Compare](https://github.com/Microsoft/msbuild/compare/master...dasMulli:feature/choose-in-targets))

If there is interest in taking this as a community contribution I could invest more work in it.

cdmihai:
Took a quick look over it and looks good! Feel free to open it up as a WIP commit to get more feedback.

My take on it:
- not sure whether to allow task executions, or to restrict it to only what's allowed in the choose elements outside of targets.
- add more tests. To find the relevant tests to duplicate (or transform them into `[Theory]`) you can search for `<choose` in all the `*_Tests.cs` files. But do send the WIP PR before you do this.

rainersigwald:
Team triage: #3851 is a solid-looking implementation of this, but we currently think we should wait for a major release to add this type of language feature. If we want to take this for 17.0, it should definitely start with reviving #3851.

voroninp:
Hello there ;-), what's the curent status of this issue?

PR #3851, opened 2018-10-11
WIP: Intrinsic choose task

dasMulli:
WIP / Prototype for enabling `<choose>` in targets.

See #2056 

cc @cdmihai @rainersigwald


dasMulli:
Current status: "it works on my machine". I couldn't find many tests that explicitly test the behaviour of choose/when/otherwise. most are construction tests.

rainersigwald:
Thank you! Sorry we let this sit without comment for so long. At the moment we think it's not worth adding this in a 16.x point release. I added some notes to the bug to consider this again for the next major version, and to do so by reviving this PR.

Thanks again!

PR #3854, opened 2018-10-11
Fail the build on target cache misses during isolated project builds

### LINKED ISSUE #3686: Fail the build on target cache misses

cdmihai:
Resolves #3686.

Finding the best spot to put the check was a bit interesting. The alternative I thought the most was to put the check in the cache itself, and then have the cache users specify whether they want enforcement or not for each cache lookup. Since performing the lookup and acting upon misses needs a lot of scheduler internal data, I decided to put everything in the Scheduler.

Another issue is how to actually stop the build from the Scheduler. The current solution returns a failed BuildResult to stop the build and then issues a separate error to have something appear to loggers. I can't find a good spot to have the failed build result automagically generate a build error in some of the more normal places (somewhere in ProjectBuilder / TargetBuilder / TaskBuilder so the error originates from within a project, like all other msbuild errors).

### DIFF
--- a/src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
+++ b/src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
@@ -1,0 +1,143 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Execution;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Shared;
+using Microsoft.Build.UnitTests;
+using Shouldly;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Build.Graph.UnitTests
+{
+    public class IsolateProjectsTests
+    {
+        private readonly string _project = @"
+                <Project DefaultTargets='BuildSelf'>
+                    <Target Name='BuildReference'>
+                        <MSBuild Projects='{0}' Targets='ReferenceTarget'/>
+                    </Target>
+
+                    <Target Name='BuildSelf'>
+                        <MSBuild Projects='$(MSBuildThisFile)' Targets='SelfTarget'/>
+                    </Target>
+
+                    <Target Name='SelfTarget'>
+                    </Target>
+                </Project>";
+
+        private readonly string _reference = @"
+                <Project>
+                    <Target Name='ReferenceTarget'>
+                    </Target>
+                </Project>";
+
+        private readonly ITestOutputHelper _testOutput;
+
+        public IsolateProjectsTests(ITestOutputHelper testOutput)
+        {
+            _testOutput = testOutput;
+        }
+
+        [Theory]
+        [InlineData(BuildResultCode.Success, new string[] {})]
+        [InlineData(BuildResultCode.Success, new[] {"BuildSelf"})]
+        public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedBuildResult, string[] targets)
+        {
+            AssertBuild(targets,
+                (result, logger) =>
+                {
+                    result.OverallResult.ShouldBe(BuildResultCode.Success);
+
+                    logger.Errors.ShouldBeEmpty();
+                });
+        }
+
+        [Fact]
+        public void CacheEnforcementShouldFailWhenReferenceWasNotPreviouslyBuilt()
+        {
+            AssertBuild(new []{"BuildReference"},
+                (result, logger) =>
+                {
+                    result.OverallResult.ShouldBe(BuildResultCode.Failure);
+
+                    logger.ErrorCount.ShouldBe(1);
+
+                    logger.Errors.First().Message.ShouldStartWith("MSB4252:");
+                });
+        }
+
+        [Fact]
+        public void CacheEnforcementShouldAcceptPreviouslyBuiltReferences()
+        {
+            AssertBuild(new []{"BuildReference"},
+                (result, logger) =>
+                {
+                    result.OverallResult.ShouldBe(BuildResultCode.Success);
+
+                    logger.Errors.ShouldBeEmpty();
+                },
+                buildReference: true);
+        }
+
+        private void AssertBuild(string[] targets, Action<BuildResult, MockLogger> assert, bool buildReference = false)
+        {
+            using (var env = TestEnvironment.Create())
+            using (var buildManager = new BuildManager())
+            {
+                var projectFile = env.CreateFile().Path;
+                var referenceFile = env.CreateFile().Path;
+
+                File.WriteAllText(projectFile, string.Format(_project, referenceFile));
+                File.WriteAllText(referenceFile, _reference);
+
+                var logger = new MockLogger(_testOutput);
+
+                var buildParameters = new BuildParameters
+                {
+                    IsolateProjects = true,
+                    Loggers = new ILogger[] {logger},
+                    EnableNodeReuse = false,
+                    DisableInProcNode = true
+                };
+
+                var rootRequest = new BuildRequestData(
+                    projectFile,
+                    new Dictionary<string, string>(),
+                    MSBuildConstants.CurrentToolsVersion,
+                    targets,
+                    null);
+
+                try
+                {
+                    buildManager.BeginBuild(buildParameters);
+
+                    if (buildReference)
+                    {
+                        buildManager.BuildRequest(
+                            new BuildRequestData(
+                                referenceFile,
+                                new Dictionary<string, string>(),
+                                MSBuildConstants.CurrentToolsVersion,
+                                new[] {"ReferenceTarget"},
+                                null))
+                            .OverallResult.ShouldBe(BuildResultCode.Success);
+                    }
+
+                    var result = buildManager.BuildRequest(rootRequest);
+
+                    assert(result, logger);
+                }
+                finally
+                {
+                    buildManager.EndBuild();
+                }
+            }
+        }
+    }
+}
--- a/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
+++ b/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
@@ -517,6 +517,12 @@ namespace Microsoft.Build.BackEnd
                                     {
                                         // We have a result, give it back to this request.
                                         currentEntry.ReportResult(cacheResponse.Results);
+
+                                        TraceEngine(
+                                            "Request {0} (node request {1}) with targets ({2}) satisfied from cache",
+                                            request.GlobalRequestId,
+                                            request.NodeRequestId,
+                                            string.Join(";", request.Targets));
                                     }
                                     else
                                     {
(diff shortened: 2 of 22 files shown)


src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1582,7 +1582,7 @@ private void HandleRequestBlockedByNewRequests(SchedulableRequest parentRequest,
                         abortRequestBatch = true;
                     }
                 }
-                else
+                else if(CacheMissEnforcement(nodeForResults, request, responses))
```
dfederm: Style nit: space between `if` and `(`

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1781,6 +1786,66 @@ private ScheduleResponse TrySatisfyRequestFromCache(int nodeForResults, BuildReq
             return null;
         }
 
+        private bool CacheMissEnforcement(int nodeForResults, BuildRequest request, List<ScheduleResponse> responses)
```
dfederm: Naming nit: Usually method names should be verbs.

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1781,6 +1786,66 @@ private ScheduleResponse TrySatisfyRequestFromCache(int nodeForResults, BuildReq
...
+            if (!request.IsRootRequest && isIsolatedBuild)
+            {
+                var configCache = (IConfigCache)_componentHost.GetComponent(BuildComponentType.ConfigCache);
+                var requestConfig = configCache[request.ConfigurationId];
+                
+                // Need the parent request. But the parent / child relationship is sometimes formed after the cache is queried, so get creative
+                var parentRequest = _schedulingData.BlockedRequests.FirstOrDefault(r => r.BuildRequest.GlobalRequestId == request.ParentGlobalRequestId);
+                if (parentRequest == null)
+                {
+                    // the parent might still be in executing requests because the scheduler might not have had a chance to mark it as blocked yet
```
dfederm: Based on the comments, do we need to worry about thread-safety here? And/or a race condition between `BlockedRequests` and `ExecutingRequests`?

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1781,6 +1786,66 @@ private ScheduleResponse TrySatisfyRequestFromCache(int nodeForResults, BuildReq
...
+                    parentRequest = _schedulingData.ExecutingRequests.FirstOrDefault(r => r.BuildRequest.GlobalRequestId == request.ParentGlobalRequestId);
+                }
+
+                ErrorUtilities.VerifyThrowInternalNull(parentRequest, nameof(parentRequest));
+                ErrorUtilities.VerifyThrow(configCache.HasConfiguration(parentRequest.BuildRequest.ConfigurationId), "All non root requests should have a parent with a loaded configuration");
+
+                var parentConfig = configCache[parentRequest.BuildRequest.ConfigurationId];
+
+                // allow self references (project calling the msbuild task on itself, potentially with different global properties)
```
dfederm: Something I just realized about self-references, what if two different nodes for the same project call into each other? IE, debug node calling into the retail node? Maybe that's an edge case that we don't particularly care about though since it doesn't make much sense anyway...

cdmihai: That should be legal under isolated builds right?

dfederm: The idea was to support multitargeting builds, so one project calling into itself. But now that we support nodes with different global props (and multiple entry points), we could have two separate nodes with different global properties, like debug and retail, which in cloudbuild would build on different machines, so semantically we should not allow them to call into each other willy nilly.

However in practice, this seems pretty edge-casey and not breaking as much as not being "correct".

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
cdmihai: I don't think so. The comment is trying to say that the scheduler sometimes checks the cache before it constructs the parent - child relationship (these are sequential steps). And in addition, when the scheduler checks the cache, there cannot be a concurrent insertion with the same data that would lead to a sort of read-before-write race.

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
cdmihai: Ah, I understand now. Detecting this situation is tricky, because there's no way that I can see to differentiate between two static graph nodes of the same project with different global props and the multi targeting scenario (where you also have the same project calling into itself with different global properties). The difference is that the former would specify the relationship in ProjectReference, but if it misses to do so, there's no way to detect it unless we do more aggressive static analysis over the msbuild task patterns to only whitelist the multi targeting pattern.

Though with our current approach it might still work out. If A -> A' (A' is the same project but different global props), and if they get assigned to different machines, executing A' would lead to its result cache getting serialized and sent over to the machine where A builds. So A would get a cache with the results of A' in it.

During a non distributed in-msbuild graph build you'd get the same result, the cached results for A' would be there when A needs to build.

The problem arises when A does not specify A' in its ProjectReference item. Because the edge is not there, A' does not get built before A. So under isolated builds, it should theoretically fail whereas with this implementation it will allow A to build A' as part of its build.

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -0,0 +1,143 @@
...
+        [InlineData(BuildResultCode.Success, new string[] {})]
+        [InlineData(BuildResultCode.Success, new[] {"BuildSelf"})]
+        public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedBuildResult, string[] targets)
+        {
+            AssertBuild(targets,
+                (result, logger) =>
+                {
+                    result.OverallResult.ShouldBe(BuildResultCode.Success);
+
+                    logger.ErrorCount.ShouldBe(0);
```
rainersigwald: I wonder if `logger.Errors.ShouldBeEmpty()` would be more diagnosable when it fails. As is it'd say "1 error, expected 0", and the natural question is "so what was the error?"

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -0,0 +1,143 @@
...
+                {
+                    result.OverallResult.ShouldBe(BuildResultCode.Failure);
+
+                    logger.ErrorCount.ShouldBe(1);
+
+                    logger.Errors.First().Message.ShouldStartWith("MSB4252: Failed isolated build constraint: References should be built before projects referencing them.");
+                });
+        }
+
+        [Fact(Skip = "test")]
```
rainersigwald: ?

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -0,0 +1,143 @@
...
+            using (var env = TestEnvironment.Create())
+            using (var buildManager = new BuildManager())
+            {
+                var projectFile = env.CreateFile().Path;
+                var referenceFile = env.CreateFile().Path;
+
+                File.WriteAllText(projectFile, string.Format(_project, referenceFile));
+                File.WriteAllText(referenceFile, _reference);
+
+                var logger = new MockLogger();
```
rainersigwald: pass the xunit output helper to the logger

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -0,0 +1,143 @@
...
+                                new Dictionary<string, string>(),
+                                MSBuildConstants.CurrentToolsVersion,
+                                new[] {"ReferenceTarget"},
+                                null))
+                            .OverallResult.ShouldBe(BuildResultCode.Success);
+                    }
+
+                    var result = buildManager.BuildRequest(rootRequest);
+
+                    //_testOutput.WriteLine(logger.FullLog);
```
rainersigwald: nit: delete

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
```diff
@@ -517,6 +517,12 @@ public void ReportConfigurationResponse(BuildRequestConfigurationResponse respon
                                     {
                                         // We have a result, give it back to this request.
                                         currentEntry.ReportResult(cacheResponse.Results);
+
+                                        TraceEngine(
+                                            "Request {0} (node request {1}) with targets ({2}) satisfied from cache",
+                                            request.GlobalRequestId,
+                                            request.NodeRequestId,
+                                            string.Join(",", request.Targets));
```
rainersigwald: nit: don't we usually join targets with `;`?

src/Build/Resources/Strings.resx
```diff
@@ -1714,13 +1714,19 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
       request a target to build itself (perhaps via a chain of other targets)
     </comment>
   </data>
+  <data name="CacheMissesNotAllowedInIsolatedGraphBuilds" UESanitized="false" Visibility="Public">
+    <value>MSB4252: Failed isolated build constraint: References should be built before projects referencing them. Project "{0}" has a reference to "{1}" ({2} target(s)) but the reference has not been previously built. In graph based builds this probably means that the reference is not explicitly specified as a project reference in "{0}".</value>
```
rainersigwald: `project reference` -> `ProjectReference`, and add a loc note to not translate it.

rainersigwald: Is this thrown when a target isn't cached, or just the whole project? sounds like the whole project; we may want to have two separate errors.

rainersigwald: Drop the `Failed isolated build constraint: References should be built before projects referencing them. `, I think, just get to the meat of it.

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
rainersigwald: The only thing I can think of to actually solve this would be to treat the global property `TargetFramework` specially here, which . . . ain't great.

src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1781,6 +1786,74 @@ private ScheduleResponse TrySatisfyRequestFromCache(int nodeForResults, BuildReq
             return null;
         }
 
+        private bool EnforceCacheMissInIsolatedProjectBuilds(int nodeForResults, BuildRequest request, List<ScheduleResponse> responses)
```
rainersigwald: The semantics of this confused me because I interpreted it backward. Maybe rename to `CacheMissIsAllowed()`?

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
cdmihai: Testing CI failures which don't repro locally.

src/Shared/UnitTests/EngineTestEnvironment.cs
```diff
@@ -4,12 +4,36 @@
...
 namespace Microsoft.Build.UnitTests
 {
     public partial class TestEnvironment
     {
+        // reset the default build manager and the state it might have accumulated from other tests
+        private object _resetBuildManager = new ResetDefaultBuildManager();
+
+        private class ResetDefaultBuildManager
+        {
+            public ResetDefaultBuildManager()
```
cdmihai: @rainersigwald Are you OK with this preventive cleanup? It does not seem to have a noticeable impact on test execution time.

src/Build/Resources/Strings.resx
cdmihai: The targets are mentioned in this error. Do you mean splitting this message into multiple errors for each target in the target list?

PR #3862, opened 2018-10-16
Evaluate nodes in parallel

yash256:


src/Build/Graph/ProjectGraph.cs
```diff
@@ -169,17 +172,19 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
-            var nodeStates = new Dictionary<ConfigurationMetadata, NodeState>();
+            var nodeStates = new Dictionary<ProjectGraphNode, NodeState>();
             var entryPointNodes = new List<ProjectGraphNode>();
+            var tasksInProgress = new ConcurrentDictionary<ConfigurationMetadata, Task>();
             foreach (var entryPoint in entryPoints)
             {
                 PropertyDictionary<ProjectPropertyInstance> globalPropertyDictionary = CreatePropertyDictionary(entryPoint.GlobalProperties);
                 var configurationMetadata = new ConfigurationMetadata(FileUtilities.NormalizePath(entryPoint.ProjectFile), globalPropertyDictionary);
-                if(!nodeStates.TryGetValue(configurationMetadata, out NodeState nodeState))
+                LoadGraph(configurationMetadata, projectCollection, globalPropertyDictionary, tasksInProgress);
```
dfederm: Can LoadGraph take a list of entrypoints now so the entrypoints can be processed in parallel? Right now it will process one entrypoint in parallel, then the next, etc.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +322,71 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+        /// <summary>
+        /// Load a graph with root node at entryProjectFile
+        /// Maintain a queue of projects to be processed and evaluate projects in parallel
+        /// </summary>
+        private void LoadGraph(ConfigurationMetadata entryProject, ProjectCollection projectCollection, PropertyDictionary<ProjectPropertyInstance> globalProperties, ConcurrentDictionary<ConfigurationMetadata, Task> tasksInProgress)
+        {
+            var waitHandle = new AutoResetEvent(false);
+            var projectsToEvalute = new ConcurrentQueue<ConfigurationMetadata>();
+            projectsToEvalute.Enqueue(entryProject);
+            while (projectsToEvalute.Count != 0 || tasksInProgress.Any())
```
dfederm: Use `tasksInProgress.Count != 0`

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +322,71 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                                    projectReferenceToParse.EvaluatedInclude,
+                                    parsedProject.ProjectInstance.FullPath));
+                            }
+
+                            string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+                            PropertyDictionary<ProjectPropertyInstance> projectReferenceGlobalProperties = GetProjectReferenceGlobalProperties(projectReferenceToParse, globalProperties);
+                            var projectReferenceConfigurationMetadata = new ConfigurationMetadata(projectReferenceFullPath, projectReferenceGlobalProperties);
+                            if (!_allParsedProjects.TryGetValue(projectReferenceConfigurationMetadata, out ProjectGraphNode parsedProjectReference))
+                            {
+                                if (!tasksInProgress.TryGetValue(projectReferenceConfigurationMetadata, out Task taskInProgress))
```
dfederm: I don't think the values of `tasksInProgress` are ever used. Maybe this should be a set instead of a dictionary.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -330,44 +396,33 @@ private enum NodeState
...
-                {
-                    throw new InvalidOperationException(string.Format(
-                        CultureInfo.InvariantCulture,
-                        ResourceUtilities.GetResourceString(
-                            "ProjectGraphDoesNotSupportProjectReferenceWithToolset"),
-                        projectReferenceToParse.EvaluatedInclude,
-                        parsedProject.ProjectInstance.FullPath));
-                }
-
                 string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
```
rainersigwald: Is this item/metadata manipulation required? Since this is now pushed to after graph construction, can you use the graph directly?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +322,71 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                                {
+                                    projectsToEvalute.Enqueue(projectReferenceConfigurationMetadata);
+                                }
+                            }
+                        }
+                    });
+
+                    if (tasksInProgress.TryAdd(projectToEvaluate, task))
+                    {
+                        task.ContinueWith(t =>
```
rainersigwald:
```suggestion
                        task.ContinueWith(_ =>
```

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +322,71 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                        {
+                            tasksInProgress.TryRemove(projectToEvaluate, out Task parsedProjectReferenceTask);
```
rainersigwald:
```suggestion
                            tasksInProgress.TryRemove(projectToEvaluate, out var _);
```

src/Build/Graph/ProjectGraph.cs
rainersigwald:
```suggestion
                                if (!tasksInProgress.TryGetValue(projectReferenceConfigurationMetadata, out var _))
```

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +322,71 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                                    ResourceUtilities.GetResourceString(
+                                        "ProjectGraphDoesNotSupportProjectReferenceWithToolset"),
+                                    projectReferenceToParse.EvaluatedInclude,
+                                    parsedProject.ProjectInstance.FullPath));
+                            }
+
+                            string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+                            PropertyDictionary<ProjectPropertyInstance> projectReferenceGlobalProperties = GetProjectReferenceGlobalProperties(projectReferenceToParse, globalProperties);
+                            var projectReferenceConfigurationMetadata = new ConfigurationMetadata(projectReferenceFullPath, projectReferenceGlobalProperties);
+                            if (!_allParsedProjects.TryGetValue(projectReferenceConfigurationMetadata, out ProjectGraphNode parsedProjectReference))
```
rainersigwald:
```suggestion
                            if (!_allParsedProjects.TryGetValue(projectReferenceConfigurationMetadata, out var _))
```

src/Build/Graph/ProjectGraph.cs
```diff
@@ -169,17 +172,19 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
             var entryPointNodes = new List<ProjectGraphNode>();
+            var tasksInProgress = new ConcurrentDictionary<ConfigurationMetadata, Task>();
             foreach (var entryPoint in entryPoints)
             {
                 PropertyDictionary<ProjectPropertyInstance> globalPropertyDictionary = CreatePropertyDictionary(entryPoint.GlobalProperties);
                 var configurationMetadata = new ConfigurationMetadata(FileUtilities.NormalizePath(entryPoint.ProjectFile), globalPropertyDictionary);
-                if(!nodeStates.TryGetValue(configurationMetadata, out NodeState nodeState))
+                LoadGraph(configurationMetadata, projectCollection, globalPropertyDictionary, tasksInProgress);
+                entryPointNodes.Add(_allParsedProjects[configurationMetadata]);
+                if (!nodeStates.TryGetValue(_allParsedProjects[configurationMetadata], out NodeState nodeState))
```
rainersigwald: Several of these dictionary accesses have a throwaway value. C# now allows a nice pattern for that with the `_` variable.

```suggestion
                if (!nodeStates.TryGetValue(_allParsedProjects[configurationMetadata], out var _))
```

src/Build/Graph/ProjectGraph.cs
yash256: the project references are not set up till this point even though all graph nodes have been constructed.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +330,73 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
         }
 
+        /// <summary>
+        /// Load a graph with root node at entryProjectFile
+        /// Maintain a queue of projects to be processed and evaluate projects in parallel
+        /// </summary>
+        private void LoadGraph(ConcurrentQueue<ConfigurationMetadata> projectsToEvaluate, ProjectCollection projectCollection, ConcurrentDictionary<ConfigurationMetadata, object> tasksInProgress)
+        {
+            var waitHandle = new AutoResetEvent(false);
+            while (projectsToEvaluate.Count != 0 || tasksInProgress.Count() != 0)
```
dfederm: > tasksInProgress.Count()

Avoid the Linq and just use `Count` (the property)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -169,17 +172,27 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
-                if(!nodeStates.TryGetValue(configurationMetadata, out NodeState nodeState))
+                projectsToEvaluate.Enqueue(configurationMetadata);
+                entryPointConfigurationMetadata.Add(configurationMetadata);
+            }
+
+            LoadGraph(projectsToEvaluate, projectCollection, tasksInProgress);
+            foreach(var configurationMetadata in entryPointConfigurationMetadata)
+            {
+                entryPointNodes.Add(_allParsedProjects[configurationMetadata]);
+                if (!nodeStates.TryGetValue(_allParsedProjects[configurationMetadata], out var _))
```
dfederm: Related to Rainer's previous comment, instead of `TryGetValue(key, out var _)` I think `ContainsKey` is even simpler.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +330,73 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                                    projectReferenceToParse.EvaluatedInclude,
+                                    parsedProject.ProjectInstance.FullPath));
+                            }
+
+                            string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+                            PropertyDictionary<ProjectPropertyInstance> projectReferenceGlobalProperties = GetProjectReferenceGlobalProperties(projectReferenceToParse, projectToEvaluate.GlobalProperties);
+                            var projectReferenceConfigurationMetadata = new ConfigurationMetadata(projectReferenceFullPath, projectReferenceGlobalProperties);
+                            if (!_allParsedProjects.TryGetValue(projectReferenceConfigurationMetadata, out var _))
+                            {
+                                if (!tasksInProgress.TryGetValue(projectReferenceConfigurationMetadata, out var _))
```
dfederm: There may be a race condition here between `_allParsedProjects` and `tasksInProgress`. Independently they're thread-safe, but the state of one might change while you're checking the other.

Maybe flipping the order would be slight better even. Right now it might not be fully parsed when you check `_allParsedProjects`, then it finishes before you check `tasksInProgress`. If you flip it though, you at least wouldn't see that specific race condition. Not sure if there may be others though.

If so, you'll need a single lock around any accesses to the state required for processing (`_allParsedProjects`, `tasksInProgress` and `projectsToEvaluate`).

rainersigwald: This is a good catch. I _think_ just flipping the order works, but I'm not that confident.

src/Build/Graph/ProjectGraph.cs
rainersigwald: This seems unfortunate. How hard would it be to fix?

yash256: One option to fix this is to construct a ProjectGraphNode object even before evaluation completes. This way, there would be no initialization done in the constructor of each node.
The other option would be to keep a list of parent nodes in tasksInProgress. Once a task completes, add project references to each of them.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +330,73 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
             return graphNode;
         }
 
+        /// <summary>
+        /// Load a graph with root node at entryProjectFile
+        /// Maintain a queue of projects to be processed and evaluate projects in parallel
+        /// </summary>
+        private void LoadGraph(ConcurrentQueue<ConfigurationMetadata> projectsToEvaluate, ProjectCollection projectCollection, ConcurrentDictionary<ConfigurationMetadata, object> tasksInProgress)
+        {
+            var waitHandle = new AutoResetEvent(false);
```
cdmihai: Rename waitHandle to something that describes the meaning :)

src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,10 +330,73 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                            }
+
+                            string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
+                            PropertyDictionary<ProjectPropertyInstance> projectReferenceGlobalProperties = GetProjectReferenceGlobalProperties(projectReferenceToParse, projectToEvaluate.GlobalProperties);
+                            var projectReferenceConfigurationMetadata = new ConfigurationMetadata(projectReferenceFullPath, projectReferenceGlobalProperties);
+                            if (!tasksInProgress.ContainsKey(projectReferenceConfigurationMetadata))
+                            {
+                                if (!_allParsedProjects.ContainsKey(projectReferenceConfigurationMetadata))
+                                {
+                                    projectsToEvaluate.Enqueue(projectReferenceConfigurationMetadata);
```
cdmihai: What if you set WaitHandle here instead? It would make the main thread loop spin new tasks faster right?

yash256: It can't be set just here. It has to be set when the thread terminates as well so that the main loop exits after parsing all targets.

cdmihai: Discussed offline: OK to set when either of the two collections gets changed. The extra sets don't seem to be harmful.

PR #3863, opened 2018-10-16
Investigate CI failure

cdmihai:
Something changes the environment variables during the added test which causes the TestEnvironment to fail when it checks that the test did not change environment variables.
Using this CI to track down what's happening.


rainersigwald:
I'm going to close this as stale. Is this still something we should be chasing @cdmihai?

PR #3866, opened 2018-10-18
[WIP] Do not set unchanged environment variables

cdmihai:
In #3854 this change was causing test `Regress265010 ` to fail with `Shouldly.ShouldAssertException : False\n should be\nTrue\n but was not\n\nAdditional Info:\n environment variable removed: BUILD_REQUESTEDFOREMAIL`

This is now passing due to a workaround in the TestEnvironment: https://github.com/Microsoft/msbuild/blob/exp/net472/src/Shared/UnitTests/TestEnvironment.cs#L372-L374

### DIFF
--- a/eng/cibuild_bootstrapped_msbuild.ps1
+++ b/eng/cibuild_bootstrapped_msbuild.ps1
@@ -131,4 +131,6 @@ finally {
   if ($prepareMachine) {
     Stop-Processes
   }
+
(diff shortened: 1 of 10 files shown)


rainersigwald:
Can you also remove the workaround to validate that this works?

rainersigwald:
Team triage: closing as stale. We can reconsider in the future.

PR #3873, opened 2018-10-23
Handle exceptions thrown by tasks during evaluation in ProjectGraph

yash256:


src/Build/Graph/ProjectGraph.cs
```diff
@@ -396,6 +426,14 @@ private void LoadGraph(ConcurrentQueue<ConfigurationMetadata> projectsToEvaluate
                     evaluationWaitHandle.WaitOne();
                 }
             }
+            if(exceptionsInTasks.Count != 0)
```
dfederm: Style nit: add newline before and add space after `if`.

AndyGerlicher:
```suggestion
            if (exceptionsInTasks.Count != 0)
```
I gotta try the new feature.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -185,28 +185,52 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
             {
-                entryPointNodes.Add(_allParsedProjects[configurationMetadata]);
-                if (!nodeStates.TryGetValue(_allParsedProjects[configurationMetadata], out var _))
+                foreach (var configurationMetadata in entryPointConfigurationMetadata)
                 {
-                    DetectCycles(_allParsedProjects[configurationMetadata], nodeStates, projectCollection, configurationMetadata.GlobalProperties);
+                    entryPointNodes.Add(_allParsedProjects[configurationMetadata]);
+                    if (!nodeStates.TryGetValue(_allParsedProjects[configurationMetadata], out var _))
+                    {
+                        try
```
dfederm: If `LoadGraph` returns true we don't have any exceptions right? Would it be worth just leaving `CircularDependencyException` a top-level exception here instead of wrapping it in the `AggregateException`?

IE, the ctor throws either a `CircularDependencyException` or a `AggregateException` containing invalid project exceptions

src/Build/Graph/ProjectGraph.cs
```diff
@@ -185,28 +185,52 @@ public ProjectGraph(IEnumerable<ProjectGraphEntryPoint> entryPoints, ProjectColl
...
+                    }
                 }
+
+                EntryPointNodes = entryPointNodes.AsReadOnly();
+                ProjectNodes = _allParsedProjects.Values.ToList();
+                GraphRoots = graphRoots.AsReadOnly();
+            }
+            else
+            {
+                exceptionInLoading = true;
```
dfederm: I think `exceptionsInLoading` can be replaced with `exceptions.Count > 0`

dfederm: Or, since `LoadGraph` only returns false when there are exceptions, jsut directly throw here instead of below.

src/Build/Graph/ProjectGraph.cs
yash256: sounds good.. will fix in the next iteration

src/Build/Graph/ProjectGraph.cs
```diff
@@ -193,8 +193,8 @@ public ProjectGraph(ProjectGraphEntryPoint entryPoint, ProjectCollection project
         /// a default implementation that calls the ProjectInstance constructor. See the remarks
         /// on <see cref="ProjectInstanceFactoryFunc"/> for other scenarios.
         /// </param>
-        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
-        /// <exception cref="System.InvalidOperationException">
+        /// <exception cref="AggregateException">If the evaluation of any project in the graph fails. The InnerException contains <see cref="InvalidOperationException"/>.</exception>
+        /// <exception cref="CircularDependencyException"> If the evaluation is successful but the project graph contains a circular dependency</exception>
         /// If a null reference is returned from <paramref name="projectInstanceFactory"/>.
```
erikma: Dangling doc comment, need to add this text into the AggEx description above

erikma:
Looks good but for the need to fix up the exception tag text as noted.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -193,8 +193,8 @@ public ProjectGraph(ProjectGraphEntryPoint entryPoint, ProjectCollection project
         /// a default implementation that calls the ProjectInstance constructor. See the remarks
         /// on <see cref="ProjectInstanceFactoryFunc"/> for other scenarios.
         /// </param>
-        /// <exception cref="InvalidProjectFileException">If the evaluation of any project in the graph fails.</exception>
-        /// <exception cref="System.InvalidOperationException">
+        /// <exception cref="AggregateException">If the evaluation of any project in the graph fails. The InnerException contains <see cref="InvalidOperationException"/>.</exception>
+        /// <exception cref="CircularDependencyException"> If the evaluation is successful but the project graph contains a circular dependency</exception>
```
rainersigwald: nit
```suggestion
        /// <exception cref="CircularDependencyException">If the evaluation is successful but the project graph contains a circular dependency</exception>
```

Issue #3690, opened 2018-09-01
Allow non-predicted targets when called from the same project


dfederm:
MSBuild tasks invoked on the same project should always be allowed, regardless of whether a the target is specified in the p2p protocol.

Example, when building cross-targeting projects, the project will MSBuild recursively for each target framework.

cdmihai:
Resolved in #3878

PR #3878, opened 2018-10-24
Constrained ms build task

### LINKED ISSUE #3688: Fail the build when the MSBuild task is used on a target not predicted by the project graph
If the target result cache does not have the results for a given invocation, fail the build.

Note that initial targets should not be allowed either, unless they were also an entry point.

cdmihai:
Closes #3688, #3690

### DIFF
--- a/src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
+++ b/src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
@@ -19,21 +19,52 @@ namespace Microsoft.Build.Graph.UnitTests
     {
         private readonly string _project = @"
                 <Project DefaultTargets='BuildSelf'>
-                    <Target Name='BuildReference'>
-                        <MSBuild Projects='{0}' Targets='ReferenceTarget'/>
+
+                    <ItemGroup>
+                        <ProjectReference Include='{0}'/>
+                    </ItemGroup>
+
+                    <Target Name='BuildDeclaredReference'>
+                        <MSBuild
+                            Projects='{1}'
+                            Targets='DeclaredReferenceTarget'
+                            {3}
+                        />
+                    </Target>
+
+                    <Target Name='BuildUndeclaredReference'>
+                        <MSBuild
+                            Projects='{2}'
+                            Targets='UndeclaredReferenceTarget'
+                            {3}
+                        />
                     </Target>
 
                     <Target Name='BuildSelf'>
-                        <MSBuild Projects='$(MSBuildThisFile)' Targets='SelfTarget'/>
+                        <MSBuild
+                            Projects='$(MSBuildThisFile)'
+                            Targets='SelfTarget'
+                            {3}
+                        />
+                    </Target>
+
+                    <Target Name='CallTarget'>
+                        <CallTarget Targets='SelfTarget'/>  
                     </Target>
 
                     <Target Name='SelfTarget'>
                     </Target>
                 </Project>";
 
-        private readonly string _reference = @"
+        private readonly string _declaredReference = @"
                 <Project>
-                    <Target Name='ReferenceTarget'>
+                    <Target Name='DeclaredReferenceTarget'>
+                    </Target>
+                </Project>";
+
+        private readonly string _undeclaredReference = @"
+                <Project>
+                    <Target Name='UndeclaredReferenceTarget'>
                     </Target>
                 </Project>";
 
@@ -45,9 +76,9 @@ namespace Microsoft.Build.Graph.UnitTests
         }
 
         [Theory]
-        [InlineData(BuildResultCode.Success, new string[] {})]
+        [InlineData(BuildResultCode.Success, new string[] { })]
         [InlineData(BuildResultCode.Success, new[] {"BuildSelf"})]
-        public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedBuildResult, string[] targets)
+        public void CacheAndTaskEnforcementShouldAcceptSelfReferences(BuildResultCode expectedBuildResult, string[] targets)
         {
             AssertBuild(targets,
                 (result, logger) =>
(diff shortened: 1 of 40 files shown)


cdmihai:
Okay, ready to review. I disabled testing multi proc builds on macos, as the failures repro even without these changes (#3915).

ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
```diff
@@ -134,6 +134,7 @@ public partial class CallTarget : Microsoft.Build.Tasks.TaskExtension
         [Microsoft.Build.Framework.OutputAttribute]
         public Microsoft.Build.Framework.ITaskItem[] TargetOutputs { get { throw null; } }
         public string[] Targets { get { throw null; } set { } }
+        [System.ObsoleteAttribute("Not supported")]
```
rainersigwald: I believe this counts as a breaking change since it adds a warning. Do we need it? We have a window for that but it's almost entirely closed now.

src/Shared/Debugging/PrintLineDebugger.cs
```diff
@@ -0,0 +1,209 @@
...
+using CommonWriterType = System.Action<string, string, System.Collections.Generic.IEnumerable<string>>;
+
+namespace Microsoft.Build.Shared.Debugging
+{
+    /// <summary>
+    ///     A class to help with printline debugging in difficult environments like CI, or when debugging msbuild through other
+    ///     tools like VS or CLI.
+    ///     See example usages in PrintLineDebugger_Tests
+    /// </summary>
+    internal class PrintLineDebugger : IDisposable
```
rainersigwald: Should we ifdef this to make sure we don't accidentally slow down released builds?

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -59,51 +90,248 @@ public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedB
...
+                    result.OverallResult.ShouldBe(BuildResultCode.Success);
+
+                    logger.Errors.ShouldBeEmpty();
+                });
+        }
+
+        [Fact(Skip = "https://github.com/Microsoft/msbuild/issues/3876")]
+        public void CacheEnforcementShouldFailWhenReferenceWasNotPreviouslyBuiltAndOnContinueOnError()
+        {
+            CacheEnforcementShouldFailWhenReferenceWasNotPreviouslyBuilt2(true);
```
rainersigwald: nit: use arg name for bool arguments

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -59,51 +90,248 @@ public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedB
...
+                    logger.ErrorCount.ShouldBe(1);
+
+                    logger.Errors.First().Message.ShouldStartWith("MSB4254:");
+                },
+                addContinueOnError: addContinueOnError);
         }
 
-        private void AssertBuild(string[] targets, Action<BuildResult, MockLogger> assert, bool buildReference = false)
+        [Fact]
+        public void TaskEnforcementShouldNotAcceptPreviouslyBuiltReferences()
```
rainersigwald: I'm not sure I understand this test name. Why shouldn't it accept this reference? Because it's previously built but also undeclared for this project?

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -59,51 +90,248 @@ public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedB
...
+                    logger.Errors.First().Message.ShouldStartWith("MSB4254:");
+                },
+                buildUndeclaredReference: true);
+        }
+
+        public static IEnumerable<object[]> TaskEnforcementShouldNormalizeFilePathsTestData
+        {
+            get
+            {
+                Func<string, string> preserve = path => path;
```
rainersigwald: nit: inconsistent capitalization here; maybe just initial-caps these first two?

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -59,51 +90,248 @@ public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedB
...
+                    return Path.Combine("..", directory, file);
+                };
+
+                Func<string, string> ToForwardSlash = path => path.ToSlash();
+
+                Func<string, string> ToBackwardSlash = path => path.ToBackwardSlash();
+
+                Func<string, string> ToDuplicateSlashes = path => path.Replace("/", "//").Replace(@"\", @"\\");
+
+                yield return new object[]
```
rainersigwald: If this is the full matrix, can you compute it dynamically instead for clarity? If it's not, can you comment what it is and why it's not?

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
```diff
@@ -59,51 +90,248 @@ public void CacheEnforcementShouldAcceptSelfReferences(BuildResultCode expectedB
...
+            Func<string, string> projectReferenceModifier = null,
+            Func<string, string> msbuildOnDeclaredReferenceModifier = null)
         {
             using (var env = TestEnvironment.Create())
             using (var buildManager = new BuildManager())
             {
+                if (NativeMethodsShared.IsOSX)
+                {
+                    // OSX links /var into /private, which makes Path.GetTempPath() to return "/var..." but Directory.GetCurrentDirectory to return "/private/var..."
+                    // this discrepancy fails the msbuild task enforcements due to failed path equality checks
```
rainersigwald: Good comment!

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -962,6 +987,24 @@ private async Task<WorkUnitResult> ExecuteInstantiatedTask(ITaskExecutionHost ta
...
+            }
+
+            var declaredProjects = _taskExecutionHost.ProjectInstance.GetItems(MSBuildConstants.ProjectReferenceItemName)
+                .Select(p => FileUtilities.NormalizePath(p.EvaluatedInclude))
+                .ToHashSet();
+            declaredProjects.Add(_taskExecutionHost.ProjectInstance.FullPath);
+
+            return msbuildTask.Projects
+                .Where(p => !declaredProjects.Contains(FileUtilities.NormalizePath(p.ItemSpec)))
+                .Select(p => p.ItemSpec)
```
rainersigwald: This seems like it might be expensive. Do you have any perf indications either way?

src/Build/Resources/Strings.resx
```diff
@@ -1728,13 +1728,20 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
       LOCALIZATION:  Do not localize the following words: ProjectInstanceFactoryFunc.
     </comment>
   </data>
+  <data name="UndeclaredMSBuildTasksNotAllowedInIsolatedGraphBuilds" UESanitized="false" Visibility="Public">
+    <value>MSB4254: The MSBuild task is building project(s) {0} which are not specified in the ProjectReference item.</value>
```
rainersigwald: Should we say something about "and you opted into isolation"?

src/Shared/FileUtilities.cs
```diff
@@ -1144,6 +1144,11 @@ internal static string ToSlash(this string s)
             return s.Replace('\\', '/');
         }
 
+        internal static string ToBackwardSlash(this string s)
```
rainersigwald: nit:

```suggestion
        internal static string ToBackslash(this string s)
```

src/Shared/Debugging/PrintLineDebugger.cs
```diff
@@ -0,0 +1,209 @@
...
+        public void Dispose()
+        {
+            ReleaseUnmanagedResources();
+            GC.SuppressFinalize(this);
+        }
+
+        public static CommonWriterType GetStaticWriter()
+        {
+            return (CommonWriterType) CommonWriterProperty.Value.GetValue(null, null);
+            ;
```
rainersigwald: nit: delete line

src/Shared/Debugging/PrintLineDebugger.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            var currentWriter = GetStaticWriter();
+
+            if (currentWriter != null)
+            {
+                ErrorUtilities.ThrowInternalError("Cannot set a new writer over an old writer. Remove the old one first");
+            }
+
+            // wrap with a lock so multi threaded logging does not break messages apart
+            CommonWriterProperty.Value.SetValue(null, (CommonWriterType) LockWrappedWriter);
```
rainersigwald: Is it safe to set the value in this way? Is there a race between multiple threads hitting this line?

ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
cdmihai: The field is currently not implemented. I'd rather we never implement it to keep things simple. Instead of deprecating should I change the comments for these fields to something like `Not implemented, never will, forget you've seen this`

src/Shared/Debugging/PrintLineDebugger.cs
cdmihai: Yes, for now it's probably for the best. Future wise, it would be nice we had something turned on all the time so we could better debug msbuild binaries already integrated in VS / CLI. A unified (as opposed to current state of having N ways) and configurable design for adding printline statements for anywhere in the code. It could be a combination of prewritten and committed groups of messages that can be turned on via unique environment variables (e.g. node communication message group, evaluation message group, scheduler message group, etc).

src/Shared/Debugging/PrintLineDebugger.cs
cdmihai: The usage intent I had in mind was to investigate one problem at a time, which means setting one writer only once (either from a test, if tests are failing, or from the main method or BuildManager when investigating under VS). Hence, no need for thready safety. If you're alright with this assumption, I'll put in a comment about it.

rainersigwald: Sounds good.

src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
cdmihai: Will rename to `TaskEnforcementShouldFailOnPreviouslyBuiltButUndeclaredReferences`

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
cdmihai: No perf experimentation, but I'll delinqify it to offset the normalization costs. In non isolated builds it gets skipped. In isolated builds, I don't expect this to be on the hot path since if there is a big p2p graph where this would method would show up (thousands of nodes), then I'd expect the parallelism to hide it.

Issue #3840, opened 2018-10-08
Test error scenarios for graph builds


dfederm:
I observed some weird error behavior, so we should explicitly test some of this stuff.

1. An entry point project is invalid
2. A non-entry point project is invalid
3. An entry point project fails to build
4. A non-entry point project failed to build
5. All of the above for both command-line and API builds

PR #3879, opened 2018-10-24
Move graph build logic to the BuildManager

dfederm:
This enables programatic graph builds by exposing a way to ask the BuildManager for a graph build instead of the logic being inside XMake.

This also fixes a couple bugs related to how graph builds were happening and gets the logic "closer" to where we want it to end up in the end (the scheduler).

Closes: #3840

### DIFF
--- a/ref/Microsoft.Build/net/Microsoft.Build.cs
+++ b/ref/Microsoft.Build/net/Microsoft.Build.cs
@@ -926,13 +926,16 @@ namespace Microsoft.Build.Execution
         public static Microsoft.Build.Execution.BuildManager DefaultBuildManager { get { throw null; } }
         public void BeginBuild(Microsoft.Build.Execution.BuildParameters parameters) { }
         public Microsoft.Build.Execution.BuildResult Build(Microsoft.Build.Execution.BuildParameters parameters, Microsoft.Build.Execution.BuildRequestData requestData) { throw null; }
+        public Microsoft.Build.Execution.GraphBuildResult Build(Microsoft.Build.Execution.BuildParameters parameters, Microsoft.Build.Execution.GraphBuildRequestData requestData) { throw null; }
         public Microsoft.Build.Execution.BuildResult BuildRequest(Microsoft.Build.Execution.BuildRequestData requestData) { throw null; }
+        public Microsoft.Build.Execution.GraphBuildResult BuildRequest(Microsoft.Build.Execution.GraphBuildRequestData requestData) { throw null; }
         public void CancelAllSubmissions() { }
         public void Dispose() { }
         public void EndBuild() { }
         ~BuildManager() { }
         public Microsoft.Build.Execution.ProjectInstance GetProjectInstanceForBuild(Microsoft.Build.Evaluation.Project project) { throw null; }
         public Microsoft.Build.Execution.BuildSubmission PendBuildRequest(Microsoft.Build.Execution.BuildRequestData requestData) { throw null; }
+        public Microsoft.Build.Execution.GraphBuildSubmission PendBuildRequest(Microsoft.Build.Execution.GraphBuildRequestData requestData) { throw null; }
         public void ResetCaches() { }
         public void ShutdownAllNodes() { }
     }
@@ -1042,6 +1045,48 @@ namespace Microsoft.Build.Execution
         public void ExecuteAsync(Microsoft.Build.Execution.BuildSubmissionCompleteCallback callback, object context) { }
     }
     public delegate void BuildSubmissionCompleteCallback(Microsoft.Build.Execution.BuildSubmission submission);
+    public sealed partial class GraphBuildRequestData
+    {
+        public GraphBuildRequestData(Microsoft.Build.Graph.ProjectGraph projectGraph, System.Collections.Generic.ICollection<string> targetsToBuild) { }
+        public GraphBuildRequestData(Microsoft.Build.Graph.ProjectGraph projectGraph, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
+        public GraphBuildRequestData(Microsoft.Build.Graph.ProjectGraph projectGraph, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public GraphBuildRequestData(Microsoft.Build.Graph.ProjectGraphEntryPoint projectGraphEntryPoint, System.Collections.Generic.ICollection<string> targetsToBuild) { }
+        public GraphBuildRequestData(Microsoft.Build.Graph.ProjectGraphEntryPoint projectGraphEntryPoint, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
+        public GraphBuildRequestData(Microsoft.Build.Graph.ProjectGraphEntryPoint projectGraphEntryPoint, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public GraphBuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, System.Collections.Generic.ICollection<string> targetsToBuild) { }
+        public GraphBuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
+        public GraphBuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public GraphBuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
+        public GraphBuildRequestData(string projectFullPath, System.Collections.Generic.IDictionary<string, string> globalProperties, System.Collections.Generic.ICollection<string> targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public Microsoft.Build.Execution.BuildRequestDataFlags Flags { get { throw null; } }
+        public Microsoft.Build.Execution.HostServices HostServices { get { throw null; } }
+        public Microsoft.Build.Graph.ProjectGraph ProjectGraph { get { throw null; } }
+        public System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> ProjectGraphEntryPoints { get { throw null; } }
+        public System.Collections.Generic.ICollection<string> TargetNames { get { throw null; } }
+    }
+    public sealed partial class GraphBuildResult
+    {
+        internal GraphBuildResult() { }
+        public bool CircularDependency { get { throw null; } }
+        public System.Exception Exception { get { throw null; } }
+        public Microsoft.Build.Execution.BuildResult this[Microsoft.Build.Graph.ProjectGraphNode node] { get { throw null; } }
+        public Microsoft.Build.Execution.BuildResultCode OverallResult { get { throw null; } }
+        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, Microsoft.Build.Execution.BuildResult> ResultsByNode { get { throw null; } }
+        public int SubmissionId { get { throw null; } }
+    }
+    public partial class GraphBuildSubmission
+    {
+        internal GraphBuildSubmission() { }
+        public object AsyncContext { get { throw null; } }
+        public Microsoft.Build.Execution.BuildManager BuildManager { get { throw null; } }
+        public Microsoft.Build.Execution.GraphBuildResult BuildResult { get { throw null; } }
+        public bool IsCompleted { get { throw null; } }
+        public int SubmissionId { get { throw null; } }
+        public System.Threading.WaitHandle WaitHandle { get { throw null; } }
+        public Microsoft.Build.Execution.GraphBuildResult Execute() { throw null; }
+        public void ExecuteAsync(Microsoft.Build.Execution.GraphBuildSubmissionCompleteCallback callback, object context) { }
+    }
+    public delegate void GraphBuildSubmissionCompleteCallback(Microsoft.Build.Execution.GraphBuildSubmission submission);
     public partial class HostServices
     {
         public HostServices() { }
@@ -1427,7 +1472,7 @@ namespace Microsoft.Build.Graph
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> EntryPointNodes { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> GraphRoots { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> ProjectNodes { get { throw null; } }
-        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, System.Collections.Immutable.ImmutableList<string>> GetTargetLists(string[] entryProjectTargets) { throw null; }
+        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, System.Collections.Immutable.ImmutableList<string>> GetTargetLists(System.Collections.Generic.ICollection<string> entryProjectTargets) { throw null; }
         public delegate Microsoft.Build.Execution.ProjectInstance ProjectInstanceFactoryFunc(string projectPath, System.Collections.Generic.Dictionary<string, string> globalProperties, Microsoft.Build.Evaluation.ProjectCollection projectCollection);
     }
     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
--- a/ref/Microsoft.Build/netstandard/Microsoft.Build.cs
+++ b/ref/Microsoft.Build/netstandard/Microsoft.Build.cs
@@ -922,13 +922,16 @@ namespace Microsoft.Build.Execution
         public static Microsoft.Build.Execution.BuildManager DefaultBuildManager { get { throw null; } }
         public void BeginBuild(Microsoft.Build.Execution.BuildParameters parameters) { }
         public Microsoft.Build.Execution.BuildResult Build(Microsoft.Build.Execution.BuildParameters parameters, Microsoft.Build.Execution.BuildRequestData requestData) { throw null; }
+        public Microsoft.Build.Execution.GraphBuildResult Build(Microsoft.Build.Execution.BuildParameters parameters, Microsoft.Build.Execution.GraphBuildRequestData requestData) { throw null; }
         public Microsoft.Build.Execution.BuildResult BuildRequest(Microsoft.Build.Execution.BuildRequestData requestData) { throw null; }
+        public Microsoft.Build.Execution.GraphBuildResult BuildRequest(Microsoft.Build.Execution.GraphBuildRequestData requestData) { throw null; }
         public void CancelAllSubmissions() { }
         public void Dispose() { }
         public void EndBuild() { }
         ~BuildManager() { }
         public Microsoft.Build.Execution.ProjectInstance GetProjectInstanceForBuild(Microsoft.Build.Evaluation.Project project) { throw null; }
         public Microsoft.Build.Execution.BuildSubmission PendBuildRequest(Microsoft.Build.Execution.BuildRequestData requestData) { throw null; }
+        public Microsoft.Build.Execution.GraphBuildSubmission PendBuildRequest(Microsoft.Build.Execution.GraphBuildRequestData requestData) { throw null; }
         public void ResetCaches() { }
         public void ShutdownAllNodes() { }
     }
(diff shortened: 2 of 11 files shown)


ref/Microsoft.Build/netstandard/Microsoft.Build.cs
```diff
@@ -1422,7 +1436,7 @@ public sealed partial class ProjectGraph
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> EntryPointNodes { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> GraphRoots { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Graph.ProjectGraphNode> ProjectNodes { get { throw null; } }
-        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, System.Collections.Immutable.ImmutableList<string>> GetTargetLists(string[] entryProjectTargets) { throw null; }
+        public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Graph.ProjectGraphNode, System.Collections.Immutable.ImmutableList<string>> GetTargetLists(System.Collections.Generic.ICollection<string> entryProjectTargets) { throw null; }
```
dfederm: Now that this is called from BuildManager instead of XMake, I can actually just make `GetTargetLists` internal (I think).

Any opinions on whether it should be internal (if possible)?

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -888,6 +905,79 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
...
+            {
+                VerifyStateInternal(BuildManagerState.Building);
+
+                try
+                {
+                    // Create the build request
+                    submission.BuildRequest = new BuildRequest(
+                        submission.SubmissionId,
+                        BackEnd.BuildRequest.InvalidNodeRequestId,
+                        GetNewConfigurationId(), // TODO: Should we create a real BuildRequestConfiguration for each graph request?
```
dfederm: `ExecuteSubmission` seems to create a unique configuration id for the project. Should I create a unique one for the graph? Or does it even matter in this case (I don't believe we will ever look this up by configuration id)?

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+                }
+
+                var projectGraph = submission.BuildRequestData.ProjectGraph;
+                if (projectGraph == null)
+                {
+                    // TODO: Need to pass through other stuff? Look how RequestBuilder creates ProjectInstance
+                    projectGraph = new ProjectGraph(submission.BuildRequestData.ProjectGraphEntryPoints);
+                }
+
+                // TODO: Handle multiple entry points
```
dfederm: This is a tough one. `BuildSubmission` has a `BuildResult`, but in our case there could me multiple `BuildResults` for each entry point. Maybe for graph builds we just always aggregate into a root build result which doesn't map to a particular project...

src/MSBuild/XMake.cs
```diff
@@ -1250,79 +1245,6 @@ private static BuildResult ExecuteBuild(BuildManager buildManager, BuildRequestD
...
         }
 
-        private static BuildResult ExecuteGraphBuild(
-            string projectFile,
-            string[] targets,
-            BuildManager buildManager,
-            ProjectCollection projectCollection,
-            Dictionary<string, string> globalProperties)
-        {
-            // TODO: Handle Ctrl-C cancellation properly.
```
dfederm: Cancellation now works correctly 😄

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+            }
+            catch (Exception ex)
+            {
+                if (ExceptionHandling.IsCriticalException(ex))
+                {
+                    throw;
+                }
+
+                // ProjectGraph throws an aggregate exception with InvalidProjectFileException inside when evaluation fails
+                if (ex is AggregateException aggregateException && aggregateException.InnerExceptions.All(innerException => innerException is InvalidProjectFileException))
```
dfederm: It'll be hard to test this e2e until #3873 is merged.

ref/Microsoft.Build/netstandard/Microsoft.Build.cs
cdmihai: Offline discussion: Users that want to distribute msbuild themselves via the new graph and cache serialization, will have to know what targets to call each graph node with in order to get those target results in the cache.

src/Build/BackEnd/BuildManager/BuildRequestData.cs
```diff
@@ -221,6 +222,170 @@ public BuildRequestData(string projectFullPath, IDictionary<string, string> glob
...
+
+        /// <summary>
+        /// Constructs a BuildRequestData for build requests based on a project graph entry points.
+        /// </summary>
+        /// <param name="projectGraphEntryPoints">The entry points to use in the build.</param>
+        /// <param name="targetsToBuild">The targets to build.</param>
+        /// <param name="hostServices">The host services to use, if any.  May be null.</param>
+        /// <param name="flags">Flags controlling this build request.</param>
+        /// <param name="propertiesToTransfer">The list of properties whose values should be transferred from the project to any out-of-proc node.</param>
+        public BuildRequestData(IEnumerable<ProjectGraphEntryPoint> projectGraphEntryPoints, string[] targetsToBuild, HostServices hostServices, BuildRequestDataFlags flags, IEnumerable<string> propertiesToTransfer)
```
cdmihai: Some of the build request data flags and that `propertiesToTransfer` sound like they either don't apply that well to graph builds or they would need extra consideration. If that's the case, there three options:
- make sure all the options have some determined behaviour for graph builds
- not care about these for now and not provide the overloads

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -765,6 +775,13 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
             ErrorUtilities.VerifyThrowArgumentNull(submission, nameof(submission));
             ErrorUtilities.VerifyThrow(!submission.IsCompleted, "Submission already complete.");
 
+            // Graph builds use a completely different mechanism for scheduling.
+            if (submission.BuildRequestData.ProjectGraph != null || submission.BuildRequestData.ProjectGraphEntryPoints != null)
```
cdmihai: Can extract to `BuildRequestData.IsGraphBased`

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: After a short poking around it seems that if a configuration ID starts flowing through the engine, it is kind of assumed it will actually exist in the ConfigCache. Assigning an id for an inexistent BuildRequestConfiguration might lead to funky behaviour. But I'm not super sure about that. 

---
In reply to: [227636190](https://github.com/Microsoft/msbuild/pull/3879#discussion_r227636190) [](ancestors = 227636190)

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -888,6 +905,79 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
...
+                        CheckSubmissionCompletenessAndRemove(submission);
+                        return;
+                    }
+
+                    if (_graphSchedulingCancellationSource == null)
+                    {
+                        _graphSchedulingCancellationSource = new CancellationTokenSource();
+                    }
+
+                    // Do the scheduling in a separate thread to unblock the calling thread
```
cdmihai: I wonder if this could deadlock, especially in /m scenarios. 
Out of proc nodes would call in on short lived packet processing threads which would end up querying the BuildManager which would make them request the already held lock.

As far as I can tell the sync lock is only held to serialize access to the BuildManager's decisions, not the whole build.

ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -981,6 +981,18 @@ public partial class BuildRequestData
...
+        public BuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, string[] targetsToBuild) { }
+        public BuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices) { }
+        public BuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags) { }
+        public BuildRequestData(System.Collections.Generic.IEnumerable<Microsoft.Build.Graph.ProjectGraphEntryPoint> projectGraphEntryPoints, string[] targetsToBuild, Microsoft.Build.Execution.HostServices hostServices, Microsoft.Build.Execution.BuildRequestDataFlags flags, System.Collections.Generic.IEnumerable<string> propertiesToTransfer) { }
```
rainersigwald: On the one hand, I hate exposing these thousand overloads. On the other, I can't be bothered to figure out which ones are most important. So 👍 I guess.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -888,6 +905,79 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
...
+                        BuildResult result = new BuildResult(submission.BuildRequest, new BuildAbortedException());
+                        submission.CompleteResults(result);
+                        submission.CompleteLogging(true);
+                        CheckSubmissionCompletenessAndRemove(submission);
+                        return;
+                    }
+
+                    if (_graphSchedulingCancellationSource == null)
+                    {
+                        _graphSchedulingCancellationSource = new CancellationTokenSource();
```
rainersigwald: I think it'd be a little cleaner to do something like

```c#
CancellationToken schedulingCancellationToken = _graphSchedulingCancellationSource?.Token ?? CancellationToken.None
```

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -888,6 +905,79 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
...
+
+                    // Do the scheduling in a separate thread to unblock the calling thread
+                    Task.Factory.StartNew(
+                        () =>
+                        {
+                            try
+                            {
+                                ExecuteGraphBuildScheduler(submission);
+                            }
+                            catch (Exception ex)
```
rainersigwald:
```suggestion
                            catch (Exception ex) when (!ExceptionHandling.IsCriticalException(ex))
```

(and remove the `if` below)

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -888,6 +905,79 @@ internal void ExecuteSubmission(BuildSubmission submission, bool allowMainThread
...
+                                HandleExecuteSubmissionException(submission, ex);
+                            }
+                        },
+                        _graphSchedulingCancellationSource.Token,
+                        TaskCreationOptions.LongRunning,
+                        TaskScheduler.Default);
+                }
+                catch (Exception ex)
+                {
+                    if (ExceptionHandling.IsCriticalException(ex))
```
rainersigwald: ditto

src/Build/BackEnd/BuildManager/BuildManager.cs
rainersigwald: Yeah, that seems reasonable--is it legal, or do we need to construct a "traversal project" entrypoint?

src/Build/BackEnd/BuildManager/BuildManager.cs
rainersigwald: Can you elaborate on the deadlock scenario? I'm not sure I understand from your description.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+            {
+                if (_shuttingDown)
+                {
+                    throw new BuildAbortedException();
+                }
+
+                var projectGraph = submission.BuildRequestData.ProjectGraph;
+                if (projectGraph == null)
+                {
+                    // TODO: Need to pass through other stuff? Look how RequestBuilder creates ProjectInstance
```
rainersigwald: I'm not sure I understand what this means, can you elaborate?

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: After reading further into the PR, it seems that you actually never use this BuildSubmission, so I guess you could just set the invalid id on it.

---
In reply to: [227900171](https://github.com/Microsoft/msbuild/pull/3879#discussion_r227900171) [](ancestors = 227900171,227636190)

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+                var finishedNodes = new HashSet<ProjectGraphNode>(projectGraph.ProjectNodes.Count);
+                var buildingNodes = new Dictionary<BuildSubmission, ProjectGraphNode>();
+                while (blockedNodes.Count > 0 || buildingNodes.Count > 0)
+                {
+                    waitHandle.WaitOne();
+
+                    lock (graphBuildStateLock)
+                    {
+                        var unblockedNodes = blockedNodes
+                            .Where(node => node.ProjectReferences.All(projectReference => finishedNodes.Contains(projectReference)))
```
rainersigwald: I think after this PR it's time to think about doing this with an array of waithandles and `WaitAny` to avoid at least one of these scans.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+
+                                waitHandle.Set();
+
+                                continue;
+                            }
+
+                            var request = new BuildRequestData(node.ProjectInstance, targetList.ToArray());
+
+                            // TODO Tack onto the existing submission instead of pending a whole new submission for every node
+                            // Among other things, this makes BuildParameters.DetailedSummary produce a summary for each node, which is not desirable.
```
rainersigwald: Do you have a clear picture of how to do this? Detailed summary is pretty important, I think.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+                                ProjectGraphNode finishedNode;
+                                lock (graphBuildStateLock)
+                                {
+                                    finishedNode = buildingNodes[finishedBuildSubmission];
+
+                                    finishedNodes.Add(finishedNode);
+                                    buildingNodes.Remove(finishedBuildSubmission);
+                                }
+
+                                if (finishedNode == entryNode)
```
cdmihai: There could be multiple entrypoints, and an entrypoint can actually not be a root. Shouldn't the results be reported only when all the ProjectGraph.GraphRoots are done? Or simpler, when the finishedNodes contains all the nodes in the graph.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1140,6 +1230,123 @@ private void IssueBuildSubmissionToScheduler(BuildSubmission submission, bool al
...
+                }
+            }
+            catch (Exception ex)
+            {
+                if (ExceptionHandling.IsCriticalException(ex))
+                {
+                    throw;
+                }
+
+                // ProjectGraph throws an aggregate exception with InvalidProjectFileException inside when evaluation fails
```
rainersigwald: More comment please. What's this block do?

src/Build/BackEnd/BuildManager/BuildManager.cs
rainersigwald: Yeah, let's block this on that. Looks like it should go in soon. @yash256 fyi.

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: If VS wants to use this for design time builds, they'd want to retrieve the results for each node in the graph. We can do a design session on how to best represent this. We could subclass the BuildResult, multi purpose it (ew), maybe have a GraphBuildSubmission, etc

src/Build/BackEnd/BuildManager/BuildManager.cs
dfederm: Well, that's semantically different. Effectively I'm just wanting to use 1 cancellation source for the BuildManager across sessions (ignoring EndBuilds + new BeginBuilds), so this is basically just lazily creating it. I don't want to use `CancellationToken.None` (which would possibly leak the thread), but instead the intent here is to use the existing cancellation source if there is one, and if there isn't one just create one and use that.

rainersigwald: Oh, I see--there is a codepath to call cancel on a token created here. Sounds good, but . . . comments :)

src/Build/BackEnd/BuildManager/BuildManager.cs
dfederm: I considered using the work queue, but I believe by default its degree of parallelism is 1, so it's not suited for a long-running thread like this.

src/Build/BackEnd/BuildManager/BuildManager.cs
dfederm: RequestBuilder creates project instances like this, which is way different than how ProjectGraph creates them:

```cs
            return new ProjectInstance(
                _requestEntry.RequestConfiguration.ProjectFullPath,
                globalProperties,
                toolsVersionOverride,
                _componentHost.BuildParameters,
                _nodeLoggingContext.LoggingService,
                new BuildEventContext(
                    _requestEntry.Request.BuildEventContext.SubmissionId,
                    _nodeLoggingContext.BuildEventContext.NodeId,
                    BuildEventContext.InvalidEvaluationId,
                    BuildEventContext.InvalidProjectInstanceId,
                    BuildEventContext.InvalidProjectContextId,
                    BuildEventContext.InvalidTargetId,
                    BuildEventContext.InvalidTaskId),
                sdkResolverService,
                _requestEntry.Request.SubmissionId,
                projectLoadSettings);

```

With Erik's recent change though I think I can more or less make this happen.

src/Build/BackEnd/BuildManager/BuildManager.cs
dfederm: No clear picture unfortunately. I can elaborate an idea of what I think should happen though.

dfederm:
Yes, I know I still need tests. The latest iteration had enough changes to warrant fresh eyes on it though.

dfederm:
Some example of console outputs:

A valid project referencing a valid project:
image

A valid project referencing an invalid project:
image

A valid project referencing a failing project:
image

src/Build/BackEnd/BuildManager/GraphBuildResult.cs
```diff
@@ -0,0 +1,97 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Build.Graph;
+
+namespace Microsoft.Build.Execution
+{
+    public sealed class GraphBuildResult
+    {
+        /// <summary>
+        /// Constructor creates a build result indicating a circular dependency was created.
```
dfederm: Woops. Should be "Constructor creates a build result with results for each graph node."

src/Build/BackEnd/BuildManager/GraphBuildRequestData.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <param name="projectFullPath">The full path to the project file.</param>
+        /// <param name="globalProperties">The global properties which should be used during evaluation of the project.  Cannot be null.</param>
+        /// <param name="targetsToBuild">The targets to build.</param>
+        /// <param name="hostServices">The host services to use.  May be null.</param>
+        /// <param name="flags">The <see cref="BuildRequestDataFlags"/> to use.</param>
+        public GraphBuildRequestData(string projectFullPath, IDictionary<string, string> globalProperties, ICollection<string> targetsToBuild, HostServices hostServices, BuildRequestDataFlags flags)
+            : this(targetsToBuild, hostServices, flags)
+        {
+            ErrorUtilities.VerifyThrowArgumentLength(projectFullPath, nameof(projectFullPath));
+            ErrorUtilities.VerifyThrowArgumentNull(globalProperties, nameof(globalProperties));
```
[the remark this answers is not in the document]
dfederm: No, the the `GraphBuildRequest` doesn't create the `ProjectGraph` just as the `BuildRequest` doesn't create the `ProjectInstance`. It just remains null and is lazily created. In the graph case, it's created in the `BuildManager` while in the non-graph case it's deferred all the way to the `RequestBuilder`.

src/Build/BackEnd/BuildManager/GraphBuildRequestData.cs
```diff
@@ -0,0 +1,193 @@
...
+            TargetNames = new List<string>(targetsToBuild);
+            Flags = flags;
+        }
+
+        /// <summary>
+        /// The requested project graph to build.
+        /// May be null.
+        /// </summary>
+        /// <value>The project graph.</value>
+        public ProjectGraph ProjectGraph { get; }
```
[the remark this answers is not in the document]
dfederm: I think generally the convention is: fields, ctors, props, methods. If this codebase does it: fields, props, ctors, methods, I can adjust.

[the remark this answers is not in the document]
dfederm: Actually `BuildRequestData` is: ctors, props (there are no fields or methods), so I'm just following that.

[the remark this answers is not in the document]
rainersigwald: Meshes with https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md, so . . . I guess I'll live, even though it's Obviously Wrong :)

src/Build/BackEnd/BuildManager/GraphBuildRequestData.cs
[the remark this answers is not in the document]
rainersigwald: How is the information persisted? Where in the just-constructed object returned from this ctor is `projectFullPath` stored?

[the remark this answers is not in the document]
dfederm: Uhhhh wat. Yea I think I messed this overload up. It should be just calling into the one that takes a `IEnumerable<ProjectGraphEntryPoint>`

src/Build/BackEnd/BuildManager/GraphBuildSubmission.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// </summary>
+        private readonly ManualResetEvent _completionEvent;
+
+        /// <summary>
+        /// The build results for each node.
+        /// </summary>
+        /// <remarks>
+        /// This collection may not contain results for all nodes until the submission is complete.
+        /// </remarks>
+        private readonly ConcurrentDictionary<ProjectGraphNode, BuildResult> _results = new ConcurrentDictionary<ProjectGraphNode, BuildResult>();
```
cdmihai: Could this be a GraphBuildResult instead?

dfederm: Lol, this field is never used. Probably a result of some refactoring and I left this hanging.

rainersigwald: Your other PR looks over sadly, shaking its head.

PR #3883, opened 2018-10-25
Fix MSBuildAssemblyVersion "Current"

rainersigwald:


src/Shared/Constants.cs
```diff
@@ -70,14 +70,14 @@ internal static class MSBuildConstants
...
         /// </summary>
-        internal static Version CurrentToolsVersionAsVersion = new Version(CurrentToolsVersion);
+        internal const string CurrentAssemblyVersion = "15.1.0.0";
 
         /// <summary>
-        /// The most current VSGeneralAssemblyVersion known to this version of MSBuild.
+        /// The current major MSBuild version for use in the automatic MSBuildAssemblyVersion property.
         /// </summary>
-        internal const string CurrentAssemblyVersion = "15.1.0.0";
+        internal const string CurrentVersionMajorMinor = "16.0";
```
jeffkl: There's also [MSBuildConstants.CurrentProductVersion](https://github.com/Microsoft/msbuild/blob/ba45c1284bbc91a60692b372062ed6b03b4aec36/src/Shared/Constants.cs#L85)

rainersigwald: Yeah, probably better to just restore that for continuity. I'll rename and recomment to match.

Issue #3889, opened 2018-10-29
[Placeholder]


cdmihai:

ladipro:
> When glob caching is enabled (either by setting MsBuildCacheFileEnumerations to 1 or programatically by using a shared EvaluationContext)

@cdmihai, in the current code `MsBuildCacheFileEnumerations` and shared `EvaluationContext` lead to two different cache implementations internally. The cache enabled by `MsBuildCacheFileEnumerations` has a comment saying:

```C#
// until Cloudbuild switches to EvaluationContext, we need to keep their dependence on global glob caching via an environment variable
```

Do you know if this is still in use?

cdmihai:
Sadly the environment flag [has become](https://github.com/search?q=MsBuildCacheFileEnumerations&type=code) more of a public API easily turn caching on without introducing evaluation context in user code. I think at this point the best we can do is unify the implementation and keep the two entry points.

PR #3890, opened 2018-10-29
Turn off caching distinct relative globs

cdmihai:
Because MSBuild globs preserve the fixed directory from the project file instead of returning fully qualified paths, it is harder than previously thought to cache distinct relative globs pointing to the same file, as the cached glob expansions would need to be rebased to multiple fixed directory parts.


cdmihai:
Issue to turn back optimization: #3889

PR #3892, opened 2018-10-30
OptProf data from 15.9 rel branch

rainersigwald:
Not sure this is necessary, but prepping it in case.


rainersigwald:
Turned out to not be necessary. If we update again or service 15.9, we should include this commit.

PR #3900, opened 2018-11-01
Remove LoadFromAssemblyName for MSBuild assemblies in the CoreCLRAssemblyLoader

### LINKED ISSUE #3352: Evaluating project file from .Net Core application throws StackOverflowException
### Steps to reproduce

I'm trying to evaluate project file from .Net Core application using `Microsoft.Build` packages. Target project uses `Microsoft.NET.Sdk`. The approach is similar to `Microsoft.Build.Locator` package, where `Microsoft.Build` dependencies are marked with `ExcludeAssets="runtime"`. Instead loading them from application directory, the application uses `AppDomain.CurrentDomain.AssemblyResolve` to load assemblies from SDK directory.

You can find sample over here: https://github.com/kovalikp/MSBuildCoreTest

### Expected  behavior

Being able to load project get list of targets from project file.

### Actual behavior

Application crashes on `Exception Unhandled: System.StackOverflowException`. Exception does not have stack trace or any other useful data.

### Environment data

OS info:
Windows 10
Version	10.0.17134 Build 17134

Microsoft Visual Studio Enterprise 2017 
Version 15.7.2
VisualStudio.15.Release/15.7.2+27703.2018

Project target framework:
netcoreapp2.0

NuGet package 
Microsoft.Build 15.7.179

dotnet 
sdk 2.1.201

JoeRobich:
From @vitek-karas,
> Handler for the Resolve event should never call LoadFromAssemblyName (or Assembly.Load) – that will lead to recursive loops. Resolve event handlers should only load assemblies from sources (files, memory streams).

Fixes #3352

### DIFF
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -73,29 +73,18 @@ namespace Microsoft.Build.Shared
             }
         }
 
-        private Assembly TryGetWellKnownAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
-        {
-            if (!_wellKnownAssemblyNames.Contains(assemblyName.Name))
-            {
-                return null;
-            }
-
-            assemblyName.Version = _currentAssemblyVersion;
-
-            return context.LoadFromAssemblyName(assemblyName);
-        }
-
         private Assembly TryResolveAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
         {
             lock (_guard)
             {
-                Assembly assembly = TryGetWellKnownAssembly(context, assemblyName);
-
-                if (assembly != null)
+                if (_wellKnownAssemblyNames.Contains(assemblyName.Name))
                 {
-                    return assembly;
+                    // Ensure we are attempting to load a matching version
+                    // of the Microsoft.Build.* assembly.
+                    assemblyName.Version = _currentAssemblyVersion;
                 }
 
+                Assembly assembly;
                 if (_namesToAssemblies.TryGetValue(assemblyName.FullName, out assembly))
                 {
                     return assembly;
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -69,70 +69,90 @@ namespace Microsoft.Build.Shared
                     return assembly;
                 }
 
-                return LoadAndCache(fullPath);
+                return LoadAndCache(AssemblyLoadContext.Default, fullPath);
             }
+        }
+
+        private Assembly TryGetWellKnownAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
+        {
+            if (!_wellKnownAssemblyNames.Contains(assemblyName.Name))
+            {
+                return null;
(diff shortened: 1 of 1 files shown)


src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -73,29 +73,18 @@ public Assembly LoadFromPath(string fullPath)
...
             {
-                Assembly assembly = TryGetWellKnownAssembly(context, assemblyName);
-
-                if (assembly != null)
+                if (_wellKnownAssemblyNames.Contains(assemblyName.Name))
                 {
-                    return assembly;
+                    // Ensure we are attempting to load a matching version
+                    // of the Microsoft.Build.* assembly.
+                    assemblyName.Version = _currentAssemblyVersion;
```
rainersigwald: Since we don't vary assembly version across builds/releases any more, I think we might have to do more here, like force the load to come from `folder-next-to-current-assembly` rather than doing the full lookup.

JoeRobich: So I should extract the candidate path searching logic and call it from `TryGetWellKnownAssembly` with the executing assembly's location? something like:

```csharp
        private Assembly TryGetWellKnownAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
        {
            if (!_wellKnownAssemblyNames.Contains(assemblyName.Name))
            {
                return null;
            }

            // Ensure we are attempting to load a matching version
            // of the Microsoft.Build.* assembly.
            assemblyName.Version = _currentAssemblyVersion;

            var searchPaths = new[] { Assembly.GetExecutingAssembly().Location };
            return TryResolveAssemblyFromPaths(context, assemblyName, searchPaths);
        }
```

rainersigwald: That sounds pretty good to me.

~I remember some funkiness around getting path-to-current-assembly in .NET Core, but I bet it's fixed now, that was a pre-1.0 memory. It _might_ be better to use `BuildEnvironmentHelper.Instance.CurrentMSBuildExePath`, but that might also accidentally trigger other loads~ nope, looks like that was all about targeting <netstandard1.3.

@tmeschter, can you dredge up any memories that might be relevant to this loader change?

tmeschter: I'm not sure how much of this code is necessary--I had to work around multiple bugs in the CoreCLR to get *anything* to work at the time, and I would assume they have been fixed now.

I can't remember now why I even needed special handling of the well-known assemblies. Since we're always using the default `AssemblyLoadContext` I would assume that all loads of `Microsoft.Build.dll`, etc, would resolve to the ones shipped with the product (regardless of version), so why would we ever need to handle those in the resolve assembly event handler? But probably I've just forgotten how `AssemblyLoadContext` works. :-)

The commit messages are your best change for understanding what I was thinking at the time.

rainersigwald:
This is good to go, right? Anyone know of a reason to wait further?

JoeRobich:
@rainersigwald I don't have any additional changes planned. =)

rainersigwald:
Thanks @JoeRobich!

PR #3931, opened 2018-11-14
Don't block explicitly referencing MSBuild 15.0

AndyGerlicher:
* Added a test and associated fixes for pointing `MSBUILD_EXE_PATH` to a Dev15 VS install path.
* Added some regression tests on scenarios that shouldn't be supported.

Note that after discussing with @rainersigwald we both agreed that other scenarios shouldn't work. The scenario here is using `Microsoft.Build.dll` version 16.0+ to evaluate, and potentially build, a project when you only have MSBuild 15 installed. The engine should be backwards compatible so using older targets _should_ not be an issue. But if we add custom logic to allow this to work easily, many scenarios are going to be broken. For example multiproc builds will call MSBuild.exe to spawn workers which will fail because the versions will mismatch. However, if the environment variable is explicitly set MSBuild should honor that even if there is unexpected behavior.

### DIFF
--- a/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
+++ b/src/Build.UnitTests/BuildEnvironmentHelper_Tests.cs
@@ -216,19 +216,30 @@ namespace Microsoft.Build.Engine.UnitTests
             }
         }
 
-        [Fact]
+        [Theory]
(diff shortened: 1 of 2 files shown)


AndyGerlicher:
@yash256 let me know if this works for you and we can get new packages pushed to myget.

PR #3949, opened 2018-11-21
Release configuration on CI

cdmihai:


src/Directory.Build.props
```diff
@@ -9,10 +12,11 @@
       <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
-  <!-- Import the repo root props -->
-  <Import Project="..\Directory.Build.props"/>
+  <PropertyGroup>
+      <ShouldUseIBCOptimization Condition="'$(BUILD_STAGINGDIRECTORY)' != '' and $(TargetFramework.StartsWith('net4')) and '$(BUILD_REASON)' != 'PullRequest'">true</ShouldUseIBCOptimization>
```
rainersigwald: I think this will break on the public pipeline when the build reason is CI. Does this make sense?
```suggestion
      <ShouldUseIBCOptimization Condition="'$(BUILD_STAGINGDIRECTORY)' != '' and $(TargetFramework.StartsWith('net4')) and !( '$(BUILD_REASON)' == 'PullRequest' or '$(BUILD_REASON)' == 'IndividualCI' )">true</ShouldUseIBCOptimization>
```

PR #3989, opened 2018-12-06
Distribute ResolveAssemblyReference cache

ccastanedaucf:
Currently, each instance of the ```SystemState``` cache copies _all_ ```FileState``` objects accessed and serializes them onto disk for each project. As pointed out in https://github.com/Microsoft/msbuild/issues/2015#issuecomment-443340515, this results in large amounts of identical data written to disk and deserialized into memory.

This adds some bookkeeping to each ```SystemState``` instance to ensure that ```FileState``` objects aren't duplicated from the process-wide cache if another instance is already responsible for serializing that object. This results in some additional IO on multi-proc builds since RAR execution order may change around and result in some cache misses, but it seems to be outweighed by the huge savings in memory and RAR times.

In both of these traces, this results in a nearly **_30% reduction in time spent on RAR_**

### WebLarge
rarcacheweblarge

### OrchardCore
rarcacheorchardcore

These are from ```/clp:PerformanceSummary /m``` so the task timings are cumulative, but this gives an idea of overall effect on build times.

**Master**
```
Target Performance Summary:
     6763 ms  Build                                    133 calls

Task Performance Summary:
    10663 ms  ResolveAssemblyReference                 132 calls
```

**Distributed cache**
```
Target Performance Summary:
     5513 ms  Build                                    133 calls

Task Performance Summary:
     7414 ms  ResolveAssemblyReference                 132 calls
```

### DIFF
--- a/src/Tasks/SystemState.cs
+++ b/src/Tasks/SystemState.cs
@@ -390,8 +390,12 @@ namespace Microsoft.Build.Tasks
             bool isInstanceFileStateUpToDate = isCachedInInstance && lastModified == cachedInstanceFileState.LastModified;
             bool isProcessFileStateUpToDate = isCachedInProcess && lastModified == cachedProcessFileState.LastModified;
 
+            // If the process-wide cache contains an up-to-date FileState, always use it
             if (isProcessFileStateUpToDate)
             {
+                // If a FileState already exists in this instance cache due to deserialization, remove it;
+                // another instance has taken responsibility for serialization, and keeping this would
+                // result in multiple instances serializing the same data to disk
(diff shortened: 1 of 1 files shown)


nguerrera:
Awesome

xoofx:
Cool, I just tried on my test case and it is reducing the cache files on the disk by a substantial margin, and good to see that ResolveAssemblyReference gets a significant benefit from this 👍

ccastanedaucf:
I definitely had to draw a bit of a flowchart when getting the logic in place

davkean:
This is great! I once took a stab at this a while back and failed dismally.

PR #3998, opened 2018-12-09
Actually test symbolic links

marcin-krystianc:


src/Tasks/NativeMethods.cs
```diff
@@ -808,6 +808,7 @@ internal static bool MakeHardLink(string newFileName, string exitingFileName, re
         // CreateSymbolicLink
         //------------------------------------------------------------------------------
         [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+        [return: MarshalAs(UnmanagedType.I1)]
```
marcin-krystianc: Native `CreateSymbolicLink` returns `BOOLEAN` which is a typedef for `BYTE`. Therefore, according to https://blogs.msdn.microsoft.com/jaredpar/2008/10/14/pinvoke-and-bool-or-should-i-say-bool/, `MarshalAs(UnmanagedType.I1)` should be used here.

src/Tasks/Copy.cs
```diff
@@ -255,6 +255,8 @@ private void LogDiagnostic(string message, params object[] messageArgs)
             else if (UseSymboliclinksIfPossible)
             {
                 TryCopyViaLink("Copy.SymbolicLinkComment", MessageImportance.Normal, sourceFileState, destinationFileState, ref destinationFileExists, out linkCreated, ref errorMessage, (source, destination, errMessage) => NativeMethods.MakeSymbolicLink(destination, source, ref errorMessage));
+                if (linkCreated && destinationFileState.Name == "foo:bar")
+                    throw new Exception("Link creation should not succeed");
```
marcin-krystianc: I cannot figure out why symbolic link creation works on CI servers for file `foo:bar`  (https://dnceng.visualstudio.com/public/_build/results?buildId=55243&view=ms.vss-test-web.test-result-details), although it is not supposed to. I cannot reproduce it locally (Win10), for me symbolic link creation always fails for file name containing colon.

Because symbolic link creation succeeds (although it should not), a test `DoNotRetryCopyNotSupportedException` fails after enabling symbolic links (https://dnceng.visualstudio.com/public/_build/results?buildId=55206&view=ms.vss-test-web.test-result-details&runId=1266334&resultId=104862).
Does someone have a clue how to explain it?

src/Tasks/NativeMethods.cs
cdmihai: Nit, optional: Can you update the other pinvokes that return bool, at least in this file?

src/Tasks/Copy.cs
```diff
@@ -305,7 +307,7 @@ private void TryCopyViaLink(string linkComment, MessageImportance messageImporta
             }
 
             linkCreated = createLink(sourceFileState.Name, destinationFileState.Name, errorMessage);
-
+
```
cdmihai: nit: spurious empty space :)

src/Tasks/Copy.cs
cdmihai: That is weird, .net core removed all path validity checks, but full framework should still have them. Do you have net472 on your machine? If not, can you get net472 and try again? Just to make sure no weird stuff happened in 472.

You can also try and debug the test to better understand what is going on. Does Copy.DoCopyWithRetries not throw at all?

src/Tasks/NativeMethods.cs
marcin-krystianc: As far as I can see, this was only a problem for `CreateSymbolicLink`, because it returns `BOOLEAN` (which is 1 byte only). Other pinvokes there, return `BOOL`  which is 4-byte type and that is correctly translated into bool.

src/Tasks/Copy.cs
marcin-krystianc: Yes I've net472 and when I locally run `DoNotRetryCopyNotSupportedException`, it passes.
I've attached a debugger and I can confirm that:
- symbolic link creation does not succeed (`NativeMethods.MakeSymbolicLink` returns false)
- following `File.Copy` operation (https://github.com/Microsoft/msbuild/blob/d13c0ebd8f7a1fe88fcc86253011a9688f958dc2/src/Tasks/Copy.cs#L269) throws `System.NotSupportedException`

The problem on CI server is that symbolic link succeeds. Because symbolic link is created, we don't call `File.Copy` and there is no `System.NotSupportedException`. 
So it is not rather .net framework version issue, more likely it is related to OS version. I'll add more checks to investigate.

cdmihai: @rainersigwald @ccastanedaucf 
Do you know of any recent windows / full framework changes that would result in the symlink `foo:bar` being accepted by windows / full framework?

marcin-krystianc: From what I can see, on CI server there is Windows 2012 used . I've tested it locally also on Windows 2012 and I also get a successful symbolic link creation for "foo:bar" (where it fails on Win10).
Interestingly, instead of file "foo:bar", a link file with name "foo" is created. Probably it has something to do with Alternate Data Streams in NTFS, where colon is used as a separator for alternate stream name.
Therefore I rekon we should simply delete `DoNotRetryCopyNotSupportedException` , because it relies on `NotSupportedException` which is framework & OS specific. 
What do you think ?

cdmihai: I think the original intent of the test was to assert no retries happen when the destination file has illegal characters. I've added a commit with more invalid characters, see how it goes :)

marcin-krystianc: Well, there is already a test called `ExitGracefullyOnInvalidPathCharacters`, which is testing the case with invalid characters and `ArgumentException`.
The intent of `DoNotRetryCopyNotSupportedException` is to test 'NotSupportedException', which I was able to reliably trigger by placing colon at the end of the file name ;-)

PR #3999, opened 2018-12-10
Expose GetFileHash & VerifyFileHash for .NET Core

iainnicol:
These tasks already exist in `Microsoft.Build.Tasks.Core.dll`, but were
not exposed with a `<UsingTask>`.  I presume this is an oversight.  In
contrast, these tasks are already exposed from
`Microsoft.Build.Tasks.v4.0`, with a `<UsingTask>`.

### DIFF
--- a/src/Tasks/Microsoft.Common.tasks
+++ b/src/Tasks/Microsoft.Common.tasks
@@ -40,7 +40,6 @@
 
     <UsingTask TaskName="Microsoft.Build.Tasks.GenerateTrustInfo"                     AssemblyName="Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="'$(MSBuildAssemblyVersion)' == ''" />
     <UsingTask TaskName="Microsoft.Build.Tasks.GetAssemblyIdentity"                   AssemblyName="Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="'$(MSBuildAssemblyVersion)' == ''" />
-    <UsingTask TaskName="Microsoft.Build.Tasks.GetFileHash"                           AssemblyName="Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="'$(MSBuildAssemblyVersion)' == ''" />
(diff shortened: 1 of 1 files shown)


rainersigwald:
Great catch! Those tasks are new (added in #3398) and _not_ exposed from the v4.0, so _only_ the new lines you're adding make sense.

Can you please **delete** the existing lines as part of this PR as well?

iainnicol:
Yes, I like the bleeding edge :).

Thanks for the explanation.  I had assumed that .NET Framework MSBuild uses ``Microsoft.Build.Tasks.v4.0.dll``, and .NET Core MSBuild uses ``Microsoft.Build.Tasks.Core.dll``.  But I now see that MSBuild, for both Framework and Core, ships ``Microsoft.Build.Tasks.Core``. Also, ``Microsoft.Build.Tasks.v4.0.dll`` is shipped separately, in the .NET Framework. Yeah?

Existing lines deleted.

japj:
ah, this explains why I couldn't use these tasks in VS2017 15.8. When will this be shipped (and does it include an update xsd by any chance)?

rainersigwald:
@japj The fix will be in 16.0 (it should be in Preview 2 released last week).

As a workaround for older versions, you can include explicit `UsingTask`s before your usage of the task. If you include a condition on `$(MSBuildVersion)` you can keep it compatible with fixed MSBuilds.

```xml
    <UsingTask TaskName="Microsoft.Build.Tasks.GetFileHash"                           AssemblyName="Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="'$(MSBuildVersion)' < '16.0.360'" />
    <UsingTask TaskName="Microsoft.Build.Tasks.VerifyFileHash"                        AssemblyName="Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="''$(MSBuildVersion)' < '16.0.360'" />

<Target Name="Whatever">
 <GetFileHash ... />
</Target>
```

> and does it include an update xsd by any chance

It does not, but should. Filed Microsoft/msbuild#4103.

japj:
@rainersigwald great! Thanks for the info and updating the documentation!

hickford:
@rainersigwald After fixing some typos, the workaround worked for me:

```xml
<UsingTask TaskName="Microsoft.Build.Tasks.GetFileHash" AssemblyName="Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="'$(MSBuildVersion)' &lt; '16.0.360'" />
<UsingTask TaskName="Microsoft.Build.Tasks.VerifyFileHash" AssemblyName="Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" Condition="'$(MSBuildVersion)' &lt; '16.0.360'" />
```

PR #4008, opened 2018-12-13
LOC CHECKIN | Microsoft/msbuild vs16.0-preview2 | 20181213

v-zbsail:


cdmihai:
In the current circumstances (release is not locked down) retargetting to master looks good.

src/MSBuild/Resources/xlf/Strings.de.xlf
```diff
@@ -126,16 +126,16 @@ Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.
...
+        <target state="translated">  -isolateProjects[:True|False]
+                     Führt dazu, dass MSBuild jedes Projekt isoliert erstellt.
 
-                     This is a more restrictive mode of MSBuild as it requires
-                     that the project graph be statically discoverable at
-                     evaluation time, but can improve scheduling and reduce
-                     memory overhead when building a large set of projects.
-                     (Short form: -isolate)
+                     Dies ist ein restriktiverer Modus von MSBuild, da er erfordert, dass
+                     das Projektdiagramm zur Auswertungszeit statisch ermittelbar ist. Er kann jedoch
```
akoeplinger: `Projectdiagramm` is a weird translation for `project graph`, we're not talking about a literal drawing here but the abstract concept of a graph as in https://en.wikipedia.org/wiki/Graph_(discrete_mathematics).

I think we can use `Projektgraph` instead since the word has the same meaning in German: https://de.wikipedia.org/wiki/Graph_(Graphentheorie)

src/MSBuild/Resources/xlf/Strings.de.xlf
```diff
@@ -155,17 +155,17 @@ Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.
...
+                     Führt dazu, dass MSBuild ein Projektdiagramm konstruiert und erstellt.
 
-                     Constructing a graph involves identifying project
-                     references to form dependencies. Building that graph
-                     involves attempting to build project references prior
-                     to the projects that reference them, differing from
-                     traditional MSBuild scheduling.
-                     (Short form: -graph)
+                     Das Konstruieren eines Diagramms beinhaltet das Identifizieren von
+                     Projektverweisen auf Formularabhängigkeiten. Beim Erstellen des Diagramms
```
akoeplinger: @v-zbsail `Formularabhängigkeiten` is wrong here as a translation for `to form dependencies`, it's talking about _creating_ dependencies and not about a form in the "sheet of paper" sense. And same issue for Diagramm vs Graph that I mentioned in the other comment.

Better translation would be:
> Das Konstruieren eines Graphs beinhaltet das Identifizieren von Projektverweisen um Abhängigkeiten zu erstellen.

akoeplinger: @janbaltus perfect, thank you very much!

PR #4027, opened 2018-12-21
LOC CHECKIN | Microsoft/msbuild vs16.0-preview2 | 20181221 | DEU fix

v-zbsail:


rainersigwald:
Merged with master which picked up some additional changes but the new file looks good to me. After this goes in I'll update the preview2 branch with the current master, so this problem doesn't happen again.

PR #4041, opened 2019-01-03
Get Roslyn from 15.0 folder in amd64 MSBuild, too

rainersigwald:
This ports the change from 7b279c8ad to apply to 64-bit MSBuild as well
as 32-bit.


rainersigwald:
@azchohfi confirmed offline that making this change manually in an installed VS worked around the problem, so let's go forward with this.

PR #4055, opened 2019-01-09
 Ignore case when hashing CoreCompile inputs

### LINKED ISSUE #4023: CoreCompile incrementality breaks on parallel builds with inconsistent path casing on case-insensitive file systems
Discovered this while building a large internal repo and seeing ```CoreCompile``` execute at seemingly random on incremental parallel builds with zero changes, making build times range from anywhere between 20 seconds and 3 minutes.

### Steps to reproduce

Minimal repro: [compile-repro.zip](https://github.com/Microsoft/msbuild/files/2700813/compile-repro.zip)

Before the ```CoreCompile``` task is executed, all of the inputs are hashed via the ```Hash``` task and stored in a ```CoreCompileInputs.cache``` file. On an incremental build, the current hash is compared to the cached hash, and if matching, compilation is skipped.

However if inputs are referenced with inconsistent casing, during a multi-proc build one of these casings will "win" over the others and be used for the remainder of the build graph. The ```Hash``` task will then produce a different hash although nothing has changed. This triggers recompilation for all dependees using a different casing whenever a new casing "wins".

Consider the following build graph:
```
  ClassLibrary1
   /    |    \
Foo     |     Bar
   \    |    /
       Baz
```
And the paths specified in each project's items:

```ClassLibrary1.csproj```:
```csproj
<ProjectReference Include="..\baz\Baz.csproj" />
<ProjectReference Include="..\Foo\Foo.csproj" />
<ProjectReference Include="..\Bar\Bar.csproj" />
```

```Foo.csproj``` and ```Bar.csproj```:
```csproj
<ProjectReference Include="..\Baz\Baz.csproj" />
```

If ```-> Foo -> Baz``` or ```-> Bar -> Baz``` build first, ```Baz\bin\Debug\net472\Baz.dll``` will be the path passed to ```CoreCompile```.

However if ```-> Baz``` builds first,  ```baz\bin\Debug\net472\Baz.dll``` will be used, generating a different hash and triggering recompilation, even though this resolves to the same path in a case-insensitive file system.

The included repro isn't large enough for there to be enough variance for this to pop up with ```/m``` (```-> Baz``` will always finish first), but you can simulate this easily.

Command line
```
msbuild /t:restore ClassLibrary1\ClassLibrary1.csproj
msbuild ClassLibrary1\ClassLibrary1.csproj
msbuild Foo\Foo.csproj
msbuild Bar\Bar.csproj
msbuild ClassLibrary1\ClassLibrary1.csproj
```

And watch as each of these builds recompile due to the different casings of ```Baz``` passed to ```Hash```. The last command will actually end up recompiling the entire graph, which shows how bad the worst case scenario is in a large repo.

On Linux where the file system is case-sensitive, this scenario behaves as expected. The build gives a warning that ```baz/Baz.csproj``` does not exist only for the project with the incorrect spelling and does not have a cascading effect on other projects in the graph. I'm guessing a ```FileExists()``` call catches this early.

### Solutions
Now the question is _where_ some change needs to be made. In general MSBuild tends to treat paths as case-insensitive when possible, such as cached directories and file exists checks. So the simplest solution may be adding a case-insensitive flag to the ```Hash``` task. Or syncing with the actual casing on-disk earlier in the build, although that seems potentially expensive.

### Expected  behavior
On case-insensitive file systems, multi-proc incremental builds should not execute ```CoreCompile``` again on unchanged projects.

### Actual behavior
On case-insensitive file systems, multi-proc incremental builds may execute ```CoreCompile``` on unchanged projects due to project build order and inconsistent path casing.

### Environment data
`msbuild /version` output: 16.0.256.5957

OS info:
Windows 10
Ubuntu 16.04

ccastanedaucf:
Fixes #4023 

Seems safe to always make hashing before ```CoreCompile``` case-insensitive regardless of file system. On a case-insensitive file system, this fixes the tagged issue. On a case-sensitive file system, adding/removing inputs that only differ by casing will still result in a different hash since we're hashing more/less characters, and breaking it seems like a very rare case.

### DIFF
--- a/src/Tasks/Hash.cs
+++ b/src/Tasks/Hash.cs
@@ -54,11 +54,8 @@ namespace Microsoft.Build.Tasks
                     {
                         foreach (var item in ItemsToHash)
                         {
-                            foreach (char c in item.ItemSpec)
-                            {
-                                stringBuilder.Append(IgnoreCase ? char.ToLower(c) : c);
-                            }
-
+                            string itemSpec = item.ItemSpec;
+                            stringBuilder.Append(IgnoreCase ? itemSpec.ToUpperInvariant() : itemSpec);
                             stringBuilder.Append(ItemSeparatorCharacter);
                         }
 
--- a/src/Tasks/Hash.cs
+++ b/src/Tasks/Hash.cs
@@ -27,6 +27,11 @@ namespace Microsoft.Build.Tasks
         public ITaskItem[] ItemsToHash { get; set; }
 
         /// <summary>
+        /// When true, will generate a case-insensitive hash.
+        /// </summary>
+        public bool IgnoreCase { get; set; }
+
+        /// <summary>
         /// Hash of the ItemsToHash ItemSpec.
         /// </summary>
         [Output]
@@ -49,7 +54,11 @@ namespace Microsoft.Build.Tasks
                     {
                         foreach (var item in ItemsToHash)
                         {
-                            stringBuilder.Append(item.ItemSpec);
+                            foreach (char c in item.ItemSpec)
+                            {
+                                stringBuilder.Append(IgnoreCase ? char.ToLower(c) : c);
+                            }
+
                             stringBuilder.Append(ItemSeparatorCharacter);
                         }
 
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -3429,7 +3429,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <CoreCompileCache Include="$(DefineConstants)" />
     </ItemGroup>
 
-    <Hash ItemsToHash="@(CoreCompileCache)" IgnoreCase="true">
+    <Hash 
+      ItemsToHash="@(CoreCompileCache)" 
+      IgnoreCase="$([MSBuild]::ValueOrDefault(`$(CoreCompileCacheIgnoreCase)`, `true`))">
       <Output TaskParameter="HashResult" PropertyName="CoreCompileDependencyHash" />
     </Hash>
 
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -3429,7 +3429,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <CoreCompileCache Include="$(DefineConstants)" />
     </ItemGroup>
 
-    <Hash ItemsToHash="@(CoreCompileCache)">
+    <Hash ItemsToHash="@(CoreCompileCache)" IgnoreCase="true">
       <Output TaskParameter="HashResult" PropertyName="CoreCompileDependencyHash" />
     </Hash>
 
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -703,6 +703,7 @@ namespace Microsoft.Build.Tasks
         public Hash() { }
         [Microsoft.Build.Framework.OutputAttribute]
         public string HashResult { get { throw null; } set { } }
+        public bool IgnoreCase { get { throw null; } set { } }
         [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem[] ItemsToHash { get { throw null; } set { } }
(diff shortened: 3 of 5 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -3429,7 +3429,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <CoreCompileCache Include="$(DefineConstants)" />
     </ItemGroup>
 
-    <Hash ItemsToHash="@(CoreCompileCache)">
+    <Hash ItemsToHash="@(CoreCompileCache)" IgnoreCase="true">
```
jeffkl: Make this use a property so users can use an escape hatch if necessary
```xml
<Hash 
  ItemsToHash="@(CoreCompileCache)" 
  IgnoreCase="$([MSBuild]::ValueOrDefault(`$(CoreCompileCacheIgnoreCase)`, `true`))">
```

src/Tasks/Hash.cs
```diff
@@ -49,7 +54,11 @@ public override bool Execute()
                     {
                         foreach (var item in ItemsToHash)
                         {
-                            stringBuilder.Append(item.ItemSpec);
+                            foreach (char c in item.ItemSpec)
+                            {
+                                stringBuilder.Append(IgnoreCase ? char.ToLower(c) : c);
```
jeffkl: Should this use `ToLowerInvariant()` instead?

src/Tasks/Hash.cs
```diff
@@ -49,7 +54,11 @@ public override bool Execute()
                     {
                         foreach (var item in ItemsToHash)
                         {
-                            stringBuilder.Append(item.ItemSpec);
+                            foreach (char c in item.ItemSpec)
```
jeffkl: If `IgnoreCase` is `false`, would it be faster to just have the `StringBuilder` append `item.ItemSpec` instead of each character since it uses pointers and strcpy?
http://index/?query=stringbuilder.app&rightProject=mscorlib&file=system%5Ctext%5Cstringbuilder.cs&line=658

src/Tasks/Hash.cs
jeffkl: Should we be worried about the perf implications of this?

AndyGerlicher: Also, use `ToUpperInvariant()`. I believe some characters don't convert correctly to lower so ToUpper is preferred in cases like this.

src/Tasks/Hash.cs
AndyGerlicher: I tried various methods:

Method | Mean | Error | StdDev | Gen 0 | Gen 1 | Allocated
-- | -- | -- | -- | -- | -- | --
StringBuilderAppend | 12.60 us | 0.0173 us | 0.0144 us | 15.6097 | 2.2278 | 96.45 KB
StringBuilderAppendIgnoreCase | 77.81 us | 0.5760 us | 0.5388 us | 24.9023 | 3.5400 | 154.33 KB
StringBuilderAppendEachChar | 59.33 us | 1.0765 us | 1.0070 us | 15.5640 | 2.1973 | 96.45 KB
StringBuilderAppendEachCharIgnoreCase | 506.68 us | 0.9487 us | 0.8410 us | 14.6484 | 1.9531 | 96.45 KB


IgnoreCase in this is just calling `ToUpperInvariant()` on the string or the character. I did a *.cs list in Microsoft.Build project for the input and pre-calculated the size. I think we should just do `Append()` when false and `ToUpperInvariant()` on the string when true, even though it uses a bit more memory.

ccastanedaucf:
After switching to ```ToUpperVariant()``` on the string, I ran a perf trace on building OrchardCore and performance is identical (left is this PR, right is master)

hash

src/Tasks/Hash.cs
cdmihai: Why ToUpper? The data is paths, which tend to be lowercase, so doing ToLower would involve fewer case changes.

AndyGerlicher: Perf is about the same. I've always used ToUpper for things like this or comparison. As I understand it (but have never directly experienced) there are languages where it is better to use ToUpper. There used to be an FxCop or StyleCop rule that would suggest always using ToUpper I believe.

Method | Mean | Error | StdDev | Gen 0 | Gen 1 | Allocated
-- | -- | -- | -- | -- | -- | --
StringBuilderAppendIgnoreCaseToUpper | 77.68 us | 0.2029 us | 0.1898 us | 24.9023 | 3.5400 | 154.33 KB
StringBuilderAppendIgnoreCaseToLower | 76.91 us | 0.1107 us | 0.0982 us | 24.9023 | 3.5400 | 154.33 KB

PR #4059, opened 2019-01-10
Fix NativeMethodsShared.GetLastWriteFileTimeUtc when the path is a

radical:
.. directory.

`GetLastWriteFileTimeUtc` depends on the incorrect assumption[1] that
`File.GetLastWriteTimeUtc` will return `DateTime.FromFileTimeUtc(0)` if
the path is a directory.

The same thing applies for the `Directory` equivalent of the same
method.

This breaks File existence checks, for example in
`ResolveAssemblyReference` when we have a reference like:

`<Reference Include="FooBar" />`

.. where `FooBar` is a directory in the current(project) directory.
ResolveAssemblyReference ends up triyng to get assembly information from
the file, because it thinks that `FooBar` is a file and thus an assembly
path.

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -838,11 +838,16 @@ namespace Microsoft.Build.Shared
                 return success;
             }
 
-            DateTime lastWriteTime = Directory.GetLastWriteTimeUtc(fullPath);
-            bool directoryExists = lastWriteTime != MinFileDate;
-
-            fileModifiedTimeUtc = directoryExists ? lastWriteTime : DateTime.MinValue;
-            return directoryExists;
+            if (Directory.Exists(fullPath))
+            {
+                fileModifiedTimeUtc = Directory.GetLastWriteTimeUtc(fullPath);
+                return true;
+            }
+            else
+            {
+                fileModifiedTimeUtc = DateTime.MinValue;
+                return false;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ccastanedaucf:
I remember this extra IO call added somewhere around an extra second on Linux vs Windows on a ~15 second build, but I guess there's no way to avoid hitting the disk twice. IO in RAR has been reduced since then (which makes up the majority of these calls), so that difference could be smaller now.

radical:
I guess [this](https://github.com/microsoft/msbuild/blob/master/src/Shared/NativeMethodsShared.cs#L823) is similarly assuming that `GetFileAttributesEx` will return `false` if the first is a directory, which is not true (https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfileattributesexa), so it will need a similar check. Am I correct?

radical:
Maybe the check should be changed to match https://github.com/microsoft/msbuild/blob/master/src/Shared/NativeMethodsShared.cs#L1596 .

ccastanedaucf:
I'm confused on that last point, since it looks like the first method performs the inverse of that FileExists condition then sets success to false if it's not a directory.

radical:
@ccastanedaucf oh, sorry, you are correct. I mixed that up. What should [this](https://github.com/microsoft/msbuild/blob/master/src/Shared/NativeMethodsShared.cs#L976) be changed to? Looking at the file attributes on [windows](https://docs.microsoft.com/en-us/windows/desktop/fileio/file-attribute-constants), IIUC, there are multiple that might indicate a file. So, maybe we should just check that it is not a `FILE_ATTRIBUTE_DIRECTORY`?

ccastanedaucf:
I think just adding ```&& (data.fileAttributes & NativeMethodsShared.FILE_ATTRIBUTE_DIRECTORY) == 0``` to the success condition would be enough there, and hopefully nothing is depending on that behavior.

radical:
Thanks for the reviews!

PR #4064, opened 2019-01-11
LOC CHECKIN | Microsoft/msbuild vs16.0-preview2 | 20190111

v-zbsail:


rainersigwald:
Merging to master since I don't think there's a strong reason to try to take this to p2 instead of waiting for p3.

PR #4067, opened 2019-01-12
Add DirectoryCreate to ensure the target directory exists

### LINKED ISSUE #4033: WriteLinesToFile should create the directory if it doesn't exist
The `WriteLinesToFile` task doesn't create the directory for the file if it doesn't exist.

It should call `Directory.CreateDirectory(Path.GetDirectoryName(File))` to help with the pit of success. Too much ceremony with requiring authors to create the directory themselves.

CodeTherapist:
Closes #4033

### DIFF
--- a/src/Tasks.UnitTests/ReadLinesFromFile_Tests.cs
+++ b/src/Tasks.UnitTests/ReadLinesFromFile_Tests.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.IO;
-using System.Threading;
 #if FEATURE_SECURITY_PERMISSIONS
 using System.Security.AccessControl;
 #endif
@@ -17,7 +16,7 @@ using Xunit;
 
 namespace Microsoft.Build.UnitTests
 {
-    sealed public class ReadLinesFromFile_Tests
+    public sealed class ReadLinesFromFile_Tests
     {
         /// <summary>
         /// Write one line, read one line.
@@ -25,21 +24,25 @@ namespace Microsoft.Build.UnitTests
         [Fact]
         public void Basic()
         {
-            string file = FileUtilities.GetTemporaryFile();
+            var file = FileUtilities.GetTemporaryFile();
             try
             {
                 // Start with a missing file.
                 File.Delete(file);
 
                 // Append one line to the file.
-                WriteLinesToFile a = new WriteLinesToFile();
-                a.File = new TaskItem(file);
-                a.Lines = new ITaskItem[] { new TaskItem("Line1") };
+                var a = new WriteLinesToFile
+                {
+                    File = new TaskItem(file),
+                    Lines = new ITaskItem[] { new TaskItem("Line1") }
+                };
                 Assert.True(a.Execute());
 
                 // Read the line from the file.
-                ReadLinesFromFile r = new ReadLinesFromFile();
-                r.File = new TaskItem(file);
+                var r = new ReadLinesFromFile
+                {
+                    File = new TaskItem(file)
+                };
                 Assert.True(r.Execute());
 
                 Assert.Single(r.Lines);
@@ -70,21 +73,25 @@ namespace Microsoft.Build.UnitTests
         [Fact]
         public void Escaping()
         {
-            string file = FileUtilities.GetTemporaryFile();
+            var file = FileUtilities.GetTemporaryFile();
             try
             {
                 // Start with a missing file.
                 File.Delete(file);
 
                 // Append one line to the file.
-                WriteLinesToFile a = new WriteLinesToFile();
-                a.File = new TaskItem(file);
-                a.Lines = new ITaskItem[] { new TaskItem("Line1_%253b_") };
+                var a = new WriteLinesToFile
+                {
+                    File = new TaskItem(file),
+                    Lines = new ITaskItem[] { new TaskItem("Line1_%253b_") }
+                };
                 Assert.True(a.Execute());
 
                 // Read the line from the file.
-                ReadLinesFromFile r = new ReadLinesFromFile();
-                r.File = new TaskItem(file);
+                var r = new ReadLinesFromFile
+                {
+                    File = new TaskItem(file)
+                };
                 Assert.True(r.Execute());
 
                 Assert.Single(r.Lines);
@@ -113,21 +120,25 @@ namespace Microsoft.Build.UnitTests
         [Fact]
         public void ANSINonASCII()
         {
-            string file = FileUtilities.GetTemporaryFile();
+            var file = FileUtilities.GetTemporaryFile();
             try
             {
                 // Start with a missing file.
                 File.Delete(file);
 
                 // Append one line to the file.
-                WriteLinesToFile a = new WriteLinesToFile();
-                a.File = new TaskItem(file);
-                a.Lines = new ITaskItem[] { new TaskItem("My special character is \u00C3") };
+                var a = new WriteLinesToFile
+                {
+                    File = new TaskItem(file),
+                    Lines = new ITaskItem[] { new TaskItem("My special character is \u00C3") }
+                };
                 Assert.True(a.Execute());
 
                 // Read the line from the file.
-                ReadLinesFromFile r = new ReadLinesFromFile();
-                r.File = new TaskItem(file);
+                var r = new ReadLinesFromFile
+                {
+                    File = new TaskItem(file)
+                };
                 Assert.True(r.Execute());
 
                 Assert.Single(r.Lines);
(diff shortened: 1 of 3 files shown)


src/Tasks/FileIO/WriteLinesToFile.cs
```diff
@@ -15,7 +16,7 @@ namespace Microsoft.Build.Tasks
     public class WriteLinesToFile : TaskExtension
     {
         // Default encoding taken from System.IO.WriteAllText()
-        private static readonly Encoding s_defaultEncoding = new UTF8Encoding(false, true);
+        private static readonly Encoding s_defaultEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
```
CodeTherapist: Added the argument names to help understand what `false` and `true` means.

src/Tasks/FileIO/WriteLinesToFile.cs
```diff
@@ -82,27 +83,31 @@ public override bool Execute()
...
+                                // just delete the file to leave everything tidy.
+                                System.IO.File.Delete(File.ItemSpec);
+                            }
                         }
                         else
                         {
-                            string contentsAsString = null;
-
+                            CreateTargetDirectory();
+                            string contentsAsString = buffer.ToString();
```
CodeTherapist: Removed the if below, since `buffer.ToString()` is always called in this `else` branch.

src/Tasks/FileIO/WriteLinesToFile.cs
```diff
@@ -116,17 +121,17 @@ public override bool Execute()
...
-                                contentsAsString = buffer.ToString();
-                            }
 
                             System.IO.File.WriteAllText(File.ItemSpec, contentsAsString, encoding);
                         }
                     }
                     else
                     {
-                        System.IO.File.AppendAllText(File.ItemSpec, buffer.ToString(), encoding);
+                        if (buffer.Length > 0)
```
CodeTherapist: Minimal performance improvement: when the buffer is empty, calling `File.AppendAllText` is a useless IO operation.

src/Tasks/FileIO/WriteLinesToFile.cs
```diff
@@ -82,27 +83,31 @@ public override bool Execute()
...
+                                // if overwrite==true, and there are no lines to write,
+                                // just delete the file to leave everything tidy.
+                                System.IO.File.Delete(File.ItemSpec);
+                            }
                         }
                         else
                         {
-                            string contentsAsString = null;
-
+                            CreateTargetDirectory();
```
Therzok: Since `GetTargetDirectory()` is called in both branches, maybe use a local variable before `if (buffer.Length == 0)` which stores `Path.GetDirectoryName(File.ItemSpec)` and use that directly.

CodeTherapist: Not sure I follow correctly. When I place that before `(buffer.Length == 0)` then the `else` branch of the `if (Overwrite)` would not work. I check the directory also when it is appending or shouldn't I do this?

Therzok: See https://github.com/Microsoft/msbuild/pull/4067#discussion_r247311672

CodeTherapist:
@Therzok I'm not sure if the build is failling because of my changes. It looks like that `File.ItemSpec` is empty - shouldn't we check that there is a valid path?

Therzok:
Looks much cleaner to me now!

Looking at master, it seems that pr builds are green, so they are caused by this change. Although, I'll defer that to @rainersigwald, as I don't have that much knowledge there.

CodeTherapist:
@Therzok I fixed it!
I found out that `File.ItemSpec` could be the file name only - without any explicit path information. For this case the `Path.GetDirectoryName()` returns (correctly) an empty string which is an invalid argument on `Directory.CreateDirectory()`.

src/Tasks/FileIO/WriteLinesToFile.cs
```diff
@@ -82,27 +83,32 @@ public override bool Execute()
 
                 try
                 {
+
+                    var directoryPath = Path.IsPathRooted(File.ItemSpec) ? Path.GetDirectoryName(File.ItemSpec) : Path.GetDirectoryName(Path.GetFullPath(File.ItemSpec));
```
cdmihai: - use FileUtilities.NormalizePath instead of Path.GetFullPath
- since you may end up calculating the full path here, might as well store it and reuse it in the places where File.ItemSpec gets sent to IO calls. That way the IO calls would avoid doing their own GetFullPath.

cdmihai:
Adding a test or two to ensure the directory gets appended on both branches would be cool :)
`Microsoft.Build.UnitTests.ReadLinesFromFile_Tests.WriteLinesWriteOnlyWhenDifferentTest` is the only test I found for this task. `TestEnvironment` is a nice utility for cleaning up test state.

src/Tasks/FileIO/WriteLinesToFile.cs
CodeTherapist: Ok - I will do that :)

CodeTherapist:
@cdmihai I will add some unit tests as well.

src/Tasks.UnitTests/ReadLinesFromFile_Tests.cs
```diff
@@ -17,29 +16,33 @@
 
 namespace Microsoft.Build.UnitTests
 {
-    sealed public class ReadLinesFromFile_Tests
+    public sealed class ReadLinesFromFile_Tests
```
rainersigwald: In the future, please don't mix this kind of style change with real code changes. It makes the PR much harder to review. In this repo, you're welcome to make these changes, but please put them in individual commits, so the "real" changes can be easily identified.

src/Tasks.UnitTests/WriteLinesToFile_Tests.cs
```diff
@@ -0,0 +1,216 @@
...
+            finally
+            {
+                File.Delete(file);
+            }
+        }
+
+        /// <summary>
+        /// Should create directory structure when target <see cref="WriteLinesToFile.File"/> does not exist.
+        /// </summary>
+        [Fact(Skip = "skip")]
```
rainersigwald: Why are you skipping all of the new tests?

src/Tasks.UnitTests/WriteLinesToFile_Tests.cs
```diff
@@ -0,0 +1,216 @@
...
+        public void WriteLinesToFileDoesCreateDirectory()
+        {
+            using (var testEnv = TestEnvironment.Create())
+            {
+                var directory = testEnv.CreateFolder(folderPath: null, createFolder: false);
+                var file = Path.Combine(directory.Path, $"{Guid.NewGuid().ToString("N")}.tmp");
+
+                var WriteLinesToFile = new WriteLinesToFile
+                {
+                    BuildEngine = new MockEngine(),
```
rainersigwald: Can you pass a test logger to this engine for improved diagnosability when things go wrong? 123e49e9bb78559ee28c9ad87be477aaa214dbb3 is an example of the easy way to do this in this codebase--once you have `_output`, you can just pass it to the `MockEngine` constructor.

src/Tasks.UnitTests/WriteLinesToFile_Tests.cs
CodeTherapist: Uh, good catch, that wasn't my intension - it doesn't make sense skipping them. Will change that.

src/Tasks.UnitTests/ReadLinesFromFile_Tests.cs
CodeTherapist: Hm.. not sure anymore if I did that as real action. It could come from hitting `Ctrl+K, Ctrl+D` (Format Document) and I expected that the solution has somewhere a editorconfig that keeps those thing align. Isn't that the case?

rainersigwald: Yeah, it's probably the case that it came from a format-document command. We haven't done a cleanup of the repo to define and match our standards in . . . a very long time.

src/Tasks.UnitTests/WriteLinesToFile_Tests.cs
CodeTherapist: Ok, got it - makes sense.

src/Tasks.UnitTests/ReadLinesFromFile_Tests.cs
CodeTherapist: It would be nice to have that, so that anyone like me hitting this shortcut (which is fairly automated on my fingers) will not get into troubles ;)

PR #4078, opened 2019-01-16
Add DesignerRuntimeImplementationProjectOutputGroup

nguerrera:
Contributes to https://github.com/dotnet/sdk/issues/2707

This adds a stub target for the new output group to common targets, which the SDK will override.

@davkean Is this the correct pattern to follow here?

FYI, @chabiss

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -6010,7 +6010,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <!--
     ============================================================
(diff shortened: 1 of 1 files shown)


davkean:
I'm not an expert on this at all, but should it be included in AllProjectOutputGroups?

rainersigwald:
> should it be included in AllProjectOutputGroups?

I lean toward "no, because it's not an output of the project in a normal build" but that opinion is _very_ lightly held.

nguerrera:
https://github.com/dotnet/sdk/pull/2895 is up with the concrete implementation. I will wait for that to be reviewed before merging this.

PR #4079, opened 2019-01-16
Perf: Reduce string.Split(), .Trim() params char[] and params string[] allocations

erikma:
Reduce useless gen0 collections on implicit params char[] and params string[] for string.Split() and string.Trim() across the codebase by using a common set of static arrays.
Recommended by: https://blog.marcgravell.com/2013/11/allocaction-allocation-allocation.html


cdmihai:
This might the thing causing the compilation errors: https://github.com/Microsoft/msbuild/blob/6716cd302418626df38c9003ea3acc6a32e3f044/src/Shared/Constants.cs#L8-L13

erikma:
Thanks, Mihai, fixed in 4th commit.

src/Tasks/ResolveSDKReference.cs
```diff
@@ -61,11 +61,6 @@ public class ResolveSDKReference : TaskExtension
         /// </summary>
         private const string SDKPlatformVersion = "PlatformVersion";
 
-        /// <summary>
-        /// Split char for strings
-        /// </summary>
-        private static readonly char[] s_semicolonSplitChar = { ';' };
```
rainersigwald: Is there a distinction between this, where you removed the private name, versus others where you kept the private name but pointed it to the `MSBuildConstants` one?

erikma: Typically I left the existing name alone when its name has more semantic meaning locally or a good doc comment. Example from XMake.cs where I left the alias because of the comment:

        /// <summary>
        /// This is a set of wildcard chars which can cause a file extension to be invalid 
        /// </summary>
        private static readonly char[] s_wildcards = MSBuildConstants.WildcardChars;

Example from ToolsetConfigurationREader.cs:

        private static readonly char[] s_separatorForExtensionsPathSearchPaths = MSBuildConstants.SemicolonChar;

danmosemsft:
In .NET Core 2.0+ there is a `String.Split` overload that takes a `char` but it was not backported.

src/Tasks/ResolveSDKReference.cs
rainersigwald: IIRC I agree with all of your decisions on that axis 👍

rainersigwald:
I don't think it's worth the extra complexity of splitting this to take advantage of the core single-char overload. Let's do this, then if that measures hot we can fix it.

rainersigwald:
Baffled by cla-bot; going to close/reopen to try to poke it.

danmosemsft:
> I don't think it's worth the extra complexity of splitting this to take advantage of the core single-char overload. Let's do this, then if that measures hot we can fix it.

Yep agreed, just mentioning it for the sake of anyone observing. 😄

Issue #308, opened 2015-10-21
GenerateResource does not fail when converting binary .resx resources to .txt resources


cdmihai:
On the full framework, if a .resx data element points to a path and there's a type specified to read that file, the conversion of the .resx resource to a .txt resource is impossible and the tasks logs an error (since .txt resources cannot have binary values).

However, the .net core behaviour is to write the file path to the .txt.

This happens because the full framework uses a special xml reader, ResXResourceReader , that reads a path in the .resx file directly to its specified .net object. However this special xml reader is not available in .net core, so a normal xml reader is used instead, which just reads the text file name.

Tests that capture failure:
- GenerateResource_Tests.UnsupportedTextType, InvalidReference, FailedResourceReader

Other implications
- .resx binary resources with specified .net types don't work.

@dsplaisted Should we copy the ResXResourceReader type to the msbuild codebase? If we try to mimic its behaviour, we end up re-implementing it.

cdmihai:
I looked at all the FEATURE_RESX_RESOURCE_READER instances, and it modified the contract in the following manner:

__Read from .resx __ - allowed, but values are cast to string, not interpreted. Therefore binary resources referenced via a file are not loaded into their .net objects. Either the contract for resgen.exe and .resx format is different for .net core, or this is a bug
__Read from .resource__ - not allowed

__Write to .resx__ - not allowed
__Write to .resource__ - yes, but by coercing values to string

rainersigwald:
This is fixable now that the ResX types are available in .NET Core 2, but MSBuild can't take advantage of that yet.

PR #4090, opened 2019-01-22
MSBuild fork of ResX reader/writer

rainersigwald:
Related: #308, #2221, https://github.com/dotnet/corefx/issues/34204.

ResX types are available and functional on .NET Core, but are in the winforms layer, which we don't want MSBuild to have to be on top of. Fortunately, they're pretty isolated, so we can pull a fork of that source into our repo and use it.

This will also probably help with future work to avoid deserialization/serialization during resource building, but doesn't directly address that yet.

Recommended review strategy:
* Commit-by-commit
* Ignore 7d32eb4 because it was just copying winforms stuff

To do:

- [x] Pull in source
- [x] make it compile
- [x] update to latest source
- [x] ResXFileRef
- [ ] Validate incremental behavior of same


rainersigwald:
Marking as WIP since https://github.com/dotnet/winforms/pull/212 just landed and I should adopt it too.

src/Tasks/winforms/misc/MultitargetUtil.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (string.IsNullOrEmpty(assemblyQualifiedName))
+                {
+                    assemblyQualifiedName = type.AssemblyQualifiedName;
+                }
+            }
+
+            return assemblyQualifiedName;
+        }
+
+        // ExecutionEngineException is obsolete and shouldn't be used (to catch, throw or reference) anymore.
```
danmosemsft: You can just remove ExecutionEngineException here.

danmosemsft:
If you didn't already, you might checkout https://github.com/dotnet/corefx/compare/master...danmosemsft:resx2?expand=1 and diff those files against yours, as I know I made various cleanup changes, possibly something will pop out that should come along. Or not.

rainersigwald:
FYI @merriemcgaw, @dreddy-work

rainersigwald:
This is no longer the approach we're going with. I'll put up a new PR shortly with an implementation that uses https://github.com/dotnet/corefx/pull/36906.

PR #4100, opened 2019-01-24
Turn GenerateCompiledExpressionsTempFile into a no-op

tmeschter:
This target in $(MSBuildFrameworkToolsPath)\Microsoft.WorkflowBuildExtensions.targets does bad things to builds inside Visual Studio. Specifically, it generates files *only* when building from within VS, for both real and design-time builds. This leads to all kinds of build problems.

In Dev16 we no longer need these files to exist on disk for Workflow Designer scenarios. We can't modify $(MSBuildFrameworkToolsPath)\Microsoft.WorkflowBuildExtensions.targets because that would break the Workflow Designer for older versions of VS using newer versions of the Framework. Instead, we effectively turn it off by overriding it with a no-op implementation here.

### DIFF
--- a/src/Tasks/Microsoft.WorkflowBuildExtensions.targets
+++ b/src/Tasks/Microsoft.WorkflowBuildExtensions.targets
@@ -24,4 +24,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
    <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.WorkflowBuildExtensions.targets" Condition="Exists('$(MSBuildFrameworkToolsPath)\Microsoft.WorkflowBuildExtensions.targets')" />
 
+   <!--
+         This target in $(MSBuildFrameworkToolsPath)\Microsoft.WorkflowBuildExtensions.targets does bad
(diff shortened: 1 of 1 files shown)


tmeschter:
Note this should not be merged in until after related changes to the Workflow Designer have been made to the main VS repo.

rainersigwald:
Marking for 16.0 per the internal thread.

When we talked about this internally, I envisioned dropping the framework import and copying the code here entirely. But this also totally works, and I think it's better:

Strategy | Pro | Con
------------ | ------------- | ---
Selective overrides | <ul><li>Understandable changes</li><li>Minimizes drift</li><li>Can be switched to full replacement later</li></ul> | <ul><li>Harder to customize more deeply</li><li>Having imports/overrides hurts evaluation performance by ε</li></ul>
Complete replacement | <ul><li>Flexible future customization</li> | <ul><li>Possiblity of unintentional drift</li><li>Moves cheese</li><li>perf</li></ul>

rainersigwald:
Per offline thread with @pilchie, this is desired for 16.0. Merge will be followed up with a VS insertion to take to the release quarterback.

PR #4102, opened 2019-01-28
Fix typo

changeworld:
ture -> true


src/Deprecated/Engine.UnitTests/DualQueue_Tests.cs
```diff
@@ -69,7 +69,7 @@ public void TestQueueEnqueueMultipleWriterOneReader()
                 resetEvent.WaitOne();
             }
           
-            // Pop off items from the queue and make ture that we got all of out items back out
+            // Pop off items from the queue and make true that we got all of out items back out
```
danmosemsft: I think it should be "sure" not "true" in both cases

PR #4125, opened 2019-02-01
Allow WriteLinesToFile to create empty files

rainersigwald:
A change intended as a performance optimization in #4067 caused a
backward-compatibility regression: `WriteLinesToFile` became unable to
create empty files. `Touch` would be more appropriate for that case, but
`WriteLinesToFile` is used in that way in the Framework-delivered
`GenerateCompiledExpressionsTempFile` target, which we can't change
easily.

Removes the condition that caused us to avoid calling `AppendAllText`
with an empty buffer.

Adds regression tests (inverting the ones that ensured deletion).

### DIFF
--- a/src/Tasks.UnitTests/WriteLinesToFile_Tests.cs
+++ b/src/Tasks.UnitTests/WriteLinesToFile_Tests.cs
@@ -179,44 +179,56 @@ namespace Microsoft.Build.Tasks.UnitTests
             }
         }
 
-        /// <summary>
-        /// Should delete the file when <see cref="WriteLinesToFile.Lines"/> is <c>null</c> and <see cref="WriteLinesToFile.Overwrite"/> is <c>true</c>.
-        /// </summary>
-        [Fact]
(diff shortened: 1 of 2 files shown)


rainersigwald:
fyi @CodeTherapist -- what I thought was a nice change in #4067 turned out to break all C# project builds in Visual Studio, which didn't turn up until a version of MSBuild with the change got pulled into a VS build. Hooray for accidental complexity! 😬

To be clear: I still really appreciate the contribution!

CodeTherapist:
@rainersigwald Oh my gosh... that is tricky. While I did my changes, I had the impression, that the code is as it schould be and there is prehaps a hidden intention about it. Unfortunately, I wasn't sceptical enough and there was no comment - that would be very useful in this case, it's not obvious...

Thanks you found it and told me. I learned something more about MSBuild - even if it looks wrong, it is probaly right! ;-)

rainersigwald:
> I learned something more about MSBuild - even if it looks wrong, it is probably right!

Or at least "even if it is wrong, somebody probably depends on it!" 😄

PR #4140, opened 2019-02-05
Disable OptProf (it's failing official builds)

rainersigwald:
@livarcocc I'd like to take this, get an official build to unblock the combined-with-Roslyn insertion, and then reenable it once things are fixed in #4134.

cc @tmat


rainersigwald:
We fixed forward instead.

PR #4142, opened 2019-02-06
Let ProjectStartedEventArgs.ParentProjectBuildEventContext.TaskId be …

MarkKharitonov:
…the Id of the MSBuild task that spawned the current project build.

This PR is my attempt to address the issue described here - https://github.com/Microsoft/msbuild/issues/3939

I have tested it on a very simple set of targets files. Unfortunately, I have not idea if it breaks any tests or how to unit test it in the first place.

### DIFF
--- a/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
+++ b/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
@@ -1127,7 +1127,7 @@ namespace Microsoft.Build.BackEnd
                         // we receive the configuration response before we enter the wait state.
                         newRequest = new BuildRequest(issuingEntry.Request.SubmissionId, GetNextBuildRequestId(),
                             request.Config.ConfigurationId, request.Targets, issuingEntry.Request.HostServices,
-                            issuingEntry.Request.BuildEventContext, issuingEntry.Request,
+                            issuingEntry.Request.CurrentMSBuildTask ?? issuingEntry.Request.BuildEventContext, issuingEntry.Request,
                             buildRequestDataFlags);
 
                         issuingEntry.WaitForResult(newRequest);
--- a/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
@@ -420,6 +420,11 @@ namespace Microsoft.Build.BackEnd
                     if (requirements != null)
                     {
                         TaskLoggingContext taskLoggingContext = _targetLoggingContext.LogTaskBatchStarted(_projectFullPath, _targetChildInstance);
+                        if (taskLoggingContext.TaskName == "MSBuild")
+                        {
+                            _buildRequestEntry.Request.CurrentMSBuildTask = taskLoggingContext.BuildEventContext;
+                        }
+
                         try
                         {
                             if (
@@ -461,6 +466,8 @@ namespace Microsoft.Build.BackEnd
                         }
                         finally
                         {
+                            _buildRequestEntry.Request.CurrentMSBuildTask = null;
+
                             // Flag the completion of the task.
                             taskLoggingContext.LogTaskBatchFinished(_projectFullPath, taskResult.ResultCode == WorkUnitResultCode.Success || taskResult.ResultCode == WorkUnitResultCode.Skipped);
 
--- a/src/Build/BackEnd/Shared/BuildRequest.cs
+++ b/src/Build/BackEnd/Shared/BuildRequest.cs
@@ -267,6 +267,8 @@ namespace Microsoft.Build.BackEnd
             }
         }
 
+        public BuildEventContext CurrentMSBuildTask { get; set; }
+
         /// <summary>
         /// The set of flags specified in the BuildRequestData for this request.
         /// </summary>


src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -420,6 +420,11 @@ private async Task<WorkUnitResult> ExecuteBucket(TaskHost taskHost, ItemBucket b
                     if (requirements != null)
                     {
                         TaskLoggingContext taskLoggingContext = _targetLoggingContext.LogTaskBatchStarted(_projectFullPath, _targetChildInstance);
+                        if (taskLoggingContext.TaskName == "MSBuild")
```
rainersigwald: It's possible for tasks other than `MSBuild` to cause other projects to build--any task that (like MSBuild itself) calls [`IBuildEngine.BuildProjectFile`](https://github.com/Microsoft/msbuild/blob/d3ae07ed0ca30ed03861215786135f9f9a699cd7/src/Framework/IBuildEngine.cs#L79-L101) or [`IBuildEngine3.BuildProjectFilesInParallel`](https://github.com/Microsoft/msbuild/blob/d3ae07ed0ca30ed03861215786135f9f9a699cd7/src/Framework/IBuildEngine3.cs#L15-L42) could be the reason here.

Let me think about the consequences of changing your `CurrentMSBuildTask` to something set more generically.

MarkKharitonov: Please, much appreciated.

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
```diff
@@ -1127,7 +1127,7 @@ private void IssueBuildRequests(BuildRequestEntry issuingEntry, FullyQualifiedBu
                         // we receive the configuration response before we enter the wait state.
                         newRequest = new BuildRequest(issuingEntry.Request.SubmissionId, GetNextBuildRequestId(),
                             request.Config.ConfigurationId, request.Targets, issuingEntry.Request.HostServices,
-                            issuingEntry.Request.BuildEventContext, issuingEntry.Request,
+                            issuingEntry.Request.CurrentMSBuildTask ?? issuingEntry.Request.BuildEventContext, issuingEntry.Request,
```
MarkKharitonov: My own concern is about this code, because it creates a regression in behavior which effect I cannot evaluate, because I do not know the code. After my change the `ChildProject.ParentProjectBuildEventContext <> ChildProject.ParentProject.BuildEventContext`. This may matter if there are maps that use BuildEventContext objects as keys. But I did not want to change the equality behavior of `BuildEventContext`, because this could mean an even greater regression. Somebody who knows the code very well should weigh in.

rainersigwald: Unfortunately, there's no such person available. I share your concern and will try to think about the implications, but I don't know if I'm going to be able to convince myself that it's "definitely ok".

MarkKharitonov: OK, how about a zero regression change where we add `TaskId2` to `BuildEventContext`? Its name reflects the current state of affairs - no one knows this area of code good enough to change it - ugly situation. So, we add to it. Writable field that does not participate in any Equals. It will be set to the MSBuild task Id in the project level `BuildEventContext`, so it is exposed in downstream projects as `ParentProjectBuildEventContext.TaskId2`.

Something has to be done. Correlating events is a nightmare.

karelz:
@rainersigwald @KirillOsenkov what can we do to move this forward?
Should we close it as "too risky"?

KirillOsenkov:
I apologize for the lack of activity. Rainer is super super busy, and I'm just super busy. We absolutely want this fix. I was intending to experiment, I think I had another idea that might be less risky. Unfortunately I never got around to it. 

Let it sit here for some more until one of us gets to it. Again, we're very sorry, but we're just super swamped and this is the unfortunate reality. We'll get to it when we can.

Thanks!

karelz:
Thanks @KirillOsenkov! Maybe it would be appropriate if @MarkKharitonov pings the PR every few weeks / every month to see how things are going and if he has to wait more ...
@MarkKharitonov are you ok with that?

KirillOsenkov:
I wouldn't worry about reminding us. We won't let it run through the cracks indefinitely. But feel free to ping the thread, it doesn't hurt either ;)

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
KirillOsenkov: I think this should be safe - the `BuildRequestId` is just returning `GetHashCode`:
https://github.com/microsoft/msbuild/blob/7bcb2f961093e87277439420d351a9f3f3a8cdb8/src/Framework/BuildEventContext.cs#L172-L175

If someone is taking a dependency on this I'd be concerned.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
KirillOsenkov: I'd argue that while this only covers the MSBuild task scenario and not the IBuildEngine APIs, there's so much value in this change that it's worth adding now, and not worrying about the remaining scenarios, which I think should be quite, if not extremely, rare.

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
Forgind: Would someone mind elaborating on why this is quite so risky? What I see is that either it does what was previously done or it uses something that's more accurate (prevented from messing up by turning it to null in the finally). I doubt someone would be too mad about us fixing a bug.

PR #4143, opened 2019-02-06
[WIP] Graph cross targeting

cdmihai:
A prototype implementation to support crosstargeting. This allows us to further evaluate the graph and find new issues. It has the following problems, which make it inadequate long term:
- project instance based graph builds now require that you clone each project instance in order to set global properties
- if a node depends directly on an inner build, the graph will overbuild and race with the node that crosstargets (eek!)
- hard to see how to elegantly extend it to allow building only one entry TF
- hard to see how to elegantly support the case where inner builds might have non-overlapping references

Also, I moved the graph types to `Microsoft.Build.Experimental.Graph` namespace. From now on, types in namespaces containing `.Experimental.` can get breaking changes at any point. When they mature, we can remove the `Experimental` part.


AndyGerlicher:
One thought I had, is it worth the complexity to not allow setting the IsGraphBuild property? I don't mind that the graph build would always add that, but it seems like if you really wanted to add that global property why block it?

Looks in general though. I like the move to experimental namespace. I don't think we can call this ready and done for 16.0.

cdmihai:
Regarding targeting vs16.0, it's probably best to have these types in the experimental namespace before dev16 ships, rather than after. I could create a separate PR just for the experimental switch. But since it is experimental, I considered it's fine to have this insufficient prototype checked in to avoid the hassle of dealing with branch drift. Though taking it out won't be as easy as deleting the branch :)

Regarding the reserved global properties, my reasoning was that if those properties are already set, should the graph overwrite them or not? If the graph does not overwrite them, the behaviour might be undefined, who knows how it will crash, or maybe it will just work. The same if the graph does overwrite them: we'd ignore user values, with undefined behaviour. Denying them seemed the most predictable thing.

Let me know if these make sense or not.

rainersigwald:
I lean toward "we should take only the Experimental move for 16.0". Even though we make no promises about the stability of code in there, I think we should still make _some_ effort to minimize churn.

cdmihai:
Ok, I extract the move to experimental into a separate PR.

cdmihai:
Replaced by #4218

PR #4148, opened 2019-02-06
Leave ProjectExtensionsElement if namespace is only diff

### LINKED ISSUE #3904: ProjectExtensionsElement doesn't do correct comparison for new and old value
### Steps to reproduce

1. Open Web Application project (ASP .NET Web Application (.NET Framework)) (or create if you don't have one)
2. Add new class to this project
3. Click "Save All" button.
4. Close VS

### Expected  behavior
<project>.csproj.user file is not updated after this operation.

### Actual behavior
<project>.csproj.user file is updated after this operation.

### Where problem is?
The problem is in the [ProjectExtensionsElement.cs](https://github.com/Microsoft/msbuild/blob/master/src/Build/Construction/ProjectExtensionsElement.cs) code in the method 'public string this[string name].set()' in this line:
if (idElement.InnerXml != value)

Here we compare serialized value of Xml in memory and new value. If it is the same we do nothing, otherwise we overwrite new value and mark node as a 'dirty' which causes re-write *.user file. Comparison is not correct, because InnerXml also adds schema information in the serialized string, but 'value' doesn't have this information. In this case this condition always 'true', even though XMLs are the same. Example:

**InnerXml**: `<FlavorProperties GUID="{349c5851-65df-11da-9384-00065b846f21}" xmlns="http://schemas.microsoft.com/developer/msbuild/2003"><WebProjectProperties><StartPageUrl></StartPageUrl><StartAction>CurrentPage</StartAction> .... and so on`

**Value**: `<FlavorProperties GUID="{349c5851-65df-11da-9384-00065b846f21}"><WebProjectProperties><StartPageUrl></StartPageUrl><StartAction>CurrentPage</StartAction>.... and so on`

rainersigwald:
When setting a ProjectExtensionsElement value, MSBuild diffs the current
and desired XML and writes only in the case that they differ. But this
is producing false negatives because the InnerXml may have a namespace
inherited from the project that isn't represented in the "real" XML.

Allow the current to be equal to the desired _modulo_ that namespace.

Fixes #3904.

### DIFF
--- a/src/Build/Construction/ProjectExtensionsElement.cs
+++ b/src/Build/Construction/ProjectExtensionsElement.cs
@@ -108,7 +108,11 @@ namespace Microsoft.Build.Construction
                     XmlElement.AppendChild(idElement);
                 }
 
-                if (idElement.InnerXml != value)
+                // The actual InnerXml may have the MSBuild namespace but be otherwise identical
+                // to the setting, in which case the namespace was probably inherited from the
+                // document and should be ignored.
+                if (idElement.InnerXml != value &&
+                    idElement.InnerXml.Replace(ProjectRootElement.EmptyProjectFileXmlNamespace, string.Empty) != value)
                 {
                     if (value.Length == 0)
                     {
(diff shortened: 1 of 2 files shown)


rainersigwald:
I applied bits with this change and it fixed the scenario in #3904. Trying to figure out if I can add a test quickly.

fyi @PavelUstinovMS

rainersigwald:
Annoyingly, the round-trip and the VS scenario have _different_ namespace-stripping behaviors. I'm going to force-push back to the manually-verified-but-no-tests case, because I don't see how the VS scenario does its thing.

PR #4152, opened 2019-02-07
Import MSBuildUserExtensionsPath\VisualStudioVersion

### LINKED ISSUE #4149: Version UserExtensionsPath imports with VisualStudioVersion
Suggested by @duncanp-sonar in https://github.com/Microsoft/msbuild/issues/3778#issuecomment-460638770

> The discussion above mentions `MSBuildExtensionsPath` but it doesn't mention `MSBuildUserExtensionsPath`. This change looks like it will break the per-user extension mechanism.
> 
> `MSBuildUserExtensionsPath` is in the form `C:\Users\[user id]\appdata\local\Microsoft\msbuild` so it is shared by all instances of MSBuild on the machine. Changing `MSBuildToolsVersion` to `Current` will break existing extensions, and it will also break the ability for per-user extension targets to be versioned by folder.
> 
> e.g. `ImportUserLocationsByWildcardBeforeMicrosoftCommonTargets` from `Microsoft.Common.CurrentVersion.targets`:
> 
> ```
> <Import Project="$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.targets\ImportAfter\*" Condition="'$(ImportUserLocationsByWildcardAfterMicrosoftCommonTargets)' == 'true' and exists('$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.targets\ImportAfter')"/>
> ```
> Per-user extension targets are important to third-party extensions as they allow targets to be injected into the build without requiring elevation (e.g. the _[Sonar Scanner for MSBuild](https://docs.sonarqube.org/display/SCAN/Analyzing+with+SonarQube+Scanner+for+MSBuild)_ that provides build integration with _[SonarQube/SonarCloud](https://sonarcloud.io/dashboard?id=sonarscanner-msbuild)_);
> 
> Could this breaking change be avoided by changing the per-user location to use a different build variable (e.g. either `VisualStudioVersion` or a new variable)?

rainersigwald:
Historically, per-VS-version customization was possible because imports
under MSBuildUserExtensionsPath were versioned. After moving to
ToolsVersion=Current, that would no longer be possible if we continued
keying off MSBuildToolsVersion.

VisualStudioVersion and MSBuildToolsVersion used to match (from 12.0 to
15.0). So switching to VisualStudioVersion shouldn't be noticeable for
folks that aren't in the minutia of MSBuild--just place the files in the
folder you logically would have anyway.

Closes #4149.

### DIFF
--- a/src/Tasks/Microsoft.CSharp.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.CSharp.CurrentVersion.targets
@@ -24,7 +24,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
        <ImportUserLocationsByWildcardAfterMicrosoftCSharpTargets Condition="'$(ImportUserLocationsByWildcardAfterMicrosoftCSharpTargets)' == ''">true</ImportUserLocationsByWildcardAfterMicrosoftCSharpTargets>
     </PropertyGroup>
 
-    <Import Project="$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.CSharp.targets\ImportBefore\*" Condition="'$(ImportUserLocationsByWildcardBeforeMicrosoftCSharpTargets)' == 'true' and exists('$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.CSharp.targets\ImportBefore')"/>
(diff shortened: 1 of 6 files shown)


rainersigwald:
cc @duncanp-sonar

rainersigwald:
On further consideration (https://github.com/Microsoft/msbuild/issues/4149#issuecomment-461633740) I've talked myself out of this. It's nice to have a single place for "all MSBuilds 16+" (saves churn on folks who use user extensions), and if it becomes necessary for a user extension to further have a dev16/dev17 (or whatever) split, it can handle that internally by further conditional imports.

Distributing an update with conditional imports would take the same amount of effort as distributing an update to support the 17.0 folder, so I think there's no reason to change away from Current.

PR #4158, opened 2019-02-07
Allow numeric comparison of $(MSBuildToolsVersion)

### LINKED ISSUE #4150: '$(MSBuildToolsVersion)' should compare as a version or number
See concerns/ideas in https://github.com/Microsoft/msbuild/issues/3778#issuecomment-441779192 and https://github.com/Microsoft/msbuild/issues/3778#issuecomment-457799037

Now that `$(MSBuildToolsVersion)` is `Current`, user projects or extensions that condition against it are likely to be broken, with an error like

```
A numeric comparison was attempted on "$(MSBuildToolsVersion)" that evaluates to "Current" instead of a number, in condition " '$(MSBuildToolsVersion)' == '' OR '$(MSBuildToolsVersion)' < '4.0' ".
```

rainersigwald:
Some extension (for sure WiX, probably others) compare
$(MSBuildToolsVersion) against a known baseline like "4.0" to determine
further behavior. This is broken by the change of ToolsVersion to
"Current":

```
A numeric comparison was attempted on "$(MSBuildToolsVersion)" that evaluates to "Current" instead of a number, in condition " '$(MSBuildToolsVersion)' == '' OR '$(MSBuildToolsVersion)' < '4.0' ".
```

Allow these comparisons again by treating expansion of that property
specially and as equivalent to CurrentVisualStudioVersion. Those two
values matched for MSBuild 12 to 15, so there shouldn't be too much user
confusion around the change.

Fixes #4150.

### DIFF
--- a/src/Build/Evaluation/Conditionals/StringExpressionNode.cs
+++ b/src/Build/Evaluation/Conditionals/StringExpressionNode.cs
@@ -42,11 +42,21 @@ namespace Microsoft.Build.Evaluation
         /// </summary>
         internal override double NumericEvaluate(ConditionEvaluator.IConditionEvaluationState state)
         {
+            if (ShouldBeTreatedAsVisualStudioVersion())
+            {
+                return ConversionUtilities.ConvertDecimalOrHexToDouble(MSBuildConstants.CurrentVisualStudioVersion);
+            }
+
             return ConversionUtilities.ConvertDecimalOrHexToDouble(GetExpandedValue(state));
         }
 
         internal override Version VersionEvaluate(ConditionEvaluator.IConditionEvaluationState state)
         {
+            if (ShouldBeTreatedAsVisualStudioVersion())
+            {
+                return Version.Parse(MSBuildConstants.CurrentVisualStudioVersion);
+            }
+
             return Version.Parse(GetExpandedValue(state));
         }
 
@@ -57,13 +67,22 @@ namespace Microsoft.Build.Evaluation
 
         internal override bool CanNumericEvaluate(ConditionEvaluator.IConditionEvaluationState state)
         {
+            if (ShouldBeTreatedAsVisualStudioVersion())
+            {
+                return true;
+            }
+
             return ConversionUtilities.ValidDecimalOrHexNumber(GetExpandedValue(state));
         }
 
         internal override bool CanVersionEvaluate(ConditionEvaluator.IConditionEvaluationState state)
         {
-            Version unused;
-            return Version.TryParse(GetExpandedValue(state), out unused);
+            if (ShouldBeTreatedAsVisualStudioVersion())
+            {
+                return true;
+            }
+
+            return Version.TryParse(GetExpandedValue(state), out _);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Build/Evaluation/Conditionals/StringExpressionNode.cs
```diff
@@ -141,6 +160,19 @@ internal override void ResetState()
...
+        /// Should this node be treated as an expansion of VisualStudioVersion, rather than
+        /// its literal meaning?
+        /// </summary>
+        /// <remarks>
+        /// Needed to provide a compat shim for numeric/version comparisons
+        /// on MSBuildToolsVersion, which were fine when it was a number
+        /// but now cause the project to throw InvalidProjectException when
+        /// ToolsVersion is "Current". https://github.com/Microsoft/msbuild/issues/4150
+        /// </remarks>
+        private bool ShouldBeTreatedAsVisualStudioVersion() =>
```
jeffkl: Should we worry about this being slow?

rainersigwald: I don't expect a straight string comparison to show up on any hot lists, even with IgnoreCase. It is slower than not having this, of course.

src/Build/Evaluation/Conditionals/StringExpressionNode.cs
```diff
@@ -42,11 +42,21 @@ internal override bool BoolEvaluate(ConditionEvaluator.IConditionEvaluationState
...
+            }
+
             return ConversionUtilities.ConvertDecimalOrHexToDouble(GetExpandedValue(state));
         }
 
         internal override Version VersionEvaluate(ConditionEvaluator.IConditionEvaluationState state)
         {
+            if (ShouldBeTreatedAsVisualStudioVersion())
+            {
+                return Version.Parse(MSBuildConstants.CurrentVisualStudioVersion);
```
jeffkl: We should have this parsed already or store its parsed value

src/Build/Evaluation/Conditionals/StringExpressionNode.cs
jeffkl: @davkean will this show up on the perf traces?

src/Build/Evaluation/Conditionals/StringExpressionNode.cs
rainersigwald: I thought about doing this and decided not to because:

* This is no more expensive than what we would always have been doing before.
* The pre-parsed value would take up memory for little value.

src/Build/Evaluation/Conditionals/StringExpressionNode.cs
rainersigwald: I can cache the result of this in a bool if you're that worried.

davkean: When you look at these sorts of things, factor in that opening the CoreFx solution caused 19,000 evaluations to show the Solution Explorer tree.

rainersigwald: Caching the result turned out to be handy to scope the quirk down further so the current rev does that.

PR #4146, opened 2019-02-06
VSIX dependency on NuGet.BuildTools

rainersigwald:
Fixes #3852 by ensuring that NuGet's tasks and targets are present
wherever MSBuild itself is, since common.targets now depends on NuGet.

Closes NuGet/Home#7390.

I did a private test build of 45c61c6e2 ([comparison view](https://github.com/Microsoft/msbuild/compare/vs16.0...exp/vsix-depend-on-nuget)), which is this commit + some build changes that shouldn't be relevant at all. DDRITs passed, Build Tools installs successfully, and the NuGet stuff is there in a C++-only Build Tools installation. Build and DDRIT results available here (Microsoft internal): https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/162986

@rrelyea, can you review or nominate another NuGet reviewer?

@DustinCampbell, this should help with bugs like https://github.com/OmniSharp/omnisharp-roslyn/issues/1311 (not completely; Build Tools could still be missing necessary C# workloads. But help some!)

PR #4159, opened 2019-02-08
VSIX dependency on NuGet.BuildTools

### LINKED ISSUE #3852: MSBuild VSIX should depend on NuGet VSIX that delivers core logic
See details in NuGet/Home#7390. Moving the manifest would need a coordinated change here and in NuGet.Client.

rainersigwald:
Fixes #3852 by ensuring that NuGet's tasks and targets are present
wherever MSBuild itself is, since common.targets now depends on NuGet.

Closes NuGet/Home#7390.


rainersigwald:
Duplicate of #4146, but targeting the right release branch.

PR #4162, opened 2019-02-10
Add "low" flag to launch processes as low priority.

bgianfo:
When building large MsBuild code bases, you often
would like to be able to use your computer while
the build is running in the background. Historically
the internal Microsoft build tools have supported
such functionality by running build processes as
low priority.

This change brings this functionality to MsBuild.


cdmihai:
Looks good. I'm okay with the change, provided the following:
- Support this for [non windows](https://github.com/Microsoft/msbuild/blob/master/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs#L442) as well. There might be an easier way to implement this crossplatform. Assuming that in all OSes the priority of the child process is inherited, you could just set the current process priority to low in XMake.cs, after parsing the cmd line arguments. As far as I know Linux processes inherit everything from the parent, and Windows child processes [sometimes inherit the priority when](https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessa):
```
... if none of the priority class flags is specified, the priority class defaults to NORMAL_PRIORITY_CLASS unless the priority class of the creating process is IDLE_PRIORITY_CLASS or BELOW_NORMAL_PRIORITY_CLASS. In this case, the child process receives the default priority class of the calling process.
```
Alternatively, skip crossplatform support but open an up-for-grabs issue on it
- Add a test which asserts that a spawned node is in low priority. Easiest way that I can think of is to do an out of proc build (`BuildParameters.DisableInProcNode = true`), call a task that returns the process priority, and assert it's low priority. You can put it in BuildManager_Tests which contains other tests doing out of proc builds.

bgianfo:
Thanks @cdmihai, setting it through the Process object model does seem to work, and sub processes inherit it correctly. 

With this change, it doesn't seem possible to write a test how you suggested though, since argument parsing options will never be processed by the new build node (unless I'm missing something). Should I write a test like those found in XMake_Test.cs which passes the command line argument to msbuild, and a task which returns it's process priority?

cdmihai:
Hmm, I see what you mean, without this being part of BuildParameters, it is impossible to react to it from inside the engine. So a question is, should this be a command line only feature, or an engine feature. I would go for command line only, as this does not seem to make sense for programatic API based access. If another process wants to use the BuildManager for low priority builds then that process can just set the priority instead. More so, it does not make sense to use multiple build managers with different priorities within the same process as they would stomp on each other, with last one winning.

Regarding tests, you are right, with this being a command line only test, you'd have to follow the pattern of the `XMake_Test` and set the environment variable `MSBUILDNOINPROCNODE` to value `1` to force out of proc builds. You can set the environment variable with a `TestEnvironment`, which resets all system mutations when it's disposed. And then assert against a task that prints the priority.

bgianfo:
Test added.

One thing I realized and had to fix in my tests, is that when building with low priority and with node re-use turned on, it's possible you could land on a build node that was previously launched as normal priority or low priority.

This is kind of a weird bug and might lead to unexpected behavior. 
Thoughts?

Pilchie:
FYI @cyrusn - this is similar to something you wanted from vbcscompiler, but better in a lot of ways, since it will effect other processes to.

@davkean / @jjmew - is this something we should think about exposing when VS launches builds?

rainersigwald:
> it's possible you could land on a build node that was previously launched as a priority which does't match the expectations of the msbuild frontend

Would it be reasonable to add process priority to the node handshake? The only downside I could see is that if there were a situation where priority wasn't inherited (for whatever reason) we might fail to connect to a just-launched node.

cdmihai:
> > it's possible you could land on a build node that was previously launched as a priority which does't match the expectations of the msbuild frontend
> 
> Would it be reasonable to add process priority to the node handshake? The only downside I could see is that if there were a situation where priority wasn't inherited (for whatever reason) we might fail to connect to a just-launched node.

I guess you could use `/nodereuse:false` with `/lowpriority`. Since the nodes are already slow, won't matter if some minor jitting time is spent as well :)

rainersigwald:
To take that a step further, `/low` could _imply_ `/nr:false`.

davkean:
I don't think /low implies /nr:false. Low means to me, do the build as fast as possible just don't interrupt my other applications that I'm using. 

Does this switch affect the vbcscompiler launch? If so, what if it's already open?

cdmihai:
> Does this switch affect the vbcscompiler launch? If so, what if it's already open?

If we'd want to reach out to task initiated build servers then we'd have to go down the route where the low priority gets stored in BuildParameters, and then we expose it to tasks via IBuildEngine. That way tasks, if they want to, can propagate the information to their servers, which, if they want to, can lower their priority for the duration of the request.

Though honestly for now i'd just go with this simple implementation and we can beef it up later based on feedback.

bgianfo:
It looks like we never reached consensus on the node reuse problem. Should I pursue making the low priority flag part of the node handshake given @davkean's objection to forcing /nr;flase?

src/MSBuild/XMake.cs
```diff
@@ -616,6 +618,14 @@ string commandLine
                         Environment.SetEnvironmentVariable("MSBUILDLOADALLFILESASWRITEABLE", "1");
                     }
 
+					// Honor the low priority flag, we place our selves below normal
+					// priority and let sub processes inherit that priority.
+					if (lowPriority)
+					{
+						Process currentProc = Process.GetCurrentProcess();
+						currentProc.PriorityClass = ProcessPriorityClass.BelowNormal;
+					}
```
danmosemsft: tabs instead of spaces?

src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -2096,6 +2138,14 @@ private string ExecuteMSBuildExeExpectSuccess(string projectContents, IDictionar
                     }
                 }
 
+                if (envsToCreate != null)
+                {
+                    foreach (var env in envsToCreate)
+                    {
+                        testEnvironment.SetEnvironmentVariable(env.Key, env.Value);
```
danmosemsft: Will this affect other tests run concurrently?

src/MSBuild/Resources/xlf/Strings.pl.xlf
```diff
@@ -178,6 +178,21 @@ Copyright (C) Microsoft Corporation. Wszelkie prawa zastrzeżone.
       LOCALIZATION: "MSBuild" should not be localized.
       LOCALIZATION: "-graphBuild" and "-graph" should not be localized.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
+    </note>
+      </trans-unit>
+      <trans-unit id="HelpMessage_37_LowPrioritySwitch">
+        <source>  -lowPriority[:True|False]
+                     Causes MSBuild to run at low priority.
+                     (Short form: -low)
```
danmosemsft: "Specifying /lowPriority is the same as specifying /lowPriority:true"

rainersigwald:
Talked this over with @BenVillalobos and @livarcocc. We think adding priority to the node handshake sounds like a good path forward. @bgianfo, would you mind doing that? microsoft/msbuild#4337 recently touched some of the handshake calculation and would be a good place to look. Let me know if you have any problems (or don't have time to do this now, we can pick it up from here if you're swamped).

bgianfo:
@rainersigwald I'll take a crack at it, will report back if I don't think I'll be able to get it done for whatever reason.

src/MSBuild/XMake.cs
bgianfo: Thanks, sorry about that.

src/MSBuild.UnitTests/XMake_Tests.cs
bgianfo: Looks like it probably will..

src/MSBuild/Resources/xlf/Strings.pl.xlf
bgianfo: Sure, thanks

bgianfo:
@rainersigwald is there a known issue with the tests hitting 1 hour timeout in Azure Devops?

rainersigwald:
No, but I've seen it a _lot_ today, please assume you're innocent until further notice.

src/MSBuild.UnitTests/XMake_Tests.cs
bgianfo: However I got this pattern from a bunch of existing tests.. so I'm not sure what the parallelism story is for the msbuild unit tests..

bgianfo:
Alright, I believe I have the node handshake coded up correctly.. all tests are passing.

src/MSBuild.UnitTests/XMake_Tests.cs
rainersigwald: No concurrency among our tests, this is ok. Or at least "not worse than before".

bgianfo:
Any chance of this getting reviewed this week?

cdmihai:
@bgianfo
We're going to discuss it in tomorrow's PR meeting, and decide on whether it needs more work, or it's good to go.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -72,17 +72,22 @@ public int AvailableNodes
...
+        /// <param name="enableLowPriority">Is the build running at low priority?</param>
+        internal static long GetHostHandshake(bool enableNodeReuse, bool enableLowPriority)
         {
             long baseHandshake = Constants.AssemblyTimestamp;
 
             baseHandshake = baseHandshake*17 + EnvironmentUtilities.Is64BitProcess.GetHashCode();
 
             baseHandshake = baseHandshake*17 + enableNodeReuse.GetHashCode();
 
+            baseHandshake = baseHandshake*17 + enableLowPriority.GetHashCode();
```
rainersigwald: Question: would it be possible to skip the plumbing of the `lowPriority` flag down to this point and look instead at the current process's priority? That might help with super-advanced cases where a user explicitly sets the entry-point MSBuild to some priority int but doesn't explicitly use `/low`. And it'd simplify the code around here.

rainersigwald: Before deciding on this please read my other comment.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -173,13 +180,20 @@ public void ShutdownConnectedNodes(bool enableReuse)
...
             // they must have been started with node reuse.
             bool nodeReuse = ComponentHost.BuildParameters?.EnableNodeReuse ?? true;
 
-            ShutdownAllNodes(NodeProviderOutOfProc.GetHostHandshake(nodeReuse), NodeProviderOutOfProc.GetClientHandshake(), NodeContextTerminated);
+            // If no BuildParameters were specified for this build,
+            // then they couldn't have been launched with priority since it defaults to off.
+            bool lowPriority = ComponentHost.BuildParameters?.LowPriority ?? false;
+
+            ShutdownAllNodes(
+                NodeProviderOutOfProc.GetHostHandshake(nodeReuse, lowPriority),
```
rainersigwald: I'm worried about only shutting down matching-priority nodes. That would make a scenario like `dotnet build -low && dotnet build-server shutdown` fail, because the second call to `ShutdownAllNodes` would not be at low priority.

We could call twice with low and not-low. But that would preclude the handshake reading priority directly.

We're also a bit concerned about the effect of such a change on VS shutdown time, though we don't have clear numbers to indicate that it would be a problem. That could probably be solved by moving the try-it-both-ways approach inside the part of the loop where we attempt to connect to enumerated processes.

src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -121,6 +121,11 @@ public class BuildParameters : ITranslatable
         private bool _enableNodeReuse = false;
 #endif
 
+        /// <summary>
+        /// Flag indicating whether the build process should run as low privilege.
```
wjk: Typo: Should be "low priority".

src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -768,6 +773,15 @@ public string OutputResultsCacheFile
         /// </summary>
         public bool DiscardBuildResults { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the build process should run as low privilege.
```
wjk: Ditto.

src/Build/BackEnd/BuildManager/BuildParameters.cs
bgianfo: Whoops, thanks!

src/Build/BackEnd/BuildManager/BuildParameters.cs
bgianfo: Thanks for spotting this, not sure what happened there..

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
bgianfo: I went with the path of trying to connect with and without the priority in the handshake.

bgianfo:
Last time I looked at this and rebased the PR, the tests are now failing in what appear to be consistently failing with some bug where  the invoked MsBuild isn't outputting the build priority. The output looks like an empty build project...  Spent a few hours debugging what's going on and didn't get very far.  ae5c2d89382817838efd03cc483e5b0bee780fb6 is one example of me trying to get more data out of CI. 

@rainersigwald Any debugging idea's?

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -173,13 +180,24 @@ public void ShutdownConnectedNodes(bool enableReuse)
...
             // other, completed build. If they're still around,
             // they must have been started with node reuse.
             bool nodeReuse = ComponentHost.BuildParameters?.EnableNodeReuse ?? true;
 
-            ShutdownAllNodes(NodeProviderOutOfProc.GetHostHandshake(nodeReuse), NodeProviderOutOfProc.GetClientHandshake(), NodeContextTerminated);
+            // To avoid issues with mismatched priorities not shutting
+            // down all the nodes on exit, we will attempt to shutdown
+            // all matching notes with and without the priroity bit set.
+            // So precompute both versions of the handshake now.
+            long hostHandshake = NodeProviderOutOfProc.GetHostHandshake(nodeReuse, enableLowPriority: false);
```
Forgind: nit:
I'd prefer if these two were inlined.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -121,7 +122,17 @@ protected void ShutdownAllNodes(long hostHandshake, long clientHandshake, NodeCo
             // For all processes in the list, send signal to terminate if able to connect
             foreach (Process nodeProcess in nodeProcesses)
             {
-                Stream nodeStream = TryConnectToProcess(nodeProcess.Id, 30/*verified to miss nodes if smaller*/, hostHandshake, clientHandshake);
+                // Verified to miss nodes if smaller.
+                int timeout = 30;
```
Forgind: It might be better to start nodes with the same priority as normal but quickly switch to low priority afterwards.
1) Nodes started with low priority are more likely to be killed between when they're started and when they finish the handshake. This leads to the hang in [this issue](https://github.com/microsoft/msbuild/issues/860)
2) If there's a 30-second timeout, and the node gets swapped out before then, this will fail.
3) Node creation is expected to be fast (<< 30 seconds), so a brief moment of high priority will likely be unnoticed.

I started a change to make node creation high priority, but I haven't finished it yet. It might be reasonable to combine the two if you like the idea of starting at high priority.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcTaskHost.cs
```diff
@@ -217,9 +217,21 @@ public void ShutdownConnectedNodes(bool enableReuse)
...
         {
-            ShutdownAllNodes(NodeProviderOutOfProc.GetHostHandshake(ComponentHost.BuildParameters.EnableNodeReuse), NodeProviderOutOfProc.GetClientHandshake(), NodeContextTerminated);
+            bool nodeReuse = ComponentHost.BuildParameters.EnableNodeReuse;
+
+            // To avoid issues with mismatched priorities not shutting
+            // down all the nodes on exit, we will attempt to shutdown
+            // all matching notes with and without the priroity bit set.
+            // So precompute both versions of the handshake now.
+            long hostHandshake = NodeProviderOutOfProc.GetHostHandshake(nodeReuse, enableLowPriority: false);
+            long hostHandshakeWithLow = NodeProviderOutOfProc.GetHostHandshake(nodeReuse, enableLowPriority: true);
```
Forgind: I think it makes more sense to only shut down matching nodes (at least if the node has started doing work). If I want to build two projects at once, one with high priority and one with low priority, shutting down one project will inadvertently shut down the other, too. I don't see a reason to want to build two projects of the same priority at once, since they'd just be using each other's resources, but with one high priority and one low priority, you might build a big thing slowly and do some small concept testing in the meantime at higher priority.

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -817,6 +817,50 @@ public void GraphBuildSwitchIdentificationTests()
...
+        public void LowPrioritySwitchIdentificationTests()
+        {
+            CommandLineSwitches.ParameterizedSwitch parameterizedSwitch;
+            string duplicateSwitchErrorMessage;
+            bool multipleParametersAllowed;
+            string missingParametersErrorMessage;
+            bool unquoteParameters;
+            bool emptyParametersAllowed;
+
+            Assert.True(CommandLineSwitches.IsParameterizedSwitch("low", out parameterizedSwitch, out duplicateSwitchErrorMessage, out multipleParametersAllowed, out missingParametersErrorMessage, out unquoteParameters, out emptyParametersAllowed));
```
Forgind: Switch to Shouldly for all these.

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -817,6 +817,50 @@ public void GraphBuildSwitchIdentificationTests()
...
+            Assert.Equal(CommandLineSwitches.ParameterizedSwitch.LowPriority, parameterizedSwitch);
+            Assert.Null(duplicateSwitchErrorMessage);
+            Assert.False(multipleParametersAllowed);
+            Assert.Null(missingParametersErrorMessage);
+            Assert.True(unquoteParameters);
+            Assert.False(emptyParametersAllowed);
+
+            Assert.True(CommandLineSwitches.IsParameterizedSwitch("LOW", out parameterizedSwitch, out duplicateSwitchErrorMessage, out multipleParametersAllowed, out missingParametersErrorMessage, out unquoteParameters, out emptyParametersAllowed));
```
Forgind: It would probably be better to switch these to a theory instead of copying ~the same code 4 times. Then you could also use `CommandLineSwitches.IsParameterizedSwitch("low", out CommandLineSwitches.ParameterizedSwitch parameterizedSwitch` and the same for the others to save a few more lines.

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -817,6 +817,50 @@ public void GraphBuildSwitchIdentificationTests()
...
+            Assert.Equal(CommandLineSwitches.ParameterizedSwitch.LowPriority, parameterizedSwitch);
+            Assert.Null(duplicateSwitchErrorMessage);
+            Assert.False(multipleParametersAllowed);
+            Assert.Null(missingParametersErrorMessage);
+            Assert.True(unquoteParameters);
+            Assert.False(emptyParametersAllowed);
+
+            Assert.True(CommandLineSwitches.IsParameterizedSwitch("lowpriority", out parameterizedSwitch, out duplicateSwitchErrorMessage, out multipleParametersAllowed, out missingParametersErrorMessage, out unquoteParameters, out emptyParametersAllowed));
```
Forgind: nit:
Please split lines this long across multiple lines.

src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -1959,7 +2008,7 @@ public void RestoreFirstClearsProjectRootElementCache()
                 }
             };
 
-            string logContents = ExecuteMSBuildExeExpectSuccess(projectContents, preExistingProps, "/restore");
+            string logContents = ExecuteMSBuildExeExpectSuccess(projectContents, filesToCreate: preExistingProps, arguments: "/restore");
```
Forgind: nit:
I don't like this sort of change, but you can wait for others' opinions if you'd like.

src/MSBuild/Resources/Strings.resx
```diff
@@ -787,6 +787,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_37_LowPrioritySwitch" UESanitized="false" Visibility="Public">
```
Forgind: UESanitized hasn't been used since 2004.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcTaskHost.cs
rainersigwald: This API does not affect currently-running builds. It exists so that MSBuild or Visual Studio can shut down _waiting, idle_ nodes left around from recent builds but not currently being used. Since you can't know what priority "all nodes" were created at, you can't pass it in here.

https://github.com/microsoft/msbuild/pull/4162#discussion_r304111430

Forgind: Makes sense to me. Thanks for explaining.

src/MSBuild/XMake.cs
```diff
@@ -620,6 +622,14 @@ string commandLine
                         Environment.SetEnvironmentVariable("MSBUILDLOADALLFILESASWRITEABLE", "1");
                     }
 
+                    // Honor the low priority flag, we place our selves below normal
+                    // priority and let sub processes inherit that priority.
+                    if (lowPriority)
+                    {
+                        Process currentProc = Process.GetCurrentProcess();
```
cdmihai: I'd move all these priority changes to a new utility class, with commands like `PriorityUtils.SwitchToUserLowPriority` that encapsulate the meaning of a "user wants to do other things with their computer" request and how to achieve it. Another use future potential use case for the util class would be encapsulating the meaning of "node spawned and needs high priority to respond back"

src/MSBuild/XMake.cs
```diff
@@ -1114,6 +1126,7 @@ string outputResultsCache
                     }
 
                     parameters.EnableNodeReuse = enableNodeReuse;
+                    parameters.LowPriority = lowPriority;
```
cdmihai: I'd rename this to something else than `BuildParameters.LowPriority`, because it looks like there might be multiple intents to playing around with priority, and the actual process/thread priority for one intent might be different then the other. How about making it a flag enum instead, something like `BuildParameters.PriorityMode`, where one value might be ``UserLowPriority`, another future one might be `NodeHandshake`. The good part is that we can keep adding new values without changing the public API.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -121,7 +122,17 @@ protected void ShutdownAllNodes(long hostHandshake, long clientHandshake, NodeCo
             // For all processes in the list, send signal to terminate if able to connect
             foreach (Process nodeProcess in nodeProcesses)
             {
-                Stream nodeStream = TryConnectToProcess(nodeProcess.Id, 30/*verified to miss nodes if smaller*/, hostHandshake, clientHandshake);
+                // Verified to miss nodes if smaller.
```
cdmihai: Ambiguous comment.

src/MSBuild/XMake.cs
Forgind: Is there any reason not to reuse the ProcessPriorityClass enum?

Also, since there is no longer another intention to change priority in the immediate future, I'm thinking it's better to leave it as LowPriority. It's messier if we want to go in and change it again, but I started making it fully generic and ran into difficulties with ShutdownAllNodes, since if it can take any value in PriorityClass, I'd either need a hash function that allocates specific bits (or equivalent) for the priority to make it easy to generate all relevant priorities or accept that I'll miss some nodes or shut down too many. Seems like too much headache for more generality that we might not use. I'll still make the `PriorityUtils.SwitchProcessPriorityTo` take a ProcessPriorityClass so it can be reasonably generic.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
danmosemsft: BTW if you do need more bits (eg., to specify the fine grained priority) you can probably remove the masking of the first byte that is done in `GetClientHandshake()`. This masking was needed when nodes might exist running versions of MSBuild written before ~2008 (?) ... which I think is now unlikely.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -72,17 +72,22 @@ public int AvailableNodes
...
+        /// /low:true build.
         /// </summary>
         /// <param name="enableNodeReuse">Is reuse of build nodes allowed?</param>
-        internal static long GetHostHandshake(bool enableNodeReuse)
+        /// <param name="enableLowPriority">Is the build running at low priority?</param>
+        internal static long GetHostHandshake(bool enableNodeReuse, bool enableLowPriority)
         {
             long baseHandshake = Constants.AssemblyTimestamp;
 
             baseHandshake = baseHandshake*17 + EnvironmentUtilities.Is64BitProcess.GetHashCode();
```
danmosemsft: Aside because it's not changed by your PR, but I hadn't noticed that the hashing scheme was now mixing all the bits. There is of course the possibility, however small, that that could cause a match when there should not be. Consider eg., if we changed the impolementation of string.GetHashCode() to return all zeros - that's absurd but valid. I thought originally this method was setting specific bits for specific features. Eg., 64 bitness could toggle one bit. Not sure why it's mixing the bits now.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
Forgind: As you pointed out below, there aren't specific bits for features anymore, but I could theoretically take advantage of this masking to make the first byte the priority. (I could really probably use the first four bits.) That would be a bit complicated, though, so I'm not sure it's worth it.

danmosemsft: Both sides of the handshake should be the same exact binaries, so this could theoretically be changed to be bit flipping. Not suggesting you do that, but it's not as if this scheme is locked in.

(Well - except that if this scheme is making a truly universally distributed bit pattern - then it actually is impossible to be certain that you won't accidentally connect to it. Unless you make the handshake longer. But now we're being even more implausible.)

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -121,7 +122,17 @@ protected void ShutdownAllNodes(long hostHandshake, long clientHandshake, NodeCo
             // For all processes in the list, send signal to terminate if able to connect
             foreach (Process nodeProcess in nodeProcesses)
             {
-                Stream nodeStream = TryConnectToProcess(nodeProcess.Id, 30/*verified to miss nodes if smaller*/, hostHandshake, clientHandshake);
+                // Some nodes can take up to this long to respond, so this is the smallest it can be without erroneously missing nodes.
```
rainersigwald: This is what the original comment meant, but let’s add context to the magic number. I believe this was a 2013 comment that didn’t provide methodology. Best to say that.

src/MSBuild/MSBuild.csproj
```diff
@@ -261,17 +262,10 @@
        in resolving references using them instead of the correct build-time references.
        Instead, add them just afterward, before assigning target paths/copying to
        the bin folder. -->
-  <Target Name="AddRefAssemblies"
-          DependsOnTargets="ResolveAssemblyReferences"
-          BeforeTargets="AssignTargetPaths"
-          Condition="'$(MonoBuild)' != 'true'">
+  <Target Name="AddRefAssemblies" DependsOnTargets="ResolveAssemblyReferences" BeforeTargets="AssignTargetPaths" Condition="'$(MonoBuild)' != 'true'">
```
rainersigwald: Can this reformatting be reverted?

src/MSBuild/PriorityUtils.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Build.CommandLine
+{
+    internal class PriorityUtils
+    {
+        public static void SwitchProcessPriorityTo(ProcessPriorityClass priority)
```
rainersigwald: This seems a bit much. Is the catch necessary now?

src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -769,6 +770,11 @@ public string OutputResultsCacheFile
         /// </summary>
         public bool DiscardBuildResults { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the build process should run as low priority.
+        /// </summary>
+        public bool LowPriority { get; set; }
```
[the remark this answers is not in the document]
Forgind: Would you mind providing more details on when something needs to survive serialization?

Priority passes automatically from parent process to child process, and it's also passed explicitly as a parameter at node creation. (Technically, we could get it from the process and avoid passing that around, but it would've made it harder if we had taken the high priority node creation change.) Is that sufficient to say it doesn't need to be translated or are there other necessary conditions?

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
Forgind: I imagine they just created a lot of processes, then ran a build to see if it would pick up the extant nodes, but I wasn't there in 2013. What was the methodology?

src/MSBuild/PriorityUtils.cs
Forgind: It is if we want to avoid crashing with no helpful message on certain operating systems if a user wants to use this to switch a node to a higher priority, even if that higher priority is changing from low priority to normal priority (which sounds very reasonable). That makes it worth it to me.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
rainersigwald: You don't have to describe the methodology, just don't state it as fact now. Otherwise, future me might email future you in five years demanding an explanation of how you knew this. And you won't know!

The internal commit that introduced this comment was http://vstfdevdiv:8080/DevDiv2/DevDiv/_versionControl/changeset/865139?_a=compare&path=%24%2FDevDiv%2Ffeature%2FVSPro_VBCS%2Fvsproject%2Fxmake%2FXMakeBuildEngine%2FBackEnd%2FComponents%2FCommunications%2FNodeProviderOutOfProcBase.cs

src/MSBuild/PriorityUtils.cs
rainersigwald: How would a user switch a node to a higher priority? The only flag we're exposing is low, right?

bgianfo:
Thanks for bringing this over the finish line @Forgind!

PR #4176, opened 2019-02-14
Don't match directories with prefix matches in globbed exclude patterns

### LINKED ISSUE #4175: Exclude is overzealous when exclude pattern is an absolute path prefix of include directory
_This was originally reported by @rla4 and refined by @peterhuene in dotnet/sdk#2813._

### Steps to reproduce

```xml
<Project>
 <PropertyGroup>
  <PrefixDirectory>$([MSBuild]::NormalizeDirectory("sub"))</PrefixDirectory>
 </PropertyGroup>
 <ItemGroup>
  <I Include="subdirectory\**" Exclude="$(PrefixDirectory)\**\*" />
 </ItemGroup>

 <Target Name="X">
  <Message Importance="High"
           Text="Prefix: $(PrefixDirectory)" />
  <Message Importance="High"
           Text="I: @(I)" />
 </Target>
</Project>
```

```
S:\repro\dotnet\sdk\issues\2813>tree /f
Folder PATH listing for volume Source
Volume serial number is E68B-B87E
S:.
│   .gitignore
│   prefix_exclude.proj
│
├───sub
└───subdirectory
        x.cs


S:\repro\dotnet\sdk\issues\2813>msbuild -v:m
Microsoft (R) Build Engine version 16.0.440-preview+gc689feb344 for .NET Framework

  Prefix: S:\repro\dotnet\sdk\issues\2813\sub\
  I:
```

rainersigwald:
Fixes #4175.

### DIFF
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -2305,7 +2305,7 @@ namespace Microsoft.Build.Shared
                         }
                         else if (excludeBaseDirectory.Length > includeBaseDirectory.Length)
                         {
-                            if (!excludeBaseDirectory.StartsWith(includeBaseDirectory, StringComparison.OrdinalIgnoreCase))
+                            if (!IsSubdirectoryOf(excludeBaseDirectory, includeBaseDirectory))
                             {
                                 //  Exclude path is longer, but doesn't start with include path.  So ignore it.
                                 continue;
@@ -2334,7 +2334,7 @@ namespace Microsoft.Build.Shared
                         else
                         {
                             //  Exclude base directory length is less than include base directory length.
-                            if (!state.BaseDirectory.StartsWith(excludeState.BaseDirectory, StringComparison.OrdinalIgnoreCase))
+                            if (!IsSubdirectoryOf(state.BaseDirectory, excludeState.BaseDirectory))
                             {
                                 //  Include path is longer, but doesn't start with the exclude path.  So ignore exclude path
                                 //  (since it won't match anything under the include path)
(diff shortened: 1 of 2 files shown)


src/Shared/FileMatcher.cs
```diff
@@ -2334,7 +2334,8 @@ static string[] CreateArrayWithSingleItemIfNotExcluded(string filespecUnescaped,
                         else
                         {
                             //  Exclude base directory length is less than include base directory length.
-                            if (!state.BaseDirectory.StartsWith(excludeState.BaseDirectory, StringComparison.OrdinalIgnoreCase))
+                            if (!(state.BaseDirectory.StartsWith(excludeState.BaseDirectory, StringComparison.OrdinalIgnoreCase) &&
+                                  directorySeparatorCharacters.Contains(state.BaseDirectory[excludeState.BaseDirectory.Length])))
```
cdmihai: Doesn't the previous `StartsWith` case also suffer from this?

rainersigwald: This one?
https://github.com/Microsoft/msbuild/blob/a4b44e00a085f90a279c0b454e2b1a8d8c77bea7/src/Shared/FileMatcher.cs#L2308-L2312

I had convinced myself that it didn't, but I can't remember why.

rainersigwald: I think it's that it just led to extra work, not an incorrect result. But we can avoid the extra work, so we should.

src/Shared/FileMatcher.cs
```diff
@@ -2440,6 +2441,12 @@ static string[] CreateArrayWithSingleItemIfNotExcluded(string filespecUnescaped,
             return files;
         }
 
+        private static bool IsSubdirectoryOf(string possibleChild, string possibleParent)
+        {
+            return possibleChild.StartsWith(possibleParent, StringComparison.OrdinalIgnoreCase) &&
+                   directorySeparatorCharacters.Contains(possibleChild[possibleParent.Length]);
```
cdmihai: Hmm, this only works if the parent does not end with a directory separator. Maybe it's best to adjust the index based on that (worst case parent could end in multiple, different directory separators). Or maybe just assert it does not end in a separator. Even if the code calling this normalizes a certain way, I can easily see it drifting away.

rainersigwald: I think other things would go wrong if the normalization failed here, but it was easy enough to account for parent-ending-with-slash, so I did that.

Issue #4181, opened 2019-02-15
Unexpandable metadata can crash loggers


rainersigwald:
_This was reported internally over email._

Reduced repro:

Project file
```xml
<Project>
 <ItemDefinitionGroup>
  <Foo>
   <MetadataFileName>a\b\%(Filename).c</MetadataFileName>
  </Foo>
 </ItemDefinitionGroup>
 <ItemGroup>
  <Foo Include="-in &quot;x\y\z&quot;" />
 </ItemGroup>

 <Target Name="X" />
</Project>
```

```
s:\>msbuild test.proj -bl
Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\15.0\bin\msbuild.exe -bl test.proj
Build started 2/15/2019 5:37:02 PM.
MSBUILD : error MSB4017: The build stopped unexpectedly because of an unexpected logger failure.
Microsoft.Build.Exceptions.InternalLoggerException: The build stopped unexpectedly because of an unexpected logger failure. ---> Microsoft.Build.Exceptions.InvalidProjectFileException: Cannot expand metadata in expression "a\b\%(Filename).c". The item metadata "%(Filename)" cannot be applied to the path "-in "x\y\z"". Illegal characters in path.
   at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject(String errorSubCategoryResourceName, IElementLocation elementLocation, String resourceName, Object[] args)
   at Microsoft.Build.Evaluation.Expander`2.MetadataExpander.ExpandMetadataLeaveEscaped(String expression, IMetadataTable metadata, ExpanderOptions options, IElementLocation elementLocation)
   at Microsoft.Build.Evaluation.Expander`2.ExpandIntoStringLeaveEscaped(String expression, ExpanderOptions options, IElementLocation elementLocation)
   at Microsoft.Build.Execution.ProjectItemInstance.TaskItem.GetMetadataEscaped(String metadataName)
   at Microsoft.Build.Execution.ProjectItemInstance.TaskItem.GetMetadata(String metadataName)
   at Microsoft.Build.Logging.BuildEventArgsWriter.Write(ITaskItem item)
   at Microsoft.Build.Logging.BuildEventArgsWriter.WriteItems(IEnumerable items)
[...]

PR #4186, opened 2019-02-19
Tolerate unexpandable metadata from definitions in logging

rainersigwald:
Using `ItemDefinitionGroup` metadata definitions, it's possible to create an item with an unexpandable metadatum: the `Include` is not a path-like thing and the metadatum uses a filesystem intrinsic metadatum name like `%(FileName)`.

Metadata is recursively expanded when logging is sufficiently high, which can lead to #4181 when 

There are a couple of options for resolving this. I'm proposing option 1 in this PR, but would like to get some opinions before going forward.

1. Log an error in the expanded metadata (including the unexpanded value and the error itself)
```
Initial Items:
Foo
    -in "x\y\z"
        MetadataFileName = Cannot expand metadata in expression "a\b\%(Filename).c". The item metadata "%(Filename)" cannot be applied to the path "-in "x\y\z"". Illegal characters in path.
```
2. Switch to logging raw, unexpanded metadata
```
Foo
    -in "x\y\z"
        MetadataFileName = a\b\%(Filename).c
```

I slightly prefer the former because if the metadata is used later, causing a build-time error, it gives a bit more of a clue about what's going wrong, and it still has the unexpanded string. But a point in favor of "just log unexpanded metadata" is that that's what happens when a new item is created inside a target.

### DIFF
--- a/src/Build.UnitTests/BinaryLogger_Tests.cs
+++ b/src/Build.UnitTests/BinaryLogger_Tests.cs
@@ -69,6 +69,31 @@ namespace Microsoft.Build.UnitTests
             ObjectModelHelpers.BuildProjectExpectSuccess(s_testProject, binaryLogger);
         }
 
+        [Fact]
+        public void BinaryLoggerShouldNotThrowWhenMetadataCannotBeExpanded()
+        {
+            var binaryLogger = new BinaryLogger
+            {
+                Parameters = $"LogFile={_logFile}"
+            };
+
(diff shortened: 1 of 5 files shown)


rainersigwald:
@KirillOsenkov, got opinions about the "correct" log format for can't-actually-expand-that metadata?

KirillOsenkov:
Nothing better comes to mind than what you already have here...

PR #4191, opened 2019-02-21
Set copyright and package license

tmat:
Prepares repo for change https://github.com/dotnet/arcade/pull/2003 by setting `Copyright` and `PackageLicenseExpression` properties. These values will be required to be set by each repository once https://github.com/dotnet/arcade/pull/2003 is merged.

In order to not break the current builds this change sets the properties conditionally. This condition can be removed once all repos switch to Arcade that has https://github.com/dotnet/arcade/pull/2003.

@markwilkie


tmat:
@rainersigwald @livarcocc PTAL

Directory.Build.props
```diff
@@ -1,7 +1,14 @@
-﻿<Project>
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
+  <PropertyGroup Condition="'$(CopyrightNetFoundation)' != ''">
+    <Copyright>$(CopyrightNetFoundation)</Copyright>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
```
rainersigwald: Does this override `PackageLicenseUrl` if it's present? Or do we need to condition that on the inverse of this condition?

https://github.com/Microsoft/msbuild/blob/f42e4b11fb675b8164a5c666cfc2d42c5755bffd/src/Directory.Build.props#L59

tmat: We need the condition on PackageLicenseUrl. Will add.

PR #4209, opened 2019-03-04
Revitalize xlf-edit check

rainersigwald:


rainersigwald:
The GitHub Checks experience for a failure here (https://github.com/Microsoft/msbuild/pull/4209/checks?check_run_id=72406585) is not what I'd call ideal, but IMO it's both better than before and (marginally) better than just printing the modified file.

nguerrera:
FWIW, other repos only set UpdateXlfOnBuild for non-CI builds, whereas here we're setting it always:

https://github.com/Microsoft/msbuild/blob/404bb795a7a25b4465b0d48d4357c40ac4b3cf61/src/Directory.Build.props#L39

Arcade has this:

https://github.com/dotnet/arcade/blob/cdd5d318667c5af1740d69c8a7755e23e7964aeb/src/Microsoft.DotNet.Arcade.Sdk/tools/Localization.targets#L11-L15

When it is false, xliff-tasks will trigger a build failure if any updates are needed.

Can we (should we) use what arcade has here?

rainersigwald:
> Can we (should we) use what arcade has here?

Yes, I think so. I slightly prefer the error experience of "dump the diff", but uniformity is good.

nguerrera:
Hmm I recall that maybe this check covered non xlf files too. Generated ref source? Cc @cdmihai.

rainersigwald:
You're right; bringing it back (but keeping the "standard" XLF behavior).

rainersigwald:
Example public API change: https://github.com/Microsoft/msbuild/pull/4209/checks?check_run_id=72443193

Example missed XLF update: https://github.com/Microsoft/msbuild/runs/72432221

(I'm now going to force-push to just the process changes I want with no deliberate errors.)

PR #4212, opened 2019-03-05
Add support mono reference assemblies folder.

mfilippov:
Add support mono reference assemblies folder passed through /p:TargetFrameworkRootPath parameter in .NET Core MSBuild. Now it doesn't work because of TargetFrameworkDirectory XML element under IsMono flag.

### DIFF
--- a/src/Utilities.UnitTests/ToolLocationHelper_Tests.cs
+++ b/src/Utilities.UnitTests/ToolLocationHelper_Tests.cs
@@ -1404,7 +1404,6 @@ namespace Microsoft.Build.UnitTests
         /// do not have anything to chain with.
         /// </summary>
         [Fact]
-        [Trait("Category", "mono-osx-failing")]
         public void ChainReferenceAssembliesRedistExistsNoRedistList()
         {
             string path = ToolLocationHelper.ChainReferenceAssemblyPath(@"PathDoesNotExistSoICannotChain");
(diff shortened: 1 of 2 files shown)


mfilippov:
@rainersigwald @radical Ready to review.

radical:
Mono supports the `RedistList/FrameworkList.xml` files having a `TargetFrameworkDirectory` attribute which can specify a different directory where the assemblies can be found.

- IIUC, this patch is trying to make that a msbuild proper feature. Is that correct?

- I'm not sure what you mean by `Add support mono reference assemblies folder passed through /p:TargetFrameworkRootPath`. Setting `TargetFrameworkRootPath` would allow you to set the path where it should be looking for the frameworks (and thus the `FrameworkList.xml` files). So, setting that shouldn't be related to the `TargetFrameworkDirectory`. Please correct me if I missed something or am wrong.

- Could you please share details of the use case that you are trying to solve?

Just in case this might be helpful, msbuild allows fallback paths for looking up target frameworks also. https://github.com/microsoft/msbuild/commit/f4dd0c94608633ffa6bf000b9ab3ae6741fd8e00

```
Add $(TargetFrameworkFallbackSearchPaths) for looking up target

.. frameworks. This contains ';' separated list of paths. When
`GetReferenceAssemblyPaths` task will look for assemblies corresponding to a
target framework, then it will try to find it in multiple locations:

	1. $(TargetFrameworkRootPath) or the default path if
	$(TargetFrameworkRootPath) is ''

	2. Each path in $(TargetFrameworkFallbackSearchPaths)
```

You can setup these extra paths with `/p:TargetFrameworkFallbackSearchPaths=...` or set them in a project file.

mfilippov:
@radical  The main goal of this patch: allows building full-framework project with mono reference assemblies. Before it was logic to pass through TargetFrameworkDirectory hidden behind condition IsMono and not working on .NET Core. In this patch, I drop IsMono condition and we have the same logic on all platforms. Also, it fixed test marked as failed on Mono. Logic with "TargetFrameworkFallbackSearchPaths"  not applicable in this case because not contains correctly selection based on the target framework name.

radical:
Ah, ok. So, you want to build a full framework project against mono's framework assemblies, using .netcore msbuild. I'm not sure what the thinking is on supporting this kinda scenario. And if it is, then whether it should be supported like this.
@rainersigwald @marek-safar @akoeplinger ^

rainersigwald:
@livarcocc @kathleendollard should also weigh in on the building using-mono-framework-assemblies using netcore MSBuild scenario.

mfilippov:
@radical Yes, it's not the default behaviour. It's work only with parameter: TargetFrameworkRootPath same as on the full framework. And also it reduces a difference in code between mono or non-Mono.

mfilippov:
Any news?

rainersigwald:
Thank you for the PR. Neither we nor the Mono team currently have plans to support targeting Mono from .NET Core or .NET Framework MSBuild. We are not planning to take this change, since it's likely just the first in a list of things that would need to change to fully support that scenario.

PR #4218, opened 2019-03-07
Graph cross targeting take two

cdmihai:
Reimplementation of #4143 to address its issues by adding explicit nodes for inner builds instead of trying to hide their complexity away. It does so by imposing a contract on what crosstargeting means, and requiring the sdk to be self-descriptive about its crosstargeting structure (the property name of the inner build (e.g. `TargetFramework`), and the property name of the inner build values (e.g. `TargetFrameworks`)).
- no more races on inner builds. They get deduped in the graph.
- inner builds can have non overlapping references
- inner builds can have arbitrary incoming edges
- clear path forward on how to filter the graph per TargetFramework. Right now if project A depends on crosstargeting project B, all of Bs innerbuilds will be added as speculative edges of A. This is no different from building a solution today, which builds all TFs. If one wants to filter, then they have to flow the desired TF through the graph (using Nuget's GetNearestTF) and clip out the unused speculative edges.

Future PRs:
- update documentation
- tag nodes and edges of interest with extra metadata: outer builds, inner builds, speculative edges
- refactorings, better inner build detection to avoid overbuilding some targets


cdmihai:
cc @erikma @dannyvv @yash256

ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -1427,6 +1427,7 @@ public sealed partial class ProjectGraph
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Experimental.Graph.ProjectGraphNode> ProjectNodes { get { throw null; } }
         public System.Collections.Generic.IReadOnlyCollection<Microsoft.Build.Experimental.Graph.ProjectGraphNode> ProjectNodesTopologicallySorted { get { throw null; } }
         public System.Collections.Generic.IReadOnlyDictionary<Microsoft.Build.Experimental.Graph.ProjectGraphNode, System.Collections.Immutable.ImmutableList<string>> GetTargetLists(System.Collections.Generic.ICollection<string> entryProjectTargets) { throw null; }
+        public string ToDot() { throw null; }
```
AndyGerlicher: Not sure we'd want this in the public API. It's pretty trivial to create it after the fact. If it helps for testing, it could be made internal?

AndyGerlicher:
Looks good! Only think I see is the `ToDot` which the more I think about it the more I think it should go.

ref/Microsoft.Build/net/Microsoft.Build.cs
cdmihai: I was thinking it would help have it public in case we decorate the Dot spec with internal information. But maybe this adds even more support to the fact that we need to decorate nodes with tags / metadata like `InnerBuild`, `SpeculativeEdge`, etc

src/Shared/UnitTests/TestEnvironment.cs
```diff
@@ -331,6 +331,11 @@ public void WriteLine(string format, params object[] args)
                 Console.WriteLine(format, args);
             }
         }
+
+        public void DoNotLaunchDebugger()
```
erikma: Need doc comment

src/Tasks/Microsoft.ManagedLanguages.targets
```diff
@@ -0,0 +1,64 @@
...
+
+This file defines common build logic for all managed languaged: C#, VisualBasic, F#
+
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+
+<Project>
+   <!-- 
+        We are doing a cross-targeting build if there is no list of target frameworks specified
```
erikma: "there is no list" => "there is a list" to match the logic below

src/Tasks/Microsoft.ManagedLanguages.targets
```diff
@@ -0,0 +1,64 @@
...
+      <InnerBuildProperty>TargetFramework</InnerBuildProperty>
+      <InnerBuildPropertyValues>TargetFrameworks</InnerBuildPropertyValues>
+   </PropertyGroup>
+
+   <PropertyGroup Condition="'$(IsGraphBuild)' == 'true'">
+      <!-- Microsoft.Net.Sdk adds all transitive project references to @ProjectReference. It does this at build time
+           via a target. This fails the undeclared msbuild reference /isolate constraint, so disable transitive
+           references.
+           Do not disable transitive references if BuildProjectReferences is false. If building references is false,
+           then msbuild won't try to build the transitive references, which won't trigger undeclared reference constraints.
```
erikma: msbuild => MSBuild

src/Tasks/Microsoft.ManagedLanguages.targets
cdmihai: Good catch

src/Shared/UnitTests/TestEnvironment.cs
```diff
@@ -331,6 +331,14 @@ public void WriteLine(string format, params object[] args)
                 Console.WriteLine(format, args);
             }
         }
+
+        /// <summary>
+        /// MSBuild launches the debugger on ErrorUtilities exceptions when in DEBUG. Disable this in tests that assert these exceptions.
+        /// </summary>
+        public void DoNotLaunchDebugger()
+        {
+            SetEnvironmentVariable("MSBUILDDONOTLAUNCHDEBUGGER", "1");
```
rainersigwald: I thought we always set this for tests. Did a failure cause a test hang?

src/Tasks/Microsoft.CSharp.targets
```diff
@@ -165,6 +156,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Import Project="$(MSBuildExtensionsPath)\4.0\Microsoft.NETFramework.props\ImportBefore\*" Condition="'$(ImportByWildcardBefore40MicrosoftNetFrameworkProps)' == 'true' and exists('$(MSBuildExtensionsPath)\4.0\Microsoft.NETFramework.props\ImportBefore')"/>
    </ImportGroup> 
 
+   <Import Project="$(MSBuildToolsPath)\Microsoft.ManagedLanguages.targets" />
```
rainersigwald: F# will need this too, eventually?

src/Tasks/Microsoft.ManagedLanguages.targets
```diff
@@ -0,0 +1,64 @@
...
+        and there is no current target framework being built individually. In that case, a crosstargeting
+        project file like Microsoft.<language>.CrossTargeting.targets gets imported.
+   -->
+   <PropertyGroup Condition="'$(TargetFrameworks)' != '' and '$(TargetFramework)' == ''">
+      <IsCrossTargetingBuild>true</IsCrossTargetingBuild>
+   </PropertyGroup>
+
+   <PropertyGroup>
+      <InnerBuildProperty>TargetFramework</InnerBuildProperty>
+      <InnerBuildPropertyValues>TargetFrameworks</InnerBuildPropertyValues>
```
rainersigwald: Does structuring this as a property mean there can be only one dimension for inner builds? That's a current limitation of cross-targeting but people keep hoping we relax it somehow.

src/Tasks/Microsoft.ManagedLanguages.targets
```diff
@@ -0,0 +1,64 @@
...
+   </PropertyGroup>
+
+   <PropertyGroup Condition="'$(IsGraphBuild)' == 'true'">
```
rainersigwald: Should we rename to `MSBuildIsGraphBuild` and make it a reserved property, so you can't set it manually?

src/Shared/UnitTests/TestEnvironment.cs
cdmihai: I had written a few tests to assert failures from ErrorUtilities.Verify* which crashed due to the debugger window launching, and setting that env var fixed them. But then I deleted those tests (because I deleted the underlying code they tested), and I decided to keep this helper because it sounds useful to have.

src/Tasks/Microsoft.CSharp.targets
cdmihai: Yeah, eventually, as this flows to master and releases. And especially once we start pilling more common code in there.

src/Tasks/Microsoft.ManagedLanguages.targets
cdmihai: I decided against generalization and went with the minimum to support the current sdk. The idea is we'll expand when necessary.
Ideally, `InnerBuildProperty` would be a set of global property names that define the inner build, and `InnerBuildPropertyValues` would contain all possible values somehow:
```
<InnerBuildProperties>TargetFramework;TargetPlatform ...
<InnerBuildPropertyValues>TargetFramework={net472;netcoreapp2.2}; TargetPlatform={win-x64;osx-x64} ...
```
Or, if one does not want a cross product of all values:
```
<InnerBuildProperties>TargetFramework;TargetPlatform ...
<InnerBuildPropertyValues>
  {
    {TargetFramework=net472; TargetPlatform=win-x64},
    {TargetFramework=netcoreapp2.2; TargetPlatform=osx-x64}
  }
```

But since these are not needed yet, not thinking about them speeds this current PR up. Let me know if we should think about them now though :)

src/Tasks/Microsoft.ManagedLanguages.targets
cdmihai: @AndyGerlicher suggests not constraining it: https://github.com/Microsoft/msbuild/pull/4143#issuecomment-461198290

Making it reserved avoids an entire class of problems that can result from people manually setting this. On the other hand, power users might enjoy shooting themselves in the foot :)

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -600,8 +628,232 @@ public void GetTargetListsNullEntryTargets()
...
+                    null,
+                    null,
+                    null,
+                    OuterBuildSpecificationWithProjectReferenceTargets +
+                    $@"<ItemGroup>
+                            <ProjectReference Include=`3.proj` Condition=`'$({InnerBuildProperty})'=='a'`/>
+
+                            <ProjectReference Include=`4.proj` Condition=`'$({InnerBuildProperty})'=='b'`/>
+                            <ProjectReference Include=`5.proj` Condition=`'$({InnerBuildProperty})'=='b'`/>
+                            <ProjectReference Include=`6.proj` Condition=`'$({InnerBuildProperty})'=='b'` Properties=`{InnerBuildProperty}=a`/>
```
rainersigwald: I'm confused on this one. 

image

Why do we get the speculative `[p1 ib=b] -> [p6 outer]` and `[p6 ib=a]` edges? My read on https://github.com/Microsoft/msbuild/blob/56e10566f8491faab2421221f9b68b109b06a014/documentation/specs/static-graph.md#multitargeting was that only references to outer builds (subject to TF negotiation) would get the speculative links.

rainersigwald: (Also, while I agree with Andy that `ToDot()` should stay internal--it's awesome :))

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -842,14 +1094,486 @@ public void TopologicalSortShouldTopologicallySort(Dictionary<int, int[]> edges)
...
+
+            AssertInnerBuildEvaluation(graph.GraphRoots.First(), true, new Dictionary<string, string>());
+
+            var nonMultitargetingNode = GetFirstNodeWithProjectNumber(graph, 2);
+
+            AssertNonMultitargetingNode(nonMultitargetingNode);
+        }
+
+        [Fact]
+        public void NonOuterBuildProjectsInTheMiddle()
```
rainersigwald:
```suggestion
        public void NonMultitargetingBuildProjectsInTheMiddle()
```
?

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -842,14 +1094,486 @@ public void TopologicalSortShouldTopologicallySort(Dictionary<int, int[]> edges)
...
+            CreateProjectFile(_env, 4, null, null, null, OuterBuildSpecification);
+
+            var graph = new ProjectGraph(root);
+
+            var dot = graph.ToDot();
+
+            graph.ProjectNodes.Count.ShouldBe(8);
+
+            AssertOuterBuildAsRoot(graph.GraphRoots.First());
+            AssertOuterBuildAsNonRoot(GetNodesWithProjectNumber(graph, 4).First(IsOuterBuild));
```
rainersigwald: Also assert 2 inner builds of p4?

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -842,14 +1094,486 @@ public void TopologicalSortShouldTopologicallySort(Dictionary<int, int[]> edges)
...
+        {
+            var extraInnerBuildReferenceSpec = OuterBuildSpecification +
+                                          $@"<ItemGroup>
+                                                <ProjectReference Condition=`'$({InnerBuildProperty})'=='b'` Include=`4.proj;5.proj`/>
+                                            </ItemGroup>".Cleanup();
+
+            var root = CreateProjectFile(_env, 1, new[] {2, 3}, null, null, extraInnerBuildReferenceSpec).Path;
+            CreateProjectFile(_env, 2, null, null, null, OuterBuildSpecification);
+            CreateProjectFile(_env, 3);
+            CreateProjectFile(_env, 4, null, null, null, OuterBuildSpecification);
```
rainersigwald: Why does this not produce 4outer->4innerA,4innerB edges?

image

src/Shared/Constants.cs
```diff
@@ -106,6 +105,30 @@ internal static class MSBuildConstants
         internal static readonly char[] PathSeparatorChar = { Path.PathSeparator };
     }
 
+    internal static class PropertyNames
+    {
+        /// <summary>
+        /// Specifies whether the current evaluation / build is happening during a graph build
+        /// </summary>
+        internal const string IsGraphBuild = "IsGraphBuild";
```
rainersigwald:
```suggestion
        internal const string IsGraphBuild = nameof(IsGraphBuild);
```

And likewise below?

src/Tasks.UnitTests/Copy_Tests.cs
```diff
@@ -533,11 +533,11 @@ public void DoCopyOverDifferentFile()
         [InlineData(true)]
         public void DoCopyOverCopiedFile(bool skipUnchangedFiles)
         {
-            var sourceFile = FileUtilities.GetTemporaryFile(null, "src", false);
-            var destinationFile = FileUtilities.GetTemporaryFile(null, "dst", false);
-
-            try
+            using (var env = TestEnvironment.Create())
```
rainersigwald: Love this but can you please pull it into a different PR?

src/Tasks/Microsoft.ManagedLanguages.targets
rainersigwald: Another possible source of inspiration here is the Azure Pipelines YAML format for specifying a [variable matrix](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#matrix).

src/Build/Graph/ProjectInterpretation.cs
```diff
@@ -0,0 +1,387 @@
...
+
+        public IEnumerable<ConfigurationMetadata> GetReferences(ProjectInstance requesterInstance)
+        {
+            IEnumerable<ProjectItemInstance> references;
+            IEnumerable<GlobalPropertiesModifier> globalPropertiesModifiers = null;
+
+            switch (GetProjectType(requesterInstance))
+            {
+                case ProjectType.OuterBuild:
+                    references = GetInnerBuildReferences(requesterInstance);
```
rainersigwald: We should find a place to write down the implication of this: an outer build **cannot** have its own distinct `ProjectReference`s. That's the case today, but I could see a situation where you might want it: a packaging project that packs its inner build outputs + "other non-TF-specific stuff" acquired through reference).

I don't think this is a bad limitation. Just worth formalizing.

src/Build/Graph/ProjectInterpretation.cs
```diff
@@ -0,0 +1,387 @@
...
+                {
+                    foreach (var innerBuild in outerBuild.ProjectReferences)
+                    {
+                        foreach (var referencingProject in outerBuild.ReferencingProjects)
+                        {
+                            referencingProject.AddProjectReference(innerBuild);
+                        }
+                    }
+
+                    outerBuild.RemoveReferences();
```
rainersigwald: Ah, my confusion reduces:

This removes links from outer->inner builds of any build that is not a graph root, because the what-targets-should-we-run-on-a-ProjectInstance code has already decided not to run `Build` (really DispatchToInnerBuilds) on the outer build.

The referencing project, then, is given results for

* Outer build: GetTargetFrameworks
* Inner builds: Actual build

And can call its negotiation code followed by an inner build using those results.

This makes sense but has a surprising (to me) property: adding a new graph root that just calls the prior graph root changes the graph in such a way that **the new graph is not a subgraph of the old one**. Is that ok? I think it probably is, but deserves documentation.

This might have negative implications for graph caching. Is the edge removal required?

src/Build/Graph/ProjectInterpretation.cs
```diff
@@ -0,0 +1,387 @@
...
+            ErrorUtilities.VerifyThrow(!(isOuterBuild && isInnerBuild), $"A project cannot be an outer and inner build at the same time: ${project.FullPath}");
+
+            return isOuterBuild
+                ? ProjectType.OuterBuild
+                : isInnerBuild
+                    ? ProjectType.InnerBuild
+                    : ProjectType.NonMultitargeting;
+        }
+
+        public static void PostProcess(ConcurrentDictionary<ConfigurationMetadata, ProjectGraphNode> allNodes)
```
rainersigwald:
```suggestion
        public static void PromoteInnerBuildReferencesToReferencingProjects(ConcurrentDictionary<ConfigurationMetadata, ProjectGraphNode> allNodes)
```

? If this is the way to keep it.

rainersigwald:
Can you add some negative tests? Like "doesn't provide the correct metadata to understand the inner build" -- what's the failure look like? `InvalidProjectFileException`?

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -600,8 +628,232 @@ public void GetTargetListsNullEntryTargets()
...
+        public void GetTargetListsUsesAllTargetsForNonMultitargetingNodes()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                var root1 = CreateProjectFile(env, 1, new[] {2}, null, null, ProjectReferenceTargetsWithMultitargeting).Path;
+                CreateProjectFile(env, 2);
+                
+                var projectGraph = new ProjectGraph(root1);
+
+                var dot = projectGraph.ToDot();
```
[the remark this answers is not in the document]
cdmihai: Just for debugging purposes. it's quite useful to visualize the graph when something goes wrong.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -600,8 +628,232 @@ public void GetTargetListsNullEntryTargets()
...
+                targetLists[GetFirstNodeWithProjectNumber(projectGraph, 3)].ShouldBe(new[] { "C" });
+            }
+        }
+
+        [Fact]
+        public void GetTargetListsUsesAllTargetsForNonMultitargetingNodes()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                var root1 = CreateProjectFile(env, 1, new[] {2}, null, null, ProjectReferenceTargetsWithMultitargeting).Path;
```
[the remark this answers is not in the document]
cdmihai: Well, R# automatically displays argument names as of some recent version.
image

But yeah, I'll add explicit arguments.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -479,28 +673,19 @@ private static IReadOnlyCollection<ProjectGraphNode> TopologicalSort(IReadOnlyCo
...
 
-                            string projectReferenceFullPath = projectReferenceToParse.GetMetadataValue(FullPathMetadataName);
-                            PropertyDictionary<ProjectPropertyInstance> projectReferenceGlobalProperties = GetProjectReferenceGlobalProperties(projectReferenceToParse, projectToEvaluate.GlobalProperties);
-                            var projectReferenceConfigurationMetadata = new ConfigurationMetadata(projectReferenceFullPath, projectReferenceGlobalProperties);
-                            if (!tasksInProgress.ContainsKey(projectReferenceConfigurationMetadata))
+                        foreach (var referenceConfig in projectInterpretation.GetReferences(parsedProject.ProjectInstance))
+                        {
+                             /*todo: fix the following double check-then-act concurrency bug: one thread can pass the two checks, loose context,
+                             meanwhile another thread passes the same checks with the same data and inserts its reference. The initial thread regains context
+                             and duplicates the information, leading to wasted work
```
rainersigwald: Just wasted work, not a correctness problem? I can live with that for a while.

src/Tasks/Microsoft.ManagedLanguages.targets
```diff
@@ -0,0 +1,64 @@
+﻿<!--
+***********************************************************************************************
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+This file defines common build logic for all managed languaged: C#, VisualBasic, F#
```
rainersigwald: @nguerrera @livarcocc and maybe @jaredpar or @agocke: any objections to introducing this new common targets file, intended to be imported from `Microsoft.CSharp.targets` and similar for other languages?

nguerrera: No real concerns, but let me bikeshed on the name while I'm here. :)

There are two analogs to this already:

* [Microsoft.Managed.Core.targets](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/MSBuildTask/Microsoft.Managed.Core.targets)
* [Microsoft.Managed.DesignTime.targets](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets)

Following an emergent pattern, this one should be Microsoft.Managed.targets.

The pattern I see is X.targets -> msbuild, X.DesignTime.targets -> project-system., X.Core.targets -> roslyn

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: Wow, good catch! It appears to be a bug. Maybe. `[p1 ib=b]` is directly referencing the inner build `[p6 ib=a]` by setting `Properties="InnerBuild=a"` metadata on `ProjectReference`. But at the same time, inner builds set the inner build property as an `UndefineProperty`, to avoid propagating their TF to the references. This undefine removes the effect of `Properties="InnerBuild=a"` and leads to P6 being transformed into an outer build.

I'll go debug how vanilla msbuild handles this situation, and adjust to that behaviour.

src/Tasks/Microsoft.ManagedLanguages.targets
cdmihai: Good observation. I'll rename it to `Microsoft.Managed.targets`

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: Mkay, this is how the sdk works (inner build A(net472) references B(net462) via `Properties="TargetFramework=net462"` metadata):
- A gets to call B(no global props), the outer build, with `GetTargetFrameworks` (GTF) and B(net462) with Build, GetNativeManifest, etc
  - Calling B(no global props) with GTF is wasted work, as the information won't get used since the Properties metadata overwrites any TF negotiation
  - A(net472) calls GTF on B outer build because [TargetFramework gets explicitly removed for calls to GTF](https://github.com/Microsoft/msbuild/blob/a972ec96c3920705e4e8d03d7ac8b6c3328450bd/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1642)
- The scenario ends up working in reality because the [SDK does not remove `TargetFramework`](https://github.com/Microsoft/msbuild/blob/a972ec96c3920705e4e8d03d7ac8b6c3328450bd/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1686-L1688) for the inner targets (Build, GetNativeManifest etc) when Nuget finds a `NearestTargetFramework` and the reference has multiple target frameworks. The static graph, on the other hand, always removes the `TargetFramework` on an inner build's references, which leads to the B(net462) getting converted to an outer build instead.

So, in this case, the static graph accidentally overbuilds, which I guess is better than underbuilding, which would cause cache miss build breaks at runtime. Since having inner builds manually reference other inner builds seems like an edge case, I propose fixing it in a separate PR. Let me know if you think otherwise.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: Already asserted in the two `AssertOuterBuildAs*` methods.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: Answered in https://github.com/Microsoft/msbuild/pull/4218#discussion_r267830872 right?

src/Build/Graph/ProjectInterpretation.cs
cdmihai: The edge removal is there to have the graph be a more truthful representation of the runtime graph. The outer -> inner edges for non root outer builds do not appear in reality. Because of this, anyone who reasons about the graph needs to know about three little idiosyncrasies:
1. that outer -> inner edges are not there in reality. So one cannot reason about a single node in isolation, one always needs to check for this case.
2. out of all the outer -> inner edges, only one manifests in reality.
3. the edge to the outer build node is a true edge

With the edge removal, the number of idiosyncrasies get reduced to only one, a variation of 2: When referencing a multitargeting node, the referencing node considers all TFs and gets to pick one. The graph contains all the edges, similar to symbolic execution, but at runtime only one path is taken.

You are right, the structural difference between outer builds as roots and outer builds as non roots destroys the nice complete-structural-sharing property for graph caching. Here's some thoughts on that:
- This issue exists with vanilla msbuild, it's not an additional artificial static graph consequence
- Even though it's not 100% sharing, I think there's still significant structural sharing. Given a graph G, we can decompose it into the graph of roots, Gr, and the graph of non-roots, Gnr. When a new node N is added to depend on G via an edge inside Gr, Gr cannot be reused, but Gnr can. The trick will be to keep track of Gr and Gnr.

Options
1. change the graph to enable full structural sharing of subgraphs. The compromise here is that the new graph structure might depart from the actual runtime graph
2. keep the graph in tune with the runtime graph, but make graph caching more complex.

For now I'd just go with 2. I'll update the spec to account for this though.

src/Build/Graph/ProjectInterpretation.cs
cdmihai: Not sure, depends how the abstractions turn out. Right now I tried to keep ProjectGraph pretty agnostic of xml contents. It does not even know about multitargeting. ProjectInterpretation is the one that knows about evaluation particularities, what multitargeting is, etc. The vague intent is to potentially decouple ProjectInterpretation into plugins provided by SDKs.

src/Build/Graph/ProjectGraph.cs
cdmihai: As far as I can tell, not a correctness problem.

src/Build/Graph/ProjectInterpretation.cs
rainersigwald: Sounds like a good plan.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
rainersigwald: Yes

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
rainersigwald: Separate PR sounds fine--can you file an issue please?

src/Build/Graph/ProjectInterpretation.cs
cdmihai: Hopefully in hindsight it still will have sounded like a good plain :)
cc @AndyGerlicher @dannyvv for visibility.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: https://github.com/Microsoft/msbuild/issues/4262

PR #4231, opened 2019-03-12
Replace GetPathsOfAllFilesAbove

tmeschter:
Replace the recently-added `GetPathsOfAllFilesAbove` with a more general set of item intrinsics that can do the same job and more:

1. `GetPathsOfAllDirectoriesAbove`: Given a set of items returns items representing all the ancestor directories.
2. `Combine`: Returns a new set of items with a given relative path appended to all the input items.
3. `Exists`: Filters a set of items to those that actually exist on disk.

With regards to .editorconfig support, we can replace this:

``` xml
<EditorConfigFiles Include="@(Compile->GetPathsOfAllFilesAbove('.editorconfig'))" />
```

with this:

``` xml
<PotentialEditorConfigFiles Include="@(Compile->GetPathsOfAllDirectoriesAbove()->Combine('.editorconfig'))" />
<EditorConfigFiles Include="@(PotentialEditorConfigFiles->Exists())" />
```

This means the project systems do not need to recompute the set of potential editor config files from the .cs inputs. Instead, that can just grab the set of PotentialEditorConfigFiles from the project.


tmeschter:
Hey look, I wrote tests that pass on Linux and OS X this time. :-)

src/Build/Evaluation/Expander.cs
```diff
@@ -2264,15 +2333,12 @@ internal static ItemTransformFunction GetItemTransformFunction(IElementLocation
                         }
                     }
 
-                    // Phase 2: Go through the directories and look for the specified file.
+                    // Phase 2: Go through the directories and return them in order
```
rainersigwald: Think we should specify, guarantee, and test this order?

tmeschter: This is more for my own sanity than anything else, so... not really. :-)

rainersigwald: When we document these we should remember to say "don't depend on the order", just in case.

tmeschter:
@davkean FYI

tmeschter:
@rainersigwald I still want to test this with .editorconfig, but other than that is there any reason to hold this?

rainersigwald:
Nope. I think it’s ready when you are.

tmeschter:
@rainersigwald Merged.

Issue #3051, opened 2018-03-06
Add fastpath for IndexOfAny in Expander.cs


KirillOsenkov:
This throws a MissingMethodException during evaluation
```
RuntimeOS.IndexOfAny(".-0123456789")
```

This is where the case needs to be added:
https://github.com/Microsoft/msbuild/blob/master/src/Build/Evaluation/Expander.cs#L3383

And a test needs to be added here:
https://github.com/Microsoft/msbuild/blob/master/src/Build.UnitTests/Evaluation/Expander_Tests.cs#L3686

KirillOsenkov:
Also:
```
[System.Version]::Parse('$(AssemblyVersion)').ToString(2)
```

KirillOsenkov:
`[System.String]::new('netcore50').Contains('aot')`

KirillOsenkov:
Also:
```
.LastIndexOf(
            '.designer.cs',
            System.StringComparison.OrdinalIgnoreCase)
```

KirillOsenkov:
```
$([System.String]::Copy('%(Identity)').LastIndexOf(
            '.designer.cs',
            System.StringComparison.OrdinalIgnoreCase))
```

PR #4232, opened 2019-03-12
Added more property function fastpaths

BenVillalobos:
Added all functions from #3051

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -3683,6 +3683,12 @@ $(
         }
 
         [Fact]
+        public void PropertyFunctionIndexOfAny()
+        {
+            TestPropertyFunction("$(prop.IndexOfAny('y'))", "prop", "x-y-z", "2");
+        }
+
+        [Fact]
         public void PropertyFunctionStringArrayIndexerGetter()
         {
             TestPropertyFunction("$(prop.Split('-')[0])", "prop", "x-y-z", "x");
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -3689,6 +3689,12 @@ $(
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


rainersigwald:
fyi @KirillOsenkov

src/Build/Evaluation/Expander.cs
```diff
@@ -3933,6 +3987,28 @@ private static bool TryGetArgs(object[] args, out int arg0, out string arg1)
...
+                arg0 = null;
+                arg1 = 0;
+
+                if (args.Length != 2)
+                {
+                    return false;
+                }
+
+                var value1 = args[1] as string;
+                arg0 = args[1] as string;
```
rainersigwald: Something about these indexes seems confused. Should it be 
```suggestion
                arg0 = args[0] as string;
```

?

src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -3682,6 +3682,36 @@ public void PropertyFunctionWithNewLines()
             result.ShouldBe("6C854");
         }
 
+        [Fact]
+        public void PropertyFunctionStringIndexOfAny()
+        {
+            TestPropertyFunction("$(prop.IndexOfAny('y'))", "prop", "x-y-z", "2");
```
rainersigwald: can you add a multi-char input to validate the char-array conversion?

src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -3682,6 +3682,36 @@ public void PropertyFunctionWithNewLines()
...
+        }
+
+        [Fact]
+        public void PropertyFunctionStringLastIndexOf()
+        {
+            TestPropertyFunction("$(prop.LastIndexOf('y'))", "prop", "x-x-y-y-y-z", "8");
+
+            TestPropertyFunction("$(prop.LastIndexOf('y', 7))", "prop", "x-x-y-y-y-z", "6");
```
rainersigwald: I totally understood how this overload worked and didn't have to look it up to figure out how that could be a reasonable answer. Completely.

😳

PR #4236, opened 2019-03-14
Enable COM remote host object

wli3:
more detail in https://github.com/Microsoft/msbuild/issues/4016


wli3:
Sees Windows full framework test failure are unrelated. any idea?

wli3:
@rainersigwald the test failure seems unrelated https://dev.azure.com/dnceng/public/_build/results?buildId=124587&view=ms.vss-test-web.build-test-results-tab and for some reason other similar leg passed any idea?

src/Build.UnitTests/Instance/HostServices_Tests.cs
```diff
@@ -281,6 +295,20 @@ public void TestNonContraditcoryHostObjectAllowed_Any()
             Assert.Equal(NodeAffinity.InProc, hostServices.GetNodeAffinity("project"));
         }
 
+#if FEATURE_COM_INTEROP
+        /// <summary>
+        /// Test which ensures the remote host object cannot affect a project which has the Any affinity specifically set.
+        /// </summary>
+        [Fact]
+        public void TestReigsterRemoteHostObjectNoAffect_Any2()
```
rainersigwald:
```suggestion
        public void TestRegisterRemoteHostObjectNoAffect_Any2()
```

src/Build.UnitTests/Instance/HostServices_Tests.cs
```diff
@@ -281,6 +295,20 @@ public void TestNonContraditcoryHostObjectAllowed_Any()
...
+        {
+            HostServices hostServices = new HostServices();
+            hostServices.SetNodeAffinity("project", NodeAffinity.Any);
+            hostServices.RegisterHostObject("project", "target", "task", "moniker");
+            Assert.Equal(NodeAffinity.Any, hostServices.GetNodeAffinity("project"));
```
rainersigwald: We've been trying to move to Shouldly assertions for new code:

```suggestion
            hostServices.GetNodeAffinity("project").ShouldBe(NodeAffinity.Any);
```

src/Build/Instance/HostServices.cs
```diff
@@ -52,6 +56,9 @@ public class HostServices
         /// </summary>
         private Dictionary<string, NodeAffinity> _projectAffinities;
 
+#if FEATURE_COM_INTEROP
+        private IRunningObjectTableWrapper _runningObjectTable = new RunningObjectTable();
+#endif
```
rainersigwald:
```suggestion
#endif

```

src/Build/Instance/HostServices.cs
```diff
@@ -69,9 +76,30 @@ public ITaskHost GetHostObject(string projectFile, string targetName, string tas
...
+            else
+            {
+                if (monikerNameOrITaskHost.IsMoniker)
+                {
+#if FEATURE_COM_INTEROP
+                    object objectFromRunningObjectTable =
+                        _runningObjectTable.GetObject(monikerNameOrITaskHost.MonikerName);
+                    return (ITaskHost)objectFromRunningObjectTable;
+#else
+                    return null;
```
rainersigwald: Should this be an error instead? Or is there a situation where we'd expect to get a moniker but not understand it?

src/Build/Instance/IRunningObjectTableWrapper.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
```
rainersigwald: Copyright headers in new files, please

src/MSBuild/MSBuild.exe.manifest
```diff
@@ -31,4 +31,18 @@
...
+            tlbid="{D8A9BA71-4724-481D-9CA7-0DA23A1D615C}"
+            version="15.1"
+            helpdir=""/>
+  </file>
+
+  <comInterfaceExternalProxyStub
+        iid="{9049A481-D0E9-414f-8F92-D4F67A0359A6}"
+        name="ITaskHost"
+        tlbid="{D8A9BA71-4724-481D-9CA7-0DA23A1D615C}"
+        proxyStubClsid32="{00020424-0000-0000-C000-000000000046}" />
```
rainersigwald: Can we document the various guids here somehow? They're Greek to me, but I assume have significant meanings.

src/Build/Instance/HostServices.cs
```diff
@@ -69,9 +76,30 @@ public ITaskHost GetHostObject(string projectFile, string targetName, string tas
...
+                return null;
+            }
+            else
+            {
+                if (monikerNameOrITaskHost.IsMoniker)
+                {
+#if FEATURE_COM_INTEROP
+                    object objectFromRunningObjectTable =
+                        _runningObjectTable.GetObject(monikerNameOrITaskHost.MonikerName);
+                    return (ITaskHost)objectFromRunningObjectTable;
```
davidwengier: Are COM exceptions here going to bubble up nicely?

Thinking about all of the failures I had while developing this, there wasn't a heap of information in the build log (ie, what moniker, what task, what target) though that could have been the Project System Tools UI not handling something.

src/Build/Instance/HostServices.cs
```diff
@@ -84,23 +112,45 @@ public void RegisterHostObject(string projectFile, string targetName, string tas
...
 
+#if FEATURE_COM_INTEROP
+        /// <summary>
+        /// Register a remote host object for a particular task/target pair.
+        /// The remote host object require registered in Running Object Table(ROT) already.
+        /// Overwrites any existing host object.
+        ///
+        /// It's caller's responsibly:
+        /// To maintain the live cycle of the host object.
+        /// Register and unregister from ROT.
```
rainersigwald: Should we have a sanity check that the object is registered when we get to this point? Or is that too much?

src/Build/Instance/HostServices.cs
rainersigwald: Yeah, thinking through and at least manually testing failure cases here is a great idea. Ideally we'd get checked-in tests but I don't know how hard it is to get the COM objects to do the testing with/set up our test infra to do it.

src/Build/Instance/HostServices.cs
wli3: I did think about it. To verify registration, we should also make sure they have the correct interface that is used by the task(which we don't know). I think do half of a job is probably the same as no verification and give people false security. Plus we would need to add another COM API "IsRunning"

But I don't hold it strongly

src/MSBuild/MSBuild.exe.manifest
wli3: Is it ok to add comment in manifest file?

src/Build/Instance/HostServices.cs
wli3: > bubble up nicely

I could try to wrap exceptions that may happen from this 2 operations and add moniker, task, target. What other info do you need? Although I think this error should be seen by us only unless the user broke the windows registry. 

> checked-in tests

That is one of the annoying problem. The most fragile part -- CPS, msbuild and the task (from xaml) have all interfaces aligned and they are registered. This need a whole VS setup from clean machine to test.

Since I mocked out runningObjectTable, serialization/deserialization and "put it in msbuild internal dictionary" are covered. Only thing that is not covered is RunningObjectTable.cs and it is very shallow.

If we add end to end test, the best we can have is extra coverage on RunningObjectTable.cs. But we still leave the interface align part off.

davidwengier: > I could try to wrap exceptions that may happen from this 2 operations and add moniker, task, target. What other info do you need?

That would be enough to investigate I'm sure, especially if we get failures from custom targets.

The other question, does an error here fail the whole build, or just the task? or target?

src/Build/Instance/HostServices.cs
```diff
@@ -52,6 +56,9 @@ public class HostServices
         /// </summary>
         private Dictionary<string, NodeAffinity> _projectAffinities;
 
+#if FEATURE_COM_INTEROP
+        private IRunningObjectTableWrapper _runningObjectTable = new RunningObjectTable();
```
davidwengier: A question/discussion on Dispose:

You're not calling it on the RunningObjectTable, and in the CPS code I am. @davkean expressed concern about disposing because it calls Marshal.ReleaseComObject which is a bad thing to do. Looking at [the documentation](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.releasecomobject?view=netframework-4.7.2) it doesn't seem bad to me (it just decrements the ref count, it doesn't ignore it like FinalReleaseComObject) but it also says to use it "only if it is absolutely required".

I'm guessing since the ROT is a singleton none of it matters, but figured we should at least do the same thing in both repos. So no disposal?

davkean: If there are 10 references to a COM on the .NET side, the ref count is only ever 1. When one .NET object calls ReleaseComObject, it's saying "no other .NET code uses this reference". It's better just to let the GC handle this.

davidwengier: Thank you. That's not apparent to me from reading the doco, but it makes sense. I'll remove the disposals on my side :)

davkean: I would delete it so that no one is tempted to call it.

davidwengier: I'm never sure with these things that get copied around to every repository - do you modify them to keep only the bits you call? Or leave them the same so one day someone can find all the dupes and clean them up

rainersigwald: The usual approach in this repo is to aim for minimal diffs, for ease of future maintenance--hopefully a diff can apply cleanly to the various different copies. But it's not a firm rule and I don't know which is better in this particular case.

wli3: i'll remove Dispose method as well.

src/Build/Instance/HostServices.cs
wli3: @rainersigwald correct me if i am wrong. There is nothing to catch this exception, so it will fail the whole build. And that is the same experience during testing

rainersigwald: This gets called from inside a task calling `buildEngine.GetHostObject()` right? In that case, an uncaught exception bubbles up to the task-running code, which catches it, writes an error, and fails the task. The details of what happen next can vary (you can run any task with `ContinueOn="ErrorAndContinue"` or `WarnAndContinue`) but generally a failed task stops execution of the target and marks the target as failed, which stops execution of the project.

src/Build/Instance/HostServices.cs
davidwengier: Just to be clear: I mean I removed the _calls_ to Dispose in the CPS code (which you already don't have). I haven't changed the definition of the RunningObjectTable class

wli3: ah, ok, i won't change that

wli3: that means GC will call Dispose at the correct time?

Pilchie: The ref count will go down *at some point*, but there isn't a guarantee of when.  If you need it to happen at a specific time, use [`Marshal.GetUniqueObjectForIUnknown`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getuniqueobjectforiunknown?view=netframework-4.7.2#System_Runtime_InteropServices_Marshal_GetUniqueObjectForIUnknown_System_IntPtr_), and then `Marshal.ReleaseComObject` on that.  We've had to do that for the Invisible editor, since it saves when it's ref count goes to 0...

davkean: GC does not call Dispose nor know about Dispose. GC knows about Runtime Callable Wrappers (RCW), and will clean them up when they fall out of scope and have no managed references.

src/Build/Instance/HostServices.cs
wli3: throwing HostObjectException now

src/MSBuild/MSBuild.exe.manifest
wli3: Add a separate doc in /documentation instead

src/Build/Instance/HostServices.cs
wli3: I think for now we don't add integration tests

src/Build/Instance/HostObjectException.cs
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Build.Execution
+{
+    [Serializable]
+    internal sealed class HostObjectException : Exception
+    {
+        private const string ErrorMessagePrefix = "Error for HostObject:";
```
wli3: @rainersigwald should we assign a error code for this? And how?

wli3:
@rainersigwald I addressed the previous review. Could you have a look again?

src/Build/Instance/HostObjectException.cs
rainersigwald: Is it a user-facing error, or one that we believe will only ever be because of a bug in VS/MSBuild? If the former, I put a doc up in https://github.com/Microsoft/msbuild/pull/4264 -- please leave comments if it's unclear.

documentation/remote-host-object.md
```diff
@@ -0,0 +1,27 @@
+# Support for remote host object
+
+Remote host object need to be registered in [Running Object Table(ROT)](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable) before hand. And msbuild will call [IRunningObjectTable::GetObject](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nf-objidl-irunningobjecttable-getobject) to get the host object by monikerName which is passed in method `void RegisterHostObject(string projectFile, string targetName, string taskName, string monikerName)`.
```
rainersigwald:
```suggestion
A remote host object must be registered in the [Running Object Table (ROT)](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable) before calling `RegisterHostObject(string projectFile, string targetName, string taskName, string monikerName)`. In the out-of-process node, MSBuild will call [`IRunningObjectTable::GetObject`](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nf-objidl-irunningobjecttable-getobject) to get the host object by the monikerName that was registered via `RegisterHostObject`.
```

documentation/remote-host-object.md
```diff
@@ -0,0 +1,27 @@
+# Support for remote host object
```
rainersigwald: Can you move this to `documentation/specs/`? I think it's more appropriate there.

documentation/remote-host-object.md
```diff
@@ -0,0 +1,27 @@
...
+
+Remote host object need to be registered in [Running Object Table(ROT)](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable) before hand. And msbuild will call [IRunningObjectTable::GetObject](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nf-objidl-irunningobjecttable-getobject) to get the host object by monikerName which is passed in method `void RegisterHostObject(string projectFile, string targetName, string taskName, string monikerName)`.
+
+[The registration of interfaces](https://docs.microsoft.com/en-us/dotnet/framework/interop/how-to-register-primary-interop-assemblies) is the only thing interop with COM that need extra care. Registration could be global with register or without. Other interface needed is registered globally by Visual Studio during installation. Ony `ITaskHost` does not use global registration. On VS side, the proxy is written. But on MsBuild side it was not. That result the change in msbuild.exe.manifest file and the change to generate tlb file for ITaskHost.
+
+## Change in  msbuild.exe.manifest file.
```
rainersigwald:
```suggestion
## Annotated additions to the msbuild.exe.manifest file.
```

documentation/remote-host-object.md
```diff
@@ -0,0 +1,27 @@
...
+```
+<file name="Microsoft.Build.Framework.tlb"> --Location of the tlb, it should be in the same directory of msbuild.exe
```
rainersigwald:
```suggestion
<file name="Microsoft.Build.Framework.tlb"> -- Location of the tlb, it should be in the same directory as msbuild.exe
```

rainersigwald: This raises an interesting point: **do we need to support this for amd64 MSBuild?**

Currently you're placing the `.tlb` only in the 32-bit MSBuild location, but VS can (if the appropriate hoops are jumped through) execute 64-bit tasks.

documentation/remote-host-object.md
```diff
@@ -0,0 +1,27 @@
+# Support for remote host object
+
+Remote host object need to be registered in [Running Object Table(ROT)](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable) before hand. And msbuild will call [IRunningObjectTable::GetObject](https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nf-objidl-irunningobjecttable-getobject) to get the host object by monikerName which is passed in method `void RegisterHostObject(string projectFile, string targetName, string taskName, string monikerName)`.
+
+[The registration of interfaces](https://docs.microsoft.com/en-us/dotnet/framework/interop/how-to-register-primary-interop-assemblies) is the only thing interop with COM that need extra care. Registration could be global with register or without. Other interface needed is registered globally by Visual Studio during installation. Ony `ITaskHost` does not use global registration. On VS side, the proxy is written. But on MsBuild side it was not. That result the change in msbuild.exe.manifest file and the change to generate tlb file for ITaskHost.
```
rainersigwald: I don't understand this. Is this a history of design decisions, or an explanation of relevant facts for the "before" situation? Both are super helpful!

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
```
rainersigwald: Copyright header here too, please.

src/Build/Instance/HostObjectException.cs
wli3: It is later. Only VS or MSBuild dev should see this. If the user see this, it is a bug

documentation/remote-host-object.md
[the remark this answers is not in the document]
wli3: https://github.com/dotnet/project-system/issues/4406 has a good write up. I will just paste it here?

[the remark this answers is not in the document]
rainersigwald: You can probably just link it. Just needs a bit more context.

documentation/remote-host-object.md
wli3: For this work only 32 bit is needed. From @davkean

src/Build/Instance/HostObjectException.cs
rainersigwald: Then I'd probably derive this from `InternalErrorException` and **not** assign it a code.

documentation/remote-host-object.md
rainersigwald: Awesome. We should probably make that explicit -- if a host object is registered by moniker and accessed in amd64 MSBuild, throw?

documentation/remote-host-object.md
wli3: My bad writing :( this are all facts. There is no other alternative solutions. Let me see if I can write it better

wli3:
And I will squash merge it

src/Build/Instance/HostObjectException.cs
rainersigwald: @wli3 pointed out offline that `InternalErrorException` a) is `sealed` and b) doesn't really convey more information here, so I agree that we should leave it as is.

PR #4237, opened 2019-03-14
More property function fastpaths

BenVillalobos:
Added all functions causing reflection when creating/building a new console project.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -3804,6 +3804,62 @@ namespace Microsoft.Build.Evaluation
                                 return true;
                             }
                         }
+                        else if (string.Equals(_methodMethodName, nameof(IntrinsicFunctions.GetCurrentToolsDirectory), StringComparison.OrdinalIgnoreCase))
+                        {
+                            if (args.Length == 0)
+                            {
+                                returnVal = IntrinsicFunctions.GetCurrentToolsDirectory();
+                                return true;
+                            }
+                        }
+                        else if (string.Equals(_methodMethodName, nameof(IntrinsicFunctions.GetToolsDirectory32), StringComparison.OrdinalIgnoreCase))
+                        {
+                            if (args.Length == 0)
+                            {
+                                returnVal = IntrinsicFunctions.GetToolsDirectory32();
+                                return true;
+                            }
+                        }
+                        else if (string.Equals(_methodMethodName, nameof(IntrinsicFunctions.GetToolsDirectory64), StringComparison.OrdinalIgnoreCase))
+                        {
+                            if (args.Length == 0)
+                            {
+                                returnVal = IntrinsicFunctions.GetToolsDirectory64();
+                                return true;
+                            }
+                        }
+                        else if (string.Equals(_methodMethodName, nameof(IntrinsicFunctions.GetMSBuildSDKsPath), StringComparison.OrdinalIgnoreCase))
+                        {
+                            if (args.Length == 0)
+                            {
+                                returnVal = IntrinsicFunctions.GetMSBuildSDKsPath();
+                                return true;
+                            }
+                        }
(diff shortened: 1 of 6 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -3618,6 +3618,62 @@ private bool TryExecuteWellKnownFunction(out object returnVal, object objectInst
                                 return true;
                             }
                         }
+                        else if (string.Equals(_methodMethodName, nameof(IntrinsicFunctions.GetCurrentToolsDirectory), StringComparison.OrdinalIgnoreCase))
```
danmosemsft: The number of if checks here is starting to get large. Also I am not sure whether OrdinalIgnoreCase means that strings of different lengths must be compared (?)

If this starts showing up in the profile it you could potentially store all the candidates upper cased, and upper case the test string then make all the tests using Ordinal comparison.

rainersigwald: Agreed that we should keep an eye out for this in the profiler, and if it does show up we have a few options for perf: length restrictions, prefix matching, generating this code from metadata to build a more-optimal matcher (that sounds pretty fun, actually). But I'm optimistic that this will only help, since reflection is so expensive.

Re: length checks, since the APIs we can fast-path are all ANSI strings (because they're mostly public API from corefx/MSBuild itself, which is all English) we should be able to short-circuit to failure on strings of different length. I'm fine with someone having a slow experience if they try to call `$([System.String]::😬😱🤦🏻‍♂️())`.

nguerrera: OrdinalIgnoreCase already bails on unequal lengths: https://github.com/dotnet/coreclr/blob/9c4d496071c2ffdc5a8793835694d1a8ac267522/src/System.Private.CoreLib/shared/System/String.Comparison.cs#L666-L670

IIRC, it is effectively Ordinal with each `char` going through `ToUpperVariant()` before being compared, and ToUpperInvariant() cannot turn one `char` into two.

nguerrera: If this is actually as bottleneck, some sort of lookup table on (receiver type, method name), maybe +args, would probably be better. It might also clean up the code.

nguerrera: I'm thinking something like 
``` C#
 void RegisterIntrinsicFastPath(Type receiver, string methodName, Func<object> handler) { /* ... */ }
 void RegisterIntrinsicFastPath<TArg0>(Type receiver, string methodName, Func<TArg0, object> handler) { /* ... */}
 void RegisterIntrinsicFastPath<TArg0, TArg1>(Type receiver, string methodName, Func<TArg0, TArg1, object> handler) { /* ... */ }
```

Used like this:
 ```
RegisterIntrinsicFastPath(typeof(string), nameof(string.Equals), (string x, string y) => string.Equals(x, y));
```

This would also be fun to implement, but yeah, unless there's evidence these comparisons are causing a problem, we can leave that for another day.

Comment (danmosemsft) [reply src/Build/Evaluation/Expander.cs]: Another option if needed is a hand-tuned switch by length like here
https://github.com/Microsoft/msbuild/issues/3466

I'll mark this resolved though.

File: src/Build/Microsoft.Build.csproj [hunk 2]
```diff
@@ -30,6 +30,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Framework\Microsoft.Build.Framework.csproj" />
+    <ProjectReference Include="..\Utilities\Microsoft.Build.Utilities.csproj" />
```
rainersigwald: This is the only part of the PR that I'm worried about. It's a change in our assembly dependency graph and thus our package dependency graph, and it might result in unexpected loading of the Utilities assembly on new paths.

Can you split this PR into the things that don't need this and the things that do? I think we should take the former and wait on the latter--probably pulling it in for 16.2 preview 1.

PR #4243, opened 2019-03-17
Add up-to-date logging for GenerateResource

danmosemsft:
Fix https://github.com/Microsoft/msbuild/issues/2950

If resource generation runs, the compile will in turn have to run, and then every downstream project. However resource generation doesn't log why it ran. This make it unnecessarily difficult to know why an overbuild is happening, such as one I'm looking at right now in my repo.

I used the same strings as for target up to date checking so it's easier to search the log (eg for "is newer than output file"). I also used the same low verbosity that those messages have. 

If the resources are all up to date, there is essentially no extra work, and no extra logging. If they are not, at most this will generate 1 extra low verbosity event per resource file. Either way, there is no new IO.

Also a trivial commit that extracts getting file extension in 2 places.

### DIFF
--- a/src/Tasks/GenerateResource.cs
+++ b/src/Tasks/GenerateResource.cs
@@ -961,8 +961,9 @@ namespace Microsoft.Build.Tasks
             // By default all file types that get here are considered dangerous
             bool dangerous = true;
 
-            if (String.Equals(Path.GetExtension(filename), ".resx", StringComparison.OrdinalIgnoreCase) ||
-                String.Equals(Path.GetExtension(filename), ".resw", StringComparison.OrdinalIgnoreCase))
+            string extension = Path.GetExtension(filename);
+            if (String.Equals(extension, ".resx", StringComparison.OrdinalIgnoreCase) ||
+                String.Equals(extension, ".resw", StringComparison.OrdinalIgnoreCase))
             {
                 // XML files are only dangerous if there are unrecognized objects in them
                 dangerous = false;
@@ -1521,8 +1522,9 @@ namespace Microsoft.Build.Tasks
 
             DateTime sourceTime = NativeMethodsShared.GetLastWriteFileUtcTime(sourceFilePath);
 
-            if (!String.Equals(Path.GetExtension(sourceFilePath), ".resx", StringComparison.OrdinalIgnoreCase) &&
-                !String.Equals(Path.GetExtension(sourceFilePath), ".resw", StringComparison.OrdinalIgnoreCase))
+            string extension = Path.GetExtension(sourceFilePath);
+            if (!String.Equals(extension, ".resx", StringComparison.OrdinalIgnoreCase) &&
+                !String.Equals(extension, ".resw", StringComparison.OrdinalIgnoreCase))
             {
                 // If source file is NOT a .resx, for example a .restext file, 
                 // timestamp checking is simple, because there's no linked files to examine, and no references.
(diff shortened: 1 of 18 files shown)


src/Tasks/GenerateResource.cs
```diff
@@ -1695,58 +1724,38 @@ private void GetStronglyTypedResourceToProcess(ref List<ITaskItem> inputsToProce
...
-                        DateTime referenceTime = NativeMethodsShared.GetLastWriteFileUtcTime(reference.ItemSpec);
+            // Check the timestamp of each of the passed-in references to find the newest;
+            // and then the additional inputs
+            var inputs = (this.References ?? Enumerable.Empty<ITaskItem>()).Concat(this.AdditionalInputs ?? Enumerable.Empty<ITaskItem>());
 
-                        if (referenceTime == DateTime.MinValue)
-                        {
-                            // File does not exist: force a build to produce an error message
-                            _foundNewestUncorrelatedInputWriteTime = true;
-                            return DateTime.MaxValue;
```
danmosemsft: Note this method was previously returning MaxValue to force a build if any input was missing.

For clarity I changed to consistently using MinValue to mean a file is missing (same as GetLastWriteFileUtcTime returns for a missing file) and explicitly checking for MinValue as needed.

danmosemsft:
cc @KirillOsenkov

danmosemsft:
@rainersigwald the CI errors are unrelated issues relating to max path tests - perhaps your CI machines got updated to a Windows version that enables paths longer than max path.

danmosemsft:
Thanks for fixing the tests @rainersigwald ! Are we missing a test for input doesn't exist and linked input doesn't exist?

rainersigwald:
I swear I ran the tests locally . . . looking

src/Tasks/GenerateResource.cs
```diff
@@ -1660,16 +1681,23 @@ private void GetStronglyTypedResourceToProcess(ref List<ITaskItem> inputsToProce
...
             DateTime outputTime = NativeMethodsShared.GetLastWriteFileUtcTime(StronglyTypedFileName);
 
-            if (sourceTime == DateTime.MinValue || outputTime == DateTime.MinValue)
+            if (sourceTime == DateTime.MinValue)
             {
                 // Source file is missing - force a build, so that resource generation
-                // will produce a nice error message; or output file is missing,
-                // need to build it
+                // will produce a nice error message
+                Log.LogMessageFromResources(MessageImportance.Low, "GenerateResource.InputDoesntExist", Sources[0].ItemSpec);
```
rainersigwald: Were you able to hit this codepath? My naive attempt to do so failed, because an earlier error fired and we never got to here.

danmosemsft: I can't remember. Feel free to remove if you like, or I can?

danmosemsft:
Do you need more than me @rainersigwald?

rainersigwald:
@danmosemsft You're good on this; I'll merge it shortly.

danmosemsft:
Thanks @rainersigwald !

danmosemsft:
Friendly ping 🙂 I added this because I would like anyone to be able to use it in our repos.

rainersigwald:
Thanks for the fix, and your patience getting it in!

danmosemsft:
Yay thanks @rainersigwald !

Issue #4247, opened 2019-03-19
MSBuild task tests fail when long-paths enabled


rainersigwald:
This test

https://github.com/Microsoft/msbuild/blob/94e11e0a773bc8956caf128335433231bb06fed5/src/Build.UnitTests/BackEnd/MSBuild_Tests.cs#L30-L93

Fails when long paths are enabled.

I think the reason is that there is a codepath that attempts to handle too-long-but-can-be-simplified-to-short-enough paths that has different behavior than actually just passing the long paths.

https://github.com/Microsoft/msbuild/blob/94e11e0a773bc8956caf128335433231bb06fed5/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/MSBuild.cs#L314

Disabling the tests against this bug.

Issue #4241, opened 2019-03-15
FileUtilities tests fail on long-path-enabled Windows


rainersigwald:
### Steps to reproduce

On a machine with Windows long paths enabled, several tests fail, all due to this

https://github.com/Microsoft/msbuild/blob/94e11e0a773bc8956caf128335433231bb06fed5/src/Shared/NativeMethodsShared.cs#L1589

returning a different result when the path is too long vs. under MAX_PATH.

PR #4248, opened 2019-03-19
Longpath test failures

rainersigwald:
It looks like the agents that Azure DevOps Pipelines use have recently flipped the registry key to enable long-path support in standard Windows APIs.

This causes several of our tests to fail, because we're testing that when we hit long-path situations, the MSBuild error produced is nice. But now there's not always an error!

I added a condition for those tests and also disabled a couple of others (for which I filed bugs: #4241, #4247). This should get us back to green CI builds.

@livarcocc this almost-test-only change (the only product-code change is `private` to `internal` for `IsMaxPathLegacyWindows`) targets vs16.0 based on our offline conversation last week: since this is a change in the hosted CI builders, it'll affect 16.0 servicing, so we should take it proactively now so that any future PR into vs16.0 has a chance of being green.

fyi @ccastanedaucf

### DIFF
--- a/documentation/specs/static-graph.md
+++ b/documentation/specs/static-graph.md
@@ -155,7 +155,7 @@ To enable further optimizations (and strictness), graph builds can run [isolated
 This is a proposal for what the public API for ProjectGraph may look like:
 
 ```csharp
-namespace Microsoft.Build.Graph
+namespace Microsoft.Build.Experimental.Graph
 {
     public class ProjectGraph
     {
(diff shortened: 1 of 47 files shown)


rainersigwald:
Retargeted to `master` while @livarcocc clears checking into the 16.0 branch with the appropriate folks. I'll do an explicit merge and we can easily merge again to the 16.0 branch later.

rainersigwald:
This supersedes #4238 by pullin in all of those changes. Seems ok to me, so I'm going to do it.

PR #4253, opened 2019-03-20
Remove YAML variable definitions

rainersigwald:
YAML variable definitions override queue-time variables, so these must
not be set in this way.

Docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#allow-at-queue-time

### DIFF
--- a/.vsts-dotnet.yml
+++ b/.vsts-dotnet.yml
@@ -2,6 +2,13 @@ trigger:
 - master
 - vs*
 
+# If defined here, these values are not overrideable
+# Once they exist, we should define these as "runtime parameters"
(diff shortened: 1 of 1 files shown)


tmat:
Yup, DevOps just started to enforce this :(. I'd recommend keeping the variables in the definition but comment them out. This is to document that the definition expects these to be set.

See also https://github.com/dotnet/roslyn/blob/master/azure-pipelines-official.yml#L5

rainersigwald:
With this, an official build (with no optprof) succeeded. Hooray!

(I'm glad there's at least a rough roadmap for the future here; I agree that the current behavior is surprising.)

PR #4259, opened 2019-03-22
WriteLinesToFile Task documentation update

### LINKED ISSUE #4257: Edcoding Property in WriteLinesToFile Task is missing from xsd and has incorrect definition
### Steps to reproduce

Either include a project sample, attach a zipped project, or provide IDE / CLI steps to create the project and repro the behaviour. Example of a project sample:

 <xs:element name="WriteLinesToFile" substitutionGroup="msb:Task">
        <xs:complexType>
            <xs:complexContent>
                <xs:extension base="msb:TaskType">
                    <xs:attribute name="Encoding" />
                    <xs:attribute name="File" use="required" />
                    <xs:attribute name="Lines" />
                    <xs:attribute name="Overwrite" type="msb:boolean" />
                </xs:extension>
            </xs:complexContent>
        </xs:complexType>
    </xs:element>

Below 2 properties are missing from xsd , also notice the comment for Encoding
 /// <summary>
        /// If true, overwrite any existing file contents.
        /// </summary>
        public string Encoding { get; set; }

        /// <summary>
        /// If true, the target file specified, if it exists, will be read first to compare against
        /// what the task would have written. If identical, the file is not written to disk and the
        /// timestamp will be preserved.
        /// </summary>
        public bool WriteOnlyWhenDifferent { get; set; }
contents 

### Expected  behavior

The properties should exists in xsd and encoding property should have correct comment

### Actual behavior
properties missing from xsd and has incorrect definition for property

### Environment data
`msbuild /version` output:

OS info:

If applicable, version of the tool that invokes MSBuild (Visual Studio, dotnet CLI, etc):

santoshyadav198613:
Fixes #4257

### DIFF
--- a/src/MSBuild/Microsoft.Build.CommonTypes.xsd
+++ b/src/MSBuild/Microsoft.Build.CommonTypes.xsd
@@ -2961,6 +2961,7 @@ elementFormDefault="qualified">
                     <xs:attribute name="File" use="required" />
                     <xs:attribute name="Lines" />
                     <xs:attribute name="Overwrite" type="msb:boolean" />
+                    <xs:attribute name="WriteOnlyWhenDifferent" type="msb:boolean" />
                 </xs:extension>
             </xs:complexContent>
         </xs:complexType>
--- a/src/Tasks/FileIO/WriteLinesToFile.cs
+++ b/src/Tasks/FileIO/WriteLinesToFile.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Build.Tasks
         public bool Overwrite { get; set; }
 
         /// <summary>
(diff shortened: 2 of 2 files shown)


santoshyadav198613:
> Can you please split these into two different commits, one for the XSD update and one for the doc comment change?

Sure. sorry i thought you mentioned this and another PR so got confused. Will revert and raise 2 separate PR.

rainersigwald:
It can be in the same PR, but I'd prefer two individual commits.

santoshyadav198613:
> It can be in the same PR, but I'd prefer two individual commits.

Done made 2 separate commits for each file.

PR #4261, opened 2019-03-22
Shorter node utilization histogram bars

grendello:
Limit the number of histogram bar segments. For long runs the number of segments
can be counted in hundreds of thousands (for instance a build which took 8061.7s
would generate a single line line 161,235 characters long) which is a bit
excessive and makes viewing text logs with histograms quite painful at times.

This commit implements a scaled approach in which the number of bar elements
calculated with the original formula is subdivided as follows:

 * <= 100 segments remains as-is, scale 1:1, character used: `.`
 * &gt; 100 <= 1000 segments, scale 1:100, character used: `+`
 * &gt; 1000 segments, scale 1:1000, character used: `#`

Additionaly, scale is printed after the histogram bar if it's different than `1:1`

### DIFF
--- a/src/Build/BackEnd/Components/Scheduler/Scheduler.cs
+++ b/src/Build/BackEnd/Components/Scheduler/Scheduler.cs
@@ -2229,7 +2229,7 @@ namespace Microsoft.Build.BackEnd
                 scale = 1000;
             }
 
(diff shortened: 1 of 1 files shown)


rainersigwald:
Thanks, @grendello!

PR #4275, opened 2019-03-27
Reduce reader-writer lock contention in ProjectCollection

erikma:
- Updated HostServices getter to use a reader lock and only upgrade to a writer when creating a value. This is a hotspot in parallel creation of projects in paths like this:
   at Microsoft.Build.Evaluation.ProjectCollection.DisposableReaderWriterLockSlim.EnterWriteLock()
   at Microsoft.Build.Evaluation.ProjectCollection.get_HostServices()
   at Microsoft.Build.Evaluation.Project.CreateProjectInstance(ILoggingService loggingServiceForEvaluation, ProjectInstanceSettings settings)
- Moved checks on param values, and pure functional modifications to parameters, out of locks.
- Moved invariant allocations out of locks where possible.
- Updated ToolsetLocations to a read-only property since it is only set in the constructor, eliminating locking on that property.

### DIFF
--- a/src/Build/Definition/ProjectCollection.cs
+++ b/src/Build/Definition/ProjectCollection.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Build.Evaluation
         // all need lock protection, but there are a lot of read cases as well, and calls to create Projects
         // call back to the ProjectCollection under locks. Use a RW lock, but default to always using
         // upgradable read locks to avoid adding reentrancy bugs.
-        private struct DisposableReaderWriterLockSlim
+        private class DisposableReaderWriterLockSlim
         {
             private readonly ReaderWriterLockSlim _lock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
             public bool IsWriteLockHeld => _lock.IsWriteLockHeld;
--- a/src/Build/Definition/ProjectCollection.cs
+++ b/src/Build/Definition/ProjectCollection.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Build.Evaluation
         // all need lock protection, but there are a lot of read cases as well, and calls to create Projects
         // call back to the ProjectCollection under locks. Use a RW lock, but default to always using
         // upgradable read locks to avoid adding reentrancy bugs.
-        private class DisposableReaderWriterLockSlim
+        private struct DisposableReaderWriterLockSlim
         {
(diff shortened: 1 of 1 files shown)


erikma:
@dfederm since we were looking at a related exception stack

src/Build/Definition/ProjectCollection.cs
```diff
@@ -621,18 +610,20 @@ public bool IsBuildEnabled
...
-
-                        eventArgs = new ProjectCollectionChangedEventArgs(ProjectCollectionChangedState.IsBuildEnabled);
+                        sendEvent = true;
                     }
                 }
 
-                OnProjectCollectionChangedIfNonNull(eventArgs);
+                if (sendEvent)
+                {
+                    OnProjectCollectionChanged(new ProjectCollectionChangedEventArgs(ProjectCollectionChangedState.IsBuildEnabled));
```
dfederm: `OnProjectCollectionChangedIfNonNull` seems like it can be deleted now.

erikma: true

src/Build/Definition/ProjectCollection.cs
```diff
@@ -1041,11 +1056,12 @@ public Project LoadProject(string fileName, string toolsVersion)
         /// <returns>A loaded project.</returns>
         public Project LoadProject(string fileName, IDictionary<string, string> globalProperties, string toolsVersion)
         {
+            ErrorUtilities.VerifyThrowArgumentLength(fileName, nameof(fileName));
+            fileName = FileUtilities.NormalizePath(fileName);
+
+            var buildEventContext = new BuildEventContext(0 /* node ID */, BuildEventContext.InvalidTargetId, BuildEventContext.InvalidProjectContextId, BuildEventContext.InvalidTaskId);
```
dfederm: Looks like this is only used in the exceptional case. Consider only allocating in the `catch` block.

erikma: nice catch

erikma:
@rainersigwald No, no appreciable perf win to report out. This is in reaction to a ~0.1% failure rate spate of problems we're seeing in QuickBuild with MSBuild parsing mixed with all the other massive thread-parallel work going on, the WaitHandle underneath the ReaderWriteLock sometimes throws when the system gets heavily loaded. It led me to look at the code as to why a read lock was being used on a property getter used on every thread, then that expanded to the rest of the PR on general principle.

rainersigwald:
Thanks @erikma.

PR #4295, opened 2019-04-08
Allow overriding the optprof source branch

rainersigwald:
When the source branch was hardcoded to `$(Build.SourceBranchName)`,
there was no good way to tell the build "build this branch, but take the
latest OptProf information from branch X". That's an important scenario
when standing up a new branch (like a release branch), or trying to
insert updated bits from a feature branch.

This requires a change to the build definition, which is the only place
(currently) that knows what variables to offer at job-scheduling time.

See https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=2568755 for an example of the override working:

```
Acquiring optimization data
Downloading optimization data from service https://devdiv.artifacts.visualstudio.com, drop OptimizationData/Microsoft/msbuild/master/20190405.1/295317/1
```

### DIFF
--- a/.vsts-dotnet.yml
+++ b/.vsts-dotnet.yml
@@ -67,7 +67,7 @@ jobs:
               -officialBuildId $(Build.BuildNumber)
               -officialSkipApplyOptimizationData $(SkipApplyOptimizationData)
               /p:RepositoryName=$(Build.Repository.Name)
-              /p:VisualStudioIbcSourceBranchName=$(Build.SourceBranchName)
+              /p:VisualStudioIbcSourceBranchName=$(IbcSourceBranchName)
               /p:VisualStudioDropAccessToken=$(System.AccessToken)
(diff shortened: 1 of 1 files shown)


rainersigwald:
@tmat, any concerns about adding this layer of indirection?

tmat:
Looks good. BTW, you might want to consider using a similar branch configuration mechanism as Roslyn does:

We configure our branches in https://github.com/dotnet/roslyn/blob/master/eng/config/PublishData.json

The only relevant version of this file is the one in the master branch.

Then we have PS functions that read that file and figure out what the right deafult IBC data branch and VS installer branch are like so: https://github.com/dotnet/roslyn/blob/master/eng/build.ps1#L272-L328

rainersigwald:
The idea of an automated indirection makes a lot of sense for the future, but I think we can defer it until we feel the pain more acutely (fortunately, we tend to have fewer branches in flight than Roslyn).

PR #4300, opened 2019-04-10
Change to MTA when creating ROT

wli3:
As a result to wait for the separate thread. It need to be async

Also make ROT init Lazy

### DIFF
--- a/src/Build/Instance/RunningObjectTable.cs
+++ b/src/Build/Instance/RunningObjectTable.cs
@@ -45,25 +45,12 @@ namespace Microsoft.Build.Execution
         /// </summary>
         public object GetObject(string itemName)
         {
-            Task<IMoniker> mkTask = CreateMoniker(itemName);
-            Task.WaitAll(_rotTask, mkTask);
-            var rot = _rotTask.Result;
+            var rot = _rotTask.GetAwaiter().GetResult();
 
-            int hr = rot.GetObject(mkTask.Result, out object obj);
-            if (hr != 0)
-            {
-                Marshal.ThrowExceptionForHR(hr);
-            }
-
-            return obj;
-        }
-
-        private Task<IMoniker> CreateMoniker(string itemName)
-        {
+            IMoniker moniker;
             if (Thread.CurrentThread.GetApartmentState() == ApartmentState.MTA)
             {
-                Ole32.CreateItemMoniker("!", itemName, out var moniker);
-                return Task.FromResult(moniker);
+                Ole32.CreateItemMoniker("!", itemName, out moniker);
             }
             else
             {
@@ -75,8 +62,16 @@ namespace Microsoft.Build.Execution
                     return mk;
                 });
 
-                return task;
+                moniker = task.GetAwaiter().GetResult();
             }
+
+            int hr = rot.GetObject(moniker, out object obj);
+            if (hr != 0)
+            {
+                Marshal.ThrowExceptionForHR(hr);
+            }
+
+            return obj;
         }
 
         private static class Ole32
--- a/src/Build/Instance/RunningObjectTable.cs
+++ b/src/Build/Instance/RunningObjectTable.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.ComTypes;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Microsoft.Build.Execution
 {
@@ -14,25 +16,28 @@ namespace Microsoft.Build.Execution
     /// <remarks>
     /// See https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable.
     /// </remarks>
-    internal class RunningObjectTable : IDisposable, IRunningObjectTableWrapper
+    internal class RunningObjectTable : IRunningObjectTableWrapper
     {
-        private readonly IRunningObjectTable rot;
-        private bool isDisposed = false;
+        private readonly Task<IRunningObjectTable> _rotTask;
 
         public RunningObjectTable()
         {
-            Ole32.GetRunningObjectTable(0, out this.rot);
-        }
-
-        public void Dispose()
-        {
-            if (this.isDisposed)
+            if (Thread.CurrentThread.GetApartmentState() == ApartmentState.MTA)
             {
-                return;
+                Ole32.GetRunningObjectTable(0, out var rot);
+                _rotTask = Task.FromResult(rot);
             }
-
-            Marshal.ReleaseComObject(this.rot);
-            this.isDisposed = true;
+            else
+            {
+                // To avoid deadlock, create ROT in a threadpool threads which guarantees to be MTA. And the
+                // object will be MTA
+                _rotTask =
+                Task.Run(() =>
+                    {
+                        Ole32.GetRunningObjectTable(0, out var rot);
+                        return rot;
+                    });
+            }
         }
 
         /// <summary>
@@ -40,8 +45,11 @@ namespace Microsoft.Build.Execution
         /// </summary>
         public object GetObject(string itemName)
         {
-            IMoniker mk = CreateMoniker(itemName);
-            int hr = this.rot.GetObject(mk, out object obj);
+            Task<IMoniker> mkTask = CreateMoniker(itemName);
+            Task.WaitAll(_rotTask, mkTask);
+            var rot = _rotTask.Result;
+
+            int hr = rot.GetObject(mkTask.Result, out object obj);
             if (hr != 0)
             {
                 Marshal.ThrowExceptionForHR(hr);
@@ -50,10 +58,25 @@ namespace Microsoft.Build.Execution
             return obj;
         }
 
-        private IMoniker CreateMoniker(string itemName)
+        private Task<IMoniker> CreateMoniker(string itemName)
         {
-            Ole32.CreateItemMoniker("!", itemName, out IMoniker mk);
-            return mk;
+            if (Thread.CurrentThread.GetApartmentState() == ApartmentState.MTA)
+            {
+                Ole32.CreateItemMoniker("!", itemName, out var moniker);
+                return Task.FromResult(moniker);
+            }
+            else
+            {
+                // To avoid deadlock, create Moniker in a threadpool threads which guarantees to be MTA. And the
+                // object will be MTA
+                var task = Task.Run(() =>
+                {
+                    Ole32.CreateItemMoniker("!", itemName, out var mk);
+                    return mk;
+                });
+
+                return task;
+            }
         }
 
         private static class Ole32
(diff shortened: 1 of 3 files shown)


src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,42 @@ namespace Microsoft.Build.Execution
...
         {
-            Ole32.GetRunningObjectTable(0, out this.rot);
-        }
-
-        public void Dispose()
-        {
-            if (this.isDisposed)
+            // To avoid deadlock, create rot in a MTA thread. And the
+            // object will be MTA
+            var tcs = new TaskCompletionSource<IRunningObjectTable>();
```
wli3: Please help me check if I use task, thread correctly

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -50,10 +60,27 @@ public object GetObject(string itemName)
...
-            Ole32.CreateItemMoniker("!", itemName, out IMoniker mk);
-            return mk;
+            // To avoid deadlock, create rot in a MTA thread. And the
+            // object will be MTA
+            var tcs = new TaskCompletionSource<IMoniker>();
+            Thread thread = new Thread(() =>
+            {
+                try
+                {
+                    Ole32.CreateItemMoniker("!", itemName, out IMoniker mk);
```
wli3: Please help me check if I use task, thread correctly.
And CreateItemMoniker should also wrap in MTA thread right?

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,42 @@ namespace Microsoft.Build.Execution
...
-        private readonly IRunningObjectTable rot;
-        private bool isDisposed = false;
+        private Task<IRunningObjectTable> _runningObjectTableTask;
 
         public RunningObjectTable()
         {
-            Ole32.GetRunningObjectTable(0, out this.rot);
-        }
-
-        public void Dispose()
```
wli3: This will never get called @AaronRobinsonMSFT

src/Build/Instance/RunningObjectTable.cs
AaronRobinsonMSFT: >And CreateItemMoniker should also wrap in MTA thread right?

Yep. Good catch.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -50,10 +60,27 @@ public object GetObject(string itemName)
             return obj;
         }
 
-        private IMoniker CreateMoniker(string itemName)
+        private Task<IMoniker> CreateMoniker(string itemName)
         {
-            Ole32.CreateItemMoniker("!", itemName, out IMoniker mk);
-            return mk;
+            // To avoid deadlock, create rot in a MTA thread. And the
```
AaronRobinsonMSFT: Comment is copy/paste

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,42 @@ namespace Microsoft.Build.Execution
...
+                }
+                catch (Exception e)
+                {
+                    tcs.SetException(e);
+                }
+            });
+            thread.SetApartmentState(ApartmentState.MTA);
+            thread.Start();
+
+            _runningObjectTableTask = tcs.Task;
```
AaronRobinsonMSFT: Remove extra line.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,42 @@ namespace Microsoft.Build.Execution
     /// <remarks>
     /// See https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable.
     /// </remarks>
-    internal class RunningObjectTable : IDisposable, IRunningObjectTableWrapper
+    internal class RunningObjectTable : IRunningObjectTableWrapper
     {
-        private readonly IRunningObjectTable rot;
-        private bool isDisposed = false;
+        private Task<IRunningObjectTable> _runningObjectTableTask;
```
AaronRobinsonMSFT: Mark the task as `readonly`.

ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -1050,7 +1050,7 @@ public partial class BuildSubmission
     public partial class HostServices
     {
         public HostServices() { }
-        public Microsoft.Build.Framework.ITaskHost GetHostObject(string projectFile, string targetName, string taskName) { throw null; }
+        public System.Threading.Tasks.Task<Microsoft.Build.Framework.ITaskHost> GetHostObject(string projectFile, string targetName, string taskName) { throw null; }
```
rainersigwald: Isn't this a breaking change for VS? My instinct is to hide all of this and call `Thread.Join()` to wait on the result of the remote thread. Is that feasible?

wli3:
@AaronRobinsonMSFT we realize use async will have breaking change in VS due to HostServices being public. I end up using Thread.Join at the first use of ROT (Thread.Join instead of Task to avoid ConfigureAwait(false))

AaronRobinsonMSFT:
@wli3 I see. This change could become very interesting. We are now introducing an explicit `Join` on the UI thread which _will_ alter the scheduling of the UI thread and could move JITs as well as GCs to different locations that are measured by the VS RPS system. I encourage you to reach out to @AArnott and other VS threading experts to hear their thoughts on potential alternatives and determine if this could impact a very sensitive VS scenario.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,29 @@ namespace Microsoft.Build.Execution
...
-            {
-                return;
-            }
-
-            Marshal.ReleaseComObject(this.rot);
-            this.isDisposed = true;
+            // To avoid deadlock, create ROT in a MTA thread. And the
+            // object will be MTA
+            _initRotThread =
+                new Thread(() => Ole32.GetRunningObjectTable(0, out _rot));
```
AArnott: Such a short-lived thread seems wasteful. And if an exception is thrown from it, it will crash the process. Why not use Task.Run for this, and store the `Task<IRunningObjectTable>` as a field instead of the `Thread`?
Also, check the current thread's apartment and do it yourself directly if it's already an MTA.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -52,7 +49,15 @@ public object GetObject(string itemName)
 
         private IMoniker CreateMoniker(string itemName)
         {
-            Ole32.CreateItemMoniker("!", itemName, out IMoniker mk);
+            IMoniker mk = null;
+
+            // To avoid deadlock, create Moniker in a MTA thread. And the
+            // object will be MTA
+            Thread thread = new Thread(() => Ole32.CreateItemMoniker("!", itemName, out mk));
```
AArnott: Again, creating a thread is relatively expensive. Why not use `Task.Run` to leverage the thread pool?
Also, check the current thread's apartment and do it yourself directly if it's already an MTA.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,29 @@ namespace Microsoft.Build.Execution
     /// <remarks>
     /// See https://docs.microsoft.com/en-us/windows/desktop/api/objidl/nn-objidl-irunningobjecttable.
     /// </remarks>
-    internal class RunningObjectTable : IDisposable, IRunningObjectTableWrapper
+    internal class RunningObjectTable : IRunningObjectTableWrapper
```
AArnott: If `IRunningObjectTableWrapper` is an internal interface, and if the callers are or could be async, consider revising this interface to define async methods so you don't have to block the caller's thread while waiting for another thread.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -50,10 +58,25 @@ public object GetObject(string itemName)
             return obj;
         }
 
-        private IMoniker CreateMoniker(string itemName)
+        private Task<IMoniker> CreateMoniker(string itemName)
```
AArnott: nit: This method should be called `CreateMonikerAsync` since it returns a `Task`, to conform to the convention.

AArnott: How often will this method get called? This signature requires that you allocate a new `Task<T>` object for every invocation just to wrap the result that you mentioned you expect to normally be executed synchronously on the caller's MTA thread. That could contribute to GC pressure if it's called a lot.
And if it is called a lot, you have a few options:
1. double-down on your expectation that you're always on an MTA anyway and just return `IMoniker` and throw if you're on an STA.
1. return `ValueTask<T>` instead (and in the Task.Run code path wrap the Task with a ValueTask).
1. move your `if` branch logic up to the caller.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,40 @@ namespace Microsoft.Build.Execution
...
         /// <summary>
         /// Attempts to retrieve an item from the ROT.
         /// </summary>
         public object GetObject(string itemName)
         {
-            IMoniker mk = CreateMoniker(itemName);
-            int hr = this.rot.GetObject(mk, out object obj);
+            Task<IMoniker> mkTask = CreateMoniker(itemName);
+            Task.WaitAll(_rotTask, mkTask);
+            var rot = _rotTask.Result;
```
AArnott: If the `_rotTask` has faulted, then `Task.Result` throws `AggregateException` instead of the original exception (the original is preserved as an inner exception). If you expect exceptions might happen and want to preserve the original, use `_rotTask.GetAwaiter().GetResult()` instead.

src/Build/Instance/RunningObjectTable.cs
```diff
@@ -14,34 +16,40 @@ namespace Microsoft.Build.Execution
...
 
         /// <summary>
         /// Attempts to retrieve an item from the ROT.
         /// </summary>
         public object GetObject(string itemName)
         {
-            IMoniker mk = CreateMoniker(itemName);
-            int hr = this.rot.GetObject(mk, out object obj);
+            Task<IMoniker> mkTask = CreateMoniker(itemName);
+            Task.WaitAll(_rotTask, mkTask);
```
AArnott: `WaitAll` will allocate a `Task[]` array. If this method is called a lot, that could needlessly cause GC pressure. 
Instead, consider just waiting for each task individually.

src/Build/Instance/RunningObjectTable.cs
wli3: I choose 3 since the method is not too long. I just inlined it

wli3: inlined

src/Build/Instance/RunningObjectTable.cs
wli3: Wait individually

wli3:
Test VS insertion looks good. Merging

Issue #4283, opened 2019-04-03
Use custom match in FileUtilities.GetFullPath


rainersigwald:
@davkean, @BenVillalobos and I were doing some exploratory PerfView analysis and noticed this.

#3969 tried to reduce the allocation of `Regex.Match` objects that aren't necessary because all we care about is `IsMatch`. That simplified the code greatly. But unfortunately, it still allocates a `Match` object under the covers.

We should replace usages of `FileUtilitiesRegex.DrivePattern`, `FileUtilitiesRegex.UncPattern`, `StartWithDrivePattern`, and `StartsWithUncPattern` with functions that do the match explicitly without allocation. The patterns are pretty straightforward, so the matcher methods should still be pretty readable.

rainersigwald:
This will help with part of #2645.

BenVillalobos:
This was fixed in #4304

PR #4304, opened 2019-04-12
FileUtilities Custom Drive Pattern & UNC Pattern Matches

BenVillalobos:
Solves #4283 

Wrote alternate functions to take the place of using these objects:
- FileUtilitiesRegex.DrivePattern
- FileUtilitiesRegex.UncPattern
- FileUtilitiesRegex.StartWithDrivePattern
- FileUtilitiesRegex.StartsWithUncPattern 

Also wrote tests for the functions above as well as the newly created functions. Switching to the new functions reduces allocations caused by the regex `IsMatch` or `Match`.

The new functions check for the patterns explicitly through iteration without the use of objects.

### DIFF
--- a/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
+++ b/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
@@ -120,7 +120,7 @@ namespace Microsoft.Build.Engine.UnitTests
         }
 
         /// <summary>
-        /// Tests DrivePattern regex using varopis invalid drive patterns.
+        /// Tests DrivePattern regex using various invalid drive patterns.
         /// </summary>
         [Fact]
         public void DoesStartWithDrivePatternInvalidFormat()
--- a/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
+++ b/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
@@ -12,8 +12,8 @@ namespace Microsoft.Build.Engine.UnitTests
 {
     public class FileUtilitiesRegex_Tests
     {
-        private string _directoryStart = new string(Path.DirectorySeparatorChar, 2);
-        private string _altDirectoryStart = new string(Path.AltDirectorySeparatorChar, 2);
+        private string _directoryStart = new string(MSBuildConstants.BackslashChar[0], 2);
+        private string _altDirectoryStart = new string(MSBuildConstants.ForwardSlash[0], 2);
 
         [Fact]
         public void DrivePatternIsMatchAllProperFormats()
--- a/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
+++ b/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
@@ -16,6 +16,8 @@ namespace Microsoft.Build.Engine.UnitTests
     {
         private string _directoryStart = new string(MSBuildConstants.BackslashChar[0], 2);
         private string _altDirectoryStart = new string(MSBuildConstants.ForwardSlash[0], 2);
+
+        //below are the legacy regex used before explcitly checking these patterns to reduce allocations
         
         // regular expression used to match file-specs comprising exactly "<drive letter>:" (with no trailing characters)
         internal static readonly Regex DrivePattern = new Regex(@"^[A-Za-z]:$", RegexOptions.Compiled);
--- a/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
+++ b/src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
@@ -15,11 +15,8 @@ namespace Microsoft.Build.Engine.UnitTests
         private string _directoryStart = new string(Path.DirectorySeparatorChar, 2);
         private string _altDirectoryStart = new string(Path.AltDirectorySeparatorChar, 2);
 
-        /// <summary>
-        /// Tests for DrivePattern regex using valid drive patterns.
-        /// </summary>
         [Fact]
-        public void DrivePatternSuccessWithRegex()
+        public void DrivePatternIsMatchAllProperFormats()
         {
             string s;
             for(char i = 'A'; i <= 'Z'; i++)
@@ -31,11 +28,8 @@ namespace Microsoft.Build.Engine.UnitTests
             }
         }
 
-        /// <summary>
-        /// Tests IsDrivePattern using valid drive patterns.
-        /// </summary>
         [Fact]
-        public void DrivePatternSuccessWithoutRegex()
+        public void IsDrivePatternAllProperFormats()
         {
             string s;
             for (char i = 'A'; i <= 'Z'; i++)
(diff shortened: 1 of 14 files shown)


src/Shared/FileUtilitiesRegex.cs
```diff
@@ -34,5 +34,180 @@ internal static class FileUtilitiesRegex
...
+            if(!MeetsUncPatternMinimumRequirements(pattern))
+            {
+                return false;
+            }
+
+            bool prevCharWasSlash = true;
+            bool hasSubfolder = false;
+            for (int i = 2; i < pattern.Length; i++)
+            {
+                if (pattern[i] == Path.DirectorySeparatorChar ||
```
rainersigwald: Probably worth a comment that real UNC paths should only be backslashes but the previous pattern accepted both so this function does too.

src/Shared/FileUtilitiesRegex.cs
```diff
@@ -34,5 +34,180 @@ internal static class FileUtilitiesRegex
...
+        /// <returns></returns>
+        internal static int DoesStartWithUncPatternMatchLength(string pattern)
+        {
+            if (!MeetsUncPatternMinimumRequirements(pattern))
+            {
+                return -1;
+            }
+
+            bool prevCharWasSlash = true;
+            bool hasSubfolder = false;
```
rainersigwald: Is this a subfolder, or the share name in `\\server\share`?

src/Shared/FileUtilitiesRegex.cs
```diff
@@ -34,5 +34,180 @@ internal static class FileUtilitiesRegex
...
+            return pattern.Length;
+        }
+
+        /// <summary>
+        /// Indicates whether or not the file-spec meets the minimum requirements of a UNC pattern.
+        /// UNC pattern requires a minimum length of 5 and first two characters must be a slash.
+        /// </summary>
+        /// <param name="pattern"></param>
+        /// <returns></returns>
+        internal static bool MeetsUncPatternMinimumRequirements(string pattern)
```
rainersigwald: we can probably be justified sprinkling a few of these around, but definitely here:

```suggestion
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        internal static bool MeetsUncPatternMinimumRequirements(string pattern)
```

src/Shared/FileUtilitiesRegex.cs
```diff
@@ -34,5 +34,180 @@ internal static class FileUtilitiesRegex
             new Regex(
                 string.Format(CultureInfo.InvariantCulture, @"{0}$", s_baseUncPattern),
                 RegexOptions.Compiled);
```
rainersigwald: Are these regexes still used? Where?

src/Shared/Modifiers.cs
```diff
@@ -472,17 +472,13 @@ internal static string GetItemSpecModifier(string currentDirectory, string itemS
...
                                 length = 2;
                             }
                             else
                             {
-                                var match = FileUtilitiesRegex.StartsWithUncPattern.Match(modifiedItemSpec);
-                                if (match.Success)
-                                {
-                                    length = match.Length;
-                                }
+                                int matchLength = FileUtilitiesRegex.DoesStartWithUncPatternMatchLength(modifiedItemSpec);
```
rainersigwald:
```suggestion
                                length = FileUtilitiesRegex.DoesStartWithUncPatternMatchLength(modifiedItemSpec);
```

?

src/Shared/FileUtilitiesRegex.cs
```diff
@@ -34,5 +34,180 @@ internal static class FileUtilitiesRegex
...
+            //Valid unc patterns don't end with slashes & have at least 1 subfolder
+            return !prevCharWasSlash && hasSubfolder;
+        }
+
+        /// <summary>
+        /// Indicates whether the specified file-spec begins with "\\<server>\<path>".
+        /// </summary>
+        /// <param name="pattern"></param>
+        /// <returns></returns>
+        internal static bool DoesStartWithUncPattern(string pattern)
```
rainersigwald: Feels like this could be a thin wrapper over the MatchLength method below.

src/Shared/FileUtilitiesRegex.cs
BenVillalobos: This is the share. Think this is due for a rename?

src/Shared/FileUtilitiesRegex.cs
BenVillalobos: After this change they're entirely unused except for in tests.

src/Shared/FileUtilitiesRegex.cs
rainersigwald: Yeah, let's use the official names:

```abnf
UNC = "\\" host-name "\" share-name [ "\" object-name ]
host-name = "[" IPv6address "]" / IPv4address / reg-name
; IPv6address, IPv4address, and reg-name as specified in [RFC3986]
share-name = 1*80pchar
pchar = %x20-21 / %x23-29 / %x2D-2E / %x30-39 / %x40-5A / %x5E-7B / %x7D-FF
object-name = *path-name [ "\" file-name ]
path-name = 1*255pchar
file-name = 1*255fchar [ ":" stream-name [ ":" stream-type ] ]
fchar = %x20-21 / %x23-29 / %x2B-2E / %x30-39 / %x3B / %x3D / %x40-5B / %x5D-7B
/ %x7D-FF
stream-name = *schar
schar = %x01-2E / %x30-39 / %x3B-5B /%x5D-FF
stream-type = 1*schar
```

(from the [Windows Data Types](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/cca27429-5689-4a16-b2b4-9325d93e4ba2) spec)

src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
```diff
@@ -0,0 +1,608 @@
...
+                FileUtilitiesRegex.DoesStartWithDrivePattern(s).ShouldBeTrue();
+                s = (char)(i + ('a' - 'A')) + ":";
+                FileUtilitiesRegex.DoesStartWithDrivePattern(s).ShouldBeTrue();
+            }
+        }
+
+        [Fact]
+        public void DrivePatternIsMatchInvalidFormat()
+        {
+            FileUtilitiesRegex.DrivePattern.IsMatch("C: ").ShouldBe(false);
```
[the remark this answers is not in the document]
BenVillalobos: Updated, thanks for pointing this out!

src/Shared/FileUtilitiesRegex.cs
rainersigwald: At this point I'm thinking maybe just pull the regexes into the test code so the product code is clean and no one's tempted to use them in the future?

src/Build.UnitTests/FileUtilitiesRegex_Tests.cs
```diff
@@ -0,0 +1,617 @@
...
+        [InlineData("::")]
+        [InlineData("C:/")]
+        [InlineData("C:\\")]
+        [InlineData("C\\")]
+        [InlineData(":/")]
+        [InlineData("  ")]
+        [InlineData("")]
+        public void IsDrivePatternInvalidFormat(string value)
+        {
+            FileUtilitiesRegex.IsDrivePattern(value).ShouldBe(false);
```
rainersigwald: Now that these are simplified, I'd pull the check-via-legacy-regex and the check-via-new-code calls into single methods for each valid/invalid case. The important thing is that they continue to match and do the expected thing.

src/Shared/FileUtilitiesRegex.cs
BenVillalobos: Comment added

src/Shared/FileUtilitiesRegex.cs
BenVillalobos: Done.

PR #4310, opened 2019-04-15
ProjectGraph 'InnerBuildsCanHaveSeparateReferences' Failed Randomly

BenVillalobos:
The issue was a dictionary lookup on the `InnerBuild` key:

`GetNodesWithProjectNumber(graph, 1).First(n => n.ProjectInstance.GlobalProperties.TryGetValue(InnerBuildPropertyName, out string p) && p == "a")`

where by chance a node with no `InnerBuild` references is evaluated before the lambda returns true


cdmihai:
Kept wanting to look into this for weeks now, thanks for checking it out!

The fix looks good, since graph construction is concurrent, the order can be arbitrary. After this gets merged I'll replace all these inner build queries with a single method.

PR #4318, opened 2019-04-17
Switch to Arcade versioning and 16.2

rainersigwald:
Removes all uses of Nerdbank.GitVersioning in favor of Arcade's versioning.

For internal folks: [test insertion to VS](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/175737). RPS passes; DDRITs show 10 passing tests but report "your test run did not complete". I've installed the build and poked around a bit; it seems fine.

```sh-session
> msbuild -version
Microsoft (R) Build Engine version 16.2.0-dev.19219.1+74698f6cd for .NET Framework

16.200.19.21901
```

This will allow us to respin builds with new optprof data without pushing an empty commit, which we've had to do close to release several times. It should also help with a lot of the Linux build issues in #4196. Also you can build from a git worktree now.


rainersigwald:
Force-pushed over (should-be) final version change for 16.1 after RIing `vs16.1`. As long as this is green I plan to merge.

Issue #4155, opened 2019-02-07
TaskRegistry wastes up to 1% of the heap on duplicate strings


davkean:
I have a dump where VS OOM'd from opening a very large solution containing http://github.com/dotnet/corefx, and I can see that nearly 1% of the heap is TaskRegistry tracking tasks in a process where we don't do builds.

![image](https://user-images.githubusercontent.com/1103906/52388901-bd660280-2ae4-11e9-8daf-d83879adc6ea.png)

![image](https://user-images.githubusercontent.com/1103906/52388929-de2e5800-2ae4-11e9-9776-01c4e327df5a.png)

Can we dedup these strings or somehow reduce this?

PR #4324, opened 2019-04-18
Task Registry Now Uses Opportunistic Intern

BenVillalobos:
Significantly improves #4155 

Opening a solution of 810 projects yields reduced string duplication.

Example string that went from 9152 duplicates to 680:
image

image


Pilchie:
FYI @drewnoakes and @davkean

davkean:
Nice!

PR #4326, opened 2019-04-19
ToolLocationHelper Expander Fastpaths

BenVillalobos:
Part 2 of #4237 

This was broken into two PR's per @rainersigwald 's [suggestion](https://github.com/Microsoft/msbuild/pull/4237#discussion_r277015713) to separate the logic that required a new project reference.


BenVillalobos:
Closing for now since it causes a new assembly dependency, which might have unintended effects.

PR #4327, opened 2019-04-20
[Static graph] Changes to enable building roslyn

cdmihai:
Things needed to build `roslyn/Compilers.sln` with no changes to the roslyn repo.

### DIFF
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -25,6 +25,7 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
         private static readonly ImmutableDictionary<string, string> EmptyGlobalProperties = new Dictionary<string, string> {{PropertyNames.IsGraphBuild, "true"}}.ToImmutableDictionary();
 
         private static readonly string InnerBuildPropertyName = "InnerBuild";
+
         private static readonly string MultitargetingSpecification = $@"<PropertyGroup>
                                                                         <InnerBuildProperty>{InnerBuildPropertyName}</InnerBuildProperty>
                                                                         <InnerBuildPropertyValues>InnerBuildProperties</InnerBuildPropertyValues>
@@ -1198,14 +1199,14 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
             Regex.Matches(dot,"label").Count.ShouldBe(graph.ProjectNodes.Count);
         }
 
-        private static void AssertOuterBuildAsRoot(ProjectGraphNode outerBuild, Dictionary<string, string> additionalGlobalProperties = null)
+        private static void AssertOuterBuildAsRoot(ProjectGraphNode outerBuild, Dictionary<string, string> additionalGlobalProperties = null, int expectedInnerBuildCount = 2)
         {
             additionalGlobalProperties = additionalGlobalProperties ?? new Dictionary<string, string>();
 
             AssertOuterBuildEvaluation(outerBuild, additionalGlobalProperties);
 
             outerBuild.ReferencingProjects.ShouldBeEmpty();
-            outerBuild.ProjectReferences.Count.ShouldBe(2);
+            outerBuild.ProjectReferences.Count.ShouldBe(expectedInnerBuildCount);
 
             foreach (var innerBuild in outerBuild.ProjectReferences)
             {
@@ -1213,7 +1214,7 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
             }
         }
 
-        private static void AssertOuterBuildAsNonRoot(ProjectGraphNode outerBuild, Dictionary<string, string> additionalGlobalProperties = null)
+        private static void AssertOuterBuildAsNonRoot(ProjectGraphNode outerBuild, Dictionary<string, string> additionalGlobalProperties = null, int expectedInnerBuildCount = 2)
         {
             additionalGlobalProperties = additionalGlobalProperties ?? new Dictionary<string, string>();
 
@@ -1230,7 +1231,7 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
                             IsInnerBuild(p) 
                             && p.ProjectInstance.FullPath == outerBuild.ProjectInstance.FullPath).ToArray();
 
-                innerBuilds.Length.ShouldBe(2);
+                innerBuilds.Length.ShouldBe(expectedInnerBuildCount);
 
                 foreach (var innerBuild in innerBuilds)
                 {
@@ -1314,6 +1315,31 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
             var outerBuild = graph.GraphRoots.First();
 
             AssertOuterBuildAsRoot(outerBuild);
+        }
+
+        [Fact]
+        public void DuplicatedInnerBuildMonikersShouldGetDeduplicated()
+        {
+            // multitarget to duplicate monikers
+            var multitargetingSpecification = MultitargetingSpecification +
+                                              @"<PropertyGroup>
+                                                    <InnerBuildProperties>a;a</InnerBuildProperties>
+                                                </PropertyGroup>";
+
+            var root = CreateProjectFile(_env, 1, new[] {2}, null, null, multitargetingSpecification).Path;
+            CreateProjectFile(_env, 2, null, null, null, multitargetingSpecification);
+
+            var graph = new ProjectGraph(root);
+
+            var dot = graph.ToDot();
+
+            graph.ProjectNodes.Count.ShouldBe(4);
+
+            var rootOuterBuild = GetNodesWithProjectNumber(graph, 1).First(IsOuterBuild);
+            var nonRootOuterBuild = GetNodesWithProjectNumber(graph, 2).First(IsOuterBuild);
+
+            AssertOuterBuildAsRoot(rootOuterBuild, null, 1);
+            AssertOuterBuildAsNonRoot(nonRootOuterBuild, null, 1);
         }
 
         [Fact]
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -10,8 +10,11 @@ using System.Text.RegularExpressions;
 using Microsoft.Build.BackEnd;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Exceptions;
+using Microsoft.Build.Execution;
+using Microsoft.Build.Internal;
 using Microsoft.Build.Shared;
 using Microsoft.Build.UnitTests;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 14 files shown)


cdmihai:
cc @AndyGerlicher @smera

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -35,12 +39,34 @@ internal ProjectGraphNode(ProjectInstance projectInstance)
...
+            // First edge wins, in accordance with vanilla msbuild behaviour when multiple msbuild tasks call into the same logical project
+            edges[(this, reference)] = projectReferenceItem;
+        }
 
-        internal void AddReferencingProject(ProjectGraphNode projectGraphNode) => _referencingProjects.Add(projectGraphNode);
+        internal void RemoveReferences(GraphBuilder.GraphEdges edges)
+        {
+            foreach (var reference in _projectReferences)
+            {
+                ErrorUtilities.VerifyThrow(reference._referencingProjects.Contains(this), "references should point the nodes referencing them");
```
rainersigwald:
```suggestion
                ErrorUtilities.VerifyThrow(reference._referencingProjects.Contains(this), "references should point to the nodes referencing them");
```

src/Build/Graph/GraphBuilder.cs
```diff
@@ -0,0 +1,397 @@
...
+                }
+            }
+
+            return propertyDictionary;
+        }
+
+        internal class GraphEdges
+        {
+            private ConcurrentDictionary<(ProjectGraphNode, ProjectGraphNode), ProjectItemInstance> ReferenceItems =
+                new ConcurrentDictionary<(ProjectGraphNode, ProjectGraphNode), ProjectItemInstance>();
```
rainersigwald: This data type raises some concerns for me. Can you expand on why it's needed?

src/Build/Graph/GraphBuilder.cs
```diff
@@ -0,0 +1,397 @@
...
+            IEnumerable<ProjectGraphEntryPoint> entryPoints,
+            ProjectCollection projectCollection,
+            ProjectGraph.ProjectInstanceFactoryFunc projectInstanceFactory
+            )
+        {
+            var entryPointConfigurationMetadata = AddGraphBuildPropertyToEntryPoints(entryPoints);
+
+            var allParsedProjects = new ConcurrentDictionary<ConfigurationMetadata, ProjectGraphNode>();
+            var tasksInProgress = new ConcurrentDictionary<ConfigurationMetadata, object>();
+            var projectsToEvaluate = new ConcurrentQueue<ConfigurationMetadata>();
```
livarcocc: Why not just pass the entryPointConfigurationMetadata to the ConcurrentQueue constructor? It is a List and therefore an IEnumerable. If you do that you won't need the foreach below.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -573,6 +530,20 @@ private static ImmutableList<string> ExpandDefaultTargets(ImmutableList<string>
...
+
+                    var expandedTargets = string.IsNullOrEmpty(targetsString)
+                        ? defaultTargets
+                        : ExpressionShredder.SplitSemiColonSeparatedList(targetsString).ToList();
+
+                    targets = targets
+                        .RemoveAt(i)
+                        .InsertRange(i, expandedTargets);
+
+                    i += defaultTargets.Count;
```
rainersigwald: `defaultTargets` or `expandedTargets`?

src/Build/Graph/ProjectInterpretation.cs
```diff
@@ -115,7 +115,12 @@ internal static ProjectType GetProjectType(ProjectInstance project)
                     : ProjectType.NonMultitargeting;
         }
 
-        public static void PostProcess(ConcurrentDictionary<ConfigurationMetadata, ProjectGraphNode> allNodes)
+        /// <summary>
+        /// To avoid calling nuget at graph construction time, the graph is initially constructed with outer build nodes referencing inner build nodes.
+        /// However, at build time, for non root outer builds, the inner builds are NOT referenced by the outer build, but by the nodes referencing the
+        /// outer build. Change the graph to mimic this behaviour.
```
rainersigwald: I don't understand. It sounds like "it starts outer->inner, but at build time it's actually outer->inner", but that doesn't make sense. Can you rephrase?

src/Build/Graph/GraphBuilder.cs
```diff
@@ -0,0 +1,397 @@
...
+                ProjectNodes = allParsedProjects.Values.ToList();
+            }
+            else
+            {
+                throw new AggregateException(exceptions);
+            }
+
+            IReadOnlyCollection<ProjectGraphNode> GetGraphRoots(IReadOnlyCollection<ProjectGraphNode> entryPointNodes)
+            {
+                var graphRoots = new List<ProjectGraphNode>(entryPointNodes.Count);
```
livarcocc: Ok to waste memory here if not all entryPoints end up being root?

src/Build/Graph/GraphBuilder.cs
```diff
@@ -0,0 +1,397 @@
...
+                foreach (var entryPoint in entryPoints)
+                {
+                    var globalPropertyDictionary = CreatePropertyDictionary(entryPoint.GlobalProperties);
+
+                    AddGraphBuildGlobalVariable(globalPropertyDictionary);
+
+                    var configurationMetadata = new ConfigurationMetadata(FileUtilities.NormalizePath(entryPoint.ProjectFile), globalPropertyDictionary);
+                    entryPointConfigurationMetadata.Add(configurationMetadata);
+                }
+                return entryPointConfigurationMetadata;
```
livarcocc: NIT: empty line after closing bracket.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -769,6 +851,78 @@ public void GetTargetListsFiltersTargetsForOuterAndInnerBuilds()
             }
         }
 
+        [Fact]
+        public void GetTargetListsDoesNotUseTargetsMetadataOnInnerBuildsFromRootOuterBuilds()
```
rainersigwald: May I beg you for a diagram of this?

src/Build/Graph/GraphBuilder.cs
```diff
@@ -0,0 +1,397 @@
...
+            allParsedProjects[configurationMetadata] = graphNode;
+            return graphNode;
+        }
+
+        /// <summary>
+        ///     Load a graph with root node at entryProjectFile
+        ///     Maintain a queue of projects to be processed and evaluate projects in parallel
+        ///     Returns false if loading the graph is not successful
+        /// </summary>
+        private bool FindGraphNodes(
```
livarcocc: Should this be LoadGraphNodes instead? Find seems to imply a search and I would not expect modifications.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -1239,6 +1419,13 @@ private static void AssertOuterBuildAsNonRoot(ProjectGraphNode outerBuild, Dicti
 
                     innerBuild.ReferencingProjects.ShouldContain(outerBuildReferencer);
                     innerBuild.ReferencingProjects.ShouldNotContain(outerBuild);
+
+                    graph.TestOnly_Edges.TestOnly_HasEdge((outerBuild, innerBuild)).ShouldBeFalse();
```
rainersigwald: Wait, why no edge for outer->inner of the same project?

src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -10,6 +11,7 @@ namespace Microsoft.Build.Experimental.Graph
     /// <summary>
     /// Represents the node for a particular project in a project graph.
     /// </summary>
+    [DebuggerDisplay(@"{ToString()}")]
```
rainersigwald: You could do this without exposing `ToString()` as public API by calling the method something else; might be worth doing.

src/Build/Graph/GraphBuilder.cs
cdmihai: Trimmed the excess memory.

src/Build/Graph/GraphBuilder.cs
cdmihai: It finds and loads them, but does not connect them. What do you think about `FindAndLoadGraphNodes`? Seems a bit superfluous.

cdmihai:
> Can you lay out what the problems were when building Roslyn that you've resolved here? That would help place the review.
> 
> The commit `Read ProjectReferenceItem.Targets` is not landing for me clearly. I think I understand the goal but not the implementation.

Some Roslyn projects set the `Targets` metadata on certain `ProjectReference` items. Static graph did not initially take this metadata into account, and therefore it inferred the wrong entry targets for the references which had that metadata set. So then, in addition to the `.default` symbol I also added `.projectReferenceTargetsOrDefault` and updated the target protocol to use it instead of `.default`.

The actual change implies remembering the project reference metadata from graph construction, so it can be read by `ProjectGraph.GetTargetLists` when it infers entry targets.

src/Build/Graph/GraphBuilder.cs
cdmihai: The change requires keeping extra state for each edge (the project reference metadata for each reference). There's two ways to keep it:
1. Enhance the OM by adding a GraphEdge class, and then have ProjectGraphNode reference lists of GraphEdge objects, rather than directly referencing parent/child ProjectGraphNodes
2. Hack it out by introducing a dictionary.

1. implies either changing the public API to add GraphEdge, or making it private by adding a dictionary on each node, which is memory wasteful. Since it's not clear whether anyone would find the extra public API useful, I decided to just go with 2. and keep everything private. If, in the future, somebody needs to access this data, then we can spend more energy in designing some cute looking API.

src/Build/Graph/ProjectGraph.cs
cdmihai: Good catch! Fortunately the defect does not affect correctness.

src/Build/Graph/ProjectInterpretation.cs
cdmihai: Updated comment with example.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: image

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: These are outer builds that are not roots. For non-root outerbuilds, the project that references the outer build also references the inner builds.

src/Build/Graph/GraphBuilder.cs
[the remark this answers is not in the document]
cdmihai: Good question. Moved it inside the method.

cdmihai:
@livarcocc @rainersigwald responded to all feedback

PR #4330, opened 2019-04-22
WIP: ExpandPropertiesLeaveTypedAndEscaped ReuseableStringBuilder Memory Improvement

### LINKED ISSUE #4084: ExpandPropertiesLeaveTypedAndEscaped (incorrectly) makes the assumption that it's ReuseableStringBuilder won't be resized
Opening a large 1000 project solution based on a solution with http://github.com/dotnet/project-system's project duplicated many times, I can see that 8.4% of all allocations during solution load (before we OOM) is ExpandPropertiesLeaveTypedAndEscaped expanding its ReuseableStringBuilder:

image

This seems to indicate that this is not required:

``` C#
                    // This method is called very often - of the order of 3,000 times per project.
                    // With the reuseable string builder, there's no particular need to initialize the length as it will already have grown.
                    using (var result = new ReuseableStringBuilder())
```

I suspect we're not caching the builder because it's grown too big so the comment will be incorrect. We should set a sensible initial size based on what we're going to put inside it.

BenVillalobos:
Meant to fix #4084 but can't seem to repro & prove the fix works.

Before creating a reuseablestringbuilder, the size is now calculated in order to prevent reallocation.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -1112,14 +1112,35 @@ namespace Microsoft.Build.Evaluation
                         return expression;
                     }
 
+                    // Calculate the expected size of the results to receive a stringbuilder
+                    // of a proper size.
+                    int expectedSize = 0;
+
+                    if(results != null)
+                    {
+                        foreach(object component in results)
+                        {
+                            expectedSize += component.ToString().Length;
+                        }
+                    }
+
+                    if(lastResult != null)
+                    {
+                        expectedSize += lastResult.ToString().Length;
+                    }
+
+                    if(expression.Length - sourceIndex > 0)
+                    {
+                        expectedSize += (expression.Length - sourceIndex);
+                    }
+
                     // We have more than one result collected, therefore we need to concatenate
                     // into the final result string. This does mean that we will lose type information.
                     // However since the user wanted contatenation, then they clearly wanted that to happen.
 
                     // Initialize our output string to empty string.
                     // This method is called very often - of the order of 3,000 times per project.
-                    // With the reuseable string builder, there's no particular need to initialize the length as it will already have grown.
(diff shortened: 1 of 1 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -1112,14 +1112,34 @@ private static class PropertyExpander<T>
...
                     }
 
+                    // Calculate the required size for the stringbuilder
+                    int expectedSize = 0;
+
+                    if(results != null)
+                    {
+                        foreach(object component in results)
+                        {
+                            expectedSize += component.ToString().Length;
```
danmosemsft: I don't have context on this change, but you are making and discarding temporary strings in a loop. Can that be avoided? Eg., some way to get length without allocating. Or a more crude way to estimate length. Resizing a StringBuilder is not super expensive, since it does not throw out the old block, it just adds another. Since the purpose of StringBuilder is to avoid temp strings, it seems odd to create temp strings in order to prepare a StringBuilder.

danmosemsft: Ah, I remember this is not a regular StringBuilder. It is supposed to get reused. Have you tried dumping (in some repro cases) what size the returned builder is reaching on Release? Perhaps MaxBuilderSize just needs increasing. That way you get reuse, which is even better than appropriately sizing initially, and was what the code was intended to achieve.

BenVillalobos:
WIP: Attempting a fix the larger issue in [another branch](https://github.com/microsoft/msbuild/tree/exp/reuseablestringbuilder-maxbuildersize) by increasing the MaxBuilderSize of the ReuseableStringBuilder so that they're not discarded so often.

rainersigwald:
Closing as we've shifted focus a bit and don't see finishing this in the short term. We can reopen when we have the cycles for it.

PR #4333, opened 2019-04-23
Use Microsoft.VisualStudio.PackageGroup.NuGet (16.0)

### LINKED ISSUE #4307: Depend on NuGet PackageGroup instead of NuGet.BuildTools
Internal bug: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/844554

Once it exists, we will need to retarget the MSBuild VSIX package to depend on the new NuGet PackageGroup, instead of:

https://github.com/Microsoft/msbuild/blob/0da7b4d3d9066daf43658fa09a13d1c299a02487/src/Package/MSBuild.VSSetup/files.swr#L8-L10

rainersigwald:
Fixes #4307, depending on the new PackageGroup that comprises the
correct NuGet package for the current SKU.

### DIFF
--- a/src/Package/MSBuild.VSSetup/files.swr
+++ b/src/Package/MSBuild.VSSetup/files.swr
@@ -6,7 +6,7 @@ package name=Microsoft.Build
         vs.package.language=neutral
 
 vs.dependencies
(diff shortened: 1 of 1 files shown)


rainersigwald:
@zivkan, look good?

Issue #4308, opened 2019-04-15
"salt" MSBuild handshake in test / dogfood environment


rainersigwald:
_From @dsplaisted on Tuesday, 29 January 2019 01:54:52_

Right now for our tests (and in the dogfood script) we override the MSBuildSDKsPath to point MSBuild to the SDK under test.  However, this doesn't seem to prevent sharing MSBuild nodes that didn't have this set, which can end up causing assembly loading or other issues.

Ideally in our test environment we would add a "salt" to the node connection handshake, so that MSBuild would only be able to connect to nodes that come from the same test environment.

This would probably require a change in MSBuild to allow us to provide such a "salt".

@rainersigwald, thoughts?

_Copied from original issue: dotnet/sdk#2880_

rainersigwald:
_From @rainersigwald on Tuesday, 29 January 2019 15:43:32_

I'd be on board with adding the value of a `MSBUILDNODEHANDSHAKESALT` environment variable to the handshake. I don't think it's generally useful, but for internal development and especially tests that require multiple invocations and need node reuse to be enabled it'd be nice.

danmosemsft:
Would it work to just hash the path to msbuild.exe into the handshake? So parent and child must be the same binary. Perhaps I misunderstand because I thought it did this actually.

dsplaisted:
@danmosemsft For the dotnet/sdk repo, we don't create an entirely new layout of the CLI for our test environment.  We use the stage 0 MSBuild, with environment variables that redirect to the SDK under test.  Because of this, we have to disable node reuse for our tests, to avoid them connecting to a stage 0 node.

BenVillalobos:
@rainersigwald I recall you mentioning a possible issue with the implementation in #4337? 

If I'm misremembering, the issue should be closed.

rainersigwald:
Oh man, I vaguely remember something there but not the details.

Perhaps it was that we should change the default to be path + version (to prevent intermixing same-sdk different-unzip-locations)? But we can leave the salt. I'm ok closing this.

PR #4337, opened 2019-04-26
MSBuild May "Salt" Node Handshake For Test Environments

BenVillalobos:
Addresses #4308

An environment variable `MSBUILDNODEHANDSHAKESALT` may be defined in order to allow MSBuild only to connect to nodes that come from the same test environment.

### DIFF
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -652,21 +652,24 @@ namespace Microsoft.Build.Internal
         /// </summary>
         /// <param name="hostContext">TaskHostContext</param>
         /// <returns>Base Handshake</returns>
-        public static long GetBaseHandshakeForContext(TaskHostContext hostContext)
+        private static long GetBaseHandshakeForContext(TaskHostContext hostContext)
         {
-            string nodeHandShakeSalt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT");
+            string salt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT");
 
-            long salt = 0;
+            long nodeHandshakeSalt = 0;
 
-            if (!string.IsNullOrEmpty(nodeHandShakeSalt))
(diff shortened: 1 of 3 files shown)


src/Shared/CommunicationsUtilities.cs
```diff
@@ -646,14 +646,27 @@ internal static void Trace(int nodeId, string format, params object[] args)
...
+
+            if (!string.IsNullOrEmpty(nodeHandShakeSalt))
+            {
+                salt = GetHandshakeHashCode(nodeHandShakeSalt);
+            }
+
+            //hostContext takes up 4 bits (value ranges from 0-4), shift it until just before the most significant byte
+            //the most significant byte gets zero'd out to avoid connecting to older builds.
+            //salt is shifted to after the fileversionhash (32 bits + 8 bits)
+            long baseHandshake = ((long)hostContext << 52) | (salt << 40) | ((long)FileVersionHash << 8);
```
rainersigwald: So there's no overlap between these, they all get ORed with zeros? If so, can you do a nice ascii-art diagram of the bitfield?

src/Shared/CommunicationsUtilities.cs
```diff
@@ -326,17 +326,17 @@ internal static long GenerateHostHandshakeFromBase(long baseHandshake, long clie
         /// Magic number sent by the host to the client during the handshake.
         /// Derived from the binary timestamp to avoid mixing binary versions.
         /// </summary>
-        internal static long GetTaskHostHostHandshake(TaskHostContext hostContext)
+        internal static long GetHostHandshake(TaskHostContext hostContext)
```
rainersigwald: I think this is a good rename; you can derive from the fact that it's passed a `TaskHostContext` that it applies to `TaskHost` handshaking.

src/Shared/CommunicationsUtilities.cs
BenVillalobos: I changed the order on these so the salt won't potentially override the hostContext, also added the diagram.

PR #4346, opened 2019-05-01
Mark a test relevant only on full framework as WindowsOnly

radical:
`Microsoft.Build.UnitTests.Move_Tests.MoveLockedFile`

- this was marked as failing on netcore
- It expects to fail trying to move a "locked"(open for writing) file
- But on dotnet and mono, the behavior is to not fail in that operation

- Thus, marking this test as Windows/full framework only.

### DIFF
--- a/src/Tasks.UnitTests/Move_Tests.cs
+++ b/src/Tasks.UnitTests/Move_Tests.cs
@@ -461,7 +461,7 @@ namespace Microsoft.Build.UnitTests
         /// Moving a locked file will fail
         /// </summary>
         [Fact]
-        [PlatformSpecific(TestPlatforms.Windows)] // "File names under Unix are case-sensitive and this test is not useful"
+        [PlatformSpecific(TestPlatforms.Windows)] // "File locking Unix differs significantly from Windows"
         public void MoveLockedFile()
         {
             string file = null;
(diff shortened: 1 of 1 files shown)


src/Tasks.UnitTests/Move_Tests.cs
```diff
@@ -461,8 +461,7 @@ public void OutputsOnlyIncludeSuccessfulMoves()
         /// Moving a locked file will fail
         /// </summary>
         [Fact]
-        [Trait("Category", "netcore-osx-failing")]
-        [Trait("Category", "netcore-linux-failing")]
+        [PlatformSpecific(TestPlatforms.Windows)] // "File names under Unix are case-sensitive and this test is not useful"
```
rainersigwald:
```suggestion
        [PlatformSpecific(TestPlatforms.Windows)] // "File locking Unix differs significantly from Windows"
```

radical: errr.. I thought I had removed that, but turns out it was in the mono commit.

rainersigwald:
@radical can I squash this in, or is this commit present in Mono so I should merge it?

rainersigwald:
I'm going to squash this in, since it doesn't look like it's in the mono fork.

PR #4347, opened 2019-05-02
Fix test on mono, failing because of difference in error string

radical:
`Microsoft.Build.Tasks.UnitTests.Unzip_Tests.LogsErrorIfReadOnlyFileCannotBeOverwitten`

### DIFF
--- a/src/Tasks.UnitTests/Unzip_Tests.cs
+++ b/src/Tasks.UnitTests/Unzip_Tests.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Build.Tasks.UnitTests
 
                 unzip.Execute().ShouldBeFalse(() => _mockEngine.Log);
 
-                _mockEngine.Log.ShouldContain("D6DFD219DACE48F8B86EFCDF98433333.txt' is denied", () => _mockEngine.Log);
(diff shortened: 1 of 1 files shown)


rainersigwald:
@radical I think some other commits snuck into this PR. Can you minimize it? Or do we need the longer list?

radical:
@rainersigwald yikes! Fixed now.

radical:
@rainersigwald Any thoughts on this?

src/Tasks.UnitTests/Unzip_Tests.cs
```diff
@@ -134,7 +134,7 @@ public void LogsErrorIfReadOnlyFileCannotBeOverwitten()
 
                 unzip.Execute().ShouldBeFalse(() => _mockEngine.Log);
 
-                _mockEngine.Log.ShouldContain("D6DFD219DACE48F8B86EFCDF98433333.txt' is denied", () => _mockEngine.Log);
+                _mockEngine.Log.ShouldContain($"D6DFD219DACE48F8B86EFCDF98433333.txt{ (NativeMethodsShared.IsMono ? "\"" : "'") } is denied", () => _mockEngine.Log);
```
rainersigwald: Is the difference in string coming from MSBuild or some other library with a slightly-different implementation on Mono?

radical: It's from mono. And as the implementations converge, this should go away.

PR #4355, opened 2019-05-06
Protect frequent logging crashes with catch all

ViktorHofer:
Contributes to https://github.com/Microsoft/msbuild/issues/4322

(the diff is hard to read - I just wrapped everything into a try catch)

cc @rainersigwald @danmosemsft

### DIFF
--- a/src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
+++ b/src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
@@ -539,36 +539,34 @@ namespace Microsoft.Build.Logging
 
         private void Write(ITaskItem item)
         {
-            // Temporarily try catch all to mitigate frequent NullReferenceExceptions in
-            // the logging code until CopyOnWritePropertyDictionary is replaced with
-            // ImmutableDictionary. Calling into Debug.Fail to crash the process in case
-            // the exception occures in Debug builds.
-            try
-            {
-                Write(item.ItemSpec);
-                IDictionary customMetadata = item.CloneCustomMetadata();
-                Write(customMetadata.Count);
+            Write(item.ItemSpec);
+            IDictionary customMetadata = item.CloneCustomMetadata();
+            Write(customMetadata.Count);
 
-                foreach (string metadataName in customMetadata.Keys)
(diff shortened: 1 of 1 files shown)


ViktorHofer:
I did not change any of this code. You now want me to actually change this code?

rainersigwald:
Yes, I'd rather see the code changed than wrapped in this way, because I think this creates a possiblity of unloadable binlogs that didn't exist before. If you'd like to have a pure addition, you could put a `catch (Exception e)` block after the existing catch inside the loop.

ViktorHofer:
> If you'd like to have a pure addition, you could put a catch (Exception e) block after the existing catch inside the loop.

I think that's the better option. Will update the code.

danmosemsft:
Any reason you don't catch only `NullReferenceException` ? Presumably any other type would be some different issue.

danmosemsft:
(I have no opinion though..)

ViktorHofer:
As this is temporary and every kind of exception in this code path is problematic for consumers I decided to catch all.

rainersigwald:
I could see going either way but this seems fine.

How do y'all take new MSBuilds? What needs to be built/published/whatever to adopt this change? I ask because we may want to try to insert what we currently have for VS, but we have to work through some build changes first. But I don't want to block this indefinitely. Trying to figure out the best path forward.

ViktorHofer:
Usually by updating the SDK we consumer a newer version of msbuild. How does the dependency flow to the SDK currently look like?

rainersigwald:
We flow to dotnet/toolset ~constantly. Looks like that's already picked up a 16.2 build so it's pulling from master: https://github.com/dotnet/toolset/blob/0a55dbc4b7a2df3d338a069747ecfc0e026a3641/eng/Version.Details.xml#L12-L15

That should flow to core-sdk pretty directly but that currently appears to be stuck: https://github.com/dotnet/core-sdk/pull/1821

Here's my plan: wait until either we have inserted to VS or that is unblocked. If we insert to VS first, just plow ahead. If that gets unblocked first, branch for VS stuff and merge this to master. Sound ok?

danmosemsft:
@rainersigwald the PR linked above seems to have merged. Is this one unblocked now? 😀

PR #4360, opened 2019-05-09
Always copy build logs and test results

rainersigwald:
`succeededOrFailed()` evaluates to `false` when the job is cancelled (as due to a
timeout), which makes diagnosing test hangs difficult. Instead, always
capture whatever logs and test output exists.

### DIFF
--- a/.vsts-dotnet-ci.yml
+++ b/.vsts-dotnet-ci.yml
@@ -36,13 +36,13 @@ jobs:
     inputs:
       PathtoPublish: 'artifacts/log/Debug'
       ArtifactName: 'FullOnWindows build logs'
-    condition: succeededOrFailed()
+    condition: always()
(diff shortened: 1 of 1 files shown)


rainersigwald:
I think having this will help diagnose an intermittent test hang, seen most recently in #4359: https://dev.azure.com/dnceng/public/_build/results?buildId=183660&view=logs&j=a964745c-2ee7-5657-5893-57d8a23ba1c9&t=8546657c-07f1-5ad9-84b8-0e3e313ae4c2&l=649

rainersigwald:
This caused CI build failures because I accidentally inserted a bogus character into the commit message (I included `` `f`` when writing it in PowerShell), which moved into the environment as `BUILD_SOURCEVERSIONMESSAGE`, which failed tests confusingly. Things should recover after the next push.

PR #4364, opened 2019-05-09
Reduce ExtractPropertyFunction Allocations

### LINKED ISSUE #4085: ExtractPropertyFunction allocates a huge amount of data during solution load
We're attempting to quadruple the number of projects that the project system can open, while at same reducing solution load time. Reducing GC time is an important part of that.

In one trace where we are opening ~1000 projects (based on projects from http://github.com/dotnet/project-system duplicated 100s of times), I'm seeing ~13.9% of all allocations (255 MB) just parsing property functions.

Gen | Count | MaxPause | MaxPeak MB | Max AllocMB/sec | TotalPause | TotalAlloc MB | Alloc MB/MSec GC | Survived MB/MSec GC | MeanPause | Induced
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
ALL | 1659 | 458.8 | 1,637.2 | 597.633 | 26,587.2 | 7,493.9 | 0.3 | 0.200 | 16.0 | 0
0 | 616 | 45.4 | 1,629.8 | 560.025 | 7,507.8 | 2,646.6 | 1.0 | 0.099 | 12.2 | 0
1 | 1039 | 48.0 | 1,637.2 | 597.633 | 18,032.1 | 4,827.3 | 2.4 | 0.172 | 17.4 | 0
2 | 4 | 458.8 | 1,580.9 | 52.451 | 1,047.3 | 20.0 | 0.1 | 0.331 | 261.8 | 0

Almost all of this is short term strings, including parsing arguments and triming the results.

image

BenVillalobos:
Meant to fix #4085 but can't seem to repro & prove the fix works.

Reduced allocations by making use of ReadOnlySpan<char>. Previous functionality saw lots of .Substring() and .Trim() calls on strings. Now they're being called on ReadOnlySpan and aren't turned into strings until needed at the end of the function.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -4456,7 +4456,7 @@ namespace Microsoft.Build.Evaluation
                 string[] functionArguments;
 
                 // The name of the function that will be invoked
-                string functionName;
+                ReadOnlySpan<char> functionName;
 
                 // What's left of the expression once the function has been constructed
                 string remainder = String.Empty;
@@ -4474,9 +4474,8 @@ namespace Microsoft.Build.Evaluation
                     string argumentsContent;
 
                     // separate the function and the arguments
-                    functionName = spanSubstring.Trim().ToString();
+                    functionName = spanSubstring.Trim();
                     //functionName = expressionFunction.Substring(methodStartIndex, argumentStartIndex - methodStartIndex).Trim();
-
 
                     // Skip the '('
                     argumentStartIndex++;
@@ -4534,10 +4533,11 @@ namespace Microsoft.Build.Evaluation
                     if (nextMethodIndex > 0)
                     {
                         methodLength = nextMethodIndex - methodStartIndex;
-                        remainder = expressionFunction.Substring(nextMethodIndex).Trim();
+                        remainder = expFuncAsSpan.Slice(nextMethodIndex).Trim().ToString();
                     }
 
-                    string netPropertyName = expressionFunction.Substring(methodStartIndex, methodLength).Trim();
+                    ReadOnlySpan<char> netPropertyName = expFuncAsSpan.Slice(methodStartIndex, methodLength).Trim();
+                    //string netPropertyName = expressionFunction.Substring(methodStartIndex, methodLength).Trim();
 
                     ProjectErrorUtilities.VerifyThrowInvalidProject(netPropertyName.Length > 0, elementLocation, "InvalidFunctionPropertyExpression", expressionFunction, String.Empty);
 
@@ -4550,7 +4550,7 @@ namespace Microsoft.Build.Evaluation
                 // either there are no functions left or what we have is another function or an indexer
                 if (String.IsNullOrEmpty(remainder) || remainder[0] == '.' || remainder[0] == '[')
                 {
-                    functionBuilder.Name = functionName;
+                    functionBuilder.Name = functionName.ToString();
                     functionBuilder.Arguments = functionArguments;
                     functionBuilder.BindingFlags = defaultBindingFlags;
                     functionBuilder.Remainder = remainder;
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -4459,8 +4459,7 @@ namespace Microsoft.Build.Evaluation
                 ReadOnlySpan<char> functionName;
 
                 // What's left of the expression once the function has been constructed
-                string remainder = String.Empty;
-                ReadOnlySpan<char> remainder2 = ReadOnlySpan<char>.Empty;
+                ReadOnlySpan<char> remainder = ReadOnlySpan<char>.Empty;
 
                 // The binding flags that we will use for this function's execution
                 BindingFlags defaultBindingFlags = BindingFlags.IgnoreCase | BindingFlags.Public;
@@ -4512,9 +4511,8 @@ namespace Microsoft.Build.Evaluation
                             // We will keep empty entries so that we can treat them as null
                             functionArguments = ExtractFunctionArguments(elementLocation, expressionFunction, argumentsContent);
                         }
-
-                        remainder = expressionFunctionAsSpan.Slice(argumentsEndIndex + 1).Trim().ToString();
-                        remainder2 = expressionFunctionAsSpan.Slice(argumentsEndIndex + 1).Trim();
+                        
+                        remainder = expressionFunctionAsSpan.Slice(argumentsEndIndex + 1).Trim();
                     }
                 }
                 else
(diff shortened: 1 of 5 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -4456,21 +4456,25 @@ private static void ConstructFunction(IElementLocation elementLocation, string e
...
                 // The binding flags that we will use for this function's execution
                 BindingFlags defaultBindingFlags = BindingFlags.IgnoreCase | BindingFlags.Public;
 
+                ReadOnlySpan<char> expressionFunctionAsSpan = expressionFunction.AsSpan();
+                
+                ReadOnlySpan<char> expressionSubstringAsSpan = argumentStartIndex > -1 ? expressionFunctionAsSpan.Slice(methodStartIndex, argumentStartIndex - methodStartIndex) : ReadOnlySpan<char>.Empty;
+
                 // There are arguments that need to be passed to the function
-                if (argumentStartIndex > -1 && !expressionFunction.Substring(methodStartIndex, argumentStartIndex - methodStartIndex).Contains("."))
+                if (argumentStartIndex > -1 && !expressionSubstringAsSpan.Contains(".".AsSpan(), StringComparison.OrdinalIgnoreCase))
```
rainersigwald: Does this overload exist? Feels cleaner if it does

```suggestion
                if (argumentStartIndex > -1 && !expressionSubstringAsSpan.Contains('.', StringComparison.OrdinalIgnoreCase))
```

src/Build/Evaluation/Expander.cs
```diff
@@ -4507,8 +4511,8 @@ private static void ConstructFunction(IElementLocation elementLocation, string e
                             // We will keep empty entries so that we can treat them as null
                             functionArguments = ExtractFunctionArguments(elementLocation, expressionFunction, argumentsContent);
                         }
-
-                        remainder = expressionFunction.Substring(argumentsEndIndex + 1).Trim();
+
```
rainersigwald: nit: delete spaces

src/Build/Microsoft.Build.csproj
```diff
@@ -33,6 +33,7 @@
 
     <PackageReference Include="System.Collections.Immutable" />
     <PackageReference Include="System.Threading.Tasks.Dataflow" />
+    <PackageReference Include="System.Memory" />
```
rainersigwald: This will require some Setup Magic, so that we have it available in VS setup.

Notes for that: #3947, https://github.com/microsoft/msbuild/pull/3648/files#diff-ebf6dde28bf300b10540e8b71b872c5d.

Let's sync on that and pair through it (maybe next week while I'm there?).

src/Build/Evaluation/Expander.cs
```diff
@@ -4542,12 +4546,12 @@ private static void ConstructFunction(IElementLocation elementLocation, string e
...
                 // either there are no functions left or what we have is another function or an indexer
-                if (String.IsNullOrEmpty(remainder) || remainder[0] == '.' || remainder[0] == '[')
+                if (remainder.IsEmpty || remainder[0] == '.' || remainder[0] == '[')
                 {
-                    functionBuilder.Name = functionName;
+                    functionBuilder.Name = functionName.ToString();
                     functionBuilder.Arguments = functionArguments;
                     functionBuilder.BindingFlags = defaultBindingFlags;
-                    functionBuilder.Remainder = remainder;
+                    functionBuilder.Remainder = remainder.ToString();
```
rainersigwald: If this still shows as a hot spot for allocations, we'll probably have to tackle `FunctionBuilder`'s need to make these into real strings. But what you've done should help quite a bit and is a great start, so I think we should take this as-is.

src/Build/Evaluation/Expander.cs
BenVillalobos: Unfortunately there are no other overloads for Contains(), it requires another ReadOnlySpan<char>

danmosemsft: @ahsonkhan  was this considered? I do not see use of StringComparison in SpanHelpers.Char.cs.

ahsonkhan: We already have the following APIs around Contains/StringComparison:

Added in 2.1 as part of adding string-like APIs: https://github.com/dotnet/corefx/issues/21395
```C#
public static bool Contains(this ReadOnlySpan<char> span, ReadOnlySpan<char> value, StringComparison comparison);
```
Added in 3.0 as an optimization over IndexOf: https://github.com/dotnet/corefx/issues/27526 
```C#
bool Contains<T>(this Span<T> span, T value);
bool Contains<T>(this ReadOnlySpan<T> span, T value);
```

I believe the request is for an API that takes a single char instead of a `ROS<char>` like this (analogous to the string API https://apisof.net/catalog/System.String.Contains(Char,StringComparison)):
```C#
public static bool Contains(this ReadOnlySpan<char> span, char value, StringComparison comparison);
```

> I do not see use of StringComparison in SpanHelpers.Char.cs.

It's in MemoryExtensions.Fast.cs:
https://github.com/dotnet/coreclr/blob/c8e9c122aa5ea827e734fdffd889afa6318619da/src/System.Private.CoreLib/shared/System/MemoryExtensions.Fast.cs#L25-L28

> was this considered?

I don't remember exactly and it might have been an oversight, since most of the Contains APIs were added based on demonstrable need (also based on IndexOf APIs which also doesn't have an overload for a single char). I don't think we added span versions of all string-like APIs. This is probably one that we missed. The workaround is fairly straightforward though.
Either:
1) `".".AsSpan()` which gives a length-1 `ROSpan<char>`, as done here.
2) stackalloc a single char array that contains the lookup char and wrap that in a `ROSpan<char>`

@stephentoub, @GrabYourPitchforks - do you remember if this was discussed or if there is any reason we shouldn't consider adding such an overload? If we were to add it, we should probably add {Last}IndexOf overloads as well.

src/Package/MSBuild.VSSetup/files.swr
```diff
@@ -161,6 +163,8 @@ folder InstallDir:\MSBuild\Current\Bin\amd64
   file source=$(X86BinPath)Microsoft.Build.Framework.dll vs.file.ngenArchitecture=all
   file source=$(X86BinPath)Microsoft.Build.Tasks.Core.dll vs.file.ngenArchitecture=all
   file source=$(X86BinPath)Microsoft.Build.Utilities.Core.dll vs.file.ngenArchitecture=all
+  file source=$(X86BinPath)System.Memory.dll vs.file.ngenArchitecture=all
+  file source=$(X86BinPath)System.Numerics.Vectors.dll vs.file.ngenArchitecture=all
```
rainersigwald: Did RPS pass on your trial insertion to VS? This caused problems there last time. If it passed, awesome ;)

src/Build/Evaluation/Expander.cs
stephentoub: Why does it need to be `".".AsSpan()` rather than just `"."`... the implicit cast isn't sufficient?

> any reason we shouldn't consider adding such an overload?

Primarily just the sheer number of overloads we're now talking about, every one of which has additional conceptual overhead, and we're talking about the difference between `"."` and `'.'`.

BenVillalobos:
Superseded by #4394

PR #4369, opened 2019-05-11
[Static graph] Add 64bit binaries to the bootstrap directory

cdmihai:
Building cloudbuild with static graph requires the 64 bit msbuild.

This PR depends on #4327, review only the latest commit.

### DIFF
--- a/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj
+++ b/src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj
@@ -48,10 +48,6 @@
     <Copy SourceFiles="$(PublishDir)$(AssemblyName).deps.json" DestinationFiles="$(PublishDir)MSBuild.deps.json" />
   </Target>
 
-  <PropertyGroup>
-    <IgnoreMSBuildTaskHost>true</IgnoreMSBuildTaskHost>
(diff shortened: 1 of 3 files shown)


src/MSBuild.Bootstrap/MSBuild.Bootstrap.csproj
```diff
@@ -48,6 +48,12 @@
     <Copy SourceFiles="$(PublishDir)$(AssemblyName).deps.json" DestinationFiles="$(PublishDir)MSBuild.deps.json" />
   </Target>
 
+  <PropertyGroup>
+    <IgnoreMSBuildTaskHost>true</IgnoreMSBuildTaskHost>
```
rainersigwald: Why though?

cdmihai: It failed to build due to some obscure nuget tfm compat reason, and since MSBuildTaskHost wasn't required for the cloudbuild scenario, I decided to just disable it, with the reasoning that whomever wants 64bit task host can debug that and enable it :)

cdmihai: Looks good. Simpler fix than what the daunting error message suggested :)

PR #4374, opened 2019-05-14
ReaderWriterLock in PropertyDictionary

### LINKED ISSUE #4368: Deadlock in PropertyDictionary equality
I was running tests locally when I noticed a hang. I attached a debugger and see that it's two `PropertyDictionary` instances trying to `.Equals` against each other:

```sh-session
> Debug.ListThreads
 Index Id     Name                           Location
--------------------------------------------------------------------------------
*1     13400  Worker Thread                  Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance>.this[string].get
 2     15468  Worker Thread                  System.Collections.Concurrent.ConcurrentDictionary<Microsoft.Build.BackEnd.ConfigurationMetadata, object>.TryRemoveInternal
 3     13888  <No Name>                      <no stack frames>
 4     15948  <No Name>                      <no stack frames>
 5     12528  Main Thread                    System.Collections.Concurrent.ConcurrentDictionary<System.__Canon, System.__Canon>.AcquireLocks
 6     8600   Worker Thread                  Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance>.this[string].get
 7     20000  <No Name>                      Xunit.Sdk.MessageBus.ReporterWorker
 8     16780  <No Name>                      <no stack frames>
 9     11244  <No Name>                      <no stack frames>
 10    10444  <No Name>                      <no stack frames>
> Debug.ListCallStack /ShowLineOffset:yes /Thread:1
Callstack for Thread 1 (Thread Id: 13400 (0x3458)):
 Index  Function
--------------------------------------------------------------------------------
 1      Microsoft.Build.dll!Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance>.this[string].get(string name) Line 189
*2      Microsoft.Build.dll!Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance>.Equals(Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance> other) Line 286
 3      Microsoft.Build.dll!Microsoft.Build.BackEnd.ConfigurationMetadata.InternalEquals(Microsoft.Build.BackEnd.ConfigurationMetadata other) Line 165
 4      Microsoft.Build.dll!Microsoft.Build.BackEnd.ConfigurationMetadata.Equals(Microsoft.Build.BackEnd.ConfigurationMetadata other) Line 148
 5      System.Private.CoreLib.dll!System.Collections.Generic.GenericEqualityComparer<Microsoft.Build.BackEnd.ConfigurationMetadata>.Equals(Microsoft.Build.BackEnd.ConfigurationMetadata x, Microsoft.Build.BackEnd.ConfigurationMetadata y)
 6      System.Collections.Concurrent.dll!System.Collections.Concurrent.ConcurrentDictionary<Microsoft.Build.BackEnd.ConfigurationMetadata, object>.TryRemoveInternal(Microsoft.Build.BackEnd.ConfigurationMetadata key, out object value, bool matchValue, object oldValue)
 7      System.Collections.Concurrent.dll!System.Collections.Concurrent.ConcurrentDictionary<Microsoft.Build.BackEnd.ConfigurationMetadata, object>.TryRemove(Microsoft.Build.BackEnd.ConfigurationMetadata key, out object value)
 8      Microsoft.Build.dll!Microsoft.Build.Experimental.Graph.ProjectGraph.FindGraphNodes.AnonymousMethod__1(System.Threading.Tasks.Task _) Line 705
 9      System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)
 10     System.Private.CoreLib.dll!System.Threading.Tasks.Task.ExecuteWithThreadLocal(ref System.Threading.Tasks.Task currentTaskSlot)
 11     System.Private.CoreLib.dll!System.Threading.ThreadPoolWorkQueue.Dispatch()

> Debug.ListCallStack /ShowLineOffset:yes /Thread:6
Callstack for Thread 6 (Thread Id: 8600 (0x2198)):
 Index  Function
--------------------------------------------------------------------------------
 1      Microsoft.Build.dll!Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance>.this[string].get(string name) Line 189
 2      Microsoft.Build.dll!Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance>.Equals(Microsoft.Build.Collections.PropertyDictionary<Microsoft.Build.Execution.ProjectPropertyInstance> other) Line 286
 3      Microsoft.Build.dll!Microsoft.Build.BackEnd.ConfigurationMetadata.InternalEquals(Microsoft.Build.BackEnd.ConfigurationMetadata other) Line 165
 4      Microsoft.Build.dll!Microsoft.Build.BackEnd.ConfigurationMetadata.Equals(Microsoft.Build.BackEnd.ConfigurationMetadata other) Line 148
 5      System.Private.CoreLib.dll!System.Collections.Generic.GenericEqualityComparer<Microsoft.Build.BackEnd.ConfigurationMetadata>.Equals(Microsoft.Build.BackEnd.ConfigurationMetadata x, Microsoft.Build.BackEnd.ConfigurationMetadata y)
 6      System.Collections.Concurrent.dll!System.Collections.Concurrent.ConcurrentDictionary<Microsoft.Build.BackEnd.ConfigurationMetadata, Microsoft.Build.Experimental.Graph.ProjectGraphNode>.TryAddInternal(Microsoft.Build.BackEnd.ConfigurationMetadata key, int hashcode, Microsoft.Build.Experimental.Graph.ProjectGraphNode value, bool updateIfExists, bool acquireLock, out Microsoft.Build.Experimental.Graph.ProjectGraphNode resultingValue)
 7      System.Collections.Concurrent.dll!System.Collections.Concurrent.ConcurrentDictionary<Microsoft.Build.BackEnd.ConfigurationMetadata, Microsoft.Build.Experimental.Graph.ProjectGraphNode>.this[Microsoft.Build.BackEnd.ConfigurationMetadata].set(Microsoft.Build.BackEnd.ConfigurationMetadata key, Microsoft.Build.Experimental.Graph.ProjectGraphNode value)
 8      Microsoft.Build.dll!Microsoft.Build.Experimental.Graph.ProjectGraph.CreateNewNode(Microsoft.Build.BackEnd.ConfigurationMetadata configurationMetadata, Microsoft.Build.Evaluation.ProjectCollection projectCollection, Microsoft.Build.Experimental.Graph.ProjectGraph.ProjectInstanceFactoryFunc projectInstanceFactory, System.Collections.Concurrent.ConcurrentDictionary<Microsoft.Build.BackEnd.ConfigurationMetadata, Microsoft.Build.Experimental.Graph.ProjectGraphNode> allParsedProjects) Line 647
 9      Microsoft.Build.dll!Microsoft.Build.Experimental.Graph.ProjectGraph.FindGraphNodes.AnonymousMethod__0() Line 676
 10     System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)
 11     System.Private.CoreLib.dll!System.Threading.Tasks.Task.ExecuteWithThreadLocal(ref System.Threading.Tasks.Task currentTaskSlot)
 12     System.Private.CoreLib.dll!System.Threading.ThreadPoolWorkQueue.Dispatch()
```

That's a deadlock because `.Equals` locks `this._properties` and then uses an indexer into `other[key]`, but the indexer locks `this._properties`:

https://github.com/microsoft/msbuild/blob/431919a7a99dda21a7a4c0a7fab2726a289f6c4b/src/Build/Collections/PropertyDictionary.cs#L282-L292

https://github.com/microsoft/msbuild/blob/431919a7a99dda21a7a4c0a7fab2726a289f6c4b/src/Build/Collections/PropertyDictionary.cs#L189-L192

Here we're doing a ton of these comparisons to get hashtable lookups into the `tasksInProgress` dictionary (of `ConfigurationMetadata` which has a `GlobalProperties` dictionary) during static graph creation, but I think this could happen in other times/places.

rainersigwald:
Fix #4368 by using a ReaderWriterLockSlim instead of a full lock. This
will allow the case where a.Equals(b) and b.Equals(a) run concurrently,
since they both only want a reader lock on their _properties: to
enumerate their own properties, and on the other to index in.

@tmeschter suggested this approach offline, and it seems obviously correct (in hindsight!).


tmeschter:
On a related note, is there any possibility that one of the `PropertyDictionary`s could be modified during an `Equals` call?

src/Build/Collections/PropertyDictionary.cs
```diff
@@ -7,6 +7,7 @@
 using System.Diagnostics;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Evaluation;
+using System.Threading;
```
tmeschter: Nit: ordering.

src/Build/Collections/PropertyDictionary.cs
```diff
@@ -113,10 +119,16 @@ internal PropertyDictionary(int capacity, IEnumerable<T> elements)
         {
             get
             {
-                lock (_properties)
+                _lock.EnterReadLock();
```
danmosemsft: Maybe worth creating a struct that implements IDisposable so you can use a using statement instead of adding a try/finally with new indent.

src/Build/Collections/PropertyDictionary.cs
```diff
@@ -497,17 +586,31 @@ internal void RemoveProperties(IEnumerable<T> other)
...
 
             return dictionary;
         }
+
+        /// <summary>
+        /// Finalizer to clean up the <see cref="ReaderWriterLockSlim" />.
+        /// </summary>
+        ~PropertyDictionary()
+        {
+            _lock?.Dispose();
```
danmosemsft: Is this needed? If the finalizer thread is used to run dispose on the lock, why not just cut out the middleman and let the finalizer thread finalize the lock.

src/Build/Collections/CopyOnReadEnumerable.cs
```diff
@@ -28,6 +29,8 @@ internal class CopyOnReadEnumerable<T> : IEnumerable<T>
         /// </summary>
         private readonly object _syncRoot;
```
danmosemsft: You can eliminate this field now and use the RW lock? It doesn't matter what the object is so long as it's readonly and private. That will claw back the size increase in this object form adding the RW lock field.

danmosemsft: Disregard, I didn't view the full diff. You still use that syncroot in another path.

src/Build/Collections/PropertyDictionary.cs
```diff
@@ -42,6 +44,11 @@ internal sealed class PropertyDictionary<T> : IEnumerable<T>, IEquatable<Propert
         [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
         private readonly RetrievableEntryHashSet<T> _properties;
 
+        /// <summary>
+        /// Lock object to guard access to backing collection.
+        /// </summary>
+        private readonly ReaderWriterLockSlim _lock = new ReaderWriterLockSlim();
```
cdmihai: Would be nice to reuse the `DisposableReaderWriterLockSlim`, and maybe make it its own file.

cdmihai: In [https://github.com/cdmihai/msbuild/blob/c8660645471a185287148f9e0897a3ec9c1ec3c8/src/Build/Definition/ProjectCollection.cs#L87](https://github.com/cdmihai/msbuild/blob/c8660645471a185287148f9e0897a3ec9c1ec3c8/src/Build/Definition/ProjectCollection.cs#L87)

---
In reply to: [302275736](https://github.com/microsoft/msbuild/pull/4374#discussion_r302275736) [](ancestors = 302275736)

src/Build/Collections/PropertyDictionary.cs
cdmihai: Does the slim version even require disposing?

---
In reply to: [285822040](https://github.com/microsoft/msbuild/pull/4374#discussion_r285822040) [](ancestors = 285822040)

rainersigwald: It's not clear to me whether it does or not. I was aping the [`ReaderWriterLockSlim` doc example](https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim?view=netframework-4.8#examples). But I see several current examples in CoreFX that don't use dispose it, like [`CodePagesEncodingProvider`](https://github.com/dotnet/corefx/blob/5e44821bd670530cd2bf1dd330079f6e2d8e5284/src/System.Text.Encoding.CodePages/src/System/Text/CodePagesEncodingProvider.cs#L14) and [`Privilege`](https://github.com/dotnet/corefx/blob/5e44821bd670530cd2bf1dd330079f6e2d8e5284/src/System.Security.AccessControl/src/System/Security/AccessControl/Privilege.cs#L41). So I guess it's ok to skip it.

danmosemsft: @stephentoub any reason we're inconsistent?

stephentoub: > any reason we're inconsistent?

In Privilege, it's static.

And CodePagesEncodingProvider exposes a static singleton that contains the instance, so it's essentially static there as well.

stephentoub: > Does the slim version even require disposing?

ReaderWriterLockSlim wraps several event wait handles.  If the RWLS isn't disposed, they'll be left for finalization.

src/Build/Collections/PropertyDictionary.cs
```diff
@@ -423,13 +497,19 @@ public T GetProperty(string name, int startIndex, int endIndex)
         /// </summary>
         IEnumerator<KeyValuePair<string, T>> IEnumerable<KeyValuePair<string, T>>.GetEnumerator()
         {
-            lock (_properties)
+            _lock.EnterReadLock();
```
stephentoub: This method was and continues to be highly suspect.  Maybe it's fine because you control all the callers as well, but the lock is going to be held while this method yields.  In other words, when I do:
```C#
PropertyDictionary pd = ...;
var e = pd.GetEnumerator();
e.MoveNext();
... // lock is held here!
```

PR #4378, opened 2019-05-15
Collect optprof profiles from d16.2stg

rainersigwald:
This should fix the profile collection that was failing in https://dev.azure.com/devdiv/DevDiv/_apps/hub/ms.vss-releaseManagement-web.cd-release-progress?releaseId=326116&environmentId=1498239&extensionId=ms.vss-test-web.test-result-in-release-environment-editor-tab&_a=release-environment-extension


rainersigwald:
I already did this in master #4365.

PR #4397, opened 2019-05-24
VS 377113: Add support for RFC3161 timestamping

sujitnayak:
[AB#377113](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/377113):
Currently in ClickOnce project settings, if you enter a RFC 3161 timestamp server URL, timestamping of the manifest and binaries fails. This is because the current code only works with authenthicode timestamp service.
This change will add support for RFC 3161 timestamping for ClickOnce manifests. It will first attempt an RFC 3161 timestamp and it that fails, fall back to using the older authenticode timestamping.

Reviewer: JohnHart;NingLi

### DIFF
--- a/src/Tasks/ManifestUtil/mansign2.cs
+++ b/src/Tasks/ManifestUtil/mansign2.cs
@@ -26,6 +26,7 @@ namespace System.Deployment.Internal.CodeSigning
         //
         // PInvoke dll's.
         //
+        internal const String CRYPT32 = "crypt32.dll";
         internal const String KERNEL32 = "kernel32.dll";
 #if (true)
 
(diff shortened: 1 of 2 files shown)


src/Tasks/ManifestUtil/mansign2.cs
```diff
@@ -707,40 +767,129 @@ private static void AuthenticodeSignLicenseDom(XmlDocument licenseDom, CmiManife
                                                              licenseDom.OuterXml + "</msrel:RelData>";
         }
 
-        private static void TimestampSignedLicenseDom(XmlDocument licenseDom, string timeStampUrl)
+        //
+        // ObtainRFC3161Timestamp
+        //
+        // This function is from mage.exe in .NET FX and is used to implement RFC 3161 timestamping.
```
rainersigwald: Can you link to the source please, even if it'll be a dead link for external contributors? That way internal folks at least can cross-reference.

sujitnayak: Mage.exe sources:
http://vstfdevdiv:8080/DevDiv2/DevDiv/_versionControl?path=%24%2FDevDiv%2FFX%2FFeature%2FNetFXDev1%2FPublic%2Finternal%2FNDP%2Finc%2Fmansign2.cs&line=1765&lineStyle=plain&lineEnd=1765&lineStartColumn=31&lineEndColumn=53

PR #4407, opened 2019-05-29
[Static Graph] Skipped results do not trigger cache miss enforcement

### LINKED ISSUE #4343: Graph cache miss constraint triggers when it shouldn't on skipped targets
### Steps to reproduce

1.csproj, the root:
```xml
<Project DefaultTargets="Build">

  <ItemGroup>
    <ProjectReference Include="2.csproj" />

    <ProjectReferenceTargets Include="Build" Targets=".default;A" />
  </ItemGroup>

  <Target Name="Build">
    <MSBuild Projects="@(ProjectReference)" Targets="A" />
    <MSBuild Projects="@(ProjectReference)" />
  </Target>

</Project>
```

2.csproj, the reference:
```xml
<Project DefaultTargets="DefaultTarget" InitialTargets="InitialTarget">

  <Target Name="A" Condition="true == false"/>

  <Target Name="DefaultTarget" Condition="true == false"/>

  <Target Name="InitialTarget" Condition="true == false"/>

</Project>
```

Command line
```
msbuild 1.csproj /graph /isolate
```
### Expected  behavior

Build succeeds

### Actual behavior

```
  E:\delete\CentralPackageVersioning\SimplifiedNoSdks\2.csproj : error : MSB4252: Project "E:\delete\CentralPackageVersioning\SimplifiedNoSdks\1.csproj" has a reference to "E:\delete\CentralPackageVersioning\SimplifiedNoSdks\2.csproj" (A target(s)) but the build resul
t for the reference is not in the engine cache. In isolated builds this could mean one of the following:
E:\delete\CentralPackageVersioning\SimplifiedNoSdks\2.csproj : error :     - the reference was not explicitly specified as a ProjectReference item in "E:\delete\CentralPackageVersioning\SimplifiedNoSdks\1.csproj"
E:\delete\CentralPackageVersioning\SimplifiedNoSdks\2.csproj : error :     - the reference was not included in the provided input caches
E:\delete\CentralPackageVersioning\SimplifiedNoSdks\2.csproj : error :     - the reference was called with a target which is not specified in the ProjectReferenceTargets item in "E:\delete\CentralPackageVersioning\SimplifiedNoSdks\1.csproj"
```

### Cause and fix

The results cache considers it a cache miss when either the requested targets, the initial targets, or the default targets (if no requested targets) are skipped: https://github.com/Microsoft/msbuild/blob/f0efb1cd0f2c91290e5e58862fc9ed56532da00b/src/Build/BackEnd/Components/Caching/ResultsCache.cs#L154-L158

The graph should be okay if any of these get skipped: a parent project requested some targets in a reference, the targets were considered, but conditioned to false.

The `BuildResult.SatisfyRequest` method should be changed to provide more detailed outcomes so the graph can still consider it a cache hit on skipped targets. Or maybe `BuildParameters.IsolateProjects` should force `BuildResult.SatisfyRequest` to always be called with `skippedResultsAreOK = true`

cdmihai:
Closes #4343

PR is based on #4379, review that one first.

If a project has skipped targets, subsequent requests for those targets trigger the re-execution of the targets. This erroneously causes the static graph constraint of no cache misses. To fix it, `BuildParameters.IsolateProjects` will cause the `ResultsCache` to not issue a cache miss on skipped targets. This seems quite reasonable because in an isolated graph build, a project is visited only once.

### DIFF
--- a/src/Build.UnitTests/BackEnd/ResultsCache_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/ResultsCache_Tests.cs
@@ -182,7 +182,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
             result.AddResultsForTarget("testTarget2", BuildResultUtilities.GetEmptySucceedingTargetResult());
             cache.AddResult(result);
 
-            ResultsCacheResponse response = cache.SatisfyRequest(request, new List<string>(), new List<string>(new string[] { "testTarget2" }), new List<string>(new string[] { "testTarget" }), skippedResultsAreOK: false);
+            ResultsCacheResponse response = cache.SatisfyRequest(request, new List<string>(), new List<string>(new string[] { "testTarget2" }), new List<string>(new string[] { "testTarget" }), skippedResultsDoNotCauseCacheMiss: false);
 
             Assert.Equal(ResultsCacheResponseType.Satisfied, response.Type);
 
--- a/src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
+++ b/src/Build.UnitTests/Graph/IsolateProjects_Tests.cs
@@ -15,7 +15,7 @@ using Xunit.Abstractions;
 
 namespace Microsoft.Build.Experimental.Graph.UnitTests
 {
-    public class IsolateProjectsTests
+    public class IsolateProjectsTests : IDisposable
     {
         private readonly string _project = @"
                 <Project DefaultTargets='BuildSelf'>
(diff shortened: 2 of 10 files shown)


src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -710,6 +714,47 @@ public List<string> GetAfterTargetsForDefaultTargets(BuildRequest request)
...
+
+                    if (items.Count == 0 || items.All(i => string.IsNullOrWhiteSpace(i.EvaluatedInclude)))
+                    {
+                        return _ => false;
+                    }
+
+                    var fragments = items.SelectMany(i => ExpressionShredder.SplitSemiColonSeparatedList(i.EvaluatedInclude));
+                    var glob = new CompositeGlob(
+                        fragments
+                            .Select(s => MSBuildGlob.Parse(Project.Directory, s)));
```
danmosemsft: Is this perf sensitive code? Historically we kept LINQ out of the codebase, or at least out of perf sensitive places. It tends to have hidden costs.

src/Build/Resources/xlf/Strings.cs.xlf
```diff
@@ -106,6 +106,13 @@
         <note>
       {StrBegin="MSB4250: "}
       LOCALIZATION:  Do not localize the following words: ProjectGraph, ProjectReference, ToolsVersion.
+    </note>
+      </trans-unit>
+      <trans-unit id="SkippedConstraintsOnRequest">
+        <source>MSB4260: Project {0} skipped graph isolation constraints on referenced project {1}</source>
```
danmosemsft: I think the other messages put " around paths.

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
cdmihai: Moved discussion to https://github.com/microsoft/msbuild/pull/4379/files#r289095009. This change belongs to #4379.

src/Build/Resources/xlf/Strings.cs.xlf
cdmihai: Moved to https://github.com/microsoft/msbuild/pull/4379/files#r289096539

PR #4414, opened 2019-06-01
Support longer relative paths in ResolveAssemblyReference HintPaths

mfilippov:
This problem introduced in https://github.com/microsoft/msbuild/pull/3700. If hint path in reference plus current directory more than MAX_PATH then MSBuild couldn't resolve the reference. But file exists and the actual path less than MAX_PATH. The problem in function GetFileAttributesEx. This function just contacted hint path with the current directory, and it doesn't work when your current directory deep into file system but the library on the top of the hierarchy. I benchmarked this fix with BenchmarkDotNet and not found a statistically significant difference between the original code and my fix.

### DIFF
--- a/src/Tasks/AssemblyDependency/HintPathResolver.cs
+++ b/src/Tasks/AssemblyDependency/HintPathResolver.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Reflection;
 using Microsoft.Build.Shared;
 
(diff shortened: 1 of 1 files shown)


mfilippov:
Benchmark results:
``` ini

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.18362
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 4 logical and 4 physical cores
.NET Core SDK=2.1.800-preview-009677
  [Host]     : .NET Core 2.1.11 (CoreCLR 4.6.27617.04, CoreFX 4.6.27617.02), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.11 (CoreCLR 4.6.27617.04, CoreFX 4.6.27617.02), 64bit RyuJIT


```
|              Method |     Mean |     Error |    StdDev | Ratio | MannWhitney(10%) | RatioSD | Rank |
|-------------------- |---------:|----------:|----------:|------:|----------------- |--------:|-----:|
| HintPathResolverOld | 25.97 us | 0.4077 us | 0.3814 us |  1.00 |             Base |    0.00 |    1 |
| HintPathResolverNew | 26.48 us | 0.2821 us | 0.2639 us |  1.02 |             Same |    0.02 |    2 |

As I understand it patch should reduce IO when you have different HintPath on the same file from different level project hierarchy.

mfilippov:
@rainersigwald Thanks for the update. I didn't know about this API.

PR #4417, opened 2019-06-03
Keep thread exceptions in ExceptionDispatchInfo

### LINKED ISSUE #4413: Avoid losing callstack for some exceptions
https://github.com/microsoft/msbuild/blob/c70d6ded175c0a7da632018b856daff9f9cd392e/src/Build/BackEnd/BuildManager/BuildManager.cs#L411-L417

https://github.com/microsoft/msbuild/blob/c70d6ded175c0a7da632018b856daff9f9cd392e/src/Build/BackEnd/BuildManager/BuildManager.cs#L776-L780

I think these could both be improved if we threw a new exception that had a stringification (including stack) of the original exception.

rainersigwald:
Fixes #4413


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -775,8 +775,7 @@ public void EndBuild()
 
                     if (_threadException != null)
                     {
-                        // Unfortunately this will reset the callstack
-                        throw _threadException;
```
ericstj: nit: did you search the codebase for `throw` followed by something other than `new`?

rainersigwald: I hadn't, but this was a great idea. I found a couple more suspicious locations but nothing that obviously fit into this pattern. I'm going to call it Good Enough for Now.

(I did search for `Unfortunately this will reset the callstack`; this was all of those.)

PR #4418, opened 2019-06-04
Fix related item output for WinMD native implementation file.

### LINKED ISSUE #4399: C++ UWP implementation output not being copied locally from project references.
### Steps to reproduce

With Visual Studio 2019 (16.1) and the latest .NET Core 3.0 preview:

1. Create a `Windows Runtime Component (Universal Windows - C++/CX)` (C++)  project.
2. Add a `WPF App (.NET Core)` (C#) project to the solution.
3. Add a reference from the WPF project to the Windows Runtime Component project.
4. Build the solution.
5. Inspect the output directory of the WPF application at `.\bin\Debug\netcoreapp3.0\`.

### Expected  behavior

The Windows Runtime Component native implementation library (i.e. the `.dll`) should be present.

### Actual behavior

Only the `.winmd`, `.pdb`, and `.pri` files were copied locally.

### Environment data

`msbuild /version` output:

```
Microsoft (R) Build Engine version 16.1.76+g14b0a930a7 for .NET Framework

16.1.76.45076
```

OS info:

```
OS: Windows 10.0.17763.475 x64
Visual Studio: 6.1 GA
dotnet: 3.0.0-preview5-27626-15
```

peterhuene:
The ResolveAssemblyReference task outputs `RelatedFiles` that should contain
the native implementation file for "native" WinMD references. This is needed to
copy the implementation file locally for P2P references.

However, if the reference to the native WinMD file already has the
`Implementation` metadata set, the task does not output the implementation file
in the `RelatedFiles` output. In the case of C++ projects that output a WinMD
file, the `Implementation` metadata gets set and referencing projects fail to
copy local the implementation file.

This commit changes the behavior so that the task will output the
implementation file even if the `Implementation` metadata is set on the WinMD
reference.

Fixes #4399.

### DIFF
--- a/src/Tasks.UnitTests/AssemblyDependency/WinMDTests.cs
+++ b/src/Tasks.UnitTests/AssemblyDependency/WinMDTests.cs
@@ -122,12 +122,19 @@ namespace Microsoft.Build.UnitTests.ResolveAssemblyReference_Tests
         /// When a project to project reference is passed in we want to verify that
         /// the winmd references get the correct metadata applied to them
         /// </summary>
-        [Fact]
-        public void VerifyP2PHaveCorrectMetadataWinMD()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void VerifyP2PHaveCorrectMetadataWinMD(bool setImplementationMetadata)
         {
             // Create the engine.
             MockEngine engine = new MockEngine(_output);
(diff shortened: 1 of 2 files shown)


peterhuene:
@rainersigwald that's correct.  It is an existing bug that impacts both .NET Core and Framework.

PR #4420, opened 2019-06-04
.NET Core resx support

rainersigwald:
Initial implementation of support for non-string resources from `.resx` in .NET Core.

Uses System.Resources.Extensions (https://github.com/dotnet/corefx/pull/36906) to embed resources in whatever serialization format they were in in the `.resx`, deserializing from that format at runtime.

Remaining work:

- [x] Rollout strategy ~(use this for 16.2, or wait for 16.3/netcore3 feature work? Want this for netcore 3 ASAP)~ initial deploy in 16.3p1/SDK 3.0.100-p7
- [ ] More test cases
- [x] Targets authoring to opt into this mechanism when targeting `netcoreapp3.0` regardless of MSBuild runtime type: https://github.com/dotnet/sdk/pull/3342
- [x] Check to ensure that the reader type is available when _not_ targeting `netcoreapp3.0`
- [x] Correct incremental builds in the presence of ResXFileRef (#1247)
- [x] Update to use new functionality from https://github.com/dotnet/corefx/pull/38012
- [x] Test against https://github.com/rainersigwald/resx-experiments/pull/1 (needs the above)

cc @ericstj @nguerrera @merriemcgaw @dreddy-work


eng/Versions.props
```diff
@@ -6,6 +6,7 @@
     <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
     <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
         $(RestoreSources);
+        https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
         https://dotnet.myget.org/F/roslyn/api/v3/index.json;
```
ericstj: nit: surprised to see MyGet is still here.

rainersigwald: I _think_ we just haven't trimmed this back to what's necessary today. microsoft/msbuild#4426 to follow up on that.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        private static string GetFullTypeNameFromAlias(string aliasedTypeName, Dictionary<string, string> aliases)
+        {
+            if (aliasedTypeName == null)
+            {
+                return StringTypeName;
+            }
+
+            int indexStart = aliasedTypeName.IndexOf(',');
+            if (aliases.TryGetValue(aliasedTypeName.Substring(indexStart + 2), out string fullAssemblyIdentity))
```
ericstj: Probably need to make sure that indexStart >= 0 &&  indexStart + 2 < aliasedTypeName.Length.  Also I'm guessing someone might not have a space in between comma and alias.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,299 @@
...
+            {
+                return aliasedTypeName.Substring(0, indexStart + 2) + fullAssemblyIdentity;
+            }
+
+            // Allow "System.String" bare or with no registered alias
+            // TODO: is this overzealous? _Hopefully_ no one is putting a System.String type in their
```
ericstj: Did you have a case where it had an alias and the alias didn't exist?  I'm OK with an equals check here.  If the entry says `System.String` and there is no alias it is reasonable for that to mean the real `System.String`, but if someone says `System.StringComparison` or `System.String, myUtf8String` it's probably not ok to treat it as string.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,299 @@
...
+            if (mimetype == null)
+            {
+                if (typename.IndexOf("System.Byte[]") != -1 && typename.IndexOf("mscorlib") != -1)
+                {
+                    byte[] byteArray = Convert.FromBase64String(value);
+
+                    // Comment and logic from https://github.com/dotnet/winforms/blob/16b192389b377c647ab3d280130781ab1a9d3385/src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs#L411-L416
+                    // Handle byte[]'s, which are stored as base-64 encoded strings.
+                    // We can't hard-code byte[] type name due to version number
+                    // updates & potential whitespace issues with ResX files.
```
ericstj: 🤦‍♂  This logic is pretty busted, but worst-case we parse as byte[] and flow through to app as byte[].  If they were expecting a different type (like `Sytem.Typle<System.String,System.Byte[]>, mscorlib`) they'll notice.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,299 @@
...
+                    ? Encoding.GetEncoding(fileRefEncoding)
+                    : Encoding.Default;
+                using (StreamReader sr = new StreamReader(fileName, textFileEncoding))
+                {
+                    resources.Add(new StringResource(name, sr.ReadToEnd(), resxFilename));
+
+                    return;
+                }
+            }
+
```
ericstj: I think you need a special case for `.ico` here, per https://github.com/dotnet/winforms/blob/33b9fe202f3dc1b8e7c4bf28492f8bd3252f1a20/src/System.Windows.Forms/src/System/Resources/ResXFileRef.cs#L309-L315

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,299 @@
...
+            return resources;
+        }
+
+        private static void ParseAssemblyAlias(Dictionary<string,string> aliases, XElement elem)
+        {
+            string alias = elem.Attribute("alias").Value;
+            string name = elem.Attribute("name").Value;
+
+            aliases.Add(alias, name);
```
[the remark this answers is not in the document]
rainersigwald: This led me down a dark rabbit hole. I agree with you! But that's not at all what ResXResourceReader did: it does last-one-wins, and while I was looking at that I discovered that you don't actually need to specify the alias, because it'll infer it for you from the name 😐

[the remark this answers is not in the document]
ericstj: Eww.  Ok to get last in wins with a Dictionary use the indexer instead of add.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        private static void ParseAssemblyAlias(Dictionary<string,string> aliases, XElement elem)
+        {
+            string alias = elem.Attribute("alias")?.Value;
+            string name = elem.Attribute("name").Value;
+
+            if (string.IsNullOrEmpty(alias))
+            {
+                AssemblyName assemblyName = new AssemblyName(name);
```
rainersigwald: @ericstj is this safe, or might it result in loading an assembly? It's what ResXResourceReader did:

https://github.com/dotnet/winforms/blob/33b9fe202f3dc1b8e7c4bf28492f8bd3252f1a20/src/System.Windows.Forms/src/System/Resources/ResXResourceReader.cs#L575-L587

src/Tasks/ResourceHandling/MSBuildResXReader.cs
rainersigwald: I thought I'd encountered one in MSBuild's tests, but evidently not. Switching to exact-match.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
rainersigwald: Thinking about this was on my to-do list, but didn't make it into the checklist above.

If I'm reading that correctly, wouldn't it apply only at _deserialization_ time? The thing that got serialized into `.resources` before was a `System.Drawing.Bitmap`; it just got created via a two-hop process instead of just calling `Activator.CreateInstance` with the stream. I don't see an analogue for that in the new world. Am I missing something?

src/Tasks/ResourceHandling/MSBuildResXReader.cs
ericstj: Yeah. AssemblyName is safe it’s just a parsed string.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
ericstj: You change the type name you pass in.  Instead of passing in System.Drawing.Bitmap, you need to pass in System.Drawing.Icon.  We can't make the decision at deserialization time because we don't know what the file extension was.

rainersigwald: I keep struggling with this, both

## How to create this situation to test

Is there a VS UI way to get this "Bitmap read from an `.ico` file" situation? I wasn't able to find one--references to icons get returned as `System.Drawing.Icon` and/or embedded, not file-refed.

I can make it manually. If I do, do we know how to validate the reported GDI+ mangling?

## What the correct behavior is

The way I'm interpreting the current behavior, it applies _only_ when you're in this kind of situation

```xml
  <data name="pictureBox1.Image" type="System.Resources.ResXFileRef, System.Windows.Forms">
    <value>test.ico;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
  </data>
```

and what you get out at runtime is a `Bitmap`, it just has different bytes than the `Bitmap` that would have been loaded using the normal codepath, because it bounced through `Icon`. Right? I'm not seeing how we could produce that other than a special case in `System.Resources.Extensions`, something like a bool `BounceBitmapThroughIcon` at deserialize time.

ericstj: > Is there a VS UI way to get this "Bitmap read from an .ico file" situation? 

What if you have a Bitmap localizable property on a control and you point it to an Icon file?  I see lots of [hits](https://github.com/search?q=%22.ico%3BSystem.Drawing.Bitmap%22&type=Code) on this pattern so there must be something that produces it.  Winforms also has a test case: https://github.com/dotnet/winforms/blob/f753f5bad925233c65a522921f9eeaf2b46b9c78/src/System.Windows.Forms/tests/UnitTests/System/Resources/ResXFileRef.ConverterTests.cs#L111-L122.

> do we know how to validate the reported GDI+ mangling?

Double check that alpha channel remains transparent.  So use an icon with some transparency then put it on top of some colored background and make sure it stays transparent rather than defaulting to some color.

> like a bool BounceBitmapThroughIcon at deserialize time.

Hmmm, that's unfortunate.  Creating API for special cases like this is not a great solution.  It's also less than ideal to store all the images in the ico when only one will ever be exposed by the app.  At the moment I can't think of a great alternative.  We could emit a suppressible warning when we see this pattern that recommends that folks export the icon to an image...  Alternatively the task could do this special casing.

rainersigwald: > What if you have a Bitmap localizable property on a control and you point it to an Icon file?

Ah, I didn't try with a localizable property.

> Alternatively the task could do this special casing.

 This feels like the best bad option at this point.

> use an icon with some transparency then put it on top of some colored background and make sure it stays transparent rather than defaulting to some color.

Trying this: https://github.com/rainersigwald/resx-experiments/tree/transparent-fileref-icon/transparent-fileref-icon, with full framework serialization:
image

with no special cases on top of this PR:
image

with special casing to change the type to Icon from Bitmap:
image

with special casing to read an icon from the file and serialize `ToBitmap` in the legacy live-object way:

ericstj: Those diffs appear to me like Icon.ToBitmap behaves differently on Core.  Looks like it is selecting the wrong bitmap, or is generating one with wrong bit depth. 

Interestingly it seems like alpha channel is missing from .net framework case: I wonder if you have to configure that control to honor alpha.

rainersigwald: Nope, I lied: the special-cased final one isn't doing the right thing. Fixing.

rainersigwald: Sorry, that was confusing: the background of the window is lime; anything with lime showing through is handling alpha correctly. The pixels of the icon itself are white+transparent, not lime+transparent.

rainersigwald: We talked about this offline. There are a few options, but anything that involves getting a live System.Drawing.Bitmap during the build is awkward, because not every .NET Core installation on Linux or macOS will have libgdiplus, which is needed for bitmaps.

For now, we're going to elide this special case. The reported transparency mangling may have been fixed (see images above: lime green is the background color so at least some alpha is being preserved). If this causes problems, the decision can be revisited.

rainersigwald:
@kevinransom I have a note to myself from long ago that you might be interested in this PR, but unfortunately I don't recall why. Have you done something with resource handling on Core that might be affected here? Any feedback welcome :)

src/Tasks/GenerateResource.cs
```diff
@@ -3284,14 +3343,17 @@ private void WriteResources(ReaderInfo reader, String filename)
...
 
                 case Format.Binary:
-#if FEATURE_RESX_RESOURCE_READER
-                    WriteResources(reader, new ResourceWriter(File.OpenWrite(filename))); // closes writer for us
-#else
-                    WriteResources(reader, new PreserializedResourceWriter(File.OpenWrite(filename))); // closes writer for us
-#endif
+                    if (_usePreserializedResources)
+                    {
+                        // TODO: check TF/references and error if System.Resources.Extensions not available
```
ericstj: Good call. 👍

src/Tasks/GenerateResource.cs
```diff
@@ -2992,40 +3028,57 @@ private void ReadResources(String filename, bool shouldUseSourcePath, String out
...
+                                        AddResource(reader, name, value, filename);
+                                    }
+                                }
+                            }
+                            else
                             {
-                                string name = dataElem.Attribute("name").Value;
-                                string value = dataElem.Element("value").Value;
-                                AddResource(reader, name, value, filename);
+                                foreach (IResource resource in MSBuildResXReader.GetResourcesFromFile(filename, shouldUseSourcePath))
```
nguerrera: (I assume this is covered by TODO: " Check to ensure that the reader type is available when not targeting netcoreapp3.0", but I want to make sure....)

When we get in here using .NET Core msbuild and we don't have S.R.Extensions (commonly because we're targeting .NETFramework, .NETStandard, or .NET Core 3.0, we need to make sure we support strings (and hopefully resx-file-ref-to-string if possible since that is commonly reported), but reject other types saying S.R.Extensions is needed. Is that the plan?

rainersigwald: Yes; I'm working on implementing this now. It's going a bit slower than I thought because a bunch of stuff was disabled in the Core build that I need to turn on so I can do the assembly-ref check.

Fileref-to-string should fall out for free, but I'll check it.

src/Tasks/ResourceHandling/LiveObjectResource.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public LiveObjectResource(string name, object value)
+        {
+            Name = name;
+            this.value = value;
+        }
+
+        public string Name { get; }
+        public object value;
+
```
nguerrera: Is there a reason why this is a field and not PascalCased?

src/Tasks/ResourceHandling/StringResource.cs
```diff
@@ -0,0 +1,55 @@
...
+            writer.AddResource(Name, Value);
+        }
+
+        public override bool Equals(object that)
+        {
+            if (that is StringResource other)
+            {
+                return Name == other.Name &&
+                    Value == other.Value;
+                // TODO: also OriginatingFile? I don't think I care
```
nguerrera: Why does this class override Equals?

src/Tasks/ResourceHandling/StringResource.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            return false;
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
nguerrera: Object.GetHashCode isn't compatible with the Equals override above.

src/Tasks.UnitTests/ResourceHandling/TextFileInShiftJIS.txt
```diff
@@ -0,0 +1 @@
+�n���[���[���h�I
```
nguerrera: In the past, corefx tests got broken by various source control moves corrupting encoding. Should we consider emitting this programatically in the test?

src/Tasks/GenerateResource.cs
```diff
@@ -3014,9 +3014,26 @@ private void ReadResources(String filename, bool shouldUseSourcePath, String out
                         ReadResources(reader, resXReader, filename);
                         break;
 #else
-                        foreach (IResource resource in MSBuildResXReader.GetResourcesFromFile(filename, shouldUseSourcePath))
+                        if (Traits.Instance.EscapeHatches.UseMinimalResxParsingInCoreScenarios)
                         {
-                            AddResource(reader, resource, filename, 0, 0);
+                            using (var xmlReader = new XmlTextReader(filename))
```
nguerrera: Consider extracting a method for this legacy escape hatch path.

src/Tasks/GenerateResource.cs
```diff
@@ -2992,40 +3028,57 @@ private void ReadResources(String filename, bool shouldUseSourcePath, String out
...
-                        using (var xmlReader = new XmlTextReader(filename))
+                        else
+#endif
                         {
-                            xmlReader.WhitespaceHandling = WhitespaceHandling.None;
-                            XDocument doc = XDocument.Load(xmlReader, LoadOptions.PreserveWhitespace);
-                            foreach (XElement dataElem in doc.Element("root").Elements("data"))
+                            if (Traits.Instance.EscapeHatches.UseMinimalResxParsingInCoreScenarios)
+                            {
+                                using (var xmlReader = new XmlTextReader(filename))
```
nguerrera: Consider extracting a method for this legacy escape hatch path.

src/Tasks/ResourceHandling/LiveObjectResource.cs
rainersigwald: It was moved from an old class that was super wonky. Updated.

src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
```diff
@@ -382,27 +384,29 @@ public void ForceOutOfDateLinked()
...
-                DateTime time = File.GetLastWriteTime(t.OutputResources[0].ItemSpec);
+                DateTime firstWriteTime = File.GetLastWriteTime(t.OutputResources[0].ItemSpec);
                 System.Threading.Thread.Sleep(200);
-                File.SetLastWriteTime(bitmap, DateTime.Now);
+                File.SetLastWriteTime(bitmap, DateTime.Now + TimeSpan.FromSeconds(2));
 
                 Utilities.ExecuteTask(t2);
 
-                Assert.True(DateTime.Compare(File.GetLastWriteTime(t2.OutputResources[0].ItemSpec), time) > 0);
+                File.GetLastWriteTime(t2.OutputResources[0].ItemSpec).ShouldBeGreaterThan(firstWriteTime);
```
cdmihai: Is there a test that asserts the task is incremental if you run t2 without updating the timestamp of the bitmap?

cdmihai: Ah, yes, there is :)

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,300 @@
...
+                {
+                    switch (elem.Name.LocalName)
+                    {
+                        case "assembly":
+                            ParseAssemblyAlias(aliases, elem);
+                            break;
+                        case "resheader":
+                            break;
+                        case "data":
+                            ParseData(filename, pathsRelativeToBasePath, resources, aliases, elem);
```
cdmihai: nit: make ParseData return `resources`, to make it clear what it's final goal is.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
```diff
@@ -0,0 +1,300 @@
...
+                return StringTypeName;
+            }
+
+            // No alias found. Hope it's sufficiently complete to be resolved at runtime
+            return aliasedTypeName;
+        }
+
+        private static void ParseData(string resxFilename, bool pathsRelativeToBasePath, List<IResource> resources, Dictionary<string,string> aliases, XElement elem)
+        {
+            string name = elem.Attribute("name").Value;
```
cdmihai: Who checks that the xml is in the right resx format?

src/Tasks/ResourceHandling/StringResource.cs
rainersigwald: It was to support some tests, but you're right that it's bogus; I'll write a helper comparer instead.

src/Tasks/ResourceHandling/MSBuildResXReader.cs
rainersigwald: Yeah, I struggled with going strict or permissive here. The current error handling approach is "assume good faith and throw". I'm optimistic that that's ok since most of these files are machine-generated and machine-edited. If it proves not to be in the previews we can make the errors much nicer.

rainersigwald: @BenVillalobos and I came up with a cheap-but-better-than-it-was improvement in 442c03b. I think that's good enough for 16.3p1.

nguerrera: I'm not a fan of that catch (Exception). If I followed right, it will lead to confounding user-facing errors like "MSB3103: Invalid Resx file. Object reference not set to an instance of an object.`" when there is actually a bug in our code.

nguerrera: On this being good enough for 16.3p1, my experience is that we don't go back and make aggressive catches less aggressive. I'd rather leave proper error handling unfinished than hacked for a preview.

PR #4425, opened 2019-06-06
LOC CHECKIN | Microsoft/msbuild master | 20190606

v-zbsail:


src/Deprecated/Conversion/Resources/xlf/Strings.it.xlf
```diff
@@ -142,7 +142,7 @@
         <target state="translated">MSB2010: l'elemento figlio &lt;{0}&gt; dell'elemento &lt;{1}&gt; non è valido.</target>
         <note>{StrBegin="MSB2010: "}It appears we've been asked to convert a project that either wasn't created by Visual Studio, or that has been hand-modified or corrupted.</note>
         <alt-trans match-quality="100%" tool="BlackBox/MSR MT">
-          <target state-qualifier="mt-suggestion">MSB2010: Elemento figlio &lt;{0}&gt; dell'elemento &lt;{1}&gt; non è valido.</target>
+          <target state-qualifier="mt-suggestion">MSB2010: Untergeordnete Element &lt;{0}&gt; Element &lt;{1}&gt; ist ungültig.</target>
```
rainersigwald: @v-zbsail Is this correct? It looks like some of these are changing from Italian to German in the `.it` file.

nguerrera: Why is this node being added at all? This should be kept only on the backend. We won't consume that. @janbaltus

janbaltus: Recently there was switch of tool used for xlf localization.
Now as source is used German version of xlf file.
This cause that generated localized xlf contain  mt-suggestion from German xlf.
This will not cause any issue in final localized build, but might be confusing to see so many differences in PR.
Generally we do not need Mt-suggestion data in xlf as current process is not able to pick this value.

tmeschter: It looks a little weird but it can be ignored.

PR #4434, opened 2019-06-12
Improve the performance of the SdkResolverManifest

jeffkl:
Read the file with an XmlReader instead of using serialization which reduces JIT and overall read time.

Related to #4025

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
@@ -6,7 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.Serialization;
+using System.Xml;
 using Microsoft.Build.BackEnd.Logging;
 using Microsoft.Build.Construction;
 using Microsoft.Build.Framework;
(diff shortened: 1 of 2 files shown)


jeffkl:
This is an alternative to #4182 which seems to have stalled

Local perf shows that before the `SdkResolverManifest.Load()` method took 119ms and how it takes just 1.5ms

image

image

rainersigwald:
Why is this preferable to #4182? I'd generally prefer to land that.

jeffkl:
> Why is this preferable to #4182? I'd generally prefer to land that.

Just a few differences
1. Use XmlReader for maximum perf improvement
2. Reuse existing error strings 

Why hasn't the other PR been merged?

jeffkl:
@rainersigwald are we going to revive #4182 or consider this PR?

jeffkl:
@rainersigwald the unit tests cover if the XML is bad and I ran them under the debugger.

https://github.com/microsoft/msbuild/blob/7b3a457540ab87cf02f6d41ce6095e6ecbbb6ea7/src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs#L211

But here's a sample of what the errors look like:

Invalid element
```
error MSB4245: SDK Resolver manifest file is invalid. This may indicate a corrupt or invalid installation of MSBuild. Manifest file path 'D:\msbuild\artifacts\bin\MSBuild.Bootstrap\Debug\net472\SdkResolvers\Test\Test.xml'. Message: The element <Foo> is unrecognized, or not supported in this context.
```

Invalid XML
```
error MSB4245: SDK Resolver manifest file is invalid. This may indicate a corrupt or invalid installation of MSBuild. Manifest file path 'D:\msbuild\artifacts\bin\MSBuild.Bootstrap\Debug\net472\SdkResolvers\Test\Test.xml'. Message: Unexpected end of file while parsing Name has occurred. Line 1, position 6.
```

Non-existent SDK resolver assembly
```
error MSB4247: Could not load SDK Resolver. A manifest file exists, but the path to the SDK Resolver DLL file could not be found. Manifest file path 'D:\msbuild\artifacts\bin\MSBuild.Bootstrap\Debug\net472\SdkResolvers\Test\Test.xml'. SDK resolver path: C:\does\not\exist.dll
```

jeffkl:
Thanks all!

Issue #2713, opened 2017-11-09
Long term effort for getting off Environment variables (for reproducible builds)


KirillOsenkov:
The fact that MSBuild initializes properties with environment variables is very bad for reproducible builds. CI servers often have different sets of environment variables set, and investigating on the CI server is usually hard.

For compatibility reasons, I doubt we'll ever get rid of reading environment, but I hope there can be "Strict Mode" where environment is not read at all. Builds could opt in to strict mode and use other means of initializing global state.

Example: 
https://github.com/NuGet/NuGet.Client/blob/2aad8d25e66e9c3737b220631f3661bb6235d14b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L555-L565

Imagine if the CI Server defines the VERSION environment variable and sets it to something like a Git SHA.

You get a build failure like this:
```
/Library/Frameworks/Mono.framework/Versions/5.8.0/lib/mono/xbuild/NuGet.targets(102,5): error : 'fbc5117dcb21f0d180c4b199d828d4f4b9d6f4a8' is not a valid version string. [/Users/builder/data/lanes/5668/fbc5117d/source/md-addins/MacSetup/build/BinaryChecker.proj]
/Library/Frameworks/Mono.framework/Versions/5.8.0/lib/mono/xbuild/NuGet.targets(102,5): error : Parameter name: value [/Users/builder/data/lanes/5668/fbc5117d/source/md-addins/MacSetup/build/BinaryChecker.proj]
```

Tracking it down from that failure to that line to NuGet.targets is highly non-trivial.

KirillOsenkov:
Example issue: https://github.com/NuGet/Home/issues/6154

nguerrera:
#1144

david-driscoll:
Years ago when I got a new PC at work (an hp), as part of the default install of windows (from hp or internally not sure) there were a few environment variables that were set, one was `PLATFORM` which completely hosed visual studio until I figured it out.

dasMulli:
Related: https://github.com/dotnet/sdk/issues/481

Also looks like some folks are using `VERSION` env vars on docker images that throw off builds. Was diagnosing the impact and workarounds with @niemyjski - e.g. `TreatAsLocalProperty`.

kzu:
Just like `/noautoresponse`, I'd name this something like `/noenvironment` or `/novars` or (uglier?) `/noenvvars`

nguerrera:
[...]

PR #4437, opened 2019-06-13
Implementing tracking of environment variables

### LINKED ISSUE #3432: Keep track of which global properties/env variables were actually used

maneely:
Implements tracking of environment variables (and uninitialized properties) for graph evaluation.

This allows users of ProjectGraph to better understand what parts of the environment will actually affect the build. This can help reduce the set of environment variables tracked and can potentially reduce unnecessary rebuilds due to environment changes.

* This is an opt-IN feature, set with a new build parameter: MSBUILDTRACKENVVARREADS

Resolves #3432 
Related to #2713

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/Build/BackEnd/BuildManager/BuildParameters.cs
@@ -97,6 +97,11 @@ namespace Microsoft.Build.Execution
         private static string s_msbuildExeKnownToExistAt;
 
         /// <summary>
+        /// Specifies whether or not MSBuild will track environment variable reads
+        /// </summary>
+        private static bool? s_trackReadsOnEnvironmentVariables;
+
+        /// <summary>
         /// The build id
         /// </summary>
         private int _buildId;
@@ -761,6 +766,15 @@ namespace Microsoft.Build.Execution
         {
             get => _outputResultsCacheFile;
             set => _outputResultsCacheFile = value;
+        }
+
+        /// <summary>
+        /// Track the environment variables actually read and used.
+        /// </summary>
+        public bool TrackReadsOnEnvironmentVariables
+        {
+            get => GetStaticBoolVariableOrDefault("MSBUILDTRACKENVVARREADS", ref s_trackReadsOnEnvironmentVariables, false);
+            set => s_trackReadsOnEnvironmentVariables = value;
         }
 
         /// <summary>
--- a/src/Build/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/Build/BackEnd/BuildManager/BuildParameters.cs
@@ -97,11 +97,6 @@ namespace Microsoft.Build.Execution
         private static string s_msbuildExeKnownToExistAt;
 
         /// <summary>
-        /// Specifies whether or not MSBuild will track environment variable reads
-        /// </summary>
-        private static bool? s_trackReadsOnEnvironmentVariables;
-
-        /// <summary>
         /// The build id
         /// </summary>
         private int _buildId;
@@ -766,15 +761,6 @@ namespace Microsoft.Build.Execution
         {
             get => _outputResultsCacheFile;
             set => _outputResultsCacheFile = value;
-        }
-
-        /// <summary>
-        /// Track the environment variables actually read and used.
-        /// </summary>
-        public bool TrackReadsOnEnvironmentVariables
-        {
-            get => GetStaticBoolVariableOrDefault("MSBUILDTRACKENVVARREADS", ref s_trackReadsOnEnvironmentVariables, false);
-            set => s_trackReadsOnEnvironmentVariables = value;
         }
 
         /// <summary>
--- a/src/Build/Definition/Project.cs
+++ b/src/Build/Definition/Project.cs
@@ -3371,7 +3371,7 @@ namespace Microsoft.Build.Evaluation
             /// <summary>
             /// Sets a property which is not derived from Xml.
             /// </summary>
-            public ProjectProperty SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved)
+            public ProjectProperty SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved, bool isEnvVar = false)
             {
                 ProjectProperty property = ProjectProperty.Create(Project, name, evaluatedValueEscaped, isGlobalProperty, mayBeReserved);
                 Properties.Set(property);
--- a/src/Build/Evaluation/Evaluator.cs
+++ b/src/Build/Evaluation/Evaluator.cs
@@ -1256,7 +1256,7 @@ namespace Microsoft.Build.Evaluation
 
             foreach (ProjectPropertyInstance environmentProperty in _environmentProperties)
             {
-                P property = _data.SetProperty(environmentProperty.Name, ((IProperty)environmentProperty).EvaluatedValueEscaped, false /* NOT global property */, false /* may NOT be a reserved name */);
+                P property = _data.SetProperty(environmentProperty.Name, ((IProperty)environmentProperty).EvaluatedValueEscaped, false /* NOT global property */, false /* may NOT be a reserved name */, true /* IS environment variable. */);
                 environmentPropertiesList.Add(property);
             }
 
--- a/src/Build/Evaluation/IEvaluatorData.cs
+++ b/src/Build/Evaluation/IEvaluatorData.cs
@@ -263,7 +263,7 @@ namespace Microsoft.Build.Evaluation
         /// <summary>
         /// Sets a property which does not come from the Xml.
         /// </summary>
-        P SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved);
+        P SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved, bool isEnvVar = false);
 
         /// <summary>
         /// Sets a property which comes from the Xml.
--- a/src/Build/Evaluation/IEvaluatorData.cs
+++ b/src/Build/Evaluation/IEvaluatorData.cs
@@ -263,7 +263,7 @@ namespace Microsoft.Build.Evaluation
         /// <summary>
         /// Sets a property which does not come from the Xml.
         /// </summary>
-        P SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved, bool isEnvVar = false);
+        P SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved, bool isEnvironmentVariable = false);
 
         /// <summary>
         /// Sets a property which comes from the Xml.
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -74,6 +75,16 @@ namespace Microsoft.Build.Experimental.Graph
         public IReadOnlyCollection<ProjectGraphNode> ProjectNodesTopologicallySorted => _projectNodesTopologicallySorted.Value;
 
         public IReadOnlyCollection<ProjectGraphNode> GraphRoots { get; }
+
+        /// <summary>
+        ///     Gets a distinct enumeration of all of the environment variables reads during evaluation.
+        /// </summary>
+        public IEnumerable<string> EnvironmentVariableReads => _projectNodesTopologicallySorted?.Value.Select(p => p.ProjectInstance).SelectMany(pi => pi.EnvironmentVariableReads).Distinct();
+
+        /// <summary>
+        ///     Get a distinct enumeration of all the uninitialized property reads during evaluation.
+        /// </summary>
+        public IEnumerable<string> UninitializedPropertyReads => _projectNodesTopologicallySorted?.Value.Select(p => p.ProjectInstance).SelectMany(pi => pi.UninitializedPropertyReads).Distinct();
 
         /// <summary>
         ///     Constructs a graph starting from the given project file, evaluating with the global project collection and no
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -94,10 +94,7 @@ namespace Microsoft.Build.Experimental.Graph
                 {
                     foreach (string environmentVariableRead in projectNode.ProjectInstance.EnvironmentVariableReads)
                     {
-                        if (!distinctReads.Contains(environmentVariableRead))
-                        {
-                            distinctReads.Add(environmentVariableRead);
-                        }
+                        distinctReads.Add(environmentVariableRead);
                     }
                 }
 
@@ -123,10 +120,7 @@ namespace Microsoft.Build.Experimental.Graph
                 {
                     foreach (string uninitializedPropertyRead in projectNode.ProjectInstance.UninitializedPropertyReads)
                     {
-                        if (!distinctUninitializedProperties.Contains(uninitializedPropertyRead))
-                        {
-                            distinctUninitializedProperties.Add(uninitializedPropertyRead);
-                        }
+                        distinctUninitializedProperties.Add(uninitializedPropertyRead);
                     }
                 }
 
--- a/src/Build/Graph/ProjectGraph.cs
+++ b/src/Build/Graph/ProjectGraph.cs
@@ -79,12 +79,65 @@ namespace Microsoft.Build.Experimental.Graph
         /// <summary>
         ///     Gets a distinct enumeration of all of the environment variables reads during evaluation.
         /// </summary>
-        public IEnumerable<string> EnvironmentVariableReads => _projectNodesTopologicallySorted?.Value.Select(p => p.ProjectInstance).SelectMany(pi => pi.EnvironmentVariableReads).Distinct();
+        public IEnumerable<string> EnvironmentVariableReads
+        {
+            get
+            {
+                if (!ProjectNodes.Any())
+                {
+                    return Enumerable.Empty<string>();
+                }
+
+                var distinctReads = new HashSet<string>();
+
+                foreach (ProjectGraphNode projectNode in ProjectNodes)
+                {
+                    foreach (string environmentVariableRead in projectNode.ProjectInstance.EnvironmentVariableReads)
+                    {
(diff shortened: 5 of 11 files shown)


src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -763,6 +768,15 @@ public string OutputResultsCacheFile
             set => _outputResultsCacheFile = value;
         }
 
+        /// <summary>
+        /// Track the environment variables actually read and used.
+        /// </summary>
+        public bool TrackReadsOnEnvironmentVariables
+        {
+            get => GetStaticBoolVariableOrDefault("MSBUILDTRACKENVVARREADS", ref s_trackReadsOnEnvironmentVariables, false);
```
smera: >GetStaticBoolVariableOrDefault("MSBUILDTRACKENVVARREADS", ref s_tr [](start = 19, length = 66)

should we expose this as part of the public API of MSBuild, instead of using an env var?
My rationale is: asking to track env vars from a command line execution of MSBuild never makes sense, and this is likely only going to be used by the static graph APIs. And an explicit flag in an API is always cleaner and more explicit than an env var

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1256,7 +1256,7 @@ private ICollection<P> AddEnvironmentProperties()
 
             foreach (ProjectPropertyInstance environmentProperty in _environmentProperties)
             {
-                P property = _data.SetProperty(environmentProperty.Name, ((IProperty)environmentProperty).EvaluatedValueEscaped, false /* NOT global property */, false /* may NOT be a reserved name */);
+                P property = _data.SetProperty(environmentProperty.Name, ((IProperty)environmentProperty).EvaluatedValueEscaped, false /* NOT global property */, false /* may NOT be a reserved name */, true /* IS environment variable. */);
```
smera: >/* IS environment variable. * [](start = 207, length = 29)

nit: you could use named args

src/Build/Instance/ProjectInstance.cs
```diff
@@ -1314,11 +1329,30 @@ ICollection<ProjectItemInstance> IItemProvider<ProjectItemInstance>.GetItems(str
...
+                        // If it's a new env var, track it.
+                        _environmentVariables.Add(name);
+                    }
+                }
+                else if (_environmentVariables.Contains(name))
+                {
+                    // If it's NOT an env var but the property has the same name,
+                    // then we're not dependent on the env var any more: remove it.
+                    _environmentVariables.Remove(name);
+                }
```
smera: I lack enough context here. But if this mean the given name was first tracked as an env var and then replaced with a property, could it happen that it was queried after it was added as an env var, but before it was replaced by a property?

If the assumption is that this never happens, maybe we should add some invariant assertion (maybe under debug if we think it will be expensive) that states that after a 
TrackPropertyRead, then we should never reach Remove(name)

src/Build/Instance/ProjectInstance.cs
```diff
@@ -2710,5 +2771,30 @@ private void CreatePropertiesSnapshot(Evaluation.Project.Data data, bool isImmut
...
         }
+
+        /// <summary>
+        /// Tracks this property read if it's from an environment variable or uninitialized.
+        /// </summary>
+        /// <param name="name">The name of the property being read.</param>
+        private void TrackPropertyRead(string name, bool propertyMissing)
+        {
+            if (string.IsNullOrEmpty(name))
+                return;
```
smera: does a null or empty name has any special semantics?
Should this be an assert instead?

src/Build/Instance/ProjectInstance.cs
```diff
@@ -2710,5 +2771,30 @@ private void CreatePropertiesSnapshot(Evaluation.Project.Data data, bool isImmut
                 _properties.Set(instance);
             }
         }
+
+        /// <summary>
+        /// Tracks this property read if it's from an environment variable or uninitialized.
+        /// </summary>
+        /// <param name="name">The name of the property being read.</param>
+        private void TrackPropertyRead(string name, bool propertyMissing)
```
smera: >propertyMissing [](start = 57, length = 15)

nit: 'propertyUninitialized' sounds more in line with what you called it above

src/Build/Graph/ProjectGraph.cs
```diff
@@ -75,6 +76,16 @@ public sealed class ProjectGraph
 
         public IReadOnlyCollection<ProjectGraphNode> GraphRoots { get; }
 
+        /// <summary>
+        ///     Gets a distinct enumeration of all of the environment variables reads during evaluation.
+        /// </summary>
+        public IEnumerable<string> EnvironmentVariableReads => _projectNodesTopologicallySorted?.Value.Select(p => p.ProjectInstance).SelectMany(pi => pi.EnvironmentVariableReads).Distinct();
```
smera: >_projectNodesTopologicallySorted [](start = 63, length = 32)

Under what circumstances this is null?
Can we return an empty collection if that's the case instead of null? I'd prefer to avoid forcing consumers to do null checks all over the place

src/Build/Instance/ProjectInstance.cs
maneely: If setting a property overwrites a previous environment variable property we won't track reads of it ANY MORE. However, if reads happened prior to when it was overwritten, those reads ARE captured.

src/Build/Graph/ProjectGraph.cs
smera: >EnvironmentVariableReads [](start = 154, length = 24)

This may be premature optimization, so feel free to ignore my comment, but it occurs to me that there is going to be a relatively big set of common env variables that are the same across every project (platform, configuration, etc).
So by having each project carry its own set, and aggregating that at the project graph level could increase memory footprint unnecessary. Each project could report reads an uninitialize vars to the project graph as they are being discovered, so each item will be there only once, and they can be added to a set or similar so de deduplication will also be unnecessary.

src/Build/Instance/ProjectInstance.cs
smera: Ah, got it. Makes sense!

---
In reply to: [293497304](https://github.com/microsoft/msbuild/pull/4437#discussion_r293497304) [](ancestors = 293497304)

src/Build/Instance/ProjectInstance.cs
maneely: I saw one instance of this and that's why I added it. MSBuild looks up a property called "InnerBuildProperty". If that isn't present, an empty string is returned and it then attempts to look up the value for that property (which is an empty string). This appeared to be current behavior and so instead of an assert I just checked explicitly.

src/Build/Graph/ProjectGraph.cs
maneely: I added them as a simple check, but am not aware of the current (or desired) behavior.

src/Build/Definition/Project.cs
```diff
@@ -3371,7 +3371,7 @@ public IItemDefinition<ProjectMetadata> GetItemDefinition(string itemType)
             /// <summary>
             /// Sets a property which is not derived from Xml.
             /// </summary>
-            public ProjectProperty SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved)
+            public ProjectProperty SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved, bool isEnvVar = false)
```
cdmihai: nit: use full name instead of abbreviations

src/Build/Evaluation/IEvaluatorData.cs
```diff
@@ -263,7 +263,7 @@ List<ProjectItemElement> EvaluatedItemElements
         /// <summary>
         /// Sets a property which does not come from the Xml.
         /// </summary>
-        P SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved);
+        P SetProperty(string name, string evaluatedValueEscaped, bool isGlobalProperty, bool mayBeReserved, bool isEnvVar = false);
```
cdmihai: My suggestion is to add environment variable reads and uninitialized property reads to Project as well, not just ProjectInstance.

In certain cases, ProjectInstances are created from Project instances via [https://github.com/microsoft/msbuild/blob/master/src/Build/Definition/Project.cs#L2696](https://github.com/microsoft/msbuild/blob/master/src/Build/Definition/Project.cs#L2696)

Quickbuild and VS do this. @dfederm, would qb use this information ?

src/Build/Instance/ProjectInstance.cs
```diff
@@ -178,6 +178,11 @@ public class ProjectInstance : IPropertyProvider<ProjectPropertyInstance>, IItem
         private bool _translateEntireState;
         private int _evaluationId = BuildEventContext.InvalidEvaluationId;
 
+        // Fields for tracking property usage:
+        private bool _trackPropertyReads = false;
+        private HashSet<string> _environmentVariables;
+        private HashSet<string> _environmentVariableReads;
```
cdmihai: Evaluations could be happening in parallel, make the state thread safe.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -75,6 +76,16 @@ public sealed class ProjectGraph
...
+
+        /// <summary>
+        ///     Get a distinct enumeration of all the uninitialized property reads during evaluation.
+        /// </summary>
+        public IEnumerable<string> UninitializedPropertyReads => _projectNodesTopologicallySorted?.Value.Select(p => p.ProjectInstance).SelectMany(pi => pi.UninitializedPropertyReads).Distinct();
```
smera: >UninitializedPropertyReads [](start = 35, length = 26)

At least at the static api level, I would rename this to something like 'AbsentEnvironmentVariableProbes'.
Uninitialized properties, even though represent the same concept, involves a couple of hops in the semantic that MSBuild gives to property/env var interactions, and looks somewhat asymmetrical to 'EnvironmentVariableReads'

smera: >public [](start = 8, length = 6)

I'd add another property 'EnvironmentVariablesImpactingBuild' (or similar), that is the union of these two. At least from a buildxl point of view, that's really what we want to know, and present vs. absent variables are not really important

src/Build/Instance/ProjectInstance.cs
cdmihai: I think issuing an event for each of these would be better than storing them in collections:
- There is a lot of duplication in this data, most evaluations will have the same environment reads. This will repetition will become non-trivial when you have 10K evaluations. On the other hand, via an event, the receiver (buildxl, qb, binary log viewer) can choose to deduplicate between multiple project evaluations
- The binary log viewer can only get information via log events, which makes the log events the only way to transport this information. @KirillOsenkov

Implementation wise you could reuse the mechanism that gets used by logging each property reassignment: [https://github.com/microsoft/msbuild/blob/master/src/Build/Evaluation/Evaluator.cs#L1400-L1422](https://github.com/microsoft/msbuild/blob/master/src/Build/Evaluation/Evaluator.cs#L1400-L1422) (@rainersigwald for opinion)

src/Build/Instance/ProjectInstance.cs
cdmihai: Oops, my bad, this is a single project, nothing going on in parallel here.

src/Build/Graph/ProjectGraph.cs
cdmihai: These will force the lazy to evaluate eagerly, essentially cancelling the benefit of the lazy. Maybe make these lazy too, which would also avoid recomputing them on each access. Or use `ProjectNodes` which is the same information, but not sorted.

src/Build/Instance/ProjectInstance.cs
maneely: Updated it.

src/Build/Definition/Project.cs
maneely: Updated.

src/Build/Graph/ProjectGraph.cs
maneely: Added.

src/Build/Instance/ProjectInstance.cs
maneely: I'm not sure how I would propagate events from the ProjectInstance, through the GraphBuilder, up to the ProjectGraph that needs it.

src/Build/Graph/ProjectGraph.cs
maneely: Updated to use ProjectNodes.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -75,6 +76,69 @@ public sealed class ProjectGraph
...
+                    return Enumerable.Empty<string>();
+                }
+
+                var distinctReads = new HashSet<string>();
+
+                foreach (ProjectGraphNode projectNode in ProjectNodes)
+                {
+                    foreach (string environmentVariableRead in projectNode.ProjectInstance.EnvironmentVariableReads)
+                    {
+                        if (!distinctReads.Contains(environmentVariableRead))
```
danmosemsft: You can skip the Contains check and just do add, saving a lookup.

danmosemsft: same below

src/Build/Instance/ProjectInstance.cs
```diff
@@ -1314,11 +1329,31 @@ ICollection<ProjectItemInstance> IItemProvider<ProjectItemInstance>.GetItems(str
...
         {
             // Mutability not verified as this is being populated during evaluation
             ProjectPropertyInstance property = ProjectPropertyInstance.Create(name, evaluatedValueEscaped, mayBeReserved, _isImmutable);
             _properties.Set(property);
+
+            if (_trackPropertyReads)
+            {
+                if (isEnvironmentVariable)
+                {
+                    if (!_environmentVariables.Contains(name))
```
danmosemsft: Again, no need to do Contains.

src/Shared/Traits.cs
```diff
@@ -77,6 +77,11 @@ public Traits()
         /// </summary>
         public readonly bool LogPropertyFunctionsRequiringReflection = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBuildLogPropertyFunctionsRequiringReflection"));
 
+        /// <summary>
+        /// Track the environment variables actually read and used.
+        /// </summary>
+        public readonly bool TrackReadsOnEnvironmentVariables = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBUILDTRACKENVVARREADS"));
```
danmosemsft: It's nice to compare against 1 rather than null/empty because that makes it possible to add other values in future without breaking existing users.

src/Build/Graph/ProjectGraph.cs
maneely: Updated.

src/Build/Instance/ProjectInstance.cs
maneely: Updated.

src/Build/Instance/ProjectInstance.cs
```diff
@@ -1314,11 +1329,28 @@ ICollection<ProjectItemInstance> IItemProvider<ProjectItemInstance>.GetItems(str
...
             _properties.Set(property);
+
+            if (_trackPropertyReads)
+            {
+                if (isEnvironmentVariable)
+                {
+                    // If it's an env var, track it.
+                    _environmentVariables.Add(name);
+                }
+                else if (_environmentVariables.Contains(name))
```
danmosemsft: You still have some cases of "if contains, then remove" or "if not contains, then add" in the context of `HashSet<T>`. In both cases, you can remove the contains to halve the lookups. The collection does not throw if you add an entry that already exists, nor if you remove an entry that does not exist.

src/Build/BackEnd/BuildManager/BuildParameters.cs
maneely: Per our discussion in last week's meeting. It is expected that environment variables are used to switch on/off rare or quirky behaviors in msbuild. I've changed where this is implemented though from BuildParameters to Traits.

src/Build/Evaluation/Evaluator.cs
maneely: It was less change to just add onto the method call.

src/Build/Evaluation/IEvaluatorData.cs
dfederm: QuickBuild mostly just uses `ProjectInstance`. We use a few small pieces of `Project` when we need unevaluated values, but only when necessary. We prefer using `ProjectInstance` when possible.

src/Build/Graph/ProjectGraph.cs
maneely: I've changed this aggregation to perform better (not leaving it to LINQ operations) in more recent commits. Others have suggested using an event mechanism, but as I responded below, I'm not sure how to nicely plumb such a system into the current architecture where it would have pass through ProjectGraph --> GraphBuilder --> ProjectInstance.

src/Build/Instance/ProjectInstance.cs
maneely: Removed with latest update.

src/Build/Instance/ProjectInstance.cs
cdmihai: This shows an interesting consequence of the current design: it tracks properties reads during evaluation, and also after evaluation, when the ProjectInstance is interogated for values via its API. Would it make sense to restrict this to recording only the reads that happen during evaluation? For me it makes sense to restrict it, otherwise it means that you may get different property read data every time you ask for it.

src/Build/Evaluation/IEvaluatorData.cs
cdmihai: Do you initialize the ProjectInstance from a Project?

src/Build/Instance/ProjectInstance.cs
cdmihai: I think it would require moving the logic to the Evaluator, which already logs a related event, that of property reassignments: https://github.com/microsoft/msbuild/blob/master/src/Build/Evaluation/Evaluator.cs#L1400-L1422.
Then either Bxl/qb or ProjectGraph via some flag would hook up a logger that picks up the event from all the project evaluations and dedupes it.

src/Shared/Traits.cs
maneely: Changed in latest commit.

maneely:
Superseded by https://github.com/microsoft/msbuild/pull/4461.
Closing.

PR #4447, opened 2019-06-17
Disable `bin` upload

rainersigwald:
This should speed up the official build considerably in addition to minimizing redundant uploads.


rainersigwald:
Symbol-check build passed: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=2773191.

PR #4452, opened 2019-06-19
WIP: Trimmed NuGet Package Sources

### LINKED ISSUE #4426: Minimize NuGet package sources
I think all of these are canonically found on a blob feed now:

https://github.com/microsoft/msbuild/blob/2b277c1383fd93ce665cac006fcd3e1bb8c2afdd/eng/Versions.props#L9-L12

The last three we may not need at all.

We should see which of these we can remove or replace with a more up to date source.

BenVillalobos:
Fix #4426

Kept `roslyn-tools` due to `MicroBuild.Plugins.SwixBuild` dependency.


BenVillalobos:
Closing this, [here's why](https://github.com/microsoft/msbuild/pull/4500#pullrequestreview-260386514)

PR #4461, opened 2019-06-21
Adding tracking of properties with relevant events.

### LINKED ISSUE #3432: Keep track of which global properties/env variables were actually used

maneely:
### Description
This change introduces opt-in functionality that will track property usage and generate three events:
1. Environment Variables Read  (new)
2. Uninitialized Properties Read (new)
3. Property Reassigned (already exists)

Opt-in is via the environment variable, `MSBuildLogPropertyTracking=1`.

It's implemented via a new class named PropertyTrackingEvaluatorDataWraper that implements IEvaluatorData<> and wraps the Evaluator's instance of IEvaluatorData<> to intercept get and set property calls.

Resolves #3432 
Related to #2713

### Customer Impact

BuildXL needs to know what environment variables took part in a build to correctly generate a graph.  Without this functionality, environment variables that modify the graph will not be properly detected which will produce incorrect cache hits.

### Regression?

No

### Risk

Low risk, net new functionality which is opt-in.  Performance testing shows no impact with these changes.

### Test changes in this PR

New unit tests

### DIFF
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4540,7 +4540,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
                 pc.RegisterLogger(logger);
                 Project project = pc.LoadProject(tempPath.Path);
 
-                bool result = project.Build();
+                project.Build().ShouldBeTrue();
                 Assert.True(result);
 
                 logger
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4541,7 +4541,6 @@ namespace Microsoft.Build.UnitTests.Evaluation
                 Project project = pc.LoadProject(tempPath.Path);
 
                 project.Build().ShouldBeTrue();
-                Assert.True(result);
 
                 logger
                     .AllBuildEvents
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -662,55 +662,6 @@ namespace Microsoft.Build.UnitTests.Evaluation
         }
 
         /// <summary>
-        /// Log when a property is being assigned a new value.
-        /// </summary>
-        [Fact]
-        public void LogPropertyAssignments()
-        {
-            string testtargets = ObjectModelHelpers.CleanupFileContents(@"
-                                <Project xmlns='msbuildnamespace'>
-                                     <PropertyGroup>
-                                         <Prop>OldValue</Prop>
-                                         <Prop>NewValue</Prop>
-                                     </PropertyGroup>
-
-                                  <Target Name=""Test""/>
-                                </Project>");
-
-            string tempPath = Path.GetTempPath();
-            string targetDirectory = Path.Combine(tempPath, "LogPropertyAssignments");
-            string testTargetPath = Path.Combine(targetDirectory, "test.proj");
-
-            bool originalValue = BuildParameters.WarnOnUninitializedProperty;
-            try
-            {
-                BuildParameters.WarnOnUninitializedProperty = true;
-                Directory.CreateDirectory(targetDirectory);
-                File.WriteAllText(testTargetPath, testtargets);
-
-                MockLogger logger = new MockLogger();
-                logger.Verbosity = LoggerVerbosity.Diagnostic;
-                ProjectCollection pc = new ProjectCollection();
-                pc.RegisterLogger(logger);
-                Project project = pc.LoadProject(testTargetPath);
-
-                bool result = project.Build();
-                Assert.True(result);
-                logger.AssertLogContains("Evaluation started");
-                logger.AssertLogContains("Property reassignment");
-                logger.AssertLogContains("Evaluation finished");
-                logger.AssertLogContains("Prop");
-                logger.AssertLogContains("OldValue");
-                logger.AssertLogContains("NewValue");
-            }
-            finally
-            {
-                BuildParameters.WarnOnUninitializedProperty = originalValue;
-                FileUtilities.DeleteWithoutTrailingBackslash(targetDirectory, true);
-            }
-        }
-
-        /// <summary>
         /// If we use a property twice make sure we warn and don't crash due to the dictionary which is holding the used but uninitialized variables..
         /// </summary>
         [Fact]
@@ -4551,6 +4502,76 @@ namespace Microsoft.Build.UnitTests.Evaluation
             project.Build(logger);
             logger.AssertLogContains(
                 ResourceUtilities.FormatResourceStringStripCodeAndKeyword("OM_GlobalProperty", "Foo"));
+        }
+
+        [Fact]
+        public void VerifyPropertyTrackingLogging()
+        {
+           string testtargets = ObjectModelHelpers.CleanupFileContents(@"
+                                <Project>
+                                     <PropertyGroup>
+                                         <Prop>$(DOES_NOT_EXIST)</Prop>
+                                         <EnvVar>$(PATH)</EnvVar>
+                                         <LocalAppData>Overwritten!</LocalAppData>
+                                         <NotEnvVarRead>$(LocalAppData)</NotEnvVarRead>
+                                         <Prop2>Value1</Prop2>
+                                         <Prop2>Value2</Prop2>
+                                     </PropertyGroup>
+                                     <Target Name='Build' />
+                                </Project>");
+
+            string tempPath = Path.GetTempPath();
+            string targetDirectory = Path.Combine(tempPath, Guid.NewGuid().ToString());
+            string testTargetPath = Path.Combine(targetDirectory, "test.proj");
+
+            using (TestEnvironment env = TestEnvironment.Create())
+            {
+                env.SetEnvironmentVariable("MsBuildLogPropertyTracking", "1");
+
+                try
+                {
+                    BuildParameters.WarnOnUninitializedProperty = true;
+                    Directory.CreateDirectory(targetDirectory);
+                    File.WriteAllText(testTargetPath, testtargets);
+
+                    MockLogger logger = new MockLogger();
+                    logger.Verbosity = LoggerVerbosity.Diagnostic;
+                    ProjectCollection pc = new ProjectCollection();
+                    pc.RegisterLogger(logger);
+                    Project project = pc.LoadProject(testTargetPath);
+
+                    bool result = project.Build();
+                    Assert.True(result);
+
+                    logger
+                        .AllBuildEvents
+                        .OfType<UninitializedPropertyReadEventArgs>()
+                        .Any(p => p.PropertyName == "DOES_NOT_EXIST")
+                        .ShouldBeTrue();
+
+                    logger
+                        .AllBuildEvents
+                        .OfType<EnvironmentVariableReadEventArgs>()
+                        .Any(ev => ev.EnvironmentVariableName == "PATH")
+                        .ShouldBeTrue();
+
+                    logger
+                        .AllBuildEvents
+                        .OfType<EnvironmentVariableReadEventArgs>()
+                        .Any(ev => ev.EnvironmentVariableName == "LocalAppData")
+                        .ShouldBeFalse();
+
+                    logger
+                        .AllBuildEvents
+                        .OfType<PropertyReassignmentEventArgs>()
+                        .Any(r => r.PropertyName == "Prop2" && r.PreviousValue == "Value1" && r.NewValue == "Value2")
+                        .ShouldBeTrue();
+                }
+                finally
+                {
+                    FileUtilities.DeleteWithoutTrailingBackslash(targetDirectory, true);
+                }
+            }
         }
 
 #if FEATURE_HTTP_LISTENER
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4511,70 +4511,61 @@ namespace Microsoft.Build.UnitTests.Evaluation
         [Fact]
         public void VerifyPropertyTrackingLogging()
         {
-           string testtargets = ObjectModelHelpers.CleanupFileContents(@"
+           string testTargets = ObjectModelHelpers.CleanupFileContents(@"
                                 <Project>
                                      <PropertyGroup>
                                          <Prop>$(DOES_NOT_EXIST)</Prop>
-                                         <EnvVar>$(PATH)</EnvVar>
-                                         <LocalAppData>Overwritten!</LocalAppData>
-                                         <NotEnvVarRead>$(LocalAppData)</NotEnvVarRead>
+                                         <EnvVar>$(DEFINED_ENVIRONMENT_VARIABLE2)</EnvVar>
+                                         <DEFINED_ENVIRONMENT_VARIABLE>Overwritten!</DEFINED_ENVIRONMENT_VARIABLE>
+                                         <NotEnvVarRead>$(DEFINED_ENVIRONMENT_VARIABLE)</NotEnvVarRead>
                                          <Prop2>Value1</Prop2>
                                          <Prop2>Value2</Prop2>
                                      </PropertyGroup>
                                      <Target Name='Build' />
                                 </Project>");
 
-            string tempPath = Path.GetTempPath();
-            string targetDirectory = Path.Combine(tempPath, Guid.NewGuid().ToString());
-            string testTargetPath = Path.Combine(targetDirectory, "test.proj");
-
             using (TestEnvironment env = TestEnvironment.Create())
             {
                 env.SetEnvironmentVariable("MsBuildLogPropertyTracking", "1");
+                env.SetEnvironmentVariable("DEFINED_ENVIRONMENT_VARIABLE", "It's Defined!");
+                env.SetEnvironmentVariable("DEFINED_ENVIRONMENT_VARIABLE2", "It's Defined!");
 
-                try
-                {
-                    BuildParameters.WarnOnUninitializedProperty = true;
-                    Directory.CreateDirectory(targetDirectory);
-                    File.WriteAllText(testTargetPath, testtargets);
+                var tempPath = env.CreateFile(Guid.NewGuid().ToString(), testTargets);
 
-                    MockLogger logger = new MockLogger();
-                    logger.Verbosity = LoggerVerbosity.Diagnostic;
-                    ProjectCollection pc = new ProjectCollection();
-                    pc.RegisterLogger(logger);
-                    Project project = pc.LoadProject(testTargetPath);
+                BuildParameters.WarnOnUninitializedProperty = true;
 
-                    bool result = project.Build();
-                    Assert.True(result);
+                MockLogger logger = new MockLogger();
+                logger.Verbosity = LoggerVerbosity.Diagnostic;
+                ProjectCollection pc = new ProjectCollection();
+                pc.RegisterLogger(logger);
+                Project project = pc.LoadProject(tempPath.Path);
 
-                    logger
-                        .AllBuildEvents
-                        .OfType<UninitializedPropertyReadEventArgs>()
-                        .Any(p => p.PropertyName == "DOES_NOT_EXIST")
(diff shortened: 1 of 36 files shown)


src/Framework/EnvironmentVariableReadEventArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+
+namespace Microsoft.Build.Framework
+{
+    /// <summary>
+    /// Arguments for the environment variable read event.
+    /// </summary>
+    [Serializable]
+    public class EnvironmentVariableReadEventArgs : BuildMessageEventArgs
```
cdmihai: nit: maybe make the names uniform, either <Propertyoperation>Property or viceversa. For example: `UninitializedPropertyReadEventArgs`, `ReassignedPropertyEventArgs`, `EnvironmentVariableAddedAsPropertyEventArgs`

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1256,7 +1274,7 @@ private ICollection<P> AddEnvironmentProperties()
 
             foreach (ProjectPropertyInstance environmentProperty in _environmentProperties)
             {
-                P property = _data.SetProperty(environmentProperty.Name, ((IProperty)environmentProperty).EvaluatedValueEscaped, false /* NOT global property */, false /* may NOT be a reserved name */);
+                P property = _data.SetProperty(environmentProperty.Name, ((IProperty)environmentProperty).EvaluatedValueEscaped, false /* NOT global property */, false /* may NOT be a reserved name */, true /* IS an environment variable.*/);
```
cdmihai: nit: can use argument names instead of comments.

src/Build/Evaluation/PropertyTrackingEvaluatorDataWrapper.cs
```diff
@@ -0,0 +1,230 @@
...
+        private readonly EvaluationLoggingContext _evaluationLoggingContext;
+
+        /// <summary>
+        /// Creates an instance of the PropertyTrackingEvaluatorDataWrapper class.
+        /// </summary>
+        /// <param name="dataToWrap">The underlying <see cref="IEvaluatorData{P,I,M,D}"/> to wrap for property tracking.</param>
+        /// <param name="evaluationLoggingContext">The <see cref="EvaluationLoggingContext"/> used to log relevant events.</param>
+        public PropertyTrackingEvaluatorDataWrapper(IEvaluatorData<P, I, M, D> dataToWrap, EvaluationLoggingContext evaluationLoggingContext)
+        {
+            _wrapped = dataToWrap ?? throw new ArgumentNullException(nameof(dataToWrap));
```
cdmihai: There's `ErrorUtilities.VerifyThrowInternalNull` and other such helpers for internal assertions.

src/Framework/EnvironmentVariableReadEventArgs.cs
maneely: The event we want is that an environment variable was read/used. Not that it was added as a property, right?

src/Build/Evaluation/Evaluator.cs
maneely: Changed with latest update.

src/Build/Evaluation/PropertyTrackingEvaluatorDataWrapper.cs
maneely: Changed with the latest commit.

src/Framework/EnvironmentVariableReadEventArgs.cs
cdmihai: Ah right, had gotten confused there :)

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1393,31 +1411,11 @@ private void EvaluatePropertyElement(ProjectPropertyElement propertyElement)
...
-                    "PropertyReassignment",
-                    property.Name,
-                    newValue,
-                    oldValue,
-                    location);
+                P property = _data.SetProperty(
+                    propertyElement,
+                    evaluatedValue,
+                    // If we're tracking property usage, don't first do a get here. The tracker will give false "uninitialized property read" events.
+                    Traits.Instance.LogPropertyTracking ? null : _data.GetProperty(propertyElement.Name));
```
cdmihai: You can change the SetProperty method to return the predecessor as an out variable. That way you can treat the predecessor handling in the wrapper and not do any checks here in the Evaluator.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -636,7 +651,10 @@ private void Evaluate(ILoggingService loggingService, BuildEventContext buildEve
 #endif
                 projectFile = String.IsNullOrEmpty(_projectRootElement.ProjectFileLocation.File) ? "(null)" : _projectRootElement.ProjectFileLocation.File;
 
-                _evaluationLoggingContext = new EvaluationLoggingContext(loggingService, buildEventContext, projectFile);
+                // This will be true when NOT logging property tracking data.
+                if (_evaluationLoggingContext == null)
```
cdmihai: Hmm, it's a bit ugly that the logging service and the build event context can now enter either via this method, or via the constructor. Can you move them all to just the constructor and have this method be parameterless?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -201,11 +196,29 @@ internal class Evaluator<P, I, M, D>
             int submissionId,
             EvaluationContext evaluationContext,
             bool profileEvaluation,
-            bool interactive)
+            bool interactive,
+            ILoggingService loggingService,
+            BuildEventContext buildEventContext)
```
cdmihai: You have to update the callsites to pass in this information right?

cdmihai: Ah, this is private, nevermind :)

---
In reply to: [296953080](https://github.com/microsoft/msbuild/pull/4461#discussion_r296953080) [](ancestors = 296953080)

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -315,6 +327,30 @@ private void Write(CriticalBuildMessageEventArgs e)
             WriteMessageFields(e);
         }
 
+        private void Write(PropertyReassignmentEventArgs e)
```
cdmihai: Also update BuildEventArgsReader

src/Build/Logging/BinaryLogger/BinaryLogRecordKind.cs
```diff
@@ -20,6 +20,9 @@ internal enum BinaryLogRecordKind
         ProjectEvaluationFinished,
         ProjectImported,
         ProjectImportArchive,
-        TargetSkipped
+        TargetSkipped,
```
cdmihai: You also have to bump up the BinaryLogger.FileFormatVErsion: [https://github.com/cdmihai/msbuild/blob/2c7610b154bce223bfefadd147109aac2c631160/src/Build/Logging/BinaryLogger/BinaryLogger.cs#L20-L34](https://github.com/cdmihai/msbuild/blob/2c7610b154bce223bfefadd147109aac2c631160/src/Build/Logging/BinaryLogger/BinaryLogger.cs#L20-L34)

src/Build/Evaluation/Evaluator.cs
maneely: This doesn't make sense to me. This is the only instance in the code where this SetProperty overload is called. It actually doesn't care about the predecessor, it just gets it for the method call.

We could entirely remove the predecessor argument from the SetProperty method definition. The only user of it is the Project.Data class. And its SetProperty implementation could just internally call GetProperty to gets the predecessor.

IMO, the above would clean up the code nicely. But I'm not sure if there's something design-wise I'm overlooking.

src/Build/Logging/BinaryLogger/BinaryLogRecordKind.cs
maneely: Updated with latest commit.

src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs
maneely: Added with latest commit.

src/Build/Evaluation/Evaluator.cs
maneely: I initially just wanted to touch as few lines of code as possible. I've updated it though so that it all happens in the constructor.

src/Build/Evaluation/Evaluator.cs
cdmihai: I agree. My initial reasoning was in the terms of "maybe the original intent is to have all accesses visible via the interface", but there's no real actual usage / benefit of that intent, and as you say, there is only one real consumer, Project.Data.SetProperty. Go ahead and remove the predecessor from the interface, and move the logic to Project.Data.SetProperty.

maneely: Updated next commit.

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4528,6 +4483,76 @@ public void VerifyMSBuildLogsAMessageWhenLocalPropertyCannotOverrideValueOfGloba
...
 
+        [Fact]
+        public void VerifyPropertyTrackingLogging()
+        {
+           string testtargets = ObjectModelHelpers.CleanupFileContents(@"
+                                <Project>
+                                     <PropertyGroup>
+                                         <Prop>$(DOES_NOT_EXIST)</Prop>
+                                         <EnvVar>$(PATH)</EnvVar>
+                                         <LocalAppData>Overwritten!</LocalAppData>
```
cdmihai: LocalAppData does not exist on Unix. Use TestEnvironment.SetEnvironmentVariable to set test env vars before the evaluation instead of using system defined env vars like Path or LocalAppData. Between different OSes and in CI environments it gets quite shaky to use predefined system env vars.

maneely: Updated with latest commit.

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4528,6 +4483,67 @@ public void VerifyMSBuildLogsAMessageWhenLocalPropertyCannotOverrideValueOfGloba
...
+                logger
+                    .AllBuildEvents
+                    .OfType<EnvironmentVariableReadEventArgs>()
+                    .Any(ev => ev.EnvironmentVariableName == "DEFINED_ENVIRONMENT_VARIABLE2")
+                    .ShouldBeTrue();
+
+                logger
+                    .AllBuildEvents
+                    .OfType<EnvironmentVariableReadEventArgs>()
+                    .Any(ev => ev.EnvironmentVariableName == "DEFINED_ENVIRONMENT_VARIABLE")
```
cdmihai: This shouldn't be an environment variable read because it's always overwritten right? So it's not sensitive to whatever value the environment variable had.

cdmihai: Ow, SholdBeFalse, nevermind :)

---
In reply to: [297766658](https://github.com/microsoft/msbuild/pull/4461#discussion_r297766658) [](ancestors = 297766658)

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -28,10 +28,12 @@ public sealed class BinaryLogger : ILogger
...
         // version 6:
-        //   -  Ids and parent ids for the evaluation locations
+        //   - Ids and parent ids for the evaluation locations
         // version 7:
-        //   -  Include ProjectStartedEventArgs.GlobalProperties
-        internal const int FileFormatVersion = 7;
+        //   - Include ProjectStartedEventArgs.GlobalProperties
+        // version 8:
+        //   - new record kinds: EnvironmentVariableRead, PropertyReassignment, UninitializedPropertyRead
+        internal const int FileFormatVersion = 8;
```
AndyGerlicher: Later on in this file (`Initialize` method) we should probably set MsBuildLogPropertyTracking=1 so that the binary logger gets the data.

AndyGerlicher:
@KirillOsenkov FYI

src/Build/Logging/BinaryLogger/BinaryLogger.cs
maneely: Added this to latest commit.

rainersigwald:
Some big questions before diving into the details:

* Does this really need to be opt-in? I expected these to be produced all the time. Is the tracking extremely expensive?
* Did you investigate the scope option we discussed earlier? Why did you go this different route?

maneely:
Thanks for responding, Rainer.

For opt-in, we didn't want each project to be storing a hashset of up to 1K properties each for this. This includes not just environment reads, but reads of uninitialized properties and it adds up.

For scope, we never discussed what you meant by "Build a scope environment" in our mail thread. I asked for clarification but didn't receive a response. In the meantime, @jeffkl , @cdmihai , @AndyGerlicher , @smera , and I contributed to what you see here. Can you explain what you had in mind?

rainersigwald:
I think using the existing scope mechanisms would resolve the impact-of-tracking concerns, especially if the uninitialized-property code follows the existing flag.

MSBuild doesn't have much notion of a [scope](https://en.wikipedia.org/wiki/Scope_(computer_science)), since all properties and items within a project are accessible almost all the time. But there is one:

https://github.com/microsoft/msbuild/blob/404539a96a51ab801862fefad5e987ee6b93e9ec/src/Build/BackEnd/Components/RequestBuilder/Lookup.cs#L20-L54

The actual Scope is later but that comment is much more informative:

https://github.com/microsoft/msbuild/blob/404539a96a51ab801862fefad5e987ee6b93e9ec/src/Build/BackEnd/Components/RequestBuilder/Lookup.cs#L1290

It's not clear to me offhand whether that can be directly used at evaluation time, or a separate related implementation would be required.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -87,9 +90,11 @@ public void Initialize(IEventSource eventSource)
         {
             _initialTargetOutputLogging = Environment.GetEnvironmentVariable("MSBUILDTARGETOUTPUTLOGGING");
             _initialLogImports = Environment.GetEnvironmentVariable("MSBUILDLOGIMPORTS");
+            _initialTrackProperties = Environment.GetEnvironmentVariable("MsBuildLogPropertyTracking");
 
             Environment.SetEnvironmentVariable("MSBUILDTARGETOUTPUTLOGGING", "true");
             Environment.SetEnvironmentVariable("MSBUILDLOGIMPORTS", "1");
+            Environment.SetEnvironmentVariable("MsBuildLogPropertyTracking", "1");
```
rainersigwald: This worries me; I don't want to slow binary logging down too much, even though this is valuable information. I don't want to give anyone an excuse to not turn binlog on. Do we have a clear picture of the impact?

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -506,9 +506,12 @@ public void EmptyPropertyIsThenSetEnvironmentVariableNotSet()
             string testTargetPath = Path.Combine(targetDirectory, "test.proj");
 
             string originalValue = Environment.GetEnvironmentVariable("MSBUILDWARNONUNINITIALIZEDPROPERTY");
+            bool originalValue2 = BuildParameters.WarnOnUninitializedProperty;
```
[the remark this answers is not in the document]
maneely: @rainersigwald , the issue here is that BuildParameters.WarnOnUninitializedProperty is static. As such changing the underlying env var itself might not affect the already read value. This only seems to be an issue in tests. However, would it be better to change this property to non-static and always read directly from the environment for the latest value?

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4528,6 +4483,67 @@ public void VerifyMSBuildLogsAMessageWhenLocalPropertyCannotOverrideValueOfGloba
...
+                Project project = pc.LoadProject(tempPath.Path);
+
+                bool result = project.Build();
+                Assert.True(result);
+
+                logger
+                    .AllBuildEvents
+                    .OfType<UninitializedPropertyReadEventArgs>()
+                    .Any(p => p.PropertyName == "DOES_NOT_EXIST")
+                    .ShouldBeTrue();
```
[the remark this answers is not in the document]
maneely: Will change next commit.

src/Build/Evaluation/PropertyTrackingEvaluatorDataWrapper.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <summary>
+        /// Logic containing what to do when a property is read.
+        /// </summary>
+        /// <param name="name">The name of the property.</param>
+        /// <param name="property">The value of the property that was read (null if there is no value).</param>
+        private void TrackPropertyRead(string name, P property)
+        {
+            // MSBuild looks up a property called "InnerBuildProperty". If that isn't present,
+            // an empty string is returned and it then attempts to look up the value for that property
+            // (which is an empty string). Thus this check.
```
rainersigwald: This check seems good, but is the described behavior a bug, @cdmihai?

src/Build/Logging/BinaryLogger/BinaryLogRecordKind.cs
```diff
@@ -20,6 +20,10 @@ internal enum BinaryLogRecordKind
...
+        PropertyReassignment,
+        UninitializedPropertyRead,
+        EnvironmentVariableRead,
+        PropertyInitialValueSet
```
rainersigwald: nit:
```suggestion
        PropertyInitialValueSet,
```

src/Build/Resources/Strings.resx
```diff
@@ -745,7 +745,7 @@
   </data>
   <data name="PropertyNameInRegistryHasZeroLength" UESanitized="false" Visibility="Public">
     <value>MSB4148: The name of a property stored under the registry key "{0}" has zero length.</value>
-    <comment>{StrBegin="MSB4148: "}</comment>
+    <comment>{StrBegin="MSB4148: "}</comment>
```
rainersigwald:
```suggestion
    <comment>{StrBegin="MSB4148: "}</comment>
```

src/Build/Resources/Strings.resx
```diff
@@ -1769,13 +1769,28 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
       LOCALIZATION: {0} and {1} are file paths
     </comment>
   </data>
+  <data name="UninitializedPropertyRead" UESanitized="false" Visibility="Public">
+    <value>MSB4261: Read uninitialized property "{0}"</value>
```
rainersigwald: These informational messages aren't usually given error/warning codes. Do you need them?

maneely: I'm not sure. When is the decision made to include one or note?

src/Build/Logging/BinaryLogger/BinaryLogger.cs
maneely: This was removed.

src/Build/Resources/Strings.resx
danmosemsft: As a user, if I see a code, I expect it is something I should probably fix, and not ignore. Otherwise I"m confused.

maneely: @rainersigwald , do you agree with this? If so, I can remove the codes from the three added messages. Thanks, @danmosemsft !

rainersigwald: My thinking is more like "is this a straightforward status message, or something I need to search for when I see it?" but I think it mostly rolls out the same.

That policy is consistent with the codebase (not shocking, since Dan was the dev lead for MSBuild for a long time). Remove 'em :)

maneely: K. Will do next commit! Thanks.

edgarrs:
AB#1550971

PR #4463, opened 2019-06-21
[Static graph] Instantiate graph from solution

cdmihai:
Supports:
- convert solution projects to graph entry points
- solution only dependencies are injected into the graph
- solution configurations to project configurations mapping via global properties

Does not support:
- solution only global properties like SolutionPath
- nested solutions (ProjectReference to sln)
- before / after solution magical imports
- multiple entry points when a solution is present
- select individual projects via solution (/t:"1:Clean")


src/Build/Graph/ProjectGraphNode.cs
```diff
@@ -3,13 +3,15 @@
...
 using System.Diagnostics;
+using Microsoft.Build.BackEnd;
 using Microsoft.Build.Execution;
 using Microsoft.Build.Shared;
 
 namespace Microsoft.Build.Experimental.Graph
 {
     /// <summary>
     /// Represents the node for a particular project in a project graph.
+    /// A node is defined by (ProjectPath, ToolsVersion, GlobalProperties).
```
dfederm: I don't think `ToolsVersion` is actually part of it. Unless things changed from the initial design.

cdmihai: [They do contribute to the project's unique ID](https://github.com/cdmihai/msbuild/blob/2c7610b154bce223bfefadd147109aac2c631160/src/Build/BackEnd/Shared/ConfigurationMetadata.cs#L165-L168), but now that we're practically ignoring ToolsVersion all over the place it's not that important anymore.

src/Build/Graph/ParallelWorkSet.cs
```diff
@@ -188,8 +188,14 @@ private void ExecuteWorkItem()
...
             {
-                TResult _ = workItem.Value;
-                Interlocked.Decrement(ref _pendingCount);
+                try
+                {
+                    TResult _ = workItem.Value;
+                }
+                finally
+                {
+                    Interlocked.Decrement(ref _pendingCount);
```
cdmihai: @ccastanedaucf I had to put the decrement in a finally block, otherwise an exception from `workItem.Value` would miss the decrement, and block the main thread forever. Probably a good idea to move this to qb as well?

ccastanedaucf: Sounds good, I also caught this today in a branch I have where I'm filling out unit tests for ParallelWorkSet

cdmihai: Nice. Maybe worth considering pulling the class into a separate repo with it's own nuget package :)

src/Build.UnitTests/Graph/GraphTestingUtilities.cs
```diff
@@ -0,0 +1,400 @@
...
+                projectNumber,
+                projectReferences,
+                projectReferenceTargets,
+                // Use "Build" when the default target is unspecified since in practice that is usually the default target.
+                defaultTargets ?? "Build",
+                extraContent);
+        }
+    }
+
+    internal class SolutionFileBuilder
```
rainersigwald: extract to its own file?

src/Build.UnitTests/Graph/GraphTestingUtilities.cs
```diff
@@ -0,0 +1,400 @@
...
+                        true)
+                    )
+                );
+
+            var sb = new StringBuilder();
+
+            sb.AppendLine($@"
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version {VisualStudioVersion.VersionLatest}
+VisualStudioVersion = {VisualStudioVersion.VersionLatest}.0.0.0
```
rainersigwald: Since the content of the file is fairly fixed I think it'd be more accurate to pin this at the version that you took the format from.

src/Build.UnitTests/Graph/GraphLoadedFromSolution_tests.cs
```diff
@@ -0,0 +1,731 @@
...
+            var graph = CreateProjectGraph(
+                _env,
+                edges,
+                new Dictionary<string, string>()
+                {
+                    {"Configuration", "Debug"},
+                    {"Platform", "AnyCPU"}
+                });
+
+            // Use ConfigutationMetadata because it is IEquatable, whereas ProjectGraphNode is node.
```
rainersigwald: node? "not"?

src/Build.UnitTests/Graph/GraphLoadedFromSolution_tests.cs
```diff
@@ -0,0 +1,731 @@
...
+                }
+                : new Dictionary<string, string>();
+
+            var graphFromSolution = new ProjectGraph(
+                new ProjectGraphEntryPoint(
+                    solutionPath,
+                    globalProperties),
+                _env.CreateProjectCollection().Collection);
+
+            // in the solution, all nodes are entry points
```
rainersigwald: An interesting idea. Why isn't the solution metaproject the entry point?

rainersigwald:
Notes:

* Caveats are ok
* @AndyGerlicher reports that this works for this repo. But it shouldn't, because Arcade builds a solution as a ProjectReference. @cdmihai can you chase that down please?

src/Build/Construction/Solution/ProjectInSolution.cs
```diff
@@ -176,13 +178,16 @@ internal set
         /// List of guids, in "{XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX}" form, mapping to projects 
         /// that this project has a build order dependency on, as defined in the solution file. 
         /// </summary>
-        public IReadOnlyList<string> Dependencies => _dependencies.AsReadOnly();
+        public IReadOnlyList<string> Dependencies => _dependenciesAsReadonly ?? (_dependenciesAsReadonly = _dependencies.AsReadOnly());
```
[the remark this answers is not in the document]
cdmihai: True, I'll reset them on mutation.

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1276,19 +1274,21 @@ internal static void AssertDictionariesEqual(IDictionary<string, string> x, IDic
                 });
         }
 
-        internal static void ShouldBeEquivalentTo<K, V>(this IReadOnlyDictionary<K, V> a, IReadOnlyDictionary<K, V> b)
+        internal static void ShouldBeEquivalentTo<K, V>(this IDictionary<K, V> a, IReadOnlyDictionary<K, V> b)
         {
             a.ShouldBeSubsetOf(b);
             b.ShouldBeSubsetOf(a);
+            a.Count.ShouldBe(b.Count);
```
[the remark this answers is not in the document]
cdmihai: Order does not matter, but count does. `[1,2,3]` is equivalent to `[3,2,1]`, but [1,2,3] is not equivalent to `[1, 1, 2, 2, 3, 3]`. For when neither order nor count is important, there's `ShouldBeSetEquivalentTo`

src/Build.UnitTests/Graph/GraphLoadedFromSolution_tests.cs
cdmihai: I thought about that idea but the metaproj looks horrible and is full of arbitrary accumulated garbage. Also, the graph does not work with in memory projects.
So I decided to be explicit about what behaviour to cherry-pick into the graph. I'd be fine going back to a metaproj, but then I'd rather redesign the metaproj to make it more human readable.

cdmihai:
> @AndyGerlicher Andy Gerlicher FTE reports that this works for this repo. But it shouldn't, because Arcade builds a solution as a ProjectReference. @cdmihai Mihai Codoban FTE can you chase that down please?

It turns out the graph based build wasn't getting propagated to the actual build. Once we fixed that, it started failing for other reasons, so no connection to this PR.

src/Build/Resources/Strings.resx
```diff
@@ -1769,13 +1769,35 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
       LOCALIZATION: {0} and {1} are file paths
     </comment>
   </data>
+  <data name="StaticGraphAcceptsSingleSolutionEntryPoint" UESanitized="false" Visibility="Public">
+    <value>MSB4261: Static graph only accepts a solution as the only entry point. Multiple entry points detected: {0}</value>
```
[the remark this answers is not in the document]
cdmihai: What about now?

PR #4468, opened 2019-06-24
Add option to not keep build results after EndBuild

panopticoncentral:
When we do design-time builds in Visual Studio, both the old and new project systems cache the build results themselves so that they can provide the results without having to start a build in MSBuild at all. That means that the cached build results that are held onto by their BuildManager after EndBuild are never going to be utilized -- either we're going to satisfy things from our own cache, or we know the build needs to be run (and the results are stale). This ends up in wasted space on the managed heap, which can become significant in a large-ish solution.

We could call ResetCaches but this imposes a small penalty because it also clears the configuration cache as well, which we would like to keep around since it likely will still apply to the next build.

This PR proposes a new property on the BuildParameters that allows specifying that the BuildManager shouldn't keep build results around after EndBuild. As always, the semantics of MSBuild can be obscure, so hopefully this is the right way to approach the problem.


panopticoncentral:
Thanks @livarcocc, I don't have permission to merge, can you hit the button for me?

davkean:
It's not clear to me on which side this cache lives on, VS or MSBuild?

panopticoncentral:
This PR allows turning off the BuildManager results cache in MSBuild. The CSProj has its own build cache in CLangBuildMgr, and CPS uses the build cache from ProjectServices. So in those cases, the MSBuild cache is not needed.

panopticoncentral:
The test solution I'm working on is a solution with about 560 projects, so not extremely large. When I open it cold (without design-time caches), after everything settles down (i.e. NuGet restores run, all design-time builds finish), about 171Mb of the 1.8Gb heap (so, ~10%) is Microsoft.Build.BackEnd.ResultCache.

src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -215,6 +215,11 @@ public class BuildParameters : ITranslatable
 
         private string _outputResultsCacheFile;
 
+        /// <summary>
+        /// Whether the build results should be cached for this build after the build ends.
```
cdmihai: Delete comment, or duplicate the property one.

src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -763,6 +768,15 @@ public string OutputResultsCacheFile
             set => _outputResultsCacheFile = value;
         }
 
+        /// <summary>
+        /// Determines whether MSBuild will cache the results of builds after EndBuild to speed up future builds.
+        /// </summary>
+        public bool CacheBuildResults
```
cdmihai: Some bikeshedding on the name.

`CacheBuildResults` sounds a bit generic. When I initially read the name I thought the PR would be about not using a ResultsCache at all, as opposed to just clearing it on EndBuild. Maybe `ResetBuildResultsCache` instead?

panopticoncentral:
@cdmihai I updated it to `SaveBuildResults` since that seems to be more accurate. How does that look?

cdmihai:
`SaveBuildResults` hints at serializing to file. What about `KeepBuildResults` / `ResetBuildResultsCache`?

panopticoncentral:
@cdmihai I took `DiscardBuildResults`

davkean:
> When we do design-time builds in Visual Studio, both the old and new project systems cache the build results themselves so that they can provide the results without having to start a build in MSBuild at all.

@panopticoncentral I don't think this is always the case in CPS - when we use our own build service (when the project has unsaved changes), we don't cache results. Be aware of that when you turn this switch on.

panopticoncentral:
@davkean Right, my plan is to only turn this switch on in ProjectServices when we are doing a remote build. It won't affect other build services.

@cdmihai @rainersigwald I don't have write access to the repo, can either of you push the button?

PR #4471, opened 2019-06-26
Add stub ResolveFrameworkReferences target

dsplaisted:
This will enable the VS project system to call this target to populate the Frameworks node in the solution explorer.  Microsoft.NET.Sdk will override this target (see https://github.com/dotnet/sdk/pull/3355) to provide ResolvedFrameworkReference items as appropriate.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2764,6 +2764,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       DependsOnTargets="ResolveComReferences" />
 
   <!--
+    ============================================================
+                                      ResolveFrameworkReferences
+
+    Overrridden by Microsoft.NET.Sdk to return
+    ResolvedFrameworkReference items in order to populate the
+    Frameworks node of the Solution Explorer in the IDE.
+  -->
+  <Target Name="ResolveFrameworkReferences" />
+  
+  <!--
     ***********************************************************************************************
     ***********************************************************************************************
                                                                 PrepareResources Section


dsplaisted:
@drewnoakes @davkean

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2763,6 +2763,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
 
+  <!--
+    ============================================================
+                                      ResolveFrameworkReferences
+
+    Overrridden by Microsoft.NET.Sdk to return
+    ResolvedFrameworkReference items in order to populate the
+    Frameworks node of the Solution Explorer in the IDE.
+  -->
+  <Target Name="ResolveFrameworkReferences" />
```
nguerrera: Some other stubs have the returns that's expected. I think this helps s documentation at least. @davkean, if the project system gets the stub does it matter if it has returns? Or will it benefits same as explicit return of empty items?

dsplaisted: I was patterning this off of [GetTargetFrameworkProperties](https://github.com/microsoft/msbuild/blob/6c53fccfab0f1a58e8d04f8c57dac058c798dcf7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1723), which doesn't return anything.

nguerrera: Ok. I don't have a strong opinion on the "doc" aspect, but I would see if there's any implication for project system functionally.

davkean: I'm pretty sure this will be fine, but @drewnoakes can you confirm?

drewnoakes: This looks fine to me.

If a project has `FrameworkReference` items and this version of MSBuild but not a version of the SDK that overrides this stub, all the `FrameworkReference` items will appear as unresolved, which seems like correct behaviour.

drewnoakes:
Any idea when we'll see this inserted somewhere for integration from the project system side?

PR #4479, opened 2019-06-27
Add BeforeTargets and AfterTargets to ProjectTargetInstance

dfederm:
`DependsOnTargets` is already exposed through `ProjectTargetInstance`, so this should be fairly non-controversial, I hope :).

The motivation here is that for consumers of the MSBuild API, like [MSBuildPrediction](https://github.com/microsoft/MSBuildPrediction), are forced to use `Project` instead of `ProjectInstance` to get the values of `BeforeTargets` and `AfterTargets` (via `ProjectTargetElement`). This helps alleviate the need for `Project` in those scenarios.

### DIFF
--- a/src/Build.UnitTests/Instance/ProjectInstance_Internal_Tests.cs
+++ b/src/Build.UnitTests/Instance/ProjectInstance_Internal_Tests.cs
@@ -749,6 +749,54 @@ namespace Microsoft.Build.UnitTests.OM.Instance
             }
         }
 
+        [Fact]
+        public void AddTargetAddsNewTarget()
+        {
+            string projectFileContent = @"
+                    <Project>
+                        <Target Name='a' />
+                    </Project>";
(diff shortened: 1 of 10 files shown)


src/Build.UnitTests/Instance/ProjectInstance_Internal_Tests.cs
```diff
@@ -1,22 +1,18 @@
...
 using System.IO;
-using System.Xml;
 using System.Linq;
+using System.Xml;
 using Microsoft.Build.BackEnd;
-using Microsoft.Build.Engine.UnitTests;
+using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Execution;
+using Microsoft.Build.Framework;
```
cdmihai: Add an AddTarget test to assert that it picks up the new target lists?

dfederm: `AddTarget` currently has no UTs, but sure I can add one to help improve coverage.

PR #4483, opened 2019-07-01
Improve static graph parallel evaluation perf

ccastanedaucf:
The current parallel graph construction model recursively creates and runs tasks as project files are encountered during evaluation. However, since there's no way to bound the number of tasks running, it ends up taking hold of several hundred threads to perform mostly CPU-bound work. This results in heavy performance degradation due to context switching on any mid-sized repo. 

This replaces the current implementation with the pattern / data structure that we built to solve a very similar problem for an internal build engine.

Perf comparison from two runs building the graph on an internal mid-sized repo on a 16 core cpu:

**Master branch**
_Uses 381 threads_
```
Constructed project graph with 733 build files in 42.51s.
```
**This PR:**
_Uses 16 threads_
```
Constructed project graph with 733 build files in 5.59s.
```


src/Build/Graph/ProjectGraph.cs
```diff
@@ -317,14 +307,105 @@ public ProjectGraph(ProjectGraphEntryPoint entryPoint, ProjectCollection project
             IEnumerable<ProjectGraphEntryPoint> entryPoints,
             ProjectCollection projectCollection,
             ProjectInstanceFactoryFunc projectInstanceFactory)
+            : this(
+                entryPoints,
+                projectCollection,
+                projectInstanceFactory,
+                Environment.ProcessorCount - 1,
```
dfederm: >Environment.ProcessorCount - 1, [](start = 16, length = 31)

Why -1 here but not in the overload below? Also, why -1 here?

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -4088,8 +4088,8 @@ public void GraphBuildInvalid()
 
             GraphBuildResult result = _buildManager.Build(_parameters, data);
             result.OverallResult.ShouldBe(BuildResultCode.Failure);
-            result.Exception.ShouldBeOfType<AggregateException>().InnerExceptions.Count.ShouldBe(1);
-            result.Exception.ShouldBeOfType<AggregateException>().InnerExceptions[0].ShouldBeOfType<InvalidProjectFileException>().ProjectFile.ShouldBe(project2);
+            result.Exception.ShouldBeOfType<InvalidProjectFileException>();
+            result.Exception.ShouldBeOfType<InvalidProjectFileException>().ProjectFile.ShouldBe(project2);
```
dfederm: Nit: You either don't need this line or don't need the `ShouldBeOfType` in the line below.

src/Build/Graph/ParallelWorkSet.cs
```diff
@@ -0,0 +1,196 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Build.Graph
```
cdmihai: The namespace of the other graph related classes is `Microsoft.Build.Experimental.Graph`

src/Build/Graph/ProjectGraph.cs
ccastanedaucf: Should be ProcessorCount, forgot to update this overload

cdmihai: Well actually, should it be -1 to account for the current thread that runs the constructor? It will participate in the graph constructor too, so -1 ensures thread count matches processor count.

cdmihai: Oh, you also decrement 1 also when instantiating the ParallelWorkSet, so yeah, no -1 here. Maybe replace -1 with a well named constant :)

ccastanedaucf: That's over in the GraphBuilder constructor:
https://github.com/microsoft/msbuild/blob/78d3a5b134554f157b690b38dec53492a50a587c/src/Build/Graph/GraphBuilder.cs#L50-L54

src/Build/Graph/GraphBuilder.cs
```diff
@@ -237,79 +238,31 @@ void AddGraphBuildGlobalVariable(PropertyDictionary<ProjectPropertyInstance> glo
...
-                                    {
-                                        projectsToEvaluate.Enqueue(referenceConfig);
-                                        evaluationWaitHandle.Set();
-                                    }
-                                }
-                            }
-                        });
+            _graphWorkSet.WaitForAllWork();
+            _graphWorkSet.Complete();
+            _graphWorkSet.WaitForCompletion();
```
rainersigwald: I find this calling pattern confusing. Why is an explicit call to `WaitForAllWork` required? I think I understand that you have to `Complete()` to signal "no new incoming work", but the other two calls both feel like "wait for stuff" to me.

ccastanedaucf: I went ahead and merged these into a single call in #4508. It's a holdover from QB where we have two parse phases and multiple work sets which can enqueue work or wait against each other, so these were split into granular methods.

PR #4394, opened 2019-05-22
Reduce ExtractPropertyFunction Allocations

BenVillalobos:
Meant to fix #4085 but can't seem to repro & prove the fix works.

Reduced allocations by making use of ReadOnlySpan. Previous functionality saw lots of .Substring() and .Trim() calls on strings. Now they're being called on ReadOnlySpan and aren't turned into strings until needed at the end of the function.

PR #4484, opened 2019-07-02
VS Setup authoring for System.Resource.Extensions

rainersigwald:
S.Resources.Extensions depends on System.Memory, so I've ripped off @BenVillalobos's S.Memory setup authoring work from #4394, but not the part that actually uses Span.

I did update to 4.5.3 to match the dependency from S.Resource.Extensions.

The actual new dependency was added in the same way.


rainersigwald:
This can't be inserted before a Roslyn build with dotnet/roslyn#36956 is, but it's ready to go.

PR #4485, opened 2019-07-03
Add SimpleProjectRootElementCache

ccastanedaucf:
```ProjectRootElementCache``` is currently the largest source of perf issues in the parsing stage of our internal build engine, holding up anywhere from 35% to 60% of total parse time, almost entirely due to lock contention. It holds a single lock for all operations on the cache as it needs to synchronize access between a dictionary of weak references and LRU cache of strong references, guard against potential concurrent modifications to ```ProjectRootElement``` state, and reload on disk changes. All of these properties are important for the Visual Studio scenario, but not for a pure evaluation where state should not change and the cache is short-lived, and has a large impact on performance when parallelizing project evaluation.

This adds a ```SimpleProjectRootElementCache``` only intended for use in evaluation, which can be enabled via environment variable. It implements the cache through a ConcurrentDictionary and avoids implementing any of the strong/weak reference methods on the original cache. Ideally these would just be removed as a requirement from the base class and  ```SimpleProjectRootElementCache``` would be usable in a full build, but several areas of MSBuild currently seem to take a hard dependency on the strong/weak reference implementation details of the original, so it would take more effort to generalize.

### **Before**
image

### **After**
image

### DIFF
--- a/src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
@@ -14,18 +14,12 @@ namespace Microsoft.Build.UnitTests.Evaluation
             collectionWithDefaultCache.ProjectRootElementCache.ShouldBeOfType<ProjectRootElementCache>();
 
             const string envKey = "MsBuildUseSimpleProjectRootElementCacheConcurrency";
-            string originalEnvVar = Environment.GetEnvironmentVariable(envKey);
 
-            try
+            using (TestEnvironment env = TestEnvironment.Create())
             {
-                Environment.SetEnvironmentVariable(envKey, "true");
-
+                env.SetEnvironmentVariable(envKey, "true");
                 var collectionWithSimpleCache = new ProjectCollection();
                 collectionWithSimpleCache.ProjectRootElementCache.ShouldBeOfType<SimpleProjectRootElementCache>();
-            }
-            finally
-            {
-                Environment.SetEnvironmentVariable(envKey, originalEnvVar);
             }
         }
     }
--- a/src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using Microsoft.Build.Evaluation;
+﻿using Microsoft.Build.Evaluation;
 using Shouldly;
 using Xunit;
 
--- a/src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using Microsoft.Build.Evaluation;
+using Shouldly;
+using Xunit;
+
+namespace Microsoft.Build.UnitTests.Evaluation
+{
+    public class ProjectCollection_Tests
+    {
+        [Fact]
+        public void ProjectRootElementCache_IsDeterminedByEnvironmentVariable()
+        {
+            var collectionWithDefaultCache = new ProjectCollection();
+            collectionWithDefaultCache.ProjectRootElementCache.ShouldBeOfType<ProjectRootElementCache>();
+
+            const string envKey = "MsBuildUseSimpleProjectRootElementCacheConcurrency";
+            string originalEnvVar = Environment.GetEnvironmentVariable(envKey);
+
+            try
+            {
+                Environment.SetEnvironmentVariable(envKey, "true");
+
+                var collectionWithSimpleCache = new ProjectCollection();
+                collectionWithSimpleCache.ProjectRootElementCache.ShouldBeOfType<SimpleProjectRootElementCache>();
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(envKey, originalEnvVar);
+            }
+        }
+    }
+}
--- a/src/Build/Definition/ProjectCollection.cs
+++ b/src/Build/Definition/ProjectCollection.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -18,7 +18,7 @@ using Microsoft.Build.Execution;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Internal;
-
+using Microsoft.Build.Utilities;
 using ForwardingLoggerRecord = Microsoft.Build.Logging.ForwardingLoggerRecord;
 using ILoggingService = Microsoft.Build.BackEnd.Logging.ILoggingService;
 using InternalLoggerException = Microsoft.Build.Exceptions.InternalLoggerException;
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

File: src/Build/Definition/ProjectCollection.cs [hunk 1]
```diff
@@ -309,7 +309,15 @@ public ProjectCollection(IDictionary<string, string> globalProperties, IEnumerab
...
             MaxNodeCount = maxNodeCount;
-            ProjectRootElementCache = new ProjectRootElementCache(false /* do not automatically reload changed files from disk */, loadProjectsReadOnly);
+
+            if (Traits.Instance.UseSimpleProjectRootElementCacheConcurrency)
+            {
+                ProjectRootElementCache = new SimpleProjectRootElementCache();
+            }
+            else
+            {
+                ProjectRootElementCache = new ProjectRootElementCache(false /* do not automatically reload changed files from disk */, loadProjectsReadOnly);
```
cdmihai: Nit: replace comment with argument name.

src/Build/Evaluation/SimpleProjectRootElementCache.cs
```diff
@@ -0,0 +1,127 @@
...
+        internal override ProjectRootElement TryGet(string projectFile, bool? preserveFormatting)
+        {
+            return Get(
+                projectFile,
+                null,
+                false,
+                preserveFormatting);
+        }
+
+        internal override void DiscardStrongReferences()
```
cdmihai: `throw new NotImplementedException();`?

src/Build/Evaluation/ProjectRootElementCacheBase.cs
```diff
@@ -0,0 +1,126 @@
...
+            var cacheDirtied = ProjectRootElementDirtied;
+            cacheDirtied?.Invoke(sender, e);
+        }
+
+        /// <summary>
+        /// Raises the <see cref="ProjectDirtied"/> event.
+        /// </summary>
+        /// <param name="sender">The dirtied project.</param>
+        /// <param name="e">Details on the Project and the change.</param>
+        internal void OnProjectDirtied(Project sender, ProjectChangedEventArgs e)
```
cdmihai: Maybe override these in the simple cache and throw, to assert the simple cache does not account for project mutations?

src/Build/Evaluation/ProjectRootElementCacheBase.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        protected void RaiseProjectRootElementAddedToCacheEvent(ProjectRootElement rootElement)
+        {
+            ProjectRootElementAddedHandler?.Invoke(this, new ProjectRootElementCacheAddEntryEventArgs(rootElement));
+        }
+
+        /// <summary>
+        /// Raises an event which is raised when a project root element is removed from the strong cache.
+        /// </summary>
+        protected void RaiseProjectRootElementRemovedFromStrongCache(ProjectRootElement projectRootElement)
```
cdmihai: Override to throw?

src/Build/Evaluation/SimpleProjectRootElementCache.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        internal override void DiscardImplicitReferences()
+        {
+            throw new NotImplementedException();
+        }
+
+        internal override void DiscardAnyWeakReference(ProjectRootElement projectRootElement)
+        {
+            ErrorUtilities.VerifyThrowArgumentNull(projectRootElement, "projectRootElement");
```
cdmihai: Is this actually called during evaluations, or can it throw?

src/Build/Definition/ProjectCollection.cs
```diff
@@ -309,7 +309,15 @@ public ProjectCollection(IDictionary<string, string> globalProperties, IEnumerab
             _loadedProjects = new LoadedProjectCollection();
             ToolsetLocations = toolsetDefinitionLocations;
             MaxNodeCount = maxNodeCount;
-            ProjectRootElementCache = new ProjectRootElementCache(false /* do not automatically reload changed files from disk */, loadProjectsReadOnly);
+
+            if (Traits.Instance.UseSimpleProjectRootElementCacheConcurrency)
```
cdmihai: Add a test to ProjectCollection to assert that setting this environment property changes affects the cache type.

src/Build/Evaluation/SimpleProjectRootElementCache.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        internal override void AddEntry(ProjectRootElement projectRootElement)
+        {
+            if (_cache.TryAdd(projectRootElement.FullPath, projectRootElement))
+            {
+                RaiseProjectRootElementAddedToCacheEvent(projectRootElement);
+            }
+        }
+
+
```
dfederm: Nit: remove extra newline

src/Build/Evaluation/SimpleProjectRootElementCache.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        internal override ProjectRootElement TryGet(string projectFile)
+        {
+            return TryGet(projectFile, null);
+        }
+
+
```
dfederm: Nit: remove extra newline

src/Build/Microsoft.Build.csproj
```diff
@@ -520,7 +520,9 @@
     <Compile Include="Evaluation\IPropertyProvider.cs" />
     <Compile Include="Evaluation\Preprocessor.cs" />
     <Compile Include="Evaluation\ProjectParser.cs" />
+    <Compile Include="Evaluation\ProjectRootElementCacheBase.cs" />
     <Compile Include="Evaluation\ProjectRootElementCache.cs" />
+        <Compile Include="Evaluation\SimpleProjectRootElementCache.cs" />
```
BenVillalobos: Nit: Remove spacing on this compile element

src/Build/Evaluation/SimpleProjectRootElementCache.cs
danmosemsft: Is there any value in `VerifyThrowArgumentNull` in a non-public API when it will only be hit if MSBuild itself has a bug and the very next line will give you a null reference exception instead?

cdmihai: It's more about making the method contract explicit as executable documentation than anything else. Yes, things will eventually fail one way or another, but at least this way it's explicit.

src/Build/Evaluation/SimpleProjectRootElementCache.cs
ccastanedaucf: Discussed w/ Mihai offline, found it easier to just no-op here since 
```ProjectCollection.UnloadProject()``` calls this, and we preferred this over doing a type check on the simple cache.

File: src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs [hunk 10]
```diff
@@ -0,0 +1,32 @@
...
+            try
+            {
+                Environment.SetEnvironmentVariable(envKey, "true");
+
+                var collectionWithSimpleCache = new ProjectCollection();
+                collectionWithSimpleCache.ProjectRootElementCache.ShouldBeOfType<SimpleProjectRootElementCache>();
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(envKey, originalEnvVar);
```
cdmihai: nit: You can use TestEnvironment to smoothly define environment variables that get automatically reverted when the TestEnvironment gets disposed.

src/Build.UnitTests/Evaluation/ProjectCollection_Tests.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Build.Evaluation;
+using Shouldly;
+using Xunit;
+
+namespace Microsoft.Build.UnitTests.Evaluation
+{
+    public class ProjectCollection_Tests
+    {
+        [Fact]
+        public void ProjectRootElementCache_IsDeterminedByEnvironmentVariable()
```
cdmihai: Maybe add another one liner test which asserts that the beefy implementation is used when the env var is not set :)

cdmihai: Ah, nevermind, the existing test does both :P

PR #4487, opened 2019-07-03
Depend on the 4.8 version of the .NET Framework SDK

tmeschter:


tmeschter:
@rainersigwald @livarcocc Are we ready to move forward with this change?

PR #4492, opened 2019-07-05
Miscellaneous test diagnosability

rainersigwald:
https://dev.azure.com/dnceng/public/_build/results?buildId=252713&view=ms.vss-test-web.build-test-results-tab&runId=6527076&resultId=100097&paneView=debug failed in a way that's not possible to diagnose:

```

Error message
Assert.True() Failure\r\nExpected: True\r\nActual: False

Stack trace
   at Microsoft.Build.UnitTests.ObjectModelHelpers.BuildTempProjectFileWithTargetsExpectSuccess(String projectFileRelativePath, String[] targets, IDictionary`2 additionalProperties) in D:\a\1\s\src\Shared\UnitTests\ObjectModelHelpers.cs:line 926
   at Microsoft.Build.UnitTests.MSBuildTask_Tests.PropertyOverridesContainSemicolon() in D:\a\1\s\src\Tasks.UnitTests\MSBuild_Tests.cs:line 429
```

with no log to look at.

Instead, force callers of that helper method to pass in a log, and plumb XUnit helpers through to all of those callsites so the log actually gets captured.

### DIFF
--- a/src/Shared/UnitTests/ObjectModelHelpers.cs
+++ b/src/Shared/UnitTests/ObjectModelHelpers.cs
@@ -910,34 +910,27 @@ namespace Microsoft.Build.UnitTests
         /// </summary>
         /// <param name="projectFileRelativePath"></param>
         /// <returns></returns>
-        internal static MockLogger BuildTempProjectFileExpectSuccess(string projectFileRelativePath)
+        internal static void BuildTempProjectFileExpectSuccess(string projectFileRelativePath, MockLogger logger)
         {
-            return BuildTempProjectFileWithTargetsExpectSuccess(projectFileRelativePath, null, null);
+            BuildTempProjectFileWithTargetsExpectSuccess(projectFileRelativePath, null, null, logger);
         }
 
         /// <summary>
         /// Builds a project file from disk, and asserts if the build does not succeed.
         /// </summary>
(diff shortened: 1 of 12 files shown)


src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -910,48 +910,27 @@ internal static string CreateFileInTempProjectDirectory(string fileRelativePath,
...
-        internal static MockLogger BuildTempProjectFileExpectFailure(string projectFileRelativePath)
-        {
-            MockLogger logger = new MockLogger();
-            bool success = BuildTempProjectFileWithTargets(projectFileRelativePath, null, null, logger);
-
-            Assert.False(success); // "Build unexpectedly succeeded.  See Standard Out tab for details"
-
-            return logger;
+            BuildTempProjectFileWithTargets(projectFileRelativePath, targets, additionalProperties, logger)
+                .ShouldBeTrue("Build failed.  See Standard Out tab for details");
```
cdmihai: nit: local builds won't have a standard out tab.

rainersigwald: Yeah, I restored the old bogus comment. Let me see if I can come up with something more descriptive.

PR #4493, opened 2019-07-08
Report long-running tests in xunit

rainersigwald:
On my dev box, the longest-running test was 24 seconds, so I'm hoping this is a reasonable time limit.


eng/cibuild_bootstrapped_msbuild.ps1
```diff
@@ -128,6 +128,9 @@ catch {
 }
 finally {
   Pop-Location
+
+  KillProcessesFromRepo
```
cdmihai: It would be comical if this kills any lingering telemetry processes that are draining events, but afaik nothing like that exists.

rainersigwald:
Looks like the current state is that it only works for Mac and Linux:

```powershell
$x = az pipelines runs artifact list --run-id 263131 --organization https://dev.azure.com/dnceng --project public
$c = $x | ConvertFrom-Json

$customobjects = $c | % { [pscustomobject]@{name=$_.name; url = $_.resource.downloadUrl} } 

$customobjects | % {Invoke-WebRequest -OutFile "$($_.name).zip" -Uri $_.url}
```

```sh-session
$ rg "\[Long " --files-with-matches
CoreOnMac build logs\CoreOnMac build logs\Microsoft.Build.CommandLine.UnitTests_netcoreapp2.1_x64.log
CoreOnLinux build logs\CoreOnLinux build logs\Microsoft.Build.CommandLine.UnitTests_netcoreapp2.1_x64.log
```

PR #4497, opened 2019-07-09
Test that timed-out Exec fails regardless of IgnoreExitCode

rainersigwald:
While doing something entirely unrelated, I asked myself "does the `Exec` task fail if it times out, even if `IgnoreExitCode == true`? It does, so I wrote a test to immortalize the current behavior.

### DIFF
--- a/src/Tasks.UnitTests/Exec_Tests.cs
+++ b/src/Tasks.UnitTests/Exec_Tests.cs
@@ -120,6 +120,26 @@ namespace Microsoft.Build.UnitTests
         }
 
         [Fact]
+        public void TimeoutFailsEvenWhenExitCodeIsIgnored()
+        {
+            // On non-Windows the exit code of a killed process is SIGTERM (143)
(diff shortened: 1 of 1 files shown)


src/Tasks.UnitTests/Exec_Tests.cs
```diff
@@ -119,6 +119,26 @@ public void Timeout()
...
 
+        [Fact]
+        public void TimeoutFailsEvenWhenExitCodeIsIgnored()
+        {
+            // On non-Windows the exit code of a killed process is SIGTERM (143)
+            int expectedExitCode = NativeMethodsShared.IsWindows ? -1 : 143;
+
+            Exec exec = PrepareExec(NativeMethodsShared.IsWindows ? ":foo \n goto foo" : "while true; do sleep 1; done");
+            exec.Timeout = 5;
+            exec.IgnoreExitCode = true;
```
rainersigwald: This is a copy-paste of the test above except for this line.

PR #4508, opened 2019-07-13
[Static Graph] Fix hangs and redundant code in ParallelWorkSet

ccastanedaucf:
- Address PR comments from #4483 
- Remove redundant code
- Add unit tests
- Fix potential thread hangs

### DIFF
--- a/src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
+++ b/src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
         private ParallelWorkSet<string, string> _workSet;
 
         [Fact]
-        public void GivenExceptionsOutsideWorker_CompletesAndThrowsException()
+        public void GivenExceptionsOnCompletionThread_CompletesAndThrowsException()
         {
             TestParallelWorkSet(new ParallelWorkSetTestCase
             {
@@ -57,7 +57,7 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
         }
 
         [Fact]
-        public void GivenExceptionsWithinWorker_CompletesAndThrowsExceptions()
+        public void GivenExceptionsOnWorkerThread_CompletesAndThrowsExceptions()
         {
             TestParallelWorkSet(new ParallelWorkSetTestCase
             {
--- a/src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
+++ b/src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
@@ -1,0 +1,253 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Build.UnitTests;
+using Shouldly;
+using Xunit;
+
+namespace Microsoft.Build.Experimental.Graph.UnitTests
+{
+    public class ParallelWorkSet_Tests
+    {
+        private class ParallelWorkSetTestCase
+        {
+            internal int DegreeOfParallelism { get; set; }
+            internal List<WorkItem> WorkItemsToAdd { get; set; } = new List<WorkItem>();
+
+            internal Dictionary<string, string> ExpectedCompletedWork =
+                new Dictionary<string, string>(StringComparer.Ordinal);
+            internal bool ShouldExpectException { get; set; }
+        }
+
+        private struct WorkItem
+        {
+            internal string Key { get; set; }
+
+            internal Func<string> WorkFunc { get; set; }
+        }
+
+        private ParallelWorkSet<string, string> _workSet;
+
+        [Fact]
+        public void GivenExceptionsOutsideWorker_CompletesAndThrowsException()
+        {
+            TestParallelWorkSet(new ParallelWorkSetTestCase
+            {
+                DegreeOfParallelism = 0,
+                WorkItemsToAdd = new List<WorkItem>
+                {
+                    new WorkItem
+                    {
+                        Key = "fooKey",
+                        WorkFunc = () => throw new Exception()
+                    },
+                    new WorkItem
+                    {
+                        Key = "barKey",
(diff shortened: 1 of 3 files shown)


src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
```diff
@@ -0,0 +1,253 @@
...
+        {
+            internal string Key { get; set; }
+
+            internal Func<string> WorkFunc { get; set; }
+        }
+
+        private ParallelWorkSet<string, string> _workSet;
+
+        [Fact]
+        public void GivenExceptionsOutsideWorker_CompletesAndThrowsException()
```
rainersigwald: What does `OutsideWorker` mean here? "Executed inline, rather than on a thread?"

src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
```diff
@@ -0,0 +1,253 @@
...
+        {
+            TestParallelWorkSet(new ParallelWorkSetTestCase());
+        }
+
+        [Fact]
+        public void GivenRecursiveWorkItemsAndMultipleWorkers_Completes()
+        {
+            TestParallelWorkSet(new ParallelWorkSetTestCase
+            {
+                DegreeOfParallelism = Environment.ProcessorCount,
```
rainersigwald: Should these be `Environment.ProcessorCount`, or hard-coded to some number to catch the scenario where the tests are run on a single-proc VM?

src/Build/Graph/ParallelWorkSet.cs
```diff
@@ -67,7 +61,11 @@ internal class ParallelWorkSet<TKey, TResult>
         /// <summary>
         /// Checks if the work set has been marked as completed.
         /// </summary>
-        internal bool IsCompleted => _isSchedulingCompleted;
+        internal bool IsCompleted
+        {
+            get => Volatile.Read(ref _isSchedulingCompleted);
+            set => Volatile.Write(ref _isSchedulingCompleted, value);
```
rainersigwald: Why is this public API surface? Is it meaningful for a caller to set it?
```suggestion
            private set => Volatile.Write(ref _isSchedulingCompleted, value);
```

src/Build/Graph/ParallelWorkSet.cs
```diff
@@ -115,51 +118,35 @@ internal void AddWork(TKey key, Func<TResult> workFunc)
...
-        internal void Complete()
+        internal void WaitForAllWorkAndComplete()
         {
-            bool schedulingCompleted = Volatile.Read(ref _isSchedulingCompleted);
-            if (schedulingCompleted)
+            if (IsCompleted)
             {
                 return;
             }
+            while (!_cancellationToken.IsCancellationRequested && Interlocked.Read(ref _pendingCount) > 0)
```
rainersigwald: If the last remaining work item is on another thread, won't this spin here really hard? Or is `_pendingCount` only for this thread?

ccastanedaucf: It will when there isn't another item in the queue (most likely start/end of the graph build), although it only effects the main/completion thread. I'm scoping this to another PR since it's a bit more involved to properly fix.

src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
ccastanedaucf: Maybe a hard-coded number if on a single-proc or below some threshold, else ```Environment.ProcessorCount```?

cdmihai: Sounds like it's time for a `[Theory]` :)

src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
```diff
@@ -0,0 +1,253 @@
...
+                WorkItemsToAdd = new List<WorkItem>
+                {
+                    new WorkItem
```
cdmihai: Should the work items be generated programmatically such that the nesting level is higher than the DegreeOfParallelism, to catch any potential starvation issues?

src/Build/Graph/ParallelWorkSet.cs
```diff
@@ -97,6 +95,11 @@ internal ParallelWorkSet(int degreeOfParallelism, IEqualityComparer<TKey> compar
         /// <param name="workFunc"></param>
         internal void AddWork(TKey key, Func<TResult> workFunc)
         {
+            if (IsCompleted)
+            {
+                throw new InvalidOperationException("Cannot add new work after work set is marked as completed.");
```
cdmihai: nit: test for this?

src/Build/Graph/ParallelWorkSet.cs
cdmihai: Rather then spinning on a `_pendingCount`, would spinning on a `remainingUnassignedWorkCount` fix it?

cdmihai: But then you'd still need to block if all remaining work has been assigned, to wait for it all to finish.

ccastanedaucf: ```remainingUnassignedWorkCount``` is essentially ```_queue.Count```. Spinning on that breaks on the scenario where a worker dequeues the last work item in the queue but that work item recursively enqueues more work at the end, then the completion thread exits the loop too early.

src/Build.UnitTests/Graph/ParallelWorkSet_Tests.cs
ccastanedaucf: Clarified this with ```WorkerThread``` and ```CompletionThread```

rainersigwald:
After addressing the feedback, please add ask mode template information. You can use microsoft/msbuild#4598 as an example.

rainersigwald:
@cdmihai can you coordinate with @ccastanedaucf about how this intersects with your recent change?

PR #4514, opened 2019-07-15
Readerwriterlock2

rainersigwald:


src/Build/Collections/PropertyDictionary.cs
```diff
@@ -497,17 +587,31 @@ internal void RemoveProperties(IEnumerable<T> other)
...
+                _lock.ExitReadLock();
+            }
 
             return dictionary;
         }
+
+        /// <summary>
+        /// Finalizer to clean up the <see cref="ReaderWriterLockSlim" />.
+        /// </summary>
+        ~PropertyDictionary()
```
drewnoakes: This finalizer doesn't add anything. `ReaderWriterLockSlim` is finalizable itself, so will be queued for finalization separately. Making `PropertyDictionary` finalizable too is overkill.

Is it possible to make `PropertyDictionary` disposable to clean up the lock and suppress its finalization?

drewnoakes: Correction: `ReaderWriterLockSlim` is not finalizable, but the wait handles it contains are (i.e. it uses the same pattern I think should be used here).

u7pro: `ReaderWriterLockSlim` is not declared as finalizable as it doesn't held any unmanaged ressources.

`PropertyDictionary` doesn't handle any unmanaged resources and should also not be finalizable.

calling `_lock.Dispose()` in this finalizer is useless as the `_lock` instance is not shared and thus its lifecycle will be as short as its parent `PropertyDictionary`.

Instead, `PropertyDictionary` should implement `IDisposable` interface to invoke `_lock.Dispose()` to give callers opportunity to free memory predictivaly instead of relying on GC events

PR #4517, opened 2019-07-16
Move official build to dev16 machines

rainersigwald:
Noticed this when investigating an official build failure that appears to have been one-off infra flakiness.


rainersigwald:
The build failed, not worth trying to figure this out for now.

PR #4528, opened 2019-07-19
Add support for optional loggers

AndyGerlicher:
When specified, do not fail the build for optional loggers.

Current behavior:
```
d:\src\MSBuildPrediction>d:\src\msbuild.fork6\artifacts\bin\bootstrap\net472\MSBuild\Current\Bin\msbuild.exe /logger:EmptyLogger2.dll /t:Restore
Microsoft (R) Build Engine version 16.3.0-dev-19369-01+79f9c820f for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1021: Cannot create an instance of the logger.
Switch: EmptyLogger2.dll
```

Optional logger behavior:
```
d:\src\MSBuildPrediction>d:\src\msbuild.fork6\artifacts\bin\bootstrap\net472\MSBuild\Current\Bin\msbuild.exe /optionallogger:EmptyLogger2.dll /t:Restore /v:diag
Microsoft (R) Build Engine version 16.3.0-dev-19369-01+79f9c820f for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : Cannot create an instance of the optional logger. Could not load file or assembly 'EmptyLogger2.dll' or one of its dependencies. The system cannot find the file specified.
... And the rest of the build here ...
```

First draft of the proposed change I mentioned previously to get feedback. Let me know if the approach looks ok.

For background, this change will allow for loggers to be specified (say in a `Directory.Build.rsp`) that are optional. We plan to use this internally to collect build telemetry (for example how long it takes to build, was the build incremental, and scope of build in a repo). We can distributed the logger via other mechanisms, but we don't want to fail the build if it hasn't been installed or deleted.

PR #4529, opened 2019-07-19
Add support to ignore rsp lines in older MSBuild

AndyGerlicher:
When adding new command-line switches to MSBuild (e.g. /restore, /optionalLogger, etc.), use of the feature will break all old versions. In general this is OK, but one might want to specify new defaults in a `Directory.Build.rsp` file to give the best experience to users of the latest. Logger verbosity is a good example as "normal" verbosity is not a great default, but it's hard to change for the product.

In my case, I would like to add logging to collect information on specific repos inside Microsoft that opt-in. To do this, the recommended approach is to add a logger. Unfortunately, this is fairly difficult to setup. I believe I've solved the distribution issues in  (#4499 and #4528), but using these features will break all users not on a sufficiently high version. This makes flighting changes to preview users not practical, and a potential lag as all users upgrade.

My proposal, since this is a somewhat corner case, is to use the `#` character to ignore rsp file lines in older versions of MSBuild and version the ignore going forward. For example:
```bash
-ConsoleLoggerParameters:Verbosity=Minimal;Summary;ForceNoAlign
-MaxCPUCount
-NodeReuse:false
-Restore
# Only attach loggers in 16.300+
#v16.300 -optionallogger:%MSBuildThisFileDirectory%\build\MyLogger.dll
# Use -graph build on 16.400+
#v16.400 -graph -isolate
```

This syntax would scope that line to be processed only in MSBuild 16.300 and above and the 2nd to 16.400 and above. As a note I don't particularly care for the syntax, but I can't think of a better way without introducing a new file or concept which is a significant amount of work and needs to be carefully thought through. I created a [gist](https://gist.github.com/AndyGerlicher/07fc5b28e5fbfa1e2198b449ccb57e72) with an idea, but I don't think that's worth the cost given this is not a common ask (although I have been asked by a large customer previously for a way to attach a logger to all builds).

Thoughts?

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1787,8 +1787,8 @@ namespace Microsoft.Build.CommandLine
                                 // ignore leading whitespace on each line
                                 string responseFileLine = responseFileContents.ReadLine().TrimStart();
 
-                                // skip comment lines beginning with #
-                                if (!responseFileLine.StartsWith("#", StringComparison.Ordinal))
+                                // Check to see if we should skip the line
+                                if (!SkipResponseFileLine(ref responseFileLine))
                                 {
                                     // Allow special case to support a path relative to the .rsp file being processed.
                                     responseFileLine = Regex.Replace(responseFileLine, responseFilePathReplacement,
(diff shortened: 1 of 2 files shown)


src/MSBuild/XMake.cs
```diff
@@ -1822,6 +1822,44 @@ private static void GatherResponseFileSwitch(string unquotedCommandLineArg, Comm
...
+            // If we don't have a version match, skip the file since it started with a #
+            if (!match.Success)
+            {
+                return true;
+            }
+
+            // We have a version in the comment, make sure it's less than or equal to the current version.
+            if (Version.TryParse(match.Groups[1].Value, out Version ver))
+            {
+                if (ver.Major <= ProjectCollection.Version.Major && ver.Minor <= ProjectCollection.Version.Minor)
```
danmosemsft: This seems wrong, it would reject 17.2 if the actual version is 18.0 because 2 <= 0

PR #4554, opened 2019-07-26
Override FileVersion to match AssemblyInformationalVersion

rainersigwald:
Fixes [AB#952777](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/952777).

This was the best fix I can think of right now. I first looked at just changing how MSBuildVersion was calculated, but that would leave the FileVersion with the unreadable Arcade versioning, and mean that you couldn't check `msbuild -version` and plug that number into a condition. I am also not confident that it would catch all the places where we expose FileVersion. So I opted to hack around Arcade's rules and directly specify FileVersion.


eng/Versions.props
```diff
@@ -32,4 +33,35 @@
...
   </PropertyGroup>
+
+  <Target Name="OverrideArcadeFileVersion"
+          AfterTargets="_InitializeAssemblyVersion">
+    <!-- MSBuild 16.2.0 didn't exist. It shipped as file version 16.200.19.32702,
+         informational version 16.2.32702+c4012a063 because of confusion around
+         Arcade versioning rules: https://github.com/dotnet/arcade/issues/3386
+
+         This caused MSBuild -version to report 16.200.19.32702 and
+         $(MSBuildVersion) to report 16.200.19, a confusing and error-prone
```
tmat: Couldn't `$(MSBuildVersion)` be changed to not use `AssemblyFileVersion` but instead some other custom attribute that you generate and completely control?

rainersigwald: As I mentioned in the comment and the PR message: yes, but what else uses file version? How can we be sure we caught it all? And why wouldn't we just have a file version that isn't munged?

tmat: The file version is compressed to fit all the information necessary to provide monotonically increasing version for Windows Installer, considering all constraints this system has. I would not use it for anything else. Other version info is much better suited for user visible version info. For example, Roslyn uses the `AssemblyInformationalVersion` value.

BTW, we shifted the minor version by 100 to accommodate for .NET SDK patch numbers greater than 100.

@nguerrera

tmat: >  yes, but what else uses file version? How can we be sure we caught it all? 

I don't see how we can guarantee anything about the exact numbers in file version. If someone relies on exact file version values then they can potentially be broken by any change to the number.

rainersigwald: I do not understand your pushback. There was a major change in the format of FileVersion as a result of the move to Arcade. That change broke assumptions in our legacy codebase. Having seen it do so already, I am concerned about the existence of further assumptions in other parts of the codebase that haven't broken loudly yet, so my preferred targeted solution is to change the FileVersion back to its former semantics, since I can't see any downside to doing so.

tmat: Up to you.

nguerrera: I hate to have to ask this, but is there any risk that there are msis out there that would run into upgrade problems from this?

I'm supportive of @rainersigwald's change.

It's really hard to satisfy all the constraints of legacy code and limited bits in format and rules outside my control like SDK must use runtime major and minor and pack the real minor and patch into the patch, etc.

rainersigwald: Re: MSIs that deliver MSBuild:

Visual Studio does not get MSBuild through MSIs; it's all vsix-delivered through Willow setup.

.NET Core SDK _does_ deliver MSBuild through an MSI, but it doesn't overwrite MSBuild in place, because it's in the (versioned) sdk folder, like `C:\Program Files\dotnet\sdk\3.0.100-preview8-013437\MSBuild.dll`, so we should be ok on that front too.

rainersigwald:
This looks good in a test upgrade install over the public release of 16.2.0. If anyone has a specific scenario they'd like to test, let me know.

I'm going to merge this and prep for insertion.

PR #4569, opened 2019-07-31
Bringing Spans Into ExtractPropertyFunction

BenVillalobos:
Using spans in ExtractPropertyFunction helps avoid one unnecessary substring call. This saves a whopping ~0.2% allocations. A small win but a win nonetheless.

proof_expander-spans

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -10,6 +10,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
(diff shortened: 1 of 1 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -3142,15 +3143,15 @@ internal string Receiver
                 // A static method is the content that follows the last "::", the rest being the type
                 if (propertyValue == null && expressionRoot[0] == '[')
                 {
-                    var typeEndIndex = expressionRoot.IndexOf(']', 1);
+                    var typeEndIndex = expressionRoot.IndexOf(']');
```
danmosemsft: Missing overload on Span?

Therzok: We have this extension method in [MonoDevelop](http://source.monodevelop.com/#MonoDevelop.Core/MemoryExtensions.cs,949db08fee756f03)

BenVillalobos: The original code I wrote here was `var typeEndIndex = expressionRoot.Slice(1).IndexOf(']');` (which is also what the MonoDevelop source code would do).

I opted for the non-sliced version assuming that letting IndexOf iterate over index 0 is more efficient than performing a slice with a startIndex of 1 each time.

PR #4598, opened 2019-08-08
Fix warning when newline in hintpath

wli3:
### Description

If there are newlines or white spaces passed to `Path.GetFullpath`, it will throw on fullframework. [Adding `Path.GetFullpath`](https://github.com/microsoft/msbuild/pull/4414/files) in previous change caused extra warning from Resolve Assembly Reference. This change added "skip when the path is not valid" to being back the "no warning but skip" behavior of 16.1.

### Customer Impact

If the customer has _warning as error_ enabled, a build would be successful in 16.1 but would be fail in the latest version

### Regression?

Yes. by https://github.com/microsoft/msbuild/pull/4414, since 16.2

### Risk

Low risk.

### Test changes in this PR

Additional unit tests and manual end to end tests.

------------------------------

fix https://github.com/microsoft/msbuild/issues/4593

This is a strict fix for the warning problem. Please look at `CannotNormalizePathWithNewLineAndSpace` test. `FileUtilities.NormalizePath` will get garbage output like (`c:\work\test   \r\n c:\work\referenced\file.dll`) when called with newline. So `File.Exists` cannot find the correct file anyway. The fix is only to make it quiet.

It behaves differently in core because Path.GetFullpath no longer throws when there is invalid char
To be discuss

1. Should we make it work? Instead of skip when contains invalid char, should we Trim it instead?
2. Is master the right branch? Do I need ask mode template for 16.3?

### DIFF
--- a/src/Tasks/AssemblyDependency/HintPathResolver.cs
+++ b/src/Tasks/AssemblyDependency/HintPathResolver.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Build.Tasks
             out bool userRequestedSpecificFile
         )
         {
-            if (!string.IsNullOrEmpty(hintPath))
+            if (!string.IsNullOrEmpty(hintPath) && !FileUtilities.PathIsInvalid(hintPath))
             {
                 if (ResolveAsFile(FileUtilities.NormalizePath(hintPath), assemblyName, isPrimaryProjectReference, wantSpecificVersion, true, assembliesConsideredAndRejected))
                 {
--- a/src/Tasks/AssemblyDependency/HintPathResolver.cs
+++ b/src/Tasks/AssemblyDependency/HintPathResolver.cs
@@ -51,6 +51,10 @@ namespace Microsoft.Build.Tasks
             out bool userRequestedSpecificFile
         )
         {
+            // If there is newline or white space `FileUtilities.NormalizePath` will get garbage result(throw on fullframework).
+            // Adding FileUtilities.NormalizePath (https://github.com/microsoft/msbuild/pull/4414) caused https://github.com/microsoft/msbuild/issues/4593
+            // It is fixed by adding skip when the path is not valid
+            // However, we should consider Trim() the hintpath https://github.com/microsoft/msbuild/issues/4603
             if (!string.IsNullOrEmpty(hintPath) && !FileUtilities.PathIsInvalid(hintPath))
             {
                 if (ResolveAsFile(FileUtilities.NormalizePath(hintPath), assemblyName, isPrimaryProjectReference, wantSpecificVersion, true, assembliesConsideredAndRejected))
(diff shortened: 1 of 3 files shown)


nguerrera:
I don't understand how the repro project I saw in 16.1 would have been quiet, but not resolve to the path. I would have expected a build warning for unresolved reference if the path was not interpreted as trimmed in 2.1. Did it resolve from another resolver than HintPath?

nguerrera:
I definitely don't want to have no warning, but also ignore the hint path. I'm assuming either warn and ignore or don't warn and trim.

nguerrera:
Do you have a binlog, or can you make a binlog for 16.1 building this: https://github.com/microsoft/msbuild/issues/4593#issuecomment-519292872

rainersigwald:
I agree: I'd like to know exactly what caused the 16.1 behavior before we decide how to fix this.

nguerrera:
Oh, I see the difference. The reference ItemSpec is to System.Net.Http so it resolves to target framework and hint path with new lines is treated as just not found which is not a warning if another resolver finds it.

nguerrera:
So this goes back to what I said originally: we should not trim in sdk conflict resolution if rar does not. What a mess.

wli3:
So which way we should we choose?
a. trim RAR hint path so it can find the actual hint path (and it should, some xml formatter will add the new line)
b. keep the change like this, and stop trimming in conflict resolution and quietly skip when there is one

I prefer a. sounds closer to what customer want

rainersigwald:
a is better behavior, but I think since it's late in 16.3, we should just revert to the 16.1 behavior. Then there's no risk of breaking people with incorrect hint paths that weren't being respected because of whitespace.

wli3:
So take this change as is?

I will later change the sdk part.

wli3:
Also do i need ask mode? 16.3 is master?

rainersigwald:
`master` is still 16.3 for the moment, yeah.

src/Tasks/AssemblyDependency/HintPathResolver.cs
```diff
@@ -51,7 +51,7 @@ public override bool Resolve
             out bool userRequestedSpecificFile
         )
         {
-            if (!string.IsNullOrEmpty(hintPath))
+            if (!string.IsNullOrEmpty(hintPath) && !FileUtilities.PathIsInvalid(hintPath))
```
rainersigwald: I think this deserves a comment. Probably with a link to the bug to consider trimming HintPath in 17.0.

wli3: Added comment, https://github.com/microsoft/msbuild/pull/4598/commits/fc762c4b478fe94d329bf15b907da88e8e2dfb92

and I do need to have ask mode write up

wli3:
get approved from email. merging

PR #4599, opened 2019-08-09
[Static Graph] ProjectGraph.GetTargetLists throws on invalid target names

cdmihai:
Just cleanup: clarifying some documentation, validating some inputs, avoiding needless iterations.

### DIFF
--- a/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
+++ b/src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
@@ -804,6 +804,42 @@ namespace Microsoft.Build.Experimental.Graph.UnitTests
             }
         }
 
+        [Fact]
+        public void GetTargetListsDoesNotPropagateEmptyTargets()
+        {
+            using (var env = TestEnvironment.Create())
+            {
+                // Target protocol produces empty target
+                TransientTestFile entryProject = CreateProjectFile(env: env, projectNumber: 1, projectReferences: new[] { 2 }, projectReferenceTargets: new Dictionary<string, string[]> { { "A", new[] { " ; ; " } }}, defaultTargets: string.Empty);
+
+                // Dependency has default targets. Even though it gets called with empty targets, B will not get called,
+                // because target propagation only equates empty targets to default targets for the root nodes.
+                CreateProjectFile(env: env, projectNumber: 2, defaultTargets: "B");
+
+                var projectGraph = new ProjectGraph(entryProject.Path);
+                projectGraph.ProjectNodes.Count.ShouldBe(2);
+
+                IReadOnlyDictionary<ProjectGraphNode, ImmutableList<string>> targetLists = projectGraph.GetTargetLists(new []{ "A" });
+                targetLists.Count.ShouldBe(projectGraph.ProjectNodes.Count);
+                targetLists[GetFirstNodeWithProjectNumber(projectGraph, 1)].ShouldBe(new []{ "A" });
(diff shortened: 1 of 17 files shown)


src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -764,6 +764,47 @@ public void GetTargetsListsShouldApplyDefaultTargetsOnlyToGraphRoots()
...
+
+                IReadOnlyDictionary<ProjectGraphNode, ImmutableList<string>> targetLists = projectGraph.GetTargetLists(new []{ "Foo" });
+                targetLists.Count.ShouldBe(projectGraph.ProjectNodes.Count);
+                targetLists[GetFirstNodeWithProjectNumber(projectGraph, 1)].ShouldBe(new []{ "Foo" });
+                targetLists[GetFirstNodeWithProjectNumber(projectGraph, 2)].ShouldBeEmpty();
+            }
+        }
+
+        [Fact]
+        public void GetTargetListsReturnsEmptyTargetsForAllNodesWhenDefaultTargetsAreRequestedAndThereAreNoDefaultTargets()
```
rainersigwald: Under what circumstance are there no default targets? Doesn't MSBuild always have a default target (if none is explicitly listed, it's the first one defined in the logical project)?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -599,6 +606,19 @@ internal string ToDot(Func<ProjectGraphNode, string> nodeIdProvider)
...
+            void ThrowOnEmptyTargetNames(ICollection<string> targetNames)
+            {
+                if (targetNames == null || targetNames.Count == 0)
+                {
+                    return;
+                }
+
+                if (targetNames.Any(targetName => string.IsNullOrWhiteSpace(targetName)))
+                {
+                    throw new ArgumentException(ResourceUtilities.FormatResourceStringIgnoreCodeAndKeyword("OM_TargetNameNullOrEmpty"));
```
rainersigwald: Would it be possible to have more context here? Right now it just says there's a null/empty target name somewhere. How do I know where?

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
cdmihai: The project file has no targets. Added a comment on it.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -597,6 +606,19 @@ internal string ToDot(Func<ProjectGraphNode, string> nodeIdProvider)
...
+            void ThrowOnEmptyTargetNames(ICollection<string> targetNames)
+            {
+                if (targetNames == null || targetNames.Count == 0)
+                {
+                    return;
+                }
+
+                if (targetNames.Any(targetName => string.IsNullOrWhiteSpace(targetName)))
+                {
+                    throw new ArgumentException(ResourceUtilities.FormatResourceStringIgnoreCodeAndKeyword("OM_TargetNameNullOrEmpty", nameof(GetTargetLists)));
```
rainersigwald: Is this enough context? I was thinking more along the lines of "what project caused this error?"

cdmihai: Ah, that's a bit harder to answer. There is no actual project here. These targets are the graph wide targets the whole graph needs to be built with. The closest thing for identification would be the set of graph roots

PR #4602, opened 2019-08-09
Linguist tweaks

rainersigwald:
Just happened across documentation that enabled me to do some things I've wanted for a while:

* Include MSBuild core props/targets in repo stats (it's a bunch of our code!)
* Exclude Arcade code from repo stats (we don't use CMake!)
* Mark XLF files as generated (collapse them by default in PRs)

Current stats:

image


rainersigwald:
Yay, it processed. Now we can CLAIM our 1.7% XML.

image

PR #4605, opened 2019-08-10

pi1024e:


danmosemsft:
Why this particular value for MaxBuilderSize? It is a tradeoff. Do you have performance measurements to support it?

rainersigwald:
@pi1024e This has the same contents as #4557, but doesn't address the feedback given there.

We are not willing to take a change to `MaxBuilderSize` without detailed justification. Do you have numbers to back up your 2048 value?

Likewise, many of the other changes in this PR seem unlikely to impact memory consumption or even performance. Do you have benchmarks proving that they do?

I am closing this PR, but if you can prove that aspects of it are performance improvements, we'd be happy to reopen it.

PR #4612, opened 2019-08-13
Add new logger options functionality

jeffkl:
### Description
Loggers are required to work otherwise a build fails.  We want to distributed loggers in NuGet packages which means that the very first build would not be able to restore since loggers are added with a command-line option.

The `-logger` argument now accepts options after the class and assembly.  One new option is to make a logger optional meaning a message will be printed if the logger can't be created rather than failing the build.

Will supersede #4528  

### Customer Impact

We need to deploy a logger via NuGet that collects telemetry for improving the inner loop developer experience.

### Regression?

No

### Risk

Low risk.

### Test changes in this PR

New unit tests

### DIFF
--- a/src/MSBuild.UnitTests/XMake_Tests.cs
+++ b/src/MSBuild.UnitTests/XMake_Tests.cs
@@ -749,8 +749,6 @@ namespace Microsoft.Build.UnitTests
                 //Should pass
                 MSBuildApp.Execute(@"c:\bin\msbuild.exe /logger:FileLogger,""Microsoft.Build, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"" " + quotedProjectFileName).ShouldBe(MSBuildApp.ExitType.Success);
 
-                //Should fail as we are not changing existing lines
-                MSBuildApp.Execute(@"c:\bin\msbuild.exe /logger:FileLogger,Microsoft.Build,Version=11111 " + quotedProjectFileName).ShouldBe(MSBuildApp.ExitType.InitializationError);
 #else
                 //Should pass
                 MSBuildApp.Execute(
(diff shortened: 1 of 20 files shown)


jeffkl:
I still need some review on this PR!

src/MSBuild/Resources/Strings.resx
```diff
@@ -252,11 +252,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <value>  -logger:&lt;logger&gt;   Use this logger to log events from MSBuild. To specify
                      multiple loggers, specify each logger separately.
                      The &lt;logger&gt; syntax is:
-                       [&lt;logger class&gt;,]&lt;logger assembly&gt;[;&lt;logger parameters&gt;]
+                       [&lt;class&gt;,]&lt;assembly&gt;[,&gt;options][;&lt;parameters&gt;]
```
rainersigwald:
```suggestion
                       [&lt;class&gt;,]&lt;assembly&gt;[,&lt;options&gt;][;&lt;parameters&gt;]
```
?

src/MSBuild/Resources/Strings.resx
```diff
@@ -406,11 +407,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
                      multiple loggers, specify each logger separately.
                      (Short form -dl)
                      The &lt;logger&gt; syntax is:
-                       [&lt;logger class&gt;,]&lt;logger assembly&gt;[;&lt;logger parameters&gt;]
+                       [&lt;class&gt;,]&lt;assembly&gt;[,options][;&lt;parameters&gt;]
```
rainersigwald:
```suggestion
                       [&lt;class&gt;,]&lt;assembly&gt;[,&lt;options&gt;][;&lt;parameters&gt;]
```

src/MSBuild/XMake.cs
```diff
@@ -3410,52 +3422,58 @@ private static LoggerDescription ParseLoggingParameter(string parameter, string
...
-            return new LoggerDescription(loggerClassName, loggerAssemblyName, loggerAssemblyFile, loggerParameters, verbosity);
+            return new LoggerDescription(loggerClassName, loggerAssemblyName, loggerAssemblyFile, loggerParameters, verbosity, isOptional);
         }
 
         /// <summary>
         /// Loads a logger from its assembly, instantiates it, and handles errors.
         /// </summary>
         /// <returns>Instantiated logger.</returns>
-        private static ILogger CreateAndConfigureLogger
+        private static bool TryCreateAndConfigureLogger
```
rainersigwald: I don't like the name `TryCreate` here when it will throw in a not-very-edge case (non-optional logger doesn't exist). I think just `CreateAndConfigureLogger`, but change the return type/out param.

src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -2042,6 +2032,33 @@ public void MultipleTargetsDoesNotCrash()
             logContents.ShouldContain("E2C73B5843F94B63B067D9BEB2C4EC52", () => logContents);
         }
 
+        [Theory]
+        [InlineData("-logger:,\"nonExistentlogger.dll\",IsOptional;Foo")]
+        [InlineData("-logger:ClassA,\"nonExistentlogger.dll\",IsOptional;Foo")]
+        [InlineData("-logger:,\"nonExistentlogger.dll\",IsOptional,OptionB,OptionC")]
+        [InlineData("-distributedlogger:,\"nonExistentlogger.dll\",IsOptional;Foo")]
+        [InlineData("-distributedlogger:ClassA,\"nonExistentlogger.dll\",IsOptional;Foo")]
+        [InlineData("-distributedlogger:,\"nonExistentlogger.dll\",IsOptional,OptionB,OptionC")]
```
rainersigwald: So the initial comma is required when specifying `IsOptional`, even if you're just passing an assembly? That seems ok, but worth calling out . . . somewhere. Not the CLI help since we're keeping that short.

PR #4617, opened 2019-08-14
Update System.Threading.Tasks.Dataflow to 4.9.0

AArnott:
#### Description

Consolidate versions used between VS and .NET Core.

#### Customer Impact

This avoids an RPS regression in VS 16.3 which is about to switch to Dataflow 4.9.0

#### Regression

No.

#### Risk

Low.

### DIFF
--- a/src/nuget/Microsoft.Build.nuspec
+++ b/src/nuget/Microsoft.Build.nuspec
@@ -58,7 +58,7 @@
         <dependency id="System.Text.RegularExpressions" version="4.1.0" />
         <dependency id="System.Threading" version="4.0.11" />
         <dependency id="System.Threading.Tasks" version="4.0.11" />
-        <dependency id="System.Threading.Tasks.Dataflow" version="4.6.0" />
+        <dependency id="System.Threading.Tasks.Dataflow" version="4.9.0" />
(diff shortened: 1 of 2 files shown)


src/nuget/Microsoft.Build.nuspec
```diff
@@ -58,7 +58,7 @@
         <dependency id="System.Text.RegularExpressions" version="4.1.0" />
         <dependency id="System.Threading" version="4.0.11" />
         <dependency id="System.Threading.Tasks" version="4.0.11" />
-        <dependency id="System.Threading.Tasks.Dataflow" version="4.6.0" />
+        <dependency id="System.Threading.Tasks.Dataflow" version="4.9.0" />
```
rainersigwald: This is harmless but I believe this file is dead; microsoft/msbuild#4618 to figure out if that's true.

rainersigwald:
Tactics approved via email. I'm going to merge this into `vs16.3` so it can be inserted to VS in isolation.

@AArnott, please be aware that we won't be able to insert this immediately; we just unblocked [the current insertion](https://dev.azure.com/devdiv/DevDiv/MSBuild/_git/VS/pullrequest/196856). I'll start an insertion with this commit as soon as that's completed.

PR #4620, opened 2019-08-15
Fix cases where UTF8 wasn't supported.

yuehuang010:
- Fix an issue where .sln files are created without BOM fallback to asci.  This limits to just EN language. 
 Instead, fallback to UTF-8, which is a superset of asci.  The VS IDE by default creates sln in UFT8.
(UNDO)- When UseCommandProcessor is true, it will create a .cmd (aka .bat) file to run.  This file encoding is hardcoded for OS lang which is sometimes incorrect, but it doesn't provide a way to override.  Using StandardOutputEncoding as a override mechanism.

### DIFF
--- a/src/Build/Construction/Solution/SolutionFile.cs
+++ b/src/Build/Construction/Solution/SolutionFile.cs
@@ -265,7 +265,7 @@ namespace Microsoft.Build.Construction
             {
                 // Open the file
                 fileStream = File.OpenRead(solutionFile);
-                reader = new StreamReader(fileStream, Encoding.UTF8); // HIGHCHAR: If solution files have no byte-order marks, then assume UFT8-8.
+                reader = new StreamReader(fileStream, Encoding.UTF8); // HIGHCHAR: If solution files have no byte-order marks, then assume UFT-8.
 
                 // Read first 4 lines of the solution file. 
                 // The header is expected to be in line 1 or 2
--- a/src/Build/Construction/Solution/SolutionFile.cs
+++ b/src/Build/Construction/Solution/SolutionFile.cs
@@ -265,7 +265,7 @@ namespace Microsoft.Build.Construction
             {
                 // Open the file
                 fileStream = File.OpenRead(solutionFile);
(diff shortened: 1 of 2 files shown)


src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -265,7 +265,7 @@ internal static void GetSolutionFileAndVisualStudioMajorVersions(string solution
             {
                 // Open the file
                 fileStream = File.OpenRead(solutionFile);
-                reader = new StreamReader(fileStream, Encoding.GetEncoding(0)); // HIGHCHAR: If solution files have no byte-order marks, then assume ANSI rather than ASCII.
+                reader = new StreamReader(fileStream, Encoding.UTF8); // HIGHCHAR: If solution files have no byte-order marks, then assume UFT8-8.
```
jgoshi: nit UTF8 misspelled in the comment

yuehuang010: Thanks. fixed.

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -265,7 +265,7 @@ internal static void GetSolutionFileAndVisualStudioMajorVersions(string solution
             {
                 // Open the file
                 fileStream = File.OpenRead(solutionFile);
-                reader = new StreamReader(fileStream, Encoding.GetEncoding(0)); // HIGHCHAR: If solution files have no byte-order marks, then assume ANSI rather than ASCII.
+                reader = new StreamReader(fileStream, Encoding.UTF8); // HIGHCHAR: If solution files have no byte-order marks, then assume UFT-8.
```
```suggestion
                reader = new StreamReader(fileStream, Encoding.UTF8); // HIGHCHAR: If solution files have no byte-order marks, then assume UTF-8 (which also supports ASCII).
```

rainersigwald:
Looks likehttps://github.com/microsoft/msbuild/blob/master/src/Build.UnitTests/Construction/SolutionFile_Tests.cs would be a good place for such a test.

yuehuang010:
I have examined further regarding code page and how the VS IDE handles ambiguity.  A simple UFT8 isn't enough, sadly.  I am abandoning this PR.

PR #4624, opened 2019-08-17
Issue #4199 says a large portion of memory used when creating a snapshot in the Project Instance class is the creation of Sync Objects, so I have optimized this portion to be more efficient. #4605

pi1024e:
I increased the MaxBuilder size from 1024 to 2048.

I took benchmarks this time: Whereas the current version of MSBuild performed all the tests in 21 minutes and 2 seconds, this build performed the tests in 20 minutes, 3 seconds.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -983,7 +983,7 @@ namespace Microsoft.Build.Evaluation
 
                 // Search for "$(" in the expression.  Loop until we don't find it 
                 // any more.
-                while (propertyStartIndex != -1)
+                do
                 {
                     if (lastResult != null)
                     {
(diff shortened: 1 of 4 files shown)


rainersigwald:
Team triage: we were hoping for much more detailed performance analysis, including details on memory usage and reuse efficiency in various scenarios. Since that's difficult to collect, let's leave the value alone until we have time to really dive into it (and use internal Visual Studio performance infrastructure to help with the analysis in real-world scenarios).

pi1024e:
How does one go about collecting this info? Because all I know is that it has done improvements from what I have noticed.

BenVillalobos:
@pi1024e [PerfView](https://github.com/microsoft/perfview) is one of the tools we use to verify reductions in memory allocation. [Here ](https://channel9.msdn.com/Series/PerfView-Tutorial)is a video series showing how to use it.

PR #4627, opened 2019-08-17
Fix IDE0008 (use explicit type instead of var) to modernize the codebase

pi1024e:
Every change was done with the IDE. This was done to make it clear what the variable's roles are, as they are used as arguments to functions that only take a certain type.


rainersigwald:
As with #4628, can you please be very clear about exactly what you did to produce these changes? In addition, when there are multiple types of automated refactoring we would like them to each be in their own commit, so that we can more easily review the large-scale changes.

pi1024e:
This was all the same type of automatic refactoring. Every one of these fixes were under IDE0008.

pi1024e:
To produce these changes, Visual Studio analyzed the solution and suggested the changes under IDE0008, and I applied its suggestions according to the code style settings in the project.

rainersigwald:

PR #4644, opened 2019-08-21
Improve incremental behavior on MSB3822

### LINKED ISSUE #4553: Binary resource error MSB3822 incorrectly skips subsequent incremental build
Project for test: https://github.com/deadman2000/BinaryResourceTest/

### Steps to reproduce

* Clean project
* Remove `obj` folders
* Build with `dotnet build`
* Getting error `MSB3822: Non-string resources require the System.Resources.Extensions assembly at runtime, but it was not found in this project's references`
* Build again with `dotnet build`
* Project builded. BUT `ResourceManager.GetObject` returns `null`

Full log: https://raw.githubusercontent.com/deadman2000/BinaryResourceTest/master/test.txt

### Environment data

.NET Core SDK
 Version:   3.0.100-preview7-012821
 Commit:    6348f1068a

rainersigwald:
## Description

Fixes #4553, an issue where a build would fail with a (correctly emitted) MSB3822 error, but subsequent builds would succeed, producing corrupted outputs without all of the resources embedded.

## Customer Impact

Fixes an error that would cause builds to fail and pass apparently randomly, while silently producing bad output when the build passed.

## Regression

Bug in new functionality introduced by .NET Core resx support for non-string types. Desktop MSBuild building desktop projects did not regress.

## Risk

Low. Impacts only the new resx codepath and uses an existing delete-or-warn function.

### DIFF
--- a/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
+++ b/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
@@ -449,6 +449,8 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
             {
                 t.Sources = new ITaskItem[] { new TaskItem(resxFile) };
 
+                string outputResource = Path.ChangeExtension(Path.GetFullPath(resxFile), ".resources");
+
 #if NETFRAMEWORK
                 if (!usePreserialized)
                 {
@@ -477,9 +479,10 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
                 else
                 {
                     Utilities.AssertLogContainsResource(t, "GenerateResource.PreserializedResourcesRequiresExtensions");
+                    Utilities.AssertLogContainsResource(t, "GenerateResource.CorruptOutput", outputResource);
                 }
 
-                File.Exists(Path.ChangeExtension(Path.GetFullPath(resxFile), ".resources"))
+                File.Exists(outputResource)
                     .ShouldBeFalse("Resources file was left on disk even though resource creation failed.");
             }
             finally
--- a/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
+++ b/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
@@ -478,6 +478,9 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
                 {
                     Utilities.AssertLogContainsResource(t, "GenerateResource.PreserializedResourcesRequiresExtensions");
                 }
+
+                File.Exists(Path.ChangeExtension(Path.GetFullPath(resxFile), ".resources"))
+                    .ShouldBeFalse("Resources file was left on disk even though resource creation failed.");
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
```diff
@@ -477,7 +479,11 @@ public void ForceOutOfDateLinked(bool usePreserialized)
                 else
                 {
                     Utilities.AssertLogContainsResource(t, "GenerateResource.PreserializedResourcesRequiresExtensions");
+                    Utilities.AssertLogContainsResource(t, "GenerateResource.CorruptOutput", outputResource);
```
nguerrera: Why does this output if we successfully delete?

rainersigwald: For example:

```
        ERROR : MSB3822: Non-string resources require the System.Resources.Extensions assembly at runtime, but it was not found in this project's references.
        WARNING MSB3555: Output file "C:\Users\raines\AppData\Local\Temp\rskgseqm.ymd\tmp363846e55b234157949d3501a7909105.resources" is possibly corrupt.
```

I don't love the idea of emitting a warning in this case in addition to the error. But I like using the existing method that does almost exactly what I want here.

nguerrera: Now I have more questions. If the corrupt file is in temp, how does it break incremental build? We go ahead and move it to final location after failure?

nguerrera: Or that's just the actual path/name for the test?

rainersigwald: Correct, that's verbatim from a test run of mine, which generates a `.resx` in temp and the `.resources` next to it. In a real build, it'd be in `obj` in a predictable location, and GenerateResources knows enough to predict that location and check its timestamp, then skip.

rainersigwald: Here's a real error situation:

```
S:\msbuild\artifacts\bin\bootstrap\net472\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(3056,5): error : MSB3822: Non-string resources require the System.Resources.Extensions assembly at runtime, but it was not found in this project's references. [S:\repro\microsoft\msbuild\issues\4553\BinaryResourceTest\BinaryResourceTest.csproj]
S:\msbuild\artifacts\bin\bootstrap\net472\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(3056,5): warning MSB3555: Output file "S:\repro\microsoft\msbuild\issues\4553\BinaryResourceTest\obj\Debug\net40\BinaryResourceTest.Resource1.resources" is possibly corrupt. [S:\repro\microsoft\msbuild\issues\4553\BinaryResourceTest\BinaryResourceTest.csproj]
```

nguerrera: The warning is a bit spammy given that it will be deleted and if we failed to delete there will be more output. But I can live with this to keep the fix scoped at this late stage of the release.

rainersigwald: My thoughts exactly. Filed #4648.

PR #4655, opened 2019-08-23

pi1024e:
I increased the MaxBuilderSize from 1024 to 2048 with the following benchmarks taken with dotnet Memory, and profiling it while compiling the open source version of osu!, I got the following results.

Cross snapshots comparison
(A) Snapshot from MSBuild with 2048 size taken 8/23/2019 5:27:34 PM
(B) Snapshot #6 from MSBuild with 1024 size taken 8/23/2019 5:41:31 PM

Type: System.Text.StringBuilder
Objects (A): 34
Objects (B): 35
Bytes (A): 1,632
Bytes (B): 1,680
Objects delta: 1
Bytes delta: 48

Additionally, at any one time, the MSBuild with 2048 size had less memory usage than the 1024 memory size by a couple megabytes.

### DIFF
--- a/src/Build/Definition/ProjectItem.cs
+++ b/src/Build/Definition/ProjectItem.cs
@@ -1077,13 +1077,14 @@ namespace Microsoft.Build.Evaluation
             /// </summary>
             public string GetEscapedValueIfPresent(string itemType, string name)
             {
+                string value = null;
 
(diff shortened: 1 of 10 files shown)


pi1024e:
MSBuild.Dev.sln was open at the time.

rainersigwald:
Team triage: Thank you for the pull request. At the moment, we don't have the bandwidth to run this through the full Visual Studio performance test suite, and we don't think we can change the constant without that kind of coverage.

PR #4657, opened 2019-08-24
Adds ErrorIfLinkFails attribute for Copy Tasks.

gregwoodio:
When ErrorIfLinkFails is set, we error if a symbolic link or hard link fails. We also error if ErrorIfLinkFails is set without specifying we should use symbolic or hard linking for Copy tasks.

Addresses #4433.


gregwoodio:
@rainersigwald please see code review changes.

gregwoodio:
Great! Thanks for the review.

rainersigwald:
Thanks again!

PR #4658, opened 2019-08-24
Fix for IDE0025 (Use Method Expression for properties)

pi1024e:
Mostly done by Visual Studio autofixers


pi1024e:
It was a fix all solution, but I checked every single one and disabled the refactorings of object collection that was incorrect.

MSBuild.Dev.sln was open at the time.

rainersigwald:

PR #4659, opened 2019-08-24
Simplify delegate invokation (Auto-fix by IDE)

pi1024e:
Done so the project can meet the project style guidelines. Also makes code easier to read.


rainersigwald:
Thanks again!

PR #4661, opened 2019-08-25
Replaced "sealed internal" modifier with "internal sealed" in Solution

pi1024e:


rainersigwald:

PR #4685, opened 2019-08-31
Fix possible null dereference pointer in test

pi1024e:
In the object model helper test, if the expected array is null, and the other one is null, the test will then compare their contents. This is not supposed to happen, as the comment clearly states "Now that we verified the arrays are nonnull and equal in length..."

Had to do a logic fix there.

### DIFF
--- a/src/Shared/UnitTests/ObjectModelHelpers.cs
+++ b/src/Shared/UnitTests/ObjectModelHelpers.cs
@@ -251,14 +251,14 @@ namespace Microsoft.Build.UnitTests
 
             for (var i = 0; i < expectedItems.Length; i++)
             {
-                if (normalizeSlashes)
+                if (!normalizeSlashes)
                 {
-                    var normalizedItem = NormalizeSlashes(expectedItems[i]);
-                    items[i].EvaluatedInclude.ShouldBe(normalizedItem);
+                    items[i].EvaluatedInclude.ShouldBe(expectedItems[i]);
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -442,16 +442,14 @@ internal static void AssertArrayContentsMatch(object[] expected, object[] actual
...
-                Assert.NotNull(actual); // "Result should be non-null."
+                return;
             }
 
+            Assert.NotNull(actual); // "Result should be non-null."
             Assert.Equal(expected.Length, actual.Length); // "Expected array length of <" + expected.Length + "> but was <" + actual.Length + ">.");
 
             // Now that we've verified they're both non-null and of the same length, compare each item in the array.
-            for (int i = 0; i < expected.Length; i++)
+            for (int i = expected.Length - 1; i >= 0; i--)
```
BenVillalobos: The other changes to this function are good, but why change the loop to iterate from end -> beginning?

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -251,14 +251,14 @@ public static void AssertItems(string[] expectedItems, IList<TestItem> items, Di
 
             for (var i = 0; i < expectedItems.Length; i++)
             {
-                if (!normalizeSlashes)
+                if (normalizeSlashes)
```
BenVillalobos: Similar to my other comment, what do the changes in this function accomplish?

PR #4689, opened 2019-09-03
[WIP] Expander experiments

Forgind:
This isn't nearly ready to merge, but I would appreciate feedback as to whether the proposed changes would be valuable if brought up to our coding standards.


rainersigwald:
Team triage: let's come back to this when we have more time to dedicate.

PR #4694, opened 2019-09-04
Allow Before/AfterTargets for dynamic sln targets

### LINKED ISSUE #4692: Cannot BeforeTargets a dynamic solution traversal target
(This was reported offline by @davkean. Thanks!)

It's possible to add targets to a solution build by using [`after.{solutionname}.sln.targets`](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#customize-the-solution-build). It's also possible to specify a target that doesn't exist in the default solution metaproject, which calls that project on all of the projects in the solution (added with #1275).

But these features interact badly: you cannot `AfterTargets="DynamicallyAddedTraversalTarget"` in the solution targets.

[sln-targets-after.zip](https://github.com/microsoft/msbuild/files/3576513/sln-targets-after.zip)

The reason this happens is that the traversal targets are injected into the solution metaproject _instance_, after it has evaluated and loaded all of its imports

https://github.com/microsoft/msbuild/blob/440c90a9e61cffb78082313d04dba64bdc8d04f4/src/Build/Construction/Solution/SolutionProjectGenerator.cs#L800-L804

But during evaluation, `BeforeTargets` and `AfterTargets` can only point to existent targets

https://github.com/microsoft/msbuild/blob/440c90a9e61cffb78082313d04dba64bdc8d04f4/src/Build/Evaluation/Evaluator.cs#L1129-L1133

It looks like this doesn't apply to the known types because of

https://github.com/microsoft/msbuild/blob/440c90a9e61cffb78082313d04dba64bdc8d04f4/src/Build/Construction/Solution/SolutionProjectGenerator.cs#L879-L885

Perhaps that could just be extended?

rainersigwald:
Fixes #4692 by ensuring that the targets that will be created exist at evaluation time, too.

### DIFF
--- a/src/Build/Construction/Solution/SolutionProjectGenerator.cs
+++ b/src/Build/Construction/Solution/SolutionProjectGenerator.cs
@@ -879,7 +879,8 @@ namespace Microsoft.Build.Construction
             // These are just dummies necessary to make the evaluation into a project instance succeed when 
             // any custom imported targets have declarations like BeforeTargets="Build"
             // They'll be replaced momentarily with the real ones.
-            foreach (string targetName in _defaultTargetNames)
+            string[] dummyTargetsForEvaluationTime = _defaultTargetNames.Union(_targetNames).ToArray();
+            foreach (string targetName in dummyTargetsForEvaluationTime)
             {
                 traversalProject.AddTarget(targetName);
             }
@@ -903,8 +904,8 @@ namespace Microsoft.Build.Construction
                 _submissionId
                 );
 
-            // Make way for the real ones                
-            foreach (string targetName in _defaultTargetNames)
+            // Make way for the real ones
+            foreach (string targetName in dummyTargetsForEvaluationTime)
             {
                 traversalInstance.RemoveTarget(targetName);
             }
--- a/src/Build/Construction/Solution/SolutionProjectGenerator.cs
+++ b/src/Build/Construction/Solution/SolutionProjectGenerator.cs
@@ -882,7 +882,9 @@ namespace Microsoft.Build.Construction
             string[] dummyTargetsForEvaluationTime = _defaultTargetNames.Union(_targetNames).ToArray();
             foreach (string targetName in dummyTargetsForEvaluationTime)
             {
-                traversalProject.AddTarget(targetName);
+                ProjectTargetElement target = traversalProject.CreateTargetElement(targetName);
+                // Prepend so that any imported target overrides these default ones.
+                traversalProject.PrependChild(target);
             }
 
             // For debugging purposes: some information is lost when evaluating into a project instance,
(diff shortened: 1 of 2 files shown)


rainersigwald:
Ok, the failing test is correct. There are two interesting cases here:

* A target in the import _references_ a dynamically-created target.
* A target in the import _overrides_ a dynamically-created target.

in the reference case, we need the target to exist during evaluation, so adding a dummy and replacing it with content later makes sense. But in the override case, the target _must not_ exist during evaluation, or it won't get the real definition from the .sln.targets file.

I don't think I see a good way out of this.

src/Build/Construction/Solution/SolutionProjectGenerator.cs
```diff
@@ -903,10 +906,15 @@ private ProjectInstance CreateTraversalInstance(string wrapperProjectToolsVersio
...
 
-            // Make way for the real ones                
-            foreach (string targetName in _defaultTargetNames)
+            // Make way for the real ones
+            foreach (string targetName in dummyTargetsForEvaluationTime)
             {
-                traversalInstance.RemoveTarget(targetName);
+                // Remove targets only if they were the dummy ones (from the metaproj path),
+                // but leave them if they're from another source (imported/overridden).
+                if (traversalInstance.Targets[targetName].Location.File == traversalProject.FullPath)
```
cdmihai: What if they're the dummy ones but there's before / after targets hooked to them? Still good to remove?

cdmihai: Talked offline. Target execution uses the object model only, luckily.

src/Build/Construction/Solution/SolutionProjectGenerator.cs
```diff
@@ -882,7 +882,9 @@ private ProjectInstance CreateTraversalInstance(string wrapperProjectToolsVersio
             string[] dummyTargetsForEvaluationTime = _defaultTargetNames.Union(_targetNames).ToArray();
             foreach (string targetName in dummyTargetsForEvaluationTime)
             {
-                traversalProject.AddTarget(targetName);
+                ProjectTargetElement target = traversalProject.CreateTargetElement(targetName);
+                // Prepend so that any imported target overrides these default ones.
+                traversalProject.PrependChild(target);
```
cdmihai: Unless it comes in via an sdk right? But that would mean an sdk in the before / after project files, which we could say it's not supported.

cdmihai: What if you enhance ProjectTargetInstance to point to the previous target it overrides? Then you could massage the override chain via some internal APIs :)

cdmihai: Or have an internal only method that hooks up a different AddTarget implementation in ProjectInstance which the solution parser hooks up to ignore the dummy targets https://github.com/microsoft/msbuild/blob/62fb89029dac66f73a5ca52363fd11b36f12190b/src/Build/Instance/ProjectInstance.cs#L1354

PR #4690, opened 2019-09-04
Exp/version 4.8

Forgind:
Updating .NET SDK to version 4.8.

Issue #4676, opened 2019-08-29
Add a .vsconfig to the repo


rainersigwald:
Visual Studio 2019 now supports [checking in a `.vsconfig` file next to a solution](https://docs.microsoft.com/en-us/visualstudio/install/import-export-installation-configurations?view=vs-2019#automatically-install-missing-components), which presents users with a nice dialog offering to install the required workloads if they're missing some.

The MSBuild repo should have one of these. That's a much nicer experience than the existing "manually install stuff based off `README.md`".

rainersigwald:
@forgind, can you take a look at this?

Forgind:
Will do

PR #4698, opened 2019-09-05
Better error when tlbexp isn't found

rainersigwald:
This situation can occur on fresh VS 16.3 installs, which install the .NET 4.8 SDK
but don't yet have #4690 and so still look for tlbimp in the .NET 4.6.2 SDK.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -108,8 +108,9 @@
   </Target>
 
   <Target Name="CreateTypeLib" BeforeTargets="AfterBuild" Condition="'$(BuildingInsideVisualStudio)' != 'true' and '$(CreateTlb)' == 'true' and '$(OsEnvironment)' == 'windows' and '$(TargetFrameworkIdentifier)' == '.NETFramework' and '$(MSBuildRuntimeType)' != 'Core'">
-    <!-- Provide a mechanism for turning on verbose TlbExp output for diagnosing issues -->
     <PropertyGroup>
+      <TlbExpPath>$([Microsoft.Build.Utilities.ToolLocationHelper]::GetPathToDotNetFrameworkSdkFile('tlbexp.exe'))</TlbExpPath>
+      <!-- Provide a mechanism for turning on verbose TlbExp output for diagnosing issues -->
       <TlbExpVerbosity Condition="'$(VerboseTlbExp)' == 'true'">/verbose</TlbExpVerbosity>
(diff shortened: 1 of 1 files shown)


Forgind:
"default .NET SDK" isn't entirely clear; a user might think that means the latest version, i.e., 4.8 right now, rather than 4.6.2. Perhaps "...MSBuild is trying to refer to the latest version of the .NET SDK to which it was updated, and that isn't installed."

What do you think?

rainersigwald:
Would it be clearer if I just removed `default`? I don't want to make this message too too complicated, just marginally clearer than the `error 9009` running `"" /silent /nologo ...`

Forgind:
That would still leave me wondering to which SDK you're referring. It would be helpful to have some specificity to make it clear which SDK is needed without giving a number (so it remains true in future updates). If you want just a marginal improvement, that would work, but I'm trying to avoid having someone see the helpful message, download a new SDK but not 4.6.2, and when it doesn't work, decide the message is just wrong and start looking elsewhere.

rainersigwald:
Updated to a "what you should do" message, rather than a "what the problem is" one. If someone encounters this it's because they don't have the right workloads installed.

src/Directory.Build.targets
```diff
@@ -122,7 +123,10 @@
       <TlbExpAssemblyPaths Include="@(_TlbExpAssemblyPaths->'%(SlashlessPath)')" />
     </ItemGroup>
 
-    <Exec Command="&quot;$([Microsoft.Build.Utilities.ToolLocationHelper]::GetPathToDotNetFrameworkSdkFile('tlbexp.exe'))&quot; $(TlbExpVerbosity) /NOLOGO @(TlbExpAssemblyPaths->'/asmpath:&quot;%(Identity)&quot;', ' ') &quot;$(TargetPath)&quot; /out:&quot;$(TargetDir)$(TargetName).tlb&quot;" />
+    <Error Condition=" '$(TlbExpPath)' == '' "
+           Text="TlbExp was not found. Ensure that you have installed everything from .vsconfig. If you have, please report a bug to MSBuild." />
```
cdmihai: `installed everything from .vsconfig` sounds ambiguous. But maybe it's one of those "you know it when you see it" kind of things.

rainersigwald: After #4676, Visual Studio itself should prompt to install workloads whenever you open one of the solutions, so _hopefully_ this is fairly rare/clear going forward. But if people stumble over it we can change it.

PR #4699, opened 2019-09-05
Fix copying of DiscardBuildResults flag for BuildParameters

panopticoncentral:
When I added the `DiscardBuildResults` flag to `BuildParameters`, I forgot to add it to the `Clone` method. So the flag is lost in the `BuildManager` since the first thing it does is clone the parameters. Also added a comment to the effect that we do not serialize this to the nodes since it's not applicable there.

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/Build/BackEnd/BuildManager/BuildParameters.cs
@@ -832,7 +832,7 @@ namespace Microsoft.Build.Execution
             // ResetCaches is not transmitted.
             // LegacyThreadingSemantics is not transmitted.
             // InputResultsCacheFiles and OutputResultsCacheFile are not transmitted, as they are only used by the BuildManager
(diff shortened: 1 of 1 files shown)


src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -831,6 +832,7 @@ void ITranslatable.Translate(ITranslator translator)
             // ResetCaches is not transmitted.
             // LegacyThreadingSemantics is not transmitted.
             // InputResultsCacheFiles and OutputResultsCacheFile are not transmitted, as they are only used by the BuildManager
+            // DiscardBuildResults is not trasmitted.
```
rainersigwald:
```suggestion
            // DiscardBuildResults is not transmitted.
```

panopticoncentral:
@rainersigwald Can you merge the PR?

PR #4703, opened 2019-09-06
Log unexpanded import expression in failed-to-import error

### LINKED ISSUE #794: MSB4019 could include the unexpanded value of the Project attribute
This error currently looks like this:

```
O:\msbuild\bin\Bootstrap\Microsoft.Xaml.targets(19,4): error MSB4019: The imported project "O:\Microsoft.Xaml.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [O:\msbuild\src\dirs.proj]
```

The first step in debugging a problem that results in this error is almost always "look in the file from the far left of the message and see what's in the `<Import>` element". The user could save a step in some cases if the error message was extended to add something like

```
Confirm that the path in the Import declaration ("$(MSBuildFrameworkToolsPath)\Microsoft.Xaml.targets") is correct, and that the file exists on disk.
```

treymerkley:
First pull on the project this size, so sorry if there are some boneheaded errors in here.


src/Build/Resources/Strings.resx
```diff
@@ -424,7 +424,7 @@
     <comment>{StrBegin="MSB4101: "}</comment>
   </data>
   <data name="ImportedProjectNotFound" UESanitized="false" Visibility="Public">
-    <value>MSB4019: The imported project "{0}" was not found. Confirm that the path in the &lt;Import&gt; declaration is correct, and that the file exists on disk.</value>
+    <value>MSB4019: The imported project "{0}" was not found. Confirm that the path in the Import declaration "{1}" is correct, and that the file exists on disk.</value>
```
rainersigwald: Can I ask why you removed the angle brackets around `Import`? I don't necessarily disagree, just curious.

treymerkley: Oh yeah, it's just because when we were talking about it, there were no brackets in the changed version you suggested. I just wanted to make sure it was *right*. If you want I can put them back.

rainersigwald:
You ran into an unfortunate collision with updated translations from the `vs16.3` release branch. I pushed a couple of updates that should fix the build.

rainersigwald:
Thanks again!

Issue #2005, opened 2017-04-24
Parallel build warning is displayed when using /m on a single-proc machine


Mertsch:
As you can see in the following log, the `/m` switch is clearly used, yet there is a message about using `/m`.

```
>"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\msbuild.exe" "C:\Build\_work\1\s\x.sln" /m
Microsoft (R) Build Engine version 15.1.1012.6693
Copyright (C) Microsoft Corporation. All rights reserved.

Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
Build started 24.04.2017 14:35:45.
```

rainersigwald:
That's . . . very odd. I do not reproduce the issue on my machine, using the same version of MSBuild:

```
s:\msbuild2>"C:\Program Files (x86)\Microsoft Visual Studio\Preview\EntInt\MSBuild\15.0\Bin\MSBuild.exe" "s:\msbuild2\src\MSBuild.sln" /m
Microsoft (R) Build Engine version 15.1.1012.6693
Copyright (C) Microsoft Corporation. All rights reserved.

Build started 4/24/2017 11:19:29 AM.
```

Are you seeing this consistently on multiple machines?

Mertsch:
@rainersigwald No, on my dev machine it does not happen. May be an issue with moving from VS2017 Preview to RTW.

Will do a repair on the build server and report back.

Mertsch:
@rainersigwald Still happens on my build server even after a "Repair" installation, any suggestion on how to proceed?

rainersigwald:
@Mertsch Unfortunately, no. The [code that emits that message](https://github.com/Microsoft/msbuild/blob/1b91be584e400d5b9ffb8711bbb926beedb25407/src/MSBuild/XMake.cs#L588-L591) looks pretty straightforward, as does the [code that interprets the command-line switch](https://github.com/Microsoft/msbuild/blob/1b91be584e400d5b9ffb8711bbb926beedb25407/src/MSBuild/XMake.cs#L2259-L2282) itself.

You could try uninstall + reinstall (I feel bad even suggesting this), diffing the MSBuild directory and assemblies against your working install, or potentially debugging into MSBuild.exe to see what's going wrong.

Mertsch:
@rainersigwald A complete uninstall / reinstall did not help.
[...]

PR #4711, opened 2019-09-10
Prevent maxcpucount warning when /m argument has been specified

### LINKED ISSUE #4705: spurious maxcpucount warning
With the following *explicit* use of `-m` switch:

```
msbuild /clp:Verbosity=minimal /nologo z-parallel-projects.sln /m:1
```

we get the following spurious warning:

```
Building the projects in this solution one at a time. To enable parallel build, please add the “/m” switch.
…
```

Not only that the message is redundant, but also our users complain and worry about the quality of an internal tooling in house which uses MSBuild in the back end.

Matthiee:
Fixes #4705 and #2005

Added an extra case to only display the message when no "/m" argument was specified. 

I'm adding WIP since I want to get some feedback first! 

I didn't create an unit test for it (yet) as this is deeply hidden inside of the `MSBuild.Execute` and I'm still thinking about the best way to proceed. 

One of the options I'm thinking about is to change the signature of `ProcessMaxCPUCountSwitch` to return a `Tuple<int, bool>` or add a new `out bool found` parameter.  

Another option is to extract the whole if statement to a seperate method.

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -608,7 +608,8 @@ namespace Microsoft.Build.CommandLine
                     // is a better way.
                     // Only display the message if /m isn't provided
                     if (cpuCount == 1 && FileUtilities.IsSolutionFilename(projectFile) && verbosity > LoggerVerbosity.Minimal
-                        && switchesNotFromAutoResponseFile[CommandLineSwitches.ParameterizedSwitch.MaxCPUCount].Length == 0)
+                        && switchesNotFromAutoResponseFile[CommandLineSwitches.ParameterizedSwitch.MaxCPUCount].Length == 0
+                        && switchesFromAutoResponseFile[CommandLineSwitches.ParameterizedSwitch.MaxCPUCount].Length == 0)
                     {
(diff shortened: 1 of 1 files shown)


src/MSBuild/XMake.cs
```diff
@@ -606,7 +606,9 @@ string commandLine
                     // Unfortunately /m isn't the default, and we are not yet brave enough to make it the default.
                     // However we want to give a hint to anyone who is building single proc without realizing it that there
                     // is a better way.
-                    if (cpuCount == 1 && FileUtilities.IsSolutionFilename(projectFile) && verbosity > LoggerVerbosity.Minimal)
+                    // Only display the message if /m isn't provided
+                    if (cpuCount == 1 && FileUtilities.IsSolutionFilename(projectFile) && verbosity > LoggerVerbosity.Minimal
+                        && switchesNotFromAutoResponseFile[CommandLineSwitches.ParameterizedSwitch.MaxCPUCount].Length == 0)
```
rainersigwald: Can you also check in `switchesFromAutoResponseFile`? I'm not sure it's reasonable to require it to be passed on the command line.

rainersigwald:
Thanks, @Matthiee!

Issue #2810, opened 2017-12-14
Log process path, command line and current directory at start of build


KirillOsenkov:
Would be great to log 3 messages at start of build:

1. Full path to the MSBuild process
2. Command line passed to MSBuild
3. Current directory

Structured Log Viewer emulates these, but only for builds initiated from the Structured Log Viewer itself.

I'm thinking if the verbosity is at least Detailed?

PR #4716, opened 2019-09-11
FileLogger writes current logging verbosity

### LINKED ISSUE #2426: File loggers should log their verbosity
I look through a lot of MSBuild logs, and when they come from other people I don't always know what verbosity they're logged at--so it's not always clear whether a log message I'm looking for is missing because that build event didn't happen, or because the verbosity isn't where I expect it.

I'd like to see a line at the top of the log that mentions what verbosity is configured.

CodeTherapist:
closes #2426

### DIFF
--- a/src/Build.UnitTests/FileLogger_Tests.cs
+++ b/src/Build.UnitTests/FileLogger_Tests.cs
@@ -369,6 +369,49 @@ namespace Microsoft.Build.UnitTests
         }
 
         /// <summary>
+        /// File logger is writting the verbosity level as soon the build starts.
+        /// </summary>
+        [Theory]
+        [InlineData(LoggerVerbosity.Quiet, false)]
+        [InlineData(LoggerVerbosity.Minimal, false)]
+        [InlineData(LoggerVerbosity.Normal, true)]
+        [InlineData(LoggerVerbosity.Detailed, true)]
+        [InlineData(LoggerVerbosity.Diagnostic, true)]
+        public void LogVerbosityMessage(LoggerVerbosity loggerVerbosity, bool shouldContain)
+        {
+            using (var testEnvironment = TestEnvironment.Create())
+            {
+                var fileLogger = new FileLogger
+                {
+                    Verbosity = loggerVerbosity
+                };
+
+                var logFile = testEnvironment.CreateFile(".log");
+                fileLogger.Parameters = "logfile=" + logFile.Path;
+
+                Project project = ObjectModelHelpers.CreateInMemoryProject(@"
+                <Project ToolsVersion=`msbuilddefaulttoolsversion` xmlns=`msbuildnamespace`>
+                    <Target Name=`Build` />
+                </Project>
+                ");
+
+                project.Build(fileLogger);
+                project.ProjectCollection.UnregisterAllLoggers();
+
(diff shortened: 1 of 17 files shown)


src/Build/Logging/FileLogger.cs
```diff
@@ -81,6 +81,11 @@ private void InitializeFileLogger(IEventSource eventSource, int nodeCount)
             // want to make decisions based on our verbosity, so we do this last.
             base.Initialize(eventSource, nodeCount);
 
+            if (!SkipProjectStartedText && Verbosity >= LoggerVerbosity.Normal)
+            {
+                eventSource.BuildStarted += (obj, args) => WriteHandler($"Verbosity is set to: {Verbosity}.");
```
rainersigwald: This string needs to be localized. Can you add a new string to [`src/Build/Resources/Strings.resx`](https://github.com/microsoft/msbuild/blob/master/src/Build/Resources/Strings.resx) and use it here with `ResourceUtilities.FormatResourceStringIgnoreCodeAndKeyword()`?

rainersigwald: I think this is a reasonable way to do this right now. I think in the future we may wish to combine this message with a solution to #2810, but we don't have to solve that right now.

CodeTherapist: Yes, sure - I appreciate your help as always, thanks 😃 
I should have known that localization is required 🙄

CodeTherapist: Do you point me to the next issue that I could do? 😉

src/Build.UnitTests/FileLogger_Tests.cs
```diff
@@ -368,6 +368,49 @@ public void EmptyErrorLogUsingWarningsErrorsOnly(string loggerOption)
...
+                <Project ToolsVersion=`msbuilddefaulttoolsversion` xmlns=`msbuildnamespace`>
+                    <Target Name=`Build` />
+                </Project>
+                ");
+
+                project.Build(fileLogger);
+                project.ProjectCollection.UnregisterAllLoggers();
+
+                string log = File.ReadAllText(logFile.Path);
+                var message = $"Logging verbosity is set to: {loggerVerbosity}.";
```
rainersigwald: This will fail on non-English OSes. It might not be the only test to do so but we try to support building and testing MSBuild on any language. We generally use

again in the tests.
```suggestion
                var message = ResourceUtilities.FormatResourceStringStripCodeAndKeyword("LogLoggerVerbosity", loggerVerbosity);
```

CodeTherapist: Good catch again, thanks for the help - that is again super obviously 🙄 I had to know that better...

src/Build/Logging/FileLogger.cs
rainersigwald: I'd recommend looking at the [up-for-grabs label](https://github.com/microsoft/msbuild/labels/up-for-grabs) and seeing if anything catches your eye.

CodeTherapist: Sure, I do exactly that, when seeking for new work 😄

rainersigwald:
Thanks again @CodeTherapist!

PR #4724, opened 2019-09-13
Microsoft-Build trace events

Forgind:
Replaces (outdated) CodeMarkers with EventSource. Also adds a few new locations for profiling in LazyItemEvaluation.


documentation/specs/event-source.md
```diff
@@ -0,0 +1,25 @@
...
+* LoadDocument: loads an XMLDocumentWithLocation from a path.
+* RemoveReferencesMarkedForExclusion: removes blacklisted references from the reference table, putting primary and dependency references in invalid file lists.
+* ComputeClosure: resolves references to, for example, properties to explicit values.
+* EvaluateCondition: checks whether a condition is true and removes false conditionals.
+* Parse: parses an XML document into a ProjectRootElement.
+* Evaluate: Evaluates a project, running several other parts of MSBuild in the process.
+* ExecuteGenerateResource: uses resource APIs to transform resource files into strongly-typed resource classes.
+* SelectItems: identifies a list of files that correspond to an item, potentially with a wildcard.
+* Apply: collects a set of items, mutates them in a specified way, and saves the results.
+* ExecuteTask: executes a task.
```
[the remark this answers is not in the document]
rainersigwald: More broadly, can you please take a look at the (internal) source for MSBuild 14, before it was open-sourced, and see if any CodeMarkers have been lost over time since they weren't building in this repo?  https://dev.azure.com/devdiv/DevDiv/_git/VS?path=%2Fsrc%2Fxmake&version=GC549bfca89f672601e9632bbde2d0ed32cfa1b6b1

documentation/specs/event-source.md
```diff
@@ -0,0 +1,25 @@
...
+* Save: saves a project to the file system if dirty, creating directories as necessary.
+* LogResults: logs the results from having executed a task.
+
+## Larger context
+
+MSBuild is the build system underlying Visual Studio. As millions of developers use Visual Studio, improvements in MSBuild's efficiency affect millions of developers and their customers.
```
rainersigwald: I think you can cut this paragraph.

documentation/specs/event-source.md
```diff
@@ -0,0 +1,25 @@
...
+
+## EventSource in MSBuild
+EventSource is primarily used to profile code. For MSBuild specifically, a major goal is to reduce the time it takes to run, as measured by the Regression Prevention System (RPS), i.e., running specific scenarios. To find which code segments were likely candidates for improvement, EventSources were added around a mix of processes. Large, high-level processes occur nearly every run of MSBuild andtake a long time. They generally only relatively few times. Smaller methods with well-defined purposes may occur numerous times. Profiling both types of events provides both broad strokes to identify large code segments that underperform and, more specifically, which parts ot them. Profiled functions include:
```
rainersigwald: Depending on scenario, wall-clock time is often not the most important performance factor for MSBuild. For instance, VS scenarios are very strongly affected by memory usage/GC pause time, so improvements there often come in the form "reduce allocations in path X". RPS measures many things besides wall-clock time, and the newer perf infrastructure is further deemphasizing wall-clock time in favor of less-variable metrics.

rainersigwald: Can you avoid using the word `process` except when referring to an OS process? I see what you're going for here, but I found it confusing. Maybe `codepath`?

rainersigwald: It's my understanding that these events aren't directly useful for _profiling_, but can augment or even replace the need for a profiler by being application-specific. Is that misguided?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -645,33 +619,23 @@ private void Evaluate(ILoggingService loggingService, BuildEventContext buildEve
 
                 ErrorUtilities.VerifyThrow(_data.EvaluationId != BuildEventContext.InvalidEvaluationId, "Evaluation should produce an evaluation ID");
 
-#if MSBUILDENABLEVSPROFILING
-        string endPass0 = String.Format(CultureInfo.CurrentCulture, "Evaluate Project {0} - End Pass 0 (Initial properties)", projectFile);
-        DataCollection.CommentMarkProfile(8816, endPass0);
-#endif
-
+                EvaluateEventSource.Log.EvaluatePhase(projectFile, 0);
```
rainersigwald: Why are these not start/stop pairs per phase?

src/Build/Evaluation/LazyItemEvaluator.cs
```diff
@@ -87,6 +88,7 @@ public bool EvaluateConditionWithCurrentState(ProjectElement element, ExpanderOp
             {
                 return true;
             }
+            EvaluateConditionEventSource.Log.EvaluateConditionStart(String.Format("Evaluate Condition {0} - Begin", condition));
```
rainersigwald: Is it worth the allocations and time to do a `string.Format` here, as opposed to just conveying the "Evaluate Condition - Begin" information with the event type?

rainersigwald: (general question, not just for this specific instance of the pattern)

src/Framework/EventSource/Keywords.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Build.Eventing
+{
+    public class Keywords
+    {
+        public const EventKeywords Task = (EventKeywords)1;
+        public const EventKeywords Item = (EventKeywords)2;
+        public const EventKeywords Project = (EventKeywords)3;
```
rainersigwald: How did you derive these keywords? How would we expect them to be used?

documentation/specs/event-source.md
```diff
@@ -0,0 +1,25 @@
+# EventSource
+
+[EventSource](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource?view=netframework-4.8) is the tool that allows Event Tracing for Windows (ETW) used in MSBuild. Among its useful features, functions with names ending in "start" and "stop" correlate between calls such that it can automatically record how long the event between the two calls took. It also provides an easy way to cheaply opt in or out, log auxiliary messages in addition to time, and add progress updates in the middle of an event as needed.
```
rainersigwald: Do the integers in the EventSource correlation ID map exactly to the various MSBuild IDs? If not, can they be made to?

src/Build/Evaluation/Evaluator.cs
Forgind: These are intermediate points within Evaluate. It's a relatively long function, so it's helpful to be able to find more precisely where the slow points are, but it doesn't "start" or "stop" at these points, only at the very beginning and end respectively.

Forgind: Also, there were CodeMarkers there originally, if I remember correctly.

src/Framework/EventSource/Keywords.cs
Forgind: We can change the keywords if you think there are more helpful options.

This is useful with IsEnabled. IsEnabled with two parameters specifies how important a particular event is (informational, error, etc.) and an EventKeywords. The first lets the user specify whether they want all the eventsources to fire or just the most critical ones (as determined arbitrarily), and the second lets them turn on/off the eventsources related to a specific goal. IsEnabled also checks the commandline argument to make sure the particular eventsource is enabled, so it actually checks three things.

documentation/specs/event-source.md
Forgind: What do you mean by correlation ID and MSBuild IDs?

rainersigwald: the "EventSource correlation ID" I mentioned is displayed as the ActivityID in PerfView:

image

My question is: do the various integers there match up with MSBuild's internal IDs like `ProjectID` or `BuildRequest.GlobalRequestId`? If not, can they be made to do so?

src/Framework/EventSource/Keywords.cs
rainersigwald: My initial reaction is "I don't know how I'd use these so do we need to put them in?"

I'd place a higher value on getting useful information out of the traces in something like the RPS case than in being able to filter specifically on the command line. That feels like an "only if we discover that some useful events are surprisingly expensive" operation.

livarcocc:
This needs validation before being merged.

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -1,5 +1,128 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+namespace Microsoft.Build.Eventing
```
brianrob: I recommend that you merge these EventSources into one EventSource class unless there isn't an assembly you can put it into that is accessible in all the necessary places.  There is cost to having lots of EventSources:

 - Initialization: There is some fixed cost to initialization that you'll pay for every one of these.
 - Usage: Folks have to know about all of these.

If the goal is to be able to enable/disable these events individually, there are better ways to do this through the keywords mechanism.

I also recommend that you use more global names.  Remember that the EventSource name is part of a global namespace of EventSources and so it's important to disambiguate.  You should use something like "Microsoft-Build" for the name of a single EventSource that includes all of these events.

I'm happy to walk through the details on how to structure the EventSource so that it meets your needs while doing things in an EventSource idiomatic way that users will understand.

src/MSBuild/XMake.cs
```diff
@@ -515,14 +516,7 @@ string commandLine
...
             {
-#if (!STANDALONEBUILD)
-                // Enable CodeMarkers for MSBuild.exe
-                CodeMarkers.Instance.InitPerformanceDll(CodeMarkerApp.MSBUILDPERF,  String.Format(CultureInfo.InvariantCulture, @"Software\Microsoft\MSBuild\{0}", MSBuildConstants.CurrentProductVersion));
-#endif
-#if MSBUILDENABLEVSPROFILING 
-                string startMSBuildExe = String.Format(CultureInfo.CurrentCulture, "Running MSBuild.exe with command line {0}", commandLine);
-                DataCollection.CommentMarkProfile(8800, startMSBuildExe);
-#endif
+                XMakeEventSource.Log.XMakeStart("Execute MSBuild - Begin");
```
brianrob: Do the strings that are being passed into each event add any value?  Generally, you want the payload to any event to be specific to that event.  The payloads in the events that I've seen thus far are essentially duplicating the meaning of the event itself.

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
Forgind: Somehow, I didn't realize that you could have them all in a single class. I can certainly do that—thanks for the suggestion!

We originally thought it might be helpful to separate them out to some extent, but if our main focus is their use in RPS, and we want to capture all information in RPS, I'm not sure we still need that.

The names were temporary until we decided what names we should use, and we never got around to changing them. It will be much simpler when they're all in one class.

brianrob: Excellent.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,338 @@
...
+using System.Globalization;
+
+
+//
+// This captures information of how various key methods of building with MSBuild ran.
+//
+namespace Microsoft.Build.Eventing
+{
+    // /OnlyProviders=*MSBuild-Profiling
+    [EventSource(Name = "MSBuild-Profiling")]
```
brianrob: Nit: I would just call this "Microsoft-Build" to match the standard naming conventions that are used across products.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,338 @@
...
+        private MSBuildEventSource() { }
+
+        #endregion
+
+        #region Events and NonEvents
+
+        /// <summary>
+        /// Call this method to notify listeners of information relevant to collecting a set of items, mutating them in a specified way, and saving the results.
+        /// </summary>
+        public void ApplyStart()
```
brianrob: Each event should also have an `EventAttribute` on it.  So for ApplyStart it would be `[Event(1)]` where the 1 matches the value that you pass to WriteEvent.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,338 @@
...
+        {
+            if (IsEnabled()) WriteEvent(1);
```
brianrob: You should remove the IsEnabled calls from all of the event methods.  IsEnabled is called inside of WriteEvent so this is just duplicate work.

If there is something that is an argument that is expensive to compute, you should put that code in an `IsEnabled` check, but that should occur outside of `Event` methods.  If you want it to happen in a method that is inside of the `EventSource` class, you can create a method and mark it as `[NonEvent]` so that it's not treated as an event.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,338 @@
...
+
+        /// <summary>
+        /// Call this method to notify listeners of how the project data was evaluated.
+        /// </summary>
+        /// <param name="file">Relevant information about where in the run of the progam it is.</param>
+        public void EvaluateStart(string file)
+        {
+            if (IsEnabled())
+            {
+                string projectFile = String.IsNullOrEmpty(file) ? "(null)" : file;
```
brianrob: You should not need this.  You can detect a null or empty argument at analysis time.  NULL values are valid values in traces.  Also, if you want to do processing of arguments it needs to happen outside of `[Event]` methods, either in a `[NonEvent]` method or at the callsite itself.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,338 @@
...
+        /// Call this method to notify listeners of progress made on Evaluate.
+        /// </summary>
+        /// <param name="file">Relevant information about where in the run of the progam it is.</param>
+        /// <param name="phaseNum">The stage in Evaluate that the program has reached.</param>
+        public void EvaluatePhase(string file, long phaseNum)
+        {
+            if (IsEnabled())
+            {
+                string projectFile = String.IsNullOrEmpty(file) ? "(null)" : file;
+                WriteEvent(12, String.Format(CultureInfo.CurrentCulture, "Evaluate Project {0}", projectFile), phaseNum);
```
brianrob: You should avoid this sort of string formatting in EventSources.  It's expensive at collection time and isn't necessary.  The standard idiom here is to just write an event that takes a `string file` and  `long phaseNum` and just emit those directly into the trace.  What's happening here is understood because in addition to the arguments, you also get the event name which has the semantic meaning that you want to impart to the viewer of the trace.

documentation/specs/event-source.md
Forgind: I believe MSBuild's internal IDs are assigned by MSBuild at build time, whereas these numbers are assigned by a combination of the eventsource code and how many times a particular event has run within this build, so they cannot be made to align, at least not from this side. It's feasible that MSBuild could choose numbers based on how eventsource will work.

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -1,5 +1,84 @@
...
+{
+    [System.Diagnostics.Tracing.EventSourceAttribute(Name="Microsoft-Build-Profiling")]
+    public sealed partial class MSBuildEventSource : System.Diagnostics.Tracing.EventSource
```
brianrob: One other thing here.  You should just make MSBuildEventSource internal.  You shouldn't need it to be public, and you won't want anyone else building an app that emits msbuild events.

src/Framework/MSBuildEventSource.cs
brianrob: I'd also remove "-Profiling" as well.  This is more generically useful than just for profiling.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -20,17 +20,12 @@
 using Microsoft.Build.Construction;
 using Microsoft.Build.Execution;
 using Microsoft.Build.Exceptions;
-using Microsoft.Build.Evaluation;
+using Microsoft.Build.Eventing;
 using Microsoft.Build.Collections;
+using Microsoft.Build.Evaluation;
```
rainersigwald: nit: can you keep this list sorted please?

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -699,27 +694,9 @@ private async Task RequestThreadProc(bool setThreadParameters)
...
-                using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildEngineBuildProjectBegin, CodeMarkerEvent.perfMSBuildEngineBuildProjectEnd))
-                {
-#if MSBUILDENABLEVSPROFILING
-                try
-                {
-                    string beginProjectBuild = String.Format(CultureInfo.CurrentCulture, "Build Project {0} - Start", requestEntry.RequestConfiguration.ProjectFullPath);
-                    DataCollection.CommentMarkProfile(8802, beginProjectBuild);
-#endif
-#endif
+                MSBuildEventSource.Log.RequestThreadProcStart();
```
rainersigwald: Path was dropped -- intentionally?

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
```diff
@@ -1046,6 +1023,8 @@ private async Task<BuildResult> BuildProject()
             // logged with the node logging context
             _projectLoggingContext = null;
 
+            MSBuildEventSource.Log.BuildProjectStart();
```
rainersigwald: Not logging path, requested targets?

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -1508,17 +1505,8 @@ public void Save(Encoding saveEncoding)
...
-            {
-                string beginProjectSave = String.Format(CultureInfo.CurrentCulture, "Save Project {0} To File - Begin", projectFileLocation.File);
-                DataCollection.CommentMarkProfile(8810, beginProjectSave);
-#endif
-
             Directory.CreateDirectory(DirectoryPath);
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildProjectSaveToFileBegin, CodeMarkerEvent.perfMSBuildProjectSaveToFileEnd))
-#endif
+            MSBuildEventSource.Log.SaveStart();
```
rainersigwald: Path?

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -2050,16 +2031,11 @@ private XmlDocumentWithLocation LoadDocument(string fullPath, bool preserveForma
                 FullPath = fullPath,
                 PreserveWhitespace = preserveFormatting
             };
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildProjectLoadFromFileBegin, CodeMarkerEvent.perfMSBuildProjectLoadFromFileEnd))
-#endif
+
             {
```
rainersigwald: nit: can you delete braces on this block now that it's never a `using`?

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -2050,16 +2031,11 @@ private XmlDocumentWithLocation LoadDocument(string fullPath, bool preserveForma
...
                 try
                 {
-#if MSBUILDENABLEVSPROFILING
-                    string beginProjectLoad = String.Format(CultureInfo.CurrentCulture, "Load Project {0} From File - Start", fullPath);
-                    DataCollection.CommentMarkProfile(8806, beginProjectLoad);
-#endif
+                    MSBuildEventSource.Log.LoadDocumentStart();
```
rainersigwald: Path?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -657,21 +631,16 @@ private void Evaluate(ILoggingService loggingService, BuildEventContext buildEve
                 }
 
                 SetAllProjectsProperty();
-
+
```
rainersigwald: nit
```suggestion

```

src/Build/Evaluation/Evaluator.cs
```diff
@@ -657,21 +631,16 @@ private void Evaluate(ILoggingService loggingService, BuildEventContext buildEve
...
                 _data.InitialTargets = initialTargets;
-#if (!STANDALONEBUILD)
-        CodeMarkers.Instance.CodeMarker(CodeMarkerEvent.perfMSBuildProjectEvaluatePass1End);
-#endif
-#if MSBUILDENABLEVSPROFILING
-        string endPass1 = String.Format(CultureInfo.CurrentCulture, "Evaluate Project {0} - End Pass 1 (Properties and Imports)", projectFile);
-        DataCollection.CommentMarkProfile(8817, endPass1);
-#endif
+                MSBuildEventSource.Log.EvaluatePhase1Stop(projectFile);
+                MSBuildEventSource.Log.EvaluatePhase2Start(projectFile);
```
rainersigwald: Extreme nit: move `Phase2Start` line after the Pass2 block comment?

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,465 @@
...
+using System.Diagnostics.Tracing;
+using System.Globalization;
+
+
+//
+// This captures information of how various key methods of building with MSBuild ran.
+//
+namespace Microsoft.Build.Eventing
+{
+    // /OnlyProviders=*MSBuild-Profiling
```
rainersigwald: Is this comment useful? If it is, is it stale?

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,465 @@
+﻿using System;
+using System.IO;
+using System.Diagnostics.Tracing;
+using System.Globalization;
+
+
+//
+// This captures information of how various key methods of building with MSBuild ran.
```
rainersigwald: We don't generally comment namespaces. If you want to, please use the XML doc comment syntax.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,465 @@
...
+//
+namespace Microsoft.Build.Eventing
+{
+    // /OnlyProviders=*MSBuild-Profiling
+    [EventSource(Name = "Microsoft-Build")]
+    internal sealed class MSBuildEventSource : EventSource
+    {
+        #region Singleton instance
```
rainersigwald: We don't usually add `#region` in new code (we often delete it in old code). No strong reason IMO, just generally out of style.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,465 @@
...
+
+        // define the singleton instance of the event source
```
rainersigwald: Comment not super useful.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,465 @@
...
+        {
+            WriteEvent(11, projectFile);
+        }
+
+        /// <summary>
+        /// Call this method to notify listeners of progress made on Evaluate.
+        /// </summary>
+        /// <param name="projectFile">Relevant information about where in the run of the progam it is.</param>
+        [Event(12)]
+        public void EvaluatePhase0Start(string projectFile)
```
rainersigwald: Evaluator pass events should be named `Pass`, instead of `Phase`.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -308,15 +294,8 @@ internal class Evaluator<P, I, M, D>
...
-#if MSBUILDENABLEVSPROFILING 
-            }
-            finally
-            {
-                string projectFile = String.IsNullOrEmpty(root.ProjectFileLocation.File) ? "(null)" : root.ProjectFileLocation.File;
-                string beginProjectEvaluate = String.Format(CultureInfo.CurrentCulture, "Evaluate Project {0} - End", projectFile);
-                DataCollection.CommentMarkProfile(8813, beginProjectEvaluate);
-            }
-#endif
+                MSBuildEventSource.Log.EvaluatePhase5Stop(root.ProjectFileLocation.File);
```
rainersigwald: Why is pass5 stop here, instead of paired with its start?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -280,17 +274,9 @@ internal class Evaluator<P, I, M, D>
...
-            {
-#if MSBUILDENABLEVSPROFILING
-            try
             {
-                string projectFile = String.IsNullOrEmpty(root.ProjectFileLocation.File) ? "(null)" : root.ProjectFileLocation.File;
-                string beginProjectEvaluate = String.Format(CultureInfo.CurrentCulture, "Evaluate Project {0} - Begin", projectFile);
-                DataCollection.CommentMarkProfile(8812, beginProjectEvaluate);
-#endif
+                MSBuildEventSource.Log.EvaluateStart(root.ProjectFileLocation.File);
+                MSBuildEventSource.Log.EvaluatePhase0Start(root.ProjectFileLocation.File);
```
rainersigwald: Why is pass0 start here instead of where it used to be?

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,465 @@
...
+        {
+            WriteEvent(12, projectFile);
+        }
+
+        /// <summary>
+        /// Call this method to notify listeners of progress made on Evaluate.
```
rainersigwald: I don't find many of the doc comments in this particularly enlightening. I'd consider just not including them; I'm not sure it's worth the effort to improve them.

src/Build/Evaluation/LazyItemEvaluator.IncludeOperation.cs
```diff
@@ -97,6 +100,7 @@ protected override ImmutableList<I> SelectItems(ImmutableList<ItemData>.Builder
                         }
 
                         string[] includeSplitFilesEscaped;
+                        MSBuildEventSource.Log.SelectItemsStart();
```
rainersigwald: I think I'd call this something clearer. `ExpandGlob`?

rainersigwald: Include `_rootDirectory`, `glob` and if not too difficult within the `TraceEvent` API `excludePatternsForGlobs`. That's critical info if this operation turns out to take a long time.

src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -51,12 +53,14 @@ protected LazyItemOperation(OperationBuilder builder, LazyItemEvaluator<P, I, M,
 
             public virtual void Apply(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
             {
+                MSBuildEventSource.Log.ApplyStart();
```
rainersigwald: Name this something like `ApplyLazyItemOperations` and pass at least `_itemElement.ItemType`. The existing evaluation profiler tracks path and line number, too, which could be very nice.

src/MSBuild/XMake.cs
```diff
@@ -515,14 +516,7 @@ string commandLine
...
             {
-#if (!STANDALONEBUILD)
-                // Enable CodeMarkers for MSBuild.exe
-                CodeMarkers.Instance.InitPerformanceDll(CodeMarkerApp.MSBUILDPERF,  String.Format(CultureInfo.InvariantCulture, @"Software\Microsoft\MSBuild\{0}", MSBuildConstants.CurrentProductVersion));
-#endif
-#if MSBUILDENABLEVSPROFILING 
-                string startMSBuildExe = String.Format(CultureInfo.CurrentCulture, "Running MSBuild.exe with command line {0}", commandLine);
-                DataCollection.CommentMarkProfile(8800, startMSBuildExe);
-#endif
+                MSBuildEventSource.Log.XMakeStart();
```
rainersigwald: Name this `MSBuildExeStart`; no need to propagate the old codename further.

src/Tasks/AssemblyDependency/ReferenceTable.cs
```diff
@@ -1343,9 +1340,7 @@ out userRequestedSpecificFile
         /// </summary>
         internal void RemoveReferencesMarkedForExclusion(bool removeOnlyNoWarning, string subsetName)
         {
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildRARRemoveFromExclusionListBegin, CodeMarkerEvent.perfMSBuildRARRemoveFromExclusionListEnd))
-#endif
+            MSBuildEventSource.Log.RemoveReferencesMarkedForExclusionStart();
```
rainersigwald: Let's set up (and document) a convention to add a prefix for task events. I'd say to call this `RarRemoveReferencesMarkedForExclusion`.

src/Tasks/AssemblyDependency/ReferenceTable.cs
```diff
@@ -1598,9 +1594,7 @@ internal void ComputeClosure
             List<Exception> exceptions
         )
         {
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildRARComputeClosureBegin, CodeMarkerEvent.perfMSBuildRARComputeClosureEnd))
-#endif
+            MSBuildEventSource.Log.ComputeClosureStart();
```
rainersigwald: `RarComputeClosureStart`

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -928,9 +926,7 @@ List<Exception> generalResolutionExceptions
         )
         {
             bool success = true;
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildRARLogResultsBegin, CodeMarkerEvent.perfMSBuildRARLogResultsEnd))
-#endif
+            MSBuildEventSource.Log.LogResultsStart();
```
rainersigwald: prefix

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -1918,9 +1916,7 @@ ReadMachineTypeFromPEHeader readMachineTypeFromPEHeader
         )
         {
             bool success = true;
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildResolveAssemblyReferenceBegin, CodeMarkerEvent.perfMSBuildResolveAssemblyReferenceEnd))
-#endif
+            MSBuildEventSource.Log.ExecuteTaskStart();
```
rainersigwald: Maybe `RarOverallStart`?

src/Tasks/GenerateResource.cs
```diff
@@ -668,9 +666,7 @@ private void GenerateResGenCommandLineWithoutResources(CommandLineBuilderExtensi
         public override bool Execute()
         {
             bool outOfProcExecutionSucceeded = true;
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildGenerateResourceBegin, CodeMarkerEvent.perfMSBuildGenerateResourceEnd))
-#endif
+            MSBuildEventSource.Log.ExecuteGenerateResourceStart();
```
rainersigwald: In keeping with the prefix idea, `GenerateResourceOverallStart`?

src/Tasks/AssemblyDependency/ReferenceTable.cs
Forgind: Isn't that "Remove Assembly References Remove References Marked for Exclusion" when written out? I could do `RarMarkedForExclusion` or `RemoveAssemblyReferencesMarkedForExclusion` if you'd like. (I prefer the latter.)

rainersigwald: I'd punctuate the expansion: "Remove Assembly References**:** Remove References Marked for Exclusion". I prefer that to your names.

Forgind: I see. Makes me want to put Rar:RemoveReferencesMarkedForExclusion or RemoveReferencesMarkedForExclusion(Rar), but then it wouldn't be a valid name.

I can see the value of having Rar on the front, though, since it would make it easy to find what has to do with RAR without looking through the whole list. I'll feel stupid if anyone asks me about it, but I think it's the best solution.

src/Framework/MSBuildEventSource.cs
Forgind: Useful, yes, stale, also yes. Will fix.

src/Tasks/AssemblyDependency/ReferenceTable.cs
rainersigwald: Why would you feel stupid? If we're using a flat namespace here (methods in a single class) prefixing as poor-dev's-namespace makes perfect sense to me.

Forgind: Oh, I agree it makes sense. Not knowing the logic behind it, I would assume I had prepended Rar without actually knowing what it was, just knowing what it's associated with. Not true here, but I would criticize myself for using an acronym without knowing what it stood for and redundancy. With multiple rar methods, though, I do think it's important here for sorting/searching without ambiguity.

src/Framework/MSBuildEventSource.cs
Forgind: Regions make the code more organized. I agree this one is unnecessary, but I'd prefer to keep the one around events if that's ok.

src/Framework/MSBuildEventSource.cs
Forgind: I think it's important to note that it's a singleton, i.e., don't make more than you need.

src/Framework/MSBuildEventSource.cs
Forgind: Although I'll admit I didn't spend too long trying to craft nice comments, and I agree it isn't worth the effort to do so, I do think they add something of value for a novice that might be hard to see for someone already familiar with how MSBuild works.

Forgind: I can remove the comments on stop events, though, since those are uniformly unhelpful.

src/Build/BackEnd/Components/RequestBuilder/RequestBuilder.cs
Forgind: This information is not captured in BuildProject. The path used in CodeMarkers does not seem to exist at this point.

src/Framework/MSBuildEventSource.cs
```diff
@@ -0,0 +1,364 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Build.Eventing
+{
+
+    //
+    // This captures information of how various key methods of building with MSBuild ran.
+    //
+    // /OnlyProviders=*Microsoft-Build
```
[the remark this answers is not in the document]
Forgind: Admittedly, it's more important without the documentation. It could be helpful if you know how to run an eventsource from the command line in general but don't know how to run this particular one. Given that it's not a surprising name, and there's documentation, I don't think it's important anymore.

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -1508,17 +1503,11 @@ public void Save(Encoding saveEncoding)
...
-                DataCollection.CommentMarkProfile(8810, beginProjectSave);
-#endif
-
             Directory.CreateDirectory(DirectoryPath);
-#if (!STANDALONEBUILD)
-            using (new CodeMarkerStartEnd(CodeMarkerEvent.perfMSBuildProjectSaveToFileBegin, CodeMarkerEvent.perfMSBuildProjectSaveToFileEnd))
-#endif
+            if (MSBuildEventSource.Log.IsEnabled())
+            {
+                MSBuildEventSource.Log.SaveStart(_projectFileLocation.LocationString);
```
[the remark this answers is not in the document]
Forgind: This is a case of a property that potentially runs several methods in a row if it hasn't been initialized. 99% of the time, it's cheap; I'm conditioning on the 1%. Most of the others just access a string that already exists.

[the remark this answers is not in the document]
rainersigwald: Worth a comment?

src/MSBuild/XMake.cs
```diff
@@ -515,14 +516,7 @@ string commandLine
...
             {
-#if (!STANDALONEBUILD)
-                // Enable CodeMarkers for MSBuild.exe
-                CodeMarkers.Instance.InitPerformanceDll(CodeMarkerApp.MSBUILDPERF,  String.Format(CultureInfo.InvariantCulture, @"Software\Microsoft\MSBuild\{0}", MSBuildConstants.CurrentProductVersion));
-#endif
-#if MSBUILDENABLEVSPROFILING 
-                string startMSBuildExe = String.Format(CultureInfo.CurrentCulture, "Running MSBuild.exe with command line {0}", commandLine);
-                DataCollection.CommentMarkProfile(8800, startMSBuildExe);
-#endif
+                MSBuildEventSource.Log.MSBuildExeStart();
```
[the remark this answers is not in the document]
Forgind: Since the command line can be passed as a single string or an array of strings, I had to wrap it in #if statements to determine which it was. Do you think that deserves a comment as to why I wrapped the array-of-strings case in IsEnabled?

src/MSBuild/XMake.cs
```diff
@@ -795,7 +801,13 @@ string commandLine
                 // Wait for any pending cancel, so that we get any remaining messages
                 s_cancelComplete.WaitOne();
 
-                MSBuildEventSource.Log.MSBuildExeStop();
+#if FEATURE_GET_COMMANDLINE
+                MSBuildEventSource.Log.MSBuildExeStop(commandLine);
+#else
+                if (MSBuildEventSource.Log.IsEnabled()) {
+                    MSBuildEventSource.Log.MSBuildExeStop(commandLine.ToString());
```
rainersigwald: Is `ToString()` sufficient here? Does it not need to be `string.Join(" ", commandLine)` or similar?

src/MSBuild/XMake.cs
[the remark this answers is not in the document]
rainersigwald: I think the different code is clear enough but would also be fine with a comment.

src/MSBuild/XMake.cs
Forgind: I think both would work, but ToString() will have more punctuation and be harder to read.

PR #4731, opened 2019-09-13
Prevent String::Copy property function from allocating

### LINKED ISSUE #4615: Special case String::Copy to just return the original string instead of calling String.Copy
Calling property functions (such as String.Replace) on item metadata values is tricky.  Generally I've seen people [use String::Copy](https://stackoverflow.com/a/8904902/) to get this to work:

```
$([System.String]::Copy('%(Filename)').Replace('config',''))
```

However, AFAIK the implementation of String.Copy creates a copy of the string data, which is unnecessary overhead.  It's possible to use `MSBuild::ValueOrDefault` to avoid this, but it's a bit more cumbersome:

```
$([MSBuild]::ValueOrDefault('%(MetadataName)', '').Replace('foo', 'bar'))
```

We should consider special-casing the `String::Copy` call to just return the string directly without the overhead of copying.

We could also add a new intrinsic function for this purpose to make it easier to understand or discover how to do this.

BenVillalobos:
Fixes #4615

Does not fix the larger issue #1155, calling instance methods on metadata.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -3686,7 +3686,7 @@ namespace Microsoft.Build.Evaluation
                         {
                             if (TryGetArg(args, out string arg0))
                             {
-                                returnVal = string.Copy(arg0);
+                                returnVal = arg0;
                                 return true;
                             }
                         }


rainersigwald:
Can you change the title before squashing this in to something like

```
Prevent String::Copy property function from allocating
```

(since it doesn't affect normal operation of `String.Copy`)

PR #4764, opened 2019-09-26
Apply DependentUpon Convention Only to .resx Files

### LINKED ISSUE #4740: Non-Resx Embedded Resources get mangled names, possible name collisions
I was able to create a small repro 

the csproj looks like this
```csproj

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <EmbeddedResource Include="MyResource.bmp" WithCulture="false" Type="Non-Resx" />
  </ItemGroup>
</Project>
```

Program.cs
```C#
namespace resourceName
{
    class Program
    {
        {
            Console.WriteLine(MyResource.Get()?.Length ?? -1);
        }
    }
}
```
MyResource.cs
```C#
namespace resourceName
{
    public static class MyResource
    {
        public static Stream Get() => typeof(MyResource).Assembly.GetManifestResourceStream($"{nameof(resourceName)}.{nameof(MyResource)}.bmp");
    }
}
```

Running this code under preview 7 gives output as 
```
C:\git\scratch\resourceName>dotnet.exe run
485514
```

where as running this code under preview 9 gives output as
```
C:\git\corefx>dotnet c:\git\scratch\resourceName\bin\Debug\netcoreapp3.0\resourceName.dll
-1
```

I checked dlls
the resource name include in .dll in previe9 is resourceName.MyResource where as under preview 7 is  resourceName.MyResource.bmp 

I hit this error while building corefx with preview9 for testData.resources file.

This works fine with 5.0 as well.

cc @ericstj @danmosemsft @livarcocc

BenVillalobos:
## Description
Fix #4740, build or runtime failures that arise because non-resx resources get embedded with an unexpected name.

## Customer Impact

Builds fail with `CS1508` or applications fail to extract a resource at runtime. Roadblock to adopting the final 3.0 SDK and to migrating to .NET Core 3.0 for applications with embedded resources (UI applications in particular).

Reported by multiple internal partners (CoreFX, EF) and customers.

### Workaround

Users can explicitly opt out of the new behavior back into preview8 behavior with a property

```xml
<EmbeddedResourceUseDependentUponConvention>false</EmbeddedResourceUseDependentUponConvention>
```

## Regression?

Yes, builds fail in cases that passed before. Introduced with #4597, a new feature to support (resx) resources without project-file impact.

## Testing
Automated tests, including a new one for this specific case. e2e validation in failing scenarios using privates.

## Risk
Low. Targeted fix in a conditioned codepath, so the existing workarounds will continue to work and can mitigate any problems.

----

Taking the suggestion made by @nguerrera to restrict dependentupon convention to .resx files. 
https://github.com/microsoft/msbuild/issues/4740#issuecomment-532962819
> Restricting to resx makes sense because that is the case that would have been persisted with DependentUpon in a classic project. I'm not aware of anything that would have put DependentUpon for non-resx in your project unless you did it manually so I don't think it should have been part of the convention to do so.

Added a check on the `Type` metadata and file extension to prevent dependent upon convention from being used on non-resx files.

Users can manually assign the `Type` metadata when including embedded resources in their projects like so:
`<EmbeddedResource Include="MyResource.txt" WithCulture="false" Type="Non-Resx" />`

If the metadata isn't set, it should automatically be populated by the time the task runs.

In case the `Type` metadata is missing for whatever reason when the task runs, a check on the file's extension will be made to determine whether or not it's a resx file.

### DIFF
--- a/src/Tasks/CreateManifestResourceName.cs
+++ b/src/Tasks/CreateManifestResourceName.cs
@@ -146,8 +146,13 @@ namespace Microsoft.Build.Tasks
                     string fileName = resourceFile.ItemSpec;
                     string dependentUpon = resourceFile.GetMetadata(ItemMetadataNames.dependentUpon);
 
-                    // If opted into convention and no DependentUpon metadata, reference "<filename>.cs" if it exists.
-                    if (UseDependentUponConvention && string.IsNullOrEmpty(dependentUpon))
+                    string fileType = resourceFile.GetMetadata("Type");
+
+                    // If it has a file type metadata is "Resx" or has the file extension ".resx", we know it's a resx file.
+                    bool isResxFile = (!string.IsNullOrEmpty(fileType) && fileType == "Resx") || Path.GetExtension(fileName) == ".resx";
+
+                    // If opted into convention and no DependentUpon metadata and is a resx file, reference "<filename>.cs" if it exists.
+                    if (isResxFile && UseDependentUponConvention && string.IsNullOrEmpty(dependentUpon))
                     {
                         string conventionDependentUpon = Path.ChangeExtension(Path.GetFileName(fileName), SourceFileExtension);
 
--- a/src/Tasks/CreateManifestResourceName.cs
+++ b/src/Tasks/CreateManifestResourceName.cs
@@ -148,10 +148,11 @@ namespace Microsoft.Build.Tasks
 
                     string fileType = resourceFile.GetMetadata("Type");
 
-                    // If it has a file type metadata is "Resx"
+                    // If it has "type" metadata and the value is "Resx"
+                    // This value can be specified by the user, if not it will have been automatically assigned by the SplitResourcesByCulture target.
                     bool isResxFile = (!string.IsNullOrEmpty(fileType) && fileType == "Resx");
 
-                    //fall back onto the extension
+                    // If not, fall back onto the extension.
                     if (string.IsNullOrEmpty(fileType))
                     {
                         isResxFile = Path.GetExtension(fileName) == ".resx";
--- a/src/Tasks/CreateManifestResourceName.cs
+++ b/src/Tasks/CreateManifestResourceName.cs
@@ -148,8 +148,14 @@ namespace Microsoft.Build.Tasks
 
                     string fileType = resourceFile.GetMetadata("Type");
 
-                    // If it has a file type metadata is "Resx" or has the file extension ".resx", we know it's a resx file.
-                    bool isResxFile = (!string.IsNullOrEmpty(fileType) && fileType == "Resx") || Path.GetExtension(fileName) == ".resx";
+                    // If it has a file type metadata is "Resx"
+                    bool isResxFile = (!string.IsNullOrEmpty(fileType) && fileType == "Resx");
+
+                    //fall back onto the extension
+                    if (string.IsNullOrEmpty(fileType))
+                    {
+                        isResxFile = Path.GetExtension(fileName) == ".resx";
+                    }
 
                     // If opted into convention and no DependentUpon metadata and is a resx file, reference "<filename>.cs" if it exists.
                     if (isResxFile && UseDependentUponConvention && string.IsNullOrEmpty(dependentUpon))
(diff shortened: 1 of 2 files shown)


src/Tasks/CreateManifestResourceName.cs
```diff
@@ -146,8 +146,19 @@ CreateFileStream createFileStream
...
                     string dependentUpon = resourceFile.GetMetadata(ItemMetadataNames.dependentUpon);
 
-                    // If opted into convention and no DependentUpon metadata, reference "<filename>.cs" if it exists.
-                    if (UseDependentUponConvention && string.IsNullOrEmpty(dependentUpon))
+                    string fileType = resourceFile.GetMetadata("Type");
+
+                    // If it has a file type metadata is "Resx"
+                    bool isResxFile = (!string.IsNullOrEmpty(fileType) && fileType == "Resx");
+
+                    //fall back onto the extension
```
nguerrera: I'm not familiar with when Type=Resx is set, is that up to the user or to targets that precede this do it?

rainersigwald: Yes, please document the current state at least in the PR and maybe in the comment here.

BenVillalobos: Users can manually set the `Type` metadata to `Resx` or `Non-Resx`. If that metadata isn't set, it is automatically populated by the time the task runs. Currently trying to track down when it gets set.

BenVillalobos: It's set through the `SplitResourcesByCulture` target [here](https://github.com/microsoft/msbuild/blob/master/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2940)

```xml
      <EmbeddedResource Include="@(_MixedResourceWithNoCulture);@(_MixedResourceWithCulture)" Condition="'%(Extension)'=='.resx' or '%(Extension)'=='.restext'">
        <Type Condition="'%(_MixedResourceWithNoCulture.Type)'=='' and '%(_MixedResourceWithCulture.Type)'==''">Resx</Type>
      </EmbeddedResource>
      <EmbeddedResource Include="@(_MixedResourceWithNoCulture);@(_MixedResourceWithCulture)" Condition="'%(Extension)'!='.resx' and '%(Extension)'!='.restext'">
        <Type Condition="'%(_MixedResourceWithNoCulture.Type)'=='' and '%(_MixedResourceWithCulture.Type)'==''">Non-Resx</Type>
      </EmbeddedResource>
```

nguerrera: My follow up here is do we need this fallback to extension, then? It seems like we should expect Type to be set?

rainersigwald: Yeah, I think I now lean toward "assume it's set and if it's not, no convention" as well. I'll push that.

nguerrera:
Tests for the convention are failing. I assume they are test defects at this point?

rainersigwald:
They are failing because the tests don't set the `Type` metadata, yes. Unfortunately, that metadata isn't documented in the [`CreateCSharpManifestResourceName` docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/createcsharpmanifestresourcename-task), so I'm wavering in my commitment to relying on it. Are our tests representative of real-world calls, or special?

I am seeing some third-party calls to the task: https://github.com/search?l=XML&q=CreateCSharpManifestResourceName&type=Code

rainersigwald:
Two relevant offline conversations yesterday:

* @nguerrera is now willing to take the double-checked extension as a belt-and-suspenders approach to direct callers of the task that don't use the usual common.targets flow to assign `Type` for inputs.
* Bar check for 16.3.x passed, so we need a build.

I'm going to merge.

BenVillalobos:
Now that the rules for manifest resource name selection have changed, refer to this [dotnet docs issue](https://github.com/dotnet/docs/issues/16964) for a detailed explanation on how this selection works now.

PR #4766, opened 2019-09-27
Remove Dev10 quirks in targets

### LINKED ISSUE #1890: Remove Dev10 quirks in targets
I'm pretty sure that this:

```xml
<!--
        QUIRKING FOR DEV10

        In the 4.0 version of the targets, we built the targets specified in the Targets metadata in addition to
        GetTargetPath when building in the IDE.  In 4.5, we changed this to just GetTargetPath because it was
        causing performance trouble with certain systems that specified targets that did a significant amount of
        work in the Targets metadata, expecting them to only build when doing a real build.

        However, that change broke C++ unit testing in Dev10 + 4.5 scenarios, because they required use of the
        Targets metadata in order to get design time builds to work properly.  Thus, we need to make sure we
        restore the Dev10 behavior when building on Dev10.
        -->

    <MSBuild
        Projects="@(_MSBuildProjectReferenceExistent)"
        Targets="%(_MSBuildProjectReferenceExistent.Targets);GetTargetPath"
        Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); %(_MSBuildProjectReferenceExistent.SetTargetFramework)"
        Condition="'%(_MSBuildProjectReferenceExistent.BuildReference)' == 'true' and '@(ProjectReferenceWithConfiguration)' != '' and ('$(BuildingInsideVisualStudio)' == 'true' or '$(BuildProjectReferences)' != 'true') and '$(VisualStudioVersion)' == '10.0' and '@(_MSBuildProjectReferenceExistent)' != ''"
        RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">

      <Output TaskParameter="TargetOutputs" ItemName="_ResolvedProjectReferencePaths" Condition="'%(_MSBuildProjectReferenceExistent.ReferenceOutputAssembly)'=='true'"/>
      <Output TaskParameter="TargetOutputs" ItemName="%(_MSBuildProjectReferenceExistent.OutputItemType)" Condition="'%(_MSBuildProjectReferenceExistent.OutputItemType)' != ''"/>

    </MSBuild>
```

Can now be safely removed because MSBuild no longer updates in-place, so any future version of MSBuild won't be loaded in dev10. Plus--is dev10 even supported any more?

CodeTherapist:
closes #1890

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1800,33 +1800,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </MSBuild>
 
     <!--
-        QUIRKING FOR DEV10
-
-        In the 4.0 version of the targets, we built the targets specified in the Targets metadata in addition to
-        GetTargetPath when building in the IDE.  In 4.5, we changed this to just GetTargetPath because it was
-        causing performance trouble with certain systems that specified targets that did a significant amount of
-        work in the Targets metadata, expecting them to only build when doing a real build.
-
-        However, that change broke C++ unit testing in Dev10 + 4.5 scenarios, because they required use of the
-        Targets metadata in order to get design time builds to work properly.  Thus, we need to make sure we
-        restore the Dev10 behavior when building on Dev10.
-        -->
-
-    <MSBuild
-        Projects="@(_MSBuildProjectReferenceExistent)"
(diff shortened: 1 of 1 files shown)


CodeTherapist:
Thanks reviewing my PR 👍 
What kind of references, same kind of as I removed already?

rainersigwald:
Yes, I think there are other places that have conditions like `'$(VisualStudioVersion)' == '10.0'` that can now be removed. But this is a good start!

PR #4776, opened 2019-10-02
 Fix #4648 

### LINKED ISSUE #4648: Remove warning on about-to-be-deleted corrupt resgen output
context: https://github.com/microsoft/msbuild/pull/4644#discussion_r316363861

Currently, there are some circumstances that will make the `GenerateResource` task run `RemoveCorruptedFile`

https://github.com/microsoft/msbuild/blob/3885a205f40944174a3e14396547f13db1196b82/src/Tasks/GenerateResource.cs#L2896-L2912

That method unconditionally emits a warning that the file may be corrupted. But it then immediately deletes it (or emits another error that it couldn't delete it).

The warning seems unnecessary. We should bump it down to info and make it more descriptive, maybe something like

> Attempting to remove possibly-corrupt output file {}.

treymerkley:


src/Tasks/GenerateResource.cs
```diff
@@ -2895,14 +2895,14 @@ private string EnsurePathIsShortEnough(string currentOutputFile, string currentO
         /// <param name="filename">Full path to file to delete</param>
         private void RemoveCorruptedFile(string filename)
         {
-            _logger.LogWarningWithCodeFromResources("GenerateResource.CorruptOutput", FileUtilities.GetFullPathNoThrow(filename));
+            _logger.LogMessageFromResources("GenerateResource.CorruptOutput. Attempting to remove possibly-corrupt output file: ", FileUtilities.GetFullPathNoThrow(filename));
```
rainersigwald: This is changing the name of the resource, and you shouldn't need to do that.
```suggestion
            _logger.LogMessageFromResources("GenerateResource.CorruptOutput", FileUtilities.GetFullPathNoThrow(filename));
```

src/Tasks/GenerateResource.cs
```diff
@@ -2895,14 +2895,14 @@ private string EnsurePathIsShortEnough(string currentOutputFile, string currentO
...
             try
             {
                 File.Delete(filename);
             }
             catch (Exception deleteException)
             {
-                _logger.LogWarningWithCodeFromResources("GenerateResource.DeleteCorruptOutputFailed", FileUtilities.GetFullPathNoThrow(filename), deleteException.Message);
+                _logger.LogInfoWithCodeFromResources("GenerateResource.DeleteCorruptOutputFailed. Failed to remove possibly-corrupt output file: ", FileUtilities.GetFullPathNoThrow(filename), deleteException.Message);
```
rainersigwald: I think you can revert this entirely; if we _fail_ to delete the file we should still log a warning (that should be rare).

src/Tasks/Resources/Strings.resx
```diff
@@ -911,7 +911,6 @@
   </data>
   <data name="GenerateResource.CorruptOutput">
     <value>MSB3555: Output file "{0}" is possibly corrupt.</value>
```
rainersigwald:
```suggestion
     <value>Output file "{0}" is possibly corrupt and will be deleted and recreated.</value>
```

src/Tasks/Resources/Strings.resx
```diff
@@ -931,7 +930,6 @@
   </data>
   <data name="GenerateResource.DeleteCorruptOutputFailed">
     <value>MSB3560: Could not delete the possibly corrupt output file "{0}". {1}</value>
-    <comment>{StrBegin="MSB3560: "}</comment>
```
rainersigwald: If we leave this as a warning we can revert this change.

src/Tasks/Resources/Strings.resx
rainersigwald: Making this change will then require running a build to propagate it to the localized `.xlf`s.

src/Tasks/GenerateResource.cs
```diff
@@ -56,7 +56,7 @@ public sealed partial class GenerateResource : TaskExtension
 
 
 #region Fields
-        
+
```
rainersigwald: Would you mind reverting the whitespace changes in this file? I think your editor (correctly) respected our `.editorconfig` but we haven't actually fixed that across the tree yet . . .

src/Tasks/GenerateResource.cs
treymerkley: I added that in because of this:

> The warning seems unnecessary. We should bump it down to info and make it more descriptive, maybe something like
>
>   ` Attempting to remove possibly-corrupt output file {}.`

How should I work that in then?

rainersigwald: Change the contents of the string (in the `.resx` file and localizations in the various `.xlf`), not its name.

As an analogy, you're doing this:

```diff
- var generateResource_CorruptOutput = "MSB3555: Output file "{0}" is possibly corrupt.";
+ var generateResource_CorruptOutput_attempting_to_remove = "MSB3555: Output file "{0}" is possibly corrupt.";
```

instead of

```diff
- var generateResource_CorruptOutput = "MSB3555: Output file "{0}" is possibly corrupt.";
+ var generateResource_CorruptOutput = "Output file "{0}" is possibly corrupt.";
```

src/Tasks/GenerateResource.cs
treymerkley: Oh sorry, I have my text editor set to always delete trailing whitespace. Is that something that I shouldn't have?

treymerkley: I just fixed it, I'll push it in just a second.

rainersigwald: I have no problem with fixing trailing whitespace, but I'd like to see it separated from the contents of this commit so it's easier to understand changes in the future.

treymerkley: wait nope, that is the issue, I think. I'll fix it after everything else in something like notepad or something.

treymerkley: sure, I'll revert it and then push it as a new commit.

src/Tasks/GenerateResource.cs
treymerkley: okay, *should* be fixed

src/Tasks/GenerateResource.cs
treymerkley: should be fixed.

src/Tasks/Resources/Strings.resx
treymerkley: okay, so just add back in the line `<comment>{StrBegin="MSB3560: "}</comment>` below 932?

rainersigwald: Yes.

treymerkley: *should* be fixed

src/Tasks/Resources/Strings.resx
```diff
@@ -910,8 +910,7 @@
     <comment>{StrBegin="MSB3554: "}</comment>
   </data>
   <data name="GenerateResource.CorruptOutput">
-    <value>MSB3555: Output file "{0}" is possibly corrupt.</value>
-    <comment>{StrBegin="MSB3555: "}</comment>
+     <value>Output file "{0}" is possibly corrupt and will be deleted and recreated.</value>
```
rainersigwald:
```suggestion
    <value>Output file "{0}" is possibly corrupt and will be deleted and recreated.</value>
```

PR #4796, opened 2019-10-09
[vs16.3] Arcade update for servicing

rainersigwald:
It appears that official builds were succeeding after #4733 only because of a coincident service outage, so updating to the latest arcade bits for publishing is likely required to support servicing 16.3.

Most of the breaking changes in #4783 are a result of the SDK update rather than Arcade changes, so this PR reverts the SDK while updating Arcade.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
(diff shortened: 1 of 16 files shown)


rainersigwald:
@mmitche This passed PR builds and [an official build (internal link)](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3127298) but that didn't try to publish since it didn't have a default channel.

PR #4797, opened 2019-10-09
Fix CachingSdkResolverService concurrency.

### LINKED ISSUE #4798: SdkResolvers can be called multiple times for the same SDK
### Steps to reproduce

Build a set of projects in an SLN and reference multiple MSBuild project SDKs

Command line
```
msbuild /bl
```
### Expected  behavior
Each SDK is resolved exactly once for the entire build and projects that use the same SDK should get a cached result.

### Actual behavior
Some SDKs are resolved multiple times

jeffkl:
The documentation for ConcurrentDictionary<T>.GetOrAdd() states:

> If you call GetOrAdd simultaneously on different threads, valueFactory may be called multiple times, but only one key/value pair will be added to the dictionary.

In all of my testing, a particular SDK was never resolved more than once so I didn't think we had a problem.  However, on an NGen'd MSBuild, I noticed that an SDK was being resolved multiple times.  This is because the `GetOrAdd()` method does not guarantee that a method is only called once.  Instead, it guarantees that if multiple threads call the method simultaneously they will all receive the same return value.

So the fix is to return a `Lazy<T>` instead.  This means that multiple threads could new up multiple instances of Lazy<T> but all threads will receive the same Lazy<T>.  They will all wait for the same result when accessing `Value` so the SDK will only be resolved once and the value is then cached for subsequent calls.

Fixes #4798

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/CachingSdkResolverService.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/CachingSdkResolverService.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Concurrent;
 using Microsoft.Build.BackEnd.Logging;
 using Microsoft.Build.Collections;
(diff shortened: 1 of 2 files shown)


src/Build/BackEnd/Components/SdkResolution/CachingSdkResolverService.cs
```diff
@@ -53,7 +54,8 @@ public override SdkResult ResolveSdk(int submissionId, SdkReference sdk, Logging
                  */
                 result = cached.GetOrAdd(
                     sdk.Name,
-                    key => base.ResolveSdk(submissionId, sdk, loggingContext, sdkReferenceLocation, solutionPath, projectPath, interactive));
+                    key => new Lazy<SdkResult>(() => base.ResolveSdk(submissionId, sdk, loggingContext, sdkReferenceLocation, solutionPath, projectPath, interactive)))
+                    .Value;
```
rainersigwald: I think I'd prefer a named variable for the lazy; this construction confused me for a second until I realized that there was a `)` between where I though the `.Value` was and where it actually was.

jeffkl: I had it like that at first and it felt too verbose.  But the tribe has spoken...

I also added/updated the comments

jeffkl:
I was able to devise a test which I ran against master and it fails. So this definitely fixes it!

PR #4812, opened 2019-10-14
Move Microsoft.Build.Experimental.Graph to Microsoft.Build.Graph

jeffkl:
Static graph is ready for production use and big partner teams are taking a dependency on it.


AndyGerlicher:
Since this is somewhat of a massive breaking change for anyone using it, do we want to do anything fancy here? Type forwarder? Copy/paste and add depreciated for a release or two then delete? I know the reason we added "Experimental" was to not have to maintain that level of comparability, but just want to make sure we're explicit about the impact.

@dfederm

dfederm:
At least for QuickBuild, I think we'd be ok as long as we take a pre-release package containing these changes *before* anyone actually started using a version of MSBuild with these changes to build. 

Specifically this is because we load up the newer of the assemblies ("ours" that comes with QuickBuild vs "theirs" they use to build). So as long as `Max(ours, theirs)` is > whatever version ends up with this change, it's fine.

cdmihai:
My personal bar for this change was to have one C# and one C++ team use static graph for either local / CI builds, but I guess we can just deal with any potential big / breaking change (in a back compat way) when we hit it.
Would we need to do anything else with documentation, beyond the [current spec](https://github.com/microsoft/msbuild/blob/master/documentation/specs/static-graph.md)?

PR #4814, opened 2019-10-14
Added overridable method after proc.Start()

### LINKED ISSUE #4611: ToolTask class should have an overridable method which is called right after tool process is started
### Steps to reproduce
There is a long standing issue when CL and Link tasks hang sometimes waiting on a pipe they create to get unicode output from the tools (the problem described here: https://docs.microsoft.com/en-us/windows/win32/ipc/pipe-handle-inheritance). 
To avoid this problem, the write handle of the pipe must be closed right after the tool process is created.

Currently, there is no way we can do it in a task derived from 
https://github.com/microsoft/msbuild/blob/master/src/Utilities/ToolTask.cs 

Please add an overridable method in ToolTask class and call it in Execute() after process.Start().

Forgind:
Fixes #4611

### DIFF
--- a/src/Utilities/ToolTask.cs
+++ b/src/Utilities/ToolTask.cs
@@ -699,6 +699,9 @@ namespace Microsoft.Build.Utilities
                 // Start the process
                 proc.Start();
 
+                // Execute process
+                this.ProcessStarted();
+
                 // Close the input stream. This is done to prevent commands from
                 // blocking the build waiting for input from the user.
                 if (NativeMethodsShared.IsWindows)
@@ -765,6 +768,14 @@ namespace Microsoft.Build.Utilities
             }
 
             return ExitCode;
+        }
+
+        /// <summary>
+        /// Optional overridable function called after proc.Start() in ExecuteTool.
+        /// </summary>
+        protected virtual void ProcessStarted()
+        {
+            
         }
 
         /// <summary>
--- a/src/Utilities/ToolTask.cs
+++ b/src/Utilities/ToolTask.cs
@@ -313,6 +313,11 @@ namespace Microsoft.Build.Utilities
         #region Overridable methods
 
         /// <summary>
+        /// Overridable function called after <see cref="Process.Start()"/> in <see cref="ExecuteTool"/>
+        /// </summary>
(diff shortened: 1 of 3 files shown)


src/Utilities/ToolTask.cs
```diff
@@ -699,6 +699,9 @@ string commandLineCommands
                 // Start the process
                 proc.Start();
 
+                // Execute process
```
rainersigwald:
```suggestion
                // Call user-provided hook for "do something right after the process starts"
```

src/Utilities/ToolTask.cs
```diff
@@ -699,6 +699,9 @@ string commandLineCommands
...
+                this.ProcessStarted();
```
rainersigwald: Should this be after the `StandardInput.Dispose()` invocation below? My instinct is to say "yes", because the overrider might do something similar and we'd rather it throw in their code than ours.

src/Utilities/ToolTask.cs
```diff
@@ -767,6 +770,14 @@ string commandLineCommands
             return ExitCode;
         }
 
+        /// <summary>
+        /// Optional overridable function called after proc.Start() in ExecuteTool.
```
rainersigwald: Some tweaks:

* Match the other comments that don't say optional
* drop reference to local variable name `proc`
* Explicit cross-references to make things more understandable

```suggestion
        /// Overridable method called after <see cref="Process.Start()"/> in <see cref="ExecuteTool" />.
```

src/Utilities/ToolTask.cs
```diff
@@ -767,6 +770,14 @@ string commandLineCommands
...
+        /// </summary>
+        protected virtual void ProcessStarted()
+        {
+            
+        }
```
rainersigwald: The usual style for empty default implementations is `{}` on a single line.

src/Utilities/ToolTask.cs
rainersigwald: @olgaark do you have an opinion on this?

jimspr: I think either should be fine for our specific purposes.

PR #4821, opened 2019-10-15
Changing source file extension to protected in CreateManifestResourceName

BenVillalobos:
Fix for #4759 

Allow users to override the default source file extension

### DIFF
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -195,6 +195,7 @@ namespace Microsoft.Build.Tasks
     public partial class CreateCSharpManifestResourceName : Microsoft.Build.Tasks.CreateManifestResourceName
     {
         public CreateCSharpManifestResourceName() { }
+        protected override string SourceFileExtension { get { throw null; } }
         protected override string CreateManifestName(string fileName, string linkFileName, string rootNamespace, string dependentUponFileName, System.IO.Stream binaryStream) { throw null; }
         protected override bool IsSourceFile(string fileName) { throw null; }
     }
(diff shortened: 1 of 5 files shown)


Forgind:
That error is actually one I ran into recently. Go to your branch, build locally, and push again, and it should be fine.

BenVillalobos:
@Forgind got it, thanks!

rainersigwald:
> I think changing to protected could cause a regression bug, and public is safer

Can you elaborate on this? Under what circumstances would it cause a regression?

BenVillalobos:
I may be mistaken, but this would only be a regression if we referenced SourceFileExtension from a class that wasn't inherited from CreateManifestResource, which we haven't done from the time it was added.

Forgind:
I was thinking along the same lines as Ben, so since we haven't referenced SourceFileExtension, it's probably not a problem in 99% of cases. Someone theoretically could have taken MSBuild, modified it to add something that uses SourceFileExtension from within the same assembly, perhaps making a patch they apply whenever they update MSBuild before building it, but if they're going to that effort, they can just revert the change anyway, so I don't think it's an issue.

rainersigwald:
As a rule, we don't worry about breaking forks of our code, so I don't think that's enough to stop us.

By my read of the [corefx breaking-changes guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members), this is ok, by

> ### Members
> &#10003; **Allowed**
> * Increasing the visibility of a member that is not `virtual`

Forgind:
I agree that it's ok but not because of that guideline. This is decreasing the visibility (in no cases based on Ben's note) of a method, so you could also classify it under:

"Decreasing the visibility of a protected member when there are accessible (public or protected) constructors and the type is not sealed"

which is not allowed. There are no guidelines on internal methods, and internal is somewhat similar to protected, but assuming they build our binaries without modifying them, we're fine, hence my approval.

rainersigwald:
> This is _decreasing_ the visibility (in no cases based on Ben's note) of a method

Isn't `internal` -> `protected` an increase?

Forgind:
My understanding is that protected allows subclasses to access it, whereas internal allows other classes in the binary to access it, so they're overlapping circles of visibility, but internal is not a subset of protected, nor is protected a subset of internal.

rainersigwald:
Sure, but only the public visibility is relevant here: we can fix any breaks we cause in our own code (that can see `internal` members). From the outside this seems equivalent to `private` -> `protected`, which is an increase.

PR #4829, opened 2019-10-17
Permissive MemoryStream typename

### LINKED ISSUE #4773: GenerateResource w/ ResXFileRef System.IO.MemoryStream, mscorlib, Version=2.0.0.0 fails at runtime
Originally reported here https://github.com/dotnet/coreclr/issues/26472#issuecomment-536176848.  Thanks @koszeggy 

### Steps to reproduce
Include resources that use ResXFileRef for a MemoryStream with 2.0.0.0 assembly identity for mscorlib.
Build.
Access resource at runtime.

### Expected  behavior
Resource is returned as UnmanagedMemoryStream.

### Actual behavior
Resource is treated as an Activator resource and we attempt to call the a MemoryStream constructor passing in the backing stream, which fails because MemoryStream doesn't have a constructor that takes a stream.

### Environment data
dotnet.exe 3.0.100

Bug is here: https://github.com/microsoft/msbuild/commit/edb3be367107d5976dd5a9b30ed067b24f982668#diff-5a51b7df470ec16b6d58bdf7234efefeR85

Here's  a repro: https://github.com/ericstj/repros/tree/master/testMemStreamResXFileRef

/cc @rainersigwald

rainersigwald:
Fixes #4773 by treating string and MemoryStream types as a prefix match on the mscorlib identity,
hoping to more closely match `Type.GetType`'s behavior.

### DIFF
--- a/src/Tasks.UnitTests/ResourceHandling/MSBuildResXReader_Tests.cs
+++ b/src/Tasks.UnitTests/ResourceHandling/MSBuildResXReader_Tests.cs
@@ -243,6 +243,39 @@ $@"  <data name='Image1' type='System.Resources.ResXFileRef, System.Windows.Form
             resource.TypeAssemblyQualifiedName.ShouldBe("System.Drawing.Bitmap, System.Drawing, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a");
         }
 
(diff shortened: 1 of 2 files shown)


rainersigwald:
@ericstj This was what you meant by using a prefix match, right?

PR #4830, opened 2019-10-18
Restore write-resource-to-text tests

### LINKED ISSUE #308: GenerateResource does not fail when converting binary .resx resources to .txt resources

rainersigwald:
Closes #308 by re-enabling the tests.

### DIFF
--- a/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
+++ b/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
@@ -1773,6 +1773,7 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
         ///  Cause failures in ResourceReader
         /// </summary>
         [Fact]
+        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp, ".NET Core MSBuild doesn't try to read binary input resources")]
         public void FailedResourceReader()
         {
             GenerateResource t = Utilities.CreateTask(_output);
--- a/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
+++ b/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
@@ -1800,6 +1800,28 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests.InProc
                 File.Delete(item.ItemSpec);
         }
 
+        [Theory]
+        [InlineData(".resources")]
+        [InlineData(".dll")]
+        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework, "This error is .NET Core only")]
+        public void ResourceReaderRejectsNonCoreCompatFormats(string inputExtension)
+        {
+            using var env = TestEnvironment.Create(_output);
+
+            GenerateResource t = Utilities.CreateTask(_output);
+            t.StateFile = new TaskItem(env.GetTempFile(".cache").Path);
+
+            // file contents aren't required since the extension is checked first
+            var resourcesFile = env.CreateFile(inputExtension).Path;
+
+            t.Sources = new ITaskItem[] { new TaskItem(resourcesFile) };
+            t.OutputResources = new ITaskItem[] { new TaskItem(env.GetTempFile(".resources").Path) };
+
+            t.Execute().ShouldBeFalse();
+
+            Utilities.AssertLogContains(t, "MSB3824");
+        }
+
         /// <summary>
         ///  Invalid STR Class name
         /// </summary>
--- a/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
+++ b/src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
@@ -3690,7 +3690,7 @@ namespace Microsoft.Build.UnitTests.GenerateResource_Tests
         /// <param name="useType">Indicates whether to include an enum to test type-specific resource encoding with assembly references</param>
         /// <param name="linkedBitmap">The name of a linked-in bitmap.  use 'null' for no bitmap.</param>
         /// <returns>The name of the resx file</returns>
-        public static string WriteTestResX(bool useType, string linkedBitmap, string extraToken, string resxFileToWrite = null)
+        public static string WriteTestResX(bool useType, string linkedBitmap, string extraToken, string resxFileToWrite = null, TestEnvironment env = null)
         {
             return WriteTestResX(useType, linkedBitmap, extraToken, useInvalidType: false, resxFileToWrite:resxFileToWrite);
         }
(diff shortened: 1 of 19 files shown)


rainersigwald:
I thought this worked on my machine but I was testing desktop framework. Sigh. Pulling for now.

src/Tasks/GenerateResource.cs
```diff
@@ -2988,7 +2994,7 @@ private void ReadResources(String filename, bool shouldUseSourcePath, String out
 #if FEATURE_ASSEMBLY_LOADFROM
                 ReadAssemblyResources(filename, outFileOrDir);
 #else
-                _logger.LogError("Reading resources from Assembly not supported on .NET Core MSBuild");
+                throw new InputFormatNotSupportedException("Reading resources from Assembly not supported on .NET Core MSBuild");
```
Forgind: Why is it advantageous to throw instead of just logging an error? Are you trying to reduce duplicate output?

src/Tasks/GenerateResource.cs
```diff
@@ -3387,8 +3393,13 @@ private bool HaveSystemResourcesExtensionsReference
 
                 PopulateAssemblyNames();
 
-                foreach (var assemblyName in _assemblyNames)
+                foreach (AssemblyNameExtension assemblyName in _assemblyNames)
                 {
+                    if (assemblyName == null)
```
Forgind: Why would they be null?

src/Tasks/ResourceHandling/InputFormatNotSupportedException.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Build.Tasks.ResourceHandling
+{
+    [Serializable]
+    internal class InputFormatNotSupportedException : Exception
```
Forgind: Is there a reason you can't use `FormatException` or `ArgumentException` instead of making your own (empty) new type of exception?

src/Tasks/Resources/xlf/Strings.es.xlf
```diff
@@ -979,6 +979,11 @@
         <target state="translated">MSB3190: ClickOnce no admite el nivel de ejecución de solicitudes '{0}'.</target>
         <note>{StrBegin="MSB3190: "}</note>
       </trans-unit>
+      <trans-unit id="GenerateResource.CoreSupportsLimitedScenarios">
+        <source>MSB3824: In order to build with .NET Core, resource inputs must be in .txt or .resx format.</source>
```
Forgind: I'm guessing this is supposed to be a localized version. When does that happen?

rainersigwald: Correct.  The loc team periodically (at least once per minor release) does a pass on strings and will submit a PR with updated localizations (#4679 was a recent one). Until that's done the English copy of the string is used.

Docs are here: https://github.com/microsoft/msbuild/blob/master/documentation/wiki/Localization.md

src/Tasks/GenerateResource.cs
rainersigwald: Logging an error here is nonfatal, so the experience with going through this codepath was:

1. This error.
2. Some other odd output when trying to move on.
3. Another fatal error that didn't directly indicate the problem.

Since there were already exception-based error paths in `ProcessFile` I opted to add another rather than change everything to error returns.

src/Tasks/GenerateResource.cs
rainersigwald: That's what happens if an assembly name can't be extracted from a reference for some reason in `PopulateAssemblyNames()`. The other codepaths were robust to this situation but this newly-added one wasn't (until running a test revealed the error).

src/Tasks/ResourceHandling/InputFormatNotSupportedException.cs
rainersigwald: Having a custom exception type makes detecting this error condition easy using a catch-specific-type block.

Forgind: Tracing how it's currently caught, I believe a FormatException will still have the exact same effect.

Also, I don't see anywhere that you try to catch it; do you have plans for adding something like that in the future? If not, you're breaking something that could otherwise be caught by an ArgumentException, failing more cleanly.

rainersigwald: > I believe a FormatException will still have the exact same effect.

Since it's a standard exception type, other code could insert a `FormatException` that would then be inadvertently caught and handled as this specific problem, rather than handled at another layer.

> Also, I don't see anywhere that you try to catch it

https://github.com/microsoft/msbuild/pull/4830/files#diff-1d2aacce8c88cda3235c1659ea67e164R2606-R2611

> If not, you're breaking something that could otherwise be caught by an ArgumentException, failing more cleanly.

This is harder to distinguish since other places could throw `ArgumentException` inside the big `ProcessFile` catch block.

src/Tasks.UnitTests/ResourceHandling/GenerateResource_Tests.cs
```diff
@@ -3711,16 +3724,27 @@ public static string WriteTestResX(bool useType, string linkedBitmap, string ext
...
-            if (string.IsNullOrEmpty(resgenFile))
+
+            string contents = GetTestResXContent(useType, linkedBitmap, extraToken, useInvalidType);
+
+            if (env == null)
+            {
+                if (string.IsNullOrEmpty(resgenFile))
+                {
+                        resgenFile = GetTempFileName(".resx");
+                        File.Delete(resgenFile);
```
BenVillalobos: (nit) Is this line necessary? According to the [File.WriteAllText documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.writealltext?view=netframework-4.8), even if the file did exist it would be overwritten

rainersigwald: It's preexisting but no, I don't think it is. I'll remove.

PR #4839, opened 2019-10-22
Enable Arcade Pipelines logging in Windows builds

rainersigwald:
Example output: https://dev.azure.com/dnceng/public/_build/results?buildId=432144&view=results

Not sure why that didn't push checks annotations except from the *nix legs that were already working, but it appears to annotate the errors correctly in AzDO.

### DIFF
--- a/eng/cibuild_bootstrapped_msbuild.ps1
+++ b/eng/cibuild_bootstrapped_msbuild.ps1
@@ -7,6 +7,10 @@ Param(
   [Parameter(ValueFromRemainingArguments=$true)][String[]]$properties
 )
 
+# Ensure that static state in tools is aware that this is
+# a CI scenario
+$ci = $true
+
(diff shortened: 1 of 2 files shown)


eng/cibuild_bootstrapped_msbuild.ps1
```diff
@@ -103,7 +111,7 @@ try {
     Move-Item -Path $ArtifactsDir -Destination $Stage1Dir -Force
   }
 
-  $buildTool = @{ Path = $buildToolPath; Command = $buildToolCommand }
+  $buildTool = @{ Path = $buildToolPath; Command = $buildToolCommand; Tool = $buildToolTool; Framework = $buildToolFramework }
```
Forgind: Can you just use `Tool = $msbuildEngine` instead?

PR #4849, opened 2019-10-24
Fix error message

### LINKED ISSUE #4332: Confusing error when using COMReference on Core
_From @fubar-coder on Tuesday, 23 April 2019 07:16:19_

I get an error which seems to be caused by the following `COMReference` project item:

```xml
  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' and '$(OS)' == 'Windows_NT' ">
    <COMReference Include="ADOX.dll">
      <Guid>00000600-0000-0010-8000-00aa006d2ea4</Guid>
      <VersionMajor>6</VersionMajor>
      <VersionMinor>0</VersionMinor>
      <WrapperTool>tlbimp</WrapperTool>
      <Lcid>0</Lcid>
      <Isolated>false</Isolated>
    </COMReference>
  </ItemGroup>
```

The error message roughly translates to something along the lines: `The "ResolveComReference" task couldn't be executed, because MSBuild couldn't create or connect to a task host with runtime "CLR4" and architecture "x86".`

I also attached the build log.

[build-log.zip](https://github.com/dotnet/core-sdk/files/3106259/build-log.zip)

_Copied from original issue: dotnet/core-sdk#1690_

Forgind:
If the platform target is explicitly set to x86, it sets the MSBuild architecture to the same. It then passes over the real problem when using dotnet build and returns a confusing error message instead. Fixes #4332.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -504,10 +504,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <PlatformTargetAsMSBuildArchitectureExplicitlySet Condition="'$(PlatformTargetAsMSBuildArchitecture)' != ''">true</PlatformTargetAsMSBuildArchitectureExplicitlySet>
     <PlatformTargetAsMSBuildArchitectureExplicitlySet Condition="'$(PlatformTargetAsMSBuildArchitecture)' == ''">false</PlatformTargetAsMSBuildArchitectureExplicitlySet>
 
-    <PlatformTargetAsMSBuildArchitecture Condition="('$(PlatformTarget)' == 'x86' and '$(MSBuildExtensionsPath86)' != '') or ('$(PlatformTarget)' == 'x64' and '$(MSBuildExtensionsPath64)' != '')">$(PlatformTarget)</PlatformTargetAsMSBuildArchitecture>
+    <PlatformTargetAsMSBuildArchitecture Condition="'$(PlatformTarget)' == 'x86' or ('$(PlatformTarget)' == 'x64' and '$(MSBuildExtensionsPath64)' != '')">$(PlatformTarget)</PlatformTargetAsMSBuildArchitecture>
     <PlatformTargetAsMSBuildArchitecture Condition="'$(PlatformTarget)' == 'arm'">x86</PlatformTargetAsMSBuildArchitecture>
     <PlatformTargetAsMSBuildArchitecture Condition="('$(PlatformTarget)' == 'ia64' and '$(MSBuildExtensionsPath64)' != '')">x64</PlatformTargetAsMSBuildArchitecture>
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -507,7 +507,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <PlatformTargetAsMSBuildArchitecture Condition="'$(PlatformTarget)' == 'x86' or ('$(PlatformTarget)' == 'x64' and '$(MSBuildExtensionsPath64)' != '')">$(PlatformTarget)</PlatformTargetAsMSBuildArchitecture>
     <PlatformTargetAsMSBuildArchitecture Condition="'$(PlatformTarget)' == 'arm'">x86</PlatformTargetAsMSBuildArchitecture>
     <PlatformTargetAsMSBuildArchitecture Condition="('$(PlatformTarget)' == 'ia64' and '$(MSBuildExtensionsPath64)' != '')">x64</PlatformTargetAsMSBuildArchitecture>
-    <PlatformTargetAsMSBuildArchitecture Condition="'$(PlatformTargetAsMSBuildArchitecture)' == ''">CurrentArchitecture</PlatformTargetAsMSBuildArchitecture>
+    <PlatformTargetAsMSBuildArchitecture Condition="'$(MSBuildRuntimeType)' == 'Core' or '$(PlatformTargetAsMSBuildArchitecture)' == ''">CurrentArchitecture</PlatformTargetAsMSBuildArchitecture>
```
rainersigwald: Would it make sense to elevate this condition to the whole PropertyGroup? I feel like "running on Core which doesn't support taskhosts" and `DisableOutOfProcTaskHost` are pretty close. It looks like that property doesn't get directly set anywhere, so we can't just condition at its definition.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -495,8 +495,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </PropertyGroup>
 
 
-  <!-- Sensible defaults for the most-commonly-desired MSBuildRuntime and MSBuildArchitecture values -->
-  <PropertyGroup Condition="'$(DisableOutOfProcTaskHost)' == ''">
+  <!-- Sensible defaults for the most-commonly-desired MSBuildRuntime and MSBuildArchitecture values. The Core and Mono runtimes do not currently support custom architectures or runtimes.
+       The change to limit this to the 'Full' runtime originated from the 'ResolveComReferences' target in this file. The team that owns that target explicitly said they are not going to port it to .NET Core, so if you remove the below check, add a similar check in ResolveComReferences. -->
```
rainersigwald:
```suggestion
       If support for out-of-proc task execution is added on other runtimes, make sure each task's logic is checked against the current state of support. -->
```

Issue #4779, opened 2019-10-03
Project-wide whitespace cleanup


treymerkley:
So while working on src/Tasks/GenerateResource.cs I accidentally found that there's a relatively large amount of whitespace in the entire MSBuild project. Can I clean this up? I'll do it programmatically of course, and (once I find one) I'll insert the function I use.

rainersigwald:
Let's limit it to only trailing whitespace, but yeah, I think we'd take a PR to do that.

Areas to be careful of:

* Try not to change whitespace inside string literals
* Is anything in our MSBuild logic whitespace-sensitive? I sure hope not!

treymerkley:
only trailing? I should be able to do that. 

Rather than trying to clean the whole project at once then, what if I went a folder at a time, starting at the root, that way any bugs that arise are immediately noticeable and easier to fix?

rainersigwald:
Seems reasonable to me!

treymerkley:
So I'm just using https://github.com/purcell/whitespace-cleanup-mode and doing it on a file-by-file basis so it's easier to track down *what* broke in the event something is whitespace dependent.

PR #4855, opened 2019-10-28
whitespace cleanup

treymerkley:
Putting more against #4779. I accidentally pushed to my own, master, still need to get Magit down :joy: :joy:


treymerkley:
Sure. I cleaned up a few more files, so I'll push those too. Is there any particular way you want me to batch them together? Just everything in a folder or something?

Issue #4823, opened 2019-10-16
Tracking: .NET Standard support in an internal repo


rainersigwald:
Tracking issue for some internal-partner-support issues that are taking a fair amount of my time this sprint.

PR #4868, opened 2019-11-01
Add Microsoft.Common.props in Current to CoreXT overlay

rainersigwald:
In the VS repo, some projects attempted to import based on $(MSBuildToolsVersion), and got confused in the '15.0 package with the 16.3 engine' situation.

The right long-term fix is to move to 16.x packages, but for now we can just duplicate this file; all other files seemed to be loaded from the right folders automatically.

This unblocks #4823. @duncanmak and I have been using it in private testing and it looks good.


src/Package/MSBuild.Engine.Corext/MsBuild.Engine.Corext.nuspec
```diff
@@ -14,6 +14,7 @@
     <file src="$repoRoot$/src/Package/MsBuild.Engine.Corext/README.md" target=""/>
     <file src="$repoRoot$/THIRDPARTYNOTICES.txt" target="v15.0/bin/THIRDPARTYNOTICES.txt" />
     <file src="$X86BinPath$/Microsoft.Common.props" target="\Extensions\15.0" />
```
cdmihai: Add a comment like `<!-- Extensions/15.0 is deprecated. Remove this in some next version -->` :)

PR #4869, opened 2019-11-01
Enable lost debug-only code

rainersigwald:
At some point we stopped defining _DEBUG for debug builds, so all this conditional code was never being built.

```sh-session
$ rep -r "#if _DEBUG" "#if DEBUG" *.cs
1993 files processed. 9 files changed. 14 replacements.
```

### DIFF
--- a/src/Tasks/UnregisterAssembly.cs
+++ b/src/Tasks/UnregisterAssembly.cs
@@ -4,6 +4,7 @@
 #if FEATURE_APPDOMAIN
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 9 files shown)


rainersigwald:
This won't pass until #4893 is merged.

src/Tasks/UnregisterAssembly.cs
```diff
@@ -4,6 +4,7 @@
 #if FEATURE_APPDOMAIN
 
 using System;
+using System.Diagnostics;
```
BenVillalobos: Is the using necessary for this PR?

rainersigwald: It is: the calls to `[System.Diagnostics.]Debug.Assert` in the `#if`s I changed wouldn't compile without it.

PR #4872, opened 2019-11-01
Allow <Exec> commands to inherit the console code page

### LINKED ISSUE #4870: MSBuild <Exec>/ToolTask resets the console output code page, prohibiting unicode output
### Steps to reproduce

Repro project (save as `repro.proj`):
```xml
    <Target Name="Print">
        <Exec Command="chcp" />
    </Target>       
</Project>
```

Repro commands in `cmd.exe`:
```
>chcp
Active code page: 850

>msbuild repro.proj
Microsoft (R) Build Engine version 16.4.0-preview-19517-01+31c846a51 for .NET Framework

Build started 11/1/2019 8:35:23 AM.
Project "repro.proj" on node 1 (default targets).
Print:
  chcp
  Active code page: 850
Done Building Project "repro.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.18

>chcp 65001
Active code page: 65001

>msbuild repro.proj
Microsoft (R) Build Engine version 16.4.0-preview-19517-01+31c846a51 for .NET Framework

Build started 11/1/2019 8:35:29 AM.
Project "repro.proj" on node 1 (default targets).
Print:
  chcp
  Active code page: 850
Done Building Project "repro.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.23
```

Note: 850 is the Western Europe/US code page, representing most latin-based characters. 65001 is UTF-8, which can represent all Unicode characters.

### Expected  behavior
When run inside MSBuild, `chcp` should output the same code page as its ancestor `cmd.exe`.

Programs should inherit the console code page from their ancestors.

### Actual behavior
When run inside MSBuild, `chcp` outputs 850 rather than 65001 as set on the ancestor `cmd.exe`.

MSBuild's `<Exec>` task executes its commands in such a way that the child processes do not inherit the console code page from MSBuild (which itself inherited it from its parent). The console code page for children processes is rather reset to the OEM code page, which cannot encode Unicode characters in any Windows OS language in its default configuration. This means that it is impossible to use MSBuild to run a program that will output Japanese characters on an English OS without it coming out as a bunch of `?????`.

Specifically, the use of `ProcessStartInfo.CreateNoWindow = true` in `ToolTask` will reset the console code page of the child process to the OEM code page. If this is fixed, the batch file should also be written in the console code page since that is what `cmd.exe` will be using to read it. Pull request coming soon.

### Environment data
`msbuild /version` output: `Microsoft (R) Build Engine version 16.4.0-preview-19517-01+31c846a51 for .NET Framework`
OS info: Windows 10

MrTrillian:
MSBuild's `<Exec>` task executes its commands in such a way that the child processes do not inherit the console code page from MSBuild (which itself inherited it from its parent). The console code page for children processes is rather reset to the OEM code page, which cannot encode Unicode characters in any Windows OS language in its default configuration. This means that it is impossible to use MSBuild's `<Exec>` task to run a program that will output Japanese characters on an English OS without it coming out as a bunch of `?????`.

Specifically, the use of `ProcessStartInfo.CreateNoWindow = true` in `ToolTask` will reset the console code page of the child process to the OEM code page. With this fixed, the batch file must also be written in the console code page since that is what `cmd.exe` will be using to read it.

With this change, a parent process can set the console output code page to 65001 (UTF-8) and have some hope of getting Unicode data back from the programs started by MSBuild.

This change has had minimal testing. I only built it and verified that `chcp` called in `<Exec>` will now output the console code page as inherited from MSBuild's parent process. There is also a risk of regression since some ill-behaved programs always output in the OEM code page, no matter what the console output code page is, and with this change they will have to face the reality of their misdeeds. Regressions would only happen when an ancestor process explicitly changes the console output code page from the default OEM code page to something else.

Fixes #4870

### DIFF
--- a/src/Utilities/ToolTask.cs
+++ b/src/Utilities/ToolTask.cs
@@ -588,8 +588,20 @@ namespace Microsoft.Build.Utilities
                 LogPrivate.LogWarningWithCodeFromResources("ToolTask.CommandTooLong", GetType().Name);
             }
 
+            // Do *not* set startInfo.CreateNoWindow = true.
+            // This causes the child process to be launched with its console code page (Win32 GetConsoleCP),
+            // and console output code page (Win32 GetConsoleOutputCP) reset to the OEM code page (Win32 GetOEMCP),
+            // rather than inherited from this process.
+            //
+            // This is problematic because processes normally write to their standard output using
+            // the console output code page. If that is reset to the default OEM code page,
+            // then it will be impossible to represent all Unicode characters,
+            // since the default configuration for all Windows OS languages is a non-Unicode-capable OEM code page.
+            //
+            // The only way to get Unicode output through the console is if an ancestor process
+            // set the console output code page to UTF-8 and text-producing programs honor it.
             ProcessStartInfo startInfo = new ProcessStartInfo(pathToTool, commandLine);
-            startInfo.CreateNoWindow = true;
+            startInfo.CreateNoWindow = false;
             startInfo.UseShellExecute = false;
             startInfo.RedirectStandardError = true;
             startInfo.RedirectStandardOutput = true;
@@ -1365,7 +1377,11 @@ namespace Microsoft.Build.Utilities
                     }
                     else
                     {
-                        File.AppendAllText(_temporaryBatchFile, commandLineCommands, Console.InputEncoding); // Or ideally a new EncodingUtilities.CurrentConsoleEncoding based on GetConsoleCP
+                        // CMD.exe reads batch files using the current console code page,
+                        // as retrieved from the Win32 GetConsoleCP API or Console.InputEncoding in .NET.
+                        // The console code page defaults to the OEM code page (from GetOEMCP),
+                        // but may have been changed by any ancestor process and inherited down.
+                        File.AppendAllText(_temporaryBatchFile, commandLineCommands, Console.InputEncoding);
 
                         string batchFileForCommandLine = _temporaryBatchFile;
 
--- a/src/Utilities/ToolTask.cs
+++ b/src/Utilities/ToolTask.cs
@@ -1365,7 +1365,7 @@ namespace Microsoft.Build.Utilities
                     }
                     else
                     {
-                        File.AppendAllText(_temporaryBatchFile, commandLineCommands, Console.InputEncoding); // Or Ideally a new EncodingUtilities.CurrentConsoleEncoding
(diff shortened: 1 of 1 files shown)


src/Utilities/ToolTask.cs
```diff
@@ -588,8 +588,20 @@ string responseFileSwitch
...
+            // This is problematic because processes normally write to their standard output using
+            // the console output code page. If that is reset to the default OEM code page,
+            // then it will be impossible to represent all Unicode characters,
+            // since the default configuration for all Windows OS languages is a non-Unicode-capable OEM code page.
+            //
+            // The only way to get Unicode output through the console is if an ancestor process
+            // set the console output code page to UTF-8 and text-producing programs honor it.
             ProcessStartInfo startInfo = new ProcessStartInfo(pathToTool, commandLine);
-            startInfo.CreateNoWindow = true;
+            startInfo.CreateNoWindow = false;
```
MrTrillian: Note: I do not know the ramifications of changing this outside of what I'm trying to fix, if any. The [documentation](https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags) says:

> CREATE_NO_WINDOW 0x08000000
> The process is a console application that is being run without a console window. Therefore, the console handle for the application is not set. This flag is ignored if the application is not a console application, or if it is used with either CREATE_NEW_CONSOLE or DETACHED_PROCESS.

MrTrillian:
Then let us hear from @rainersigwald :)

src/Utilities/ToolTask.cs
MrTrillian: If this is problematic, I propose an alternate solution in my comment on #4870

rainersigwald:
This definitely makes me very nervous. Last time we mucked with cmd.exe codepages it caused a year-long bug tail and didn't fix things.

Doesn't avoiding this API cause the executed script to have access to MSBuild's console? That seems like a pretty bad idea: tools could then directly emit output and overwrite the build log.

Have you discussed this change with the Windows console folks?

MrTrillian:
I meant this PR to be a reference to what a fix might look like, but the main point is to make sure the original internal bug gets tracked, and I realize that getting into the implementation is distracting from this goal. I will reassign that internal bug to your path and update the github issue to better reflect it.

Issue #822, opened 2016-07-26
Syntactic sugar for importing files if they exist


mhutch:
A common pattern in MSBuild is to import targets files only if they exist.

```xml
    <Import Project="Foo.targets" Condition="Exists('Foo.targets')"/>
```

This could be made much more readable by adding explicit support for this pattern via a `ImportIfExists` element: 

```xml
<ImportIfExists Project="Foo.targets" />`
```

Nirmal4G:
Really need something like this.

Nirmal4G:
Since, SDKs came along, we might not need it more than usual.

Nirmal4G:
How does this work for `Imports` with `Sdk` attribute?

mhutch:
> How does this work for `Imports` with `Sdk` attribute?

I think this could work fine:

```xml
<ImportIfExists Sdk="Foo.Sdk" Project="Bar.targets" />
```

Not sure how useful it use, but it's not actually something that's possible today.

Nirmal4G:
It'll be useful for composing and detecting multiple SDKs all at once. I was thinking long the lines of Custom SDK injection...

#### Some examples

```xml
<Import Sdk="NET.Defaults.Sdk" Project="Defaults.targets" Condition="$(UseNETSdkDefaults)"/>

<Import Sdk="$(PlatformName).Sdk" Project="Core.targets" Condition="SdkExists('$(PlatformName).Sdk')"/>
```

mhutch:
+ @dsplaisted this use case overlaps with our discussions of possible fallbacks for missing sdks ^

PR #4879, opened 2019-11-02
822 - Syntactic Sugar for Importing files if exist

tjt60716:
#822 

Changed
```xml
<Import Project="$(MSBuildExtensionsPath)\Foo\Bar.targets"
    Condition="Exists('$(MSBuildExtensionsPath)\Foo\Bar.targets')"
/>
```

to

```xml
<ImportIfExists Project="$(MSBuildExtensionsPath)\Foo\Bar.targets" />`
```

with
@stuleelight
Stuart Lightfoot

This is our first time opening a pull request in this project!

### DIFF
--- a/src/Tasks/Microsoft.CSharp.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.CSharp.CurrentVersion.targets
@@ -24,10 +24,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
        <ImportUserLocationsByWildcardAfterMicrosoftCSharpTargets Condition="'$(ImportUserLocationsByWildcardAfterMicrosoftCSharpTargets)' == ''">true</ImportUserLocationsByWildcardAfterMicrosoftCSharpTargets>
     </PropertyGroup>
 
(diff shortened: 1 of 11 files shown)


rainersigwald:
Thanks for the contribution! However, I don't think we'll be able to accept it at this time.

The `ImportIfExists` element is not currently supported. #822 proposes adding support for it, at which time a PR like this could be made to adopt it.

If you're interested in adding support for the element, we should talk through its design and implications in #822 before starting to code it up. Historically, we've been hesitant to add new elements, since using a new element means you can no longer load a project in an older Visual Studio or MSBuild. That's not necessarily reason to not do it, though--we have to start somewhere! But we should talk about the timing.

stuleelight:
Rainer,
I super appreciate the reply. If I am being honest, this was part of a Hackathon, and I believe our prompt was a touch off. It seemed too easy and my partner and I asked several times about it. 
However, I’m decent with C# most through game development, but I’m obviously a touch green still. Where or how would you start this? 
Am I understanding correctly we are looking to create a type of function that checks ifExists first then proceed with the import?
Any help is appreciated.

Sincerely,
Stuart

PR #4888, opened 2019-11-05
Include path to MSBuild in node handshake salt

### LINKED ISSUE #3137: Avoid crosstalk between same-MSBuild-but-different-SDKs nodes
~We should add the environment variables `DOTNET_HOST_PATH` and `DOTNET_ROOT` to the node-compatibility handshake computation.~

### Steps to reproduce

* Have two copies of the .NET Core SDK that use the same version of MSBuild but that aren't identical
* Run a build with copy A (created nodes will wait for reuse after the build completes)
* Run a build with copy B

### Expected  behavior

The build in copy B's dotnet host does not reuse nodes from copy A

### Actual behavior

This will be mostly true now, because the handshake includes `AssemblyTimestamp` which is `LastWriteTime` which will _probably_ be distinct, but isn't necessarily.

Forgind:
Prevent unwanted node reuse.

Fixes #3137.

### DIFF
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -655,7 +655,9 @@ namespace Microsoft.Build.Internal
         /// <returns>Base Handshake</returns>
         private static long GetBaseHandshakeForContext(TaskHostContext hostContext)
         {
-            string salt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT");
+            string salt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT") +
+                Environment.GetEnvironmentVariable("DOTNET_HOST_PATH") +
+                Environment.GetEnvironmentVariable("DOTNET_ROOT");
 
             long nodeHandshakeSalt = 0;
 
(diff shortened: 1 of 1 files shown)


src/Shared/CommunicationsUtilities.cs
```diff
@@ -655,7 +655,9 @@ internal static void Trace(int nodeId, string format, params object[] args)
         /// <returns>Base Handshake</returns>
         private static long GetBaseHandshakeForContext(TaskHostContext hostContext)
         {
-            string salt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT");
+            string salt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT") +
+                Environment.GetEnvironmentVariable("DOTNET_HOST_PATH") +
+                Environment.GetEnvironmentVariable("DOTNET_ROOT");
```
rainersigwald: Should we be using environment variables for these, or determining them programmatically? I expected the latter, but if the env var is sufficient that's easy . . .

Forgind: You mentioned they were environment variables in the issue. I can try to figure them out another way if you'd prefer. I just set them as desired on my computer.

rainersigwald: We shouldn't require user intervention to set these. I believe `dotnet build` and `dotnet msbuild` both set them for the child MSBuild process. Can you confirm that? If they both do, using the env var is probably enough.

@dsplaisted do you have an opinion on this?

dsplaisted: Isn't the path to MSBuild.dll also part of the handshake?  In that case I don't think we would additionally need to add the path to `dotnet`, as that should be less unique than the MSBuild path.

PR #4896, opened 2019-11-07
whitespace cleanup

treymerkley:
I think this is including some previous commits for some reason, let me know if I need to remove them.
#4779


src/Build.UnitTests/ConsoleLogger_Tests.cs
```diff
@@ -560,7 +560,7 @@ public void ColorTest(string expectedMessageType, string expectedColor, bool par
 
                 if (expectedMessageType.Equals("message"))
                 {
-                    
+
```
drewnoakes: Remove this line altogether?

```suggestion
```

PR #4897, opened 2019-11-07
Completely automate experimental branch builds

BenVillalobos:
Builds prefixed with `exp/` use `master` for their `IbcSourceBranchName`
Builds NOT prefixed with `exp/` use the default `IbcSourceBranchName` (which is the branch name itself).

With this change inserted, branches prefixed with `exp/` will no longer fail because of optprof data not existing for that branch.

Yay automation!

### DIFF
--- a/.vsts-dotnet.yml
+++ b/.vsts-dotnet.yml
@@ -91,7 +91,27 @@ stages:
                 /p:TeamName=MSBuild
                 /p:DotNetPublishUsingPipelines=true
       displayName: Build
-      condition: succeeded()
+      condition: and(succeeded(), not(startsWith(variables['Build.SourceBranch'], 'refs/heads/exp/')))
+      
+    - script: eng/CIBuild.cmd
+                -configuration $(BuildConfiguration)
+                -officialBuildId $(Build.BuildNumber)
+                -officialSkipApplyOptimizationData $(SkipApplyOptimizationData)
+                /p:RepositoryName=$(Build.Repository.Name)
+                /p:VisualStudioIbcSourceBranchName=master
+                /p:VisualStudioDropAccessToken=$(System.AccessToken)
+                /p:VisualStudioDropName=$(VisualStudio.DropName)
+                /p:DotNetSignType=$(SignType)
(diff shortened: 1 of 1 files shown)


.vsts-dotnet.yml
```diff
@@ -91,7 +91,27 @@ stages:
                 /p:TeamName=MSBuild
                 /p:DotNetPublishUsingPipelines=true
       displayName: Build
-      condition: succeeded()
+      condition: and(succeeded(), not(startsWith(variables['Build.SourceBranch'], 'refs/heads/exp/')))
```
rainersigwald: Is it possible to define a new variable with this condition and use that in a single invocation of `cibuild.cmd`?

BenVillalobos: Livar just pointed me to an example for that. Much cleaner!

.vsts-dotnet.yml
```diff
@@ -11,6 +11,10 @@ trigger:
 #   SkipApplyOptimizationData: false
 
 variables:
+  - ${{ if startsWith(variables['Build.SourceBranch'], 'refs/heads/exp/')) }}:
+    SourceBranch: master
```
Forgind: Why is this format different from the lines just below?

.vsts-dotnet.yml
```diff
@@ -78,7 +82,7 @@ stages:
                 -officialBuildId $(Build.BuildNumber)
                 -officialSkipApplyOptimizationData $(SkipApplyOptimizationData)
                 /p:RepositoryName=$(Build.Repository.Name)
-                /p:VisualStudioIbcSourceBranchName=$(IbcSourceBranchName)
+                /p:VisualStudioIbcSourceBranchName=SourceBranch
```
Forgind: Doesn't this need to be accessed via $(SourceBranch)?

BenVillalobos: Yep! Thanks

.vsts-dotnet.yml
BenVillalobos: Changed it to match the format below

PR #4899, opened 2019-11-07
Apply a solution filter

### LINKED ISSUE #4097: Consider supporting .slnf solution filters
This is a new feature in VS: https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019. MSBuild doesn't understand `.slnf` files so you can't build them at the command line, but I assume (without looking into it at all) that it would be possible to build a solution metaproject using a `.slnf` and its `.sln` and build it.

Forgind:
This fixes the fact that, when building, we always build a complete solution even if only a small part of it is really necessary. One workaround is to have a stripped-down .sln file, but .slnf has a simpler format and less duplicated information. This creates support for .slnf files in MSBuild.

Fixes #4097.

Filter files currently look like this:
```{
  "solution": {
    "path": "C:\\Users\\namytelk\\Documents\\GitHub\\msbuild\\MSBuild.Dev.sln",
    "projects": [
      "src\\Build\\Microsoft.Build.csproj",
      "src\\Framework\\Microsoft.Build.Framework.csproj",
      "src\\MSBuild\\MSBuild.csproj",
      "src\\Tasks.UnitTests\\Microsoft.Build.Tasks.UnitTests.csproj"
    ]
  }
}
```

Documentation can be found [here](https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019).


src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -195,7 +197,16 @@ internal string FullPath
             {
                 // Should already be canonicalized to a full path
                 ErrorUtilities.VerifyThrowInternalRooted(value);
-                _solutionFile = value;
+                if (FileUtilities.IsSolutionFilterFilename(value))
+                {
+                    _solutionFilterFile = value;
+                    _solutionFile = value.Substring(0, value.Length - 1);
```
rainersigwald: Don't you need to parse the solution filename out of the slnf? For example, in Roslyn's `Compilers.slnf` it points to `Roslyn.sln`:

https://github.com/dotnet/roslyn/blob/7bc44488c661fd6bbb6c53f39512a6fe0cc5ef84/Compilers.slnf#L3

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -261,10 +272,13 @@ internal static void GetSolutionFileAndVisualStudioMajorVersions(string solution
             solutionVersion = 0;
             visualStudioMajorVersion = 0;
 
+            // Solutions can be passed as either a solution or a solution filter or truncated list of what projects to build.
+            string realSolutionFile = FileUtilities.IsSolutionFilterFilename(solutionFile) ? solutionFile.Substring(0, solutionFile.Length - 1) : solutionFile;
```
rainersigwald: Does this method need to be aware of slnf at all? Couldn't it just always be given a real solution?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -697,13 +750,20 @@ private void ParseProject(string firstLine)
             ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile(line != null, "SubCategoryForSolutionParsingErrors",
                 new BuildEventFileInfo(FullPath), "SolutionParseProjectEofError", proj.ProjectName);
 
-            // Add the project to the collection
-            AddProjectToSolution(proj);
-            // If the project is an etp project then parse the etp project file 
-            // to get the projects contained in it.
-            if (IsEtpProjectFile(proj.RelativePath))
+            if (slnFilter == null || slnFilter.Contains(proj.RelativePath))
```
rainersigwald: Is this valid? I would have expected that we'd need to understand all projects in the solution for configuration purposes, even if they're not valid entry points.

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -431,16 +445,55 @@ internal void ParseSolution()
 
             ParseFileHeader();
 
+            HashSet<string> slnFilter = null;
+            if (_solutionFilterFile != null)
+            {
+                slnFilter = new HashSet<string>();
+                string filterFile = File.ReadAllText(_solutionFilterFile);
+                int projectsInd = filterFile.IndexOf('[', filterFile.IndexOf("\"projects\":"));
```
rainersigwald: Use `System.Text.Json` to parse the file. VS uses a JSON library so it'll be tolerant of various quirks and we should be similarly tolerant.

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -431,16 +445,55 @@ internal void ParseSolution()
...
+                bool inQuote = false;
+                StringBuilder sb = null;
+                if (projectsInd == -1)
+                {
+                    throw new InvalidDataException("Solution filter file is incorrectly formatted.");
```
rainersigwald: These need to produce localized user-facing errors with MSB codes. How are other sln-parsing errors handled? Can we do things the same way?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -632,7 +685,7 @@ private void ValidateSolutionFileVersion(string versionString)
         ///  EndProject
         /// 
         /// </summary>
-        private void ParseProject(string firstLine)
+        private void ParseProject(string firstLine, HashSet<string> slnFilter, HashSet<string> guidFilter)
```
rainersigwald: What is `guidFilter` for?

src/Shared/FileUtilities.cs
```diff
@@ -959,7 +959,12 @@ internal static bool FileOrDirectoryExistsNoThrow(string fullPath, IFileSystem f
         /// </summary>
         internal static bool IsSolutionFilename(string filename)
         {
-            return HasExtension(filename, ".sln");
+            return HasExtension(filename, ".sln") || HasExtension(filename, ".slnf");
```
rainersigwald: Why does this need to be extended?

src/Build/Construction/Solution/SolutionFile.cs
Forgind: What do you have to do to include System.Text.Json? `using System.Text.Json` didn't work.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: When you make a solution filter file, it automatically checks for dependencies, so my understanding is that you will either have all the necessary projects in the filter file, or it should fail.

src/Shared/FileUtilities.cs
Forgind: This is used a number of times to check if the input is a solution rather than, say, a project. Extending this function cleanly brings the solution filter file to SolutionFile.cs where it can be parsed.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: It's for nested projects; they refer to each other by guid instead of path, so I needed a second way of determining if a project should be included or not.

eng/Packages.props
```diff
@@ -39,6 +39,7 @@
     <PackageReference Update="System.Runtime.Serialization.Primitives" Version="4.1.1" />
     <PackageReference Update="System.Security.Principal.Windows" Version="4.3.0" />
     <PackageReference Update="System.Text.Encoding.CodePages" Version="4.0.1" />
+    <PackageReference Update="System.Text.Json" Version="4.6.0" />
```
rainersigwald: This will require setup authoring changes to make sure we deliver it to VS.

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -195,7 +200,39 @@ internal string FullPath
...
-                _solutionFile = value;
+                if (FileUtilities.IsSolutionFilterFilename(value))
+                {
+                    using (JsonDocument text = JsonDocument.Parse(File.ReadAllText(value)))
+                    {
+                        try
+                        {
+                            JsonElement solution = text.RootElement.GetProperty("solution");
+                            _solutionFile = solution.GetProperty("path").GetString();
+                            _solutionFilter = new HashSet<string>(NativeMethodsShared.IsLinux ? StringComparer.Ordinal : StringComparer.OrdinalIgnoreCase);
```
rainersigwald: Where did we land on this case insensitivity/how does VS currently behave?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -195,7 +200,39 @@ internal string FullPath
...
+                            foreach (JsonElement project in solution.GetProperty("projects").EnumerateArray())
+                            {
+                                _solutionFilter.Add(project.GetString());
+                            }
+                        }
+                        catch (Exception)
+                        {
+                            ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile
+                                (
+                                    false /* just throw the exception */,
```
rainersigwald: Can we preserve information about the exception to pass up to the user?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -218,6 +255,11 @@ internal string FullPath
 
         #region Methods
 
+        internal bool IsFiltered(string projectFile)
```
```suggestion
        internal bool FilterIncludesProject(string projectFile)
```

?

src/Build/Construction/Solution/SolutionProjectGenerator.cs
```diff
@@ -637,6 +637,11 @@ private static string DetermineLikelyActiveSolutionConfiguration(SolutionFile so
         /// </summary>
         private static bool WouldProjectBuild(SolutionFile solutionFile, string selectedSolutionConfiguration, ProjectInSolution project, ProjectConfigurationInSolution projectConfiguration)
         {
+            if (!solutionFile.IsFiltered(project.RelativePath))
```
rainersigwald: I think this deserves a comment.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: If I modify a .sln file by changing case, VS handles it fine. If I modify a .slnf file to change case, it has trouble finding it. Since .sln files have existed much longer than .slnf files, I wouldn't be surprised if .slnf files change at some point, so I would prefer to stick with ignoring case.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: Does this not preserve information—in the form of an exception the user can catch and inspect?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -218,6 +255,11 @@ internal string FullPath
 
         #region Methods
 
+        public bool IsFiltered(string projectFile)
+        {
+            return _solutionFilter == null || _solutionFilter.Contains(projectFile);
+        }
+
```
Forgind: Do you think this method should be more intelligent? In particular, I was wondering if we should check if there's a mismatch between full and relative paths. One extra complication for this is that paths in the .slnf are, by default, relative from the .sln file, which might include some `..`s that would have to be simplified out.

rainersigwald: I don't understand the question. Can you clarify? I would expect the path constraints on projects in the solution to match the slnf; is that not true?

Forgind: Currently, both the .sln and .slnf use relative paths from the .sln, so there's no problem. As you noted at one point, however, .slnfs are easy to modify, so this would make it resistant to adding an absolute path into the .slnf.

rainersigwald: I'd rather reject such an slnf than try to tolerate it (unless VS does)

src/Build/Construction/Solution/SolutionFile.cs
rainersigwald: No, because a primary use case for this functionality is using `msbuild.exe foo.slnf` on the command line, where you get only MSBuild errors (not exceptions).

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -461,6 +527,31 @@ internal void ParseSolution()
                 }
             }
 
+            if (_solutionFilter != null)
+            {
+                HashSet<string> projectNamesInOrder = new HashSet<string>(_projectsInOrder.Count);
```
rainersigwald: Should this have the same comparer case sensitivity as the solution filter?

src/Build/Instance/ProjectInstance.cs
```diff
@@ -2035,9 +2035,15 @@ internal static ProjectInstance[] LoadSolutionForBuild(string projectFile, Prope
             else
             {
                 int solutionVersion;
-                int visualStudioVersion;
-                SolutionFile.GetSolutionFileAndVisualStudioMajorVersions(projectFile, out solutionVersion, out visualStudioVersion);
-
+                int visualStudioVersion = 0;
+                if (FileUtilities.IsSolutionFilterFilename(projectFile))
+                {
+                    solutionVersion = 16;
```
rainersigwald: This might age badly if there's a 17 that supports slnf and some new thing. Can you get it from the referenced `.sln`?

src/Shared/FileUtilities.cs
rainersigwald: Update the doc comment with `<remarks>` saying that, please

src/Shared/Resources/Strings.shared.resx
```diff
@@ -262,10 +262,26 @@
     <value>Path: {0} exceeds the OS max path limit. The fully qualified file name must be less than {1} characters.</value>
     <comment></comment>
   </data>
+  <data name="JsonParsingError" UESanitized="false" Visibility="Public">
```
rainersigwald: Convention in shared resources is to start with the `Shared.` prefix.

rainersigwald: Give these a bit of a "namespace" like maybe `SolutionFilterJsonParseError`.

ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -458,6 +458,7 @@ public sealed partial class SolutionFile
         public string GetDefaultConfigurationName() { throw null; }
         public string GetDefaultPlatformName() { throw null; }
         public static Microsoft.Build.Construction.SolutionFile Parse(string solutionFile) { throw null; }
+        public bool ProjectShouldBuild(string projectFile) { throw null; }
```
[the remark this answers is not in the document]
Forgind: It needs to be visible to SolutionFile_Tests.cs, and given how long it took to resolve the previous IVT change, I'm somewhat reluctant to make it internal and add more IVT changes.

src/Build/Instance/ProjectInstance.cs
Forgind: Yes. My rationale is that it wouldn't make any difference now, and this way avoids file i/o (slow) and a string search. Would you be ok with it if I just put a comment like `// TODO: parse file to get accurate solution version`?

src/Shared/Resources/Strings.shared.resx
Forgind: There are a lot without that prefix (see `SolutionParseVersionMismatchError` or `UnrecognizedSolutionComment` for instance). Do you want me to change those as well?

danmosemsft: Aside for @rainersigwald afaik `UESanitized` has not been used since approx 2004 and can be removed everywhere unless you know differently

Forgind: Also, I don't see any examples from SolutionFile with the Shared. prefix. Maybe there's an exception for some reason?

src/Shared/Resources/Strings.shared.resx
```diff
@@ -262,10 +262,26 @@
...
+    <value>MSB5025: Json in solution filter file "{0}" is incorrectly formatted.</value>
+    <comment>{StrBegin="MSB: 5025"}UE: The solution filename is provided separately to loggers.</comment>
+  </data>
+  <data name="MissingSolutionError" UESanitized="false" Visibility="Public">
+    <value>MSB5026: The solution filter file at "{0}" specifies there will be a solution file at "{1}", but that file does not exist.</value>
+    <comment>{StrBegin="MSB: 5026"}UE: The solution filename is provided separately to loggers.</comment>
```
danmosemsft: Some of these StrBegin are wrong. `{StrBegin="MSB: 5026"}` should be `{StrBegin="MSB5026: "}`

@rainersigwald The purpose of these annotations was to support pseudo localizer tools without breaking the processing of the string by removing the necessary error/warning prefix. I do not know whether pseudo localization is still relevant these days -- if not, they can all be removed -- unlreated to this change of course

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -195,7 +200,63 @@ internal string FullPath
...
+                    try
+                    {
+                        using JsonDocument text = JsonDocument.Parse(File.ReadAllText(value));
+                        JsonElement solution = text.RootElement.GetProperty("solution");
+                        _solutionFile = solution.GetProperty("path").GetString();
+                        if (!File.Exists(_solutionFile))
+                        {
+                            ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile
+                            (
+                                false /* just throw the exception */,
```
danmosemsft: There is an overload of `ProjectFileErrorUtilities.ThrowInvalidProjectFile` with 5 parameters that is missing here. It is identical to what is written here but assumes false for the first parameter

Forgind: I see three `ProjectFileErrorUtilities.ThrowInvalidProjectFile`s, and they all take `bool condition` as the first argument.

src/Shared/Resources/Strings.shared.resx
rainersigwald: I don't know differently. @tmeschter should we remove 'em all?

rainersigwald: I wouldn't go back and change the others, but I'd expect anything in this file (compiled into multiple assemblies) to have the prefix. I don't know why the existing solution ones don't, that's funky.

tmeschter: You can remove them all. They don't affect the .xlf files in any way, and so aren't part of the localization pipeline.

src/Shared/Resources/Strings.shared.resx
tmeschter: The loc team does use annotations like these for purposes beyond pseudo loc; I recommend correcting them when you see issues rather than removing them.

src/Shared/Resources/Strings.shared.resx
```diff
@@ -48,107 +48,107 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.3500.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="BuildAborted" UESanitized="false" Visibility="Public">
+  <data name="BuildAborted" Visibility="Public">
     <value>MSB4188: Build was canceled.</value>
     <comment>{StrBegin="MSB4188: "} Error when the build stops suddenly for some reason. For example, because a child node died.</comment>
   </data>
-  <data name="BuildEngineCallbacksInTaskHostUnsupported" UESanitized="false" Visibility="Public">
+  <data name="BuildEngineCallbacksInTaskHostUnsupported" Visibility="Public">
```
rainersigwald: Please don't remove `UESanitized` in this PR; it's not related to this work.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
```diff
@@ -16,6 +16,84 @@ namespace Microsoft.Build.UnitTests.Construction
...
+                ")]
+        public void InvalidSolutionFilters(string slnfValue)
+        {
+            Assert.False(File.Exists("C:\\notAPath2\\MSBuild.Dev.sln"));
+            using (TestEnvironment testEnvironment = TestEnvironment.Create())
+            {
+                TransientTestFolder folder = testEnvironment.CreateFolder(createFolder: true);
+                TransientTestFile sln = testEnvironment.CreateFile(folder, "Dev.sln");
+                TransientTestFile slnf = testEnvironment.CreateFile(folder, "Dev.sln", slnfValue.Replace(@"C:\\notAPath\\MSBuild.Dev.sln", sln.Path.Replace("\\", "\\\\")));
+                Shouldly.Should.Throw<InvalidProjectFileException>(() => { SolutionFile sf = SolutionFile.Parse(slnf.Path); });
```
rainersigwald: Add a `using` for `Shouldly` and simplify this reference.

rainersigwald:
```suggestion
                Shouldly.Should.Throw<InvalidProjectFileException>(() => SolutionFile.Parse(slnf.Path));
```

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
```diff
@@ -16,6 +16,84 @@ namespace Microsoft.Build.UnitTests.Construction
...
+                      {""path"": ""src\\Framework\\Microsoft.Build.Framework.csproj""},
+                      {""path"": ""src\\MSBuild\\MSBuild.csproj""},
+                      {""path"": ""src\\Tasks.UnitTests\\Microsoft.Build.Tasks.UnitTests.csproj""}
+                    ]
+                    }
+                }
+                ")]
+        public void InvalidSolutionFilters(string slnfValue)
+        {
+            Assert.False(File.Exists("C:\\notAPath2\\MSBuild.Dev.sln"));
```
rainersigwald: Is this relevant? Don't you just replace it?

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
```diff
@@ -16,6 +16,84 @@ namespace Microsoft.Build.UnitTests.Construction
     /// </summary>
     public class SolutionFile_Tests
     {
+        [Theory]
+        [InlineData(@"
```
rainersigwald: It's not obvious to me what's wrong with these filters. Can you either

1. abstract the current method to a helper and make each of these a named method describing the problem
2. specify additional data in the theory input to give names (or inner exception types?)

That would also allow you to have more detailed checks on the assertion--"doesn't accept bad input of type X" is bare minimum and I'd like to test that the correct, specific, diagnosable errors are produced.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
```diff
@@ -16,6 +16,84 @@ namespace Microsoft.Build.UnitTests.Construction
...
+                {
+                  ""solution"": {
+                    ""path"": ""C:\\notAPath\\MSBuild.Dev.sln"",
```
rainersigwald: Are the paths in `.slnf` files really absolute? That feels pretty broken to me.

src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
```diff
@@ -60,6 +60,121 @@ public void Dispose()
...
+                TransientTestFolder classLibFolder = testEnvironment.CreateFolder(Path.Combine(folder.Path, "ClassLibrary"), createFolder: true);
+                TransientTestFolder classLibSubFolder = testEnvironment.CreateFolder(Path.Combine(classLibFolder.Path, "ClassLibrary"), createFolder: true);
+                TransientTestFile classLibrary = testEnvironment.CreateFile(classLibSubFolder, "ClassLibrary.csproj",
+                    @"<Project>
+                  <Target Name=""ClassLibraryTarget"">
+                      <Message Text=""ClassLibraryBuilt""/>
+                  </Target>
+                  </Project>
+                    ");
+                TransientTestFile classLibrary1Sol = testEnvironment.CreateFile(classLibFolder, "ClassLibrary.sln",
```
rainersigwald: What does this solution do?

src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
```diff
@@ -60,6 +60,121 @@ public void Dispose()
...
+                  </Target>
+                  </Project>
+                    ");
+                TransientTestFile solutionFile = testEnvironment.CreateFile(simpleProjectFolder, "SimpleProject.sln",
+                    @"
+                    Microsoft Visual Studio Solution File, Format Version 12.00
+                    # Visual Studio Version 16
+                    VisualStudioVersion = 16.0.29326.124
+                    MinimumVisualStudioVersion = 10.0.40219.1
+                    Project(""{9A19103F-16F7-4668-BE54-9A1E7A4F7556}"") = ""SimpleProject"", """ + simpleProject.Path + @""", ""{79B5EBA6-5D27-4976-BC31-14422245A59A}""
```
rainersigwald: Solutions generally have relative paths in them. Is it important to the test that it be absolute/nonstandard?

src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
```diff
@@ -60,6 +60,121 @@ public void Dispose()
...
+                        EndGlobalSection
+                    EndGlobal
+                ");
+                TransientTestFile filterFile = testEnvironment.CreateFile(folder, "solutionFilter.slnf",
+                    @"
+                {
+                  ""solution"": {
+                    ""path"": """ + solutionFile.Path.Replace("\\", "\\\\") + @""",
+                    ""projects"": [
+                      """ + simpleProject.Path.Replace("\\", "\\\\") + @"""
```
rainersigwald: Shouldn't these all be relative paths too?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -18,6 +20,8 @@
 using ExceptionUtilities = Microsoft.Build.Shared.ExceptionHandling;
 using System.Collections.ObjectModel;
 using Microsoft.Build.Shared;
+using Microsoft.Build.Shared.FileSystem;
+using System.Linq.Expressions;
```
rainersigwald: nit: alphabetize this into the `System.` block above

src/Build/Construction/Solution/SolutionFile.cs
rainersigwald: I think Dan's suggesting adding a new overload, and I agree.

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -195,7 +200,62 @@ internal string FullPath
...
                 ErrorUtilities.VerifyThrowInternalRooted(value);
-                _solutionFile = value;
+                if (FileUtilities.IsSolutionFilterFilename(value))
+                {
+                    try
+                    {
+                        using JsonDocument text = JsonDocument.Parse(File.ReadAllText(value));
+                        JsonElement solution = text.RootElement.GetProperty("solution");
+                        _solutionFile = solution.GetProperty("path").GetString();
+                        if (!File.Exists(_solutionFile))
```
rainersigwald: We have an abstraction for filesystem that we hope to build some caching into that should be used here:

```suggestion
                        if (!FileSystems.Default.FileExists(_solutionFile))
```

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -195,7 +200,62 @@ internal string FullPath
...
+                                new BuildEventFileInfo(_solutionFile),
+                                "SolutionFilterMissingSolutionError",
+                                value,
+                                _solutionFile
+                            );
+                        }
+                        _solutionFilter = new HashSet<string>(NativeMethodsShared.IsLinux ? StringComparer.Ordinal : StringComparer.OrdinalIgnoreCase);
+                        foreach (JsonElement project in solution.GetProperty("projects").EnumerateArray())
+                        {
+                            if (!File.Exists(Path.Combine(Path.GetDirectoryName(_solutionFile), project.GetString())) && !File.Exists(project.GetString()))
```
rainersigwald: Why not defer the check to the normal time that the solution would deal with the file?

I would expect an error when the filter's project path isn't a project in the `.sln` file (that's `SolutionFilterFilterContainsProjectNotInSolution`, right?), but not when the project doesn't exist on disk (at this point).

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -461,6 +526,31 @@ internal void ParseSolution()
                 }
             }
 
+            if (_solutionFilter != null)
+            {
+                HashSet<string> projectPaths = new HashSet<string>(_projectsInOrder.Count, NativeMethodsShared.IsLinux ? StringComparer.Ordinal : StringComparer.OrdinalIgnoreCase);
```
rainersigwald: I don't like sprinkling this case-sensitivity check here, since it's wrong in some cases (you can have a case-sensitive FS on macOS, and you can have case-insensitive ones on Linux). I don't think we need to fix that right now (it's basically impossible to get right), but I'd like to at least have it in one place--should we create a `NativeMethodsShared.IsFileSystemCaseSensitive` or `NativeMethodsShared.FileSystemComparer` or put a comparer into `IFileSystem` and use it here? @cdmihai feel free to chime in :)

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -461,6 +526,31 @@ internal void ParseSolution()
...
+                for (int a = 0; a < _projectsInOrder.Count; a++)
```
rainersigwald: why not a foreach?

src/Shared/FileUtilities.cs
rainersigwald: ping on extending comment here

src/Shared/Resources/Strings.shared.resx
Forgind: True, but it doesn't hurt, and you previously recommended removing them all.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
Forgind: I don't replace it, which means this test will fail if someone actually makes `C:\notAPath2\MSBuild.Dev.sln`, and I imagine they would be extremely confused when they went to investigate. This means it will still fail, but the user would have a hint in the right direction. I can replace it with a comment if you'd prefer.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
Forgind: That's the default. Not only are they absolute, you might notice the @ symbol at the front—they're escaped.

src/Build.UnitTests/Construction/SolutionProjectGenerator_Tests.cs
Forgind: It isn't critical. This was slightly easier for me so I didn't have to worry about slashes. I could replace it with `Path.Combine("SimpleProject", "SimpleProject.csproj")` if you'd prefer.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: Ah, I see. It might be easiest to just make it a default argument.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: Because this is a problem with the .slnf, not with the .sln. You're right that that other error is for if there's something in the .slnf that isn't in the .sln.

I think it's valuable to have both types of exceptions. If we let projects exist in the .slnf that don't exist elsewhere, the user could try to modify the .slnf, make a small typo, and be very confused as to why their project wasn't being built.

src/Build/Construction/Solution/SolutionFile.cs
Forgind: I replaced it with a call to NativeMethodsShared.OSUsesCaseSensitivePaths instead—for now, that just returns NativeMethodsShared.IsLinux, but if we decide it's time to implement a more complicated but more correct solution, this will be silently fixed.

src/Build/Construction/Solution/SolutionFile.cs
rainersigwald: I don't understand what you're saying here. Can you clarify?

The slnf filters the solution, right? So the entries in the slnf should match the projects in the solution. I don't see what presence/absence on disk has to do with anything.

Forgind: The motivating scenario here is if the user wants to include a project in what is being built, so they add it to the .slnf but make a typo. This throws an error if they do that.

Another way of looking at it is that if a file doesn't exist, we can't build it anyway, so it shouldn't be in the .slnf.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
```diff
@@ -91,6 +78,13 @@ public void InvalidSolutionFilters(string slnfValue, string exceptionReason)
                 TransientTestFolder folder = testEnvironment.CreateFolder(createFolder: true);
                 TransientTestFile sln = testEnvironment.CreateFile(folder, "Dev.sln");
                 TransientTestFile slnf = testEnvironment.CreateFile(folder, "Dev.slnf", slnfValue.Replace(@"C:\\notAPath\\MSBuild.Dev.sln", sln.Path.Replace("\\", "\\\\")));
+                try
+                {
+                    SolutionFile.Parse(slnf.Path);
+                } catch (Exception ex)
+                {
+                    Console.WriteLine(ex);
+                }
```
rainersigwald: ??

Forgind: I need to remember to take out code only used for testing _before_ committing.

src/Shared/Resources/Strings.shared.resx
```diff
@@ -262,10 +262,22 @@
...
   </data>
+  <data name="SolutionFilterJsonParsingError" Visibility="Public">
+    <value>MSB5025: Json in solution filter file "{0}" is incorrectly formatted.</value>
+    <comment>{StrBegin="MSB5025: "}UE: The solution filename is provided separately to loggers.</comment>
+  </data>
+  <data name="SolutionFilterMissingSolutionError" Visibility="Public">
+    <value>MSB5026: The solution filter file at "{0}" specifies there will be a solution file at "{1}", but that file does not exist.</value>
+    <comment>{StrBegin="MSB5026: "}UE: The solution filename is provided separately to loggers.</comment>
+  </data>
+  <data name="SolutionFilterFilterContainsProjectNotInSolution" Visibility="Public">
```
GearedToWar: Will this result in duplicated errors for a missing project file?
I see that when building a solution we get 

> MSB3202: The project file "<>" was not found.

 for a project in the solution that does not exist.

Forgind: MSB3202 is for if a project in the .sln doesn't exist on disk; this new error is for if a project is in the filter but is not in the solution file. Together, they ensure that any project in the .slnf is in both the .sln and on disk as desired, but they can never both be thrown because one requires the project to be in the .sln and the other cannot have the project in the .sln.

GearedToWar: Ah, got it. Thanks for the explanation.

PR #4905, opened 2019-11-09
Parallellized node creation

### LINKED ISSUE #860: Hang when child process exits without writing to IPC pipe
After MSBuild starts a child/worker process, it [tries to handshake](https://github.com/Microsoft/msbuild/blob/d1bdd68a3f2b1eec2e9a17ec62ebe70c59e9019b/src/XMakeBuildEngine/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs#L412-L416) with the newly-created node by writing to an IPC pipe and waiting for the handshake response.

If process creation fails, an error is propagated upward and MSBuild doesn't hang. If, however, process creation succeeds but the child quickly terminates instead of waiting for the handshake, the indefinite blocking read on handshake response turns into a mysterious MSBuild hang.

We should either have a timeout for handshake success, or at the very least fail early in `WaitForConnectionFromProcess` if the `nodeProcessId` doesn't belong to a running process.

Forgind:
This makes node creation faster and solves the problem in which waiting 30 seconds for each of many nodes seems like a hang.

Fixes #860.

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildManager.cs
+++ b/src/Build/BackEnd/BuildManager/BuildManager.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Collections.ObjectModel;
@@ -102,7 +103,7 @@ namespace Microsoft.Build.Execution
         /// <summary>
         /// Set of active nodes in the system.
         /// </summary>
-        private readonly HashSet<NGen<int>> _activeNodes;
+        private readonly ConcurrentDictionary<int, byte> _activeNodes;
 
         /// <summary>
         /// Event signalled when all nodes have shutdown.
@@ -262,7 +263,7 @@ namespace Microsoft.Build.Execution
             _buildSubmissions = new Dictionary<int, BuildSubmission>();
             _graphBuildSubmissions = new Dictionary<int, GraphBuildSubmission>();
             _noActiveSubmissionsEvent = new AutoResetEvent(true);
-            _activeNodes = new HashSet<NGen<int>>();
+            _activeNodes = new ConcurrentDictionary<int, byte>();
             _noNodesActiveEvent = new AutoResetEvent(true);
             _nodeIdToKnownConfigurations = new Dictionary<NGen<int>, HashSet<NGen<int>>>();
             _unnamedProjectInstanceToNames = new Dictionary<ProjectInstance, string>();
@@ -1789,8 +1790,8 @@ namespace Microsoft.Build.Execution
         private void HandleNodeShutdown(int node, NodeShutdown shutdownPacket)
         {
             _shuttingDown = true;
-            ErrorUtilities.VerifyThrow(_activeNodes.Contains(node), "Unexpected shutdown from node {0} which shouldn't exist.", node);
-            _activeNodes.Remove(node);
+            ErrorUtilities.VerifyThrow(_activeNodes.Keys.Contains(node), "Unexpected shutdown from node {0} which shouldn't exist.", node);
+            _activeNodes.TryRemove(node, out _);
 
             if (shutdownPacket.Reason != NodeShutdownReason.Requested)
             {
@@ -1924,24 +1925,34 @@ namespace Microsoft.Build.Execution
 
                     case ScheduleActionType.CreateNode:
                         var newNodes = new List<NodeInfo>();
-
-                        for (int i = 0; i < response.NumberOfNodesToCreate; i++)
+                        GetNodeConfiguration();
+                        try
                         {
-                            NodeInfo createdNode = _nodeManager.CreateNode(GetNodeConfiguration(), response.RequiredNodeType);
-
-                            if (null != createdNode)
+                            Parallel.For(0, response.NumberOfNodesToCreate, i =>
                             {
-                                _noNodesActiveEvent.Reset();
-                                _activeNodes.Add(createdNode.NodeId);
-                                newNodes.Add(createdNode);
-                                ErrorUtilities.VerifyThrow(_activeNodes.Count != 0, "Still 0 nodes after asking for a new node.  Build cannot proceed.");
-                            }
-                            else
-                            {
-                                BuildEventContext buildEventContext = new BuildEventContext(0, Scheduler.VirtualNode, BuildEventContext.InvalidProjectInstanceId, BuildEventContext.InvalidProjectContextId, BuildEventContext.InvalidTargetId, BuildEventContext.InvalidTaskId);
-                                ((IBuildComponentHost)this).LoggingService.LogError(buildEventContext, new BuildEventFileInfo(String.Empty), "UnableToCreateNode", response.RequiredNodeType.ToString("G"));
+                                NodeInfo createdNode = _nodeManager.CreateNode(GetNodeConfiguration(), response.RequiredNodeType);
 
-                                throw new BuildAbortedException(ResourceUtilities.FormatResourceStringStripCodeAndKeyword("UnableToCreateNode", response.RequiredNodeType.ToString("G")));
+                                if (createdNode != null)
+                                {
+                                    _noNodesActiveEvent.Reset();
+                                    _activeNodes.TryAdd(createdNode.NodeId, 0);
+                                    newNodes.Add(createdNode);
+                                    ErrorUtilities.VerifyThrow(_activeNodes.Count != 0, "Still 0 nodes after asking for a new node.  Build cannot proceed.");
+                                }
+                                else
+                                {
+                                    BuildEventContext buildEventContext = new BuildEventContext(0, Scheduler.VirtualNode, BuildEventContext.InvalidProjectInstanceId, BuildEventContext.InvalidProjectContextId, BuildEventContext.InvalidTargetId, BuildEventContext.InvalidTaskId);
+                                    ((IBuildComponentHost)this).LoggingService.LogError(buildEventContext, new BuildEventFileInfo(String.Empty), "UnableToCreateNode", response.RequiredNodeType.ToString("G"));
+
+                                    throw new BuildAbortedException(ResourceUtilities.FormatResourceStringStripCodeAndKeyword("UnableToCreateNode", response.RequiredNodeType.ToString("G")));
+                                }
+                            });
+                        }
+                        catch (AggregateException e)
+                        {
+                            foreach (Exception ex in e.InnerExceptions)
+                            {
+                                throw ex;
                             }
                         }
 
--- a/src/Build/BackEnd/BuildManager/BuildManager.cs
+++ b/src/Build/BackEnd/BuildManager/BuildManager.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Build.Execution
         /// <summary>
         /// Set of active nodes in the system.
         /// </summary>
-        private readonly ConcurrentDictionary<int, byte> _activeNodes;
+        private readonly HashSet<NGen<int>> _activeNodes;
 
         /// <summary>
         /// Event signalled when all nodes have shutdown.
@@ -263,7 +263,7 @@ namespace Microsoft.Build.Execution
             _buildSubmissions = new Dictionary<int, BuildSubmission>();
             _graphBuildSubmissions = new Dictionary<int, GraphBuildSubmission>();
             _noActiveSubmissionsEvent = new AutoResetEvent(true);
-            _activeNodes = new ConcurrentDictionary<int, byte>();
+            _activeNodes = new HashSet<NGen<int>>();
             _noNodesActiveEvent = new AutoResetEvent(true);
             _nodeIdToKnownConfigurations = new Dictionary<NGen<int>, HashSet<NGen<int>>>();
             _unnamedProjectInstanceToNames = new Dictionary<ProjectInstance, string>();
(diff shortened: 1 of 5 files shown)


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1924,24 +1925,34 @@ private void PerformSchedulingActions(IEnumerable<ScheduleResponse> responses)
...
-                                _noNodesActiveEvent.Reset();
-                                _activeNodes.Add(createdNode.NodeId);
-                                newNodes.Add(createdNode);
-                                ErrorUtilities.VerifyThrow(_activeNodes.Count != 0, "Still 0 nodes after asking for a new node.  Build cannot proceed.");
-                            }
-                            else
-                            {
-                                BuildEventContext buildEventContext = new BuildEventContext(0, Scheduler.VirtualNode, BuildEventContext.InvalidProjectInstanceId, BuildEventContext.InvalidProjectContextId, BuildEventContext.InvalidTargetId, BuildEventContext.InvalidTaskId);
-                                ((IBuildComponentHost)this).LoggingService.LogError(buildEventContext, new BuildEventFileInfo(String.Empty), "UnableToCreateNode", response.RequiredNodeType.ToString("G"));
+                                NodeInfo createdNode = _nodeManager.CreateNode(GetNodeConfiguration(), response.RequiredNodeType);
```
rainersigwald: I don't think `_nodeManager` or the `NodeProviderOutOfProc` it calls into is thread safe, but they're being called in parallel here.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1923,25 +1924,35 @@ private void PerformSchedulingActions(IEnumerable<ScheduleResponse> responses)
                         break;
 
                     case ScheduleActionType.CreateNode:
-                        var newNodes = new List<NodeInfo>();
-
-                        for (int i = 0; i < response.NumberOfNodesToCreate; i++)
+                        var newNodes = new ConcurrentBag<NodeInfo>();
```
cdmihai: nit: this case branch is becoming big, I'd extract it to a new method.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1923,25 +1924,35 @@ private void PerformSchedulingActions(IEnumerable<ScheduleResponse> responses)
...
+                        GetNodeConfiguration();
```
cdmihai: Why are you calling this without consuming its return value?

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -102,7 +103,7 @@ public class BuildManager : INodePacketHandler, IBuildComponentHost, IDisposable
         /// <summary>
         /// Set of active nodes in the system.
         /// </summary>
-        private readonly HashSet<NGen<int>> _activeNodes;
+        private readonly ConcurrentDictionary<int, byte> _activeNodes;
```
cdmihai: To minimize the change, and to not give false impressions that BuildManager handles packets in parallel, you can avoid making this a concurrent data structure by hiding the parallel node creation with local concurrent data structures which you then dump into `_activeNodes` after all nodes get created.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1923,25 +1924,35 @@ private void PerformSchedulingActions(IEnumerable<ScheduleResponse> responses)
...
+                                    ErrorUtilities.VerifyThrow(_activeNodes.Count != 0, "Still 0 nodes after asking for a new node.  Build cannot proceed.");
+                                }
+                                else
+                                {
+                                    BuildEventContext buildEventContext = new BuildEventContext(0, Scheduler.VirtualNode, BuildEventContext.InvalidProjectInstanceId, BuildEventContext.InvalidProjectContextId, BuildEventContext.InvalidTargetId, BuildEventContext.InvalidTaskId);
+                                    ((IBuildComponentHost)this).LoggingService.LogError(buildEventContext, new BuildEventFileInfo(String.Empty), "UnableToCreateNode", response.RequiredNodeType.ToString("G"));
 
-                                throw new BuildAbortedException(ResourceUtilities.FormatResourceStringStripCodeAndKeyword("UnableToCreateNode", response.RequiredNodeType.ToString("G")));
+                                    throw new BuildAbortedException(ResourceUtilities.FormatResourceStringStripCodeAndKeyword("UnableToCreateNode", response.RequiredNodeType.ToString("G")));
```
cdmihai: nit: AFAIK Parallel.For bundles all exceptions from all threads in an AggregateException. Might want to catch that and rethrow the containing exceptions to avoid changing the exceptions that API users will get.

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: Yup, would need to make the whole object tree under `NodeManager` thread safe.

cdmihai:
>             _nodesShutdown = false;

This write is not thread safe anymore.

---
Refers to: src/Build/BackEnd/Components/Communications/NodeManager.cs:124 in 247bc78. [](commit_id = 247bc788484813ab091eb2f69d8ffcdf195031db, deletion_comment = False)

src/Build/BackEnd/BuildManager/BuildManager.cs
Forgind: `GetNodeConfiguration` calculates and returns _nodeConfiguration, but it caches it so that if it's already been calculated, it doesn't calculate it again. This makes sure that it's already been calculated.

src/Build/BackEnd/BuildManager/BuildManager.cs
Forgind: I'm confused. Is that not what I did? I have a try block around the parallel for, and the catch (of an AggregateException) just throws all the inner exceptions.

Forgind:
I think (?) I remember having looked through for what possible types the _nodeManager could be and found that the only way `_nodesShutdown` could be read or written to is in exactly where you noticed it, in which case any interleaving of `_nodesShutdown = false` with `_nodesShutdown = false` is equally valid.

src/Build/BackEnd/Components/Communications/NodeManager.cs
```diff
@@ -361,6 +367,18 @@ private int AttemptCreateNode(INodeProvider nodeProvider, NodeConfiguration node
...
+                {
+                    ((NodeProviderInProc)nodeProvider)._nodeAttemptInProgress = false;
+                }
+                else if (nodeProvider is NodeProviderOutOfProc)
+                {
+                    ((NodeProviderOutOfProc)nodeProvider)._nodeAttemptsInProgress--;
+                }
+            }
+
             _nodeIdToProvider.Add(nodeId, nodeProvider);
```
cdmihai: `_nodeIdToProvider` not thread safe

cdmihai:
> I think (?) I remember having looked through for what possible types the _nodeManager could be and found that the only way `_nodesShutdown` could be read or written to is in exactly where you noticed it, in which case any interleaving of `_nodesShutdown = false` with `_nodesShutdown = false` is equally valid.

It's set [here](https://github.com/cdmihai/msbuild/blob/9c33693c492a0cb99474dcb703bfd0947056d8a9/src/Build/BackEnd/Components/Communications/NodeManager.cs#L158), but I didn't look at the interleavings. Since code shifts around over time, probably safest to just make it thread safe. Otherwise any small change can start to have hard to understand consequences.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -1895,6 +1896,54 @@ private void CheckForActiveNodesAndCleanUpSubmissions()
...
+        /// Handles creating new nodes.
+        /// </summary>
+        /// <param name="response">Details of how to make the nodes (the number, type, and so on).</param>
+        private void CreateNodes(ScheduleResponse response)
+        {
+            var newNodes = new ConcurrentBag<NodeInfo>();
+            GetNodeConfiguration();
+            try
+            {
+                Parallel.For(0, response.NumberOfNodesToCreate, i =>
```
cdmihai: Do you have any idea whether it's common to have multiple nodes requested by the scheduler at the same time, and how much time parallelization would save? If it's the common case, then the extra parallelization complexity might make sense. If not, it might be better to have less complex code.

cdmihai:
>                 _nodeContexts[nodeId] = context;

`_nodeContexts` is not thread safe

---
Refers to: src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs:130 in 247bc78. [](commit_id = 247bc788484813ab091eb2f69d8ffcdf195031db, deletion_comment = False)

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -188,12 +192,15 @@ protected NodeContext GetNode(string msbuildLocation, string commandLineArgs, in
                     _processesToIgnore.Add(nodeLookupKey);
```
cdmihai: You could get the same key added multiple times in the bag. Won't affect correctness, but maybe it makes the code more clear if you use `ConcurrentDictionary.GetOrAdd`

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -188,12 +192,15 @@ protected NodeContext GetNode(string msbuildLocation, string commandLineArgs, in
...
 
                     // Attempt to connect to each process in turn.
-                    Stream nodeStream = TryConnectToProcess(nodeProcess.Id, 0 /* poll, don't wait for connections */, hostHandshake, clientHandshake);
-                    if (nodeStream != null)
+                    lock (createNodeLock)
```
cdmihai: This will serialize all parallel attempts to connect to existing nodes. I'd change it to have one thread per nodeid trying to connect. First one to connect wins and the others back out and relinquish the node. But it might get too hairy.

cdmihai:
>             CommunicationsUtilities.Trace("Could not connect to existing process, now creating a process...");

All these trace messages will get duplicated by each thread. I'd prepend the thread id to aid debugging.

---
Refers to: src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs:213 in 247bc78. [](commit_id = 247bc788484813ab091eb2f69d8ffcdf195031db, deletion_comment = False)

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -229,10 +236,13 @@ protected NodeContext GetNode(string msbuildLocation, string commandLineArgs, in
                     }
                 }
 #endif
-
-                // Create the node process
-                int msbuildProcessId = LaunchNode(msbuildLocation, commandLineArgs);
-                _processesToIgnore.Add(GetProcessesToIgnoreKey(hostHandshake, clientHandshake, msbuildProcessId));
+                int msbuildProcessId = -1;
+                lock (createNodeLock)
```
cdmihai: I think you can remove the lock, everything in `LaunchNode` seems stack confined.

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: Ah, my bad, didn't notice the loop in the catch block :blush:

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: Sounds good. Can you add a quick comment? Or rename method to something like `GetAndCacheNodeConfiguration` to avoid the comment.

src/Build/BackEnd/BuildManager/BuildManager.cs
Forgind: I believe it will be the default on .NET Core. Ultimately, creating a node itself doesn't take too much time; this is more important in the unusual case in which nodes are repeatedly failing right after they're made as in #860. What appeared to be a hang was probably just up to 32 consecutive 30-second timeouts—one per failed node.

I'd be happy to come over and talk about it if you think we might want to go with a change that's easier to get right.

cdmihai: I think in #860 you get a true hang, not a timeout apparent hang. If the main node connects to the child, but the child does not write any handshake back, the main node blocks indefinitely on full framework. I see that [on .net core there's a timeout](https://github.com/microsoft/msbuild/blob/0f85cc7476897ff7bb6f099a1b78ab5fff64b50c/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs#L348-L352), so probably full framework does not have the timeout API.

Thinking back on the main problem, maybe parallel creation is not the ideal fix here. If the main node fails to spawn a child node due to a child node corruption, there's a high chance all nodes will fail that way, since the way the main node sets up a child node does not vary. One caveat is that, before the child node writes a handshake, we don't know whether it's corrupted, or whether the machine is so slow the node didn't get to it yet. Maybe a better way is to:
- fail the build if two node creations failed with the same non-timeout issues.
- for slow machines, figure out some better solution. For example maybe temporarily create the node processes with super high priority so the OS scheduler is forced to schedule their process, and then lower their priority right after they've replied their handshake. This way maybe we can do without the timeout, or have a much shorter timeout.

Forgind:
This is probably the wrong direction. @cdmihai suggested creating child nodes with high priority instead, and since that's a much safer direction, I will look into that further.

PR #4911, opened 2019-11-12
Add intrinsic property function to compare versions

### LINKED ISSUE #3212: Need a performace and unified way to compare versions
In msbuild condition. msbuild build will try to [parse it as dec or hex ]( https://github.com/Microsoft/msbuild/blob/e8b480c373eadd9daa975c6e96e4dbbabcd9a4fe/src/Build/Evaluation/Conditionals/NumericComparisonExpressionNode.cs#L44)first. So _2.11_ will be smaller than _2.2_.

```xml
<Project ToolsVersion="15.0" DefaultTargets="Try">
  <Target Name="Try">
   <Error Condition=" '2.11' &lt; 2.2 "
           Text='unexpected' />
  </Target>
</Project>
```

At the same time. If appending 0 at the end as in _2.2.0_, it will cause parse as number fail and consider it as Version. However, Version class's comparison does not consider _2.1_ and _2.1.0_ to be equal.[ In fact ](
https://msdn.microsoft.com/en-us/library/s3bf0xb2(v=vs.110).aspx)_2.1 < 2.1.0_. This will also occur in Property Function. 

```xml
<Project ToolsVersion="15.0" DefaultTargets="Try">
  <Target Name="Try">
   <Error Condition=" '2.1' &lt; 2.1.0 "
           Text='unexpected' />
  </Target>
</Project>
```

It is different behavior of [NuGet version](https://docs.microsoft.com/en-us/nuget/reference/package-versioning).

Also there is performance concern.

nguerrera:
Usage: 
```
$([MSBuild]::VersionEquals(a, b))
$([MSBuild]::VersionNotEquals(a, b))
$([MSBuild]::VersionLessThan(a, b))
$([MSBuild]::VersionLessThanOrEquals(a, b))
$([MSBuild]::VersionLessThan(a, b))
$([MSBuild]::VersionGreaterThanOrEquals(a, b))
```

Versions are parsed like System.Version, with the following exceptions:

 * Leading v or V is ignored. Allows comparison to $(TargetFrameworkVersion)

 * Everything from first '-' or '+' to end of string is ignored. Allows passing in semantic versions, though the order is not the same as semver. Instead, prerelease specified and build metadata do not have any sorting weight. This can be useful, for example, to turn on a feature for >= x.y and have it kick in on x.y.z-pre.

 * Remaining part of string from above is parsed with Version.Parse, then normalized to have 4 explicit parts to make x == x.0 == x.0.0 == x.0.0.0.

 * Whitespace is not allowed in integer components

 * major version only is valid ("3" is equal to "3.0.0.0")

 * `+`  is not allowed as positive sign in integer components (it is treated as semver metadata and ignored per above)

The parser is adapted from System.Version source code, but simplified and tweaked to handle quirks above, with trivial opportunities to improve perf / reduce allocations taken. Tests are also adapted from corefx.

Fix #3212

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -2790,6 +2790,77 @@ namespace Microsoft.Build.UnitTests.Evaluation
             Assert.Equal("True", result);
         }
 
+        [Theory]
+        [InlineData("NotAVersion")]
+        [InlineData("1.2.3.4.5")]
+        [InlineData("1,2,3,4")]
+        public void PropertyFunctionVersionComparisonsFailsWithInvalidArguments(string badVersion)
+        {
+            var pg = new PropertyDictionary<ProjectPropertyInstance>();
+            var expander = new Expander<ProjectPropertyInstance, ProjectItemInstance>(pg, FileSystems.Default);
+            string expectedMessage = ResourceUtilities.GetResourceString("InvalidVersionFormat");
+
+            AssertThrows($"$([MSBuild]::VersionGreaterThan('{badVersion}', '1.0.0'))");
+            AssertThrows($"$([MSBuild]::VersionGreaterThan('1.0.0', '{badVersion}'))");
+
+            AssertThrows($"$([MSBuild]::VersionGreaterThanOrEquals('{badVersion}', '1.0.0'))");
+            AssertThrows($"$([MSBuild]::VersionGreaterThanOrEquals('1.0.0', '{badVersion}'))");
+
+            AssertThrows($"$([MSBuild]::VersionLessThan('{badVersion}', '1.0.0'))");
+            AssertThrows($"$([MSBuild]::VersionLessThan('1.0.0', '{badVersion}'))");
+
+            AssertThrows($"$([MSBuild]::VersionLessThanOrEquals('{badVersion}', '1.0.0'))");
+            AssertThrows($"$([MSBuild]::VersionLessThanOrEquals('1.0.0', '{badVersion}'))");
+
+            AssertThrows($"$([MSBuild]::VersionEquals('{badVersion}', '1.0.0'))");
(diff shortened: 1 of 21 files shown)


src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2790,6 +2790,37 @@ public void IsOsPlatformShouldBeCaseInsensitiveToParameter()
             Assert.Equal("True", result);
         }
 
+        [Theory]
+        [InlineData("vv1.0")]
+        [InlineData("NotAVersion")]
+        [InlineData("1.2.3.4~5")]
```
rainersigwald: Add a test for a bare number like `3`?

nguerrera:
```
Microsoft (R) Visual C# Interactive Compiler version 3.5.0-beta1-19558-01 ()

Type "#help" for more information.
> System.Version.Parse("3")
System.ArgumentException: Version string portion was too short or too long.
```

Good catch. I think this should work, but basically requires that we do not use Version.Parse, but rather have our own parser. Which was something I was only wanting to do if there was a significant perf win.

nguerrera:
I like the idea of having better ergonomics for conditions.

So, how about all of these:

VersionGreaterThan
VersionGreaterThanOrEqual
VersionLessThan
VersionLessThanOrEqual
VersionEqual

Do we even need the one that returns an int if we have all of the above?

I added Than because VersionLess read like version-less (as in having no version) to me.

nguerrera:
Perhaps, VersionNotEquals is unnecessary since leading `!$([MSBuild]::VersionEquals(a, b))` is pretty readable and fewer characters anyway.

eerhardt:
Did we consider having a `IsVersion` or `TryParse` kind of method along with these? That way someone could use it for `$(LangVersion)`, which may be of the form `x.y`, but it also may be `latest`, `preview`, etc.

PR #4916, opened 2019-11-13
Load plugins in an AssemblyLoadContext

### LINKED ISSUE #1754: Each MSBuild task should exist in its own AssemblyLoadContext
Each MSBuild Task should load in its own AssemblyLoadContext. At least 3rd party tasks. This prevents build failures due to two independently developed tasks sharing a dependency but with different versions.

Also, the default AssemblyLoadContext should be very nearly empty, so that each Task can define its own dependencies instead of inheriting them from the default context. For example, Roslyn assemblies should *not* be in the default AssemblyLoadContext so that MSBuild tasks can use Roslyn assemblies of the version it was compiled against.

@tmat summarized this well after [a discussion we had in another issue](https://github.com/Microsoft/msbuild/issues/1737#issuecomment-282571709).

CC: @nguerrera

### LINKED ISSUE #4635: AssemblyLoadContext for loggers
#1754, but for loggers, which have the same constraints: user-controlled assemblies with arbitrary assembly dependencies.

rainersigwald:
Use `AssemblyLoadContext` and the new `AssemblyDependencyResolver` to allow task implementations to have independent, incompatible dependencies on .NET Core.

Follows the spec from #4133 and the [tutorial on creating .NET Core plugin hosts](https://docs.microsoft.com/en-us/dotnet/core/tutorials/creating-app-with-plugin-support).

Since `AssemblyDependencyResolver` is a .NET Core 3.0 type, required targeting .NET Core 3.0.

Fixes #1754 (task ALC)
Fixes #4635 (logger ALC)

### DIFF
--- a/src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
            );
         }
 
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
         /// <summary>
         /// Test that an exception is thrown when the path to the task assembly is empty
         /// </summary>
@@ -267,7 +267,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
             TaskHostConfiguration deserializedConfig = packet as TaskHostConfiguration;
 
             Assert.Equal(config.TaskName, deserializedConfig.TaskName);
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
             Assert.Equal(config.TaskLocation, deserializedConfig.TaskLocation);
 #endif
             Assert.Null(deserializedConfig.TaskParameters);
@@ -305,7 +305,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
             TaskHostConfiguration deserializedConfig = packet as TaskHostConfiguration;
 
             Assert.Equal(config.TaskName, deserializedConfig.TaskName);
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
             Assert.Equal(config.TaskLocation, deserializedConfig.TaskLocation);
 #endif
             Assert.NotNull(deserializedConfig.TaskParameters);
@@ -348,7 +348,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
             TaskHostConfiguration deserializedConfig = packet as TaskHostConfiguration;
 
             Assert.Equal(config.TaskName, deserializedConfig.TaskName);
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
             Assert.Equal(config.TaskLocation, deserializedConfig.TaskLocation);
 #endif
             Assert.NotNull(deserializedConfig.TaskParameters);
@@ -389,7 +389,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
             TaskHostConfiguration deserializedConfig = packet as TaskHostConfiguration;
 
             Assert.Equal(config.TaskName, deserializedConfig.TaskName);
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
             Assert.Equal(config.TaskLocation, deserializedConfig.TaskLocation);
 #endif
             Assert.NotNull(deserializedConfig.TaskParameters);
@@ -429,7 +429,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
             TaskHostConfiguration deserializedConfig = packet as TaskHostConfiguration;
 
             Assert.Equal(config.TaskName, deserializedConfig.TaskName);
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
             Assert.Equal(config.TaskLocation, deserializedConfig.TaskLocation);
 #endif
             Assert.NotNull(deserializedConfig.TaskParameters);
--- a/src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
+++ b/src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if FEATURE_ASSEMBLYLOADCONTEXT
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Shared;
+using Microsoft.Build.Utilities;
+using System.Runtime.Loader;
+
+namespace Microsoft.Build.UnitTests
+{
+    public class ValidateAssemblyLoadContext : Task
+    {
+        public override bool Execute()
+        {
+            var thisLoadContext = AssemblyLoadContext.GetLoadContext(typeof(ValidateAssemblyLoadContext).Assembly);
+
+            // Check by name because this always reports false, presumably due to ALC shenanigans:
+            // if (thisLoadContext is MSBuildLoadContext context)
+            if (thisLoadContext.GetType().FullName == typeof(MSBuildLoadContext).FullName)
+            {
+#if NETCOREAPP && !NETCOREAPP2_1 // TODO: enable this functionality when targeting .NET Core 3.0+
+                if (!thisLoadContext.Name.EndsWith(typeof(ValidateAssemblyLoadContext).Assembly.GetName().Name + ".dll"))
+                {
+                    Log.LogError($"Unexpected AssemblyLoadContext name: \"{thisLoadContext.Name}\", but the current executing assembly was {typeof(ValidateAssemblyLoadContext).Assembly.GetName().Name}");
+                }
+                else
+                {
+                    Log.LogMessage(MessageImportance.High, $"Task {nameof(ValidateAssemblyLoadContext)} loaded in AssemblyLoadContext named {thisLoadContext.Name}");
+                }
+#endif
+            }
+            else
+            {
+                Log.LogError($"Load context was a {thisLoadContext.GetType().AssemblyQualifiedName} instead of an {typeof(MSBuildLoadContext).AssemblyQualifiedName}");
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+    }
+}
+#endif
--- a/src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
+++ b/src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
@@ -16,8 +16,11 @@ namespace Microsoft.Build.UnitTests
         {
             var thisLoadContext = AssemblyLoadContext.GetLoadContext(typeof(ValidateAssemblyLoadContext).Assembly);
 
-            // Check by name because this always reports false, presumably due to ALC shenanigans:
-            // if (thisLoadContext is MSBuildLoadContext context)
+            // The straightforward implementation of this check:
+            //   if (thisLoadContext is MSBuildLoadContext context)
+            // fails here because MSBuildLoadContext (in this test assembly) is from MSBuild.exe via
+            // IVT, but the one that actually gets used for task isolation is in Microsoft.Build.dll.
+            // This probably doesn't need to be how it is forever: https://github.com/microsoft/msbuild/issues/5041
             if (thisLoadContext.GetType().FullName == typeof(MSBuildLoadContext).FullName)
             {
 #if NETCOREAPP && !NETCOREAPP2_1 // TODO: enable this functionality when targeting .NET Core 3.0+
--- a/src/MSBuild/MSBuild.csproj
+++ b/src/MSBuild/MSBuild.csproj
@@ -138,6 +138,7 @@
     <Compile Include="..\Shared\OutOfProcTaskHostTaskResult.cs" />
     <Compile Include="..\Shared\TaskHostTaskCancelled.cs" />
     <Compile Include="..\Shared\TaskLoader.cs" />
+    <Compile Include="..\Shared\MSBuildLoadContext.cs" Condition="'$(TargetFrameworkIdentifier)'!='.NETFramework'" />
     <Compile Include="..\Shared\TypeLoader.cs" />
     <Compile Include="..\Shared\LoadedType.cs">
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -5,8 +5,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
-using System.Runtime.Loader;
-using Microsoft.Build.Shared.FileSystem;
 
 namespace Microsoft.Build.Shared
 {
@@ -16,35 +14,8 @@ namespace Microsoft.Build.Shared
     internal sealed class CoreClrAssemblyLoader
     {
         private readonly Dictionary<string, Assembly> _pathsToAssemblies = new Dictionary<string, Assembly>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, Assembly> _namesToAssemblies = new Dictionary<string, Assembly>();
-        private readonly HashSet<string> _dependencyPaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         private readonly object _guard = new object();
 
-        private bool _resolvingHandlerHookedUp = false;
-
-        private static readonly string[] _extensions = new[] { "ni.dll", "ni.exe", "dll", "exe" };
-        private static readonly Version _currentAssemblyVersion = new Version(Microsoft.Build.Shared.MSBuildConstants.CurrentAssemblyVersion);
-        private static readonly HashSet<string> _wellKnownAssemblyNames = new HashSet<string>(
-            new[]
-            {
-                "Microsoft.Build",
-                "Microsoft.Build.Framework",
-                "Microsoft.Build.Tasks.Core",
-                "Microsoft.Build.Utilities.Core"
-            });
-
-        public void AddDependencyLocation(string fullPath)
-        {
-            if (fullPath == null)
-            {
-                throw new ArgumentNullException(nameof(fullPath));
-            }
-
-            lock (_guard)
-            {
-                _dependencyPaths.Add(fullPath);
-            }
-        }
 
         public Assembly LoadFromPath(string fullPath)
         {
@@ -57,113 +28,19 @@ namespace Microsoft.Build.Shared
 
             lock (_guard)
             {
-                if (!_resolvingHandlerHookedUp)
-                {
-                    AssemblyLoadContext.Default.Resolving += TryResolveAssembly;
-                    _resolvingHandlerHookedUp = true;
-                }
-
                 Assembly assembly;
                 if (_pathsToAssemblies.TryGetValue(fullPath, out assembly))
                 {
                     return assembly;
                 }
 
-                return LoadAndCache(AssemblyLoadContext.Default, fullPath);
+
+                var contextForAssemblyPath = new MSBuildLoadContext(fullPath);
+
+                assembly = contextForAssemblyPath.LoadFromAssemblyPath(fullPath);
+
+                return assembly;
             }
-        }
-
-        private Assembly TryGetWellKnownAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
-        {
-            if (!_wellKnownAssemblyNames.Contains(assemblyName.Name))
-            {
-                return null;
-            }
-
-            // Ensure we are attempting to load a matching version
-            // of the Microsoft.Build.* assembly.
-            assemblyName.Version = _currentAssemblyVersion;
-
-            var searchPaths = new[] { Assembly.GetExecutingAssembly().Location };
-            return TryResolveAssemblyFromPaths(context, assemblyName, searchPaths);
-        }
-
-        private Assembly TryResolveAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
-        {
-            lock (_guard)
-            {
-                Assembly assembly = TryGetWellKnownAssembly(context, assemblyName);
-
-                if (assembly != null)
-                {
-                    return assembly;
-                }
-
-                if (_namesToAssemblies.TryGetValue(assemblyName.FullName, out assembly))
-                {
-                    return assembly;
-                }
-
-                return TryResolveAssemblyFromPaths(context, assemblyName, _dependencyPaths);
-            }
-        }
-
-        private Assembly TryResolveAssemblyFromPaths(AssemblyLoadContext context, AssemblyName assemblyName, IEnumerable<string> searchPaths)
-        {
-            foreach (var cultureSubfolder in string.IsNullOrEmpty(assemblyName.CultureName)
-                // If no culture is specified, attempt to load directly from
-                // the known dependency paths.
-                ? new[] { string.Empty }
-                // Search for satellite assemblies in culture subdirectories
-                // of the assembly search directories, but fall back to the
-                // bare search directory if that fails.
-                : new[] { assemblyName.CultureName, string.Empty })
-            {
-                foreach (var searchPath in searchPaths)
-                {
-                    foreach (var extension in _extensions)
-                    {
-                        var candidatePath = Path.Combine(searchPath,
-                            cultureSubfolder,
-                            $"{assemblyName.Name}.{extension}");
-
-                        if (IsAssemblyAlreadyLoaded(candidatePath) ||
-                            !FileSystems.Default.FileExists(candidatePath))
-                        {
-                            continue;
-                        }
-
-                        AssemblyName candidateAssemblyName = AssemblyLoadContext.GetAssemblyName(candidatePath);
-                        if (candidateAssemblyName.Version != assemblyName.Version)
-                        {
-                            continue;
-                        }
-
-                        return LoadAndCache(context, candidatePath);
-                    }
-                }
-            }
-
-            return null;
-        }
-
-        /// <remarks>
-        /// Assumes we have a lock on _guard
-        /// </remarks>
-        private Assembly LoadAndCache(AssemblyLoadContext context, string fullPath)
-        {
-            var assembly = context.LoadFromAssemblyPath(fullPath);
-            var name = assembly.FullName;
-
-            _pathsToAssemblies[fullPath] = assembly;
-            _namesToAssemblies[name] = assembly;
-
-            return assembly;
-        }
-
-        private bool IsAssemblyAlreadyLoaded(string path)
-        {
-            return _pathsToAssemblies.ContainsKey(path);
         }
     }
 }
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -34,10 +34,14 @@ namespace Microsoft.Build.Shared
                     return assembly;
                 }
 
-
                 var contextForAssemblyPath = new MSBuildLoadContext(fullPath);
 
                 assembly = contextForAssemblyPath.LoadFromAssemblyPath(fullPath);
+
+                if (assembly != null)
+                {
+                    _pathsToAssemblies[fullPath] = assembly;
+                }
 
                 return assembly;
             }
--- a/src/Shared/MSBuildLoadContext.cs
+++ b/src/Shared/MSBuildLoadContext.cs
@@ -78,6 +78,18 @@ namespace Microsoft.Build.Shared
                 }
             }
 
+            // If the Assembly is provided via a file path, the following rules are used to load the assembly:
+            // - if the simple name of the assembly exists in the same folder as msbuild.exe, then that assembly gets loaded, indifferent of the user specified path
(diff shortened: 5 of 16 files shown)


src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
```diff
@@ -336,7 +336,7 @@ public void TestTranslationWithEmptyDictionary()
             TaskHostConfiguration deserializedConfig = packet as TaskHostConfiguration;
 
             Assert.Equal(config.TaskName, deserializedConfig.TaskName);
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
             Assert.Equal(config.TaskLocation, config.TaskLocation);
```
rainersigwald: These look fishy: separate PR to make it

```suggestion
            Assert.Equal(config.TaskLocation, deserializedConfig.TaskLocation);
```
?

src/MSBuild/MSBuild.csproj
```diff
@@ -13,6 +13,9 @@
          versions of this project -->
     <RuntimeIdentifiers>win7-x86;win7-x64</RuntimeIdentifiers>
 
+    <!-- Having the .exe around confuses things -->
```
rainersigwald:
```suggestion
    <!-- Don't produce the .exe, because the CLI directly invokes the DLL and so should tests -->
```

src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
rainersigwald: Done in #4297

rainersigwald:
@vitek-karas would you mind taking a look at this PR using `AssemblyDependencyResolver` in MSBuild?

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,69 @@
...
+                "MSBuild",
+                "Microsoft.Build",
+                "Microsoft.Build.Framework",
+                "Microsoft.Build.Tasks.Core",
+                "Microsoft.Build.Utilities.Core",
+            }.ToImmutableHashSet();
+
+
+        public MSBuildLoadContext(string assemblyPath) :
+            base(name: assemblyPath, isCollectible: false) // TODO: make this collectible?
```
vitek-karas: Is there some way the MSBuild can decide to unload a task? Is there a natural trigger for it? If the answer is no, then don't enable unloading - just enabling it comes with a perf cost (not a big one, but still).

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.Build.Shared
+{
+    /// <summary>
+    /// This class is used to isolate the types used by an MSBuild plugin
+    /// (SDK resolver, logger, or task).
+    /// </summary>
+    internal class MSBuildLoadContext : AssemblyLoadContext
+    {
+        private readonly AssemblyDependencyResolver _resolver;
+        private readonly string _directory;
```
vitek-karas: This is not a directory - it holds the full path to the task dll.

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,69 @@
...
+                "Microsoft.Build.Tasks.Core",
+                "Microsoft.Build.Utilities.Core",
+            }.ToImmutableHashSet();
+
+
+        public MSBuildLoadContext(string assemblyPath) :
+            base(name: assemblyPath, isCollectible: false) // TODO: make this collectible?
+        {
+            _directory = assemblyPath;
+            _resolver = new AssemblyDependencyResolver(_directory);
```
vitek-karas: I just want to make sure this is intentional - this will look for `task.deps.json` next to the `task.dll` and if it's there it will use it - new and potentially breaking behavior.
If it's not there it will use all `.dll` files in the same folder as the dependencies - which is probably closer to the current behavior.

I'm not saying we should not do this - in fact not using the `.deps.json` will basically fix only half of the problem we could fix with this change. I'm just pointing out that this is potentially a breaking change.

src/Shared/TypeLoader.cs
```diff
@@ -152,35 +152,14 @@ private static Assembly LoadAssembly(AssemblyLoadInfo assemblyLoadInfo)
...
                 }
                 else
                 {
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
                     loadedAssembly = Assembly.UnsafeLoadFrom(assemblyLoadInfo.AssemblyFile);
 #else
-                    // If the Assembly is provided via a file path, the following rules are used to load the assembly:
-                    // - if the simple name of the assembly exists in the same folder as msbuild.exe, then that assembly gets loaded, indifferent of the user specified path
-                    // - otherwise, the assembly from the user specified path is loaded, if it exists.
```
vitek-karas: Similarly we're making a breaking change here.
Again, probably a good thing - but I think we need to be intentional about it.

src/Shared/MSBuildLoadContext.cs
rainersigwald: Talked about this with @livarcocc. We're planning to be conservative now (use the same behavior but in an ALC) and reconsider adding an `AssemblyDependencyResolver` later (probably in the .NET 5 timeframe). We agree that it's better moving forward but don't want to break folks if possible.

src/Shared/MSBuildLoadContext.cs
rainersigwald: This is moot now that I backed down to 2.1 and don't have the option, but: we don't have a great place to unload tasks. There's a fairly natural point for it--when a worker node is released, or when a build completes--but we're also likely to want to go past that point for the sake of performance: we don't really want to slow down inner-loop builds reloading `Csc` for every build. So I think we should just not make it collectible.

src/Shared/TypeLoader.cs
rainersigwald: In the implementation I just pushed I added the MSBuild directory as a fallback within the ALC, which I think makes sense as a middle ground between "keep task assemblies with poorly-packaged dependencies happy" (mostly, unless they deeply depended on the overriding) and "best isolation" (a well-packaged task can use a different version of something in our directory now).

src/Shared/MSBuildLoadContext.cs
vitek-karas: This is now resolved - as it is now a directory.

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,96 @@
...
+            // If the Assembly is provided via a file path, the following rules are used to load the assembly:
+            // - if the simple name of the assembly exists in the same folder as msbuild.exe, then that assembly gets loaded, indifferent of the user specified path
+            // - otherwise, the assembly from the user specified path is loaded, if it exists.
+
+            var assemblyNameInExecutableDirectory = Path.Combine(BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory,
+                assemblyName.Name);
+
+            if (FileSystems.Default.FileExists(assemblyNameInExecutableDirectory))
+            {
+                return LoadFromAssemblyPath(assemblyNameInExecutableDirectory);
```
vitek-karas: This will load a "copy" of the assembly into the task's ALC - it will not "share" it across ALCs.
Let's assume that there's an assembly "BuildUtils.dll" which lives in the msbuild directory. With the code as is, each task which has a dependency on `BuildUtils` will load its own copy of `BuildUtils`.
Not sure if this is the intended behavior - if so, I would explicitly state it in the comment.

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                // Force MSBuild assemblies to be loaded in the default ALC
+                // and unify to the current version.
+                return null;
+            }
+
+            foreach (var cultureSubfolder in string.IsNullOrEmpty(assemblyName.CultureName)
+                // If no culture is specified, attempt to load directly from
+                // the known dependency paths.
+                ? new[] { string.Empty }
```
cdmihai: nit: can cache the empty string array

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,96 @@
...
+                // Search for satellite assemblies in culture subdirectories
+                // of the assembly search directories, but fall back to the
+                // bare search directory if that fails.
+                : new[] { assemblyName.CultureName, string.Empty })
+            {
+                foreach (var extension in _extensions)
```
cdmihai: what if multiple extensions exist? does the array order actually signify a priority? If so, put a comment on it :)

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,96 @@
...
+                        continue;
+                    }
+
+                    return LoadFromAssemblyPath(candidatePath);
+                }
+            }
+
+            // If the Assembly is provided via a file path, the following rules are used to load the assembly:
+            // - if the simple name of the assembly exists in the same folder as msbuild.exe, then that assembly gets loaded, indifferent of the user specified path
+            // - otherwise, the assembly from the user specified path is loaded, if it exists.
```
cdmihai: is this `- otherwise` case implemented by returning null and letting the base context resolve it? Would it be more explicit to just call `base.Load`?

vitek-karas: `base.Load` will always return `null`, so there's no real difference. And it's the documented contract - return null from `ALC.Load` -> fallback to the `Default` ALC resolution.

src/Shared/MSBuildLoadContext.cs
cdmihai: Maybe key the task load contexts based on the submission entry point's project path? If the new build submission is based on a different entry point, unload the previous one. Else, keep it.

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,96 @@
...
+
+
+        public MSBuildLoadContext(string assemblyPath)
+        {
+            _directory = Directory.GetParent(assemblyPath).FullName;
+        }
+
+        protected override Assembly? Load(AssemblyName assemblyName)
+        {
+            if (_wellKnownAssemblyNames.Contains(assemblyName.Name!))
```
cdmihai: What if the requested msbuild version missmatches the currently running msbuild? Wouldn't it cause all sorts of type missmatches, and also pin the first request?

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        private static readonly string[] _extensions = new[] { "ni.dll", "ni.exe", "dll", "exe" };
+
+
+        public MSBuildLoadContext(string assemblyPath)
+        {
+            _directory = Directory.GetParent(assemblyPath).FullName;
+        }
+
+        protected override Assembly? Load(AssemblyName assemblyName)
```
cdmihai: What about framework assemblies? Should they be "binding redirected" to assemblies shared between all plugins? Otherwise you could get stuff like `String cannot be cast to String`, if communicate types between themselves (e.g. tasks and loggers)

src/Shared/MSBuildLoadContext.cs
cdmihai: In this case though, wouldn't it be better to load that dll in the current context rather than default? Let's say a task has a dependency to an arbitrary path `/a/b/newtonsoft.dll`. This will hardcode the version of `newtonsoft.dll` between all tasks that point to newtonsoft via an arbitrary location.

src/Shared/MSBuildLoadContext.cs
cdmihai: This is basically the reverse of your comment https://github.com/microsoft/msbuild/pull/4916#discussion_r355402062: If the arbitrary dependency is under the msbuild bin directory, you do want sharing. If the arbitrary dependency is not in the msbuild bin dir, then I think you do not want sharing.

src/Shared/MSBuildLoadContext.cs
cdmihai: This is basically the reverse of your comment #4916 (comment): If the arbitrary dependency is under the msbuild bin directory, you do want sharing. If the arbitrary dependency is not in the msbuild bin dir, then I think you do not want sharing.

src/Shared/TypeLoader.cs
```diff
@@ -152,35 +152,18 @@ private static Assembly LoadAssembly(AssemblyLoadInfo assemblyLoadInfo)
             {
                 if (assemblyLoadInfo.AssemblyName != null)
                 {
-#if FEATURE_ASSEMBLY_LOADFROM
+#if !FEATURE_ASSEMBLYLOADCONTEXT
```
cdmihai: Shouldn't the else branch use the ALC to avoid loading in the default ALC?

cdmihai:
Another good set of test assertions would be to update the tests in `TypeLoader_Tests` and `TypeLoader_Dependencies_Tests` with assertions checking that no new types were loaded in the default context.

jzabroski:
@rainersigwald 
A great test case is to load System.ComponentModel.Annotations 4.7 nuget package via net48 and netstandard2.0 tasks, as net48 maps to assemblyVersion 4.2.1 (UGH!!!) and netstandard2.0 maps to assemblyVersion 4.2.0.

rainersigwald:
@jzabroski can you please be more specific? I don't understand the situation you're describing based on your description.

src/Shared/MSBuildLoadContext.cs
vitek-karas: Currently SDK has no way to build a self-contained plugin (classlib project), so one would have to go outside of the supported SDK scenarios to create something like that. But in general plugins must not bring framework assemblies with them, otherwise all kinds of things can break. MSBuild could blacklist/whitelist assemblies just like it does so for MSBuild framework assemblies to make this a bit more robust. In theory MSBuild could enumerate all assemblies in the Default load context (not an easy thing to do, but possible) and "pin" all of them.  

Another solution to this could be to not do anything in `Load` method and instead register event handler to the `Resolving` event on the ALC. The code flow then is that all bind requests will fall back to Default ALC first - and if Default can resolve it, it will do so. And only if Default can't resolve it it would get into the custom ALC's event handler. This would guarantee that all framework assemblies are loaded through Default only. But it would also mean that any other assemblies MSBuild itself uses would get the same treatment. So for example if MSBuild itself would use Newtonsoft.Json, all plugins would get that one version only. Doing this would probably require careful review of all MSBuild dependencies (And potentially work to move some of those into secondary ALCs - out of the Default ALC).

Note that there's "framework" assembly and then there's a different kind of "framework" assembly. Some framework assemblies (this happens specifically in ASP.NET, but in theory in any framework) can be overridden by the application. For example if a customer wants a patch for a specific issue and so on. Currently we allow apps to carry some framework assemblies with them and we pick the higher version from the app/framework pair to load. So blindly whitelisting all framework assemblies would disable this behavior for plugins. Maybe not a big deal... but worth knowing about.

And to answer the question - currently there's no standard way to ask if an assembly is a framework assembly. We could come up with pretty good "guesses", but without doing specific feature work in the runtime/BCL, it would never be 100%. (also as noted above, it's not always 100% clear what is a framework assembly in some cases).

src/Shared/MSBuildLoadContext.cs
vitek-karas: OK - this means that tasks with native (or platform specific dependencies) will still not be easy to write. For example SourceLink is such a task - where it needs to implement a two-layer approach and implement it's own ALC and custom resolution and so on. Such tasks carry some of their dependencies in subfolders which are keyed by the platform - so for example `win/NativeLib.dll` and `linux/NativeLib.so` - similarly for managed dependencies which are different per-platform. Using `AssemblyDependencyResolver` in some way would make this possible as it will be able to correctly pick the right one. The current solution will not help with these tasks.

I'm writing this down just to make it clear, I fully understand the desire to keep this as risk-free as possible.

There might be some way to combine both `AssemblyDependencyResolver` and plain directory search together to achieve a middle ground - but it would require .NET Core 3.0, so we're not there yet anyway.

src/Shared/MSBuildLoadContext.cs
vitek-karas: This looks good.

src/Shared/MSBuildLoadContext.cs
vitek-karas: With the code the way it is now, plugins will get their own version if they have their own copy (as we search the plugins folder first and load into custom ALC if it's there). Only if the plugin doesn't have it, we fallback to Default.

I do agree that a slightly cleaner way to implement this would be to return `null` here and implement an event handler for `AssemblyLoadContext.Default.Resolving` which would resolve assemblies by looking into the msbuild directory (and load them to Default). Such a solution would make it consistent regardless of where the load comes from (task/plugin or msbuild itself). But if we're concern only with plugins/tasks then the end result is the same.

src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.Build.UnitTests
+{
+    public class ValidateAssemblyLoadContext : Task
+    {
+        public override bool Execute()
+        {
+            var thisLoadContext = AssemblyLoadContext.GetLoadContext(typeof(ValidateAssemblyLoadContext).Assembly);
+
+            // Check by name because this always reports false, presumably due to ALC shenanigans:
+            // if (thisLoadContext is MSBuildLoadContext context)
```
jaredpar: Did you file a follow up issue for that commented out section? That definitely seems like a bug or at least an item worthy of documentation.

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -57,113 +28,23 @@ public Assembly LoadFromPath(string fullPath)
...
-            var searchPaths = new[] { Assembly.GetExecutingAssembly().Location };
-            return TryResolveAssemblyFromPaths(context, assemblyName, searchPaths);
-        }
-
-        private Assembly TryResolveAssembly(AssemblyLoadContext context, AssemblyName assemblyName)
-        {
-            lock (_guard)
-            {
-                Assembly assembly = TryGetWellKnownAssembly(context, assemblyName);
+                assembly = contextForAssemblyPath.LoadFromAssemblyPath(fullPath);
```
jaredpar: Loading inside a lock seems a bit dicey. Are there re-entrancy issues to be worried about here?

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -57,113 +28,23 @@ public Assembly LoadFromPath(string fullPath)
...
 
                 if (assembly != null)
```
jaredpar: Shouldn't `null` loads be recorded as well? If not failed loads will just be attempted over and over again correct?

src/Shared/MSBuildLoadContext.cs
cdmihai: One "temporary" way, until the BCL gains some self descriptive capabilities, would be to look at msbuild's project.assets.json and define a whitelist. At the very least, look at what types are ferried across public msbuild interfaces / types (ITask, ILogger, etc)

src/Shared/MSBuildLoadContext.cs
cdmihai: Last case: what if an assembly is neither under the plugin's directory, nor under the msbuild bin directory? Currently it would be loaded in the default context, which could pollute other future tasks that load the same dll from a different, arbitrary path.

src/Shared/MSBuildLoadContext.cs
vitek-karas: Couple of notes (my own opinions):
* Defining the list of types which can be shared between the MSbuild core and tasks is definitely worthwhile. One output of that can be the whitelist of assemblies (where those types are defined) for this purpose. It is a good idea no matter what as that will clearly define expectations on plugins.
* We currently don't have any plans to teach BCL to determine "framework" assemblies. Note that marking something as "framework" is not as clear cut as it seems (especially for apps which override some framework assemblies and for self-contained apps). Also just having that functionality would not be enough for MSBuild - you would still need the whitelist of MSBuild's own "framework" on top of it.
- SDK way of building plugins today effectively enforces "framework dependent" plugins and thus framework assemblies will not be part of the plugin in any way so there should be no collisions anyway. This is not 100% guaranteed, but it's close enough to reduce the importance of trying to solve this in some other way in my opinion.

src/Shared/CoreCLRAssemblyLoader.cs
vitek-karas: Well - runtime caches the results of bind operations (including failures), so overall this should not be called too often really. That said if the cache is already in place, it would definitely help to use if for all assemblies, but I would not expect the perf gain to be substantial.

src/Shared/MSBuildLoadContext.cs
rainersigwald: Filed microsoft/msbuild#5037 to make sure we come back to reevaluate this.

src/Shared/MSBuildLoadContext.cs
jzabroski: > * SDK way of building plugins today effectively enforces "framework dependent" plugins and thus framework assemblies will not be part of the plugin in any way so there should be no collisions anyway. This is not 100% guaranteed, but it's close enough to reduce the importance of trying to solve this in some other way in my opinion.

@vitek-karas  Can you elaborate on how that statement is true and what the scope of that statement is?  Are you saying if new AssemblyLoadContext's don't respect `<RollForward>LatestMajor</RollForward>` and I can't hack it by editing `runtimeconfig.json`?  I'm not internal to Microsoft so I try to wrap my head around this as a customer.

src/Shared/CoreCLRAssemblyLoader.cs
rainersigwald: Looking around, it looks like there are static fields that keep a `CoreCLRAssemblyLoader` around for the lifetime of the process, which for a worker node could be long. If you:

1. build
1. it fails to load a task because you didn't restore so the file doesn't exist (or something)
1. restore
1. build

We need to try to load again in step 4 even if it failed in step 2.

(Did I make the change and get 80% though typing up a commit message justifying it before realizing this? Yes I did.)

src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
rainersigwald: Yeah, I don't understand what's going on here.

I patched it a bit to get more details (again, I think I did this before):

```C#
Log.LogMessage(MessageImportance.High, $"thisLoadContext is MSBuildLoadContext context : {thisLoadContext is MSBuildLoadContext context}");
Log.LogMessage(MessageImportance.High, $"thisLoadContext            loaded in AssemblyLoadContext {AssemblyLoadContext.GetLoadContext(thisLoadContext.GetType().Assembly)}");
Log.LogMessage(MessageImportance.High, $"typeof(MSBuildLoadContext) loaded in AssemblyLoadContext {AssemblyLoadContext.GetLoadContext(typeof(MSBuildLoadContext).Assembly)}");
```

```
  thisLoadContext is MSBuildLoadContext context : False
  thisLoadContext            loaded in AssemblyLoadContext "Default" System.Runtime.Loader.DefaultAssemblyLoadContext #2
  typeof(MSBuildLoadContext) loaded in AssemblyLoadContext "Default" System.Runtime.Loader.DefaultAssemblyLoadContext #2
```

My theory was that they were in different ALCs but that doesn't seem to be true.

@vitek-karas do you know what might cause this?

src/Shared/MSBuildLoadContext.cs
vitek-karas: @jzabroski What the statement I wrote meant is that:
* Plugins should be built as classlib projects (producting a dll not an exe)
* If that is true SDK currently has no way to produce a stand-alone classlib (as in classlib which contains frameworks copied into the output). All classlibs will be framework dependent (the build assumes that all the necessary dependencies will be available at runtime).
* If the classlib is consumed via build (ProjectReference/PackageReference) the build system is the one to figure out all the dependencies and make sure they're available
* If the classlib is loaded as a plugin then this job is left to the runtime/ALC. That's what the AssemblyDependencyResolver class is for - it implements the necessary logic to parse .deps.json (dependencies description produced by the build of the classlib) and resolve them at runtime. The implementation in this PR doesn't use AssemblyDependencyResolver (as it's only available in 3.0 and above), but it implements something which should work for the most part (searches for the assemblies in several locations)

Given the above it's very unlikely that framework assemblies (let's sat System.Xml.dll as an example) will be part of the plugins directory. Technically there is a way to force SDK to put it into the output, but it's an explicit action by the developer of the plugin and works on one assembly at a time. So almost no plugins will do this (why would they).

I must admit I don't understand how RollForward and runtimeconfig relates to this topic though. When loading plugins purely from managed code (As in this case), no runtimeconfig or rollforward is considered. It is assumed that the host (msbuild in this case) can resolve dependencies and if it can't it will fail.

Plugins may have runtimeconfig but that's for cases where they are loaded via native hosting APIs (and so they may start a runtime, in which case the runtimeconfig is used to determine which runtime/frameworks to load).

src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
vitek-karas: The different ALCs would be my theory as well - but as you mention that doesn't seem to be the case. You can try to GetType on both ends of the comparison and look it up in the debugger (make object ID) to double check. If you can prove that the Type objects are the same, but "is" doesn't work - that would be a problem.

src/Shared/MSBuildLoadContext.cs
cdmihai: > Note that marking something as "framework" is not as clear cut as it seems

True. Though one can conceive a list of "core language types", like String, IEnumerable, IDisposable, etc. Especially the types that the compiler knows about. But this analysis is beyond the scope of this PR :)

src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
nguerrera: I feel uneasy that this isn't resolved. Please replace "shenanigans" with an explanation as what is happening. It feels like we don't even know if this is a bug or not yet.

src/Shared/CoreCLRAssemblyLoader.cs
nguerrera: Is there a conflict between above needing to work and the comment just above that says runtime caches failures?

src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
rainersigwald: Thanks to Nick offline I think I now understand what's going on. Because `MSBuildLoadContext` is compiled as source into `MSBuild.exe` and `Microsoft.Build.dll`, there are two copies of it available at runtime. MSBuild's copy is exposed to the test assembly via IVT, but Microsoft.Build's is the one that is relevant for loading the task. So in fact they don't match--even though they have the same namespace names they have different assembly-qualified names.

I'll amend the comment to describe this--I don't think it's worth juggling IVT right now. Filing an issue to come back and look at that.

src/Shared/MSBuildLoadContext.cs
```diff
@@ -0,0 +1,97 @@
...
+            new[]
+            {
+                "MSBuild",
+                "Microsoft.Build",
+                "Microsoft.Build.Framework",
+                "Microsoft.Build.Tasks.Core",
+                "Microsoft.Build.Utilities.Core",
+            }.ToImmutableHashSet();
+
+        private static readonly string[] _extensions = new[] { "ni.dll", "ni.exe", "dll", "exe" };
```
nguerrera: Separate, but if I'm not mistaken ni.dll and ni.exe are holdovers from early days of CoreCLR and not something we need to support. Potentially could speed things up to stop probing for them. Or at least move them to the end of the list? Probably best as a separate issue, though, either way.

src/Shared/MSBuildLoadContext.cs
jzabroski: Thanks, Vitek.  I think I am slowly closing my gap of knowledge on all the knobs you can turn in .NET Core to configure how assemblies get loaded.  I wish there was a "CLR VIA C#" type chapter that explained all this, but I don't think there is one source of truth for all this information.

src/Shared/MSBuildLoadContext.cs
rainersigwald: I was thinking along the same lines but didn't bother to follow up. Filed microsoft/msbuild#5042

src/MSBuild.UnitTests/ValidateAssemblyLoadContext.cs
jzabroski: What does IVT stand for?  Intermediate Value Theorem?? Interrupt Vector Table?? InternalsVisibleTo?

src/Shared/MSBuildLoadContext.cs
vitek-karas: @jzabroski Thanks for the feedback - it's hard for me to tell which parts of the system are tricky to understand as I've been living in it for the last year 😄 . We have some docs in this area.
The overview is here: https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/overview
And more details are in the Loading details section (in the TOC on the left). Of note are probably:
* https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing
* https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/loading-managed

Some more details on how the managed assemblies in .deps.json are found is here: https://github.com/dotnet/runtime/blob/master/docs/design/features/host-probing.md (although it's probably a tiny bit out of date).

And then framework resolution is described here: https://github.com/dotnet/runtime/blob/master/docs/design/features/framework-version-resolution.md - but be warned, this is a really detailed doc - only if you're feeling hardcore about it 😉

rseanhall:
@rainersigwald Do you mind editing the description of this PR? The final version doesn't use `AssemblyDependencyResolver`, follow the referenced spec, or change the target of msbuild to .NET Core 3.0.

PR #4918, opened 2019-11-13
Add arcade telemetry for build failures in Build-Repo

Forgind:
Helps us correctly classify build errors as such when running them through Arcade.

### DIFF
--- a/eng/build.ps1
+++ b/eng/build.ps1
@@ -174,6 +174,8 @@ try {
   }
 }
 catch {
+  Write-Host $_
+  Write-Host $_.Exception
   Write-Host $_.ScriptStackTrace
   Write-PipelineTelemetryError -Category "Build" -Message $_
   ExitWithExitCode 1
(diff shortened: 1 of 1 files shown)


eng/build.ps1
```diff
@@ -174,9 +174,8 @@ try {
   }
 }
 catch {
-  Write-Host $_
-  Write-Host $_.Exception
   Write-Host $_.ScriptStackTrace
+  Write-PipelineTelemetryError -Category "Build" -Message $_
```
rainersigwald: Is this file is the only relevant change? It looks like you pulled in some other commits from another PR.

Forgind: You're right—I pulled in the parallelism branch by accident. I'll sort that out.

rainersigwald:
The root cause of the Arcade misclassification is that we're not using the Arcade logger in our Windows legs. To fix it, we should pick up #4839. I think we should abandon this PR in favor of that one.

rainersigwald:
Closing in favor of #4839.

PR #4921, opened 2019-11-14
Diagnosability refactoring

rainersigwald:
This is a collection of small things I discovered while implementing/testing/debugging #4916. Pulling into their own separate review to not distract from the important feature there.

### DIFF
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4275,6 +4275,8 @@ namespace Microsoft.Build.UnitTests.Evaluation
         }
 
 #if FEATURE_HTTP_LISTENER
+        private Exception _httpListenerThreadException = null;
+
         /// <summary>
         /// Verify that DTD processing is disabled when loading a project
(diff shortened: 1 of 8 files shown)


src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4326,7 +4328,8 @@ public void VerifyDTDProcessingIsDisabled2()
                 Thread.Sleep(500);
 
                 // Expect server to be alive and hung up unless a request originating from DTD processing was sent
-                Assert.True(t.IsAlive);
+                _httpListenerThreadException.ShouldBeNull();
```
BenVillalobos: is this check necessary? `_httpListenerThreadException` is not being set anywhere in this test

rainersigwald: It may be set indirectly: this test fires off a thread running `HttpServerThread`, which may set the exception if it fails in some bad way (this is how I found the problem: for some reason it couldn't open the port on my machine).

BenVillalobos: Ah I missed that.

PR #4922, opened 2019-11-15
Don't stomp on *DependsOn properties

AArnott:
This allows the property to be set within a project file without its value being replaced by the imported .targets at the bottom. Particularly in SDK style projects where properties cannot be set after the imported .targets file, this unblocks a key scenario.

### DIFF
--- a/src/Tasks/Microsoft.CSharp.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.CSharp.CurrentVersion.targets
@@ -164,7 +164,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
 
     <PropertyGroup>
-        <CoreCompileDependsOn>_ComputeNonExistentFileProperty;ResolveCodeAnalysisRuleSet</CoreCompileDependsOn>
+        <CoreCompileDependsOn>$(CoreCompileDependsOn);_ComputeNonExistentFileProperty;ResolveCodeAnalysisRuleSet</CoreCompileDependsOn>
         <ExportWinMDFile Condition="'$(ExportWinMDFile)' == '' and '$(OutputType)' == 'WinMDObj'">true</ExportWinMDFile>
     </PropertyGroup>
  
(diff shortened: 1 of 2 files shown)


rainersigwald:
A few notes here:

* This is a breaking change for users that have defined something that has been up 'til now ignored/overwritten. Hopefully that's rare?
* It's not true that you can't overwrite these when using SDKs; you can [desugar the SDK reference](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2019#reference-a-project-sdk), or you can run an `InitialTarget` to change dependencies.

I'm willing to try this but we may need to back it out if we get user pain from it.

AArnott:
> It's not true that you can't overwrite these when using SDKs; you can desugar the SDK reference

That's not an option that nuget packages that contribute build authoring can make. They get imported wherever they get imported. And when I was testing, my nuget package got imported earlier than this stomping property definition.

> or you can run an InitialTarget to change dependencies.

Yes, this is [what I ended up doing](https://github.com/AArnott/Nerdbank.GitVersioning/pull/405/files).

> I'm willing to try this but we may need to back it out if we get user pain from it.

Totally reasonable. And thank you.

Nirmal4G:
I stumbled upon this recently when I tried `CoreCompileDependsOn` for VisualBasic and F# projects. The change in this PR was not applied there.

@AArnott Does it affect VB/F# with NBGV?

AArnott:
@Nirmal4G, no, NB.GV is not adversely impacted by this since we took an execution time workaround to append to the property for all languages.

Nirmal4G:
Does `InitialTargets` workaround have any issues or gotchas I should be aware of?

AArnott:
You shouldn't set `DefaultTargets`. That only affects the target to build when the user doesn't specify it anyway.

But your own msbuild file can set `InitialTargets` without concern. All the InitialTargets set in every imported file gets collected and executed. But these should be very lightweight targets since they run for every single build, even design-time builds.

rainersigwald:
> But these should be very lightweight targets since they run for every single build, even design-time builds.

And the build that happens in command line `Restore`. For these reasons, we mostly migrated away from them in the common targets. But changing a property is lightweight, and these properties will only matter when getting to the target that `DependsOn` them, so I don't think those factors apply to this specific case.

PR #4931, opened 2019-11-18
Filtering for duplicate content files with referencing projects

sfoslund:
Addressing the issue of dependency content files overwriting the referencing project's content files of the same name (see dotnet/sdk#3871). 
Test coverage in the SDK [here](https://github.com/sfoslund/sdk/blob/CopyAppSettingsIssue/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToReferenceAProject.cs#L138).

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -4472,9 +4472,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         SkipNonexistentTargets="true"
         RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
 
-      <Output TaskParameter="TargetOutputs" ItemName="_AllChildProjectItemsWithTargetPath"/>
+      <Output TaskParameter="TargetOutputs" ItemName="_UnfilteredAllChildProjectItemsWithTargetPath"/>
 
     </MSBuild>
+    
+    <!-- Filter out dependency items that could overwrite referencer content -->
+    <CreateItem Include="@(_UnfilteredAllChildProjectItemsWithTargetPath)"  AdditionalMetadata="ParentContent=%(ContentWithTargetPath.Identity)">
+        <Output TaskParameter="Include" ItemName="_FilteredChildProjectItemsWithTargetPath"/>
+    </CreateItem>
+    <ItemGroup>
+      <_ParentChildContentDuplicates Include="@(_FilteredChildProjectItemsWithTargetPath)" Condition="'%(_FilteredChildProjectItemsWithTargetPath.TargetPath)'=='%(ParentContent)'" />
+      <_FilteredChildProjectItemsWithTargetPath Remove="@(_ParentChildContentDuplicates)" />
+      <_FilteredChildProjectItemsWithTargetPath>
+        <ParentContent></ParentContent> <!-- Remove unnecessary metadata -->
+      </_FilteredChildProjectItemsWithTargetPath>
+      <_AllChildProjectItemsWithTargetPath Include="@(_FilteredChildProjectItemsWithTargetPath->Distinct())" />
+    </ItemGroup>
 
     <!-- Target outputs must be full paths because they will be consumed by a different project. -->
     <ItemGroup>
@@ -4484,7 +4497,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <!-- Remove items which we will never again use - they just sit around taking up memory otherwise -->
     <ItemGroup>
-      <_AllChildProjectItemsWithTargetPath       Remove="@(_AllChildProjectItemsWithTargetPath)"/>
+      <_AllChildProjectItemsWithTargetPath            Remove="@(_AllChildProjectItemsWithTargetPath)"/>
+      <_UnfilteredAllChildProjectItemsWithTargetPath  Remove="@(_UnfilteredAllChildProjectItemsWithTargetPath)"/>
+      <_FilteredChildProjectItemsWithTargetPath       Remove="@(_FilteredChildProjectItemsWithTargetPath)"/>
+      <_ParentChildContentDuplicates                  Remove="@(_ParentChildContentDuplicates)"/>
     </ItemGroup>
 
     <!-- Copy paste _GetCopyToOutputDirectoryItemsFromThisProject but keep the items that came from other projects via ProjectReference's OutputItemType metadata -->
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -4569,7 +4569,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="TargetOutputs" ItemName="_ThisProjectItemsToCopyToOutputDirectory" />
     </CallTarget>
 
-    <ItemGroup>
+    <ItemGroup Condition="'$(IgnoreConflictingTransitiveContent)' == 'false' or ('$(IgnoreConflictingTransitiveContent)' == '' and '$(MSBuildRuntimeType)' == 'Core')">
       <_TransitiveItemsToCopyToOutputDirectory Remove="@(_ThisProjectItemsToCopyToOutputDirectory)" MatchOnMetadata="TargetPath" MatchOnMetadataOptions="PathLike" />
     </ItemGroup>
 
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -4472,22 +4472,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         SkipNonexistentTargets="true"
         RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
 
-      <Output TaskParameter="TargetOutputs" ItemName="_UnfilteredAllChildProjectItemsWithTargetPath"/>
+      <Output TaskParameter="TargetOutputs" ItemName="_AllChildProjectItemsWithTargetPath"/>
 
     </MSBuild>
-    
-    <!-- Filter out dependency items that could overwrite referencer content -->
-    <CreateItem Include="@(_UnfilteredAllChildProjectItemsWithTargetPath)"  AdditionalMetadata="ParentContent=%(ContentWithTargetPath.Identity)">
-        <Output TaskParameter="Include" ItemName="_FilteredChildProjectItemsWithTargetPath"/>
-    </CreateItem>
-    <ItemGroup>
-      <_ParentChildContentDuplicates Include="@(_FilteredChildProjectItemsWithTargetPath)" Condition="'%(_FilteredChildProjectItemsWithTargetPath.TargetPath)'=='%(ParentContent)'" />
-      <_FilteredChildProjectItemsWithTargetPath Remove="@(_ParentChildContentDuplicates)" />
-      <_FilteredChildProjectItemsWithTargetPath>
-        <ParentContent></ParentContent> <!-- Remove unnecessary metadata -->
-      </_FilteredChildProjectItemsWithTargetPath>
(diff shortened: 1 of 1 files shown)


rainersigwald:
I think this will conflict with #4865. @cdmihai would you mind taking a look at this one with that in mind?

cdmihai:
> I think this will conflict with #4865. @cdmihai Mihai Codoban FTE would you mind taking a look at this one with that in mind?

Yes, it will. I propose reimplementing this over  #4865, because  #4865 nicely separates the content items coming coming the current project from the content items coming from references.

sfoslund:
@cdmihai Sounds good to me, I'll update this once #4865 is in.

rainersigwald:
@sfoslund We merged the other PR; can you reimplement this in the new approach? Let me and @cdmihai know if you have any questions.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4472,9 +4472,22 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         SkipNonexistentTargets="true"
         RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)">
 
-      <Output TaskParameter="TargetOutputs" ItemName="_AllChildProjectItemsWithTargetPath"/>
+      <Output TaskParameter="TargetOutputs" ItemName="_UnfilteredAllChildProjectItemsWithTargetPath"/>
```
cdmihai: Wouldn't it be easier to implement this filtering in the target `GetCopyToOutputDirectoryItems` [after the two CallTarget calls](https://github.com/microsoft/msbuild/pull/4931/files#diff-02872751bc64e0ff785c9b6c1cd919b5R4587), via something like:

```xml
<ItemGroup>
   <_TransitiveItemsToCopyToOutputDirectory Remove="@(_ThisProjectItemsToCopyToOutputDirectory)" />
</ItemGroup>
```

The above removes all the transitive content items that also appear in the current project's content items, thus preventing any transitive content items from overriding the current project's content items.

Another reason for doing it this way is that all the rest of the includes in this target (`_GetCopyToOutputDirectoryItemsFromTransitiveProjectReferences`) can bring transitive content items, so moving it in `GetCopyToOutputDirectoryItems` nicely handles all these items.

cdmihai:
This is a breaking change for legacy style projects.

Should this behaviour be under a flag that only the `Microsoft.Net.Sdk` turns on? Or is the actual intent to force this upon everyone by default?

sfoslund:
@cdmihai Yes, a flag is a good idea. Do you have a certain flag in mind?

cdmihai:
> Do you have a certain flag in mind?

How about `IgnoreConflictingTransitiveContent`

src/Tasks/Microsoft.Common.CurrentVersion.targets
sfoslund: Moving it makes sense, but I think we would still need the more complicated logic. My understanding is that that above line would filter out duplicate items if they had the same path but I'm looking to filter out items with the same target path, see my test project:

Screenshot (7)

Is there an easier way to find the overlap of two item groups based on matching metadata than my original implementation?

cdmihai: You're right, filtering needs to happen on matching `TargetPath` metadata. Fortunately, doing it in the target `GetCopyToOutputDirectoryItems` should guarantee that all items have `TargetPath` defined.

Unfortunately it gets quite hairy to write the logic. This is the shortest I came up with to remove all `I1` items from `I2` that match on metadata `M`:
```xml
<Project>
    <ItemGroup>
        <I1 Include="a1" M="1"/>
        <I1 Include="b1" M="2"/>
        <I1 Include="c1" M="3"/>
        <I1 Include="d1" M="4"/>

        <I2 Include="a2" M="x"/>
        <I2 Include="b2" M="2"/>
        <I2 Include="c2" M="3"/>
        <I2 Include="d2" M="y"/>
    </ItemGroup>

    <Target Name="Build">
        <ItemGroup>
            <I1_a Include="%(I1.M)" PreviousIdentity="%(I1.Identity)" />
            <I2_a Include="%(I2.M)" PreviousIdentity="%(I2.Identity)" />

            <I2_b Include="@(I2_a)" KeepMetadata="PreviousIdentity"/>

            <I2_a Remove="@(I1_a)" />

            <I2_c Include="@(I2_b)" KeepMetadata="PreviousIdentity"/>
            <I2_c Remove="@(I2_a)" />

            <I2_d Include="%(I2_c.PreviousIdentity)" />

            <I2 Remove="@(I2_d)" />
        </ItemGroup>
        <Message Importance="High" Text="I2: %(I2.Identity) M=[%(I2.M)]" />
    </Target>
</Project>
```

It would be much more useful if we implemented a new msbuild item operation that does all this in one line, something like `<I2 Remove="I1" On="M"/>`. @sfoslund, if I prototype the new operation, would you be willing to finish it (fill in edge cases, make code pretty, tests, shepherd PR, etc)? (cc @livarcocc )

sfoslund: @cdmihai  Sure, that's fine with me

cdmihai: Awesome. Here's a first draft, that implements this for Remove under targets (but not for remove outside of targets): https://github.com/microsoft/msbuild/pull/4997

sfoslund:
Thanks @Forgind! Looking back on previous feedback, I'll also want to add a flag to enable this behavior before merge, as discussed with @cdmihai

Forgind:
So more like `<ItemGroup Condition="'$(IgnoreConflictingTransitiveContent)' == 'false'>`? Should the flag be on by default or off?

sfoslund:
This is a breaking change for legacy style projects so it should be disabled by default but enabled for sdk style projects.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4569,6 +4569,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="TargetOutputs" ItemName="_ThisProjectItemsToCopyToOutputDirectory" />
     </CallTarget>
 
+    <ItemGroup Condition="'$(IgnoreConflictingTransitiveContent)' == 'false' or ('$(IgnoreConflictingTransitiveContent)' == '' and '$(MSBuildRuntimeType)' == 'Core')">
```
rainersigwald, line 4572: What's the motivation for the is-Core check here? In general we'd rather push this to the SDK so that we get identical behavior between VS, `msbuild.exe`, and `dotnet build`. 

Can we instead just consume a "bool" (tri-state) here, and opt into the new behavior in the SDK?

sfoslund: That sounds good to me. @Forgind did you have a specific motivation for adding this check or does opting in from the SDK sound okay to you?

Forgind: @cdmihai [noted](https://github.com/microsoft/msbuild/pull/4931#issuecomment-566809124) that this is a breaking change for legacy projects, so I wanted to make it opt-in for VS and opt-out for the SDK. We can do it in the SDK instead. If the SDK explicitly opts in by default, that works, too; it's just an extra PR.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4569,6 +4569,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="TargetOutputs" ItemName="_ThisProjectItemsToCopyToOutputDirectory" />
     </CallTarget>
 
+    <ItemGroup Condition="'$(IgnoreConflictingTransitiveContent)' == 'true'">
```
Forgind, line 4572: Wait, why true? Shouldn't it be false? If it is true that we are ignoring conflicts, we should not be doing anything, i.e., we should not remove anything from _TransitiveItemsToCopyToOutputDirectory.

Forgind: Note that whether it's true or false doesn't affect whether it's opt-in or opt-out, since the default is nothing.

sfoslund: My bad, I misread the name of the property. I've changed it so the condition is `IgnoreConflictingTransitiveContent == false`

Forgind: Also, have you prepped a change in the SDK to set IgnoreConflictingTransitiveContent?

src/Tasks/Microsoft.Common.CurrentVersion.targets
sfoslund: Just updated the PR with this change, I'll make note to make a PR into the SDK once this is in.

src/Tasks/Microsoft.Common.CurrentVersion.targets
sfoslund: I haven't prepped the change, feel free to make it if you want. I don't think you would want to put the property in Directory.Build.props, though, since I believe that's only used for building the SDK itself. Instead, we want it in one of [the .props files that ships with the SDK](https://github.com/dotnet/sdk/tree/master/src/Tasks/Microsoft.NET.Build.Tasks/targets). I would put it in [Microsoft.NET.Sdk.props](https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L105)

Forgind: I put it where you suggested, but I don't fully understand the implications. I thought both would be included in any SDK-style project by default, so putting it in either places would have the same effect?

sfoslund: Putting it in any of [these files](https://github.com/dotnet/sdk/tree/master/src/Tasks/Microsoft.NET.Build.Tasks/targets) would have the same effect, since all of them are included in the SDK by default, so among those it's just convention. However, the original link pointed to the Directory.Build.props file at the root of the repo, which isn't included in the SDK as it's used to customize our build (note that we do some things that are specific to building the SDK in that file like import arcade and set warnings as errors).

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4569,6 +4569,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="TargetOutputs" ItemName="_ThisProjectItemsToCopyToOutputDirectory" />
     </CallTarget>
 
+    <ItemGroup Condition="'$(IgnoreConflictingTransitiveContent)' == 'false'">
```
rainersigwald, line 4572:
```suggestion
    <ItemGroup Condition="'$(CopyConflictingTransitiveContent)' == 'true'">
```

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4569,6 +4569,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="TargetOutputs" ItemName="_ThisProjectItemsToCopyToOutputDirectory" />
     </CallTarget>
 
+    <ItemGroup Condition="'$(CopyConflictingTransitiveContent)' == 'true'">
```
rainersigwald, line 4572: You did exactly what I said, but I was wrong. Sorry.

The normal behavior is that we copy conflicting transitive content. So we want to explictly stop doing that.

```suggestion
    <ItemGroup Condition="'$(CopyConflictingTransitiveContent)' == 'false'">
```

right?

Forgind: Yes; I agree.

"Ignore" and "Copy" sound like opposites, but it was ignore the _conflicts_ or copy the _files_, which actually does the same thing. I'm wondering if we should ask another opinion on how to make this less confusing.

sfoslund: Right, totally missed that, sorry!

PR #4939, opened 2019-11-21
Implement feature that allows tasks to get the global properties

jeffkl:
This is an implementation of https://github.com/microsoft/msbuild/issues/4925 which allows tasks to get the current project's global properties.

I added an `IBuildEngine6` and plumbed the global properties to the appropriate task hosts.

### DIFF
--- a/src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
@@ -185,7 +185,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
                 "TaskName",
                 @"c:\MyTasks\MyTask.dll",
                 null,
-                    null);
+                null);
 
             IDictionary<string, object> parameters = new Dictionary<string, object>();
             TaskHostConfiguration config3 = new TaskHostConfiguration(
@@ -204,7 +204,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
                 "TaskName",
                 @"c:\MyTasks\MyTask.dll",
                 parameters,
-                    null);
+                null);
 
             IDictionary<string, object> parameters2 = new Dictionary<string, object>();
             parameters2.Add("Text", "Hello!");
@@ -228,7 +228,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
                 "TaskName",
                 @"c:\MyTasks\MyTask.dll",
                 parameters2,
-                    null);
+                null);
         }
 
         /// <summary>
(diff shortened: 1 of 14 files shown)


src/Build.UnitTests/BackEnd/TaskHostConfiguration_Tests.cs
```diff
@@ -178,7 +184,8 @@ public void TestValidConstructors()
                 _continueOnErrorDefault,
                 "TaskName",
                 @"c:\MyTasks\MyTask.dll",
-                null);
+                null,
+                    null);
```
rainersigwald: nit: wonky spacing

src/Framework/IBuildEngine6.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// This interface extends <see cref="IBuildEngine5" /> to allow tasks to get the current project's global properties.
+    /// </summary>
+    public interface IBuildEngine6 : IBuildEngine5
+    {
+        /// <summary>
+        /// Gets the global properties for the current project.
+        /// </summary>
+        /// <returns>A <see cref="Dictionary{String, String}" /> containing the global properties of the current project.</returns>
+        Dictionary<string, string> GetGlobalProperties();
```
jeffkl: This should probably return `IDictionary<TKey, TValue>`.  I couldn't return `IReadOnlyDictionary<TKey, TValue>` since its not in .NET 3.5

rainersigwald:
```suggestion
        ReadOnlyDictionary<string, string> GetGlobalProperties();
```

?

jeffkl: ReadOnlyDictionary isn't in .NET 3.5 so the MSBuildTaskHost wouldn't compile :(

jeffkl: So I made sure that I copy the dictionary before returning it

rainersigwald: You can wrap the implementation like the current ones:

https://github.com/microsoft/msbuild/blob/74c08313395ee87d58ef53f8a7918a937082be01/src/MSBuild/OutOfProcTaskHostNode.cs#L380-L381

jeffkl:
It seems a little weird to log a message but it does sort of make sense.  I had coded up something but I couldn't decide on the message as I'm not sure what it should say?

```
TaskA requested the current global properties.

Fetching the current global properties for TaskA

Fetched global properties for TaskA

TaskA requested the following global properties:
  Property1: Value1
```

If there's a message you like, I'll gladly put it in there.

src/Framework/IBuildEngine6.cs
jeffkl: Nevermind, you are right!  I was trying to use `IReadOnlyDictionary` in `TaskHostConfiguration` which was causing me build errors in MSBuildTaskHost.  But I can just use a Dictionary in that class and use IReadOnlyDictionary everywhere else.

AndyGerlicher:
I would say no on logging? If we want that maybe in the style of #4461 which was reverted. Logging the reads seems like a natural extension of that.

jeffkl:
@AndyGerlicher I agree, let's add logging events later as part of the bigger effort to capture things that effect the build, what do you think @rainersigwald?

jeffkl:
@rainersigwald gentle ping...

jeffkl:
Thanks all!

PR #4952, opened 2019-12-03
Discourage building with MSBUILDDEBUGONSTART

Forgind:


build.cmd
```diff
@@ -1,3 +1,10 @@
 @echo off
+set CACHEDMSBUILDDEBUGONSTART=%MSBUILDDEBUGONSTART%
+if %MSBUILDDEBUGONSTART% == 0 goto build
+set /p ans=Did you mean to have MSBUILDDEBUGONSTART equal %MSBUILDDEBUGONSTART%? (y/n)
```
rainersigwald: Should we prompt or just scream loudly about it and use a `REALLY_MSBUILDDEBUGONSTART=1` to opt out of unsetting it?

build.cmd
```diff
@@ -1,3 +1,10 @@
 @echo off
+set CACHEDMSBUILDDEBUGONSTART=%MSBUILDDEBUGONSTART%
+if %MSBUILDDEBUGONSTART% == 0 goto build
+set /p ans=Did you mean to have MSBUILDDEBUGONSTART equal %MSBUILDDEBUGONSTART%? (y/n) 
+if %ans% == y goto build
+set MSBUILDDEBUGONSTART=0
+:build
 powershell -NoLogo -NoProfile -ExecutionPolicy ByPass -Command "& """%~dp0eng\common\build.ps1""" -build -restore %*"
+set MSBUILDDEBUGONSTART=%CACHEDMSBUILDDEBUGONSTART%
```
rainersigwald: I'd put 

```cmd
setlocal
```

at the top of the file (under `@echo off`); then you won't have to worry about resetting state.

build.cmd
```diff
@@ -1,3 +1,10 @@
 @echo off
+set CACHEDMSBUILDDEBUGONSTART=%MSBUILDDEBUGONSTART%
+if %MSBUILDDEBUGONSTART% == 0 goto build
+set /p ans=Did you mean to have MSBUILDDEBUGONSTART equal %MSBUILDDEBUGONSTART%? (y/n) 
+if %ans% == y goto build
+set MSBUILDDEBUGONSTART=0
```
rainersigwald:
```suggestion
set MSBUILDDEBUGONSTART=
```

I prefer unsetting because we're not super consistent about which of our environment variables get "if the environment variable is set to anything at all" versus "set to specific value" treatment.

build.cmd
```diff
@@ -1,3 +1,10 @@
 @echo off
+set CACHEDMSBUILDDEBUGONSTART=%MSBUILDDEBUGONSTART%
+if %MSBUILDDEBUGONSTART% == 0 goto build
```
rainersigwald:
```suggestion
if defined MSBUILDDEBUGONSTART goto build
```

build.cmd
Forgind: The advantage of prompting is that if you actually want MSBUILDDEBUGONSTART=1, you won't be temporarily confused why you can't debug a build anymore and have to learn a new environment variable. The advantage of having an extra environment variable is that if it was unintentional, you don't have to answer a question, which wastes time.

I have a very slight preference for prompting, but I could go either way.

rainersigwald: I feel I'm personally pretty likely to type `./build`, hit enter, and immediately alt-tab away to do something else while it builds, then come back later and be annoyed that it blocked on a prompt. And I don't know if I've ever _intentionally_ debugged our build in this way.

PR #4953, opened 2019-12-04
Node tracing uses correct exe name

BenVillalobos:
Previous functionality had hardcoded strings such as `Successfully launched msbuild.exe node with PID {0}`, even when using `MSBuildTaskHost.exe`. This PR changes these strings to `Successfully launched {1} node with PID {0}` where {1} is whatever the current process is.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
@@ -110,7 +110,7 @@ namespace Microsoft.Build.BackEnd
             // INodePacketFactory
             INodePacketFactory factory = new NodePacketFactory();
 
-            List<Process> nodeProcesses = GetPossibleRunningNodes();
+            List<Process> nodeProcesses = GetPossibleRunningNodes().Item2;
 
             // Find proper MSBuildTaskHost executable name
             string msbuildtaskhostExeName = NodeProviderOutOfProcTaskHost.TaskHostNameForClr2TaskHost;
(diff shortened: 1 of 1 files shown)


src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -169,7 +169,7 @@ protected NodeContext GetNode(string msbuildLocation, string commandLineArgs, in
             {
                 var candidateProcesses = GetPossibleRunningNodes(msbuildLocation);
 
-                CommunicationsUtilities.Trace("Attempting to connect to each existing msbuild.exe process in turn to establish node {0}...", nodeId);
+                CommunicationsUtilities.Trace("Attempting to connect to each existing {1} process in turn to establish node {0}...", nodeId, Path.GetFileName(msbuildLocation));
```
Forgind: I don't think msbuildLocation is guaranteed to have a value at this point, so you might want to add default behavior.

rainersigwald: Maybe change `GetPossibleRunningNodes` to return `(expectedProcessName, candidateProcesses)` and log that? That'll be helpful for the .NET Core host case too.

BenVillalobos: Fixed!

BenVillalobos: Whoops, didn't see the most recent comment.  I'll get on that.

BenVillalobos: Using the name gathered from `GetPossibleRunningNodes` worked out better in this case.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -110,7 +110,7 @@ protected void ShutdownAllNodes(long hostHandshake, long clientHandshake, NodeCo
             // INodePacketFactory
             INodePacketFactory factory = new NodePacketFactory();
 
-            List<Process> nodeProcesses = GetPossibleRunningNodes();
+            List<Process> nodeProcesses = GetPossibleRunningNodes().Item2;
```
rainersigwald: Can I show you a [neat trick](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#tuples)?

https://github.com/microsoft/msbuild/commit/44bdf900b6aac51c965aca60b6a6f932fca23d59

PR #4964, opened 2019-12-09
Initial draft of my-build-is-slow doc

rainersigwald:
"My build is slow" is a common complaint. There's a bunch of information about how to figure out what to do about it, but we don't have much here. Here's a quick first draft doc based on some stuff I've sent to customers recently.

### DIFF
--- a/documentation/diagnosing-slow-builds.md
+++ b/documentation/diagnosing-slow-builds.md
@@ -1,0 +1,44 @@
+# Diagnosing slow builds
+
+No build is ever fast enough. It's often worth investing some effort on understanding where MSBuild spends its time when building your code, and thinking about whether it can be improved.
+
+[Obligatory XKCD](https://www.xkcd.com/303/)
+
+Logs at MSBuild’s “normal” level don't capture most of the really interesting information, though.
+
+Try [capturing a binary log](https://aka.ms/msbuild/binlog) in your build; that preserves detailed information and allows for offline analysis. Changing log verbosity levels can affect build performance, but doesn't usually cause dramatic slowdown for full builds. Cross-check the results of with-logging and without-logging builds to make sure they're not completely out of whack before delving deeply into analyzing the binary log.
+
+## Setting expectations
+
+Do not expect to see 100% processor utilization during the entire build. Builds are often constrained by disk I/O, and the build process is also constrained by the dependencies between projects that are being built. If 100 projects depend on a common library, that library must be built before they can start building.
+
+For this reason, avoid unnecessary dependencies. Dependencies are usually `ProjectReference` items, but can also be expressed in a solution file as "solution build dependencies".
+
+Dependencies between projects also usually keep `-maxcpucount` values from scaling linearly.
+
+## Building in parallel
+
+The biggest performance win is going from single-threaded build to multi-process build by adding `-m` to your command-line build invocation (or equivalent in build automation).
+
+Standard project types should support multiprocessor build without any work on your part, but customizations or very old projects may have race conditions that need to be fixed. It's worth the effort!
+
+## `DetailedSummary` and `PerformanceSummary`
+
+At minimum, [enable DetailedSummary and PerformanceSummary logging](https://docs.microsoft.com/visualstudio/msbuild/msbuild-command-line-reference). That captures information about how MSBuild schedules work to its available nodes and what parts of the build process take up the most time.
+
+TODO: sample outputs and analysis
+
+📝 The performance summary can be misleading with complicated builds, because it counts time spent waiting on another project to build redundantly ([microsoft/msbuild#4189](https://github.com/microsoft/msbuild/issues/4189)).
+
+## If a task or target dominates build time
+
+If most of the build time is being spent in a specific task:
+
+* Can the thing be avoided? Often it cannot (sadly, we can't skip running the compiler or linker), but if it can, that's a huge win.
+* Does the target declare correct inputs and outputs for [incremental build](https://docs.microsoft.com/visualstudio/msbuild/incremental-builds)?
+* Can the inputs be simplified? For instance, can generated code that causes compiler delays be restructured?
+* Should the tool be improved? Is there a bug that can be filed against the task or tool owner?
+
+## If MSBuild project evaluation dominates build time
+
+See [evaluation profiling](evaluation-profiling.md).


documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
+# Diagnosing slow builds
+
+No build is ever fast enough. It's often worth investing some effort on understanding where MSBuild spends its time when building your code, and thinking about whether it can be improved.
+
+[Obligatory XKCD](https://www.xkcd.com/303/)
+
+Logs at MSBuild’s “normal” level don't capture most of the really interesting information, though.
+
+Try [capturing a binary log](https://aka.ms/msbuild/binlog) in your build; that preserves detailed information and allows for offline analysis. Changing log verbosity levels can affect build performance, but doesn't usually cause dramatic slowdown for full builds. Cross-check the results of with-logging and without-logging builds to make sure they're not completely out of whack before delving deeply into analyzing the binary log.
```
Forgind:
```suggestion
Try [capturing a binary log](https://github.com/microsoft/msbuild/blob/master/documentation/wiki/Providing-Binary-Logs.md) in your build. The binary log preserves detailed information and allows for offline analysis. Changing log verbosity levels can affect build performance but doesn't usually slow full builds too dramatically. Cross-check the results of with-logging and without-logging builds to make sure they're reasonable before delving deeply into analyzing the binary log.
```

Forgind: By `Cross-check the results...to make sure they're reasonable`, do you mean making sure the outputs and build times are similar?

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Setting expectations
+
+Do not expect to see 100% processor utilization during the entire build. Builds are often constrained by disk I/O, and the build process is also constrained by the dependencies between projects that are being built. If 100 projects depend on a common library, that library must be built before they can start building.
```
Forgind:
```suggestion
Do not expect to see full processor utilization during the entire build. Builds are often constrained by disk I/O, and the build process is also constrained by the dependencies between projects that are being built. If 100 projects depend on a common library, that library must be built before they can start building.
```

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+For this reason, avoid unnecessary dependencies. Dependencies are usually `ProjectReference` items, but can also be expressed in a solution file as "solution build dependencies".
```
Forgind:
```suggestion
In other words, unnecessary dependencies may slow your build. Dependencies are usually `ProjectReference` items but can also be expressed in a solution file as "solution build dependencies".
```

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Building in parallel
+
+The biggest performance win is going from single-threaded build to multi-process build by adding `-m` to your command-line build invocation (or equivalent in build automation).
+
+Standard project types should support multiprocessor build without any work on your part, but customizations or very old projects may have race conditions that need to be fixed. It's worth the effort!
+
+## `DetailedSummary` and `PerformanceSummary`
+
+At minimum, [enable DetailedSummary and PerformanceSummary logging](https://docs.microsoft.com/visualstudio/msbuild/msbuild-command-line-reference). That captures information about how MSBuild schedules work to its available nodes and what parts of the build process take up the most time.
```
Forgind:
```suggestion
[Enable DetailedSummary and PerformanceSummary logging](https://docs.microsoft.com/visualstudio/msbuild/msbuild-command-line-reference) to capture information about how MSBuild schedules work to its available nodes and what parts of the build process take up the most time.
```

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+TODO: sample outputs and analysis
+
+📝 The performance summary can be misleading with complicated builds, because it counts time spent waiting on another project to build redundantly ([microsoft/msbuild#4189](https://github.com/microsoft/msbuild/issues/4189)).
```
Forgind:
```suggestion
📝 The performance summary can be misleading with complicated builds, because it counts time spent waiting on another project to build redundantly (This issue is being tracked with [microsoft/msbuild#4189](https://github.com/microsoft/msbuild/issues/4189)).
```

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+## If a task or target dominates build time
+
+If most of the build time is being spent in a specific task:
+
+* Can the thing be avoided? Often it cannot (sadly, we can't skip running the compiler or linker), but if it can, that's a huge win.
```
Forgind:
```suggestion
* Can it be avoided? Often it cannot (sadly, we can't skip running the compiler or linker), but if it can, that's a huge win.
```

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+* Does the target declare correct inputs and outputs for [incremental build](https://docs.microsoft.com/visualstudio/msbuild/incremental-builds)?
```
Forgind:
```suggestion
* Does the target declare correct inputs and outputs for [incremental builds](https://docs.microsoft.com/visualstudio/msbuild/incremental-builds)?
```

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+* Can the inputs be simplified? For instance, can generated code that causes compiler delays be restructured?
```
Forgind: To some extent, this sounds like the "unnecessary dependencies" problem. Perhaps combine with that?

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+* Should the tool be improved? Is there a bug that can be filed against the task or tool owner?
+
+## If MSBuild project evaluation dominates build time
+
+See [evaluation profiling](evaluation-profiling.md).
```
Forgind: I don't think that's a real link?

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+## `DetailedSummary` and `PerformanceSummary`
+
+At minimum, [enable DetailedSummary and PerformanceSummary logging](https://docs.microsoft.com/visualstudio/msbuild/msbuild-command-line-reference). That captures information about how MSBuild schedules work to its available nodes and what parts of the build process take up the most time.
+
+TODO: sample outputs and analysis
+
+📝 The performance summary can be misleading with complicated builds, because it counts time spent waiting on another project to build redundantly ([microsoft/msbuild#4189](https://github.com/microsoft/msbuild/issues/4189)).
+
+## If a task or target dominates build time
```
Forgind: How would they figure this out? https://github.com/microsoft/msbuild/pull/4950

rainersigwald: By looking at the performance summary.

documentation/diagnosing-slow-builds.md
Forgind: https://github.com/microsoft/msbuild/blob/master/documentation/evaluation-profiling.md

documentation/diagnosing-slow-builds.md
Forgind: I was suggesting they could also use the events I added.

documentation/diagnosing-slow-builds.md
cdmihai: Is there anything out there that prints the critical path, so people can easily see it and work on reducing it? I regularly print out the p2p graph with [this thing](https://github.com/microsoft/MSBuildStaticGraphTester/tree/master/src/GraphGen) whenever I need to debug static graph, but I guess it could be used to visually look and identity critical paths. It can also be reasonably easily changed to add an option to color the critical path. Or to prune everything else out.

documentation/diagnosing-slow-builds.md
cdmihai: I think it's worth mentioning that sdk style projects inject the entire dependency closure into each project's `ProjectReference`. People that care about perf might want to turn this off.

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+Try [capturing a binary log](https://aka.ms/msbuild/binlog) in your build; that preserves detailed information and allows for offline analysis. Changing log verbosity levels can affect build performance, but doesn't usually cause dramatic slowdown for full builds. Cross-check the results of with-logging and without-logging builds to make sure they're not completely out of whack before delving deeply into analyzing the binary log.
+
+## Setting expectations
+
+Do not expect to see 100% processor utilization during the entire build. Builds are often constrained by disk I/O, and the build process is also constrained by the dependencies between projects that are being built. If 100 projects depend on a common library, that library must be built before they can start building.
+
+For this reason, avoid unnecessary dependencies. Dependencies are usually `ProjectReference` items, but can also be expressed in a solution file as "solution build dependencies".
+
+Dependencies between projects also usually keep `-maxcpucount` values from scaling linearly.
```
cdmihai: `MSBuild /m /graph` will build bottom up, in parallel, allowing users to squeeze some extra parallelism juice.

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Building in parallel
+
+The biggest performance win is going from single-threaded build to multi-process build by adding `-m` to your command-line build invocation (or equivalent in build automation).
+
+Standard project types should support multiprocessor build without any work on your part, but customizations or very old projects may have race conditions that need to be fixed. It's worth the effort!
```
cdmihai: Any good advice on detecting races? Like the binlog's double write detector, or other patterns to look for in the binlog?

danmosemsft: re races, this could be very useful:
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/BinClashLogger.cs

It's not necessarily ready to reuse, but picking it up and using it could be far cheaper than trying to figure out by eyeball why projects are building concurrently.

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
...
+
+Do not expect to see 100% processor utilization during the entire build. Builds are often constrained by disk I/O, and the build process is also constrained by the dependencies between projects that are being built. If 100 projects depend on a common library, that library must be built before they can start building.
+
+For this reason, avoid unnecessary dependencies. Dependencies are usually `ProjectReference` items, but can also be expressed in a solution file as "solution build dependencies".
+
+Dependencies between projects also usually keep `-maxcpucount` values from scaling linearly.
+
+## Building in parallel
+
+The biggest performance win is going from single-threaded build to multi-process build by adding `-m` to your command-line build invocation (or equivalent in build automation).
```
danmosemsft: In some cases `-m` may over subscribe the machine, specifically if there are build steps that add their own parallelism, for example a compiler that compiles its inputs in parallel. one sign of this may be that you see many more active processes than you haev cores. try specifying a smaller number for `-m`

documentation/diagnosing-slow-builds.md
danmosemsft: Avoid file references to anything that's an output of your build. It will like cause a race. Use project references OR use manually specified project dependencies in the solution file (advanced -- add these using VS)

documentation/diagnosing-slow-builds.md
danmosemsft: For slow builds, I think there are two distinct investigations
1. speed of clean builds. I guess that's this doc
1. speed of incremental builds. that is done differently. typically I start by building a 2nd time and verifying that nothing actually happened. if it did happen, check inputs/outputs. then once that's fixed, try touching specific inputs and verifying the right thing happens. for most people, they're not using custom builds steps, so hopefully they never need this kind of investigation. but it is a *different* process

documentation/diagnosing-slow-builds.md
stevenbrix: how do you avoid races? in our build we have dozens of projects in our solution that are dependent on a fairly common project being built first, and I see a bunch of file write errors in the `CopyFilesToOutputDirectory` because other instances of MSBuild have a lock on the file

documentation/diagnosing-slow-builds.md
BenVillalobos: I'm fine with this suggested change except for going away from the aka.ms link (so long as one of us owns that aka.ms link)

documentation/diagnosing-slow-builds.md
```diff
@@ -0,0 +1,44 @@
+# Diagnosing slow builds
+
+No build is ever fast enough. It's often worth investing some effort on understanding where MSBuild spends its time when building your code, and thinking about whether it can be improved.
+
+[Obligatory XKCD](https://www.xkcd.com/303/)
+
+Logs at MSBuild’s “normal” level don't capture most of the really interesting information, though.
```
BenVillalobos:
```suggestion
"Normal" level MSBuild logs don't capture the most interesting information, though.
```

documentation/diagnosing-slow-builds.md
BenVillalobos:
```suggestion
## [Capture a binary log](https://aka.ms/msbuild/binlog)
Start with capturing a `.binlog`. This preserves detailed information and allows for offline analysis. Changing log verbosity levels can affect build performance, but doesn't usually cause dramatic slowdown for full builds. Cross-check the results of with-logging and without-logging builds to make sure they're not completely out of whack before delving deeply into analyzing the binary log.
```
Capturing a binary log is an important step such that I'd suggest giving it it's own header.

Forgind: I was wrong when I suggested moving away from the aka.ms link. Glancing at my history, it seems I typed `aka.ms/msbuild/buildlog` by accident, got nowhere, and assumed it was a dead link.

stevenbrix: can you add documentation about how to do this in Visual Studio? I haven't figured out how to do this and command line build are often different from Visual Studio, so when I drop down to the command line to build I get a different set of errors.

documentation/diagnosing-slow-builds.md
stevenbrix: if you add dependencies with VS, does that pretty much mean you can never build from the command line? or do you always have to build the .sln to get consistent results?

documentation/diagnosing-slow-builds.md
aolszowka: "Is there anything out there that prints the critical path"

I too am interested in this, recently we spun up a 64 Core VB with 128GB of RAM in Azure, we allocated a 64GB RAM Drive to try and rule out any I/O limitations. Historically I have used a utility that I open sourced awhile back (https://github.com/aolszowka/MsBuildProjectReferenceDependencyGraph) however based on my graphs I don't think this is the issue we are encountering.

rainersigwald:
I'd like to get back to this . . . someday.

PR #4969, opened 2019-12-10
Avoid loading NuGet.targets if already loaded

chipitsine:
Signed-off-by: Ilya Shipitsin <chipitsine@gmail.com>

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -6077,7 +6077,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'==''">$(MSBuildToolsPath)\NuGet.targets</NuGetRestoreTargets>
   </PropertyGroup>
 
-  <Import Condition="Exists('$(NuGetRestoreTargets)')" Project="$(NuGetRestoreTargets)" />
+  <Import Condition="'$(IsRestoreTargetsFileLoaded)' != 'true' and Exists('$(NuGetRestoreTargets)')" Project="$(NuGetRestoreTargets)" />
 
   <Import Project="$(CustomAfterMicrosoftCommonTargets)" Condition="'$(CustomAfterMicrosoftCommonTargets)' != '' and Exists('$(CustomAfterMicrosoftCommonTargets)')"/>
 


rainersigwald:
@chipitsine Can you provide some details about why this is needed?

cc @rrelyea, @jeffkl

chipitsine:
in complex environments NuGet.targets might be occasioanely imported many times

```
     xxx\Current\bin\Microsoft.Common.CurrentVersion.targets(6086,3): warning MSB4011: "xxx\Current\bin\NuGet.targets" cannot be imported again. It was already imported at "xxx\yyy.props (63,3)". This is most likely a build authoring error. This subsequent import will be ignored. [xxx\yyy.proj]
```

`IsRestoreTargetsFileLoaded` is property that means NuGet.targets is already imported

rrelyea:
Including @nkolev92 

What is impact of loading multiple times in one project? Is there a repro of how this happens?

rainersigwald:
@rrelyea the impact is that warning (which could be treated as an error if a user has their build configured that way), and confusion around import ordering. This would only address the former.

I would also like to know more details about exactly what projects are causing this warning--what are the "complex environments" like?

chipitsine:
I'm afraid I cannot provide particular details due to some NDA.

but google shows similar [issues](https://www.google.com/search?ei=trHzXe7xObOCk74Pp-mSkAc&q=Microsoft.Common.CurrentVersion.targets+warning+MSB4011+NuGet.targets+cannot+be+imported+again.+It+was+already+imported+at+This+is+most+likely+a+build+authoring+error.+This+subsequent+import+will+be+ignored&oq=Microsoft.Common.CurrentVersion.targets+warning+MSB4011+NuGet.targets+cannot+be+imported+again.+It+was+already+imported+at+This+is+most+likely+a+build+authoring+error.+This+subsequent+import+will+be+ignored&gs_l=psy-ab.3...3856.13099..13534...0.0..0.0.0.......3....1..gws-wiz.sGMtdrLfjT4&ved=0ahUKEwjupNfT-7LmAhUzwcQBHae0BHIQ4dUDCAo&uact=5)

chipitsine:
yep, the impact is to mute warnings

rainersigwald:
@nkolev92, do you have a preference on this? The MSBuild team is willing to take it.

The only downside is that if you have a situation where you import `NuGet.targets` early, then in the normal location, it might be harder to set properties to configure NuGet behavior. That's true today if you're in that situation, but you get the warning.

One possible workaround: `-warnAsMessage:MSB4011` on the command line should quiet this.

nkolev92:
I do not recall getting a lot of feedback (read: any) around duplicate imports of the NuGet targets file. 
So I am inclined to say, I'm ok with this change. 

Is there even a default template that could lead you to this situation?

PR #4980, opened 2019-12-12
Fix Incremental Builds (_GenerateCompileDependencyCache)

BenVillalobos:
Pairs with https://github.com/dotnet/wpf/issues/2287 to fix https://github.com/microsoft/msbuild/issues/4963

Fixed the _**incredibly complicated**_ boolean logic issue that prevented _GenerateCompileDependencyCache from running at all for builds (which is when it should be running)

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -3424,7 +3424,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     contribute to incremental build inconsistencies.
     ============================================================
     -->
-  <Target Name="_GenerateCompileDependencyCache" Condition="'$(DesignTimeBuild)' != 'true' and '$(BuildingProject)' != 'true'" DependsOnTargets="ResolveAssemblyReferences">
+  <Target Name="_GenerateCompileDependencyCache" Condition="'$(DesignTimeBuild)' != 'true' and '$(BuildingProject)' == 'true'" DependsOnTargets="ResolveAssemblyReferences">
     <ItemGroup>
       <CustomAdditionalCompileInputs Include="$(IntermediateOutputPath)$(MSBuildProjectFile).CoreCompileInputs.cache" />
       <CoreCompileCache Include="@(Compile)" />


BenVillalobos:
## I've verified that `_GenerateCompileDenendencyCache` runs:

### With fix
**Design Time Builds**
CPS: No ✅
Legacy Project System: No ✅

**Standard Builds**
CPS: yes ✅
Legacy Project System: yes ✅

### Without fix

**Design Time Build**
CPS: No✅
Legacy Project System: No ✅

**Standard Builds**
CPS: **No** ❌
Legacy Project System: **No** ❌

BenVillalobos:
In the interest of servicing `vs16.4`, I'll base this change off of 16.4 and cherry pick it from master.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -3400,7 +3400,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     contribute to incremental build inconsistencies.
     ============================================================
     -->
-  <Target Name="_GenerateCompileDependencyCache" Condition="'$(DesignTimeBuild)' != 'true' and '$(BuildingProject)' != 'true'" DependsOnTargets="ResolveAssemblyReferences">
+  <Target Name="_GenerateCompileDependencyCache" Condition="'$(DesignTimeBuild)' != 'true' and '$(BuildingProject)' == 'true'" DependsOnTargets="ResolveAssemblyReferences">
```
rainersigwald: Pedantically checking ourselves here:

The [documented check for "is a design time build"](https://github.com/dotnet/project-system/blob/master/docs/design-time-builds.md#determining-whether-a-target-is-running-in-a-design-time-build) is 

```
'$(DesignTimeBuild)' == 'true' OR '$(BuildingProject)' != 'true'
```

We want its inverse here, so

```
!( '$(DesignTimeBuild)' == 'true' OR '$(BuildingProject)' != 'true' )
```

We can use [DeMorgan's theorems](https://www.allaboutcircuits.com/textbook/digital/chpt-7/demorgans-theorems/):

```
!('$(DesignTimeBuild)' == 'true') AND !('$(BuildingProject)' != 'true')

'$(DesignTimeBuild)' != 'true') AND '$(BuildingProject)' == 'true'
```

Which is what this PR does. So 👍.

dasMulli:
It took me an hour to figure out this ""broke"" my incremental builds in docker files :)
They all built the .sln file and then ran a `dotnet build -o /app/build` on individual projects to utilize the docker cache. (our bulids basically slowed down by 4x)
Turns out up until now, you could do a `dotnet build`  with different output directories and it wouldn't rebuild any source code - which is somehow okay since there weren't changes. But the path to referenced project outputs do change when you use the command line argument so now all the code gets recompiled.
So.. it's more correct now but may cause it to rebuild in places where you'd wonder why. 😅

rainersigwald:
Oof, that's an unexpected side effect. Unfortunately, I don't think there's any way to know at the MSBuild level that the files are logically the same to continue to short-circuit work. The paths are different which _might_ result in different outputs.

The ideal situation would be to change the output directory _only_ for the entry-point project you're building. But preventing its flow as a global property is difficult.

PR #4982, opened 2019-12-12
Fixes ClickOnce problem

### LINKED ISSUE #4824: Save in ClickOnce scenario does not update ProjectRootElement._lastWriteTimeWhenRead (regression)
See [AB#984122](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/984122)

More Info: 
1. The bug reproduces on master-29312.03 & Dev16.3 Pre4 d16.3-29311.1183.
2. This bug doesn't reproduce on master-29311.112 & Dev16.3 Pre3 d16.3-29230.61.
3. This bug doesn't repro when setting NuGet package management format to Packages.config.
4. This bug doesn't repro after rebuilt the project. 

Steps to reproduce:
1. Create a project, such as WPF/Winforms project.
2. Right click on Reference open Manage NuGet Packages… 
3. Click Settings and set the package management format to PackageReference .
4. Install a NuGet package such as Newtonsoft.Json.
5. Publish and install the project.
6. On Signing page uncheck the 'Sign the ClickOnce manifests'.
7. Publish and install the project.

Actual result: 
Cannot start the application

The warning information is:
Application validation did not succeed. Unable to continue.
See Details txt file in attachment.
Expect result:
The application should be add successfully.

Forgind:
Fixes #4824
It fixes it by essentially forcing a rebuild of all project files if this save method is ever called. That is not optimal. An alternative solution proposed is to write a special version of Save in ProjectRootElement and have users call that instead. This version is simpler but not as efficient.

### DIFF
--- a/src/Build/Construction/ProjectRootElement.cs
+++ b/src/Build/Construction/ProjectRootElement.cs
@@ -1599,6 +1599,10 @@ namespace Microsoft.Build.Construction
                 XmlDocument.Save(projectWriter);
             }
 
+            if (DateTime.Now.CompareTo(_lastWriteTimeWhenRead) > 0)
+            {
+                _lastWriteTimeWhenRead = DateTime.Now;
+            }
             _versionOnDisk = Version;
         }
 


Forgind:
This will break because we only add the last modified project to MSBuildAllProjects now for performance reasons.

PR #4984, opened 2019-12-12
Prevent thrashing

### LINKED ISSUE #1308: Extreme slowdowns when process memory usage is high
Whenever MSBuild completes a `BuildRequest`, it [checks](https://github.com/Microsoft/msbuild/blob/4e405a802d587368581117daf3402491beea8de7/src/XMakeBuildEngine/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs#L771) to see if the current process is using "too much" memory (defined as 80% of the virtual memory limit). If it is, the engine [attempts to free memory](https://github.com/Microsoft/msbuild/blob/4e405a802d587368581117daf3402491beea8de7/src/XMakeBuildEngine/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs#L854-L862) by flushing build results to a file-based cache and calling for garbage collection.

This system likely worked well at some point, when process memory usage came primarily from MSBuild and flushing results to disk released a significant amount of memory. But if the bulk of the memory usage is coming from something _else_ in the process, this can cause extreme slowdowns: every build request completion does synchronous I/O _and_ a stop-the-world GC. This is especially painful if the process in question is `devenv.exe` and it's in the midst of a solution-configuration change (i.e. "do a design-time build of every project in the solution").

Forgind:
Fixes #1308

Although I think this is a helpful fix in preventing extreme slowdowns in what is probably an edge case, it should only be rolled out to a limited number of customers at first to see if it causes major problems.

### DIFF
--- a/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
+++ b/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
@@ -111,6 +111,18 @@ namespace Microsoft.Build.BackEnd
         private readonly bool _debugForceCaching;
 
         /// <summary>
+        /// Determines the acceptable fraction of virtual memory we can use.
+        /// </summary>
+        private int memoryLimitFractionIndex = 0;
+
+        /// <summary>
+        /// The acceptable fractions of virtual memory, depending on our previous success at clearing memory. ** This is designed to prevent
+        /// something similar to thrasing as we try to clear memory, clear only a very small amount of memory, and trigger more expensive
+        /// operations in trying to clear that small amount of memory. **
+        /// </summary>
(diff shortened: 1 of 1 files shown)


Forgind:
I thought you were worried about how long it would take to implement rather than whether it's a useful fix. My mistake.

Could we theoretically do something with the VS installer, in which it generates a random float during installation and stores it in some file for us to access? Then, if we want to roll out a feature to, say, 5% of customers, we check the value in the file and only activate it if it's < 0.05.

rainersigwald:
VS has feature flags and gradual rollout capabilities already; we should not reinvent our own if we can avoid it.

Forgind:
Why can't we use theirs, then?

rainersigwald:
We have many scenarios that aren't "inside VS": command line builds and the .NET Core SDK. I don't know whether their approach works in those situations.

PR #4995, opened 2019-12-18
Fix+enable parallel build support on mono

radical:


src/Shared/NamedPipeUtil.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            if (NativeMethodsShared.IsUnixLike)
+            {
+                // If we're on a Unix machine then named pipes are implemented using Unix Domain Sockets.
+                // Most Unix systems have a maximum path length limit for Unix Domain Sockets, with
+                // Mac having a particularly short one. Mac also has a generated temp directory that
+                // can be quite long, leaving very little room for the actual pipe name. Fortunately,
+                // '/tmp' is mandated by POSIX to always be a valid temp directory, so we can use that
+                // instead.
+                return Path.Combine("/tmp", pipeName);
```
rainersigwald: I remember fighting with this and I could have sworn we had some other fix. But I can't seem to find any evidence of that.

src/Shared/NamedPipeUtil.cs
```diff
@@ -0,0 +1,26 @@
+using System;
```
rainersigwald: Copyright banner please

src/Shared/NodeEndpointOutOfProcBase.cs
```diff
@@ -196,34 +196,48 @@ internal void InternalConstruct(string pipeName)
...
-            _pipeServer = new NamedPipeServerStream
-                (
-                pipeName,
-                PipeDirection.InOut,
-                1, // Only allow one connection at a time.
-                PipeTransmissionMode.Byte,
-                PipeOptions.Asynchronous | PipeOptions.WriteThrough,
-                PipeBufferSize, // Default input buffer
-                PipeBufferSize  // Default output buffer
-#if FEATURE_NAMED_PIPE_SECURITY_CONSTRUCTOR
```
rainersigwald: Could we just avoid defining this for mono?

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -441,10 +441,14 @@ private int LaunchNode(string msbuildLocation, string commandLineArgs)
...
             string exeName = msbuildLocation;
 
-#if RUNTIME_TYPE_NETCORE
-            // Run the child process with the same host as the currently-running process.
-            exeName = GetCurrentHost();
-            commandLineArgs = "\"" + msbuildLocation + "\" " + commandLineArgs;
+#if RUNTIME_TYPE_NETCORE || MONO
+            // Mono automagically uses the current mono, to execute a managed assembly
+            if (!NativeMethodsShared.IsMono)
+            {
```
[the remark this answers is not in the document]
radical: It would, but I was trying to go towards having the same flags for netcore and mono, which would facilitate towards eventually having a single build.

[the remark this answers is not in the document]
radical: or net4*. Essentially, just trying to use runtime checks instead of the compile time ones.

src/Shared/NodeEndpointOutOfProcBase.cs
radical: Same idea here, trying to gradually add runtime checks instead of compile time ones. I can drop them, if you prefer that.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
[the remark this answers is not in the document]
rainersigwald: Sounds like a good plan.

src/Shared/NamedPipeUtil.cs
radical: Just clarifying, do I need to do anything here?

rainersigwald: Nope!

PR #5000, opened 2019-12-21
Target protocol with GetTargetPath

cdmihai:
Quickbuild calls each project with `BuildProjectReferences=false`, which makes the main p2p target be `GetTargetPath` instead of the default (default target or the `ProjectReference.Targets`). This PR updates the target protocol for static graph to reflect this.

Unfortunately, Microsoft.Managed.targets was imported before the common targets which define `BuildProjectReferences`. So I split it up in before / after imports.

### DIFF
--- a/scripts/Deploy-MSBuild.ps1
+++ b/scripts/Deploy-MSBuild.ps1
@@ -49,6 +49,8 @@ $filesToCopyToBin = @(
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.CSharp.CurrentVersion.targets"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.CSharp.targets"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.Managed.targets"
+    "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.Managed.Before.targets"
+    "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.Managed.After.targets"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.Net.props"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.NetFramework.CurrentVersion.props"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.NetFramework.CurrentVersion.targets"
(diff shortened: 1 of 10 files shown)


src/Tasks/Microsoft.CSharp.targets
```diff
@@ -17,25 +17,25 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <!--
-        In VS 2010 SP1 and VS 2012, both supported for asset compatibility, the MSBuild installed 
-        as part of them did not enforce using the local ToolsVersion (4.0) in all cases, but instead 
-        just used whatever ToolsVersion was in the project file if it existed on the machine, and 
-        only forced 4.0 if that ToolsVersion did not exist.
```
cdmihai: I don't know why VSCode did all these whitespace changes, but they seem better than what was there initially.

Forgind: True, but it also makes it harder to figure out what you actually changed. I saw you split Microsoft.Managed.targets into Microsoft.Managed.Before.targets and Microsoft.Managed.After.targets, both as far as references to them and the code, and that you (conditionally) modified _MainReferenceTarget instead of _RecursiveTargetForContentCopying. Were those the only changes?

cdmihai: Yup, I split the target file in two, because some things need to be set before the common targets (common targets read them), and some things after (common targets compute stuff read by them)

rainersigwald:
`Microsoft.Managed.targets` can be directly imported, such as here https://github.com/microsoft/MSBuildSdks/blob/0111035365b5381b868db7169f56f753c989b431/src/NoTargets/Sdk/Sdk.targets#L74-L78

Can you keep it around with just import before/after? It won't be used in normal operation but would preserve compat.

cdmihai:
@rainersigwald 
I brought the original file back from the grave :)

rainersigwald:
@cdmihai is f7ae982 ok? That way we don't have copy/paste.

scripts/Deploy-MSBuild.ps1
```diff
@@ -49,6 +49,8 @@ $filesToCopyToBin = @(
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.CSharp.CurrentVersion.targets"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.CSharp.targets"
     "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.Managed.targets"
+    "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.Managed.Before.targets"
```
Forgind: Since Before/After are both part of Microsoft.Managed.targets now, these lines are unnecessary, right?

rainersigwald: It's the other way around: Managed.targets now imports the other two. So they need to be deployed/updated.

cdmihai:
> @cdmihai is [f7ae982](https://github.com/microsoft/msbuild/commit/f7ae982b3fbd4e5cc881478f2ad336b3e373d7a9) ok? That way we don't have copy/paste.

OMG, that's brilliant, it never occurred to me I can avoid  the copy paste :)

PR #4973, opened 2019-12-11
Report long-running tests in xunit

rainersigwald:
On my dev box, the longest-running test was 24 seconds, so I'm hoping this is a reasonable time limit.

PR #5001, opened 2019-12-21
DownloadFile behaves unexpectedly in the case of OperationCanceledExceptions

ChadNedzlek:
We have some builds that are calling the DownloadFile task, and they seem to be
silently misbehaving: no errors, just hangs for a few hours and then we have to kill the process.

Looking at the code, there are a few paths through the code that result in either infinite loops or silently downloading nothing and returning success.

In particular, if a server returns a failed response, like a 500, and then the task is cancelled, it enters an infinite loop, because canRetry isn't refreshed in the loop.

Also, if a request times out the first time, where HttpClient.Send throws an OperationCanceledException, the task returns success without downloading anything.

I added tests for these two cases, and they fail without this change.

Hopefully this change is an acceptable one... I tried hard not to change any existing code behaviors (other than infinite loop and silent failure).  This will still fail and not-retry if any request times out, but that was the existing behavior (it just correctly reports and error and returns false now, instead of not downloading the requested file), and perhaps that behavior isn't desirable, in which case, IsRetriable should have a clause for OperationCanceled that returns true.

### DIFF
--- a/src/Tasks.UnitTests/DownloadFile_Tests.cs
+++ b/src/Tasks.UnitTests/DownloadFile_Tests.cs
@@ -246,6 +246,26 @@ namespace Microsoft.Build.Tasks.UnitTests
         }
 
         [Fact]
+        public void RetryOnTimeout()
+        {
+            // Http timeouts manifest as "OperationCanceledExceptions" from the handler
+            CancellationTokenSource timeout = new CancellationTokenSource();
+            timeout.Cancel();
+            DownloadFile downloadFile = new DownloadFile()
+            {
+                BuildEngine = _mockEngine,
+                HttpMessageHandler = new MockHttpMessageHandler((message, token) => throw new OperationCanceledException(timeout.Token)),
+                Retries = 1,
+                RetryDelayMilliseconds = 100,
+                SourceUrl = "http://notfound/foo.txt"
+            };
+
+            downloadFile.Execute().ShouldBeFalse(() => _mockEngine.Log);
+
+            _mockEngine.Log.ShouldContain("MSB3923", () => _mockEngine.Log);
+        }
+
+        [Fact]
         public void SkipUnchangedFiles()
         {
             using (TestEnvironment testEnvironment = TestEnvironment.Create())
--- a/src/Tasks.UnitTests/DownloadFile_Tests.cs
+++ b/src/Tasks.UnitTests/DownloadFile_Tests.cs
@@ -301,7 +301,12 @@ namespace Microsoft.Build.Tasks.UnitTests
             await Task.Delay(TimeSpan.FromSeconds(1));
             runaway.IsCompleted.ShouldBeTrue("Task did not cancel");
             if (!runaway.IsCompleted)
+            {
+                // If this task isn't completed now, it never will, so we have an early return here
+                // so that the test fails rather than enter an infinite loop if the assert above doesn't
+                // throw for whatever reason.
                 return;
+            }
 
             var result = await runaway;
             result.ShouldBeFalse(() => _mockEngine.Log);
@@ -359,7 +364,7 @@ namespace Microsoft.Build.Tasks.UnitTests
 
             _mockEngine.Log.ShouldContain("MSB3922", () => _mockEngine.Log);
         }
-        
+
         private class MockHttpContent : HttpContent
         {
             private readonly Func<Stream, Task> _func;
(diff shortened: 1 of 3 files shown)


src/Tasks/DownloadFile.cs
```diff
@@ -77,36 +83,46 @@ public override bool Execute()
             }
 
             int retryAttemptCount = 0;
-            bool canRetry = false;
+            
+            CancellationToken cancellationToken = _cancellationTokenSource.Token;
 
-            do
+            while(true)
```
Forgind: Why did you change the type of loop? I personally would prefer to see a single while or do/while loop without break statements if possible, that is, putting `canRetry = false;` in instead of breaks.

src/Tasks/DownloadFile.cs
```diff
@@ -77,36 +83,46 @@ public override bool Execute()
...
                 try
                 {
-                    Download(uri);
+                    await DownloadAsync(uri, cancellationToken);
                     break;
                 }
-                catch (OperationCanceledException)
+                catch (OperationCanceledException e) when (e.CancellationToken == cancellationToken)
                 {
+                    // This task is being cancelled, exit the loop
```
Forgind:
```suggestion
                    // This task is being cancelled. Exit the loop.
```

src/Tasks.UnitTests/DownloadFile_Tests.cs
```diff
@@ -245,6 +245,68 @@ public void RetryOnResponseError()
...
+                }),
+                Retries = 2,
+                RetryDelayMilliseconds = 100,
+                SourceUrl = "http://notfound/foo.txt"
+            };
+
+            var runaway = Task.Run(() => downloadFile.Execute());
+            await Task.Delay(TimeSpan.FromSeconds(1));
+            runaway.IsCompleted.ShouldBeTrue("Task did not cancel");
+            if (!runaway.IsCompleted)
```
Forgind: Isn't this redundant given the previous line? If not, can you add braces around it or put it on one line?

src/Tasks/DownloadFile.cs
```diff
@@ -69,6 +70,11 @@ public void Cancel()
         }
 
         public override bool Execute()
+        {
+            return ExecuteAsync().GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
ChadNedzlek: This is the (unforunate) sync wrapper around the async code, because, to my knowledge, there is no async entry point into a build Task.  So this is the only place that can't use it.

src/Tasks/DownloadFile.cs
ChadNedzlek: That is the source of the bugs.  There are many cases that _should_ break out of the loop, but failed to do so before, because it carried the value from the previous iteration of the loop unexpectedly.  There is no way to scope the variable to the body of loop, so you HAVE to carry the value between iterations, even though each iteration has nothing to do with the previous one.  By using explicit breaks, it's super clear when something means to break out of the loop.

I could change it back, and then just make sure to reset "canRetry" to false at the beginning of every loop... and then make sure it's set to true in every case where we want to retry... but that seems like a complicated and error prone way of saying "break" when we want to break.

src/Tasks.UnitTests/DownloadFile_Tests.cs
ChadNedzlek: I suppose it is, if the assumption is that a failed "Should" will _always_ abort the test.  The problem is that if this isn't trye, the link 306 will wait infinitely.  I wanted to be extra sure the test didn't enter an infinite loop, since it's really hard to report on those.

I'd rather leave the valve, but I'll add the braces and add a comment explaining why it's there.

ChadNedzlek:
I think I resolved all the outstanding issues and have pushed an updated change merged with master

src/Tasks.UnitTests/DownloadFile_Tests.cs
rainersigwald: Since xUnit assertions work via exception, I think it's safe to assume that the `ShouldBeTrue` will end the test if it fails. We also have #4973 and https://github.com/dotnet/arcade/pull/3255 coming which should help catch hung tests in the general case.

ChadNedzlek: The arcade situation will almost certainly cause it to just... lose the test results, but I'm find removing this test hardening.  Committed an update.

PR #5003, opened 2019-12-23
Error on cyclical nesting Fixes #1599

### LINKED ISSUE #1599: Prevent infinite recursion in Solution parser when a project is nested under itself
We don't check for invalid project nestings in the solution file, so it's possible to create a solution that crashes MSBuild with stack overflow:

https://github.com/Microsoft/msbuild/blob/xplat/src/XMakeBuildEngine/Construction/Solution/ProjectInSolution.cs#L393

An example solution is:
https://github.com/dotnet/corefx/blob/f6ab339f446e37505f93ef9f430801fa1dcc05b8/src/System.Collections.Immutable/System.Collections.Immutable.sln#L40-L41

Note that we should check for indirect recursion as well (loops).

Forgind:
Fixes #1599

### DIFF
--- a/src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
+++ b/src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
@@ -22,10 +22,8 @@ namespace Microsoft.Build.UnitTests.Construction
         [Fact]
         public void ParseInvalidProjectNesting()
         {
-            Assert.Throws<InvalidProjectFileException>(() =>
-            {
-                string solutionFileContents =
-                    @"
+            string solutionFileContents =
+                @"
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26124.0
@@ -51,10 +49,7 @@ Global
                 {933E47E9-37B5-4AB3-AD1C-0E48043575D7} = {933E47E9-37B5-4AB3-AD1C-0E48043575D7}
         EndGlobalSection
 EndGlobal";
-                ParseSolutionHelper(solutionFileContents);
-                Assert.True(false, "Should not get here");
-            }
-           );
+            Shouldly.Should.Throw<InvalidProjectFileException>(() => ParseSolutionHelper(solutionFileContents));
         }
 
         /// <summary>
@@ -63,10 +58,8 @@ EndGlobal";
         [Fact]
         public void ParseInvalidProjectNesting2()
         {
-            Assert.Throws<InvalidProjectFileException>(() =>
-            {
-                string solutionFileContents =
-                    @"
+            string solutionFileContents =
+                @"
 
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
@@ -94,10 +87,7 @@ Global
 	EndGlobalSection
 EndGlobal
 ";
-                ParseSolutionHelper(solutionFileContents);
-                Assert.True(false, "Should not get here");
-            }
-           );
+            Shouldly.Should.Throw<InvalidProjectFileException>(() => ParseSolutionHelper(solutionFileContents));
         }
         
         /// <summary>
--- a/src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
+++ b/src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
@@ -17,6 +17,90 @@ namespace Microsoft.Build.UnitTests.Construction
     public class SolutionFile_Tests
     {
         /// <summary>
+        /// Test that a project invalid project nestings is seen as invalid.
+        /// </summary>
+        [Fact]
+        public void ParseInvalidProjectNesting()
+        {
+            Assert.Throws<InvalidProjectFileException>(() =>
+            {
+                string solutionFileContents =
+                    @"
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26124.0
+MinimumVisualStudioVersion = 15.0.26124.0
+Project(""{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}"") = ""tempCode"", ""tempCode.csproj"", ""{933E47E9-37B5-4AB3-AD1C-0E48043575D7}""
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{933E47E9-37B5-4AB3-AD1C-0E48043575D7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+	EndGlobalSection
+        GlobalSection(NestedProjects) = preSolution
+                {933E47E9-37B5-4AB3-AD1C-0E48043575D7} = {933E47E9-37B5-4AB3-AD1C-0E48043575D7}
+        EndGlobalSection
+EndGlobal";
+                ParseSolutionHelper(solutionFileContents);
+                Assert.True(false, "Should not get here");
+            }
+           );
+        }
+
+        /// <summary>
+        /// Test that a project invalid project nestings is seen as invalid.
+        /// </summary>
+        [Fact]
+        public void ParseInvalidProjectNesting2()
+        {
+            Assert.Throws<InvalidProjectFileException>(() =>
+            {
+                string solutionFileContents =
+                    @"
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26124.0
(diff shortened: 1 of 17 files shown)


src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1211,50 @@ internal void ParseNestedProjects()
...
+                if (!DFS(parents, a))
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        // -2 indicates that there is no cycle among ancestors. -1 indicates that it has not yet been checked.
+        private bool DFS(int[] parents, int ind) {
```
danmosemsft: Naming nits:
Ind->index?
DFS->DepthFirstSearch?
a->i

Also open paren on new line

src/Shared/Resources/Strings.shared.resx
```diff
@@ -208,6 +208,10 @@
     <value>MSB5023: Error parsing the nested project section in solution file. A project with the GUID "{0}" is listed as being nested under project "{1}", but does not exist in the solution.</value>
     <comment>{StrBegin="MSB5023: "}UE: The solution filename is provided separately to loggers.</comment>
   </data>
+  <data name="SolutionContainsCyclicallyNestedProjectFiles" Visibility="Public">
+    <value>MSB5025: Nested project section in solution file {0} has a circular nesting.</value>
```
danmosemsft: I think the messages normally use quotes around file paths

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1211,51 @@ internal void ParseNestedProjects()
...
+                if (!DepthFirstSearch(parents, i))
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        // -2 indicates that there is no cycle among ancestors. -1 indicates that it has not yet been checked.
+        private bool DepthFirstSearch(int[] parents, int index)
```
BenVillalobos: Can we change the name of `DepthFirstSearch` or add a comment here? Maybe `DepthFirstSearch_FindCycle` to add context to this function (particularly about what it means when it returns true or false)

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1211,51 @@ internal void ParseNestedProjects()
...
+            string parentGuid = _projectsInOrder[index].ParentProjectGuid;
+            if (parentGuid == null)
+            {
+                parents[index] = -2;
+                return true;
+            }
+            int parent = _projectsInOrder.IndexOf(_projects[parentGuid]);
+            if (parents[parent] == -2)
+            {
+                parents[index] = -2;
```
cdmihai: nit: use an Enum to name the indexes.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
```diff
@@ -16,6 +16,80 @@ namespace Microsoft.Build.UnitTests.Construction
...
+        EndGlobalSection
+EndGlobal";
+            Shouldly.Should.Throw<InvalidProjectFileException>(() => ParseSolutionHelper(solutionFileContents));
+        }
+
+        /// <summary>
+        /// Test that a project invalid project nestings is seen as invalid.
+        /// </summary>
+        [Fact]
+        public void ParseInvalidProjectNesting2()
```
cdmihai: I suggest adding many more graph shapes (single nodes, multiple notes, cross edges, diamond dependencies) X (with cycles, without cycles). To avoid the nuisance of specifying the whole solution each time, you can generate it instead. You can reuse / adapt the code from here: [https://github.com/cdmihai/msbuild/blob/7bcb2f961093e87277439420d351a9f3f3a8cdb8/src/Build.UnitTests/Graph/GraphLoadedFromSolution_tests.cs#L263](https://github.com/cdmihai/msbuild/blob/7bcb2f961093e87277439420d351a9f3f3a8cdb8/src/Build.UnitTests/Graph/GraphLoadedFromSolution_tests.cs#L263)

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1218,50 @@ internal void ParseNestedProjects()
...
+
+        private bool AncestorsContainNoCycle(SearchState[] parents, int index)
+        {
+            string parentGuid = _projectsInOrder[index].ParentProjectGuid;
+            if (parentGuid == null)
+            {
+                parents[index] = SearchState.AncestorsContainNoCycles;
+                return true;
+            }
+            int parent = _projectsInOrder.IndexOf(_projects[parentGuid]);
```
cdmihai: A project can have multiple parents right? What if the cycle is not on the first parent that's found?  `a->[b, c], b->d, c->e, e->[f, a], d->f`
Add a test case for this.

Forgind: Not in this case; it's what you're directly nested under.

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
Forgind: Although I agree the testing could be more rigorous (As an alternative, I could make it into a theory and very cleanly tried a variety of different shapes), the original issue was posted almost three years ago and got no attention until I mentioned to Rainer that it looked like a quick, fun fix. I don't think it's worth taking the time to add lots of test cases to make it safer.

src/Shared/Resources/Strings.shared.resx
```diff
@@ -208,6 +208,10 @@
     <value>MSB5023: Error parsing the nested project section in solution file. A project with the GUID "{0}" is listed as being nested under project "{1}", but does not exist in the solution.</value>
     <comment>{StrBegin="MSB5023: "}UE: The solution filename is provided separately to loggers.</comment>
   </data>
+  <data name="SolutionContainsCyclicallyNestedProjectFiles" Visibility="Public">
+    <value>MSB5025: Nested project section in solution file "{0}" has a circular nesting.</value>
```
rainersigwald: You're using MSB5025 in #4899

rainersigwald: I sorta understand what this message means, but it's not very actionable. What does VS do in this case? Can we crib a message from there? Can we include the nature of the cycle? Or at least one project ID in the cycle?

src/Build.OM.UnitTests/Construction/SolutionFile_Tests.cs
rainersigwald:
```suggestion
        public void ParseInvalidProjectNestingIndirect()
```

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1218,50 @@ internal void ParseNestedProjects()
...
+            for (int i = 0; i < _projectsInOrder.Count; i++) {
+                if (!AncestorsContainNoCycle(parents, i))
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        private bool AncestorsContainNoCycle(SearchState[] parents, int index)
```
```suggestion
        private bool AncestorsContainNoCycle(SearchState[] state, int index)
```

? I'm finding all the indexing very confusing.

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1218,50 @@ internal void ParseNestedProjects()
...
+        {
```
rainersigwald: Should there be an early check/out here on `state[index] == SearchState.AncestorsContainNoCycles`? That could happen if an early project was parented to a later one, right?

src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -1211,6 +1218,50 @@ internal void ParseNestedProjects()
...
+            ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile(ContainsNoCyclicalNesting(), "SubCategoryForSolutionParsingErrors",
+                new BuildEventFileInfo(FullPath, _currentLineNumber, 0), "SolutionContainsCyclicallyNestedProjectFiles", _solutionFile);
+        }
+
+        private bool ContainsNoCyclicalNesting()
+        {
+            SearchState[] parents = new SearchState[_projectsInOrder.Count];
+            for (int a = 0; a < _projectsInOrder.Count; a++) {
+                parents[a] = SearchState.Unchecked;
+            }
```
rainersigwald: Change the enum so that `Unchecked` has the backing value `0` (is listed first, optionally with `= 0`). Then the default-initialized array will do this for you.

[sharplab demonstration of this](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUBgBjwwGYACAbzzJrIAcwBLANwENgBTMjqCAWzIBlDqzABjABaDg7LiDIAjAJ6dqtKrlpayAVSiSOYgNYc4ZALxkCMNdrIBBfRwDOwAPZhnAYTdQZjKAA5Ny8lMQAbFxtNOzIvCUMTOFsyAF88FIx0ShStYVFJaVkAbQBdelEeYGcLMigOAHchEXEpGU5izAJSgG5c2gAzDzIACgDgMkZagh7JsgAeCrAq5wA6ABkeAHNgCVnGREQASn6aDVjtVEwAThG6Sr9nYsZSo76Yu3SPtLxUoA)

src/Build/Construction/Solution/SolutionFile.cs
rainersigwald: Negative names in method names are confusing but this is contained and I think I already thought through it so I won't make you change it now.

src/Shared/Resources/Strings.shared.resx
Forgind: I know, but that's an easy change, so to (hopefully) avoid confusion as to why I skipped some numbers, I took the next one instead.

PR #5004, opened 2019-12-24
Introduce ExecuteTask ETW

### LINKED ISSUE #1581: Print time taken in executing individual task
Hello,

Is it possible to print the time taken in executing individual task in msbuild? I know there is /displaysummary option in msbuild but it prints the aggregate time of all invocation of tasks. What I am looking at is the time taken for executing 1 invocation of task. Below given are the kind of logs that where I want time taken to be printed.

Done executing task "CL". [I need something like **Done executing task "CL" in 15 seconds**.]
Done executing task "MSBuild".
Done executing task "AddToManifest".

This will be very helpful to us to investigate slow build issues.

Ravinder

Forgind:
This creates a new eventsource pair: ExecuteTaskStart and ExecuteTaskStop. Together, they allow a user to see the name of a task and how long it took to execute (including exact start and stop time) for each task. Another pair is added to capture how long each of yield and reacquire takes (if relevant).

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
@@ -326,7 +326,7 @@ namespace Microsoft.Build.BackEnd
                 if (MSBuildEventSource.Log.IsEnabled())
                 {
                     startTime = MSBuildEventSource.GetThreadTime();
-                    MSBuildEventSource.Log.ExecuteTaskStart(_taskNode == null ? null : _taskNode.Name);
+                    MSBuildEventSource.Log.ExecuteTaskStart(_taskNode?.Name);
                 }
                 // Loop through each of the batch buckets and execute them one at a time
                 for (int i = 0; i < buckets.Count; i++)
@@ -345,7 +345,8 @@ namespace Microsoft.Build.BackEnd
                 if (MSBuildEventSource.Log.IsEnabled())
                 {
                     long endTime = MSBuildEventSource.GetThreadTime();
-                    MSBuildEventSource.Log.ExecuteTaskStop(_taskNode == null ? null : _taskNode.Name, startTime == -1 || endTime == -1 ? null : string.Empty + (endTime - startTime));
+                    TimeSpan time = TimeSpan.FromMilliseconds(startTime == -1 || endTime == -1 ? -1 : endTime - startTime);
+                    MSBuildEventSource.Log.ExecuteTaskStop(_taskNode?.Name, time.TotalMilliseconds == -1 ? null : string.Empty + time.Duration().Milliseconds);
                 }
                 taskResult = aggregateResult;
             }
--- a/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
@@ -15,6 +15,7 @@ using System.Threading.Tasks;
 using Microsoft.Build.Collections;
 using Microsoft.Build.Construction;
 using Microsoft.Build.Evaluation;
+using Microsoft.Build.Eventing;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.Execution;
 using Microsoft.Build.Framework;
@@ -324,6 +325,8 @@ namespace Microsoft.Build.BackEnd
 
                 WorkUnitResult aggregateResult = new WorkUnitResult();
 
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                MSBuildEventSource.Log.ExecuteTaskStart(_taskNode == null ? null : _taskNode.Name);
                 // Loop through each of the batch buckets and execute them one at a time
                 for (int i = 0; i < buckets.Count; i++)
                 {
@@ -337,6 +340,8 @@ namespace Microsoft.Build.BackEnd
                         break;
                     }
                 }
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                MSBuildEventSource.Log.ExecuteTaskStop(_taskNode == null ? null : _taskNode.Name);
 
                 taskResult = aggregateResult;
             }
--- a/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
@@ -2,18 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
-using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Build.Collections;
-using Microsoft.Build.Construction;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Eventing;
 using Microsoft.Build.Exceptions;
@@ -326,7 +322,12 @@ namespace Microsoft.Build.BackEnd
                 WorkUnitResult aggregateResult = new WorkUnitResult();
 
                 // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
-                MSBuildEventSource.Log.ExecuteTaskStart(_taskNode == null ? null : _taskNode.Name);
+                long startTime = -1;
+                if (MSBuildEventSource.Log.IsEnabled())
+                {
+                    startTime = MSBuildEventSource.GetThreadTime();
+                    MSBuildEventSource.Log.ExecuteTaskStart(_taskNode == null ? null : _taskNode.Name);
+                }
                 // Loop through each of the batch buckets and execute them one at a time
                 for (int i = 0; i < buckets.Count; i++)
                 {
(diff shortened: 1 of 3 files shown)


src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -324,6 +325,8 @@ private async Task<WorkUnitResult> ExecuteTask(TaskExecutionMode mode, Lookup lo
 
                 WorkUnitResult aggregateResult = new WorkUnitResult();
 
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                MSBuildEventSource.Log.ExecuteTaskStart(_taskNode == null ? null : _taskNode.Name);
```
danmosemsft: What if the task calls IBuildEngineX.Yield? The events will bracket more than the time the task spent executing.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -337,7 +341,12 @@ private async Task<WorkUnitResult> ExecuteTask(TaskExecutionMode mode, Lookup lo
                         break;
                     }
                 }
-
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                if (MSBuildEventSource.Log.IsEnabled())
+                {
+                    long endTime = MSBuildEventSource.GetThreadTime();
```
BenVillalobos: @Forgind mentioned that some tests were coming up with negative times in perfview. @brianrob would you happen to know any reason why this would happen?

Inspiration for this idea came from [here](https://www.codeproject.com/Articles/31152/ExecutionStopwatch)

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -324,6 +321,13 @@ private async Task<WorkUnitResult> ExecuteTask(TaskExecutionMode mode, Lookup lo
 
                 WorkUnitResult aggregateResult = new WorkUnitResult();
 
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                long startTime = -1;
+                if (MSBuildEventSource.Log.IsEnabled())
+                {
+                    startTime = MSBuildEventSource.GetThreadTime();
+                    MSBuildEventSource.Log.ExecuteTaskStart(_taskNode == null ? null : _taskNode.Name);
```
BenVillalobos:
```suggestion
                    MSBuildEventSource.Log.ExecuteTaskStart(_taskNode?.Name);
```

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -337,7 +341,12 @@ private async Task<WorkUnitResult> ExecuteTask(TaskExecutionMode mode, Lookup lo
                         break;
                     }
                 }
-
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                if (MSBuildEventSource.Log.IsEnabled())
+                {
+                    long endTime = MSBuildEventSource.GetThreadTime();
+                    MSBuildEventSource.Log.ExecuteTaskStop(_taskNode == null ? null : _taskNode.Name, startTime == -1 || endTime == -1 ? null : string.Empty + (endTime - startTime));
```
BenVillalobos:
```suggestion
                    MSBuildEventSource.Log.ExecuteTaskStop(_taskNode?.Name, startTime == -1 || endTime == -1 ? null : string.Empty + (endTime - startTime));
```

src/Framework/MSBuildEventSource.cs
```diff
@@ -361,6 +370,32 @@ public void MSBuildExeStop(string commandLine)
...
+        }
+
+        #endregion
+
+        #region Non-Events
+        [NonEvent]
+        public static long GetThreadTime()
+        {
+            long kernelTime, userTime;
+            if (!Convert.ToBoolean(GetThreadTimes(GetCurrentThread(), out _, out _, out kernelTime, out userTime)))
```
BenVillalobos: Is it a realistic scenario that the values for kernelTime and userTime would overflow? If so, how can we test for that?

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
Forgind: I experimented some more, and with the latest commit, they now show up as always non-negative (as far as I can tell). I think the problem was that, as I noted in our call, there can be overflow between if a "stopwatch" starts near the max time and ends shortly after it, since that would seem like "negative" time. When I used TimeSpan.Duration(), that problem went away.

src/Framework/MSBuildEventSource.cs
Forgind: See above comment. I think the fact that Duration() made them all non-negative in building MSBuild.sln means there was overflow, and it is now corrected.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -324,6 +321,13 @@ private async Task<WorkUnitResult> ExecuteTask(TaskExecutionMode mode, Lookup lo
 
                 WorkUnitResult aggregateResult = new WorkUnitResult();
 
+                // Some tests do not provide an actual taskNode; checking if _taskNode == null prevents those tests from failing.
+                long startTime = -1;
+                if (MSBuildEventSource.Log.IsEnabled())
+                {
+                    startTime = MSBuildEventSource.GetThreadTime();
+                    MSBuildEventSource.Log.ExecuteTaskStart(_taskNode?.Name);
```
cdmihai: I think logging the task ID would be better. It would help users correlate the msbuild log task started / ended events and the ETW events.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
cdmihai: @danmosemsft What would be the cleanest way of subtracting yields?

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
Forgind: I don't see the use case for this? The original issue mentioned wanting to see `Done executing task "CL" in 15 seconds` or the equivalent to help with investigating slow builds. That means you would need to know what function (its name) was slow but no more.

Also, I didn't see an ID attribute of _taskNode.

cdmihai: The issue refers to tasks that run multiple times in a build (e.g. for each project), and the user needs to differentiate them (otherwise, for singular tasks, they can just in the existing perf summary). I think it'd be helpful for them to be able to place the task in the event hierarchy, using the task id, so they can know which project is running that task. TaskBuilder should have the id somewhere, it's the one that logs task started / ended events, which contain the ID.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
brianrob: I would strongly recommend that rather than calculating times in your code that you do this calculation on the resulting trace.  If you're using PerfView it will do the calculation for you automatically in the Events view.

src/Framework/MSBuildEventSource.cs
brianrob: I'd recommend staying out of the business of doing this type of calculation and just use the trace time stamps to calculate the time.  Also note that this won't work on Unix as it depends on Windows-specific pinvokes.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
Forgind: That's DURATION_MSEC, I imagine? I'd thought that was wall clock time rather than execution time. @danmosemsft, @cdmihai, does that resolve the IBuildEngineX.Yield problem?

brianrob: Are you trying to capture CPU time?  If so, I'd recommend using an ETW trace that captures CPU samples rather than asking the kernel for this information.  The reason is that what ends up happening is you get a trace that contains the numbers, but then the next question is "where is that time spent?", and you can answer that with a CPU trace, and the barrier to collection of CPU events is the same as collecting the new event that you're adding here.

src/Build/BackEnd/Components/RequestBuilder/TaskHost.cs
```diff
@@ -361,7 +364,11 @@ public void Reacquire()
                 IRequestBuilderCallback builderCallback = _requestEntry.Builder as IRequestBuilderCallback;
                 ErrorUtilities.VerifyThrow(_yieldThreadId != -1, "Cannot call Reacquire() before Yield().");
                 ErrorUtilities.VerifyThrow(_yieldThreadId == Thread.CurrentThread.ManagedThreadId, "Cannot call Reacquire() on thread {0} when Yield() was called on thread {1}", Thread.CurrentThread.ManagedThreadId, _yieldThreadId);
+                MSBuildEventSource.Log.ExecuteTaskYieldStop(_taskLoggingContext.TaskName, _taskLoggingContext.BuildEventContext.TaskId);
+                MSBuildEventSource.Log.ExecuteTaskReacquireStart(_taskLoggingContext.TaskName, _taskLoggingContext.BuildEventContext.TaskId);
                 builderCallback.Reacquire();
+                MSBuildEventSource.Log.ExecuteTaskReacquireStop(_taskLoggingContext.TaskName, _taskLoggingContext.BuildEventContext.TaskId);
+                MSBuildEventSource.Log.ExecuteTaskStart(_taskLoggingContext.TaskName, _taskLoggingContext.BuildEventContext.TaskId);
```
rainersigwald: We talked offline about leaving TaskStart/Stop total wall-clock end-to-end time.

PR #5006, opened 2019-12-26
Add ability to supply an EvaluationContext to a directly created ProjectInstance

dfederm:


src/Build/Instance/ProjectInstance.cs
```diff
@@ -607,6 +646,43 @@ private ProjectInstance(ProjectInstance that, bool isImmutable, RequestedProject
...
 
+        /// <summary>
+        /// Create a file based ProjectInstance.
+        /// </summary>
+        /// <param name="file">The file to evaluate the ProjectInstance from.</param>
+        /// <param name="options">The <see cref="ProjectOptions"/> to use.</param>
+        /// <returns></returns>
+        public static ProjectInstance FromFile(string file, ProjectOptions options)
+        {
+            // string projectFile, IDictionary<string, string> globalProperties, string toolsVersion, string subToolsetVersion, ProjectCollection projectCollection, EvaluationContext evaluationContext
```
rainersigwald: ?

dfederm: Lol oops. I had pasted this from the ctor I was targeting to get the right arguments in the right order since intellisense kept choosing the wrong one.

PR #5007, opened 2019-12-26
Read switches from any _MSBUILD_ env var

danmoseley:
Fixes https://github.com/microsoft/msbuild/issues/5005 

(In case you believe this is worthwhile.)

### DIFF
--- a/src/MSBuild.UnitTests/XMake_Tests.cs
+++ b/src/MSBuild.UnitTests/XMake_Tests.cs
@@ -2061,29 +2061,36 @@ namespace Microsoft.Build.UnitTests
         }
 
         [Theory]
-        [InlineData("", "[v0][]")] // empty does nothing
-        [InlineData(" ", "[v0][]")] // whitespace does nothing
-        [InlineData("   /p:p1=v1a  /p:p1=v1b   ", "[v1b][]")]   // simple case, override order, lead/trail whitespace
-        [InlineData("/p:p1=v1 /p:p2=\"v2a v2b\"", "[v1][v2a v2b]")] // split quoted values correctly
+        [InlineData("", "[v0a][][v3][v4]")] // empty does nothing
+        [InlineData(" ", "[v0a][][v3][v4]")] // whitespace does nothing
+        [InlineData("   /p:p1=v1a  /p:p1=v1b   ", "[v1b][][v3][v4]")]   // simple case, override order, lead/trail whitespace
+        [InlineData("/p:p1=v1 /p:p2=\"v2a v2b\"", "[v1][v2a v2b][v3][v4]")] // split quoted values correctly
         [InlineData("/p:p1=\"username is %username%\"", "[username is %username%][]")] // expand env vars, like for response file content
         public void ArgumentsPulledFromEnvironmentVariable(string value, string expected)
         {
+            expected = Environment.ExpandEnvironmentVariables(expected);
+
             try
             {
                 Environment.SetEnvironmentVariable("_MSBUILD_", value);
 
-                string projectContents = ObjectModelHelpers.CleanupFileContents(@"<Project xmlns=""http://schemas.microsoft.com/developer/msbuild/2003"">
+                using (TestEnvironment testEnvironment = UnitTests.TestEnvironment.Create())
+                {
+                    var testProject = testEnvironment.CreateFile("test.proj", ObjectModelHelpers.CleanupFileContents(@"<Project xmlns=""http://schemas.microsoft.com/developer/msbuild/2003"">
+                      <Target Name=""Build"">
+                        <Message Text=""[$(p1)][$(p2)][$(p3)][$(p4)]"" />
+                      </Target>
+                    </Project>"));
 
-  <Target Name=""Build"">
-    <Message Text=""[$(p1)][$(p2)]"" />
-  </Target>
+                    // pass a property by response file to ensure that it is lower in precedence
+                    var responseFile = testEnvironment.CreateFile("my.rsp", "/p:p1=v0b /p:p3=v3");
 
-</Project>");
+                    // pass a property to msbuild.exe to ensure that it is lower in precedence
+                    string output = RunnerUtilities.ExecMSBuild($"\"{testProject.Path}\" @\"{responseFile.Path}\" /p:p1=v0a /p:p4=v4", out bool success, _output);
 
-                expected = Environment.ExpandEnvironmentVariables(expected);
-                string logContents = ExecuteMSBuildExeExpectSuccess(projectContents, null, new string[] { "/p:p1=v0" } );
-
-                logContents.ShouldContain(expected);
+                    success.ShouldBeTrue(output);
+                    output.ShouldContain(expected);
+                }
             }
             finally
             {
--- a/src/MSBuild.UnitTests/XMake_Tests.cs
+++ b/src/MSBuild.UnitTests/XMake_Tests.cs
@@ -2063,7 +2063,7 @@ namespace Microsoft.Build.UnitTests
         [Theory]
         [InlineData("", "[v0][]")] // empty does nothing
         [InlineData(" ", "[v0][]")] // whitespace does nothing
-        [InlineData("/p:p1=v1", "[v1][]")]   // simple case
+        [InlineData("   /p:p1=v1a  /p:p1=v1b   ", "[v1b][]")]   // simple case, override order, lead/trail whitespace
         [InlineData("/p:p1=v1 /p:p2=\"v2a v2b\"", "[v1][v2a v2b]")] // split quoted values correctly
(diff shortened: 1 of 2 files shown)


Forgind:
Sounds good to me. Do you think it would be better to be able to parse values without `/p:`? That is, `p1=v1 p2="v2a v2b"` instead of `/p:p1=v1 /p:p2="v2a v2b"`.

cdmihai:
Why can't you use a `directory.build.rsp` file to achieve the same thing?

Forgind:
You could put it in a .gitignore, but otherwise it would still close up the git diff. Also, it might be useful in a remote environment if you send the code there, then want to modify the build without adding an extra pre-build step.

danmosemsft:
> Why can't you use a directory.build.rsp file to achieve the same thing?

As @Forgind suggested, sometimes it's easier to control the command line than the machine content, and it also is non persistent/doesn't show in the diff. (There might even be one in the repo already). Just another option.

> Do you think it would be better to be able to parse values without /p:? That is, p1=v1 p2="v2a v2b" instead of /p:p1=v1 /p:p2="v2a v2b".

I'm not sure I follow -- in my use case, I wanted to add a logger. So it needs to accept arbitrary parameters -- exactly the same as a .rsp file.

Forgind:
> I'm not sure I follow -- in my use case, I wanted to add a logger. So it needs to accept arbitrary parameters -- exactly the same as a .rsp file.

It probably isn't worth it. I was wondering if you could check each switch as it comes in and automatically append `/p:` if necessary. It would slightly reduce typing for future users, but it would add significantly the complexity.

cdmihai:
Worried this might hurt debug-ability. If a build is crashing due to an environment switch, there'd be little trace of it. You'd have to look in the msbuild log, in a *root* project's property list (the env var will leak into the msbuild properties). This is very indirect, and I can easily see how people might waste hours on this (I know I would :)).

Forgind:
True. I know someone who did exactly that recently. He wasn't happy.

danmosemsft:
>  You'd have to look in the msbuild log, in a root project's property list (the env var will leak into the msbuild properties). 

Good point. Couldn't the same occur with the existing directory.build.rsp mechanism though? Assuming the directory.build.rsp is already committed so it doesn't show up as a diff. Although, perhaps it would have gone through CI.

Also, arguably this applies to the myriad MSBUILD* environment variables.

danmosemsft:
I suppose if `_MSBUILD_` is being picked up, it could write a line prominently to stdout/console.

I actually didn't know about `directory.build.rsp` - I assume that doesn't apply when you're building in VS, as there's no command line parser - does it cause discrepancies between VS and msbuild.exe builds?

cdmihai:
> I assume that doesn't apply when you're building in VS, as there's no command line parser - does it cause discrepancies between VS and msbuild.exe builds? 

It gets read only in msbuild.exe, so build manager based builds won't get it.

>it could write a line prominently to stdout/console.What do you think?

Yea, I like that idea. In general, I think it would help debugging if msbuild printed all the sources it loaded cmd line args from. Biggest caveat that I see here is that MSBuild.exe cannot easily log build events, as there's no `LoggingService` at that point. I wouldn't be opposed to changing that, since it would help with other things, like logging the command line arguments. But I guess stdout printing would also be a good enough solution.

cdmihai:
Regarding logging these command line switches, looks like @KirillOsenkov started working on it in #5026. After he's done with that PR, you could reuse the logic to log command line args sources and their order.

rainersigwald:
Team triage: closing for now to pick up more discussion. This implementation looks good and should be the basis of the feature if we decide to take it in the future (except for adding logging). More details: https://github.com/microsoft/msbuild/issues/5005#issuecomment-591084947

If reopening this PR, add logging using the deferred mechanism introduced in #5055.

KirillOsenkov:
I strongly think we should do this. @mikestall recently requested a way to add /bl when the MSBuild command line was buried deep in the system and not easily accessible.

KirillOsenkov:
@danmoseley would you like to pick this up? Or should me or someone from the team adopt this PR and land it?

Forgind:
Related: Is there a way to read switches from properties? I had someone say they couldn't generate a real binlog because their project created a Project object and called Build on it. There was no obvious command line, and I didn't see an overload for adding one.

danmoseley:
@KirillOsenkov thanks for asking. I don't plan to pick this up again. You are welcome to, and I still think it would be valuable . Feel free to close this PR or repurpose it.

BenVillalobos:
@marcpopMSFT did you want to put this into 16.10?

marcpopMSFT:
@KathleenDollard this needs PM input.

Forgind:
@KirillOsenkov, what do you think still needs to be done? This looks pretty complete to me. I made a couple modifications (not pushed yet) that resolve the compile error and add a unit test case, but it looks otherwise good to me.

KirillOsenkov:
Just add a logging message at the beginning of the build like Rainer says: https://github.com/dotnet/msbuild/pull/5007#issuecomment-591085041

Using the additional command-line arguments “***” specified in the  _MSBUILD_ environment variable.

Probably in the same place that warns about single-core build or logs the binlog file name.

marcpopMSFT:
We had a discussion with the engineering system team and there are some concerns.  The recommendation is to hold off until @rainersigwald is back and then get together to talk through what folks are trying to achieve with this variable to determine if this is the right path forward.  The concerns were around reliability since customers could become reliant on that variable for the builds to work and be consistent which is the top issue facing internal builds.

Forgind:
https://github.com/danmoseley/msbuild/pull/1

danmoseley:
Just curious, are you planning to continue with this? I can't forget it as when I look at my list of PR's it's been there since 2019 ..!

rainersigwald:
I still don't think this is worth the prospect of additional hard-to-debug complexity.

danmoseley:
@rainersigwald could you say a little more? My change was very simple and localized. I know it would be valuable to me. Did this grow into a more complex proposal?

KirillOsenkov:
I’d love to see this happen too, and if someone is setting that env var, they know what they’re doing. We’ll be logging the var and the extra arguments anyway. It’s just like an .rsp file.

KirillOsenkov:
I primarily need it to inject /bl in a build where msbuild is launched indirectly from some blackbox system, like cake, nuke etc. Without this, and without modifying the blackbox system, you’re just out of luck.

KirillOsenkov:
Feel free to give the env var a scarier name

cdmihai:
> I primarily need it to inject /bl in a build where msbuild is launched indirectly from some blackbox system, like cake, nuke etc. Without this, and without modifying the blackbox system, you’re just out of luck.

@KirillOsenkov Now you can inject a binlogger in the BuildManager by setting `MSBuildDebugEngine`: https://github.com/dotnet/msbuild/blob/main/src/Build/BackEnd/BuildManager/BuildManager.cs#L524
Not quite the full power of this PR though.

rainersigwald:
I like the idea of being able to easily get a binlog (or whatever) when debugging in an unfamiliar environment. I am concerned about this becoming a bugbear we have to constantly check on when debugging "works on my machine" type scenarios.

The `MSBuildDebugEngine` stuff sidesteps this by affecting only logging which I'm prepared to treat as "safe".

Does anyone have a use case for the environment-variable feature _besides_ logging?

danmoseley:
I like to add `/flp9;v=d` but I may be an old timer and I should get with the `/bl` program.

My other scenarios are eg
*  `/pp:pp%random%.txt`
*  `/p:foo=bar` (where foo might be MyDumpExtraInfo property or ForcePassingOptimizedFlagToCsc or PathToSomeLocalTaskDllThatFeedsToAUsingTask or something)

danmoseley:
Also, it's easier to know about `__MSBUILD__` (not least it follows existing pattern for CL etc) than it is to know to set both MSBuildDebugEngine=1 and MSBuildDebugProcessName=msbuild.exe (actually, is it dotnet.exe?)

cdmihai:
> Also, it's easier to remember `__MSBUILD__` (not least it follows existing pattern for CL etc) than it is to know to set both `MSBuildDebugEngine=1 `and `MSBuildDebugProcessName=msbuild.exe` (actually, is it dotnet.exe?)

Not necessary to set `MSBuildDebugProcessName`, if it's empty it just fallbacks to checking `MSBuildDebugEngine`. MSBuildDebugProcessName is there when debugging a complicated multi proc app that calls msbuild from different processes (like VS) and you want to restrict logs to only one of the differently named processes

danmoseley:
> I am concerned about this becoming a bugbear we have to constantly check on when debugging "works on my machine" type scenarios.

If it's fully logged, would it be much different to a regular command line flag? These can already come from an nth-nested response file.

PR #5008, opened 2019-12-27
Expose project imports on ProjectInstance

dfederm:
Currently the only reason QuickBuild and [MSBuildPrediction](https://github.com/Microsoft/MSBuildPrediction) use `Project` instead of `ProjectInstance` is because imports are not exposed through the `ProjectInstance`. This would allow both to avoid `Project` and use `ProjectInstance` directly.

Note that I decided to just expose full paths and not the `ResolvedImport` because `ResolvedImport` is not currently translatable as it contains a pretty deep object graph which also isn't translatable. Instead of implementing `ITranslatable` on a whole bunch of stuff (including XML...), I decided to simplify and only expose the full file paths.

### DIFF
--- a/src/Build/Instance/ProjectInstance.cs
+++ b/src/Build/Instance/ProjectInstance.cs
@@ -1882,7 +1882,7 @@ namespace Microsoft.Build.Execution
 
         /// <summary>
         /// Translate the project instance to or from a stream.
-        /// Only translates global properties, properties, items, imports, and mutability.
+        /// Only translates global properties, properties, items, and mutability.
         /// </summary>
         void ITranslatable.Translate(ITranslator translator)
         {
@@ -1919,16 +1919,13 @@ namespace Microsoft.Build.Execution
             translator.Translate(ref _isImmutable);
             translator.Translate(ref _evaluationId);
 
-            translator.Translate(ref _imports);
-            Imports = _imports.AsReadOnly();
-
-            translator.Translate(ref _importsIncludingDuplicates);
-            ImportsIncludingDuplicates = _importsIncludingDuplicates.AsReadOnly();
-
             translator.TranslateDictionary(
                 ref _itemDefinitions,
                 ProjectItemDefinitionInstance.FactoryForDeserialization,
                 capacity => new RetrievableEntryHashSet<ProjectItemDefinitionInstance>(capacity, MSBuildNameIgnoreCaseComparer.Default));
+
+            // ignore _imports/Imports. Only used by public API users, not nodes
+            // ignore _importsIncludingDuplicates/ImportsIncludingDuplicates. Only used by public API users, not nodes
         }
 
         private void TranslateToolsetSpecificState(ITranslator translator)
@@ -2638,18 +2635,26 @@ namespace Microsoft.Build.Execution
         /// </summary>
         private void CreateImportSnapshot(Evaluation.Project.Data data)
         {
-            _imports = new List<string>(data.ImportClosure.Count);
+            _imports = new List<string>(data.ImportClosure.Count - 1 /* outer project */);
             foreach (var resolvedImport in data.ImportClosure)
             {
-                _imports.Add(resolvedImport.ImportedProject.FullPath);
+                // Exclude outer project itself
+                if (resolvedImport.ImportingElement != null)
+                {
+                    _imports.Add(resolvedImport.ImportedProject.FullPath);
+                }
             }
 
             Imports = _imports.AsReadOnly();
 
-            _importsIncludingDuplicates = new List<string>(data.ImportClosureWithDuplicates.Count);
+            _importsIncludingDuplicates = new List<string>(data.ImportClosureWithDuplicates.Count - 1 /* outer project */);
             foreach (var resolvedImport in data.ImportClosureWithDuplicates)
             {
-                _importsIncludingDuplicates.Add(resolvedImport.ImportedProject.FullPath);
+                // Exclude outer project itself
+                if (resolvedImport.ImportingElement != null)
+                {
+                    _importsIncludingDuplicates.Add(resolvedImport.ImportedProject.FullPath);
+                }
             }
 
             ImportsIncludingDuplicates = _importsIncludingDuplicates.AsReadOnly();
--- a/src/Build/Instance/ProjectInstance.cs
+++ b/src/Build/Instance/ProjectInstance.cs
@@ -89,9 +89,9 @@ namespace Microsoft.Build.Execution
 
         private List<string> _initialTargets;
 
-        private List<string> _imports;
+        private List<string> _importPaths;
 
-        private List<string> _importsIncludingDuplicates;
+        private List<string> _importPathsIncludingDuplicates;
 
         /// <summary>
         /// The global properties evaluation occurred with.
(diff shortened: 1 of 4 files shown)


src/Build/Instance/ProjectInstance.cs
```diff
@@ -1358,22 +1383,23 @@ ICollection<ProjectItemInstance> IItemProvider<ProjectItemInstance>.GetItems(str
...
+            ((IEvaluatorData<ProjectPropertyInstance, ProjectItemInstance, ProjectMetadataInstance, ProjectItemDefinitionInstance>)this).RecordImportWithDuplicates(importElement, import, versionEvaluated);
         }
 
         /// <summary>
         /// Record an import opened during evaluation. Include duplicates
-        /// Does nothing: not needed for project instances.
         /// </summary>
         void IEvaluatorData<ProjectPropertyInstance, ProjectItemInstance, ProjectMetadataInstance, ProjectItemDefinitionInstance>.RecordImportWithDuplicates(ProjectImportElement importElement, ProjectRootElement import, int versionEvaluated)
         {
+            _importsIncludingDuplicates.Add(import.FullPath);
```
Forgind: These methods don't seem like they're ever used?

Forgind: Also, if either of these is called, it would mean `Imports` and `ImportsWithDuplicates` would be out-of-date relative to `_imports` and `_importsIncludingDuplicates`.

src/Build/Instance/ProjectInstance.cs
```diff
@@ -773,6 +786,18 @@ public string FullPath
             { return _itemDefinitions; }
         }
 
+        /// <summary>
+        /// The full file paths of all the files that during evaluation contributed to this project instance.
+        /// This does not include projects that were never imported because a condition on an Import element was false.
+        /// The outer ProjectRootElement that maps to this project instance itself is not included.
+        /// </summary>
+        public IReadOnlyList<string> Imports { get; private set; }
```
Forgind: Is there a reason these have to be read only? I don't know why you'd want to modify the list of imports, but I don't think it really hurts to give people the option, since that means you have to use `_imports.AsReadOnly()` instead of just `_imports`, for instance, which takes up a little extra time and space.

src/Build/Instance/ProjectInstance.cs
```diff
@@ -1893,6 +1919,12 @@ private void TranslateAllState(ITranslator translator)
             translator.Translate(ref _isImmutable);
             translator.Translate(ref _evaluationId);
 
+            translator.Translate(ref _imports);
```
Forgind: Would you mind fixing the comment for [Translate](https://github.com/microsoft/msbuild/blob/master/src/Build/Instance/ProjectInstance.cs#L1859)?

src/Build/Instance/ProjectInstance.cs
```diff
@@ -2597,6 +2633,28 @@ private void CreateTargetsSnapshot(Evaluation.Project.Data data)
             _targets = CreateCloneDictionary(data.Targets);
         }
 
+        /// <summary>
+        /// Create various import snapshots
+        /// </summary>
+        private void CreateImportSnapshot(Evaluation.Project.Data data)
+        {
+            _imports = new List<string>(data.ImportClosure.Count);
+            foreach (var resolvedImport in data.ImportClosure)
```
Forgind: I think you could condense these loops into one.

src/Build/Instance/ProjectInstance.cs
dfederm: It only makes sense as read only. If one we're to add to the collection, it wouldn't actually import anything for example.

src/Build/Instance/ProjectInstance.cs
dfederm: They're used during evaluation, when any of the overloads which call Initialize are used (not the one used when creating a ProjectInstance from a Project)

dfederm: Regarding out-of-dateness, the AsReadOnly is a wrapper around the collection. It's not a snapshot copy of the collection.

src/Build/Instance/ProjectInstance.cs
dfederm: They're different collections with (potentially) different content.

src/Build/Instance/ProjectInstance.cs
Forgind: I understand, but it seems contradictory to say there's no reason to modify it (so it should be read only) but also modify it in RecordImport and RecordImportWithDuplicates.

If you have a way of importing more files after it's created, you might want to keep the list updated.

src/Build/Instance/ProjectInstance.cs
Forgind: One way:
Make Import a set instead of a list. It's not supposed to have duplicates anyway; this enforces that.
Alternative:
If the duplicates list is sorted, check if the next element is the same as the last one and don't add it to _imports if so.

src/Build/Instance/ProjectInstance.cs
dfederm: It's read-only in the *public* interface. ie a consumer of MSBuild itself (the whole point of this change).

src/Build/Instance/ProjectInstance.cs
dfederm: This is following the implementation in Project. Deviating a lot between the two seems like it'll cause more problems than solve any.

The List vs Set thing is enforced by the Evaluator. This is merely being reported to a public-facing API.

src/Build/Instance/ProjectInstance.cs
AndyGerlicher: `_imports` and `_importsIncludingDuplicates` are `List<string>`. Since the List type implements that interface there's no time/space concern calling `.AsReadOnly()`. And as David said giving external API consumers the a mutable `List` here would be the wrong choice.

src/Build/Instance/ProjectInstance.cs
```diff
@@ -1893,6 +1919,12 @@ private void TranslateAllState(ITranslator translator)
             translator.Translate(ref _isImmutable);
             translator.Translate(ref _evaluationId);
 
+            translator.Translate(ref _imports);
+            Imports = _imports.AsReadOnly();
```
cdmihai: The assignments to `Imports` and `ImportsIncludingDuplicates` should only be made when `translator.Mode == TranslationDirectory.ReadFromStream`. It would avoid creating unneeded new list wrapper instances for the import properties when the same ProjectInstance is serialized multiple times (which might screw up user code if they put the readonly lists in dictionaries and such).

dfederm: Nice catch

dfederm: You know what, I see instance of stuff like this:
```
            // ignore EvaluatedItemElements. Only used by public API users, not nodes
            // ignore itemsByEvaluatedInclude. Only used by public API users, not nodes
```

So since this is only for the public API, it seems that I don't need to translate this at all.

src/Build/Instance/ProjectInstance.cs
cdmihai: I think you want it a List to preserve information about the import order. Single loop does not make sense to me as the number of iterations can differ, and this code is not in a hot path so intense that would warrant such optimizations.

PR #5009, opened 2019-12-27
Various minor cleanup

dfederm:
Various minor cleanup.

Should review commit-by-commit (this will not be squashed on completion).

### DIFF
--- a/src/Build/Definition/ResolvedImport.cs
+++ b/src/Build/Definition/ResolvedImport.cs
@@ -13,12 +13,6 @@ namespace Microsoft.Build.Evaluation
     /// between a ProjectImportElement and the ProjectRootElement of the
     /// imported project.
     /// </summary>
-    /// <comment>
-    /// This struct is functionally identical to KeyValuePair, but is necessary to avoid
-    /// CA908 warnings (types that in ngen images that will JIT).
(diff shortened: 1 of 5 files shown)


src/Build/Definition/ResolvedImport.cs
```diff
@@ -13,12 +13,6 @@ namespace Microsoft.Build.Evaluation
     /// between a ProjectImportElement and the ProjectRootElement of the
     /// imported project.
     /// </summary>
-    /// <comment>
-    /// This struct is functionally identical to KeyValuePair, but is necessary to avoid
-    /// CA908 warnings (types that in ngen images that will JIT).
-    /// It works because although this is a value type, it is not defined in mscorlib.
-    /// Essentially we would use KeyValuePair except for this technical reason.
-    /// </comment>
```
Forgind: Why did you delete this?

dfederm: It's not longer true. `ResolvedImport` is more than just a `KeyValuePair` at this point; in fact it has 5 fields. So if anything it would need to be a value-typed 5-tuple, but that would be a breaking change and not to mention pretty terrible to have to deal with, so overall this comment just doesn't make sense.

PR #5013, opened 2019-12-31
Pass MSBuild Task Id via project's ParentProjectBuildEventContext

KirillOsenkov:
This is based on @MarkKharitonov's fix https://github.com/microsoft/msbuild/pull/4142 for https://github.com/Microsoft/MSBuild/issues/3939. Same change but rebases on latest and adds a unit-test.

@rainersigwald I've played with Mark's change and thought about it, and I think the best thing to do is to take Mark's change as is. I agree that it doesn't capture scenarios where a project's build is spawned via IBuildEngine* APIs and not via the MSBuild task, but we don't care about those as much. Even if we just cover the MSBuild task like this, it will be a huge improvement that covers most if not all scenarios that we really care about.

It's pretty low-risk and non-invasive. I've checked that in this expression:
```
issuingEntry.Request.CurrentMSBuildTask ?? issuingEntry.Request.BuildEventContext
```

all the fields on the build event context are the same, except `BuildRequestId`, `TargetId`and `TaskId`. Since `BuildRequestId` is just `GetHashCode()` I'm not worried about it. `TargetId` and `TaskId` acquire meaningful values instead of being set to -1. I've verified that the valid task id is propagated all the way to the `ParentProjectBuildEventContext` for the project being built. I've added a unit-test to validate that.

I've also made a change in the binlog viewer to take advantage of this if present. It is a major improvement that will significantly clean up the tree and get rid of most low-relevance "project builds", or rather reparent them where they belong and get them out of the way.

image

### DIFF
--- a/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
+++ b/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
@@ -1137,7 +1137,7 @@ namespace Microsoft.Build.BackEnd
                             configurationId: request.Config.ConfigurationId,
                             escapedTargets: request.Targets,
                             hostServices: issuingEntry.Request.HostServices,
-                            parentBuildEventContext: issuingEntry.Request.BuildEventContext,
+                            parentBuildEventContext: issuingEntry.Request.CurrentMSBuildTask ?? issuingEntry.Request.BuildEventContext,
                             parentRequest: issuingEntry.Request,
                             buildRequestDataFlags: buildRequestDataFlags,
                             requestedProjectState: null,
--- a/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
+++ b/src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
@@ -1137,7 +1137,7 @@ namespace Microsoft.Build.BackEnd
                             configurationId: request.Config.ConfigurationId,
                             escapedTargets: request.Targets,
                             hostServices: issuingEntry.Request.HostServices,
-                            parentBuildEventContext: issuingEntry.Request.CurrentMSBuildTask ?? issuingEntry.Request.BuildEventContext,
+                            parentBuildEventContext: issuingEntry.Request.CurrentTaskContext ?? issuingEntry.Request.BuildEventContext,
                             parentRequest: issuingEntry.Request,
                             buildRequestDataFlags: buildRequestDataFlags,
                             requestedProjectState: null,
--- a/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
@@ -421,6 +421,11 @@ namespace Microsoft.Build.BackEnd
                     if (requirements != null)
                     {
                         TaskLoggingContext taskLoggingContext = _targetLoggingContext.LogTaskBatchStarted(_projectFullPath, _targetChildInstance);
+                        if (taskLoggingContext.TaskName == "MSBuild")
+                        {
+                            _buildRequestEntry.Request.CurrentMSBuildTask = taskLoggingContext.BuildEventContext;
+                        }
+
                         try
                         {
                             if (
@@ -462,6 +467,8 @@ namespace Microsoft.Build.BackEnd
                         }
                         finally
                         {
+                            _buildRequestEntry.Request.CurrentMSBuildTask = null;
+
                             // Flag the completion of the task.
                             taskLoggingContext.LogTaskBatchFinished(_projectFullPath, taskResult.ResultCode == WorkUnitResultCode.Success || taskResult.ResultCode == WorkUnitResultCode.Skipped);
 
--- a/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
@@ -421,10 +421,7 @@ namespace Microsoft.Build.BackEnd
                     if (requirements != null)
                     {
                         TaskLoggingContext taskLoggingContext = _targetLoggingContext.LogTaskBatchStarted(_projectFullPath, _targetChildInstance);
-                        if (taskLoggingContext.TaskName == "MSBuild")
-                        {
-                            _buildRequestEntry.Request.CurrentMSBuildTask = taskLoggingContext.BuildEventContext;
-                        }
+                        _buildRequestEntry.Request.CurrentTaskContext = taskLoggingContext.BuildEventContext;
 
                         try
                         {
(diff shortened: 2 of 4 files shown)


MarkKharitonov:
Thank you guys. Can't wait to see it released. If only it could be backported to msbuild 15 ...

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -421,6 +421,11 @@ private async Task<WorkUnitResult> ExecuteBucket(TaskHost taskHost, ItemBucket b
                     if (requirements != null)
                     {
                         TaskLoggingContext taskLoggingContext = _targetLoggingContext.LogTaskBatchStarted(_projectFullPath, _targetChildInstance);
+                        if (taskLoggingContext.TaskName == "MSBuild")
```
cdmihai: Do you have enough info here to check the type rather than the string name?

cdmihai: Actually, why restrict this to the MSBuild task? If the current task id of any task would be saved, the `IBuildEngine` based build requests would also get parented to their spawning tasks, right?

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
```diff
@@ -1137,7 +1137,7 @@ private void IssueBuildRequests(BuildRequestEntry issuingEntry, FullyQualifiedBu
                             configurationId: request.Config.ConfigurationId,
                             escapedTargets: request.Targets,
                             hostServices: issuingEntry.Request.HostServices,
-                            parentBuildEventContext: issuingEntry.Request.BuildEventContext,
+                            parentBuildEventContext: issuingEntry.Request.CurrentMSBuildTask ?? issuingEntry.Request.BuildEventContext,
```
cdmihai: I think `CurrentMSBuildTask` is written by a different thread than the one reading it here. Should it be volatile to flush the value to memory, or am I just too paranoid? `BuildEventContext` should be fine since it gets set very early during the request creation.

src/Tasks.UnitTests/MSBuild_Tests.cs
```diff
@@ -1408,5 +1407,76 @@ public void NonexistentTargetFailsAfterSkipped()
...
+                Assert.True(project.Build(logger));
+
+                var text = logger.Text;
+                var taskId = Regex.Match(text, @"TaskId=(\d+)").Groups[1].Value;
+                var parentTaskId = Regex.Match(text, @"ParentTaskId=(\d+)").Groups[1].Value;
+                Assert.Equal(taskId, parentTaskId);
+            }
+            finally
+            {
+                File.Delete(projectFile1);
```
cdmihai: You can use the fancy [TestEnvironment](https://github.com/microsoft/msbuild/blob/master/src/Shared/UnitTests/TestEnvironment.cs) to avoid manual cleanup.

src/Tasks.UnitTests/MSBuild_Tests.cs
```diff
@@ -1408,5 +1407,76 @@ public void NonexistentTargetFailsAfterSkipped()
...
+
+            try
+            {
+                Project project = new Project(projectFile2);
+                var logger = new MSBuildTaskIdLogger();
+
+                Assert.True(project.Build(logger));
+
+                var text = logger.Text;
+                var taskId = Regex.Match(text, @"TaskId=(\d+)").Groups[1].Value;
```
cdmihai: You can use [MockLogger.ProjectStartedEvents](https://github.com/cdmihai/msbuild/blob/7bcb2f961093e87277439420d351a9f3f3a8cdb8/src/Shared/UnitTests/MockLogger.cs#L87) to avoid string matching

src/Tasks.UnitTests/MSBuild_Tests.cs
KirillOsenkov: I followed the pattern in the file :) We can convert to TestEnvironment later in a separate change. Might as well do it for all other tests then.

src/Tasks.UnitTests/MSBuild_Tests.cs
KirillOsenkov: Good idea, fixed!

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
KirillOsenkov: Yes, it's written by the RequestBuilder thread, and read in BuildRequestEngine (dataflow callback ProcessMessage). However while it's read the RequestBuilder thread is not running, it's awaiting task completion so there's no parallelism.

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
KirillOsenkov: Good idea, fixed. Also renamed CurrentMSBuildTask -> CurrentTaskContext.

KirillOsenkov:
OK second pass ready for review!

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
cdmihai: I was thinking about the case where the value is not yet available in the reading thread's core's cache. AFAIK `volatile` (and at least in Java, other synchronization primitives) flush the writes from the cache, and force the reads from memory (https://igoro.com/archive/volatile-keyword-in-c-memory-model-explained/)

src/Tasks.UnitTests/MSBuild_Tests.cs
```diff
@@ -1408,5 +1406,44 @@ public void NonexistentTargetFailsAfterSkipped()
             // 2nd invocation of t_missing should fail the build resulting in target not found error (MSB4057)
             logger.FullLog.ShouldContain("MSB4057");
         }
+
+        [Fact]
+        public void MSBuildTaskPassesTaskIdToSpawnedBuilds()
```
cdmihai: My last request: add a test for the `IBuildEngine.BuildProjectFile` case :)

cdmihai:
Overall looks good. Slightly worried that task yielding might screw up `BuildRequest.CurrentTaskContext` somehow, but I don't see how. Even if you have multiple yielding tasks at the same time, each one is tied to its own `BuildRequest`.

src/Build/BackEnd/Components/BuildRequestEngine/BuildRequestEngine.cs
KirillOsenkov: this is too deep for me ;)

cdmihai: I looked into it a bit more, and I think it's actually fine. The read is inside a lock block, which refreshes the cache on entry. And the write is inside an async method, which apparently also does it: https://stackoverflow.com/q/6581848/417199

Safest is to make `BuildRequest.CurrentMSBuildTask ` volatile, but I doubt anybody will remove the lock or the async.

src/Tasks.UnitTests/MSBuild_Tests.cs
KirillOsenkov: Done! 🎄

KirillOsenkov:
The Exec_Tests.Timeout fails with:
`Assert.Equal() Failure\nExpected: 143\nActual: 259`

https://github.com/microsoft/msbuild/blob/7bcb2f961093e87277439420d351a9f3f3a8cdb8/src/Tasks.UnitTests/Exec_Tests.cs#L104-L107

MarkKharitonov:
Thank you, guys. In what version of msbuild are we expected to get it?

rainersigwald:
@MarkKharitonov 16.6

PR #5020, opened 2020-01-01
Log global properties in cache miss error

cdmihai:
This reduces the time needed to debug `/isolate` errors. The errors would look like this:

image


src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1870,6 +1870,11 @@ BuildEventContext NewBuildEventContext()
                 var parentConfiguration = configCache[parentRequest.BuildRequest.ConfigurationId];
                 return (buildRequestConfiguration, parentConfiguration);
             }
+
+            string ConcatenateGlobalProperties(BuildRequestConfiguration configuration)
+            {
+                return string.Join(";", configuration.GlobalProperties.GetCopyOnReadEnumerable().Select(p => $"{p.Name}={p.EvaluatedValue}"));
```
Forgind: This makes a copy of the properties, right? Why is that necessary; could you condense it to just selecting name and its value in the dictionary?

Also, nit: add a space after the semicolon. Makes it look cleaner.

PR #5021, opened 2020-01-01
RAR: Prefer the .dll extension first when resolving assemblies.

KirillOsenkov:
RAR: Reorders the AllowedReferenceAssemblyFileExtensions to place .dll first and .winmd second. Statistically the vast majority of references are to .dll files.

We avoid a large amount of File.Exists checks if we try the .dll first.

Even on a small simple build (30 projects) the difference is noticeable:

.winmd first:
========

.binlog size: 14,685 KB
"Considered .winmd" messages: 2217
"Considered but it didn't exist": 6151

.dll first:
=====

.binlog size: 14,221 KB
"Considered .winmd" messages: 700
"Considered but it didn't exist": 4634

Reduction in the number of log entries will also indirectly translate in a slight reduction of build times, slight reduction of file system access operations (File.Exists checks). It also declutters the logs.

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -578,8 +578,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         Put the most likely extensions first for reference resolution speed.
         -->
     <AllowedReferenceAssemblyFileExtensions Condition=" '$(AllowedReferenceAssemblyFileExtensions)' == '' ">
-      .winmd;
       .dll;
+      .winmd;
       .exe
     </AllowedReferenceAssemblyFileExtensions>
 


KirillOsenkov:
Fixes https://github.com/microsoft/msbuild/issues/5018

KirillOsenkov:
Note: for most builds, just setting this to `.dll` would be enough. Numbers for just the .dll:

.winmd first:
.binlog size: 14,685 KB
"Considered .winmd" messages: 2217
"Considered but it didn't exist": 6151

.dll first:
.binlog size: 14,221 KB
"Considered .winmd" messages: 700
"Considered but it didn't exist": 4634

Just the .dll:
.binlog size: 13,943 KB
"Considered .winmd" messages: 0
"Considered but it didn't exist": 3234

This is almost halving File.Exists checks during RAR and reducing the log size by 2%. Of course we can't do that in common targets, but this is something to recommend to folks tuning their builds.

Therzok:
Should exe also go before winmd?

KirillOsenkov:
I’d argue that people reference .exes just as rarely as .winmd so there’s no discernible difference there.

cdmihai:
I wonder if the ordering actually represents some app model loading semantics that nobody knows about anymore, and thus breaking it if changing the order. But, there's only one way to find out :D

KirillOsenkov:
We're just changing RAR, what could go wrong?

KirillOsenkov:
image

cdmihai:
>There's no detail about how critical the reordering is relative to the other work in that changeset.

Sounds like there would be benefit in releasing it and reverting if it breaks something: without a proper comment saying why the order is important, future people will spend time thinking why they can't reorder, or duplicating this very PR. With a concrete failure at hand, at least we can properly document it and spare the future some pain :)

Forgind:
😔 
I would be much more comfortable doing something like that if we had a nice way of releasing it to only a small number of customers, so that if it turns out to be a bigger problem than anticipated, fewer people will be mad.

cdmihai:
> It will break things.

Cool, then we can replace the PR changes with a comment explaining why order matters.

davkean:
@cdmihai What fun would that be?

rainersigwald:
Added some reasons it's not easy in https://github.com/microsoft/msbuild/issues/5018#issuecomment-590441271

PR #5022, opened 2020-01-01
Avoid a first-chance exception in DeleteNoThrow.

KirillOsenkov:
It makes debugging unit-tests with first-chance exceptions a bit better.

### DIFF
--- a/src/Shared/FileUtilities.cs
+++ b/src/Shared/FileUtilities.cs
@@ -754,7 +754,11 @@ namespace Microsoft.Build.Shared
         {
             try
             {
-                File.Delete(FixFilePath(path));
+                path = FixFilePath(path);
+                if (File.Exists(path))
+                {
+                    File.Delete(path);
+                }
             }
             catch (Exception ex) when (ExceptionHandling.IsIoRelatedException(ex))
             {


KirillOsenkov:
Are you worried about perf? Hopefully the file system makes this negligible. I also hope to offset this by removing thousands of File.Exists checks for .winmd files in RAR with my other PR ;)

Functionally, the highly unlikely case when the file system changes in between the two calls can fall into two buckets:
1. file didn't exist but it starts existing - we don't enter the condition and the file survives - but arguably if the condition wasn't in place, there's high chance of us trying to delete the file before it appears, so we're not losing anything in correctness here
2. file existed but it disappears before the second call - then we just get an exception like we had before, so this is functionally equivalent to the old code.

Realistically and practically though, I reduce the chance of a first-chance exception from 100% to 0.001% and I call it a good thing.

Therzok:
Other than in test code, DeleteNoThrow is used in 2 places:

https://github.com/microsoft/msbuild/blob/037db14d698d4bf0322c2bc69ea00b4c08f1e078/src/Tasks/Copy.cs#L310

https://github.com/microsoft/msbuild/blob/87255d576bcd260dceb8cc52a369564529ac98e7/src/Build/BackEnd/Shared/BuildRequestConfiguration.cs#L906

The latter does a FileExists check itself. Maybe the caller code that triggers the first chance exception should be updated to do that check?

KirillOsenkov:
That's a good idea, also less invasive.

KirillOsenkov:
Hmm, I forgot which test was problematic, and it's used in a lot of places.

Our options:
1. Just close the PR and have a more targeted fix next time
2. Add an optional parameter whether to check for file existence and default to false to match old behavior. Turn it on in places that don't want to do their own null check. This would force clarify the contract at the callsites. This is more work.

I'm now leaning towards option 1 as I certainly don't want to make the Copy operation more expensive...

KirillOsenkov:
Yes I'm just going to close this. Even if we want to fix something, it should go elsewhere.

PR #5026, opened 2020-01-02
Log execution context like the current process, command line, working directory, etc

KirillOsenkov:
It can be useful to investigate issues.

I spent some time finding a good place to emit these messages for all loggers, but it's unclear where would be a good place. You want to log it exactly once per process, you want to log after the loggingService has been initialized, and you want to log regardless of Restore, Build or Graph Build.

BinaryLogger initialization seems like a good enough spot.

Fixes https://github.com/microsoft/msbuild/issues/2810

This is how it would look like in the viewer:

image

And this is how it would look like when the binlog is played back:

image


cdmihai:
@danmosemsft is also looking into logging rsp sources in #5007. What if, instead of implementing the logic only in the binary logger, you add a new `BuildManager.StartBuild(IEnumerable<string> initialMessages)` overload. `StartBuild` is the one creating the `LoggingService`.  The command line app would call `StartBuild` with a list of messages containing your stuff, and @danmosemsft's stuff

cdmihai:
Nice to see TestBinaryLoggerRoundtrip catching the binary format change: https://dev.azure.com/dnceng/public/_build/results?buildId=470636&view=ms.vss-test-web.build-test-results-tab&runId=14975240&resultId=102647&paneView=debug

KirillOsenkov:
Hmm, I'm not sure I want to mess with BuildManager at this level... let's see what others say.

For now regardless of what we decide to do, I've updated the test to reflect that the binlog will contain more information than the text log.

rainersigwald:
My instinct was also to provide a bit more structure to this, like Mihai proposed. But this gets us an awful lot of the way there at pretty low cost, so I think I'm on board with it.

KirillOsenkov:
That was my thinking. Cheap, non-invasive, safe, gets the job done and open for future improvements.

Forgind:
By StartBuild, do you mean BeginBuild? Either I'm missing something, or StartBuild doesn't exist outside comments in BuildManager.cs.

I'd be happier with it in BuildManager.BeginBuild, partially because that puts it right next to `MSBuildEventSource.Log.BuildStart`—putting logging in one place. You can also add the same information to my BuildStart/Stop events, too, if you'd like. From a functional perspective, though, this works well enough.

cdmihai:
Yeah, `BeginBuild`

KirillOsenkov:
To be clear, I'm down with whatever you folks prefer, but I feel like a different change should be done by a team member (because I don't know what I'm doing). I wouldn't be comfortable messing with the heart of BuildManager myself. Maintainers have the rights to push to my fork, be my guest ;)

Or, we just merge this as-is and move on to more pressing challenges ;)

cdmihai:
>the heart of BuildManager myself

It's more like a mouth than a heart :)

I'll see if I have some spare cycles to prototype it today.

src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -944,6 +944,68 @@ public void SimpleBuildWithFailure()
             _logger.AssertLogContains("[errormessage]");
         }
 
+        [Fact]
+        public void DeferredMessageShouldGetLogged()
```
Forgind: nit:
```suggestion
        public void DeferredMessageShouldBeLogged()
```
(Slightly more formal, and it matches the shouldly asserts better)

src/Build.UnitTests/BinaryLogger_Tests.cs
```diff
@@ -44,20 +44,21 @@ public void TestBinaryLoggerRoundtrip()
 
             binaryLogger.Parameters = _logFile;
 
-            var mockLogger1 = new MockLogger();
+            var mockLogFromBuild = new MockLogger();
```
Forgind: 👍

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -358,6 +360,37 @@ public static BuildManager DefaultBuildManager
...
+        /// <summary>
+        /// Prepares the BuildManager to receive build requests.
+        /// </summary>
+        /// <param name="parameters">The build parameters.  May be null.</param>
+        /// <param name="deferredBuildMessages"> Build messages to be logged before the build begins. </param>
+        /// <exception cref="InvalidOperationException">Thrown if a build is already in progress.</exception>
+        public void BeginBuild(BuildParameters parameters, IEnumerable<DeferredBuildMessage> deferredBuildMessages)
+        {
+            // deferredBuildMessages cannot be an optional parameter on a single BeginBuild method because it would break binary compatibility.
+
```
Forgind: nit:
delete space

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -2301,6 +2336,19 @@ private ILoggingService CreateLoggingService(IEnumerable<ILogger> loggers, IEnum
...
         }
 
+        private static void LogDeferredMessages(ILoggingService loggingService, IEnumerable<DeferredBuildMessage> deferredBuildMessages)
+        {
+            if (deferredBuildMessages == null)
+            {
+                return;
+            }
+
+            foreach (var message in deferredBuildMessages)
```
Forgind: nit:
type

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -2301,6 +2336,19 @@ private ILoggingService CreateLoggingService(IEnumerable<ILogger> loggers, IEnum
             return loggingService;
         }
 
+        private static void LogDeferredMessages(ILoggingService loggingService, IEnumerable<DeferredBuildMessage> deferredBuildMessages)
```
Forgind: It would be simpler to inline this function.

src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -2301,6 +2336,19 @@ private ILoggingService CreateLoggingService(IEnumerable<ILogger> loggers, IEnum
...
+        {
+            if (deferredBuildMessages == null)
```
Forgind: if (deferredBuildMessages != null)
{
    // foreach below
}

src/Build/Logging/BinaryLogger/BinaryLogger.cs
```diff
@@ -139,9 +142,23 @@ public void Initialize(IEventSource eventSource)
 
             binaryWriter.Write(FileFormatVersion);
 
+            LogInitialInfo();
+
             eventSource.AnyEventRaised += EventSource_AnyEventRaised;
         }
 
+        private void LogInitialInfo()
```
Forgind: Inline this?

src/MSBuild/Resources/Strings.resx
```diff
@@ -1225,6 +1225,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: {0} contains the exception message explaining why the
       logger could not be created -- this message comes from the CLR/FX and is localized.
     </comment>
+  </data>
+  <data name="Process" UESanitized="true" Visibility="Public">
```
Forgind: We don't need UESanitized

src/Shared/Traits.cs
```diff
@@ -89,6 +89,11 @@ private static int ParseIntFromEnvironmentVariableOrDefault(string environmentVa
 
     internal class EscapeHatches
     {
+        /// <summary>
+        /// Do not log command line information to build loggers. Useful to unbreak people who parse the msbuild log and who are unwilling to change their code.
+        /// </summary>
+        public readonly bool DoNotSendDeferredMessagesToBuildManager = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MsBuildDoNotSendDeferredMessagesToBuildManager"));
```
Forgind:
```suggestion
        public readonly bool DoNotSendDeferredMessagesToBuildManager = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBuildDoNotSendDeferredMessagesToBuildManager"));
```

src/Shared/Traits.cs
```diff
@@ -89,6 +89,11 @@ private static int ParseIntFromEnvironmentVariableOrDefault(string environmentVa
 
     internal class EscapeHatches
     {
+        /// <summary>
+        /// Do not log command line information to build loggers. Useful to unbreak people who parse the msbuild log and who are unwilling to change their code.
```
Forgind:
```suggestion
        /// Do not log command line information to build loggers. Useful for unbreaking people who parse the MSBuild log but don't want to change their code.
```

cdmihai:
@KirillOsenkov We decided to move the logging in BuildManager so even API based builds log this information (which should be nice for logs comming from within devenv & friends). I'll create a separate PR for deferred build message, which should enable @danmosemsft's scenario.

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: Style preference. I prefer less nested code than more nested code.

src/Build/Logging/BinaryLogger/BinaryLogger.cs
cdmihai: I prefer the compile to do that, and keep the functions to improve readability.

rainersigwald:
Since we slipped this to 16.6 and already took #5026, I reverted some of the more recent changes that avoided that dependency and merged master in my latest push.

cdmihai:
Wouldn't it be better to actually keep the logging of these things in the BuildManager.BeginBuild, to allow devenv based builds to also include them? Though honestly I don't know how useful they would be there.

@davkean, would any of the extra logged stuff in this PR (current process, command line, working directory, etc) be useful when investigating msbuild logs under devenv?

PR #5028, opened 2020-01-03
Fix nullref in binlog playback when CentralLogger is null

KirillOsenkov:
When playing back a .binlog with using -distributedfilelogger there was a crash because CentralLogger was null.

Fixes https://github.com/microsoft/msbuild/issues/4788

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -3519,13 +3519,14 @@ namespace Microsoft.Build.CommandLine
 
             foreach (var distributedLoggerRecord in distributedLoggerRecords)
             {
-                if (distributedLoggerRecord.CentralLogger is INodeLogger nodeLogger)
+                var centralLogger = distributedLoggerRecord.CentralLogger;
+                if (centralLogger is INodeLogger nodeLogger)
                 {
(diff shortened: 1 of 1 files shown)


src/MSBuild/XMake.cs
```diff
@@ -3558,7 +3559,11 @@ private static void ReplayBinaryLog
 
             foreach (var distributedLoggerRecord in distributedLoggerRecords)
             {
-                distributedLoggerRecord.CentralLogger.Shutdown();
```
rainersigwald: Just do

```c#
distributedLoggerRecord.CentralLogger?.Shutdown();
```

?

src/MSBuild/XMake.cs
```diff
@@ -3519,13 +3519,14 @@ private static void ReplayBinaryLog
 
             foreach (var distributedLoggerRecord in distributedLoggerRecords)
             {
-                if (distributedLoggerRecord.CentralLogger is INodeLogger nodeLogger)
+                var centralLogger = distributedLoggerRecord.CentralLogger;
```
Forgind:
```suggestion
                ILogger centralLogger = distributedLoggerRecord.CentralLogger;
```

PR #5029, opened 2020-01-03
Normalize the path before opening an app.config file.

KirillOsenkov:
Fixes https://github.com/microsoft/msbuild/issues/4335

### DIFF
--- a/src/Tasks/AppConfig/AppConfig.cs
+++ b/src/Tasks/AppConfig/AppConfig.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Build.Tasks
 
                 // it's important to normalize the path as it may contain two slashes
                 // see https://github.com/microsoft/msbuild/issues/4335 for details.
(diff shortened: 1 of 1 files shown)


src/Tasks/AppConfig/AppConfig.cs
```diff
@@ -23,6 +24,11 @@ internal void Load(string appConfigFile)
             try
             {
                 var readerSettings = new XmlReaderSettings { DtdProcessing = DtdProcessing.Ignore };
+
+                // it's important to normalize the path as it may contain two slashes
+                // see https://github.com/microsoft/msbuild/issues/4335 for details.
+                appConfigFile = Path.GetFullPath(appConfigFile);
```
rainersigwald:
```suggestion
                appConfigFile = FileUtilities.NormalizePath(appConfigFile);
```

? That's the one we generally use everywhere.

PR #5032, opened 2020-01-06
Implement Command-line option for displaying targets

### LINKED ISSUE #33: Command-line option for displaying targets
Most build systems that allow the creation of custom targets also have a command-line option for displaying the available targets in a project:

gradle tasks
rake --tasks
grunt --help

Developers working on projects that use MSBuild would benefit from a similar feature.

szaliszali:
This feature request was lying around for a few years, I'm submitting a simple implementation for review and further discussion. As @AndyGerlicher kindly pointed out the already existing `preprocess` option, it seemed logical to take the same approach for this one.

Sample output for a C# project:
```
Microsoft (R) Build Engine version 16.5.0-dev-20056-01+fb441875f for .NET Framework

CreateManifestResourceNames
ResolveCodeAnalysisRuleSet
XamlPreCompile
ShimReferencePathsWhenCommonTargetsDoesNotUnderstandReferenceAssemblies
_BeforeVBCSCoreCompile
[.... 200+ lines omitted ....]
XsdResolveReferencePath
CleanXsdCodeGen
_SetTargetFrameworkMonikerAttribute
EnsureNuGetPackageBuildImports
```

Fixes #33


src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -731,6 +731,31 @@ public void PreprocessSwitchIdentificationTests()
             Assert.True(unquoteParameters);
         }
 
+        [Fact]
+        public void TargetsSwitchIdentificationTests()
+        {
+            CommandLineSwitches.ParameterizedSwitch parameterizedSwitch;
+            string duplicateSwitchErrorMessage;
```
Forgind: nit:
You can inline these variable declarations such as `out string duplicateSwitchErrorMessage` below

src/MSBuild/Resources/Strings.resx
```diff
@@ -1154,6 +1170,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <value>MSBUILD : error MSB1047: File to preprocess to is not valid. {0}</value>
     <comment>{StrBegin="MSBUILD : error MSB1047: "}</comment>
   </data>
+  <data name="InvalidTargetsPath">
+    <value>MSBUILD : error MSB1047: File to print targets to is not valid. {0}</value>
```
Forgind: I know you copied the style of the error for preprocessing, but I don't think that message is perfect. Would you mind changing to something simple like:
`Targets could not be printed.`
Also, you reused MSB1047. Can you change the number?

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -731,6 +731,31 @@ public void PreprocessSwitchIdentificationTests()
             Assert.True(unquoteParameters);
         }
 
+        [Fact]
```
Forgind: Perhaps add a test for failure?

src/MSBuild/XMake.cs
```diff
@@ -1056,6 +1060,7 @@ string outputResultsCache
                 ToolsetDefinitionLocations toolsetDefinitionLocations = ToolsetDefinitionLocations.Default;
 
                 bool preprocessOnly = preprocessWriter != null && !FileUtilities.IsSolutionFilename(projectFile);
+                bool targetsOnly = targetsWriter != null && !FileUtilities.IsSolutionFilename(projectFile);
```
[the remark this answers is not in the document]
szaliszali: This does not seem logical for me at first glance, because printing the targets should be a read-only operation.

Also, it looks like the setting will be ignored anyway (see [this comment](https://github.com/Microsoft/msbuild/blob/08d1cb5532d98dd4b9f2f4ecdb7c9c7a037aed1c/src/Build/Xml/XmlReaderExtension.cs#L83-L85)).

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
```diff
@@ -731,6 +731,31 @@ public void PreprocessSwitchIdentificationTests()
...
+        public void TargetsSwitchIdentificationTests()
+        {
+            CommandLineSwitches.ParameterizedSwitch parameterizedSwitch;
+            string duplicateSwitchErrorMessage;
+            bool multipleParametersAllowed;
+            string missingParametersErrorMessage;
+            bool unquoteParameters;
+            bool emptyParametersAllowed;
+
+            Assert.True(CommandLineSwitches.IsParameterizedSwitch("targets", out parameterizedSwitch, out duplicateSwitchErrorMessage, out multipleParametersAllowed, out missingParametersErrorMessage, out unquoteParameters, out emptyParametersAllowed));
```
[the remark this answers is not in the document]
szaliszali: committed

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
szaliszali: committed

src/MSBuild/Resources/Strings.resx
szaliszali: committed

src/MSBuild.UnitTests/CommandLineSwitches_Tests.cs
szaliszali: added one for multiple occurrences, switch parser does not validate filenames, it accepts anything (like `"\0"`, for example)

src/MSBuild/XMake.cs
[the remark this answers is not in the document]
Forgind: I thought I was suggesting it should be read-only, but I might have gotten it flipped. @rainersigwald, do you know better? Otherwise, this looks good to me.

cdmihai:
>_BeforeVBCSCoreCompile
[.... 200+ lines omitted ....]
XsdResolveReferencePath

I doubt this would actually be practically useful to anyone.  Since MSBuild does not have any information hiding capabilities (you get ALL the hundreds of meaningless targets, not just the entry graph wide operations supported by the current mix of "SDKs"), and since MSBuild does not differentiate the graph wide entry targets (Build, Clean, Rebuild, Test, Restore, Pack) in any way from the implementation targets, this feature will just lead to information overload. It would interesting though, if we introduced graph wide entry targets as first class entities in the language.

I'd be quite curious to learn of a practical usage to this.

Forgind:
One possible use I thought of was if you know roughly what you need but don't know exactly what the target is called. This would make it easier to find the exact target. You can also try to add custom targets, and if they don't show up, you know you didn't add them properly.

The original issue has 30 upvotes and has gotten some activity within the past 6 months, so although it's an old issue, I think it's worth taking this pr.

AndyGerlicher:
One thing I worry about on this feature is it might give the implication of a target contract. Looking at the other commands provided as examples, `grunt --help` or `rake --tasks` to me seem to imply those are operational things that will occur or are available to be used. In MSBuild, many of the targets are internal implementations and should really be considered "private" and not exposed like this. The language really isn't setup as a discoverable catalog of functions. Maybe one mitigation would be to filter out all targets starting with `_`? It's largely convention based, but it would help some with that.

The other is this is pretty trivial to get as a one-off. It seems surprising to me that this would be common enough to go into MSBuild proper. Again unless it's more of a let's see what's available to hook into which is my original concern.

src/Build/Definition/Project.cs
```diff
@@ -1417,6 +1417,11 @@ public void SaveLogicalProject(TextWriter writer)
             implementation.SaveLogicalProject(writer);
         }
 
+        public void PrintTargets(TextWriter writer)
```
cdmihai: I'd rather have the Project API only return targets, and not deal with the extra problem domain of printing object model things to streams.

szaliszali:
I'm happy to see that the PR sparked a bit of conversation [also, a bit less happy about the merge conflict]. I'll update as soon as I have a bit of free time.

About usability of the output: I also think that dumping 200 targets is not the most useful thing we could do. For now, I can consider this feature more like a debugging tool than quick help for the confused end-user. Also, would not recommend skipping targets based on naming convention.

szaliszali:
Short explanation for the additional error handler: during testing I experienced several error modes:
- missing project file: prints error and halts before doing any work
image
- fatal error during project file loading (e.g. a random text file as project): finds project file and starts loading, then prints error (in red) and exits
image
- valid project with something broken during processing build files (e.g. SDK resolver is missing from msbuild directory): in this case, I could not figure out why the error message is printed in one case and not in the other
    - error message (in red) was printed
    image
    - no error message was printed and this could confuse users what is (not) happening, this is why I decided to add the extra error handler
    image

I chose this simple way of displaying the message instead of real loggers because I found the logger setup and logging too complicated for this simple task

src/MSBuild/XMake.cs
```diff
@@ -1292,6 +1303,25 @@ string outputResultsCache
...
+                Project project = projectCollection.LoadProject(projectFile, globalProperties, toolsVersion);
+
+                targetsWriter.WriteLine(string.Join(Environment.NewLine, project.Targets.Keys));
+
+                projectCollection.UnloadProject(project);
+                return true;
+            }
+            catch (Exception ex)
+            {
+                var message = ResourceUtilities.FormatResourceStringStripCodeAndKeyword("TargetsCouldNotBePrinted", ex.Message);
```
rainersigwald:
```suggestion
                var message = ResourceUtilities.FormatResourceStringStripCodeAndKeyword("TargetsCouldNotBePrinted", ex);
```

This preserves the call stack which can make fixing bugs that report this error MUCH easier.

src/MSBuild/XMake.cs
```diff
@@ -1292,6 +1303,25 @@ string outputResultsCache
             return success;
         }
 
+        private static bool PrintTargets(string projectFile, string toolsVersion, Dictionary<string, string> globalProperties, TextWriter targetsWriter, ProjectCollection projectCollection)
+        {
+            try
+            {
+                Project project = projectCollection.LoadProject(projectFile, globalProperties, toolsVersion);
+
+                targetsWriter.WriteLine(string.Join(Environment.NewLine, project.Targets.Keys));
```
rainersigwald: Can you call this in a loop?
```suggestion
                foreach (string target in project.Targets.Keys)
                {
                    targetsWriter.WriteLine(target);
                }
```

That way we don't have to allocate a big string? (it's not super critical here, but it's good practice)

rainersigwald:
Thanks @szaliszali! This should go out with Visual Studio 16.6 preview 1 (at some point in the future).

videoguy:
Is it available in VS 2019?
I tried, it doesn't seem to be supported.

Forgind:
It should be. What exact version of VS 2019 are you using? 16.11?

videoguy:
VS 2019 16.9.25 build

videoguy:
After the update, it is working. I am wondering if there is something for solution file (.sln) as well.

rainersigwald:
@videoguy not today; that's tracked by dotnet/msbuild#7697.

PR #5034, opened 2020-01-07
Add CPS F# project type GUID

saul:
Currently F# SDK-style projects are treated at `SolutionProjectType.Unknown`. This is PR fixes this and treats them as `KnownToBeMSBuildFormat`.


rainersigwald:
@cartermp can you confirm this is the right project-type GUID? Any others that we should know about while we're in here?

cartermp:
Not sure about others but the [whole file](https://github.com/dotnet/project-system/blob/138f25249f2170a9cf3c3df1b6cea419f4dc469b/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/ProjectType.cs) might have some that are interesting for MSBuild.

PR #5043, opened 2020-01-09
Detect and switch to UTF-8 codepage in ToolTask with UseCommandProcessor

### LINKED ISSUE #4904: ToolTask UseCommandProcessor should handle Unicode commands
Originally filed as [AB#1012708](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1012708) by @MrTrillian.

When a `ToolTask`-derived task (like VC++'s `CustomBuild`) opts into running using `cmd.exe`, the batch file generated by `ToolTask`

https://github.com/microsoft/msbuild/blob/b6e7d1e07e28ac7f8e70ebee110dcfe795c30adf/src/Utilities/ToolTask.cs#L1368

may be in an encoding incapable of representing the full string.

`ToolTask` should do something like #143. But make sure to use the _current_ version of that special-casing, since it has had a number of bug fixes over the years.

rainersigwald:
Fix #4904 using the same detect-and-switch-codepage code as we use in `Exec`.

### DIFF
--- a/src/Shared/EncodingUtilities.cs
+++ b/src/Shared/EncodingUtilities.cs
@@ -148,5 +148,38 @@ namespace Microsoft.Build.Shared
                 return StartsWithPreamble(stream);
             }
         }
+
+        /// <summary>
+        /// Checks to see if a string can be encoded in a specified code page.
+        /// </summary>
+        /// <remarks>Internal for testing purposes.</remarks>
+        /// <param name="codePage">Code page for encoding.</param>
+        /// <param name="stringToEncode">String to encode.</param>
+        /// <returns>True if the string can be encoded in the specified code page.</returns>
+        internal static bool CanEncodeString(int codePage, string stringToEncode)
+        {
+            // We have a System.String that contains some characters. Get a lossless representation
+            // in byte-array form.
+            var unicodeEncoding = new UnicodeEncoding();
+            var unicodeBytes = unicodeEncoding.GetBytes(stringToEncode);
+
+            // Create an Encoding using the desired code page, but throws if there's a
+            // character that can't be represented.
+            var systemEncoding = Encoding.GetEncoding(codePage, EncoderFallback.ExceptionFallback,
+                DecoderFallback.ExceptionFallback);
+
+            try
+            {
+                var oemBytes = Encoding.Convert(unicodeEncoding, systemEncoding, unicodeBytes);
+
+                // If Convert didn't throw, we can represent everything in the desired encoding.
+                return true;
+            }
+            catch (EncoderFallbackException)
+            {
+                // If a fallback encoding was attempted, we need to go to Unicode.
+                return false;
+            }
+        }
     }
 }
--- a/src/Shared/EncodingUtilities.cs
+++ b/src/Shared/EncodingUtilities.cs
@@ -14,6 +14,8 @@ namespace Microsoft.Build.Shared
     /// </summary>
     internal static class EncodingUtilities
     {
+        internal static readonly Encoding Utf8WithoutBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+
         private static Encoding s_currentOemEncoding;
 
         internal const string UseUtf8Always = "ALWAYS";
--- a/src/Shared/EncodingUtilities.cs
+++ b/src/Shared/EncodingUtilities.cs
@@ -16,6 +16,11 @@ namespace Microsoft.Build.Shared
     {
         private static Encoding s_currentOemEncoding;
 
+        internal const string UseUtf8Always = "ALWAYS";
+        internal const string UseUtf8Never = "NEVER";
+        internal const string UseUtf8Detect = "DETECT";
+        internal const string UseUtf8System = "SYSTEM";
+
         /// <summary>
         /// Get the current system locale code page, OEM version. OEM code pages are used for console-based input/output
         /// for historical reasons.
--- a/src/Shared/EncodingUtilities.cs
+++ b/src/Shared/EncodingUtilities.cs
@@ -188,5 +188,59 @@ namespace Microsoft.Build.Shared
                 return false;
             }
         }
+
+        /// <summary>
+        /// Find the encoding for the batch file.
+        /// </summary>
+        /// <remarks>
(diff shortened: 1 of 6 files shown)


src/Utilities/ToolTask.cs
```diff
@@ -1365,7 +1365,24 @@ public override bool Execute()
...
+                        {
+                            encoding = EncodingUtilities.CurrentSystemOemEncoding;
+                        }
+                        else
+                        {
+                            encoding = EncodingUtilities.BatchFileEncoding(commandLineCommands + _temporaryBatchFile, EncodingUtilities.UseUtf8Detect);
+
+                            if (encoding.CodePage != EncodingUtilities.CurrentSystemOemEncoding.CodePage)
+                            {
+                                File.AppendAllText(_temporaryBatchFile, $@"%SystemRoot%\System32\chcp.com {encoding.CodePage}>nul{Environment.NewLine}"); // first line is ANSI
```
MrTrillian: First line is decoded with the console code page which will default to the OEM code page (may be distinct from ANSI, though all chars here should be ASCII).

src/Shared/EncodingUtilities.cs
```diff
@@ -148,5 +155,92 @@ internal static bool FileStartsWithPreamble(string file)
...
+            var unicodeBytes = unicodeEncoding.GetBytes(stringToEncode);
+
+            // Create an Encoding using the desired code page, but throws if there's a
+            // character that can't be represented.
+            var systemEncoding = Encoding.GetEncoding(codePage, EncoderFallback.ExceptionFallback,
+                DecoderFallback.ExceptionFallback);
+
+            try
+            {
+                var oemBytes = Encoding.Convert(unicodeEncoding, systemEncoding, unicodeBytes);
```
MrTrillian: Curious, is this different from/better than calling `systemEncoding.GetBytes(stringToEncode)`?

Oh, this was moved from elsewhere.

src/Shared/EncodingUtilities.cs
```diff
@@ -14,8 +14,15 @@ namespace Microsoft.Build.Shared
...
     internal static class EncodingUtilities
     {
+        internal static readonly Encoding Utf8WithoutBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+
         private static Encoding s_currentOemEncoding;
 
+        internal const string UseUtf8Always = "ALWAYS";
+        internal const string UseUtf8Never = "NEVER";
+        internal const string UseUtf8Detect = "DETECT";
+        internal const string UseUtf8System = "SYSTEM";
```
MrTrillian: Those would be nicer as an enum.

src/Tasks/Exec.cs
```diff
@@ -203,7 +198,7 @@ public ITaskItem[] Outputs
         /// </summary>
         private void CreateTemporaryBatchFile()
         {
-            var encoding = BatchFileEncoding();
+            var encoding = EncodingUtilities.BatchFileEncoding(Command + WorkingDirectory, UseUtf8Encoding);
 
             // Temporary file with the extension .Exec.bat
             _batchFile = FileUtilities.GetTemporaryFile(".exec.cmd");
```
MrTrillian: Unrelated to this PR but FYI: From what I learned from the console team, the code slightly below those lines isn't how things should work "in an ideal world" (not sure how this is actually feasible)

```
_standardOutputEncoding = encoding;
_standardErrorEncoding = encoding;
```

The fact that we use chcp to change the console code page and console output code page to some arbitrary code page shouldn't affect the encoding of the stdout written by well-behaved apps (again, according to the console team's ideal world). Because the output is redirected to msbuild and not sent to a console buffer, apps should always write UTF-8 in order to preserve all unicode chars. Apps should only honor the console output code page when writing to a conhost's buffer. This differs from what I previously thought was the preferred behavior.

src/Shared/EncodingUtilities.cs
cdmihai: Mmmmmm, enums 🍰

.vsts-dotnet.yml
```diff
@@ -147,6 +147,13 @@ stages:
         ArtifactName: logs
       condition: succeededOrFailed()
 
+    - task: PublishBuildArtifacts@1
+      displayName: 'Publish Artifact: symbols'
+      inputs:
+        PathtoPublish: 'artifacts\bin'
+        ArtifactName: symbols
+      condition: succeededOrFailed()
```
Forgind: To differentiate from...unfinished? I'm guessing this variable is set externally, but if we have access to it, is it possible to change it to finished?

eng/cibuild_bootstrapped_msbuild.ps1
```diff
@@ -106,6 +106,11 @@ try {
 
   if ($buildStage1)
   {
+    if (Test-Path $Stage1Dir)
```
Forgind: I think you included code from outside your pr. This is an example from Mihai. Would you mind cleaning it up before I look further?

.vsts-dotnet.yml
rainersigwald: This is not part of the PR; GitHub has gotten confused about what commits will actually be merged to 16.5 here.

But the answer to your question is "to distinguish from 'cancelled'", where you might not want to bother uploading partial artifacts, even if you want them from completed-but-failed builds. https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#succeededorfailed

src/Utilities/ToolTask.cs
rainersigwald: Clarified this comment, thanks.

src/Shared/EncodingUtilities.cs
rainersigwald: This is moved code, too. I'm not sure why it wasn't an enum in the first place but I'd rather leave it as-is now.

src/Shared/EncodingUtilities.cs
rainersigwald: Looks like it's pretty much just that: https://github.com/dotnet/runtime/blob/843d66b238817c7bdbe42556bb17ab1e9f10973a/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.cs#L202

PR #5045, opened 2020-01-10
Enable ResolveNonMSBuildProjectOutput on .NET Core

### LINKED ISSUE #4770: Design-time build tasks aren't all available in .NET Core
See https://github.com/dotnet/roslyn/issues/36072#issuecomment-536088468 for a nice repro by @d-dizhevsky.

> I have created repro project here: https://github.com/d-dizhevsky/owss
> 
> It contains:
> 
> * `src/OpenWebSecurityScanner` project which targets `netcoreapp2.1` and runs MSBuildWorkspace
> * `testProjects/WebAPI1` which was created with `dotnet new webapi`
> * `testProjects/TestCoreLib` which was created with `dotnet new classlib`
> 
> The `WebAPI1` has a project reference on `TestCoreLib`.
> Just do `cd src/OpenWebSecurityScanner/` and then `dotnet run`.
> 
> Thats my output on OSx:
> image

----

This is happening because some tasks that are used in design-time builds (and design-time-like builds like Roslyn `MSBuildWorkspace`) aren't ported to .NET Core.

Right know there's only one known failure; I suspect there may be a few more after fixing this one:

- [ ] [`ResolveNonMSBuildProjectOutput`](https://github.com/microsoft/msbuild/blob/49729650232e1e7c1f7a1016c83c5d6f254cfa04/src/Tasks/ResolveNonMSBuildProjectOutput.cs).

rainersigwald:
This task isn't required in normal operation on .NET Core, but
compiles and runs just fine so there's no reason to have a discrepancy.

Fixes #4770.

### DIFF
--- a/src/Tasks.UnitTests/ResolveNonMSBuildProjectOutput_Tests.cs
+++ b/src/Tasks.UnitTests/ResolveNonMSBuildProjectOutput_Tests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.IO;
 using System.Text;
 using System.Linq;
 using System.Xml;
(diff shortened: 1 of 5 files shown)


src/Tasks.UnitTests/ResolveNonMSBuildProjectOutput_Tests.cs
```diff
@@ -119,32 +128,32 @@ public void TestResolve()
 
             // non matching project in string
             projectOutputs = new Hashtable();
-            projectOutputs.Add("{11111111-1111-1111-1111-111111111111}", @"obj\wrong.dll");
+            projectOutputs.Add("{11111111-1111-1111-1111-111111111111}", Path.Combine("obj", "wrong.dll"));
```
Forgind: This seems like something that might have been failing for quite a while but went unnoticed because we don't test full framework/macOS. Is that possible? If so, do you think we should add that as one of the standard checks?

rainersigwald: .NET ("full framework") runs only on Windows. There is no "full framework/macOS" scenario.

rainersigwald:
> the point of Core is that it doesn't have everything, but what it has is necessary

I wouldn't say that. MSBuild on .NET Core is cross-platform. It has a subset of functionality because the initial port was focused on command line .NET Core build scenarios and not everything has been ported (or can be, depending on specific functionality). There is no feature that .NET Core MSBuild "shouldn't" have. It's just MSBuild, running on a different runtime--and with a different set of user scenarios and thus priorities.

> and used a lot.

As the linked issue indicates, this task is used in Roslyn `MSBuildWorkspace` scenarios for design-time-like builds in .NET Core environments.

PR #5046, opened 2020-01-10
Revert "Apply a solution filter"

rainersigwald:
Reverts microsoft/msbuild#4899 -- it caused failures when inserting the updated build to Visual Studio.


rainersigwald:
We're going to branch retroactively, so closing this.

PR #5055, opened 2020-01-14
Enable logging messages created before BuildManager.BeginBuild

cdmihai:
Should enable https://github.com/microsoft/msbuild/pull/5007 to log places where msbuild searches for command line arguments.


BenVillalobos:
Merging this since master has snapped to 16.6.

PR #5069, opened 2020-01-23
Keep track of ProjectRootElements that change via save to a stream Fixes #4824

### LINKED ISSUE #4824: Save in ClickOnce scenario does not update ProjectRootElement._lastWriteTimeWhenRead (regression)

Forgind:
Writing to a stream may or may not involve modifying a project, so we previously did not update _lastWriteTimeWhenRead to reflect changes from a stream. When we changed how we calculate MSBuildAllProjects to only include the most recently modified project (so we don't have to look through a potentially long list of irrelevant projects), we started missing out on project files that were written via writing a stream but not modified explicitly with a non-stream write—these would not be added to MSBuildAllProjects because they weren't the "most recently modified" project, causing them never to be built if the most recently written project according to MSBuildAllProjects had been written before the last build. This caused the error in ClickOnce publishing, so this Fixes #4824.

### DIFF
--- a/ref/Microsoft.Build/net/Microsoft.Build.cs
+++ b/ref/Microsoft.Build/net/Microsoft.Build.cs
@@ -226,7 +226,6 @@ namespace Microsoft.Build.Construction
     public partial class ProjectRootElement : Microsoft.Build.Construction.ProjectElementContainer
     {
         internal ProjectRootElement() { }
-        public System.Nullable<System.DateTime> StreamTime;
         public System.Collections.Generic.ICollection<Microsoft.Build.Construction.ProjectChooseElement> ChooseElements { get { throw null; } }
         public override string Condition { get { throw null; } set { } }
         public override Microsoft.Build.Construction.ElementLocation ConditionLocation { get { throw null; } }
--- a/ref/Microsoft.Build/net/Microsoft.Build.cs
+++ b/ref/Microsoft.Build/net/Microsoft.Build.cs
@@ -226,6 +226,7 @@ namespace Microsoft.Build.Construction
     public partial class ProjectRootElement : Microsoft.Build.Construction.ProjectElementContainer
     {
         internal ProjectRootElement() { }
+        public System.Nullable<System.DateTime> StreamTime;
         public System.Collections.Generic.ICollection<Microsoft.Build.Construction.ProjectChooseElement> ChooseElements { get { throw null; } }
         public override string Condition { get { throw null; } set { } }
         public override Microsoft.Build.Construction.ElementLocation ConditionLocation { get { throw null; } }
--- a/src/Build/Construction/ProjectRootElement.cs
+++ b/src/Build/Construction/ProjectRootElement.cs
@@ -1590,7 +1590,7 @@ namespace Microsoft.Build.Construction
                 XmlDocument.Save(projectWriter);
             }
 
-            StreamTime = DateTime.Now;
+            StreamTime = DateTime.Now.ToLocalTime();
             _versionOnDisk = Version;
         }
 
--- a/src/Build/Construction/ProjectRootElement.cs
+++ b/src/Build/Construction/ProjectRootElement.cs
@@ -617,7 +617,7 @@ namespace Microsoft.Build.Construction
         /// This can be used to see whether the file has been changed on disk
         /// by an external means.
         /// </summary>
-        public DateTime LastWriteTimeWhenRead => Link != null ? RootLink.LastWriteTimeWhenRead : _lastWriteTimeWhenRead;
+        public DateTime LastWriteTimeWhenRead => Link != null ? RootLink.LastWriteTimeWhenRead : _lastWriteTimeWhenRead.ToLocalTime();
 
         internal DateTime? StreamTime = null;
 
@@ -1532,7 +1532,7 @@ namespace Microsoft.Build.Construction
                 // come from disk.
                 if (fileInfo != null)
                 {
-                    _lastWriteTimeWhenRead = fileInfo.LastWriteTime;
+                    _lastWriteTimeWhenRead = fileInfo.LastWriteTime.ToUniversalTime();
                     if (_lastWriteTimeWhenRead > StreamTime)
                     {
                         StreamTime = null;
@@ -1590,7 +1590,7 @@ namespace Microsoft.Build.Construction
                 XmlDocument.Save(projectWriter);
             }
 
-            StreamTime = DateTime.Now.ToLocalTime();
+            StreamTime = DateTime.Now.ToUniversalTime();
             _versionOnDisk = Version;
         }
 
@@ -2061,7 +2061,7 @@ namespace Microsoft.Build.Construction
                     XmlDocument.FullPath = fullPath;
                 }
 
-                _lastWriteTimeWhenRead = FileUtilities.GetFileInfoNoThrow(fullPath).LastWriteTime;
+                _lastWriteTimeWhenRead = FileUtilities.GetFileInfoNoThrow(fullPath).LastWriteTime.ToUniversalTime();
                 if (StreamTime < _lastWriteTimeWhenRead)
                 {
                     StreamTime = null;
--- a/src/Build/Construction/ProjectRootElement.cs
+++ b/src/Build/Construction/ProjectRootElement.cs
@@ -619,7 +619,7 @@ namespace Microsoft.Build.Construction
         /// </summary>
         public DateTime LastWriteTimeWhenRead => Link != null ? RootLink.LastWriteTimeWhenRead : _lastWriteTimeWhenReadUtc.ToLocalTime();
 
-        internal DateTime? StreamTime = null;
+        internal DateTime? StreamTimeUtc = null;
 
         /// <summary>
         /// This does not allow conditions, so it should not be called.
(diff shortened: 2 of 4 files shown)


ref/Microsoft.Build/net/Microsoft.Build.cs
```diff
@@ -226,6 +226,7 @@ public partial class ProjectPropertyGroupElement : Microsoft.Build.Construction.
     public partial class ProjectRootElement : Microsoft.Build.Construction.ProjectElementContainer
     {
         internal ProjectRootElement() { }
+        public System.Nullable<System.DateTime> StreamTime;
```
rainersigwald: Does this actually need to be public, or can it be `internal`?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1988,6 +1993,11 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
                                 _lastModifiedProject = importedProjectElement;
                             }
 
+                            if (importedProjectElement.StreamTime != null && importedProjectElement.StreamTime > _lastModifiedProject.LastWriteTimeWhenRead)
+                            {
+                                _streamImports.Append(";").Append(importedProjectElement.FullPath);
```
rainersigwald: I think it would be less allocatey to keep a list of elements and do a `Join(';')` at the end.

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -1584,6 +1586,7 @@ public void Save(TextWriter writer)
                 XmlDocument.Save(projectWriter);
             }
 
+            StreamTime = DateTime.Now;
```
jeffkl: Is this guaranteed to write to disk and to the same path that represents the file?  Why not just update the last write time?  If its a memory stream, or anything not on disk, then it wouldn't trigger a rebuild anyway since the datetime stamp would be calculated off of the file during build.  So it seems like it would be better to just assume the file was updated on disk if the stream was updated so everything works the same.  Unless of course I'm missing something.

Forgind: The "last modified project" [is calculated](https://github.com/microsoft/msbuild/blob/master/src/Build/Evaluation/Evaluator.cs#L1986-L1989) purely based on `_lastWriteTimeWhenRead`. Since we only keep the last modified project to add to MSBuildAllProjects, it would break that should the last thing we wrote to not be a file on disk.

jeffkl: > Since we only keep the last modified project to add to MSBuildAllProjects, it would break that should the last thing we wrote to not be a file on disk.

I'm still confused.  If we didn't write a file to disk, the up-to-date check wouldn't pick it up either right?  Just adding the path of the import/project to MSBuildAllProjects isn't enough, the timestamp on disk has to change as well.

Maybe we just need a new LastWriteTime property which reflects when the file is modified but doesn't cause the file to be re-read from disk?  And then this calculation would come from that.

Forgind: This is the scenario I mean:
```
Write to file a
Build
Write to file b
Write to stream c that is not represented by a file on disk
Build
```
In this case, if we update the last write time of the projectrootelement corresponding to c, then in the part of Evaluator I linked above, it will think the last modified project was c, not b, so it will prepend that to MSBuildAllProjects.
When we check to see if we need to rebuild, there's no corresponding file on disk for the last project "written to," so since we only kept the last project (for performance) instead of keeping all of them, our up-to-date check fails—we miss that file b was written to.
In other words, the problem is that when we write to a stream, it could be writing to a file (in which case, we should update _lastWriteTimeWhenRead) or not, in which case doing so would prevent us from noticing the real most recently written ProjectRootElement.

Could you explain your LastWriteTime idea a little more and how we can detect that a file has been modified without reading it? I don't understand how it differs from _lastWriteTimeWhenRead.

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -619,6 +619,8 @@ public int Version
         /// </summary>
         public DateTime LastWriteTimeWhenRead => Link != null ? RootLink.LastWriteTimeWhenRead : _lastWriteTimeWhenRead;
 
+        internal DateTime? StreamTime = null;
```
danmosemsft: Seems that some DateTime's in this type are stored as Utc and some are not. Would it be helpful to consistently use UTC? Or else maybe consider a suffix on the field name eg `NotUtc`

Forgind: It's important for StreamTime to be in the same time zone as the file's time and _lastWriteTimeWhenRead, and I think using local time is the least confusing. I changed it to explicitly convert it to local time just for clarity.

rainersigwald: I prefer the other direction: use UTC everywhere. That's what we do for most things. Unfortunately, we didn't catch this when we added `LastWriteTimeWhenRead` as a public field instead of `LastWriteTimeWhenReadUtc`. So I guess converging on local time in this class is probably the way to go.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -2367,12 +2378,12 @@ private void SetAllProjectsProperty()
...
                 P oldValue = _data.GetProperty(Constants.MSBuildAllProjectsPropertyName);
-
+                string streamImports = string.Join(";", _streamImports.ToArray());
                 P newValue = _data.SetProperty(
                     Constants.MSBuildAllProjectsPropertyName,
                     oldValue == null
-                        ? _lastModifiedProject.FullPath
-                        : $"{_lastModifiedProject.FullPath};{oldValue.EvaluatedValue}",
+                        ? $"{_lastModifiedProject.FullPath};{streamImports}"
+                        : $"{_lastModifiedProject.FullPath};{streamImports};{oldValue.EvaluatedValue}",
```
rainersigwald: Hmm. I don't love having extra semicolons in this list, which we will now in the "normal" (no stream save) case. But not a huge deal, everything should be perfectly capable of stripping them out later.

src/Build/Construction/ProjectRootElement.cs
danmosemsft: The public properties need not have the Utc suffix, but the backing fields could be Utc, and have utc suffix, without breaking anyone (unless the machine is changing timezones)

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -1584,6 +1590,7 @@ public void Save(TextWriter writer)
                 XmlDocument.Save(projectWriter);
             }
 
+            StreamTime = DateTime.Now.ToUniversalTime();
```
rainersigwald:
```suggestion
            StreamTime = DateTime.UtcNow;
```

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1988,6 +1995,12 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
                                 _lastModifiedProject = importedProjectElement;
                             }
 
+                            if (importedProjectElement.StreamTime != null && importedProjectElement.StreamTime?.ToLocalTime() > _lastModifiedProject.LastWriteTimeWhenRead)
```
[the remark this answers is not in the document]
Forgind: DateTime? doesn't have ToLocalTime() without the ?, so I removed the first null check instead.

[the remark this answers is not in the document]
rainersigwald: Ah, you can do `.Value` to get through a nullable. But removing the first one sounds good.

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -617,7 +617,7 @@ public int Version
         /// This can be used to see whether the file has been changed on disk
         /// by an external means.
         /// </summary>
-        public DateTime LastWriteTimeWhenRead => Link != null ? RootLink.LastWriteTimeWhenRead : _lastWriteTimeWhenRead.ToLocalTime();
+        public DateTime LastWriteTimeWhenRead => Link != null ? RootLink.LastWriteTimeWhenRead : _lastWriteTimeWhenReadUtc.ToLocalTime();
 
         internal DateTime? StreamTime = null;
```
rainersigwald: Should this be 
```suggestion
        internal DateTime? StreamTimeUtc = null;
```

?

PR #5070, opened 2020-01-23
Publish x64 Architecture Symbols

### LINKED ISSUE #5060: Symbols not published for amd64 binaries in 16.5
See, for example, [AB#1050605](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1050605).

BenVillalobos:
fixes #5060 

This fixes our symbol check failures on our VS PRs by copying over the relevant `.pdb`s to where they need to be to get published during a pipeline build.

### DIFF
--- a/eng/AfterSigning.targets
+++ b/eng/AfterSigning.targets
@@ -1,0 +1,12 @@
+<Project>
+    <Target Name="CopyAMD64Symbols" BeforeTargets="Build">
+        <Copy 
+            SourceFiles="$(ArtifactsBinDir)MSBuild\x64\Release\net472\MSBuild.pdb"
+            DestinationFolder="$(ArtifactsSymStoreDirectory)\MSBuild\net472\amd64"
+            />
+        <Copy 
+            SourceFiles="$(ArtifactsBinDir)MSBuildTaskHost\x64\Release\net35\MSBuildTaskHost.pdb"
+            DestinationFolder="$(ArtifactsSymStoreDirectory)\MSBuildTaskHost\net35\amd64"
+            />
+    </Target>
+</Project>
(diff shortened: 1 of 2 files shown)


eng/AfterSigning.targets
```diff
@@ -0,0 +1,12 @@
+<Project>
+    <Target Name="CopyAMD64Symbols" BeforeTargets="Build">
```
rainersigwald: Condition this on `'$(OfficialBuild)' == 'true'`?

src/MSBuild/MSBuild.csproj
```diff
@@ -36,7 +36,7 @@
     <NoPackageAnalysis>true</NoPackageAnalysis>
     <AddAppConfigToBuildOutputs>false</AddAppConfigToBuildOutputs>
 
-    <DebugType Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">full</DebugType><!-- Work around arcade stomping on symbols for same-program-different-arches. -->
+    <DebugType Condition="'$(Platform)' == 'x64'">full</DebugType><!-- Work around arcade stomping on symbols for same-program-different-arches. -->
```
rainersigwald: Can you fix my incorrect comment? point to AfterSigning.targets too, and describe how it works for x64 and non-x64 flavors.

BenVillalobos: Here's my working comment. I'm not sure what to put in place of 'more compatible', can you provide some input for that? or reword as necessary.
 ```Setting DebugType here goes hand in hand with AfterSigning.targets. This is to prompt the x64 build to produce a 'full' .pdb that's `more compatible` then 'portable' and 'embedded' .pdbs. This doesn't get set on 32 bit architecture, which will default to 'embedded' and 'pdb2pdb' will convert those as needed. See github pull #5070 for context. ```

rainersigwald: looks ok to me, but use a full URL for the PR. and a full local path for aftersigning.

eng/AfterSigning.targets
```diff
@@ -0,0 +1,12 @@
+<Project>
```
Forgind: How does this get called? This is a new file, so there can't be any references to it previously, and you didn't add one here.

rainersigwald: https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md#engaftersigningtargets-optional

PR #5074, opened 2020-01-24
Updating BootstrapUtil to look for packages in additional folders

MSLukeWest:
We want teams to be able to install bootstrapper packages from outside the VS MSIs where they currently come from.  Right now we only enumerate packages from a single directory that's defined in the HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\GenericBootstrapper reg key.  I'm making it so if there's any additional paths specified under the AdditionalPackagePaths sub key we'll look for bootstrapper packages in those paths as well.

One small issue with this change is that when a bootstrapper specified in a clickonce or setup project isn't found we give a warning like this:

_Item '.NETFramework,Version=v4.7.2' could not be located in 'C:\Program Files (x86)\Microsoft SDKs\ClickOnce Bootstrapper\'_

Ideally this would be updated to also mention any possible additional paths that have been specified using this new support.  But since that would require a string change and a new public interface to properly support I've decided to delay that part of the change until 16.6.

### DIFF
--- a/src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
+++ b/src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
@@ -444,13 +444,20 @@ namespace Microsoft.Build.Tasks.Deployment.Bootstrapper
 
             foreach (string file in files)
             {
-                string folder = System.IO.Path.GetDirectoryName(file);
-                if (folder.Substring(0, invariantPath.Length).ToLowerInvariant().CompareTo(invariantPath) == 0)
+                List<string> packagePaths = new List<string>() { invariantPath };
+                packagePaths.AddRange(Util.AdditionalPackagePaths.Select(p => Util.AddTrailingChar(p.ToLowerInvariant(), System.IO.Path.DirectorySeparatorChar)));
+                foreach (string packagePath in packagePaths)
                 {
-                    string relPath = folder.Substring(invariantPath.Length);
-                    if (!folders.Contains(relPath))
+                    string folder = System.IO.Path.GetDirectoryName(file);
+                    if (folder.Length >= packagePath.Length && folder.Substring(0, packagePath.Length).ToLowerInvariant().CompareTo(packagePath) == 0)
                     {
-                        folders.Add(relPath);
+                        string relPath = folder.Substring(packagePath.Length);
+                        if (!folders.Contains(relPath))
+                        {
+                            folders.Add(relPath);
+                        }
+
+                        break;
                     }
                 }
             }
@@ -576,18 +583,23 @@ namespace Microsoft.Build.Tasks.Deployment.Bootstrapper
             _xmlNamespaceManager.AddNamespace(BOOTSTRAPPER_PREFIX, BOOTSTRAPPER_NAMESPACE);
 
             XmlElement rootElement = _document.CreateElement("Products", BOOTSTRAPPER_NAMESPACE);
-            string packagePath = PackagePath;
 
-            if (FileSystems.Default.DirectoryExists(packagePath))
+            List<string> packagePaths = new List<string>() { PackagePath };
+            packagePaths.AddRange(Util.AdditionalPackagePaths);
+            foreach (string packagePath in packagePaths)
             {
-                foreach (string strSubDirectory in Directory.GetDirectories(packagePath))
+                if (FileSystems.Default.DirectoryExists(packagePath))
                 {
-                    int nStartIndex = packagePath.Length;
-                    if ((strSubDirectory.ToCharArray())[nStartIndex] == System.IO.Path.DirectorySeparatorChar)
+                    foreach (string strSubDirectory in Directory.GetDirectories(packagePath))
                     {
-                        nStartIndex = nStartIndex + 1;
+                        int nStartIndex = packagePath.Length;
+                        if ((strSubDirectory.ToCharArray())[nStartIndex] == System.IO.Path.DirectorySeparatorChar)
+                        {
+                            nStartIndex = nStartIndex + 1;
+                        }
+
+                        ExploreDirectory(strSubDirectory.Substring(nStartIndex), rootElement, packagePath);
                     }
-                    ExploreDirectory(strSubDirectory.Substring(nStartIndex), rootElement);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


MSLukeWest:
@SvetBonev Please take a look

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
```diff
@@ -444,13 +444,20 @@ public string[] GetOutputFolders(string[] productCodes, string culture, string f
...
+                foreach (string packagePath in packagePaths)
                 {
-                    string relPath = folder.Substring(invariantPath.Length);
-                    if (!folders.Contains(relPath))
+                    string folder = System.IO.Path.GetDirectoryName(file);
+                    if (folder.Length >= packagePath.Length && folder.Substring(0, packagePath.Length).ToLowerInvariant().CompareTo(packagePath) == 0)
                     {
-                        folders.Add(relPath);
+                        string relPath = folder.Substring(packagePath.Length);
+                        if (!folders.Contains(relPath))
```
SvetBonev: Is the logic still correct? We might get the same relative path from multiple packages and I am not sure the results will be used correctly.

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
```diff
@@ -444,13 +444,20 @@ public string[] GetOutputFolders(string[] productCodes, string culture, string f
...
             {
-                string folder = System.IO.Path.GetDirectoryName(file);
-                if (folder.Substring(0, invariantPath.Length).ToLowerInvariant().CompareTo(invariantPath) == 0)
+                List<string> packagePaths = new List<string>() { invariantPath };
+                packagePaths.AddRange(Util.AdditionalPackagePaths.Select(p => Util.AddTrailingChar(p.ToLowerInvariant(), System.IO.Path.DirectorySeparatorChar)));
+                foreach (string packagePath in packagePaths)
                 {
-                    string relPath = folder.Substring(invariantPath.Length);
-                    if (!folders.Contains(relPath))
+                    string folder = System.IO.Path.GetDirectoryName(file);
```
SvetBonev: Move outside the foreach.

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
```diff
@@ -444,13 +444,20 @@ public string[] GetOutputFolders(string[] productCodes, string culture, string f
 
             foreach (string file in files)
             {
-                string folder = System.IO.Path.GetDirectoryName(file);
-                if (folder.Substring(0, invariantPath.Length).ToLowerInvariant().CompareTo(invariantPath) == 0)
+                List<string> packagePaths = new List<string>() { invariantPath };
+                packagePaths.AddRange(Util.AdditionalPackagePaths.Select(p => Util.AddTrailingChar(p.ToLowerInvariant(), System.IO.Path.DirectorySeparatorChar)));
```
SvetBonev: Consider keeping the processed list in a local variable outside the foreach.

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
MSLukeWest: Yes, this is still correct.  This method is used to determine the additional subfolders needed in the click once publish directory when you have the "Download prerequisites from the same location as my application" option selected.  The relative path comes from the relative path of the package under the packages directory (Example: "C:\Program Files (x86)\Microsoft SDKs\ClickOnce Bootstrapper\Packages\DotNetFX472" => "DotNetFX472").  In the unlikely case where an identical sub folder name is used between two packages that come from different package folders the package files from both would just get copied into this same subfolder.

SvetBonev:
Looks good to me.

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
SvetBonev: Sounds good.

src/Tasks/BootstrapperUtil/Util.cs
```diff
@@ -145,6 +149,44 @@ public static string GetDefaultPath(string visualStudioVersion)
...
         }
 
+        // Gets the list of additional paths to inspect for packages as defined in the registry
+        public static List<string> AdditionalPackagePaths
+        {
+            get
+            {
+                if (s_additionalPackagePaths == null)
+                {
+                    s_additionalPackagePaths = new List<string>();
```
rainersigwald: There shouldn't really be a race condition here because the task shouldn't be reentrant or anything, but can you accumulate this in a local variable and only assign it to the static field at the end?

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
```diff
@@ -442,15 +442,24 @@ public string[] GetOutputFolders(string[] productCodes, string culture, string f
             var files = new List<string>();
             BuildPackages(settings, null, null, files, null);
 
+            List<string> packagePaths = new List<string>() { invariantPath };
+            packagePaths.AddRange(Util.AdditionalPackagePaths.Select(p => Util.AddTrailingChar(p.ToLowerInvariant(), System.IO.Path.DirectorySeparatorChar)));
```
rainersigwald: What's the motivation for `ToLowerInvariant` here? It's a loc problem (Turkish `İ`), and the filesystem is already case-insensitive (usually, for NTFS and friends on Windows).

rainersigwald: Ah, I see you're following the existing pattern. Do you know if it's required for some reason? If it is, please comment why--and do the normalization when you read the reg key, rather than on every use.

src/Tasks/BootstrapperUtil/Util.cs
MSLukeWest: Fixed

src/Tasks/BootstrapperUtil/BootstrapperBuilder.cs
MSLukeWest: As you say, I'm doing it this way to follow the existing pattern.  I'm also not sure why ToLowerInvariant is necessary here.

Also, since the other code path that uses AdditionalPackagePaths (see RefreshProducts) doesn't call ToLowerInvariant on them I'm going to leave the call here.

src/Tasks/BootstrapperUtil/Util.cs
```diff
@@ -175,12 +175,14 @@ public static List<string> AdditionalPackagePaths
...
                                     {
-                                        s_additionalPackagePaths.Add(path);
+                                        additionalPackagePaths.Add(path);
                                     }
                                 }
                             }
                         }
                     }
+
+                    s_additionalPackagePaths = new List<string>(additionalPackagePaths);
```
rainersigwald:
```suggestion
                    s_additionalPackagePaths = additionalPackagePaths;
```
?

MSLukeWest: Updated

Issue #2892, opened 2018-01-23
MSBuild /help should link to docs


rainersigwald:
Somewhere in the output of `msbuild /help` we should put a link to https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild.

(Thanks for the suggestion, @whoisj: https://github.com/Microsoft/msbuild/issues/2887#issuecomment-359844657.)

joseotavioq:
Hi @rainersigwald!

Could it be something like this, when command `msbuild /help` is executed?

![image](https://user-images.githubusercontent.com/1444234/72208774-a282d780-3485-11ea-9e88-9bbcd87f42ee.png)

PR #5091, opened 2020-01-29
Added the message at the end of the command /help output

joseotavioq:
A suggestion to solve the issue #2892

### DIFF
--- a/src/MSBuild/Resources/Strings.resx
+++ b/src/MSBuild/Resources/Strings.resx
@@ -787,10 +787,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
-  <data name="HelpMessage_37_DocsLink" UESanitized="false" Visibility="Public">
-    <value>For more detailed information access the link:
-                     https://docs.microsoft.com/visualstudio/msbuild/msbuild
-    </value>
+  <data name="HelpMessage_37_DocsLink" Visibility="Public">
+    <value>For more detailed information, see https://aka.ms/msbuild/docs</value>
   </data>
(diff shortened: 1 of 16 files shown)


src/MSBuild/Resources/Strings.resx
```diff
@@ -787,6 +787,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       LOCALIZATION: None of the lines should be longer than a standard width console window, eg 80 chars.
     </comment>
   </data>
+  <data name="HelpMessage_37_DocsLink" UESanitized="false" Visibility="Public">
```
Forgind: We don't use this anymore
```suggestion
  <data name="HelpMessage_37_DocsLink" Visibility="Public">
```

src/MSBuild/Resources/Strings.resx
```diff
@@ -787,6 +787,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <value>For more detailed information access the link:
```
Forgind: nit:
```suggestion
    <value>For more detailed information, see:
```

rainersigwald:
```suggestion
    <value>For more detailed information, see https://aka.ms/msbuild/docs
```

? That shortlink points to `https://docs.microsoft.com/visualstudio/msbuild/msbuild` but can be easily moved if we need to later on.

src/MSBuild/Resources/Strings.resx
joseotavioq: @Forgind  Done!

src/MSBuild/Resources/Strings.resx
joseotavioq: @rainersigwald Done!

Issue #5037, opened 2020-01-08
Update plugin assembly loading for .NET 3/5


rainersigwald:
Several improvements to #4916 are possible once we're moved to target .NET 3+:

- [x] Name AssemblyLoadContexts
- [ ] Use `AssemblyDependencyResolver` to make native dependencies easier to handle (https://github.com/microsoft/msbuild/pull/4916#discussion_r363712530)
- [ ] Audit MSBuild's dependencies and consider handling `Resolving` rather than `Load` (https://github.com/microsoft/msbuild/pull/4916/files/6d4c3b1e2adcd6ce97233b85489c514b2b990949#r363710290, https://github.com/microsoft/msbuild/pull/4916#discussion_r363715494)

rainersigwald:
Note: this was already mostly done in https://github.com/microsoft/msbuild/commit/90ac0c3110ddbc112780fdb009079dc6288e5bd7. We could mostly pull back to that, but we'll need to keep the escape hatches and other behaviors of the current implementation.

AArnott:
Please to take a look at the unique requirements proposed by #1755 which was recently duped against this one. The goal being that now that msbuild isn't a Windows, .NET Framework only process, we need a more scalable, simpler model for plugins to ship without shipping the full transitive closure of their own runtime dependencies for every single OS+runtime, while still being able to run on each of these.

rainersigwald:
@AArnott are you saying there are MSBuild-specific requirements that are not covered by [`AssemblyDependencyResolver`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblydependencyresolver?view=netcore-3.1)?

AArnott:
@rainersigwald I'm not sure. I don't see anything in the `AssemblyDependencyResolver` doc to suggest that it can load a plugin's dependencies from the nuget package cache.

mwpowellhtx:
@rainersigwald Are there examples how to plugin with this in view? Thank you...

rainersigwald:
You might be interested in https://docs.microsoft.com/dotnet/core/tutorials/creating-app-with-plugin-support#plugin-with-library-dependencies, @mwpowellhtx.

marcin-krystianc:
Hi @rainersigwald, I've recently come across a memory leak caused by the fact that assemblies are loaded into `AssemblyLoadContext` but they are never unloaded. 
[...]

PR #5098, opened 2020-01-31
Add an escape hatch for ALC isolation

### LINKED ISSUE #5086: Escape hatch for legacy Core ALC behavior
Allow disabling #4916 via an environment variable, in case of further bugs or #5084 being more severe than I currently expect.

rainersigwald:
Restore the original load-with-funky-rules-into-the-default-ALC behavior under an environment-variable escape hatch. Fixes #5086.

This looks like a lot of code, but it's almost all just undeletions. I cherry-picked the change on top of the existing ALC changes, so you can see that more clearly: https://github.com/rainersigwald/msbuild/compare/alc-escape-hatch-base...rainersigwald:alc-escape-hatch-cleanerdiff?expand=1

Compare that to the diff in #4916 to see that it makes it look like this is just adding functionality.

### DIFF
--- a/src/Shared/CoreCLRAssemblyLoader.cs
+++ b/src/Shared/CoreCLRAssemblyLoader.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using Microsoft.Build.Shared.FileSystem;
+using Microsoft.Build.Utilities;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
+using System.Runtime.Loader;
 
 namespace Microsoft.Build.Shared
 {
@@ -14,8 +17,26 @@ namespace Microsoft.Build.Shared
     internal sealed class CoreClrAssemblyLoader
     {
         private readonly Dictionary<string, Assembly> _pathsToAssemblies = new Dictionary<string, Assembly>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, Assembly> _namesToAssemblies = new Dictionary<string, Assembly>();
+        private readonly HashSet<string> _dependencyPaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         private readonly object _guard = new object();
 
+        private bool _resolvingHandlerHookedUp = false;
+
+        private static readonly Version _currentAssemblyVersion = new Version(Microsoft.Build.Shared.MSBuildConstants.CurrentAssemblyVersion);
+
+        public void AddDependencyLocation(string fullPath)
+        {
+            if (fullPath == null)
+            {
+                throw new ArgumentNullException(nameof(fullPath));
+            }
+
+            lock (_guard)
+            {
+                _dependencyPaths.Add(fullPath);
+            }
+        }
 
         public Assembly LoadFromPath(string fullPath)
         {
@@ -31,6 +52,38 @@ namespace Microsoft.Build.Shared
             // folders in a NuGet package).
             fullPath = FileUtilities.NormalizePath(fullPath);
 
+            if (Traits.Instance.EscapeHatches.UseSingleLoadContext)
+            {
+                return LoadUsingLegacyDefaultContext(fullPath);
+            }
+            else
+            {
+                return LoadUsingPluginContext(fullPath);
+            }
+        }
+
+        private Assembly LoadUsingLegacyDefaultContext(string fullPath)
+        {
+            lock (_guard)
+            {
+                if (!_resolvingHandlerHookedUp)
+                {
+                    AssemblyLoadContext.Default.Resolving += TryResolveAssembly;
+                    _resolvingHandlerHookedUp = true;
+                }
+
+                Assembly assembly;
+                if (_pathsToAssemblies.TryGetValue(fullPath, out assembly))
+                {
+                    return assembly;
+                }
+
+                return LoadAndCache(AssemblyLoadContext.Default, fullPath);
+            }
+        }
+
+        private Assembly LoadUsingPluginContext(string fullPath)
+        {
             lock (_guard)
             {
                 Assembly assembly;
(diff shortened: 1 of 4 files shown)


src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -51,5 +104,98 @@ public Assembly LoadFromPath(string fullPath)
...
+            foreach (var cultureSubfolder in string.IsNullOrEmpty(assemblyName.CultureName)
+                // If no culture is specified, attempt to load directly from
+                // the known dependency paths.
+                ? new[] { string.Empty }
+                // Search for satellite assemblies in culture subdirectories
+                // of the assembly search directories, but fall back to the
+                // bare search directory if that fails.
+                : new[] { assemblyName.CultureName, string.Empty })
+            {
+                foreach (var searchPath in searchPaths)
```
Forgind: At what point are _dependencyPaths added? Is it somewhere buried in `ComputeClosure`?

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -51,5 +104,98 @@ public Assembly LoadFromPath(string fullPath)
...
+                {
+                    foreach (var extension in MSBuildLoadContext.Extensions)
+                    {
+                        var candidatePath = Path.Combine(searchPath,
```
Forgind: This makes me slightly nervous just because it vaguely resembles @BenVillalobos's DependentUpon change.

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -51,5 +104,98 @@ public Assembly LoadFromPath(string fullPath)
...
+                Assembly assembly = TryGetWellKnownAssembly(context, assemblyName);
+
+                if (assembly != null)
+                {
+                    return assembly;
+                }
+
+                if (_namesToAssemblies.TryGetValue(assemblyName.FullName, out assembly))
+                {
+                    return assembly;
```
Forgind: Is there no need to check version here?

src/Shared/Traits.cs
```diff
@@ -179,6 +179,11 @@ public bool LogProjectImports
         /// </summary>
         public readonly bool DisableNuGetSdkResolver = Environment.GetEnvironmentVariable("MSBUILDDISABLENUGETSDKRESOLVER") == "1";
 
+        /// <summary>
+        /// Disable AssemblyLoadContext isolation for plugins.
+        /// </summary>
+        public readonly bool UseSingleLoadContext = Environment.GetEnvironmentVariable("MSBUILDSINGLELOADCONTEXT") == "1";
```
Forgind: Do you think this should be opt-in?

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -51,5 +104,98 @@ public Assembly LoadFromPath(string fullPath)
...
+                }
+
+                return TryResolveAssemblyFromPaths(context, assemblyName, _dependencyPaths);
```
Forgind: This looks like mostly the same as the previous version—it just adds the possibility of cultures and custom search paths. Is that right?

src/Shared/MSBuildLoadContext.cs
```diff
@@ -19,9 +19,8 @@ namespace Microsoft.Build.Shared
     internal class MSBuildLoadContext : AssemblyLoadContext
     {
         private readonly string _directory;
-        private readonly object _guard = new object();
 
-        private static readonly ImmutableHashSet<string> _wellKnownAssemblyNames =
+        internal static readonly ImmutableHashSet<string> WellKnownAssemblyNames =
```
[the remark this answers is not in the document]
rainersigwald: It generally doesn't need to be because MSBuild is what does the loading and thus doesn't need to be unified. I added it here to smooth out some of our unit tests that do things differently.

src/Shared/Traits.cs
rainersigwald: No, we want the new behavior by default, because we expect it to be better for most people. This PR just buys us time to fix problems if it's not . . .

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -31,6 +52,38 @@ public Assembly LoadFromPath(string fullPath)
...
+                return LoadUsingLegacyDefaultContext(fullPath);
+            }
+            else
+            {
+                return LoadUsingPluginContext(fullPath);
+            }
+        }
+
+        private Assembly LoadUsingLegacyDefaultContext(string fullPath)
+        {
```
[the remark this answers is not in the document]
rainersigwald: This is an undelete for compat reasons; I'd rather not add additional points of failure at the same time.

src/Shared/CoreCLRAssemblyLoader.cs
```diff
@@ -51,5 +104,98 @@ public Assembly LoadFromPath(string fullPath)
...
+            if (!MSBuildLoadContext.WellKnownAssemblyNames.Contains(assemblyName.Name))
+            {
+                return null;
+            }
+
+            // Ensure we are attempting to load a matching version
+            // of the Microsoft.Build.* assembly.
+            assemblyName.Version = _currentAssemblyVersion;
+
+            var searchPaths = new[] { Assembly.GetExecutingAssembly().Location };
```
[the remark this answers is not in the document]
rainersigwald: I am extremely averse to any "magical" assembly loading technique like that. This whole mess is bad enough already.

src/Shared/CoreCLRAssemblyLoader.cs
rainersigwald: Undeleting, so don't want to change behavior. But no, because .NET Core allows only one assembly per simple name in the same ALC, so if it's already loaded at _some_ version, that's the one you get.

src/Shared/CoreCLRAssemblyLoader.cs
rainersigwald: The same as what previous version? It should be the same as the v16.4 version.

src/Shared/CoreCLRAssemblyLoader.cs
rainersigwald: This emulates the .NET Framework's loader behavior. It can blessedly be deleted with #5037, but until then we have to do it ourselves. And again: this is not new code, just undeleted.

src/Shared/CoreCLRAssemblyLoader.cs
rainersigwald: Nice catch! Fixed with 44af1ff6

src/Shared/CoreCLRAssemblyLoader.cs
Forgind: This was just a bad question. That's not what I meant, but I can't tell you what I did mean.

PR #5099, opened 2020-01-31
vs.relatedProcessFiles for MSBuild assemblies

### LINKED ISSUE #5025: Use vs.relatedProcessFiles to prevent torn state on VS upgrade
It's possible for an application outside of VS to use and lock MSBuild assemblies. If Visual Studio Setup then tries to update MSBuild assemblies, the files in use will require a reboot, and possibly be mismatched across files (for instance, if `Microsoft.Build.Framework.dll` is in use but the engine isn't, there might be `Microsoft.Build.dll` might be updated to a newer version while Framework is stuck waiting for reboot.

The VS installer supports `vs.relatedProcessFile` and `vs.relatedProcessDirectory` to tell the Windows Restart Manager to check whether the file/files in the folder are in use and offer to kill the process at setup/upgrade time. We should add MSBuild assemblies (or maybe the whole folder?) to this in our `.swr`.

[Internal-only doc link](https://microsoft.sharepoint.com/teams/DD_Willow/_layouts/15/Doc.aspx?sourcedoc={9f9aba5f-0856-4a8d-a1bc-efb69c4ffe54}&action=edit&wd=target%28Dev%20Docs.one%7C39a1cebd-3922-4c5a-b90c-e1c6bb99ee00%2FAuthoring%20related%20process%20directories%20%5C%2F%20files%20and%20services%7C7ca87616-4a45-461c-bda2-73f03af9acf1%2F%29)

rainersigwald:
Add an explicit list of MSBuild assemblies that might be loaded by an
application outside of the Visual Studio installation directory, so that
VS Setup can find any applications using these files and shut them down
before updating VS.

Closes #5025.

### DIFF
--- a/src/Package/MSBuild.VSSetup/files.swr
+++ b/src/Package/MSBuild.VSSetup/files.swr
@@ -9,6 +9,16 @@ vs.dependencies
   vs.dependency id=Microsoft.VisualStudio.PackageGroup.NuGet
                 version=[15.0,17.0)
 
+vs.relatedProcessFiles
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\Microsoft.Build.dll"
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\Microsoft.Build.Framework.dll"
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\Microsoft.Build.Tasks.Core.dll"
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\Microsoft.Build.Utilities.Core.dll"
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\amd64\Microsoft.Build.dll"
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\amd64\Microsoft.Build.Framework.dll"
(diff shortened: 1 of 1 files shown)


rainersigwald:
Example upgrade from the private build of this PR executed as an upgrade after creating a trivial MSBuildLocator-using application (`ConsoleApplication1`) to load MSBuild from VS.

image

src/Package/MSBuild.VSSetup/files.swr
```diff
@@ -9,6 +9,16 @@ vs.dependencies
   vs.dependency id=Microsoft.VisualStudio.PackageGroup.NuGet
                 version=[15.0,17.0)
 
+vs.relatedProcessFiles
+  vs.relatedProcessFile Path="[InstallDir]\MSBuild\Current\Bin\Microsoft.Build.dll"
```
Forgind: Should we also add MSBuild.exe/dll, Microsoft.Build.Conversion.Core.dll, Microsoft.Build.Engine.dll, and MSBuildTaskHost.exe?

rainersigwald: The `.exe`s are handled automatically by Setup logic (it already scans for _programs_ under the VS install directory).

The deprecated assemblies I'm less sure about. I _hope_ no one is using them, and MSBuildLocator doesn't help people load them. So I'm not sure how to trade off the cost of scanning for them (it is supposedly nontrivial--which is why we have to use this explicit list instead of having Setup just do "everything under VS") vs. the likelihood of them actually being in use by non-VS programs (hopefully none? But I don't have numbers).

I'd rather do this now and we can always change it later if we find more users.

PR #5101, opened 2020-02-03
Write generated AssemblyAttributes file to intermediate output path

tmat:
Fixes https://github.com/microsoft/msbuild/issues/1479 as suggested in the issue description.

### DIFF
--- a/src/Tasks.UnitTests/RegressionTests.cs
+++ b/src/Tasks.UnitTests/RegressionTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.IO;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.UnitTests;
 using Xunit;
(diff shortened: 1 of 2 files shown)


tmat:
@rainersigwald PTAL

rainersigwald:
Can you explain why the old file location is a problem after dotnet/roslyn#39136?

The test is failing on Mac and Linux with

```
/home/vsts/work/1/s/artifacts/bin/Microsoft.Build.Tasks.UnitTests/Debug/netcoreapp2.1/Microsoft.Common.CurrentVersion.targets(1177,5):  error MSB3644: The reference assemblies for .NETFramework,Version=v4.0 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks
```

I'm ok with conditionalizing the test to run only on Windows with a comment that it depends on .NET 4 targeting.

tmat:
> Can you explain why the old file location is a problem after dotnet/roslyn#39136?

The file does not contain source for any method bodies it used to be not included in the document table in the PDB. With dotnet/roslyn#39136 the file is included in the PDB, which leads to more issues. For example, a deterministic build won't normalize random paths outside of the repo root, path to %temp% directory leaks the user name, etc.

tmat:
@rainersigwald The test now runs on Windows only.

PR #5129, opened 2020-02-15
Import NuGet.props

cristinamanum:
Import NuGet.props file.

### DIFF
--- a/src/Tasks/Microsoft.Common.CrossTargeting.targets
+++ b/src/Tasks/Microsoft.Common.CrossTargeting.targets
@@ -136,7 +136,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     This will import NuGet restore targets. We need restore to work before any package assets are available.
   -->
   <PropertyGroup>
-    <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'=='' and '$([MSBuild]::IsRunningFromVisualStudio())'=='true'">$(MSBuildToolsPath32)\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets</NuGetRestoreTargets>
+    <MSBuildUseVisualStudioDirectoryLayout Condition="'$(MSBuildUseVisualStudioDirectoryLayout)'==''">$([MSBuild]::IsRunningFromVisualStudio())</MSBuildUseVisualStudioDirectoryLayout>
+    <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'=='' and '$(MSBuildUseVisualStudioDirectoryLayout)'=='true'">$(MSBuildToolsPath32)\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets</NuGetRestoreTargets>
     <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'==''">$(MSBuildToolsPath)\NuGet.targets</NuGetRestoreTargets>
   </PropertyGroup>
 
--- a/src/Tasks/Microsoft.Common.props
+++ b/src/Tasks/Microsoft.Common.props
@@ -169,8 +169,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         Import NuGet.props file.
     -->
   <PropertyGroup>
-    <IsRunningFromVisualStudio Condition="'$(IsRunningFromVisualStudio)'==''">$([MSBuild]::IsRunningFromVisualStudio())</IsRunningFromVisualStudio>
-    <NuGetProps Condition="'$(NuGetProps)'=='' and '$(IsRunningFromVisualStudio)'=='true'">$(MSBuildToolsPath32)\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.props</NuGetProps>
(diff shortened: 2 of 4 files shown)


src/Tasks/Microsoft.Common.props
```diff
@@ -164,4 +164,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <Import Project="$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Imports\Microsoft.Common.props\ImportAfter\*" Condition="'$(ImportUserLocationsByWildcardAfterMicrosoftCommonProps)' == 'true' and exists('$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Imports\Microsoft.Common.props\ImportAfter')"/>
     <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Imports\Microsoft.Common.props\ImportAfter\*" Condition="'$(ImportByWildcardAfterMicrosoftCommonProps)' == 'true' and exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Imports\Microsoft.Common.props\ImportAfter')"/>
   </ImportGroup>
+
+  <!-- 
+        Import NuGet.props file.
+    -->
+  <PropertyGroup>
+    <IsRunningFromVisualStudio Condition="'$(IsRunningFromVisualStudio)'==''">$([MSBuild]::IsRunningFromVisualStudio())</IsRunningFromVisualStudio>
```
nkolev92: I assume updating the targets to use this prop is out of scope here?

rainersigwald:
```suggestion
    <MSBuildUseVisualStudioDirectoryLayout Condition="'$(IsRunningFromVisualStudio)'==''">$([MSBuild]::IsRunningFromVisualStudio())</MSBuildUseVisualStudioDirectoryLayout>
```

? We talked about the property function being confusingly named so we have a chance to make it clearer here.

rainersigwald: I'd like to see that change made here while we're touching things; we might forget later.

src/Tasks/Microsoft.Common.props
```diff
@@ -164,4 +164,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <NuGetProps Condition="'$(NuGetProps)'=='' and '$(IsRunningFromVisualStudio)'=='true'">$(MSBuildToolsPath32)\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.props</NuGetProps>
+    <NuGetProps Condition="'$(NuGetProps)'==''">$(MSBuildToolsPath)\NuGet.props</NuGetProps>
+  </PropertyGroup>
+  <Import Condition="Exists('$(NuGetProps)')" Project="$(NuGetProps)" />
```
rainersigwald: _Should_ this be conditioned? We now expect it to be part of every MSBuild distribution, right? And NuGet is licensed compatibly so that people could distribute it?

cristinamanum: If the question is if we should have 
``` xml
Condition="Exists('$(NuGetProps)')"
```

I think that is good to not try importing without checking the existence of the file.

rainersigwald: For files we expect to preserve core functionality, I don't think that's true: it transforms a semi-actionable error "NuGet.props doesn't exist" into hidden behavior changes like "my project-wide package versions aren't being respected".

src/Tasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -136,7 +136,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     This will import NuGet restore targets. We need restore to work before any package assets are available.
   -->
   <PropertyGroup>
-    <NuGetRestoreTargets Condition="'$(NuGetRestoreTargets)'=='' and '$([MSBuild]::IsRunningFromVisualStudio())'=='true'">$(MSBuildToolsPath32)\..\..\..\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets</NuGetRestoreTargets>
+    <MSBuildUseVisualStudioDirectoryLayout Condition="'$(MSBuildUseVisualStudioDirectoryLayout)'==''">$([MSBuild]::IsRunningFromVisualStudio())</MSBuildUseVisualStudioDirectoryLayout>
```
rainersigwald: I wanted to say that this would be guaranteed to be already set by common.props, but I couldn't prove it to myself so I'm ok leaving the redundancy here. It shouldn't be expensive since it gets conditioned out if already set anyway.

src/Tasks/Microsoft.Common.props
cristinamanum: Updated

cristinamanum: Updated

src/Tasks/Microsoft.Common.props
cristinamanum: I see your point. Removed.

rainersigwald: May have been a good idea in theory but I just undid it since it caused our bootstrap builds to fail. This also means there's no strong order dependency between merging this and distributing NuGet.props in your packages.

PR #5139, opened 2020-02-22
Don't propagate TargetPath to RAR related files

### LINKED ISSUE #4756: ResolveAssemblyReferences propagates TargetPath metadata to related files
_contributes to internal bug [devdiv#980195](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/980195)_

RAR copies metadata from primary source items to outputs, including the `TargetPath` metadatum, and including to "related" non-primary output files. This means that a C# project that references a C++ WinMD project can get an item that looks like this:

```
C:\source\repos\WapProjTemplate36\Debug\WindowsRuntimeComponent1\WindowsRuntimeComponent1.pdb
    TargetPath = WindowsRuntimeComponent1.winmd
```

And that `TargetPath` may overlap with another real file.

There's already

https://github.com/microsoft/msbuild/blob/ed928d8d06ef21eb5795f6a7cc767d37d31e57bc/src/Tasks/AssemblyDependency/ReferenceTable.cs#L2918-L2926

Possibly just add `TargetPath` to that would solve the problem. But we need to think through the consequences of that.

rainersigwald:
Fixes #4756 by adding TargetPath to the list of metadata that shouldn't
be forwarded to "child"/related files discovered in RAR.

Adds an escape hatch for the behavior since we're not completely sure
there isn't user code that depends on the (surprising) behavior.

### DIFF
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -197,6 +197,11 @@ namespace Microsoft.Build.Utilities
         public readonly bool DisableNuGetSdkResolver = Environment.GetEnvironmentVariable("MSBUILDDISABLENUGETSDKRESOLVER") == "1";
 
         /// <summary>
+        /// Don't delete TargetPath metadata from associated files found by RAR.
+        /// </summary>
+        public readonly bool TargetPathForRelatedFiles = Environment.GetEnvironmentVariable("MSBUILDTARGETPATHFORRELATEDFILES") == "1";
+
+        /// <summary>
         /// Disable AssemblyLoadContext isolation for plugins.
         /// </summary>
         public readonly bool UseSingleLoadContext = Environment.GetEnvironmentVariable("MSBUILDSINGLELOADCONTEXT") == "1";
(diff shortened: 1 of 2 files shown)


rainersigwald:
@rainersigwald validate that this fixes the problem, please.

src/Shared/Traits.cs
```diff
@@ -196,6 +196,11 @@ public bool LogTaskInputs
         /// </summary>
         public readonly bool DisableNuGetSdkResolver = Environment.GetEnvironmentVariable("MSBUILDDISABLENUGETSDKRESOLVER") == "1";
 
+        /// <summary>
+        /// Don't delete TargetPath metadata from associated files found by RAR.
+        /// </summary>
+        public readonly bool TargetPathForRelatedFiles = Environment.GetEnvironmentVariable("MSBUILDTARGETPATHFORRELATEDFILES") == "1";
```
Forgind: This feels a little like a double-negative; maybe change this to `...== "";` here and remove the `!` in ReferenceTable.cs?

rainersigwald: I was trying to avoid a double negative in the name.

PR #5142, opened 2020-02-24
Revert slnf feature (and associated System.Text.Json dependency)

Forgind:


davkean:
Why was this reverted?

Forgind:
VS hadn't had a dependency on System.Text.Json previously, and I was having some trouble with bringing it in properly, since it also required updating some other assemblies to more recent versions. [The latest build](https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/230232?_a=overview) looks hopeful, but we have things we really want in preview 1 , and this was making it hard.

It will hopefully come back in soon.

PR #5144, opened 2020-02-24
Do not respect NodeReuse for TaskHosts Fixes #3141

### LINKED ISSUE #3141: Node reuse locks task assembly file
We have build scripts that produce tasks as a local assembly. These assembly files are locked by the nodes, which means we can't use nodeReuse on repos that use custom local tasks.

### Steps to reproduce

1. Git clone https://github.com/aspnet/BuildTools/tree/6e2c86240ffd0056d202eb8ede1fba1e251804b4
2. `build.cmd /p:SkipTests=true`
3. `build.cmd /p:SkipTests=true`

### Expected  behavior
It would be great if nodeReuse was savvy enough to handle changes to task assemblies on disk.

### Actual behavior
```
C:\Users\namc\.dotnet\x64\sdk\2.1.300-preview2-008396\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(137,5): warning MSB3026: Could not copy "obj\Release\netcoreapp2.0\RepoTasks.dll" to "C:\dev\aspnet\BuildTools\build\tasks\bin\publish\RepoTasks.dll". Beginning retry 1 in 1000ms. The process cannot access the file 'C:\dev\aspnet\BuildTools\build\tasks\bin\publish\RepoTasks.dll' because it is being used by another process.  [C:\dev\aspnet\BuildTools\build\tasks\RepoTasks.csproj]
```

The process locking the file is
```
C:\Users\namc\.dotnet\x64\dotnet.exe "C:\Users\namc\.dotnet\x64\sdk\2.1.300-preview2-008396\MSBuild.dll" C:\Users\namc\.dotnet\x64\sdk\2.1.300-preview2-008396\MSBuild.dll /nologo /nodemode:1 /nodeReuse:true
```

### Environment data
```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.300-preview2-008396
 Commit:    400e670faa

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Users\namc\.dotnet\x64\sdk\2.1.300-preview2-008396\
```

## Workaround
```
build.cmd /nodeReuse:false
```

Forgind:
They can lock the node, causing errors.

Resolves #3141.


src/MSBuild/OutOfProcTaskHostNode.cs
```diff
@@ -693,11 +693,11 @@ private void CancelTask()
         /// <summary>
         /// Handles the NodeBuildComplete packet.
         /// </summary>
-        private void HandleNodeBuildComplete(NodeBuildComplete buildComplete)
+        private void HandleNodeBuildComplete()
         {
             ErrorUtilities.VerifyThrow(!_isTaskExecuting, "We should never have a task in the process of executing when we receive NodeBuildComplete.");
 
-            _shutdownReason = buildComplete.PrepareForReuse ? NodeEngineShutdownReason.BuildCompleteReuse : NodeEngineShutdownReason.BuildComplete;
+            _shutdownReason = NodeEngineShutdownReason.BuildComplete;
```
rainersigwald: Worth a comment that there's no reuse scenario here because ...

cdmihai:
Wouldn't this slow down people whose builds call into out of proc task nodes many times? An escape hatch might be useful. I guess it depends which population is larger, the repos with frequently called out of proc tasks, or the repos that build their own tasks. Probably the later ...

Forgind:
I could add an escape hatch. I also considered trying to detect if the relevant task was a custom task or MSBuild task, but it got a little too complicated.

src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
```diff
@@ -21,6 +22,29 @@ public WarningsAsMessagesAndErrorsTests(ITestOutputHelper output)
...
+                string introToPID = "PID to shut down is ";
+                string afterPID = "(EndPID)";
+                string pidTaskProject = $@"<Project>
+    <UsingTask TaskName=""ProcessIdTask"" AssemblyName=""Microsoft.Build.Engine.UnitTests"" TaskFactory=""TaskHostFactory"" />
+    <Target Name='AccessPID'>
+        <ProcessIdTask />
+    </Target>
+      </Project>";
+                TransientTestFile project = env.CreateFile("testProject.csproj", pidTaskProject);
+                MockLogger logger = new MockLogger();
```
rainersigwald:
```suggestion
                MockLogger logger = new MockLogger(_output);
```

src/Shared/Traits.cs
```diff
@@ -163,6 +163,11 @@ public bool LogTaskInputs
         /// </summary>
         public readonly bool UseSymlinkTimeInsteadOfTargetTime = Environment.GetEnvironmentVariable("MSBUILDUSESYMLINKTIMESTAMP") == "1";
 
+        /// <summary>
+        /// Allow node reuse on TaskHost nodes. This breaks if there are custom tasks.
```
rainersigwald:
```suggestion
        /// Allow node reuse of TaskHost nodes. This results in task assemblies locked past the build lifetime, but may improve performance.
```

src/Build.UnitTests/WarningsAsMessagesAndErrors_Tests.cs
```diff
@@ -21,6 +22,30 @@ public WarningsAsMessagesAndErrorsTests(ITestOutputHelper output)
...
+    <Target Name='AccessPID'>
+        <ProcessIdTask />
+    </Target>
+      </Project>";
+                TransientTestFile project = env.CreateFile("testProject.csproj", pidTaskProject);
+                MockLogger logger = new MockLogger(_output);
+                ObjectModelHelpers.BuildTempProjectFileWithTargetsExpectSuccess(project.Path, null, null, logger);
+                int start = logger.FullLog.IndexOf(introToPID) + introToPID.Length;
+                int length = logger.FullLog.IndexOf(afterPID) - start;
+                string pid = logger.FullLog.Substring(start, length);
```
rainersigwald: Instead of grepping through the logging, can you change the test to return the PID of its process as a property, capture the property into a named property when you call it, and read the property out of the built project instance? That's a much cleaner way to do things (and a good exercise for you in using our API).

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
rainersigwald: copyright header please

src/Build.UnitTests/BackEnd/ProcessIdTask.cs
```diff
@@ -11,13 +12,16 @@ namespace Microsoft.Build.UnitTests
     /// </summary>
     public class ProcessIdTask : Task
     {
+        [Output]
+        public Int32 pid { get; set; }
```
rainersigwald:
```suggestion
        public int Pid { get; set; }
```

style nits: use the C# type where possible, and properties should start with a capital letter.

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
```diff
@@ -0,0 +1,38 @@
...
+using Shouldly;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Build.Engine.UnitTests.BackEnd
+{
+    public sealed class TaskHostFactory_Tests
+    {
+        [Fact]
+        [SkipOnTargetFramework(TargetFrameworkMonikers.Netcoreapp, "https://github.com/microsoft/msbuild/issues/5158")]
```
rainersigwald: Does this work on Mono?

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
```diff
@@ -0,0 +1,38 @@
...
+    <Target Name='AccessPID'>
+        <ProcessIdTask>
+            <Output PropertyName=""PID"" TaskParameter=""pid"" />
+        </ProcessIdTask>
+    </Target>
+</Project>";
+                TransientTestFile project = env.CreateFile("testProject.csproj", pidTaskProject);
+                ProjectInstance projectInstance = new ProjectInstance(project.Path);
+                projectInstance.Build().ShouldBeTrue();
+                Should.Throw<ArgumentException>(() => Process.GetProcessById(Int32.Parse(projectInstance.GetPropertyValue("PID"))));
```
rainersigwald: I'd split this up to help with diagnosability when it fails (as it's doing now).

```suggestion
                string pidProperty = projectInstance.GetPropertyValue("PID");
                string.IsNullOrEmpty(pidProperty).ShouldBeFalse();

                // also log the PID that we're looking for

                int.TryParse(pidProperty, out int pid).ShouldBeTrue();

                Should.Throw<ArgumentException>(() => Process.GetProcessById(pid));
```

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
Forgind: Does Mono support out-of-proc task hosts? I was going to find out with the automated tests.

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
```diff
@@ -36,8 +36,15 @@ public void TaskNodesDieAfterBuild()
                 string.IsNullOrEmpty(processId).ShouldBeFalse();
                 Int32.TryParse(processId, out int pid).ShouldBeTrue();
                 Process.GetCurrentProcess().Id.ShouldNotBe<int>(pid);
-                Thread.Sleep(1000);
-                Should.Throw<ArgumentException>(() => Process.GetProcessById(pid));
+                try {
+                    Process taskHostNode = Process.GetProcessById(pid);
+                    taskHostNode.WaitForExit(2000);
```
rainersigwald: Assert that this reported that the process ended (and didn't time out).

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
```diff
@@ -36,8 +36,15 @@ public void TaskNodesDieAfterBuild()
...
+                    Should.Throw<ArgumentException>(() => Process.GetProcessById(pid));
+                }
+                catch (Exception e)
+                {
+                    e.ShouldBeOfType<ArgumentException>();
+                }
```
rainersigwald: Why this catch?

Forgind: We want taskHostNode to die. If it dies before this try statement, `Process.GetProcessById` will throw an ArgumentException, so this test succeeded if either the `Should.Throw` statement succeeds or the node dies fast, and `Process.GetProcessById` throws an ArgumentException.

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
Forgind: That's caught by the Should.Throw immediately afterwards. If it times out, `Process.GetProcessById` won't throw, and the test will fail.

rainersigwald: Not necessarily though, given PID reuse.

src/Build.UnitTests/BackEnd/TaskHostFactory_Tests.cs
rainersigwald: Comment with that reasoning, catch only the exact exception, minimize the scope of the catch, and can you assert details of it?

Forgind:
Checked with @rainersigwald, and this does not work on mono, so disabling the test seems reasonable.

PR #5152, opened 2020-03-02
Find amd64/MSBuild.exe when run from a 64-bit process

### LINKED ISSUE #5151: 64-bit host processes can't build multiproc using API
### Steps to reproduce

[amd64-msbuild-client.zip](https://github.com/microsoft/msbuild/files/4277239/amd64-msbuild-client.zip)

`dotnet run` is sufficient.

### Expected behavior

Success building the trivial project across multiple nodes.

### Actual behavior

Success building but on only one node, followed by

```
Unhandled Exception: Microsoft.Build.Exceptions.BuildAbortedException: Build was canceled. Failed to successfully launch or connect to a child MSBuild.exe process. Verify that the MSBuild.exe "C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Current\Bin\MSBuild.exe" launches successfully, and that it is loading the same microsoft.build.dll that the launching process loaded. If the location seems incorrect, try specifying the correct location in the BuildParameters object, or with the MSBUILD_EXE_PATH environment variable.
   at Microsoft.Build.BackEnd.NodeProviderOutOfProc.CreateNode(Int32 nodeId, INodePacketFactory factory, NodeConfiguration configuration)
   at Microsoft.Build.BackEnd.NodeManager.AttemptCreateNode(INodeProvider nodeProvider, NodeConfiguration nodeConfiguration)
   at Microsoft.Build.BackEnd.NodeManager.CreateNode(NodeConfiguration configuration, NodeAffinity nodeAffinity)          at Microsoft.Build.Execution.BuildManager.PerformSchedulingActions(IEnumerable`1 responses)
   at Microsoft.Build.Execution.BuildManager.ProcessPacket(Int32 node, INodePacket packet)
   at Microsoft.Build.Execution.BuildManager.ProcessWorkQueue(Action action)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Build.Execution.BuildManager.EndBuild()
   at Microsoft.Build.Execution.BuildManager.Build(BuildParameters parameters, BuildRequestData requestData)
   at amd64_msbuild_client.Program.Build() in S:\play\amd64-msbuild-client\Program.cs:line 34
   at amd64_msbuild_client.Program.Main(String[] args) in S:\play\amd64-msbuild-client\Program.cs:line 18
```

### Environment data
`msbuild /version` output: `16.6.0-preview-20126-02+13cfe7fc5`

OS info: Windows 10 amd64

rainersigwald:
Fixes #5151 by inserting amd64 into the path if run from a 64-bit process
and discovering MSBuild in a Visual Studio installation.

### DIFF
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -334,13 +334,7 @@ namespace Microsoft.Build.Shared
                 "MSBuild",
                 CurrentToolsVersion,
                 "Bin",
-#if NET35
-                IntPtr.Size == 8
-#else
-                System.Runtime.InteropServices.Marshal.SizeOf<IntPtr>() == 8
-#endif
-                  ? "amd64"
-                  : string.Empty,
+                IntPtr.Size == 8 ? "amd64" : string.Empty,
                 "MSBuild.exe");
         }
 
--- a/src/Shared/BuildEnvironmentHelper.cs
+++ b/src/Shared/BuildEnvironmentHelper.cs
@@ -329,7 +329,19 @@ namespace Microsoft.Build.Shared
 
         private static string GetMSBuildExeFromVsRoot(string visualStudioRoot)
         {
-            return FileUtilities.CombinePaths(visualStudioRoot, "MSBuild", CurrentToolsVersion, "Bin", "MSBuild.exe");
+            return FileUtilities.CombinePaths(
+                visualStudioRoot,
+                "MSBuild",
+                CurrentToolsVersion,
+                "Bin",
+#if NET35
+                IntPtr.Size == 8
+#else
+                System.Runtime.InteropServices.Marshal.SizeOf<IntPtr>() == 8
+#endif
+                  ? "amd64"
+                  : string.Empty,
+                "MSBuild.exe");
         }
 
         private static bool? _runningTests;


src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -329,7 +329,19 @@ private static string GetVsRootFromMSBuildAssembly(string msBuildAssembly)
...
+            return FileUtilities.CombinePaths(
+                visualStudioRoot,
+                "MSBuild",
+                CurrentToolsVersion,
+                "Bin",
+#if NET35
+                IntPtr.Size == 8
+#else
+                System.Runtime.InteropServices.Marshal.SizeOf<IntPtr>() == 8
+#endif
```
rainersigwald: We have this, but it's not easily accessible in this file and uses .NET 4+ features in both the interesting function and other functions in the file that make it very annoying to use.

https://github.com/microsoft/msbuild/blob/13cfe7fc5a97772e4045671c71fa31a3c0856af0/src/Shared/EnvironmentUtilities.cs#L11

So I duplicated the code and included a net35-compatible implementation.

cdmihai:
Shouldn't all the other MSBuild.exe searches do this, not just VS?

rainersigwald:
I think so, but we also run into https://github.com/microsoft/msbuild/issues/1941, which I'm leaving open for now. This unblocks the most critical scenario when the calling code uses Locator.

src/Shared/BuildEnvironmentHelper.cs
Forgind: `IntPtr.Size == 8` would work for new versions or older versions, right? Could we just use that version instead?

src/Shared/BuildEnvironmentHelper.cs
```diff
@@ -329,7 +329,19 @@ private static string GetVsRootFromMSBuildAssembly(string msBuildAssembly)
...
+                  ? "amd64"
+                  : string.Empty,
```
Forgind: I imagine this will create a double-slash, but I don't think there's a way around that that improves performance.

cdmihai: Probably doing two if-then-else Combine calls depending on bitness would be faster than the unfathomable machinery triggered in .net and windows by the two `\\`

rainersigwald: I was curious about this as I implemented it. It actually does The Right Thing:

```sh-session
$ type .\Program.cs
using System;

namespace path_thing
{
    class Program
    {
        {
            Console.WriteLine(System.IO.Path.Combine("a", "b", string.Empty, "c"));
        }
    }
}

/play/path-thing via •NET v3.1.200-preview-015002
$ dotnet run
a\b\c
```

src/Shared/BuildEnvironmentHelper.cs
rainersigwald: that's fair and cleaner-looking

src/Shared/BuildEnvironmentHelper.cs
cdmihai: Curious, in CSI I get:

```c#
> System.IO.Path.Combine("a", string.Empty, "b")
"a\\b"
```

cdmihai: Ah, CSI pretty prints the string and escapes the slash, which I judge as a very dubious choice

```c#
> var a = System.IO.Path.Combine("a", "", "b");
> foreach (char i in a) System.Console.WriteLine(i);
a
\
b
```

PR #5165, opened 2020-03-06
Remove FEATURE_VARIOUS_EXCEPTIONS

rainersigwald:
While working on something else I noticed this old (.NET Core 1) feature flag that we don't need any more since we have the guarded APIs everywhere now.

### DIFF
--- a/eng/Packages.props
+++ b/eng/Packages.props
@@ -37,7 +37,8 @@
     <PackageReference Update="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
     <PackageReference Update="System.Runtime.Loader" Version="4.0.0" />
     <PackageReference Update="System.Runtime.Serialization.Primitives" Version="4.1.1" />
-    <PackageReference Update="System.Security.Principal.Windows" Version="4.3.0" />
+    <PackageReference Update="System.Security.Permissions" Version="4.7.0" />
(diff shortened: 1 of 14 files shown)


eng/Packages.props
```diff
@@ -37,7 +37,8 @@
     <PackageReference Update="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
     <PackageReference Update="System.Runtime.Loader" Version="4.0.0" />
     <PackageReference Update="System.Runtime.Serialization.Primitives" Version="4.1.1" />
-    <PackageReference Update="System.Security.Principal.Windows" Version="4.3.0" />
+    <PackageReference Update="System.Security.Permissions" Version="4.7.0" />
```
rainersigwald: Needed this to get XmlSyntaxException on Core.

eng/Packages.props
```diff
@@ -37,7 +37,8 @@
...
+    <PackageReference Update="System.Security.Principal.Windows" Version="4.7.0" />
```
rainersigwald: Taking the latest of System.Security.Permissions conflicted with the older version here so I bumped it. It's core-only so shouldn't impact VS.

eng/Packages.props
ericstj: This is CAS are you sure you want this?  Did you examine the impact to your install footprint?

PR #5177, opened 2020-03-12
Remove block and publish tlb both platforms

davidwengier:
I'm still trying to test this locally, but the MSBuild bits seem to work fine and I think the problem now lies in the Xaml markup compiler. Thought I'd put this up early just to confirm, because to say I'm unfamiliar with your build infrastructure is a gross understatement :)

(Hopefully) fixes [AB#1077964](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1077964)

### DIFF
--- a/src/Package/MSBuild.VSSetup/files.swr
+++ b/src/Package/MSBuild.VSSetup/files.swr
@@ -177,7 +177,7 @@ folder InstallDir:\MSBuild\Current\Bin\amd64
   file source=$(X86BinPath)Microsoft.Build.dll vs.file.ngenArchitecture=all
   file source=$(MSBuildConversionBinPath)Microsoft.Build.Engine.dll vs.file.ngenArchitecture=all
   file source=$(X86BinPath)Microsoft.Build.Framework.dll vs.file.ngenArchitecture=all
-  file source=$(X86BinPath)Microsoft.Build.Framework.tlb
+  file source=$(X64BinPath)Microsoft.Build.Framework.tlb
   file source=$(X86BinPath)Microsoft.Build.Tasks.Core.dll vs.file.ngenArchitecture=all
   file source=$(X86BinPath)Microsoft.Build.Utilities.Core.dll vs.file.ngenArchitecture=all
   file source=$(X86BinPath)System.Buffers.dll vs.file.ngenArchitecture=all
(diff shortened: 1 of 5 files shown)


src/Package/MSBuild.VSSetup/files.swr
```diff
@@ -177,6 +177,7 @@ folder InstallDir:\MSBuild\Current\Bin\amd64
   file source=$(X86BinPath)Microsoft.Build.dll vs.file.ngenArchitecture=all
   file source=$(MSBuildConversionBinPath)Microsoft.Build.Engine.dll vs.file.ngenArchitecture=all
   file source=$(X86BinPath)Microsoft.Build.Framework.dll vs.file.ngenArchitecture=all
+  file source=$(X86BinPath)Microsoft.Build.Framework.tlb
```
rainersigwald:
```suggestion
  file source=$(X64BinPath)Microsoft.Build.Framework.tlb
```

? Or is the `.tlb` not arch-specific?

davidwengier: > Or is the .tlb not arch-specific?

It never even occurred to me that this was possible, but it makes sense, and potentially is a flaw in the current build scripts that the tlb produced will be whatever bitness the machine doing the build is. I'll update this to produce both types and see if there are differences.

davidwengier:
> If it's hard to validate this

I don't think its harder to validate locally than from a full VS build, but will keep the option in mind.

davidwengier:
Thanks @rainersigwald your change looks good on my machine (the .vsix has two tlbs and the one in the amd64 directory is different). I've validated that this works as expected when running the 64bit msbuild when i build locally when used in this build of VS: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3564683&_a=summary&view=results

How do you want to move forward? I think this is ready to merge, but if you want to do a test VS insertion first that's cool.

rainersigwald:
@davidwengier Probably worth trying the insertion just to be sure. I pushed a branch causing https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3569475 to build; when it's complete it should produce an experimental insertion.

davidwengier:
Thanks @rainersigwald I took that insertion, merged it with the other required changes into [a test PR](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/236974) and have validated that the scenarios work :)

davidwengier:
Thanks for the approval @rainersigwald .. don't forget I can't merge this though :D

BenVillalobos:
For myself future generations: What's a .tlb file and what context is it used in?

Otherwise, lgtm!

rainersigwald:
@BenVillalobos thanks for volunteering to be our local COM expert. Start [here](https://docs.microsoft.com/en-us/windows/win32/midl/com-dcom-and-type-libraries). It's all your problem now!

:)

BenVillalobos:
Quick recap on my previous question: `.tlb` files are **Type Library** files. They're used by other applications to determine a COM or DCOM object's properties and methods, primarily (exclusively?) through the interfaces they implement.

PR #5182, opened 2020-03-18
Apply a solution filter (revert revert)

Forgind:
This reverts commit 2e4c1867b08f0fe9ae2591cde6cb73ab304efeff.

Original slnf change was #4899

Revert was #5142. #4899 was reverted because it created a System.Text.Json dependency that has now been corrected.

### DIFF
--- a/src/Build/Construction/Solution/SolutionFile.cs
+++ b/src/Build/Construction/Solution/SolutionFile.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Text;
 using System.Globalization;
 using System.Security;
+using System.Text.Json;
 using System.Text.RegularExpressions;
 
 using ErrorUtilities = Microsoft.Build.Shared.ErrorUtilities;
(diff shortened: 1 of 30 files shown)


src/Build/Construction/Solution/SolutionFile.cs
```diff
@@ -462,6 +509,30 @@ internal void ParseSolution()
...
+                foreach (ProjectInSolution project in _projectsInOrder)
+                {
+                    projectPaths.Add(project.RelativePath);
+                }
+                foreach (string project in _solutionFilter)
+                {
+                    if (!projectPaths.Contains(project))
+                    {
+                        ProjectFileErrorUtilities.VerifyThrowInvalidProjectFile
+                        (
```
danmosemsft, line 524: is this missing the first parameter (false)?

danmosemsft: oops, no idea how I ended up commenting on an old PR :)

Forgind: This theoretically will still go in if System.Text.Json starts working. This PR also specifies what VerifyThrowInvalidProjectFile should do if it's missing the first parameter: assume false.

danmosemsft: Ah - I thought the overloads that assume false were named `ThrowInvalidProjectFile()`

PR #5191, opened 2020-03-23
Do not emit "unlogged error" when a task fails because of calling into another project

### LINKED ISSUE #5179: MSB4132 is emitted from GenerateTemporaryTargetAssembly
XAML projects generate a temporary project and build it inside the target `GenerateTemporaryTargetAssembly`. If this fails, MSBuild now emits

```
:\windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.WinFx.targets(444,8): error MSB4132: The "GenerateTemporaryTargetAssembly" task returned false but did not log an error. [c:\project-
system\src\Microsoft.VisualStudio.Editors\Microsoft.VisualStudio.Editors.vbproj]
```

This should be special-cased like the MSBuild and CallTarget tasks:

https://github.com/microsoft/msbuild/blob/acff60145c49281a35fb4492b08d95b19da87495/src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs#L950-L953

_Ideally_, the engine would detect use of `IBuildEngine.BuildProjectFile`/`IBuildEngine3.BuildProjectFilesInParallel`, and apply the exception to any task that builds other projects that failed. But that might be hard to do right now. We could special-case `GenerateTemporaryTargetAssembly` easily.

Forgind:
When a task builds other tasks, and one of those tasks fails and logs an error, the building task should fail as well, but since the error is already logged, it does not need to log an error itself. This caused the failure message for "this task failed but did not log an error" to erroneously think an error had not been logged. This fixes #5179—that problem.


Forgind:
azp /run

src/Build/BackEnd/Components/RequestBuilder/TaskHost.cs
```diff
@@ -241,6 +241,8 @@ internal bool IsOutOfProc
             }
         }
 
+        public bool ParallelTasksSucceeded { get; private set; } = true;
```
Forgind: This is a bad name, but I had some trouble with it. Please opine.

rainersigwald: `BuildRequestFailed`?

src/Build/BackEnd/Components/RequestBuilder/TaskBuilder.cs
```diff
@@ -950,7 +947,8 @@ private async Task<WorkUnitResult> ExecuteInstantiatedTask(ITaskExecutionHost ta
                 // When a task fails it must log an error. If a task fails to do so,
                 // that is logged as an error. MSBuild tasks are an exception because
                 // errors are not logged directly from them, but the tasks spawned by them.
-                if (!isMSBuildTaskOrCallTargetTask && taskReturned && !taskResult && !taskLoggingContext.HasLoggedErrors)
+                IBuildEngine be = host.TaskInstance.BuildEngine;
+                if (taskReturned && !taskResult && !taskLoggingContext.HasLoggedErrors && (be is TaskHost ? ((TaskHost)be).BuildRequestsSucceeded : false))
```
rainersigwald:
```suggestion
                if (taskReturned && !taskResult && !taskLoggingContext.HasLoggedErrors && (be is TaskHost th ? th.BuildRequestsSucceeded : false))
```

src/Build.UnitTests/BackEnd/GenerateTemporaryTargetAssembly_Tests.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.Build.Engine.UnitTests.BackEnd
+{
+    sealed public class GenerateTemporaryTargetAssembly_Tests
+    {
+        [Fact]
+        public void FailsWithOnlyTargetErrors()
+        {
+            MockLogger logger = ObjectModelHelpers.BuildProjectExpectFailure(@"
+<Project>
+    <UsingTask TaskName=""FailingBuilderTask"" AssemblyName=""Microsoft.Build.Engine.UnitTests"" />
```
rainersigwald: Can you get the assembly name via reflection?

src/Build.UnitTests/BackEnd/FailingBuilderTask.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections;
```
rainersigwald: Copyright header

src/Build.UnitTests/BackEnd/FailingBuilderTask.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.Build.UnitTests
+{
+    public  class FailingBuilderTask : Task
+    {
+        public FailingBuilderTask()
+            : base(null)
+        { }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
```
rainersigwald: shouldn't need this

src/Build.UnitTests/BackEnd/FailingBuilderTask.cs
```diff
@@ -0,0 +1,29 @@
...
+        public override bool Execute()
+        {
+            try
+            {
+                BuildEngine.BuildProjectFile(CurrentProject, new string[] { "CoreCompile" }, new Hashtable(), null);
```
rainersigwald: Comment that this calling pattern is like the XAML thing we care about specifically.

src/Build.UnitTests/BackEnd/GenerateTemporaryTargetAssembly_Tests.cs
```diff
@@ -0,0 +1,22 @@
...
+    <Target Name=""MyTarget"">
+        <FailingBuilderTask CurrentProject="".\otherproj.csproj"" />
```
rainersigwald: Can you make the other project a real project that has a target with an error, and assert that the sole error is that error?

src/Build.UnitTests/BackEnd/GenerateTemporaryTargetAssembly_Tests.cs
```diff
@@ -0,0 +1,34 @@
...
+    </Target>
+</Project>");
+                MockLogger logger = ObjectModelHelpers.BuildProjectExpectFailure(@$"
+<Project>
+    <UsingTask TaskName=""FailingBuilderTask"" AssemblyFile=""{Assembly.GetExecutingAssembly().Location}"" />
+    <Target Name=""MyTarget"">
+        <FailingBuilderTask CurrentProject=""{otherproj.Path}"" />
+    </Target>
+</Project>");
+                logger.ErrorCount.ShouldBe(1);
```
rainersigwald: Also assert that the error is the the one you threw in the other project.

Forgind:
I created a WPF app and intentionally introduced a compile-time error in a C# file. I then tried building with a version of MSBuild from before MSB4132 was introduced and found three errors. Building with a version from after it was introduced, MSB4132 was added to those three errors. Building with code from this PR led to three errors—the three from before MSB4132 was introduced.

PR #5195, opened 2020-03-24
Don't swallow AfterTargets failures in multi-proc builds

### LINKED ISSUE #3345: Build reports success when an AfterTargets errors
### Steps to reproduce

1.proj
```xml
<Project DefaultTargets="Build">
    <Target Name="Build">
        <MSBuild Projects="2.proj" Targets="Build"/>
    </Target>
</Project>
```
2.proj
```xml
<Project DefaultTargets="Build">
    <Target Name="Build" />

    <Target Name="IndirectBuild" AfterTargets="Build">
        <Error Text="Induce failure" />
    </Target>
</Project>
```

build with
```
set MSBUILDNOINPROCNODE=1

msbuild.exe 1.proj /clp:ShowCommandLine /m:1 /verbosity:n /nr:false /bl /fl /flp:v=diagnostic
```

### Expected  behavior

Build fails because of the `<Error>` task

### Actual behavior
Build succeeds:

```
Build started 5/24/2018 5:35:56 PM.
Project "e:\delete\repro\1.proj" on node 2 (default targets).
Project "e:\delete\repro\1.proj" (1) is building "e:\delete\repro\2.proj" (2) on node 2 (Build target(s)).
e:\delete\repro\2.proj(5,9): error : Induce failure
Done Building Project "e:\delete\repro\2.proj" (Build target(s)) -- FAILED.

Done Building Project "e:\delete\repro\1.proj" (default targets).

Deferred Messages

  Detailed Build Summary
  ======================


  ============================== Build Hierarchy (IDs represent configurations) =====================================================
  Id                  : Exclusive Time   Total Time   Path (Targets)
  -----------------------------------------------------------------------------------------------------------------------------------
  0                   : 0.101s           0.120s       e:\delete\repro\1.proj ()
  . 1                 : 0.018s           0.018s       e:\delete\repro\2.proj (Build)

  ============================== Node Utilization (IDs represent configurations) ====================================================
  Timestamp:            2        Duration   Cumulative
  -----------------------------------------------------------------------------------------------------------------------------------
  636628053566714538:   0        0.096s     0.096s #
  636628053567674882:   1        0.018s     0.114s
  636628053567855025:   0        0.007s     0.121s
  -----------------------------------------------------------------------------------------------------------------------------------
  Utilization:          100.0    Average Utilization: 100.0

Build succeeded.

"e:\delete\repro\1.proj" (default target) (1) ->
"e:\delete\repro\2.proj" (Build target) (2) ->
(IndirectBuild target) ->
  e:\delete\repro\2.proj(5,9): error : Induce failure

    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.22
```

ladipro:
`BuildResult` had code in one of its constructors specifically to handle AfterTargets failures, making sure that they are propagated to the referenced target. This didn't work in multi-proc builds for multiple reasons:

1) The process orchestrating the build may have had no visibility into targets other than those requested to be executed by the worker process. This means that the `existingResults` argument typically didn't contain results for the targets with AfterTargets specified, meaning it had no way of knowing that they had failed.

2) The process orchestrating the build may not have loaded the project file handled by the worker process. This means that the `additionalTargetsToCheck` argument could have been `null`, again making it impossible for `BuildResult` to detect the dependency.

This change moves the logic of calculating target failures to the worker process, namely to `TargetBuilder`. A new flag `AfterTargetsHaveFailed` is introduced and calculated for each requested target after all targets have executed. This flag is then passed back as part of `TargetResult` and trivially used in `BuildResult`'s `OverallResult`. Note that targets with failed AfterTargets can still succeed in terms of their `ResultCode`, i.e. the failure is kept only in form of this bolt-on bit.

The now unused `GetAfterTargetsForDefaultTargets` is deleted.

Fixes #3345

### DIFF
--- a/src/Build.UnitTests/BackEnd/ResultsCache_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/ResultsCache_Tests.cs
@@ -168,8 +168,8 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
         /// <summary>
         /// If a result had multiple targets associated with it and we only requested some of their 
-        /// results, the returned result should only contain the targets we asked for, BUT the overall 
-        /// status of the result should remain the same.  
+        /// results, the returned result should only contain the targets we asked for, and the overall
+        /// status of the result should reflect the targets we asked for as well.
         /// </summary>
         [Fact]
         public void TestRetrieveSubsetTargetsFromResult()
@@ -182,13 +182,13 @@ namespace Microsoft.Build.UnitTests.BackEnd
             result.AddResultsForTarget("testTarget2", BuildResultUtilities.GetEmptySucceedingTargetResult());
             cache.AddResult(result);
 
-            ResultsCacheResponse response = cache.SatisfyRequest(request, new List<string>(), new List<string>(new string[] { "testTarget2" }), new List<string>(new string[] { "testTarget" }), skippedResultsDoNotCauseCacheMiss: false);
+            ResultsCacheResponse response = cache.SatisfyRequest(request, new List<string>(), new List<string>(new string[] { "testTarget2" }), skippedResultsDoNotCauseCacheMiss: false);
 
             Assert.Equal(ResultsCacheResponseType.Satisfied, response.Type);
 
             Assert.True(AreResultsIdenticalForTarget(result, response.Results, "testTarget2"));
             Assert.False(response.Results.HasResultsForTarget("testTarget"));
-            Assert.Equal(BuildResultCode.Failure, response.Results.OverallResult);
+            Assert.Equal(BuildResultCode.Success, response.Results.OverallResult);
         }
 
         [Fact]
--- a/src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
@@ -16,6 +16,7 @@ using Microsoft.Build.Collections;
 using Microsoft.Build.Evaluation;
 using Microsoft.Build.Execution;
 using Microsoft.Build.Shared;
+using Shouldly;
 
 using ILoggingService = Microsoft.Build.BackEnd.Logging.ILoggingService;
 using ProjectLoggingContext = Microsoft.Build.BackEnd.Logging.ProjectLoggingContext;
@@ -691,6 +692,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -716,6 +718,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "AfterTask" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -744,6 +747,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -786,6 +790,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask", "AfterTask", "Error2" });
+            Assert.False(result.ResultsByTarget["PostBuild"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -811,6 +816,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask", "AfterTask" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -836,6 +842,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask", "AfterTask" });
+            Assert.False(result.ResultsByTarget["Build;Me"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -866,6 +873,53 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask", "AfterTask", "AfterTask2" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
+        }
+
+        /// <summary>
+        /// Test a failing after target
+        /// </summary>
+        [Fact]
+        public void TestAfterTargetsWithFailure()
+        {
+            string projectBody = @"
+<Target Name='Build'>
+    <BuildTask/>
+</Target>
+
+<Target Name='After' AfterTargets='Build'>
+    <ErrorTask1/>
+</Target>
+";
+
+            BuildResult result = BuildSimpleProject(projectBody, new string[] { "Build" }, failTaskNumber: 2 /* Fail on After */);
+            result.ResultsByTarget["Build"].ResultCode.ShouldBe(TargetResultCode.Success);
+            result.ResultsByTarget["Build"].AfterTargetsHaveFailed.ShouldBe(true);
+        }
+
+        /// <summary>
+        /// Test a transitively failing after target
+        /// </summary>
+        [Fact]
+        public void TestAfterTargetsWithTransitiveFailure()
+        {
+            string projectBody = @"
+<Target Name='Build'>
+    <BuildTask/>
+</Target>
+
+<Target Name='After1' AfterTargets='Build'>
+    <BuildTask/>
+</Target>
+
+<Target Name='After2' AfterTargets='After1'>
+    <ErrorTask1/>
+</Target>
+";
+
+            BuildResult result = BuildSimpleProject(projectBody, new string[] { "Build" }, failTaskNumber: 3 /* Fail on After2 */);
+            result.ResultsByTarget["Build"].ResultCode.ShouldBe(TargetResultCode.Success);
+            result.ResultsByTarget["Build"].AfterTargetsHaveFailed.ShouldBe(true);
         }
 
         /// <summary>
@@ -1080,6 +1134,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             BuildResult result = builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
             AssertTaskExecutionOrder(new string[] { "BuildTask", "BeforeErrorTargetTask", "ErrorTargetTask", "AfterErrorTargetTask" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
         }
 
         /// <summary>
@@ -1482,6 +1537,27 @@ namespace Microsoft.Build.UnitTests.BackEnd
         private ProjectLoggingContext GetProjectLoggingContext(BuildRequestEntry entry)
         {
             return new ProjectLoggingContext(new NodeLoggingContext(_host, 1, false), entry, null);
+        }
+
+        /// <summary>
+        /// Builds a project using TargetBuilder and returns the result.
+        /// </summary>
+        /// <param name="projectBody">The project contents.</param>
+        /// <param name="targets">The targets to build.</param>
+        /// <param name="failTaskNumber">The task ordinal to fail on.</param>
+        /// <returns>The result of building the specified project/tasks.</returns>
+        private BuildResult BuildSimpleProject(string projectBody, string[] targets, int failTaskNumber)
+        {
+            ProjectInstance project = CreateTestProject(projectBody);
+
+            MockTaskBuilder taskBuilder = (MockTaskBuilder)_host.GetComponent(BuildComponentType.TaskBuilder);
+            taskBuilder.FailTaskNumber = failTaskNumber;
+
+            TargetBuilder builder = (TargetBuilder)_host.GetComponent(BuildComponentType.TargetBuilder);
+            IConfigCache cache = (IConfigCache)_host.GetComponent(BuildComponentType.ConfigCache);
+            BuildRequestEntry entry = new BuildRequestEntry(CreateNewBuildRequest(1, targets), cache[1]);
+
+            return builder.BuildTargets(GetProjectLoggingContext(entry), entry, this, entry.Request.Targets.ToArray(), CreateStandardLookup(project), CancellationToken.None).Result;
         }
 
         /// <summary>
--- a/src/Build/BackEnd/Components/Caching/ResultsCacheWithOverride.cs
+++ b/src/Build/BackEnd/Components/Caching/ResultsCacheWithOverride.cs
@@ -77,14 +77,12 @@ namespace Microsoft.Build.Execution
             BuildRequest request,
             List<string> configInitialTargets,
             List<string> configDefaultTargets,
-            List<string> additionalTargetsToCheckForOverallResult,
             bool skippedResultsDoNotCauseCacheMiss)
         {
             var overrideRequest = _override.SatisfyRequest(
                 request,
                 configInitialTargets,
                 configDefaultTargets,
-                additionalTargetsToCheckForOverallResult,
                 skippedResultsDoNotCauseCacheMiss);
 
             if (overrideRequest.Type == ResultsCacheResponseType.Satisfied)
@@ -95,7 +93,6 @@ namespace Microsoft.Build.Execution
                         request,
                         configInitialTargets,
                         configDefaultTargets,
-                        additionalTargetsToCheckForOverallResult,
                         skippedResultsDoNotCauseCacheMiss)
                         .Type == ResultsCacheResponseType.NotSatisfied,
                     "caches should not overlap");
@@ -107,7 +104,6 @@ namespace Microsoft.Build.Execution
                 request,
                 configInitialTargets,
                 configDefaultTargets,
-                additionalTargetsToCheckForOverallResult,
                 skippedResultsDoNotCauseCacheMiss);
         }
 
--- a/src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
@@ -181,6 +181,7 @@ namespace Microsoft.Build.BackEnd
             }
 
             // Gather up outputs for the requested targets and return those.  All of our information should be in the base lookup now.
+            ComputeAfterTargetFailures(targetNames);
             BuildResult resultsToReport = new BuildResult(_buildResult, targetNames);
 
             // Return after-build project state if requested.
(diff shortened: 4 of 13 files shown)


ladipro:
cc @donJoseLuis who's also looked into this issue.

src/Build.UnitTests/BackEnd/ResultsCache_Tests.cs
```diff
@@ -168,8 +168,8 @@ public void TestRetrieveSubsetResults()
 
         /// <summary>
         /// If a result had multiple targets associated with it and we only requested some of their 
-        /// results, the returned result should only contain the targets we asked for, BUT the overall 
-        /// status of the result should remain the same.
```
ladipro: I hope this comment was wrong. As far as I can tell the idea behind `OverallResult` was exactly the opposite, i.e. correctly calculating results for subsets of targets.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+                    targetsToCheckForAfterTargets.Enqueue(targetName);
+
+                    while (targetsToCheckForAfterTargets.Count > 0)
+                    {
+                        string targetToCheck = targetsToCheckForAfterTargets.Dequeue();
+                        IList<TargetSpecification> targetsWhichRunAfter = _requestEntry.RequestConfiguration.Project.GetTargetsWhichRunAfter(targetToCheck);
+
+                        foreach (TargetSpecification afterTarget in targetsWhichRunAfter)
+                        {
+                            if (!resultsWithAfterTargets.TryGetValue(afterTarget.TargetName, out bool afterTargetFailed))
```
donJoseLuis: minor: should we log anything if `TryGetValue` returns true?

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+                            {
+                                _buildResult.ResultsByTarget.TryGetValue(afterTarget.TargetName, out TargetResult result);
```
donJoseLuis: minor: is there any value to assessing the outcome of `_buildResult.ResultsByTarget.TryGetValue(` ?

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+                                afterTargetFailed = result != null && result.ResultCode == TargetResultCode.Failure && !result.TargetFailureDoesntCauseBuildFailure;
+
+                                resultsWithAfterTargets[afterTarget.TargetName] = afterTargetFailed;
+                                if (afterTargetFailed)
+                                {
+                                    _buildResult.ResultsByTarget[targetName].AfterTargetsHaveFailed = true;
+                                }
+
+                                // We haven't seen this target yet, add it to the list to check.
+                                targetsToCheckForAfterTargets.Enqueue(afterTarget.TargetName);
```
donJoseLuis: I'm probably not 100% sure about the flow, as I see a potential never ending `while`: if we enqueue items inside the body of the loop, on the same queueue we're monitoring to exit the loop...how do we protect against infinite spinning? 

Due diligence would be for me to clone the branch and step through.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
 
             return false;
         }
+
+        private void ComputeAfterTargetFailures(string[] targetNames)
+        {
```
donJoseLuis: could `targetNames` be null?

src/Build/BackEnd/Shared/TargetResult.cs
```diff
@@ -141,6 +146,18 @@ internal bool TargetFailureDoesntCauseBuildFailure
             set => _targetFailureDoesntCauseBuildFailure = value;
         }
 
+        /// <summary>
+        /// Sets or gets a flag indicating whether at least one target which has run after us (transitively via AfterTargets) failed.
+        /// </summary>
+        internal bool AfterTargetsHaveFailed
```
donJoseLuis: why is a field needed behind this property?  `internal bool AfterTargetsHaveFailed { get; set; }`

donJoseLuis: I see: a field is needed, as the value is passed by ref later in the code

src/Build/BackEnd/Components/Caching/ResultsCacheWithOverride.cs
```diff
@@ -77,14 +77,12 @@ public BuildResult GetResultsForConfiguration(int configurationId)
             BuildRequest request,
             List<string> configInitialTargets,
             List<string> configDefaultTargets,
-            List<string> additionalTargetsToCheckForOverallResult,
             bool skippedResultsDoNotCauseCacheMiss)
         {
             var overrideRequest = _override.SatisfyRequest(
```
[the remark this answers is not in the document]
ladipro: This project actually has a well-defined coding style (linked from the docs, although the link is broken -- will fix that momentarily). This is what it has to say for `var`:

`
We only use var when it's obvious what the variable type is (e.g. var stream = new FileStream(...) not var stream = OpenStandardInput()).
`

I'll go over my changes and make sure I'm following this. Thanks!

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+            // Cache intermediate results so we don't visit a target more than once.
+            Dictionary<string, bool> resultsWithAfterTargets = new Dictionary<string, bool>();
+
+            foreach (string targetName in targetNames)
```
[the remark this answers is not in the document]
ladipro: I'd personally vote for common sense rather than strict rules based on such things as loop depth and cyclomatic complexity. In this particular case I do see methods with 3 nested loops in the code base so I would assume it's ok.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: No, I don't believe that this is a log-worthy event. We're simply caching the results in a local dictionary to avoid getting into the O(n^2) run-time territory. A cache hit/miss is pure implementation and not interesting outside of this method.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: Yes, the outcome is assessed on the following line with `result != null`. We want to make this defensive check anyway and can take advantage of the `TryGetValue` contract which says that if the key is not found, the "default value for the type of the value parameter" is returned. TValue is a reference type so the its default value is `null`. Two birds with one stone, if you will.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: I would also not be 100% sure, despite having stepped through the code. I do have a couple of good reasons to believe that this program eventually terminates, though:

1) We are operating on an internal representation of the build files which, presumably, has validation logic in place to make sure that there are no cycles in the task graph. Handling cycles later, like in this case when actually executing the tasks, would be very problematic.

2) If you squint hard enough, an analogous loop exists in `ProcessTargetStack` in this same file.

3) Pretty much the same loop exists in `GetAfterTargetsForDefaultTargets` which I am deleting as part of this PR.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: No, it is treated as non-null by the caller and since this method is a private helper I don't think arguments need to be validated. Looking at other helpers in this file, for example `PushTargets`, I don't see such a validation either.

src/Build/BackEnd/Components/Caching/ResultsCacheWithOverride.cs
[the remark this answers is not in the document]
ladipro: I have replaced two concrete types with `var` in accordance with the coding guidelines and to match what `GetAfterTargetsForDefaultTargets` had.

Also added a commit fixing the broken doc links. Hope we're not too crazy about strictly keeping such changes in separate PRs.

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
```diff
@@ -866,6 +872,71 @@ public void TestAfterTargetsWithTwoReferringToOne()
...
             AssertTaskExecutionOrder(new string[] { "BuildTask", "AfterTask", "AfterTask2" });
+            Assert.False(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
+        }
+
+        /// <summary>
+        /// Test a failing after target
+        /// </summary>
+        [Fact]
+        public void TestAfterTargetsWithFailure()
+        {
```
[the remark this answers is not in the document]
ladipro: Thank you, yes, there's a lot of code duplication in this file. I have added a helper method and rewrote the two new tests to use it. It should be possible to have a bunch of existing tests converted to use the helper as well but better leave it for another change to make reviewing of this PR easier.

ladipro:
Thank you for the thorough review @donJoseLuis.

> 1. we should ensure all unit test cases pass, though we know 75 do not succeed when run in VS

Totally. Although getting all-green UT locally seems unrelated to this change. Let's sync up on that offline.

src/Build/BackEnd/Components/Caching/ResultsCacheWithOverride.cs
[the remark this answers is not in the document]
Forgind: Consistent coding practice is good; the problem is determining what that coding practice should be. cdmihai is in favor of var even when it's very not obvious to me because it forces him to use better variable names. I'm in favor of concrete types everywhere with very rare exceptions when we can't figure out the type until runtime. It's clearer if you want to know what type something is without looking at more than a single spot, and it helps with VS's "go to definition," which is sometimes confused by imprecision and takes me somewhere useless instead of what I want.
I suggested talking about this at some point at a time when it was bothering me regularly, but we never had it, and it isn't bothering me as often right now. I'd still be happy to have such a discussion.

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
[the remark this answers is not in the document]
Forgind: You can also put the asserts in the other method. Alternatively, you can squish the two together in a theory. The only differences seem to be the project files and 2 vs. 3.

For asserts, please you shouldly—`TargetResultCode.Success.ShouldBe(result.ResultsByTarget["Build"].ResultCode`, for instance. It makes debugging easier if this test breaks in the future.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
Forgind: Targets definitely cycle-detection logic. I'm not sure about tasks, but I'm also having trouble making tasks call each other, since there isn't a nice way to do that, as far as I know.

src/Build.UnitTests/BackEnd/ResultsCache_Tests.cs
Forgind: I'm not confident of this answer, but I think both versions of the comment are saying the same (correct) thing: if you only request some results, the Overall status reflects the results of all the targets, both those we asked for and those we didn't.

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
```diff
@@ -1484,6 +1538,30 @@ private ProjectLoggingContext GetProjectLoggingContext(BuildRequestEntry entry)
...
+        /// </summary>
+        /// <param name="projectBody">The project contents.</param>
+        /// <param name="targets">The targets to build.</param>
+        /// <param name="failTaskNumber">The task ordinal to fail on, or null if no task should fail.</param>
+        /// <returns>The result of building the specified project/tasks.</returns>
+        private BuildResult BuildSimpleProject(string projectBody, string[] targets, int? failTaskNumber = null)
+        {
+            ProjectInstance project = CreateTestProject(projectBody);
+
+            if (failTaskNumber.HasValue)
```
Forgind: Does it ever not have a value? Could you just make it non-nullable?

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+                    {
+                        string targetToCheck = targetsToCheckForAfterTargets.Dequeue();
+                        IList<TargetSpecification> targetsWhichRunAfter = _requestEntry.RequestConfiguration.Project.GetTargetsWhichRunAfter(targetToCheck);
+
+                        foreach (TargetSpecification afterTarget in targetsWhichRunAfter)
+                        {
+                            if (!resultsWithAfterTargets.TryGetValue(afterTarget.TargetName, out bool afterTargetFailed))
+                            {
+                                _buildResult.ResultsByTarget.TryGetValue(afterTarget.TargetName, out TargetResult result);
+                                afterTargetFailed = result != null && result.ResultCode == TargetResultCode.Failure && !result.TargetFailureDoesntCauseBuildFailure;
```
Forgind: nit:
```suggestion
                                afterTargetFailed = result?.ResultCode == TargetResultCode.Failure && !result.TargetFailureDoesntCauseBuildFailure;
```

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+
+                                resultsWithAfterTargets[afterTarget.TargetName] = afterTargetFailed;
```
Forgind: Move this into if statement?

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
Forgind: What I'm more wondering about is how it gets the value if the aftertarget hasn't run yet. It seems to use `_buildResult.ResultsByTarget.TryGetValue` to get results, presumably from other nodes that are building the aftertargets, but does that work in a single proc build?

src/Build/BackEnd/Shared/BuildResult.cs
```diff
@@ -191,7 +191,7 @@ internal BuildResult(BuildRequest request, BuildResult existingResults, Exceptio
         /// <param name="targetNames">The list of target names that are the subset of results that should be returned.</param>
         /// <param name="additionalTargetsToCheck">The additional targets that need to be taken into account when computing the overall result, if any.</param>
         /// <param name="exception">The exception, if any</param>
-        internal BuildResult(BuildRequest request, BuildResult existingResults, string[] targetNames, List<string> additionalTargetsToCheck, Exception exception)
+        internal BuildResult(BuildRequest request, BuildResult existingResults, string[] targetNames, Exception exception)
```
cdmihai: Remove the parameter from docs.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,45 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+                                if (afterTargetFailed)
+                                {
+                                    _buildResult.ResultsByTarget[targetName].AfterTargetsHaveFailed = true;
```
cdmihai: As an optimization, if you found one failing after target, why not exit the while loop prematurely?

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
```diff
@@ -866,6 +872,53 @@ public void TestAfterTargetsWithTwoReferringToOne()
...
+            BuildResult result = BuildSimpleProject(projectBody, new string[] { "Build" }, failTaskNumber: 2 /* Fail on After */);
+            Assert.Equal(TargetResultCode.Success, result.ResultsByTarget["Build"].ResultCode);
+            Assert.True(result.ResultsByTarget["Build"].AfterTargetsHaveFailed);
+        }
+
+        /// <summary>
+        /// Test a transitively failing after target
+        /// </summary>
+        [Fact]
+        public void TestAfterTargetsWithTransitiveFailure()
```
cdmihai: Could you also add an equivalent test but which runs the build in an out of proc node? You can use [BuildManagerSession](https://github.com/microsoft/msbuild/blob/a1a53cb88da5d05eb854f02be5ee19edd0c2f399/src/Shared/UnitTests/ObjectModelHelpers.cs#L1863) and send it `BuildParameters.DisableInprocNode` to force builds to be out of proc.

ladipro: Turns out such tests already exist in BuildManager_Tests. I have tweaked them to also run with `BuildParameters.DisableInprocNode`. And verified that they were failing without the changes in this PR and with the flag set.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: Great point, done!

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: Originally I wanted to cache negative results as well. But you are bringing up a really good point, the logic as implemented was not correct. To really cache all results, the search would have to be depth-first so that we really have a valid result for each intermediate target. It's probably not worth adding more complexity given the expected number of targets. I have removed the cache. Thank you!

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: Done, thank you.

src/Build/BackEnd/Shared/BuildResult.cs
ladipro: Done, thank you.

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
ladipro: I wanted to encourage reuse of this method by tests that do not require a task failure. Fixed as you're suggesting to avoid confusion.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: Reading the logic in this file and `ProcessTargetStack` specifically, it looks like we are guaranteed that aftertargets have run by the time we get to this point, and that `_buildResult` contains their results.

For what it's worth I see it working in single proc build, both in unit tests and using the repro project file in #3345. Curious to learn more about your concern.

src/Build.UnitTests/BackEnd/TargetBuilder_Tests.cs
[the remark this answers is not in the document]
ladipro: I've kept the methods separate -- the amount of code is so small that I think the readability hit from converting it to a theory would overweight the code reuse.

I have converted the asserts to `Shoudly`, although my understanding is that the statement should look like this:

`result.ResultsByTarget["Build"].ResultCode.ShouldBe(TargetResultCode.Success);`

i.e. the expression that *should* have the expected value is what `Should` is invoked on and the expected value is the argument. Please correct me if I got it wrong.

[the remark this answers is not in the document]
Forgind: Nope! You're right.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -766,5 +767,37 @@ private async Task<bool> CompleteOutstandingActiveRequests(string targetName)
...
+                    {
+                        string targetToCheck = targetsToCheckForAfterTargets.Dequeue();
+                        IList<TargetSpecification> targetsWhichRunAfter = _requestEntry.RequestConfiguration.Project.GetTargetsWhichRunAfter(targetToCheck);
+
+                        foreach (TargetSpecification afterTarget in targetsWhichRunAfter)
+                        {
+                            _buildResult.ResultsByTarget.TryGetValue(afterTarget.TargetName, out TargetResult result);
+                            if (result?.ResultCode == TargetResultCode.Failure && !result.TargetFailureDoesntCauseBuildFailure)
+                            {
+                                _buildResult.ResultsByTarget[targetName].AfterTargetsHaveFailed = true;
```
Forgind: If target A has aftertarget B, and B has aftertarget C, and C fails, do you also need to specify that `_buildResult.ResultsByTarget[B].AfterTargetsHaveFailed = true`?

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
```diff
@@ -791,12 +791,13 @@ private void ComputeAfterTargetFailures(string[] targetNames)
...
                             {
                                 _buildResult.ResultsByTarget.TryGetValue(afterTarget.TargetName, out TargetResult result);
-                                afterTargetFailed = result != null && result.ResultCode == TargetResultCode.Failure && !result.TargetFailureDoesntCauseBuildFailure;
+                                afterTargetFailed = result?.ResultCode == TargetResultCode.Failure && !result.TargetFailureDoesntCauseBuildFailure;
 
                                 resultsWithAfterTargets[afterTarget.TargetName] = afterTargetFailed;
                                 if (afterTargetFailed)
                                 {
                                     _buildResult.ResultsByTarget[targetName].AfterTargetsHaveFailed = true;
+                                    break;
```
cdmihai: This only breaks the innermost foreach, not the while as well right? You could empty the queue or use a bool.

Forgind: Could also use `return`, since there isn't anything after that.

cdmihai: I think returning would be incorrect, as the outermost foreach over `targetNames` needs to explore whether each target's aftertargets closure contains a failing target.

ladipro: Fixed, thank you!

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
ladipro: No need to do that as we're going to forget the results of all but top-level targets anyway. Right after returning from this method, actually:

```
BuildResult resultsToReport = new BuildResult(_buildResult, targetNames);
```

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
Forgind: You're right. Too many comments—I missed the third loop.

src/Build/BackEnd/Components/Caching/ResultsCacheWithOverride.cs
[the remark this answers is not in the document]
rainersigwald: Roslyn now supports formalizing style questions like this in a `.editorconfig`, and then you can apply fixes across the entire source tree. I've been meaning to put together my set of proposed ones but haven't gotten around to it.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
rainersigwald: Agreed. We also need to develop a "what do we do for `#nullable`" strategy for MSBuild. Right now I've only explicitly enabled it in a few new files. We need to decide whether we want to go with the "turn it on as you touch stuff" or the "turn it off in every file" strategy. Here, no action needed.

src/Build/BackEnd/Components/RequestBuilder/TargetBuilder.cs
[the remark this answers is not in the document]
rainersigwald: No existing rules.

ladipro:
> LGTM! Would you mind squashing to two commits, one for the fixed links and one for the main change?

Done, thank you!

PR #5196, opened 2020-03-24
Use environment variable for handshake Resolves #4961

### LINKED ISSUE #4961: New Tracing & Hash Computation Does Not Apply to Non-TaskHost Worker Nodes
The following trace does not apply to non-taskhost msbuild worker nodes:

https://github.com/microsoft/msbuild/blob/master/src/Shared/CommunicationsUtilities.cs#L665

Other potential areas the trace logic should move to:
https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs#L78

https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs#L93

Pairs with #4960

Forgind:
Adds use of an environment variable for non-task host nodes as well. Some duplicated logic should ideally be refactored pending acceptance of an altered handshake.

Fixes #4961.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
@@ -14,6 +14,7 @@ using System.Security.Permissions;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.Internal;
+using Microsoft.Build.Utilities;
 
 namespace Microsoft.Build.BackEnd
 {
@@ -71,8 +72,8 @@ namespace Microsoft.Build.BackEnd
         /// <param name="enableLowPriority">Is the build running at low priority?</param>
         internal static long GetHostHandshake(bool enableNodeReuse, bool enableLowPriority)
         {
-            CommunicationsUtilities.Trace("MSBUILDNODEHANDSHAKESALT=\"{0}\", msbuildDirectory=\"{1}\", enableNodeReuse={2}, enableLowPriority={3}", Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT"), BuildEnvironmentHelper.Instance.MSBuildToolsDirectory32, enableNodeReuse, enableLowPriority);
-            return CommunicationsUtilities.GetHostHandshake(CommunicationsUtilities.GetNodeProviderContext(enableNodeReuse, enableLowPriority, EnvironmentUtilities.Is64BitProcess));
+            CommunicationsUtilities.Trace("MSBUILDNODEHANDSHAKESALT=\"{0}\", msbuildDirectory=\"{1}\", enableNodeReuse={2}, enableLowPriority={3}", Traits.MSBuildNodeHandshakeSalt, BuildEnvironmentHelper.Instance.MSBuildToolsDirectory32, enableNodeReuse, enableLowPriority);
+            return CommunicationsUtilities.GetHostHandshake(CommunicationsUtilities.GetHandshakeOptions(taskHost: false, nodeReuse: enableNodeReuse, lowPriority: enableLowPriority, is64Bit: EnvironmentUtilities.Is64BitProcess));
         }
 
         /// <summary>
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
@@ -3,24 +3,15 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
-using System.Text;
-using System.IO;
-using System.IO.Pipes;
 using System.Diagnostics;
-using System.Threading;
-using System.Runtime.InteropServices;
-using System.Security;
 #if FEATURE_SECURITY_PERMISSIONS
 using System.Security.AccessControl;
 #endif
-using System.Security.Principal;
 #if FEATURE_SECURITY_PERMISSIONS
 using System.Security.Permissions;
 #endif
 
 using Microsoft.Build.Shared;
-using Microsoft.Build.Framework;
 using Microsoft.Build.Exceptions;
 using Microsoft.Build.Internal;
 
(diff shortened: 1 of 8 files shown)


src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -80,7 +80,12 @@ public int AvailableNodes
         /// <param name="enableLowPriority">Is the build running at low priority?</param>
         internal static long GetHostHandshake(bool enableNodeReuse, bool enableLowPriority)
         {
-            long baseHandshake = Constants.AssemblyTimestamp;
+            CommunicationsUtilities.Trace("MSBUILDNODEHANDSHAKESALT=\"{0}\", msbuildDirectory=\"{1}\", enableNodeReuse={2}, enableLowPriority={3}", Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT"), BuildEnvironmentHelper.Instance.MSBuildToolsDirectory32, enableNodeReuse, enableLowPriority);
+
+            string salt = Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT") + BuildEnvironmentHelper.Instance.MSBuildToolsDirectory32;
```
cdmihai: There's also a `BuildEnvironment.MSBuildToolsDirectory64`. Shouldn't the handshake reflect that bitness as well?

cdmihai:
It would sure be nice if there would one single place that computes the handshake, like a `MSBuildHandshakeBuilder`. Also, is that salt and tools directory really useful? Could the unified handshake not contain them? Who sets `MSBUILDNODEHANDSHAKESALT`? Searched the VS codebase and also googled it and didn't find anyone who sets it.

src/Shared/CommunicationsUtilities.cs
```diff
@@ -490,86 +493,85 @@ internal static async Task<int> ReadAsync(Stream stream, byte[] buffer, int byte
         /// <summary>
         /// Given the appropriate information, return the equivalent TaskHostContext.  
         /// </summary>
-        internal static TaskHostContext GetTaskHostContext(IDictionary<string, string> taskHostParameters)
+        internal static HandshakeOptions GetTaskHostContext(IDictionary<string, string> taskHostParameters)
```
cdmihai, line 496: Rename to GetHandshakeOptions to keep in line with the new enum name?

src/Shared/CommunicationsUtilities.cs
```diff
@@ -292,9 +293,8 @@ internal static void SetEnvironment(IDictionary<string, string> newEnvironment)
 
         /// <summary>
         /// Given a base handshake, generates the real handshake based on e.g. elevation level.  
-        /// Client handshake required for comparison purposes only.  Returns the update handshake.  
         /// </summary>
-        internal static long GenerateHostHandshakeFromBase(long baseHandshake, long clientHandshake)
+        internal static long GenerateHostHandshakeFromBase(long baseHandshake)
```
cdmihai, line 297: Make private or inline, no need to expose it anymore.

src/Shared/CommunicationsUtilities.cs
cdmihai: Actually, wouldn't it be cleaner if you moved all the GetTaskHostContext methods as static constructors on the enum instead, and rename to `From`: `HandshakeOptions.From(bool, bool)`

---
In reply to: [412594545](https://github.com/microsoft/msbuild/pull/5196#discussion_r412594545) [](ancestors = 412594545)

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcTaskHost.cs
```diff
@@ -388,9 +388,13 @@ internal static void ClearCachedTaskHostPaths()
         /// <summary>
         /// Given a TaskHostContext, returns the name of the executable we should be searching for.
         /// </summary>
-        internal static string GetTaskHostNameFromHostContext(TaskHostContext hostContext)
+        internal static string GetTaskHostNameFromHostContext(HandshakeOptions hostContext)
         {
-            if (hostContext == TaskHostContext.X64CLR4 || hostContext == TaskHostContext.X32CLR4)
+            ErrorUtilities.ThrowInternalErrorUnreachable((hostContext & HandshakeOptions.TaskHost) == HandshakeOptions.TaskHost);
```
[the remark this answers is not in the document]
Forgind: If the condition is _false_, it throws, so this throws if it is not a task host that is passed in. I modeled it after cases like `ErrorUtilities.VerifyThrow(bool condition, string unformattedMessage)` where the first line of the body is `if (!condition)`.

If I wanted to do it in a switch/case of if/else, I'd either have to start with checking whether it's a task host (i.e., this) or put it at the end with every condition before it verifying that it's a task host. Since it's an error condition rather than an expected case, I think separating it somehow is better, and this is cleaner than adding a condition to every other case.

src/Shared/CommunicationsUtilities.cs
Forgind: 👍

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
Forgind: 👍

src/Shared/CommunicationsUtilities.cs
```diff
@@ -25,32 +23,35 @@ namespace Microsoft.Build.Internal
     /// <summary>
     /// Enumeration of all possible (currently supported) types of task host context.
```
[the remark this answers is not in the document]
Forgind: 👍

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -80,15 +71,8 @@ public int AvailableNodes
...
-            long baseHandshake = Constants.AssemblyTimestamp;
-
-            baseHandshake = baseHandshake*17 + EnvironmentUtilities.Is64BitProcess.GetHashCode();
-
-            baseHandshake = baseHandshake*17 + enableNodeReuse.GetHashCode();
-
-            baseHandshake = baseHandshake*17 + enableLowPriority.GetHashCode();
-
-            return CommunicationsUtilities.GenerateHostHandshakeFromBase(baseHandshake, GetClientHandshake());
+            CommunicationsUtilities.Trace("MSBUILDNODEHANDSHAKESALT=\"{0}\", msbuildDirectory=\"{1}\", enableNodeReuse={2}, enableLowPriority={3}", Environment.GetEnvironmentVariable("MSBUILDNODEHANDSHAKESALT"), BuildEnvironmentHelper.Instance.MSBuildToolsDirectory32, enableNodeReuse, enableLowPriority);
```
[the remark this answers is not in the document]
Forgind: 👍

src/Shared/CommunicationsUtilities.cs
Forgind: Can an enum have a method like a class? I think I'd have to replace the enum with a class, and that would make it much bigger.

cdmihai: You can write extension methods for it: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-create-a-new-method-for-an-enumeration

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcTaskHost.cs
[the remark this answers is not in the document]
rainersigwald:
```suggestion
            ErrorUtilities.VerifyThrowInternalErrorUnreachable((hostContext & HandshakeOptions.TaskHost) == HandshakeOptions.TaskHost);
```

src/Shared/Traits.cs
```diff
@@ -61,6 +61,11 @@ public Traits()
         /// </summary>
         public readonly bool EnableRestoreFirst = Environment.GetEnvironmentVariable("MSBUILDENABLERESTOREFIRST") == "1";
 
+        /// <summary>
+        /// Allow the user to specify that two processes should not be communicating via an environment variable.
+        /// </summary>
+        public static readonly int MSBuildNodeHandshakeSalt = ParseIntFromEnvironmentVariableOrDefault("MSBUILDNODEHANDSHAKESALT", -1);
```
rainersigwald, line 67:
```suggestion
        public static readonly string MSBuildNodeHandshakeSalt = ParseIntFromEnvironmentVariableOrDefault("MSBUILDNODEHANDSHAKESALT", -1);
```

src/Shared/CommunicationsUtilities.cs
Forgind: I put a change [here](https://github.com/Forgind/msbuild/commit/001c9b69a6e5b8bcf68355f753b5f2234c11fa6b), but I don't think I like it. It would probably be good if there were a lot of classes that were going to touch `HandshakeOptions`, but since all references to it are focused in this class, it seems unnecessary.

PR #5200, opened 2020-03-25
Don't import Microsoft.WinFX.targets for sdk style projects

BenVillalobos:
Part 1 to fix https://github.com/microsoft/msbuild/issues/4948

Part 2 (sets the ImportWinFXTargets property to false): https://github.com/dotnet/sdk/pull/10998

This conditionally imports winfx.targets for all non-sdk style builds. Pairs with another PR for the sdk 
that ensures the condition will fall through.

### DIFF
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -109,6 +109,8 @@ namespace Microsoft.Build.Utilities
         /// </summary>
         public readonly bool AlwaysUseContentTimestamp = Environment.GetEnvironmentVariable("MSBUILDALWAYSCHECKCONTENTTIMESTAMP") == "1";
 
+        public readonly bool LogTaskInputsVerbose = Environment.GetEnvironmentVariable("MSBUILDLOGTASKINPUTSVERBOSE") == "1";
+
         /// <summary>
         /// Emit events for project imports.
         /// </summary>
--- a/src/Shared/Traits.cs
+++ b/src/Shared/Traits.cs
@@ -109,8 +109,6 @@ namespace Microsoft.Build.Utilities
         /// </summary>
         public readonly bool AlwaysUseContentTimestamp = Environment.GetEnvironmentVariable("MSBUILDALWAYSCHECKCONTENTTIMESTAMP") == "1";
 
-        public readonly bool LogTaskInputsVerbose = Environment.GetEnvironmentVariable("MSBUILDLOGTASKINPUTSVERBOSE") == "1";
-
         /// <summary>
         /// Emit events for project imports.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.WinFx.targets
```diff
@@ -9,6 +9,6 @@
         the shim will internally simply redirect to the real copy of the targets file in the .NET Framework. 
    -->
 
-   <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets" Condition="Exists('$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets')" />
+   <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets" Condition="Exists('$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets') and '$(ImportWinFXTargets)' != 'false'" />
```
cdmihai: Check the property first to avoid doing the IO existence check. The and operator is short circuiting.

BenVillalobos: Good call, done!

src/Tasks/Microsoft.WinFx.targets
```diff
@@ -9,6 +9,6 @@
         the shim will internally simply redirect to the real copy of the targets file in the .NET Framework. 
    -->
 
-   <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets" Condition="Exists('$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets')" />
+   <Import Project="$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets" Condition="'$(ImportWinFXTargets)' != 'false' and Exists('$(MSBuildFrameworkToolsPath)\Microsoft.WinFx.targets')" />
```
Nirmal4G: Why not use `ImportInboxWinFXTargets` or `ImportFrameworkWinFXTargets`, seems more clear as to what we're not importing and where it's from. I prefer the former name.

BenVillalobos: Good suggestion. I opted for ImportFrameworkWinFXTargets. 

Thanks!

BenVillalobos:
Verified that this works locally 👍

src/Shared/Traits.cs
```diff
@@ -109,6 +109,8 @@ internal class EscapeHatches
         /// </summary>
         public readonly bool AlwaysUseContentTimestamp = Environment.GetEnvironmentVariable("MSBUILDALWAYSCHECKCONTENTTIMESTAMP") == "1";
 
+        public readonly bool LogTaskInputsVerbose = Environment.GetEnvironmentVariable("MSBUILDLOGTASKINPUTSVERBOSE") == "1";
```
rainersigwald, line 112: Did this leak in from your other change?

BenVillalobos: Whoops! I'll revert that.

Thanks!

PR #5206, opened 2020-03-27
Revert "Support host objects in 64-bit MSBuild (#5177)"

Forgind:
This reverts commit 2d82e1a861d890fce68c8e2d42b569e5bbaf5687.

It failed regression tests here:
https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/237504

Reverting it allowed tests to pass again:
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/237771

Reverting this while we investigate.

### DIFF
--- a/src/Build/Instance/HostServices.cs
+++ b/src/Build/Instance/HostServices.cs
@@ -90,6 +90,11 @@ namespace Microsoft.Build.Execution
                 {
 #if FEATURE_COM_INTEROP
 
+                    if (Environment.Is64BitProcess)
+                    {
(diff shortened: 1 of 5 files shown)


Forgind:
azp /run

Forgind:
@davidwengier FYI: our current policy is to revert changes that cause regressions while we investigate so we can unblock insertions.

davidwengier:
I maintain my innocence! :P

Whilst I don't know anything about investigating these failures, let me know if there is anything I need to do to help.

PR #5210, opened 2020-03-30
Option to Truncate Task Input-Logging

BenVillalobos:
After a limit of 10k characters, stop logging task inputs.

Fixes https://github.com/microsoft/msbuild/issues/4907

cc: @rainersigwald @cdmihai How would we go about using a localized string instead of the hard-coded english one I used?

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -8,6 +8,7 @@ using System.Globalization;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 using System.Reflection;
+using Microsoft.Build.Utilities;
 
 namespace Microsoft.Build.BackEnd
 {
@@ -75,7 +76,14 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("        ");
                     }
 
-                    sb.Append(GetStringFromParameterValue(parameterValue[i]));
+                    string data = GetStringFromParameterValue(parameterValue[i]);
+
+                    if(!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && data.Length > )
+                    {
+
+                    }
+
+                    sb.Append(data);
 
                     if (!specialTreatmentForSingle && i < parameterValue.Count - 1)
                     {
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Build.BackEnd
                     sb.Append("\n");
                 }
 
-                bool logVerbose = Traits.Instance.EscapeHatches.LogTaskInputsVerbose;
+                bool truncateLogging = Traits.Instance.EscapeHatches.TruncateTaskInputLogging;
 
                 for (int i = 0; i < parameterValue.Count; i++)
                 {
@@ -95,7 +95,7 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("\n");
                     }
 
-                    if (!logVerbose && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
+                    if (truncateLogging && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
                     {
                         sb.Append(ResourceUtilities.GetResourceString("LogTaskInputs.Truncated"));
                         break;
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Build.BackEnd
         /// <summary>
         /// The default character limit for logging parameters. 10k is somewhat arbitrary, see https://github.com/microsoft/msbuild/issues/4907.
         /// </summary>
-        internal static int parameterCharacterLimit = 10_000;
+        internal static int parameterCharacterLimit = 40_000;
 
         /// <summary>
         /// The default parameter limit for logging. 200 is somewhat arbitrary, see https://github.com/microsoft/msbuild/pull/5210.
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Build.BackEnd
                     sb.Append("\n");
                 }
 
-                bool truncateLogging = Traits.Instance.EscapeHatches.TruncateTaskInputLogging;
+                bool logAllTaskInputs = Traits.Instance.EscapeHatches.LogAllTaskInputs;
 
                 for (int i = 0; i < parameterValue.Count; i++)
                 {
@@ -95,7 +95,7 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("\n");
                     }
 
-                    if (truncateLogging && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
+                    if (!logAllTaskInputs && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
                     {
                         sb.Append(ResourceUtilities.GetResourceString("LogTaskInputs.Truncated"));
                         break;
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -97,7 +97,7 @@ namespace Microsoft.Build.BackEnd
 
                     if (!logVerbose && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
                     {
-                        sb.Append("...\nThe parameters have been truncated beyond this point. To view all parameters, set environment variable MSBUILDLOGTASKINPUTSVERBOSE to 1");
+                        sb.Append(ResourceUtilities.GetResourceString("LogTaskInputs.Truncated"));
                         break;
                     }
                 }
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -7,8 +7,8 @@ using System.Collections.Generic;
 using System.Globalization;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
-using System.Reflection;
 using Microsoft.Build.Utilities;
+using System.Reflection;
 
 namespace Microsoft.Build.BackEnd
 {
@@ -18,9 +18,14 @@ namespace Microsoft.Build.BackEnd
     internal static class ItemGroupLoggingHelper
     {
         /// <summary>
-        /// The default character limit for logging parameters.
+        /// The default character limit for logging parameters. 10k is somewhat arbitrary, see https://github.com/microsoft/msbuild/issues/4907.
         /// </summary>
-        internal static int parameterTextLimit = 10000;
+        internal static int parameterCharacterLimit = 10_000;
+
+        /// <summary>
+        /// The default parameter limit for logging. 200 is somewhat arbitrary, see https://github.com/microsoft/msbuild/pull/5210.
+        /// </summary>
+        internal static int parameterLimit = 200;
 
         /// <summary>
         /// Gets a text serialized value of a parameter for logging.
@@ -69,6 +74,8 @@ namespace Microsoft.Build.BackEnd
                     sb.Append("\n");
                 }
 
+                bool logVerbose = Traits.Instance.EscapeHatches.LogTaskInputsVerbose;
+
                 for (int i = 0; i < parameterValue.Count; i++)
                 {
                     if (parameterValue[i] == null)
@@ -88,9 +95,9 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("\n");
                     }
 
-                    if (!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && sb.Length >= parameterTextLimit)
+                    if (!logVerbose && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
                     {
-                        sb.Append("...\nThe parameters have been truncated beyond this point. To view all parameters, set environment variable MSBUILDLOGTASKINPUTSVERBOSE=1");
+                        sb.Append("...\nThe parameters have been truncated beyond this point. To view all parameters, set environment variable MSBUILDLOGTASKINPUTSVERBOSE to 1");
                         break;
                     }
                 }
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -18,6 +18,11 @@ namespace Microsoft.Build.BackEnd
     internal static class ItemGroupLoggingHelper
     {
         /// <summary>
+        /// The default character limit for logging parameters.
+        /// </summary>
+        internal static int parameterTextLimit = 10000;
+
+        /// <summary>
         /// Gets a text serialized value of a parameter for logging.
         /// </summary>
         internal static string GetParameterText(string prefix, string parameterName, params object[] parameterValues)
@@ -76,18 +81,17 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("        ");
                     }
 
-                    string data = GetStringFromParameterValue(parameterValue[i]);
-
-                    if(!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && data.Length > )
-                    {
-
-                    }
-
-                    sb.Append(data);
+                    sb.Append(GetStringFromParameterValue(parameterValue[i]));
 
                     if (!specialTreatmentForSingle && i < parameterValue.Count - 1)
                     {
                         sb.Append("\n");
+                    }
+
+                    if (!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && sb.Length >= parameterTextLimit)
+                    {
+                        sb.Append("...\nThe parameters have been truncated beyond this point. To view all parameters, set environment variable MSBUILDLOGTASKINPUTSVERBOSE=1");
+                        break;
                     }
                 }
 
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Build.BackEnd
                     sb.Append("\n");
                 }
 
-                bool logAllTaskInputs = Traits.Instance.EscapeHatches.LogAllTaskInputs;
+                bool truncateTaskInputs = Traits.Instance.EscapeHatches.TruncateTaskInputs;
 
                 for (int i = 0; i < parameterValue.Count; i++)
                 {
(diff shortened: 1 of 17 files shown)


src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -16,6 +17,11 @@ namespace Microsoft.Build.BackEnd
     /// </summary>
     internal static class ItemGroupLoggingHelper
     {
+        /// <summary>
+        /// The default character limit for logging parameters.
```
Forgind: Might want to explain why. I think this was arbitrary, but that would be my first question.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -81,6 +87,12 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
                     {
                         sb.Append("\n");
                     }
+
+                    if (!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && sb.Length >= parameterTextLimit)
+                    {
+                        sb.Append("...\nThe parameters have been truncated beyond this point. To view all parameters, set environment variable MSBUILDLOGTASKINPUTSVERBOSE=1");
```
Forgind: Use ResourceUtilities.FormatString (possibly with an option)

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -81,6 +87,12 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
                     {
                         sb.Append("\n");
                     }
+
+                    if (!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && sb.Length >= parameterTextLimit)
```
Forgind: Do we also want to limit the number of strings outputted?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind:
```suggestion
                        sb.Append("...\nThe parameters have been truncated beyond this point. To view all parameters, set the environment variable MSBUILDLOGTASKINPUTSVERBOSE to 1.");
```

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
BenVillalobos: Based on the original issue it looked like we would rather limit the raw data it uses as opposed to the number of items.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
BenVillalobos: The beauty is that it is arbitrary 😁. I'll link to the original issue for some context.

> We spitballed a limit like "the first 10k characters of the item inputs as text".

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
BenVillalobos: To format what exactly?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind: I'd limit both. Just as you don't care about the 100,000th character, you probably don't care about the 100,000th string. You _might_ care about the last string, so ideally, you might capture the first 9k and the last 1k or something like that.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
cdmihai: Here's an example: https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/BuildManager/BuildManager.cs#L1983

cdmihai:
@KirillOsenkov Does the binlog viewer parse the the task input string? If yes, then the bin logger needs to opt into not truncating the string right?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
BenVillalobos: I'll spitball 200 here, I saw one edge case(?) that was about 600 items. When building msbuild the parameters seemed to hover around 6-15.

KirillOsenkov: Consider extracting `Traits.Instance.EscapeHatches.LogTaskInputsVerbose` into a local variable outside the loop.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 using System.Reflection;
+using Microsoft.Build.Utilities;
```
KirillOsenkov: Mind sorting the usings here?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: First 200 items seems about right to me.

KirillOsenkov:
Actually let me do some research tomorrow and see how many strings exceed 10K in existing binlogs and how the binlogs would look with this change. Honestly I agree with Nick here:
https://github.com/microsoft/msbuild/issues/2200#issuecomment-307229017

We shouldn't be concatenating or using a stream builder, or truncating. We should be streaming the existing strings to the loggers as is, and it'd be up to the loggers to either efficiently write the string out to their streams, or do something else. Basically this is totally worth fixing, but there's a more proper fundamental fix.

KirillOsenkov:
I had to go to great lengths in the binlog viewer to parse these strings back out, and I still don't get it fully right as you lose information when you concatenate. Examples: 

https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/291
https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/224

If these come in a structure then I won't need to parse them and this will significantly speed up loading binlogs as well as present the items correctly.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -16,6 +17,11 @@ namespace Microsoft.Build.BackEnd
     /// </summary>
     internal static class ItemGroupLoggingHelper
     {
+        /// <summary>
+        /// The default character limit for logging parameters.
+        /// </summary>
+        internal static int parameterTextLimit = 10000;
```
rainersigwald: extreme nit:
```suggestion
        internal static int parameterTextLimit = 10_000;
```

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
rainersigwald: We've seen 10k items in C++ builds.

I don't object to an item-count limit, but it makes things more complex, because a single item can go over the desired character-count limit. So you still need that, too.

BenVillalobos: A combination of both seems like the way to go here. 200 items or 10k characters, whichever happens first will result in the rest of the results being truncated.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind: Note that that commit didn't actually sort the usings.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -18,9 +18,14 @@ namespace Microsoft.Build.BackEnd
...
-        /// The default character limit for logging parameters.
+        /// The default character limit for logging parameters. 10k is somewhat arbitrary, see https://github.com/microsoft/msbuild/issues/4907.
+        /// </summary>
+        internal static int parameterCharacterLimit = 10_000;
+
+        /// <summary>
+        /// The default parameter limit for logging. 200 is somewhat arbitrary, see https://github.com/microsoft/msbuild/pull/5210.
         /// </summary>
-        internal static int parameterTextLimit = 10000;
+        internal static int parameterLimit = 200;
```
Forgind: Rainer said 10k or perhaps even 20k is more reasonable here.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -88,9 +95,9 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
                         sb.Append("\n");
                     }
 
-                    if (!Traits.Instance.EscapeHatches.LogTaskInputsVerbose && sb.Length >= parameterTextLimit)
+                    if (!logVerbose && (sb.Length >= parameterCharacterLimit || i > parameterLimit))
```
Forgind: Grab the first 0.9 * parameterLimit and the last 0.1 * parameterLimit.

rainersigwald: I don't think this is worth the additional complexity.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
rainersigwald: No, I didn't. I think it's good for the _character_ limit.

src/Shared/Traits.cs
```diff
@@ -115,6 +115,11 @@ internal class EscapeHatches
         /// </summary>
         public readonly bool AlwaysUseContentTimestamp = Environment.GetEnvironmentVariable("MSBUILDALWAYSCHECKCONTENTTIMESTAMP") == "1";
 
+        /// <summary>
+        /// Stop truncating task inputs when logging them.
+        /// </summary>
+        public readonly bool LogTaskInputsVerbose = Environment.GetEnvironmentVariable("MSBUILDLOGTASKINPUTSVERBOSE") == "1";
```
rainersigwald:
```suggestion
        public readonly bool LogTaskInputsVerbose = Environment.GetEnvironmentVariable("MSBUILDLOGTASKINPUTSCOMPLETE") == "1";
```

?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind: "We've seen 10k items in C++ builds" was very confusing, then.

KirillOsenkov:
@rainersigwald yeah I wasn't blocking the PR. It's a good first step.

Just saying that the next step I'd like is remove string concatenations in this file (write straight to the string builder). That would be step number 2 and I volunteer to experiment with this.

Step number three is more advanced and complicated - add structure on TaskStartedEventArgs that would contain the input properties and items as is, instead of a giant string. Then move the writing logic to individual loggers. Preserve legacy behavior for loggers that don't support the new structure. I'd probably also volunteer to explore this, although given that this would take much longer I'm not sure if I can make a time commitment.

src/Shared/Traits.cs
BenVillalobos: Other options. Not sure I have a favorite.

MSBUILDLIMITLOGTASKINPUTS
MSBUILDLOGALLTASKINPUTS
MSBUILDLOGTASKINPUTSNOLIMIT

KirillOsenkov:
So I've gotten some stats from a real world 140 MB binlog (on the upper end of typical). We would truncate 2,314 strings out of 15,970 tasks:

image

The largest string is ~ 5MB and it quickly drops down from there.

src/Shared/Traits.cs
rainersigwald: I like `MSBUILDLOGALLTASKINPUTS` best I think.

BenVillalobos:
@KirillOsenkov Thank you for gathering this data! Would you recommend any changes to the values we've estimated based on this?

KirillOsenkov:
No idea yet, but I'm looking. So far 10,000 looks reasonable. Here are the tasks that log the most:

```
RemoveDuplicates 4998577
RemoveDuplicates 3357896
RestoreTask 3217404
Csc 1883626
CreateCSharpManifestResourceName 1449534
AssignCulture 1113606
RemoveDuplicates 606092
RemoveDuplicates 567005
RemoveDuplicates 506056
RemoveDuplicates 479566
RemoveDuplicates 468576
AssignTargetPath 467715
RemoveDuplicates 443453
FilterLocalCopyReferences 441703
RemoveDuplicates 438016
RemoveDuplicates 432498
FindUnderPath 431132
RemoveDuplicates 426718
Copy 401155
RemoveDuplicates 385957
Csc 383570
RemoveDuplicates 381534
Copy 375446
Hash 373980
RemoveDuplicates 365493
FilterLocalCopyReferences 362844
RemoveDuplicates 358903
Hash 348189
FindUnderPath 346383
RemoveDuplicates 346364
RemoveDuplicates 346186
RemoveDuplicates 343564
ConvertToAbsolutePath 340771
RemoveDuplicates 339519
RemoveDuplicates 336046
RemoveDuplicates 332830
Csc 315097
RemoveDuplicates 314012
RemoveDuplicates 311447
Csc 311201
```

KirillOsenkov:
Feels like significant savings can be had by just not logging the inputs and outputs of RemoveDuplicates... I'll think about it more.
https://source.dot.net/#Microsoft.Build.Tasks.Core/ListOperators/RemoveDuplicates.cs,6ee306d2d90af177

KirillOsenkov:
Seems like the binlog viewer is able to deal with truncated strings just fine:

image

KirillOsenkov:
I suppose the main concern is that you can't trust the search in binlogs anymore because you don't know whether the thing is really not there or it happened to be truncated. Really Csc is the task I worry about the most... It's crucial to know what references are being passed to the compiler and they are the most likely to get truncated.

image

cdmihai:
@KirillOsenkov 
> I suppose the main concern is that you can't trust the search in binlogs anymore because you don't know whether the thing is really not there or it happened to be truncated. Really Csc is the task I worry about the most...

I see that the method that does the clipping is called from [TaskExecutionHost](https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs#L1371), which has access to the ITask instance. If you really want to exclude `csc` from clipping the info, I guess we can add a new protected `ToolTask.DoNotClip` field, which `csc` sets to true.

KirillOsenkov:
Let's keep this PR alive for now, and in a few days we'll compare which is better ;) I am now hungry for all of this low hanging fruit.

KirillOsenkov:
I have lots of thoughts already after researching into this, and will have lots more in the next few days.

For now, the most important epiphany is:
1) we do need to log all items (maybe clip after say 200).
2) we don't need to log metadata. The bulk is in metadata, and it's noisy and rarely useful. I think we should explicitly turn on logging metadata for just a few chosen tasks. RAR inputs and Csc inputs are literally the only two tasks that come to mind where I'd want to see item metadata in task inputs.

Otherwise just log metadata in two places: initial items for the project, and any time items are added/removed and/or metadata is changed. There's 0 reason to log metadata in task inputs or outputs if it doesn't change.

BenVillalobos:
[Consensus via teams](https://teams.microsoft.com/l/message/19:3212bf033f4c4b5198643a04fa1048fa@thread.skype/1585693964033?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&parentMessageId=1585693964033&teamName=.NET%20Developer%20Experience&channelName=MSBuild&createdTime=1585693964033) is that this change should start _disabled_ with the option to _enable_. This is a blanket escape hatch for those hitting OOM issues.

src/Build/Resources/Strings.resx
```diff
@@ -616,6 +616,9 @@
   <data name="EnvironmentHeader" UESanitized="false" Visibility="Public">
     <value>Environment at start of build:</value>
   </data>
+  <data name="LogTaskInputs.Truncated" UESanitized="false" Visibility="Public">
```
Forgind, line 619: You don't need `UESanitized="false"`

src/Shared/Traits.cs
```diff
@@ -115,6 +115,11 @@ internal class EscapeHatches
         /// </summary>
         public readonly bool AlwaysUseContentTimestamp = Environment.GetEnvironmentVariable("MSBUILDALWAYSCHECKCONTENTTIMESTAMP") == "1";
 
+        /// <summary>
+        /// Stop truncating task inputs when logging them.
+        /// </summary>
+        public readonly bool TruncateTaskInputLogging = Environment.GetEnvironmentVariable("MSBUILDTRUNCATETASKINPUTLOGGING") == "1";
```
Forgind, line 121: https://github.com/microsoft/msbuild/pull/5210#discussion_r401197592

BenVillalobos: I opted for a name change to make it easier to understand. But I realize we could use the same name and have the condition be something like this: `public readonly bool LogAllTaskInputs = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBUILDLOGALLTASKINPUTS"));` so if it is empty (as it should be by default), it'll log everything. To disable it you'd have to set it to any value. It makes slightly less sense to me this way but that seems to be the convention.

KirillOsenkov:
I looked at the data more and I'm thinking we should bump the threshold from 10K to 40K, and here's why. The Large Object Heap threshold is 85K. Strings are 2 bytes each + header. So if you roughly divide that in half you get 40K.

From the data, by increasing the limit from 10K to 40K we still stay out of the LOH, and gain back more than half of the tasks where we are lossy.

This would still fulfil the primary goal of avoiding OOM since we still have a failsafe against worst offenders, but this maximizes the usable binlogs without fragmenting the LOH.

src/Shared/Traits.cs
BenVillalobos: After changing it back to `LogAllTaskInputs`, I reverted back to the `TruncateTaskInputs` convention. It just makes more sense to me.

src/Build/Resources/Strings.resx
rainersigwald:
```suggestion
  <data name="LogTaskInputs.Truncated" Visibility="Public">
```

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -16,6 +17,16 @@ namespace Microsoft.Build.BackEnd
     /// </summary>
     internal static class ItemGroupLoggingHelper
     {
+        /// <summary>
+        /// The default character limit for logging parameters. 10k is somewhat arbitrary, see https://github.com/microsoft/msbuild/issues/4907.
+        /// </summary>
+        internal static int parameterCharacterLimit = 10_000;
```
rainersigwald, line 23:
```suggestion
        internal static int parameterCharacterLimit = 40_000;
```

? To just miss the LOH?

KirillOsenkov: Right

src/Shared/Traits.cs
```diff
@@ -115,6 +115,11 @@ internal class EscapeHatches
         /// </summary>
         public readonly bool AlwaysUseContentTimestamp = Environment.GetEnvironmentVariable("MSBUILDALWAYSCHECKCONTENTTIMESTAMP") == "1";
 
+        /// <summary>
+        /// Stop truncating task inputs when logging them. Setting any value disables this.
```
rainersigwald, line 119:
```suggestion
        /// Truncate task inputs when logging them. This can reduce memory pressure
        /// at the expense of log usefulness.
```

Forgind:
I can restart RPS, but I'd be very reluctant to merge anything more with insertions blocked.

KirillOsenkov:
🎉

PR #5212, opened 2020-03-31
Implement an empty SGen .NET Core task

### LINKED ISSUE #3583: SGen on .NET Core should point to Microsoft.XmlSerializer.Generator package
In the early days of .NET Core there was no sgen, and we didn't port the sgen task.

There's a new .NET Core serializer generator: https://github.com/dotnet/corefx/issues/25138. Docs at  https://github.com/dotnet/core/blob/master/samples/xmlserializergenerator-instructions.md.

We should promote that when someone attempts to use the SGen task on .NET Core. Probably just emit an error pointing to an aka.ms link to those docs?

cc @huanwu in case there's a better thing to do.

ladipro:
The SGen task has not been implemented on .NET Core and building a project with `<GenerateSerializationAssemblies>` is failing with what looks like an infrastructure error.

This PR is adding an empty implementation of the task which throws a nicer error pointing the user to the build-time serialization mechanism supported on .NET Core - Microsoft.XmlSerializer.Generator. The changes here are analogous to how the ResolveComReference task has been implemented to allow erroring out on Core.

Fixes #3583

### DIFF
--- a/src/Tasks/Resources/Strings.resx
+++ b/src/Tasks/Resources/Strings.resx
@@ -1968,7 +1968,7 @@
     <comment>{StrBegin="MSB3473: "}</comment>
   </data>
   <data name="SGen.TaskNotSupported">
-    <value>MSB3474: The task "{0}" is not supported on the .NET Core version of MSBuild. Please instead use the Microsoft XML Serializer Generator. See https://go.microsoft.com/fwlink/?linkid=858594 for further details.</value>
+    <value>MSB3474: The task "{0}" is not supported on the .NET Core version of MSBuild. Use the Microsoft XML Serializer Generator package instead. See https://go.microsoft.com/fwlink/?linkid=858594 for more information.</value>
     <comment>{StrBegin="MSB3474: "}</comment>
   </data>
   <!--
--- a/src/Tasks/Resources/Strings.resx
+++ b/src/Tasks/Resources/Strings.resx
@@ -1967,6 +1967,10 @@
     <value>MSB3473: Path for "{0}" is invalid. {1}</value>
     <comment>{StrBegin="MSB3473: "}</comment>
   </data>
+  <data name="SGen.TaskNotSupported">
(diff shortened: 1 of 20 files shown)


src/Tasks/SGen.cs
```diff
@@ -11,10 +11,123 @@
 
 namespace Microsoft.Build.Tasks
 {
+    internal interface ISGenTaskContract
```
rainersigwald: Doc comment for why this exists?

src/Tasks/Resources/Strings.resx
```diff
@@ -1967,6 +1967,10 @@
     <value>MSB3473: Path for "{0}" is invalid. {1}</value>
     <comment>{StrBegin="MSB3473: "}</comment>
   </data>
+  <data name="SGen.TaskNotSupported">
+    <value>MSB3474: The task "{0}" is not supported on the .NET Core version of MSBuild. Please instead use the Microsoft XML Serializer Generator. See https://go.microsoft.com/fwlink/?linkid=858594 for further details.</value>
```
rainersigwald: Tweaks--@kathleendollard, do you have any input on this error?

```suggestion
    <value>MSB3474: The task "{0}" is not supported on the .NET Core version of MSBuild. Use the Microsoft XML Serializer Generator package instead. See https://go.microsoft.com/fwlink/?linkid=858594 for more information.</value>
```

src/Tasks/SGen.cs
ladipro: Added, thanks.

src/Tasks/Resources/Strings.resx
ladipro: Synced up with @timheuer on this in Kathleen's absence. Tim is fine with the error but points out that the doc article it links to needs some love. Following up in this issue: https://github.com/dotnet/docs/issues/14862

ladipro: I've tweaked the error as you suggested. Thank you!

PR #5215, opened 2020-03-31
Add back support for host objects in 64-bit MSBuild

Forgind:
This reverts commit 28ce126ed4f2220d87317b7f95d03d116c72bb4b.

The original commit failed RPS twice, so it was reverted. It then passed RPS twice, so this reverts its revert.


BenVillalobos:
Is there a link to the build where RPS passed you can paste here? (for the sake of leaving a clean trail)

Forgind:
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/237504

PR #5217, opened 2020-04-01
Reduce allocations in ItemGroupLoggingHelper.

KirillOsenkov:
This code is a very hot path (though only hit when LogTaskInputs is set).
There was an n^2 string concatenation problem alongside with other easily avoidable allocations.
Cache resource strings to avoid extracting them from resources every time.
Use a ThreadStatic field to avoid allocating an array every time.
Sorting key value pairs in place is 3x faster than LINQ OrderBy.

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -93,6 +93,12 @@ namespace Microsoft.Build.BackEnd
         /// </summary>
         internal static string GetStringFromParameterValue(object parameterValue, bool includeMetadata = false)
         {
+            // fast path for the common case
+            if (parameterValue is string valueText)
+            {
+                return valueText;
+            }
+
             using (var sb = new ReuseableStringBuilder())
             {
                 AppendStringFromParameterValue(sb, parameterValue, includeMetadata);
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -16,6 +16,11 @@ namespace Microsoft.Build.BackEnd
     /// </summary>
     internal static class ItemGroupLoggingHelper
     {
+        internal static string ItemGroupIncludeLogMessagePrefix = ResourceUtilities.GetResourceString("ItemGroupIncludeLogMessagePrefix");
+        internal static string ItemGroupRemoveLogMessage = ResourceUtilities.GetResourceString("ItemGroupRemoveLogMessage");
+        internal static string OutputItemParameterMessagePrefix = ResourceUtilities.GetResourceString("OutputItemParameterMessagePrefix");
+        internal static string TaskParameterPrefix = ResourceUtilities.GetResourceString("TaskParameterPrefix");
+
         /// <summary>
         /// Gets a text serialized value of a parameter for logging.
         /// </summary>
@@ -56,7 +61,8 @@ namespace Microsoft.Build.BackEnd
                     sb.Append("\n    ");
                 }
 
-                sb.Append(parameterName + "=");
+                sb.Append(parameterName);
+                sb.Append('=');
 
                 if (!specialTreatmentForSingle)
                 {
@@ -75,7 +81,7 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("        ");
                     }
 
-                    sb.Append(GetStringFromParameterValue(parameterValue[i]));
+                    AppendStringFromParameterValue(sb, parameterValue[i]);
 
                     if (!specialTreatmentForSingle && i < parameterValue.Count - 1)
                     {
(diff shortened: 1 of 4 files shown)


KirillOsenkov:
Relevant benchmark:
https://github.com/KirillOsenkov/Benchmarks/blob/e9b3af701eb4bf98a598b866cc252349b50819e3/Benchmarks/Tests/SortDictionary.cs#L56

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -16,18 +16,15 @@ namespace Microsoft.Build.BackEnd
...
     internal static class ItemGroupLoggingHelper
     {
-        /// <summary>
-        /// Gets a text serialized value of a parameter for logging.
-        /// </summary>
-        internal static string GetParameterText(string prefix, string parameterName, params object[] parameterValues)
-        {
-            return GetParameterText(prefix, parameterName, (IList)parameterValues);
-        }
+        internal static string ItemGroupIncludeLogMessagePrefix = ResourceUtilities.GetResourceString("ItemGroupIncludeLogMessagePrefix");
```
Forgind, line 19: Rather than caching four specific strings (presumably the most oft-used?), could you initialize an empty dictionary and cache strings in there as you need them?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -56,7 +53,8 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
                     sb.Append("\n    ");
                 }
 
-                sb.Append(parameterName + "=");
+                sb.Append(parameterName);
+                sb.Append('=');
```
Forgind, line 57: At first, I thought this would allocate a new string for the `=` each time, but I see now it's a character, so that shouldn't be a problem.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -95,53 +101,76 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
...
-                    result += "\n";
-                    var names = new List<string>();
+                    sb.Append('\n');
 
-                    foreach (string name in customMetadata.Keys)
+                    // need to initialize the thread static on each new thread
+                    if (keyValuePairList == null)
                     {
-                        names.Add(name);
+                        keyValuePairList = new List<KeyValuePair<string, string>>(count);
```
rainersigwald: There's not a way to ensure `count` capacity in the already-allocated case, right? Because setting `Capacity` could result in reallocating to smaller, which isn't what we want.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -93,55 +91,92 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
...
-
-                var customMetadata = item.CloneCustomMetadata();
+                sb.Append(item.ItemSpec);
 
-                if (customMetadata.Count > 0)
+                if (includeMetadata)
                 {
-                    result += "\n";
-                    var names = new List<string>();
+                    var customMetadata = item.CloneCustomMetadata();
```
Forgind, line 120: Do we need to clone the metadata or can we just access it?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -93,55 +91,92 @@ internal static string GetParameterText(string prefix, string parameterName, ILi
...
+                        {
+                            keyValuePairList.Sort((l, r) => StringComparer.OrdinalIgnoreCase.Compare(l.Key, r.Key));
+                        }
 
-                        if (i < names.Count - 1)
+                        for (int i = 0; i < count; i++)
                         {
-                            result += "\n";
+                            var kvp = keyValuePairList[i];
+                            sb.Append("                ");
```
Forgind, line 157: Can this string be allocated outside the function so we don't allocate it multiple times?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind: Actually, these are the only strings that are used in normal conditions, right? So this seems reasonable.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: This list will always grow and never shrink. So once it grows to a sufficient size it stops rejiggering altogether. Clear() at the end keeps the Capacity unchanged.

KirillOsenkov: There's an EnsureCapacity() on List<T> which is exactly what we want, but it's internal. No biggie, I think it will only resize the first few times and that's it.

KirillOsenkov: Actually we could do if (Capacity < count) Capacity = count; but I'd rather rely on List's internal doubling algorithm.

Forgind: Just for clarity, since this is under `if (keyValuePairList == null)`, it will only look at count once per thread, so it will resize beyond when one gets to maximum capacity, but it shouldn't get smaller.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: Yes, the strings are fixed always.

There wasn't an OrderBy but I was considering various approaches to sort, and decided to measure properly. Good thing I didn't go with it ;) https://twitter.com/KirillOsenkov/status/1245180011457306624

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: String literals that are compile-time constants are stored in the Strings metadata heap and don't need to be allocated.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: The original code had CloneCustomMetadata() and I think it's the blessed way to retrieve a thread-safe read-only snapshot of custom metadata only.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: These aren't allocated. It's just `ldstr` which gets you a pointer to the Strings metadata heap cheaply: https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLRIMaOQSwG4SoAOsMECAdsgDQwg4A2APgAIBMAjALABQLAzAAJ2ggMKCA3r0EzhQlgBZBAWQAUASknTZOlhwCcqgERgA9qaPqA3NpkBfXnaA=

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind: Is there anything wrong with an unstable sort here? Metadata all have unique names.

KirillOsenkov: We should be good here. I doubt they really have to be sorted in the first place. The viewer could sort when loading the binlog. But I guess there's some merit to it if they always appear in the same order, maybe some log parsing tools rely on that. 

Although if a log parsing tool isn't using binlog they just deserve the pain they get ;)

Forgind: I see an argument both ways, since having a consistent order is good, but so is allocating less. I'm tempted to put out a no-sort version under a feature flag and see what people think, but it probably isn't worth it.

KirillOsenkov: Yes, there's just so much more low-hanging fruit everywhere. I'd love to fix that first.

KirillOsenkov:
Yes, I should tease these out to be two separate PRs. This kind of evolved as I was learning :)

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
Forgind: After we've reached this point, the custom metadata won't change, right? In which case, since we're also not modifying it, does it need to be thread-safe or read-only?

KirillOsenkov:
Here's the allocation reduction without the change to metadata logging:
https://github.com/microsoft/msbuild/pull/5224

KirillOsenkov:
I suggest let's close this PR, and use #5224 instead. After this merges I can start another one for turning off logging of metadata.

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
BenVillalobos: We should take note of other places where we can cache resource strings like this 👍

PR #5218, opened 2020-04-01
CloneCustomMetadata with capacity to avoid resizes

KirillOsenkov:


src/Build/Instance/ProjectItemInstance.cs
```diff
@@ -1386,9 +1386,10 @@ public void CopyMetadataTo(ITaskItem destinationItem, bool addOriginalItemSpec)
             /// <returns>The cloned metadata.</returns>
             public IDictionary CloneCustomMetadata()
             {
-                Dictionary<string, string> clonedMetadata = new Dictionary<string, string>(MSBuildNameIgnoreCaseComparer.Default);
+                var metadata = MetadataCollection;
+                Dictionary<string, string> clonedMetadata = new Dictionary<string, string>(metadata.Count, MSBuildNameIgnoreCaseComparer.Default);
```
Forgind, line 1390: Any reason you couldn't just use MetadataCollection.Count instead of making a new variable?

KirillOsenkov: Accessing MetadataCollection is expensive as it creates a new copy every time.

KirillOsenkov: This whole area has too many allocations. Hope one day we can switch to immutable collections and clean all of this up.

Forgind: Ah, I see. And you can't just cache it and avoid that problem because metadata can change, just not between these two points. Sounds good!

src/Build/Instance/ProjectItemInstance.cs
```diff
@@ -1386,9 +1386,10 @@ public void CopyMetadataTo(ITaskItem destinationItem, bool addOriginalItemSpec)
             /// <returns>The cloned metadata.</returns>
             public IDictionary CloneCustomMetadata()
             {
-                Dictionary<string, string> clonedMetadata = new Dictionary<string, string>(MSBuildNameIgnoreCaseComparer.Default);
+                var metadata = MetadataCollection;
```
cdmihai, line 1389: Are  you capturing the reference here to avoid data races between access to count and the enumerator?

KirillOsenkov: Calling MetadataCollection creates a new copy on each access, check the getter :)

PR #5222, opened 2020-04-01
Fix cache issues with graph exempt projects

### LINKED ISSUE #4386: Projects excluded from static graph constraint checks should not appear in the caches
#4379 enables projects to skip isolation constraints on certain references. These references should not appear in the serialized caches when `-outputResultsCache` is used. If the projects appear in the serialized caches, it can introduce a side channel through undeclared dependencies leak, leading to harder to understand, unstable builds.

cdmihai:
Fixes #4386

This PR fixes the bug cascade that results when exempting projects from graph isolation that also exist in the graph. It implements this region of the docs: https://github.com/microsoft/msbuild/compare/master...cdmihai:cacheAndIsolationDocumentation?expand=1#diff-5de82af4d8f52787d69b38fd58c509caR44

When a reference is exempt from isolation constraints, there can be multiple situations:
- the reference does not exist in the graph, or in the input caches files. This is what the original exemption feature was designed for, and it worked great for enabling WPF's on the fly generated projects work with isolation. All the following scenarios were not supported, and fixed in this PR.
- the reference exists in one or more input cache files. This causes part of the failures in #4386.  There are multiple sub cases here:
  - the reference is in the input cache, but its build results are not a superset of the current project's requests. This happens when the exempt project is in the input cache with results for other targets than the ones requested by the current project. 
  - the reference is in the  input cache and is a superset of the current project's requests.

Things this PR changes:
1. If the CacheAggregator detects conflicting incoming cache entries (conflicting instances of BuldRequestConfiguration and BuildResult), instead of erroring it will now merge them using the "first entry wins" strategy. BuildResults are merged at the level of individual target results. This allows multiple projects to exempt the same reference.
3. A BuildRequestConfiguration is marked as exempt (new bool field) whenever a building project marks it as exempt.
4. The override caches (ConfigCacheWithOverride, BuildResultsWithOverride) always ensure an entry is either in the override cache, or the current cache, but not in both.
  a. An exception here is the BuildResultsWithOverride cache, where an exempt configuration can appear in both caches, This happens when the exempt project is present in the input caches, but does not have target results that the current project wants to build. In this case, two things happen. In the ConfigWithOverride cache, the entry is migrated from the override cache to the current cache. In the BuildResultsWithOverride cache, The override entry gets the new, additional targets (not ideal, but it would be super hard to change), and the current cache gets only the newly built targets, to ensure the invariant that the current cache should only contain newly built things, and should never contain results from deserialized caches..

### DIFF
--- a/scripts/Deploy-MSBuild.ps1
+++ b/scripts/Deploy-MSBuild.ps1
@@ -8,15 +8,30 @@ Param(
   [string] $runtime = "Desktop"
 )
 
+Set-StrictMode -Version "Latest"
+$ErrorActionPreference = "Stop"
+
 function Copy-WithBackup ($origin) {
-    $destinationPath = Join-Path -Path $destination -ChildPath (Split-Path $origin -leaf)
+    $directoryPart = Join-Path -Path $destination $origin.IntermediaryDirectories 
+    $destinationPath = Join-Path -Path $directoryPart (Split-Path $origin.SourceFile -leaf)
 
     if (Test-Path $destinationPath -PathType Leaf) {
         # Back up previous copy of the file
         Copy-Item $destinationPath $BackupFolder -ErrorAction Stop
     }
 
-    Copy-Item $origin $destinationPath -ErrorAction Stop
+    if (!(Test-Path $directoryPart)) {
+        [system.io.directory]::CreateDirectory($directoryPart)
+    }
+
+    Copy-Item $origin.SourceFile $destinationPath -ErrorAction Stop
+
+    echo "Copied $($origin.SourceFile) to $destinationPath"
+}
+
+function FileToCopy([string] $sourceFileRelativeToRepoRoot, [string] $intermediaryDirectories)
+{
+    return [PSCustomObject]@{"SourceFile"=$([IO.Path]::Combine($PSScriptRoot, "..", $sourceFileRelativeToRepoRoot)); "IntermediaryDirectories"=$intermediaryDirectories}
 }
 
 # TODO: find destination in PATH if not specified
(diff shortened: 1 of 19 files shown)


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -511,22 +513,24 @@ void InitializeCaches()
                 _configCache = ((IBuildComponentHost)this).GetComponent(BuildComponentType.ConfigCache) as IConfigCache;
                 _resultsCache = ((IBuildComponentHost)this).GetComponent(BuildComponentType.ResultsCache) as IResultsCache;
 
-                if (!usesInputCaches && (_buildParameters.ResetCaches || _configCache.IsConfigCacheSizeLargerThanThreshold()))
+                if (usesInputCaches)
```
cdmihai, line 516: Affects non-isolation execution: just a condition simplification, should be benign.

src/Build/BackEnd/Shared/BuildRequestConfiguration.cs
```diff
@@ -547,6 +548,12 @@ internal int ResultsNodeId
             set => _resultsNodeId = value;
         }
 
+        public bool SkippedFromStaticGraphIsolationConstraints
```
cdmihai, line 551: Affects non-isolation execution: new translatable bool that never gets set or read outside isolation execution

cdmihai:
I've commented on all the diffs that affect non-isolation code paths, you can search for them with the string `Affects non-isolation execution:`

scripts/Deploy-MSBuild.ps1
```diff
@@ -36,58 +51,65 @@ if ($runtime -eq "Desktop") {
     $targetFramework = "netcoreapp2.1"
 }
 
+$bootstrapBinDirectory = "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework"
```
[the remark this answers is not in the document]
cdmihai: Honestly I do not know. I also saw that, but envisioned the hell of trying to understand the diffs so I didn't get into it. @rainersigwald, do you know more? :)

rainersigwald:
Talked to @cdmihai offline; the plan is to go a different direction.

PR #5223, opened 2020-04-01
Refactor translator APIs

dsplaisted:
ITranslator methods now take an ObjectTranslator delegate which can translate an object in both directions, as opposed to requiring that objects implement ITranslatable and that a factory be passed in.

The TranslatorHelpers class includes extension methods which support the old APIs, so call sites that use ITranslatable objects and NodePacketValueFactory still work.

### DIFF
--- a/src/Shared/ITranslator.cs
+++ b/src/Shared/ITranslator.cs
@@ -17,6 +17,13 @@ namespace Microsoft.Build.BackEnd
     internal delegate T NodePacketValueFactory<T>(ITranslator translator);
 
     /// <summary>
+    /// Delegate for users that want to translate an arbitrary structure that doesn't implement <see cref="ITranslatable"/> (e.g. translating a complex collection)
+    /// </summary>
+    /// <param name="translator">the translator</param>
+    /// <param name="objectToTranslate">the object to translate</param>
+    internal delegate void ObjectTranslator<T>(ITranslator translator, ref T objectToTranslate);
+
+    /// <summary>
     /// This delegate is used to create arbitrary collection types for serialization.
     /// </summary>
     /// <typeparam name="T">The type of dictionary to be created.</typeparam>
@@ -161,7 +168,7 @@ namespace Microsoft.Build.BackEnd
         /// <param name="list">The list to be translated.</param>
         /// <param name="factory">factory to create type T</param>
         /// <typeparam name="T">A TaskItemType</typeparam>
-        void Translate<T>(ref List<T> list, NodePacketValueFactory<T> factory) where T : ITranslatable;
+        void Translate<T>(ref List<T> list, ObjectTranslator<T> objectTranslator);
 
         /// <summary>
         /// Translates a list of T where T implements INodePacketTranslateable using a collection factory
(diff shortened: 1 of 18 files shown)


Forgind:
I like delegates, but is there any reason ITranslatable can't move to Microsoft.Build.Framework?

dsplaisted:
> I like delegates, but is there any reason ITranslatable can't move to Microsoft.Build.Framework?

Microsoft.Build.Framework is our public API and we can't make breaking changes to it.  We'd have to pull ITranslator down together with it, and then we could no longer make API breaking changes to it.

src/Shared/ITranslator.cs
```diff
@@ -16,6 +16,13 @@ namespace Microsoft.Build.BackEnd
     /// <typeparam name="T">The type to be translated.</typeparam>
     internal delegate T NodePacketValueFactory<T>(ITranslator translator);
 
+    /// <summary>
+    /// Delegate for users that want to translate an arbitrary structure that doesn't implement <see cref="ITranslatable"/> (e.g. translating a complex collection)
+    /// </summary>
+    /// <param name="translator">the translator</param>
+    /// <param name="objectToTranslate">the object to translate</param>
+    internal delegate void ObjectTranslator<T>(ITranslator translator, ref T objectToTranslate);
```
cdmihai, line 24: Why not delete `NodePacketValueFactory` and replace it with `ObjectTranslator`?

cdmihai:
I guess I'm a little bit confused how this enables types in Microsoft.Build.Framework to be translated without having them know about ITranslator or ITranslatable. Those new types wouldn't be able to provide ObjectTranslator delegates, because the delegate signature takes in an ITranslatable. Nor can they use the new TranslatorHelpers since it also depends on ITranslator. Can you please provide an example on how this would work on a new type in Microsoft.Build.Framework that needs translation?

src/Shared/ITranslator.cs
dsplaisted: There's lots of code which already uses NodePacketValueFactory.  A type that implements ITranslatable either needs to have a default constructor or have something like NodePacketValueFactory.  I wasn't sure I wanted to commit to saying that all those types need to support default constructors.

dsplaisted:
@cdmihai the delegate takes an `ITranslator`, not an `ITranslatable`:

```c#
internal delegate void ObjectTranslator<T>(ITranslator translator, ref T objectToTranslate);
```

Here's an example of a helper method that can be implemented in the engine to translate a "POCO" from Microsoft.Build.Framework:

```c#
        public static void Translate(this ITranslator t, ref SdkReference sdkReference)
        {
            string name = null;
            string version = null;
            string minimumVersion = null;

            if (t.Mode == TranslationDirection.WriteToStream)
            {
                name = sdkReference.Name;
                version = sdkReference.Version;
                minimumVersion = sdkReference.MinimumVersion;
            }

            t.Translate(ref name);
            t.Translate(ref version);
            t.Translate(ref minimumVersion);

            if (t.Mode == TranslationDirection.ReadFromStream)
            {
                sdkReference = new SdkReference(name, version, minimumVersion);
            }
        }
```

src/Shared/ITranslator.cs
cdmihai: But they can just as easily provide an ObjectTranslator, instead of NodePacketValueFactory, right? ObjectTranslator seems to me to be able to completely replace NodePacketValueFactory.

cdmihai:
Ok, so translation logic will be in Microsoft.Build, not Microsoft.Build.Framework. And ObjectTranslator is a convenient new delegate, equivalent to NodePacketValueFactory. The benefits over NodePacketValueFactory are that it avoids having to define an extra class / closure to hold the reference for the instance that needs translating, and it has a better name.

Forgind:
Heads-up @dsplaisted:
I resolved a merge conflict with https://github.com/microsoft/msbuild/pull/5270. It wasn't too complicated, but let me know if I messed anything up.

dsplaisted:
These changes are also part of #5269, so we will probably just use that PR.  I also resolved the conflict there.

rainersigwald:
Closing in favor of #5269; if you want to go in parts I'd be fine with that too.

PR #5226, opened 2020-04-03
Fast path when evaluating string.PadLeft(int, char)

KirillOsenkov:
The previous fast path was only expecting the args of int and string, but was actually given int and char. So the fast path was not taken for expressions such as:

$(VersionSuffixBuildOfTheDay.PadLeft(2, $([System.Convert]::ToChar(`0`))))

Try to interpret int and char as int and string, this ensures the fast path is taken.

This fixes a first-chance MissingMethodException.

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -4251,6 +4251,11 @@ namespace Microsoft.Build.Evaluation
                 }
 
                 arg1 = args[1] as string;
+                if (arg1 == null && args[1] is char ch)
+                {
+                    arg1 = ch.ToString();
+                }
+
                 if (TryConvertToInt(args[0], out arg0) &&
                     arg1 != null)
                 {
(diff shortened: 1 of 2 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -4251,6 +4251,11 @@ private static bool TryGetArgs(object[] args, out int arg0, out string arg1)
                 }
 
                 arg1 = args[1] as string;
+                if (arg1 == null && args[1] is char ch)
```
cdmihai, line 4254: not affecting this PR: At some point somebody has to extract all this C# interpretation code outside of Expander.cs and into its own class :)

KirillOsenkov: At some point somebody should rewrite the Evaluator and kill this class altogether ;)

cdmihai: They mythical "some point" when everything is made nice :)

PR #5234, opened 2020-04-03
Adding target framework intrinsic functions

sfoslund:


sfoslund:
@rainersigwald this will add a dependency on nuget. Do you have guidance on how to get that hooked up properly?

src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2800,34 +2800,23 @@ public void PropertyFunctionVersionComparisonsFailsWithInvalidArguments(string b
             var expander = new Expander<ProjectPropertyInstance, ProjectItemInstance>(pg, FileSystems.Default);
             string expectedMessage = ResourceUtilities.GetResourceString("InvalidVersionFormat");
 
-            AssertThrows($"$([MSBuild]::VersionGreaterThan('{badVersion}', '1.0.0'))");
-            AssertThrows($"$([MSBuild]::VersionGreaterThan('1.0.0', '{badVersion}'))");
+            AssertThrows(expander, $"$([MSBuild]::VersionGreaterThan('{badVersion}', '1.0.0'))", expectedMessage);
```
Forgind, line 2803: Why did you have to change all these? It'll make `git blame` harder.

eng/Packages.props
```diff
@@ -18,6 +18,7 @@
     <PackageReference Update="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="1.16.30" />
     <PackageReference Update="Microsoft.Win32.Registry" Version="4.3.0" />
     <PackageReference Update="NuGet.Build.Tasks" Version="$(NuGetBuildTasksVersion)" />
+    <PackageReference Update="NuGet.Frameworks" Version="$(NuGetBuildTasksVersion)" />
```
Forgind, line 21: Is this already in VS? I imagine so, but I haven't checked.

src/Build.UnitTests/Evaluation/Expander_Tests.cs
sfoslund: Reusing code. I had to change the function signature to make it available for more than one test.

eng/Packages.props
sfoslund: Can you clarify what you mean? AFAIK MSBuild doesn't take a dependency on Nuget right now, this will introduce that.

Forgind: I found it [here](https://dev.azure.com/devdiv/DevDiv/_git/VS?path=%2Fsrc%2FProductData%2FEngineering%2FExternalWhiteList.csv), so I think it's ok. It might be good to merge this into an `exp/` branch before we merge it into master to make sure RPS doesn't complain about adding an assembly.

This was my issue with the solution filters—nothing in VS had used System.Text.Json previously, so I needed to introduce it, which also required updating the versions of other assemblies, which entangled my change with work Roslyn had to do, which meant that it sat forever in limbo. I don't think you'll hit that problem, but it would be good to make sure.

src/Build.UnitTests/Evaluation/Expander_Tests.cs
Forgind: I mean, I do think it's good to reuse code, but the three instances I see where you used your new method, you started the method with the exact same two lines. Why not put those in the same method? Note that this is a test, so it doesn't have to be optimized. (And having theories the way you (and nguerrera) did effectively does the same thing.)

eng/Packages.props
rainersigwald: Yes, we'll definitely have VS Insertion Fun™️ with this one. Especially because I think NuGet places itself in different relative locations between .NET SDK and VS. Let's punt this for now and try to sync on it when we get a chance, though.

rainersigwald, line 21: Since this reference is used for dev code and not just test/bootstrap building, we'll need to add a darc subscription. This will be . . . unfun, because we'll have to juggle it across the many NuGet/VS/MSBuild releases between now and .NET 5 GA. However, we may be forced to avoid this by the VS Setup problems--if they're only resolvable via reflection-loading NuGet we don't have to have the direct reference anyway.

rainersigwald: A quick and easy test: can you use `scripts\Deploy-MSBuild.ps1` to deploy the output of on top of a VS installation (sacrificial! don't use your primary VS). Does that work?

src/Build.UnitTests/Evaluation/Expander_Tests.cs
rainersigwald: I don't have a concern with this change. It might be a good idea to make a class-level `expander` for use in multiple tests if they all need the same one (like for all the ones that don't need properties and don't touch the filesystem). But that can wait for a test-perf effort.

src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -480,6 +481,21 @@ internal static bool VersionLessThanOrEquals(string a, string b)
...
+        {
+            return NuGetFramework.Parse(tfm).Framework;
+        }
+
+        internal static string GetTargetFrameworkVersion(string tfm)
+        {
+            return NuGetFramework.Parse(tfm).Version.ToString(2);
+        }
+
+        internal static bool IsTargetFrameworkCompatible(string tfm1, string tfm2)
```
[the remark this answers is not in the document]
sfoslund: I don't know I'm not familiar with `AssetTargetFallback`. What does that entail?

eng/Packages.props
Forgind: And while you're at it, you can try it with your SDK folder, adding `-runtime Core`.

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
Forgind: [comment in code](https://github.com/microsoft/msbuild/blob/master/src/MSBuild/MSBuild/Microsoft.Build.CommonTypes.xsd#L1637)
[docs](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#assettargetfallback)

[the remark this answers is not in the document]
sfoslund: [NuGetFrameworkUtility.IsCompatibleWithFallbackCheck](https://github.com/NuGet/NuGet.Client/blob/3803820961f4d61c06d07b179dab1d0439ec0d91/src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkUtility.cs#L122) does check for compatibility with fallback TFMs. I'm not sure about AssetTargetFallback specifically but it makes sense to me to be consistent with the way NuGet determines compatibility.

eng/Packages.props
sfoslund: No, didn't work. I'm getting `error MSB4184: The expression "[MSBuild]::GetTargetFrameworkIdentifier(netcoreapp3.1)" cannot be evaluated. Could not load file or assembly 'NuGet.Frameworks, Version=5.4.0.2, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.`

When I try to add the `-runtime Core` option VS can't load projects with the error `Could not load file or assembly 'Microsoft.VisualStudio.VC.RecommendationProvider, Version=16.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.`.

Forgind: `-runtime Core` has to go into your sdk folder and be called through `dotnet build` rather than into your VS folder, so something like `C:\Program Files\dotnet\sdk\3.1.200-preview-015002`.

rainersigwald: microsoft/msbuild#5245: we should have docs on the script, it's not obvious

Unfortunately, that's the failure I expected from the VS scenario. Let's sync tomorrow about causes/possible solutions.

Forgind: Include me?
And do you want me to send you the video about adding an assembly to VS? I imagine this is quite different, since the assembly is coming from a different source, and VS already seems to have some basic conception of it, but it might still be helpful.

sfoslund: Sure, if you think it's helpful you can send it over.

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
rainersigwald: We'll have to check with NuGet folks about the shape of the API. I was envisioning something like `IsTargetFrameworkCompatible(candidate, ideal, fallback1, fallback2, fallback3)`. So you could do something like `IsTargetFrameworkCompatible($(thisprojecttfm), $(whatever), $(AssetTargetFallback)`?

But maybe this is a bad idea or unnecessary, not sure.

sfoslund:
@rainersigwald I've pushed my first pass at using reflection. I've tested in VS using `scripts\Deploy-MSBuild.ps1` and that's working. It's a little messy so let me know where you want this code to go.

src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -32,6 +32,23 @@ internal static class IntrinsicFunctions
...
 
+        /// <summary>
+        /// Resolve the location of the NuGet.Frameworks assembly
+        /// </summary>
+        private static Assembly GetNuGetAssembly()
+        {
+            var assemblyDirectory = IsRunningFromVisualStudio() ?
+                Path.Combine(GetVsInstallRoot(), "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet") :
+                BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory;
+            return Assembly.LoadFile(Path.Combine(assemblyDirectory, "NuGet.Frameworks.dll"));
```
rainersigwald, line 43: What's the behavior if this fails? Can we present a nice error like "hey, couldn't find NuGet"?

src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -32,6 +32,23 @@ internal static class IntrinsicFunctions
...
+        }
+
+        /// <summary>
+        ///  NuGet Types
+        /// </summary>
+        private static Type NuGetFramework = GetNuGetAssembly().GetType("NuGet.Frameworks.NuGetFramework");
+        private static Type NuGetFrameworkUtility = GetNuGetAssembly().GetType("NuGet.Frameworks.NuGetFrameworkUtility");
```
rainersigwald, line 50: Can you make these lazies instead of statics, so that we don't have to pay reflection cost on load for projects that don't use the new property functions?

src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -480,6 +497,26 @@ internal static bool VersionLessThanOrEquals(string a, string b)
             return SimpleVersion.Parse(a) <= SimpleVersion.Parse(b);
         }
 
+        private static object ParseNuGetFramework(string tfm)
+        {
+            return NuGetFramework.GetMethod("Parse", new Type[] { typeof(string) }).Invoke(null, new object[] { tfm });
```
rainersigwald, line 502: Can the `GetMethod`-like calls be cached? Feels like they should be but I don't have perf numbers to back it up so it can definitely wait.

src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -480,6 +497,26 @@ internal static bool VersionLessThanOrEquals(string a, string b)
             return SimpleVersion.Parse(a) <= SimpleVersion.Parse(b);
         }
 
+        private static object ParseNuGetFramework(string tfm)
```
rainersigwald, line 500: Putting these here exposes them as `[MSBuild]::` functions. That's what a lot of our stuff does. Should we consider making it more like `[NuGet]::`? That's a bigger change for not a ton of benefit. I think I lean toward "nah".

eng/TestMSBuild.targets
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <!--
+  Copies the NuGet dll's required for testing to testing folders
+   -->
+  <Target Name="CopyNuGetDependencies" AfterTargets="ResolvePackageAssets">
+    <ItemGroup>
+        <_NuGetDependencies Include="$(RepoRoot).dotnet\sdk\$(DotNetCliVersion)\Sdks\Microsoft.NET.Sdk\tools\net472\NuGet.Frameworks.dll" />
```
rainersigwald, line 7: Should these instead have a dev-only (`PrivateAssets=all`?) reference to NuGet packages, and let that copy?

src/Build/Resources/Strings.resx
```diff
@@ -53,214 +112,214 @@
...
     <value>Importing the file "{0}" into the file "{1}" results in a circular dependency.</value>
     <comment>
        {0} is a file imported into the file "{1}" such that it results in a circular dependency. For e.g. if t1.targets imports
        t2.targets and t2.targets tries to import t1.targets, then it results in a circular dependency.
     </comment>
   </data>
-  <data name="SearchPathsForMSBuildExtensionsPath">
-      <value>Search paths being used for {0} are {1}</value>
+  <data name="SearchPathsForMSBuildExtensionsPath" xml:space="preserve">
+    <value>Search paths being used for {0} are {1}</value>
```
rainersigwald, line 317: Did you edit this file with VS? Could I get you to revert it and go back and manually edit it? Lots of probably-not-needed changes in here.

nkolev92:
I'll take a look. 
Tagging some people that might be interested as well. 

@zkat @zivkan @rrelyea

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
nkolev92: IsCompatibleWithFallbackCheck does not consider ATF. 

We prefer that ATF remains a 2 methods approach because you need to ensure that it's only applied when necessary. 
Allows you to raise warnings and related better. 

We usually use NuGetFrameworkUtility.GetNearest for everything, but you probably just need to call IsCompatible. 

We basically do a 2 pass: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/GetReferenceNearestTargetFrameworkTask.cs. 

While it does feel like it'd be nice to have an ATF intrinsic, what's a real scenario for that?

eng/Packages.props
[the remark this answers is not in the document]
nkolev92: nvm you are doing that :D This is just an "Update"

src/Build/Utilities/NuGetFrameworkWrapper.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        /// <summary>
+        /// NuGet Types
+        /// </summary>
+        private static Type NuGetFramework = GetNuGetAssembly().GetType("NuGet.Frameworks.NuGetFramework");
+        private static Type NuGetFrameworkUtility = GetNuGetAssembly().GetType("NuGet.Frameworks.NuGetFrameworkUtility");
+        private static MethodInfo ParseMethod = NuGetFramework.GetMethod("Parse", new Type[] { typeof(string) });
+        private static MethodInfo IsCompatibleMethod = NuGetFrameworkUtility.GetMethod("IsCompatibleWithFallbackCheck");
```
nkolev92, line 40: I'd just use IsCompatible. 
IsCompatibleWithFallback check from strings will never return anythign diffferent from IsCompatible.

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
rainersigwald: Seems reasonable to go without until we find a concrete need.

@sfoslund did this get considered and rejected?

> you probably just need to call IsCompatible.

src/Build/Utilities/NuGetFrameworkWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+            try
+            {
+                var assemblyDirectory = BuildEnvironmentHelper.Instance.Mode == BuildEnvironmentMode.VisualStudio ?
+                    Path.Combine(BuildEnvironmentHelper.Instance.VisualStudioInstallRootDirectory, "Common7", "IDE", "CommonExtensions", "Microsoft", "NuGet") :
+                    BuildEnvironmentHelper.Instance.CurrentMSBuildToolsDirectory;
+                return Assembly.LoadFile(Path.Combine(assemblyDirectory, "NuGet.Frameworks.dll"));
+            }
+            catch
+            {
+                throw new InternalErrorException(AssemblyResources.GetString("NuGetAssemblyNotFound"));
```
rainersigwald, line 30: Worth passing the path where we looked for it to the error? Might speed up an investigation if it's hit.

src/Build/Utilities/NuGetFrameworkWrapper.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Wraps the NuGet.Frameworks assembly, which is referenced by reflection.
+    /// </summary>
+    internal class NuGetFrameworkWrapper
+    {
+        /// <summary>
+        /// Resolve the location of the NuGet.Frameworks assembly
+        /// </summary>
+        private static Assembly GetNuGetAssembly()
```
rainersigwald, line 19: Should this be in a static ctor so it's called only once? Right now it'll be called for each exposed type, right?

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
sfoslund: No, I switched over to IsCompatible: https://github.com/microsoft/msbuild/pull/5234/files#diff-5af5289df4605c1a0a27f11cb43a8d23R41

dsplaisted:
@sfoslund @rainersigwald Should this be retargeted to the 16.7 branch?

src/Build/Evaluation/IntrinsicFunctions.cs
Nirmal4G: All these are dotnet related, not just NuGet specific, so why not under `[DotNet]::`. It is easy to understand that way!

Putting it under MSBuild, atleast for me, is not clear, I read `IsCompatibleWith` and I'm confused as to, for what part of the tech stack is it comparing to, is it .NET? VC? PHP? etc...

Forgind:
> @sfoslund Sarah Oslund FTE @rainersigwald Rainer Sigwald FTE Should this be retargeted to the 16.7 branch?

master currently goes to VS 16.7.

sfoslund:
@rainersigwald I think I'm caught up on feedback here, unless we want to switch to `[NuGet]::` instead of `[MSBuild]::`. Do you have anything else?

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
rainersigwald: Ah, I was confused because the local name is still `IsCompatibleWithFallbackCheck`. Should that match?

src/MSBuild.UnitTests/Microsoft.Build.CommandLine.UnitTests.csproj
```diff
@@ -73,5 +73,5 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
-
+
```
rainersigwald, line 76: nit: revert whitespace

src/Build/Evaluation/IntrinsicFunctions.cs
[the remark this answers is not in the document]
sfoslund: Yup, it does now :)

sfoslund:
@rainersigwald sounds good, just pushed those updates. 😄

src/Build/Evaluation/IntrinsicFunctions.cs
```diff
@@ -480,6 +480,21 @@ internal static bool VersionLessThanOrEquals(string a, string b)
...
+        {
+            return NuGetFramework.Value.GetTargetFrameworkIdentifier(tfm);
+        }
+
+        internal static string GetTargetFrameworkVersion(string tfm)
+        {
+            return NuGetFramework.Value.GetTargetFrameworkVersion(tfm);
+        }
+
+        internal static bool IsTargetFrameworkCompatible(string tfm1, string tfm2)
```
nkolev92, line 493: nit: Maybe use the descriptive names?
Compatibility is a one-way relationship the order matters. 

Call the target and candidate, similar to what the NuGet API calls them?

sfoslund: Sure, just made that change. (Had to force push since the PR didn't pick it up when I updated the branch the first time).

PR #5235, opened 2020-04-03
Allow specifying task parameter logging options via UsingTask.

KirillOsenkov:
This enables fine-grained control over whether:
 * to log log each parameter (whether input or output)
 * or whether to log item metadata for each ITaskItem[] parameter.

When LogTaskInputs is set the default behavior is still to log all parameters and all item metadata for ITaskItem[] parameters. Since this is very verbose and hurts performance without adding any useful informations it is valuable to be able to turn this logging off in certain situations.

This approach allows controlling logging via an extension to the UsingTask XML element. Even for the default AssemblyTaskFactory a ParameterGroup child element is now allowed, with subelements for parameters that can carry the new Log attribute or the new LogTask.

I've identified the specific tasks and parameters that we want to restrict logging for that would give us the most gains without losing any significant useful info:

https://github.com/KirillOsenkov/MSBuildStructuredLog/wiki/Task-Parameter-Logging

### DIFF
--- a/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
@@ -27,14 +27,6 @@ namespace Microsoft.Build.BackEnd
         /// </summary>
         internal static int parameterLimit = 200;
 
-        /// <summary>
-        /// Gets a text serialized value of a parameter for logging.
-        /// </summary>
-        internal static string GetParameterText(string prefix, string parameterName, params object[] parameterValues)
-        {
-            return GetParameterText(prefix, parameterName, (IList)parameterValues);
-        }
-
         internal static string ItemGroupIncludeLogMessagePrefix = ResourceUtilities.GetResourceString("ItemGroupIncludeLogMessagePrefix");
         internal static string ItemGroupRemoveLogMessage = ResourceUtilities.GetResourceString("ItemGroupRemoveLogMessage");
         internal static string OutputItemParameterMessagePrefix = ResourceUtilities.GetResourceString("OutputItemParameterMessagePrefix");
@@ -43,7 +35,7 @@ namespace Microsoft.Build.BackEnd
         /// <summary>
         /// Gets a text serialized value of a parameter for logging.
         /// </summary>
-        internal static string GetParameterText(string prefix, string parameterName, IList parameterValue)
+        internal static string GetParameterText(string prefix, string parameterName, IList parameterValue, bool logItemMetadata = true)
         {
             if (parameterValue == null || parameterValue.Count == 0)
             {
@@ -94,7 +86,7 @@ namespace Microsoft.Build.BackEnd
                         sb.Append("        ");
                     }
 
-                    AppendStringFromParameterValue(sb, parameterValue[i]);
+                    AppendStringFromParameterValue(sb, parameterValue[i], logItemMetadata);
 
                     if (!specialTreatmentForSingle && i < parameterValue.Count - 1)
                     {
@@ -118,7 +110,7 @@ namespace Microsoft.Build.BackEnd
         /// First line is already indented.
         /// Indent of any subsequent line should be 12 spaces.
         /// </summary>
-        internal static string GetStringFromParameterValue(object parameterValue)
+        internal static string GetStringFromParameterValue(object parameterValue, bool logItemMetadata = true)
         {
             // fast path for the common case
             if (parameterValue is string valueText)
@@ -128,7 +120,7 @@ namespace Microsoft.Build.BackEnd
 
             using (var sb = new ReuseableStringBuilder())
             {
-                AppendStringFromParameterValue(sb, parameterValue);
+                AppendStringFromParameterValue(sb, parameterValue, logItemMetadata);
                 return sb.ToString();
             }
         }
@@ -138,7 +130,7 @@ namespace Microsoft.Build.BackEnd
         [ThreadStatic]
         private static List<KeyValuePair<string, string>> keyValuePairList;
 
-        private static void AppendStringFromParameterValue(ReuseableStringBuilder sb, object parameterValue)
+        private static void AppendStringFromParameterValue(ReuseableStringBuilder sb, object parameterValue, bool logItemMetadata = true)
         {
             if (parameterValue is string text)
             {
@@ -147,6 +139,11 @@ namespace Microsoft.Build.BackEnd
             else if (parameterValue is ITaskItem item)
             {
                 sb.Append(item.ItemSpec);
+
+                if (!logItemMetadata)
+                {
+                    return;
+                }
 
                 var customMetadata = item.CloneCustomMetadata();
                 int count = customMetadata.Count;
--- a/src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
+++ b/src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
@@ -430,12 +430,12 @@ namespace Microsoft.Build.BackEnd
                 if (TaskParameterTypeVerifier.IsAssignableToITask(type))
                 {
                     ITaskItem[] outputs = GetItemOutputs(parameter);
-                    GatherTaskItemOutputs(outputTargetIsItem, outputTargetName, outputs, parameterLocation);
+                    GatherTaskItemOutputs(outputTargetIsItem, outputTargetName, outputs, parameterLocation, parameter);
                 }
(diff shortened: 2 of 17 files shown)


KirillOsenkov:
This is how we can specify Log and LogMetadata attributes on the `UsingTask`:

image

KirillOsenkov:
Before:
image

After:
image

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
```diff
@@ -114,7 +114,7 @@ internal static string GetStringFromParameterValue(object parameterValue)
 
                 var customMetadata = item.CloneCustomMetadata();
```
Forgind, line 115: It would be nice to avoid this, too, when logItemMetadata is false.

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
```diff
@@ -1283,10 +1283,13 @@ private bool InitializeTaskVectorParameter
         /// </remarks>
         private bool InternalSetTaskParameter(TaskPropertyInfo parameter, IList parameterValue)
         {
-            if (LogTaskInputs && !_taskLoggingContext.LoggingService.OnlyLogCriticalEvents && parameterValue.Count > 0)
+            if (LogTaskInputs &&
```
Forgind, line 1286: Rather than threading parameter through as you did here, could we instead add a (private) set method to LogTaskInputs? The only potential drawback I see is that that would set it for a TaskExecutionHost rather than setting it for a particular task, but you could presumably add something that changes it and reverts it to its previous value afterwards, and that would still consolidate this logic in one place.

If you want to add the option to log inputs but not their metadata, you could also change LogTaskInputs into an enum with options like `LogAllInputs`, `LogInputsWithoutMetadata`, and `DoNotLogInputs`. Since TaskExecutionHost is internal, I don't think that's a breaking change.

src/Build/Construction/UsingTaskParameterGroupElement.cs
```diff
@@ -129,12 +129,15 @@ private static void VerifyCorrectParent(ProjectElementContainer parent)
...
-            // Now since there is not goign to be a TaskElement on the using task we need to validate and make sure there is a TaskFactory attribute on the parent element and 
+            // Now since there is not going to be a TaskElement on the using task we need to validate and make sure there is a TaskFactory attribute on the parent element and 
             // that it is not empty
             if (parentUsingTask.TaskFactory.Length == 0)
             {
                 ErrorUtilities.VerifyThrow(parentUsingTask.Link == null, "TaskFactory");
-                ProjectXmlUtilities.VerifyThrowProjectRequiredAttribute(parent.XmlElement, "TaskFactory");
+
+                // Parameters under UsingTask is now allowed even if TaskFactory is not specified on it.
+                // This is used to set Log or LogItemMetadata on individual parameters.
```
Forgind, line 139: Can probably just remove these three lines. I assume you made it not need TaskFactory so anyone could specify the two parameters you added?

src/Build/Xml/ProjectXmlUtilities.cs
```diff
@@ -42,6 +42,21 @@ internal static void VerifyThrowProjectNoChildElements(XmlElementWithLocation el
             }
         }
 
+        /// <summary>
+        /// Throw an invalid project exception if there are any child elements with a different name than expected
+        /// </summary>
+        internal static void VerifyThrowProjectOnlyAllowedChildrenWithName(XmlElementWithLocation element, string allowedChildName)
```
Forgind, line 48: I'm a little confused by this. All children need to have a single name? I thought there were two you cared about? Maybe add a test?

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: Once https://github.com/microsoft/msbuild/pull/5224 is merged, I'll have to fix merge conflicts and will fix it then :)

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
KirillOsenkov: Yes, I think LogTaskInputs needs reworking. It has various issues, e.g. users have no control over it because it's coupled with verbosity. 

I didn't quite understand which set method you're suggesting to add and where? I think we want to control whether to log item metadata per task.

src/Build/Construction/UsingTaskParameterGroupElement.cs
KirillOsenkov: Right. Didn't want to be too destructive and wanted to leave a historical record here ;) This is one thing about this PR that might be potentially worrisome as we are effectively relaxing the XML schema a bit.

src/Build/Xml/ProjectXmlUtilities.cs
KirillOsenkov: Yeah basically there was a test that verified that the node had no children in case there wasn't a TaskFactory specified. Now I'm relaxing the test to verify that it only has a child of a given name (Parameters). Since we still want to disallow the actual body, but we do want to allow Parameters.

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
Forgind: Looking at this again, I think my method wouldn't work because it would change logging behavior for all tasks executing in parallel on a TaskExecutionHost. We could presumably make separate TaskExecutionHosts, each of which has a unique value for LogTaskInputs and never executes a task that doesn't agree with their version, but that would be a heavyweight solution.

Although I no longer think it's a great solution, here's what I was thinking:
protected set method [here](https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/TaskExecutionHost/AddInParts/ITaskExecutionHost.cs#L58)
Choose a couple key places (like [here](https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs#L429), for instance) where you can set LogTaskInputs to a new value (with LogTaskInputs being an enum including whether to log inputs and whether to log metadata). Then you can reset it in a finally block [here](https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs#L510). If there's only one thread running on a TaskExecutionHost at a time, or they all have the same logging behavior, that would mean you wouldn't have to thread it through so many function calls.

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
```diff
@@ -1283,10 +1283,13 @@ private bool InitializeTaskVectorParameter
...
+                !_taskLoggingContext.LoggingService.OnlyLogCriticalEvents &&
+                parameterValue.Count > 0 &&
+                parameter.Log)
```
cdmihai, line 1289: For the sake of debuggability , it would be nice to log a comment whenever an item or metadata isn't because these two new knobs are false. And maybe add an escape hatch to ignore them and log everything (maybe reuse the one here [https://github.com/microsoft/msbuild/pull/5210/files#diff-5f4088c7b8512fa52f7faf88f89661f3R122](https://github.com/microsoft/msbuild/pull/5210/files#diff-5f4088c7b8512fa52f7faf88f89661f3R122))

src/Build/BackEnd/Components/RequestBuilder/IntrinsicTasks/ItemGroupLoggingHelper.cs
KirillOsenkov: Fixed.

src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs
KirillOsenkov: We can always add this in the viewer post-build, since we know declaratively which tasks have parameters with logging disabled or constrained. No need to log. E.g. for each Hash task we could easily add an icon or a message: this parameter's item metadata was not logged.

KirillOsenkov: And if people are using text logs, it's their fault. (no offense @rainersigwald)

KirillOsenkov:
Here's another approach, via simple properties:
https://github.com/microsoft/msbuild/pull/5268

rainersigwald:
Team triage: closing in favor of #5268.

PR #5236, opened 2020-04-03
Allow overriding task parameter logging options in task classes

KirillOsenkov:
This is an alternative approach to https://github.com/microsoft/msbuild/pull/5235.

Instead of specifying the task parameter logging options in UsingTask XML, we introduce an internal virtual method on TaskExtension to let implementations override the defaults.

A new internal interface has to be added to Framework to connect the task implementations to the TaskLoggingHelper that needs to consume the new method.

### DIFF
--- a/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
+++ b/ref/Microsoft.Build.Tasks.Core/net/Microsoft.Build.Tasks.Core.cs
@@ -171,6 +171,7 @@ namespace Microsoft.Build.Tasks
         [Microsoft.Build.Framework.RequiredAttribute]
         public Microsoft.Build.Framework.ITaskItem[] Paths { get { throw null; } set { } }
         public override bool Execute() { throw null; }
+        protected override Microsoft.Build.Utilities.Task.ParameterLoggingOptions GetParameterLoggingOptions(string parameterName) { throw null; }
     }
     public partial class Copy : Microsoft.Build.Tasks.TaskExtension, Microsoft.Build.Framework.ICancelableTask, Microsoft.Build.Framework.ITask
     {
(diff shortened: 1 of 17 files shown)


cdmihai:
How about keeping both mechanisms? Tasks that no for sure those items or metadata do not add any logging value can remove them via this PR. For tasks that cannot be changed, there's #5235

KirillOsenkov:
I don't mind ;) Will fix conflicts shortly.

Was curious what the team thinks since I'm not 100% on either approach.

KirillOsenkov:
I'm thinking I like the UsingTask approach much better as it's more extensible and doesn't require to change the task classes.

https://github.com/microsoft/msbuild/pull/5235

Let's keep this PR around for a bit but if all goes well I'll just abandon it.

cdmihai:
We discussed it a bit in the PR review today. Using task is seen as not that great, because it's hard for the user to override UsingTask elements. Unfortunately first one wins, not last one wins. This assumes the scenario where the user wants to tone down a task and has to add using task to do so. He has to make sure his using task element is the first.

Regarding the TaskExtension approach, I don't remember any points against it, other than it should be in Task, not TaskExtension.

I'll invite you to the next PR review so we can brain storm.

KirillOsenkov:
I think there was a reason it is on `TaskExtension`, not `Task`. Task is in `Microsoft.Build.Utilities.Core` and TaskExtension is in `Microsoft.Build.Tasks.Core`. I think it was IVT related...

Assemblies

KirillOsenkov:
Also I honestly don't foresee users needing to change that all that often - the mechanism is predominantly for us, and I don't even think we'll need to extend it in the future, maybe just for NuGet tasks. And for that we can just do that at the original UsingTask.

KirillOsenkov:
Here's another approach, via simple properties:
#5268

rainersigwald:
Team triage: closing in favor of #5268.

Issue #2308, opened 2017-07-16
BaseIntermediateOutputPath for C++ projects


kokole:
Why isn't this supported?

am11:
`vcxproj`s support `IntDir` for the same purpose, for instance:

```xml
<PropertyGroup>
  <OutDir>$(SolutionDir)bin\$(Configuration)\</OutDir>
  <IntDir>$(OutDir)obj\</IntDir>
</PropertyGroup>
```

In `csproj`, `fsproj and` `vbproj` etc., if we try to use these, they get overridden by the commons. The consumer-facing properties are `OutputDir` and `BaseIntermediateOutputPath`.

Though not a show-stopper, it would be nice to have unified syntax across the board: either `OutputDir` / `BaseIntermediateOutputPath` or the shorter ones; `OutDir` / `IntDir`.

puxu-msft:
The problem still exists on VS 2022, I debugged with this for hours, and found the definition in `Microsoft/VC/v170/Microsoft.Cpp.MSVC.Toolset.Common.props`

```xml
<IntDir Condition="'$(IntDir)'=='' AND '$(IntermediateOutputPath)'!=''">$(IntermediateOutputPath)</IntDir>
<IntDir Condition="'$(IntDir)'=='' AND '$(IntermediateOutputPath)'=='' and '$(Platform)' == 'Win32'">$(Configuration)\</IntDir>
<IntDir Condition="'$(IntDir)'=='' AND '$(IntermediateOutputPath)'=='' and '$(Platform)' != 'Win32'">$(Platform)\$(Configuration)\</IntDir>
```

Before this, there is no process from `BaseIntermediateOutputPath` to `IntermediateOutputPath`. And I found no helpful pages in MSDN..

Nirmal4G:
VC Team should update their targets to remove usages of `IntermediateOutputPath`! This property is no longer present in Common props. The `BaseIntermediateOutputPath` is in Common props but I'm proposing to it move it into targets and replace it with a common `BuildDir` since we already have a `PublishDir` (_should be promoted to top-level folder instead!_). This will be a breaking change but one that I think is beneficial in the long run.

PR #5238, opened 2020-04-05
Add BaseOutputPath to common targets

### LINKED ISSUE #1664: BaseOutputPath is not respected as per the documentation
Going back to Visual Studio 2008 (and up to the current version) it is documented that `BaseOutputPath` is a "Common MSBuild Project Property".

Specifically it is specified:
> Specifies the base path for the output file. If it is set, MSBuild will use OutputPath = $(BaseOutputPath)\$(Configuration)\. 

However, `BaseOutputPath` is not actually defined or used anywhere in the Common props or targets files.

Nirmal4G:
While this is not officially supported on non-sdk projects, NuGet Pack targets need a common output path to put '.nupkg' across both type of projects.

Required for NuGet/Home#9234 in PR NuGet/NuGet.Client#3270

Also Fixes #1664

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -144,17 +144,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
     <ConfigurationName Condition="'$(ConfigurationName)' == ''">$(Configuration)</ConfigurationName>
 
-    <BaseOutputPath Condition="'$(BaseOutputPath)' == ''">bin\</BaseOutputPath>
-    <BaseOutputPath Condition="!HasTrailingSlash('$(BaseOutputPath)')">$(BaseOutputPath)\</BaseOutputPath>
+    <BaseOutputPath>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(BaseOutputPath)', 'bin'))))</BaseOutputPath>
     <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' == 'AnyCPU'">$(BaseOutputPath)$(Configuration)\</OutputPath>
     <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' != 'AnyCPU'">$(BaseOutputPath)$(PlatformName)\$(Configuration)\</OutputPath>
-    <OutputPath Condition="!HasTrailingSlash('$(OutputPath)')">$(OutputPath)\</OutputPath>
+    <OutputPath>$([MSBuild]::EnsureTrailingSlash('$(OutputPath)'))</OutputPath>
 
-    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)' == ''">obj\</BaseIntermediateOutputPath>
-    <BaseIntermediateOutputPath Condition="!HasTrailingSlash('$(BaseIntermediateOutputPath)')">$(BaseIntermediateOutputPath)\</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(BaseIntermediateOutputPath)', 'obj'))))</BaseIntermediateOutputPath>
     <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == '' and '$(PlatformName)' == 'AnyCPU'">$(BaseIntermediateOutputPath)$(Configuration)\</IntermediateOutputPath>
     <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == '' and '$(PlatformName)' != 'AnyCPU'">$(BaseIntermediateOutputPath)$(PlatformName)\$(Configuration)\</IntermediateOutputPath>
-    <IntermediateOutputPath Condition="!HasTrailingSlash('$(IntermediateOutputPath)')">$(IntermediateOutputPath)\</IntermediateOutputPath>
+    <IntermediateOutputPath>$([MSBuild]::EnsureTrailingSlash('$(IntermediateOutputPath)'))</IntermediateOutputPath>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -212,15 +210,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <!-- Required for enabling Team Build for packaging app package-generating projects -->
     <OutDirWasSpecified Condition=" '$(OutDir)'!='' and '$(OutDirWasSpecified)'=='' ">true</OutDirWasSpecified>
 
-    <OutDir Condition=" '$(OutDir)' == '' ">$(OutputPath)</OutDir>
     <!-- Example, bin\Debug\ -->
     <!-- Ensure OutDir has a trailing slash, so it can be concatenated -->
-    <OutDir Condition="'$(OutDir)' != '' and !HasTrailingSlash('$(OutDir)')">$(OutDir)\</OutDir>
+    <OutDir>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(OutDir)', '$(OutputPath)'))))</OutDir>
     <ProjectName Condition=" '$(ProjectName)' == '' ">$(MSBuildProjectName)</ProjectName>
     <!-- Example, MyProject -->
 
     <!-- For projects that generate app packages or ones that want a per-project output directory, update OutDir to include the project name -->
-    <OutDir Condition="'$(OutDir)' != '' and '$(OutDirWasSpecified)' == 'true' and (('$(WindowsAppContainer)' == 'true' and '$(GenerateProjectSpecificOutputFolder)' != 'false') or '$(GenerateProjectSpecificOutputFolder)' == 'true')">$(OutDir)$(ProjectName)\</OutDir>
+    <OutDir Condition="'$(OutDir)' != '' and '$(OutDirWasSpecified)' == 'true' and (('$(WindowsAppContainer)' == 'true' and '$(GenerateProjectSpecificOutputFolder)' != 'false') or '$(GenerateProjectSpecificOutputFolder)' == 'true')">$([MSBuild]::EnsureTrailingSlash('$(OutDir)$(ProjectName)'))</OutDir>
 
     <AssemblyName Condition=" '$(AssemblyName)'=='' ">$(MSBuildProjectName)</AssemblyName>
     <TargetName Condition="'$(TargetName)' == '' and '$(OutputType)' == 'winmdobj' and '$(RootNamespace)' != ''">$(RootNamespace)</TargetName>
@@ -306,7 +303,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         Condition intentionally omitted on this one, because it causes problems
         when we pick up the value of an environment variable named TargetDir
         -->
-    <TargetDir Condition="'$(OutDir)' != ''">$([MSBuild]::Escape($([System.IO.Path]::GetFullPath(`$([System.IO.Path]::Combine(`$(MSBuildProjectDirectory)`, `$(OutDir)`))`))))</TargetDir>
+    <TargetDir Condition="'$(OutDir)' != ''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(OutDir)'))</TargetDir>
 
     <!-- Example, C:\MyProjects\MyProject\bin\Debug\MyAssembly.dll -->
     <TargetPath Condition=" '$(TargetPath)' == '' ">$(TargetDir)$(TargetFileName)</TargetPath>
@@ -390,12 +387,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <PropertyGroup Condition="'$(_DebugSymbolsProduced)' == 'true' and '$(OutputType)' == 'winmdobj'">
     <WinMDExpOutputPdb Condition="'$(WinMDExpOutputPdb)' == ''">$(IntermediateOutputPath)$(TargetName).pdb</WinMDExpOutputPdb>
-    <_WinMDDebugSymbolsOutputPath>$([System.IO.Path]::Combine('$(OutDir)', $([System.IO.Path]::GetFileName('$(WinMDExpOutputPdb)'))))</_WinMDDebugSymbolsOutputPath>
+    <_WinMDDebugSymbolsOutputPath>$(OutDir)$([System.IO.Path]::GetFileName('$(WinMDExpOutputPdb)'))</_WinMDDebugSymbolsOutputPath>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(OutputType)' == 'winmdobj' and '$(DocumentationFile)'!=''">
     <WinMDOutputDocumentationFile Condition="'$(WinMDOutputDocumentationFile)' == ''">$(IntermediateOutputPath)$(TargetName).xml</WinMDOutputDocumentationFile>
-    <_WinMDDocFileOutputPath>$([System.IO.Path]::Combine('$(OutDir)', $([System.IO.Path]::GetFileName('$(WinMDOutputDocumentationFile)'))))</_WinMDDocFileOutputPath>
+    <_WinMDDocFileOutputPath>$(OutDir)$([System.IO.Path]::GetFileName('$(WinMDOutputDocumentationFile)'))</_WinMDDocFileOutputPath>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(WinMDExpOutputWindowsMetadataFilename)' != ''">
@@ -467,8 +464,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <!-- Output location for publish target. -->
   <PropertyGroup>
-    <PublishDir Condition="'$(PublishDir)' != '' and !HasTrailingSlash('$(PublishDir)')">$(PublishDir)\</PublishDir>
-    <PublishDir Condition="'$(PublishDir)'==''">$(OutputPath)app.publish\</PublishDir>
+    <PublishDir>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(PublishDir)', '$(OutputPath)app.publish'))))</PublishDir>
   </PropertyGroup>
 
   <!--
@@ -3376,7 +3372,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <PropertyGroup Condition="'$(TargetFrameworkMoniker)' != ''">
     <!-- Do not clean if we are going to default the path to the temp directory -->
     <TargetFrameworkMonikerAssemblyAttributesFileClean Condition="'$(TargetFrameworkMonikerAssemblyAttributesFileClean)' == '' and '$(TargetFrameworkMonikerAssemblyAttributesPath)' != ''">true</TargetFrameworkMonikerAssemblyAttributesFileClean>
-    <TargetFrameworkMonikerAssemblyAttributesPath Condition="'$(TargetFrameworkMonikerAssemblyAttributesPath)' == ''">$([System.IO.Path]::Combine('$(IntermediateOutputPath)','$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)'))</TargetFrameworkMonikerAssemblyAttributesPath>
+    <TargetFrameworkMonikerAssemblyAttributesPath Condition="'$(TargetFrameworkMonikerAssemblyAttributesPath)' == ''">$(IntermediateOutputPath)$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)</TargetFrameworkMonikerAssemblyAttributesPath>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -4581,7 +4577,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <Output TaskParameter="DestinationFiles" ItemName="FinalWinmdExpArtifacts" />
     </Copy>
 
-    <Message Importance="High" Text="$(MSBuildProjectName) -&gt; $([System.IO.Path]::GetFullPath('$(_WindowsMetadataOutputPath)'))" Condition="'$(SkipCopyWinMDArtifact)' != 'true' and '$(_WindowsMetadataOutputPath)' != ''" />
+    <Message Importance="High" Text="$(MSBuildProjectName) -&gt; $([MSBuild]::NormalizePath('$(_WindowsMetadataOutputPath)'))" Condition="'$(SkipCopyWinMDArtifact)' != 'true' and '$(_WindowsMetadataOutputPath)' != ''" />
 
   </Target>
 
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -326,7 +326,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <TargetRefPath Condition=" '$(TargetRefPath)' == '' and '$(ProduceReferenceAssembly)' == 'true' ">$([MSBuild]::NormalizePath($(TargetDir), 'ref', $(TargetFileName)))</TargetRefPath>
 
     <!-- Example, C:\MyProjects\MyProject\ -->
-    <ProjectDir Condition=" '$(ProjectDir)' == '' ">$([MSBuild]::EnsureTrailingSlash($(MSBuildProjectDirectory))</ProjectDir>
+    <ProjectDir Condition=" '$(ProjectDir)' == '' ">$([MSBuild]::EnsureTrailingSlash($(MSBuildProjectDirectory)))</ProjectDir>
 
     <!-- Example, C:\MyProjects\MyProject\MyProject.csproj -->
     <ProjectPath Condition=" '$(ProjectPath)' == '' ">$(ProjectDir)$(ProjectFileName)</ProjectPath>
@@ -1718,7 +1718,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <!-- Pass the CurrentProjectTargetPlatform parameter to the task only if GetReferenceNearestTargetFrameworkTaskSupportsTargetPlatformParameter is true.  This means
          that we are using a version of NuGet which supports that parameter on this task. -->
-    
+
     <GetReferenceNearestTargetFrameworkTask AnnotatedProjectReferences="@(_ProjectReferenceTargetFrameworkPossibilities)"
                                             CurrentProjectTargetFramework="$(ReferringTargetFrameworkForProjectReferences)"
                                             CurrentProjectTargetPlatform="$(TargetPlatformMoniker)"
@@ -1787,7 +1787,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <TargetPlatformMonikers>@(_TargetFrameworkInfo->'%(TargetPlatformMonikers)')</TargetPlatformMonikers>
 
         <HasSingleTargetFramework>true</HasSingleTargetFramework>
-        
+
         <!-- indicate to caller that project is RID agnostic so that a global property RuntimeIdentifier value can be removed -->
         <IsRidAgnostic>false</IsRidAgnostic>
         <IsRidAgnostic Condition=" '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == '' ">true</IsRidAgnostic>
@@ -4029,7 +4029,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       Name="_DeploymentGenerateLauncher"
       Condition="'$(GenerateClickOnceManifests)'=='true' and '$(_DeploymentLauncherBased)' == 'true'">
 
-    <!-- 
+    <!--
       If apphost based built EXE is found, use that as the Launcher.exe's entry point otherwise
       use the built DLL as the entry point
     -->
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -68,6 +68,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Import Project="Microsoft.NET.props" Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' or '$(TargetFrameworkIdentifier)' == '.NETCoreApp'"/>
 
   <PropertyGroup>
+    <!-- Generates full paths for the 'File' property in errors, warnings and messages in many targets -->
+    <GenerateFullPaths Condition="'$(GenerateFullPaths)' == ''">true</GenerateFullPaths>
     <!-- Yield optimization properties -->
     <YieldDuringToolExecution Condition="'$(YieldDuringToolExecution)' == ''">true</YieldDuringToolExecution>
   </PropertyGroup>
@@ -109,7 +111,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     OutDir can be used to gather multiple project outputs in one location. In addition,
     OutDir is included in AssemblySearchPaths used for resolving references.
 
+    BaseOutputPath:
+    This is the top level folder where all configuration specific output folders will be created.
+    Default value is bin\
+
     OutputPath:
+    This is the full Output Path, and is derived from BaseOutputPath, if none specified
+    (eg. bin\Debug). If this property is overridden, then setting BaseOutputPath has no effect.
+
+    For projects loaded with MPFProj (Legacy project system):
     This property is usually specified in the project file and is used to initialize OutDir.
     OutDir and OutputPath are distinguished for legacy reasons, and OutDir should be used if at all possible.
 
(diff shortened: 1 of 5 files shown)


Nirmal4G:
I don't know who to tag. So, @rainersigwald NuGet Pack target needs this. Can you or anyone from the team can take a look at this?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,30 +127,35 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    (eg. obj\Debug). If this property is overridden, then setting BaseIntermediateOutputPath has no effect.
+
+    Ensure any and all path property has a trailing slash, so it can be concatenated.
     -->
 
   <PropertyGroup>
-    <!-- Ensure any OutputPath has a trailing slash, so it can be concatenated -->
-    <OutputPath Condition="'$(OutputPath)' != '' and !HasTrailingSlash('$(OutputPath)')">$(OutputPath)\</OutputPath>
+    <BaseOutputPath Condition=" '$(BaseOutputPath)' == '' ">bin\</BaseOutputPath>
+    <BaseOutputPath Condition="!HasTrailingSlash('$(BaseOutputPath)')">$(BaseOutputPath)\</BaseOutputPath>
```
rainersigwald, line 137:
```suggestion
    <BaseOutputPath>$([MSBuild]::EnsureTrailingSlash('$([MSBuild]::ValueOrDefault('$(BaseOutputPath)', 'bin\'))'))</BaseOutputPath>
```

?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,30 +127,35 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
         Be careful not to give OutputPath a default value in the case of an invalid Configuration/Platform.
         We use OutputPath specifically to check for invalid configurations/platforms.
         -->
-    <OutputPath Condition=" '$(Platform)'=='' and '$(Configuration)'=='' and '$(OutputPath)'=='' ">bin\Debug\</OutputPath>
+    <OutputPath Condition=" '$(Platform)'=='' and '$(Configuration)'=='' and '$(OutputPath)'=='' ">$(BaseOutputPath)</OutputPath>
     <_OriginalConfiguration>$(Configuration)</_OriginalConfiguration>
     <_OriginalPlatform>$(Platform)</_OriginalPlatform>
     <Configuration Condition=" '$(Configuration)'=='' ">Debug</Configuration>
-    <ConfigurationName Condition=" '$(ConfigurationName)' == '' ">$(Configuration)</ConfigurationName>
+    <_Configuration Condition=" '$(_Configuration)' == '' ">$(Configuration)</_Configuration>
```
rainersigwald, line 148: What's the motivation for this change? Is it separable from the other part?

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: The lines are similar to how it was presented in the SDK targets and to `BaseIntermediateOutputPath` lines.

I have the refactored version of the targets (in a separate branch). If it's okay with you, I can create a separate PR for those changes too!

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: Are `ConfigurationName` and `PlatformName` is being used anywhere besides the Common targets?

I have looked in all the targets that shipped with Dev10 until now and I didn't find any. In the refactored version, I removed all these workarounds and used them directly.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,30 +127,35 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <BaseOutputPath Condition=" '$(BaseOutputPath)' == '' ">bin\</BaseOutputPath>
+    <BaseOutputPath Condition="!HasTrailingSlash('$(BaseOutputPath)')">$(BaseOutputPath)\</BaseOutputPath>
+
     <AssemblyName Condition=" '$(AssemblyName)'=='' ">$(MSBuildProjectName)</AssemblyName>
     <!--
         Be careful not to give OutputPath a default value in the case of an invalid Configuration/Platform.
         We use OutputPath specifically to check for invalid configurations/platforms.
         -->
-    <OutputPath Condition=" '$(Platform)'=='' and '$(Configuration)'=='' and '$(OutputPath)'=='' ">bin\Debug\</OutputPath>
+    <OutputPath Condition=" '$(Platform)'=='' and '$(Configuration)'=='' and '$(OutputPath)'=='' ">$(BaseOutputPath)</OutputPath>
```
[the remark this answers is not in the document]
Nirmal4G: I could append `Debug` there.

But do we have to? I mean, No `Configuration` or `Platform` is defined, so that means the build is happening without them, which might be single true build. So, isn't it better to place it in the root dir itself instead of creating a sub-dir?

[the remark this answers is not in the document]
Forgind: Whatever the best answer is, I'd be hesitant to change something that others can have relied on if, for instance, someone explicitly looks for `bin\Debug\...`, their code would now start failing where it hadn't previously.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: They are "public"--not prefixed with a `_`--and can be accessed by arbitrary build logic in user projects. As such, removing them is a breaking change.

src/Tasks/Microsoft.Common.CurrentVersion.targets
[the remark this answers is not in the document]
rainersigwald: Exactly: don't change behavior of existing projects without a very good reason.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: Sure, you can follow the convention here.

src/Tasks/Microsoft.Common.CurrentVersion.targets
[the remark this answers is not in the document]
Nirmal4G: I'll append `Debug` here.

It seems that Configuration is indeed set to Debug when empty.

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: OK. I'll also group the output logic together.

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: OK. I'll revert them. So, is it meant to be an alias for the original properties?!

rainersigwald: No idea what the intent was, unfortunately. But we have enough experience to be very scared of making changes.

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: I think I'd prefer inline changes without moves to make the changes super clear.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -161,6 +172,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
 
+  <PropertyGroup Condition=" $(OutputPath) == '' ">
+    <OutputPath Condition=" '$(PlatformName)' == 'AnyCPU' ">$(BaseOutputPath)$(Configuration)\</OutputPath>
+    <OutputPath Condition=" '$(PlatformName)' != 'AnyCPU' ">$(BaseOutputPath)$(PlatformName)\$(Configuration)\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup>
+    <OutputPath>$([MSBuild]::EnsureTrailingSlash('$(OutputPath)'))</OutputPath>
+  </PropertyGroup>
+
   <PropertyGroup Condition=" '$(OutputPath)' == '' ">
```
Nirmal4G, line 183: If we're allowing defaults for `OutputPath`, then, how can we check for invalid configuration/platform combinations?

1. We could check for `PlatformTarget` instead of `OutputPath` but that won't be a complete solution.
2. We could send in the available combinations, and check against current value but that'll only work with solution that defines all combinations.
3. There are no invalid configurations/platform (only missing properties). In this case, let the project system handle it.

Any suggestions?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -161,6 +173,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <OutputPath Condition=" '$(PlatformName)' == 'AnyCPU' ">$(BaseOutputPath)$(Configuration)\</OutputPath>
+    <OutputPath Condition=" '$(PlatformName)' != 'AnyCPU' ">$(BaseOutputPath)$(PlatformName)\$(Configuration)\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup>
+    <OutputPath Condition="!HasTrailingSlash('$(OutputPath)')">$(OutputPath)\</OutputPath>
+  </PropertyGroup>
+
   <PropertyGroup Condition=" '$(OutputPath)' == '' ">
     <!--
         A blank OutputPath at this point means that the user passed in an invalid Configuration/Platform
```
Nirmal4G, line 185: If we're allowing defaults for `OutputPath`, then, how can we check for invalid configuration/platform combinations?

1. We could check for `PlatformTarget` instead of `OutputPath` but that won't be a complete solution.
2. We could send in the available combinations, and check against current value but that'll only work with solution that defines all combinations.
3. There are no invalid configurations/platform (only missing properties). In this case, let the project system handle it.

Any suggestions?

Nirmal4G: If we're allowing defaults for `OutputPath`, then, how can we check for invalid configuration/platform combinations?

1. We could check for `PlatformTarget` instead of `OutputPath` but that won't be a complete solution.
2. We could send in the available combinations, and check against current value but that'll only work with solution that defines all combinations.
3. There are no invalid configurations/platform (only missing properties). In this case, let the project system handle it.

Any suggestions?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -161,6 +172,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <OutputPath Condition=" '$(PlatformName)' == 'AnyCPU' ">$(BaseOutputPath)$(Configuration)\</OutputPath>
+    <OutputPath Condition=" '$(PlatformName)' != 'AnyCPU' ">$(BaseOutputPath)$(PlatformName)\$(Configuration)\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup>
+    <OutputPath>$([MSBuild]::EnsureTrailingSlash('$(OutputPath)'))</OutputPath>
+  </PropertyGroup>
+
   <PropertyGroup Condition=" '$(OutputPath)' == '' ">
     <!--
         A blank OutputPath at this point means that the user passed in an invalid Configuration/Platform
```
Nirmal4G, line 185: If we're allowing defaults for `OutputPath`, then, how can we check for invalid configuration/platform combinations?

1. We could check for `PlatformTarget` instead of `OutputPath` but that won't be a complete solution.
2. We could send in the available combinations, and check against current value but that'll only work with solution that defines all combinations.
3. There are no invalid configurations/platform (only missing properties). In this case, let the project system handle it.

Any suggestions?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -333,6 +333,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <AutoUnifyAssemblyReferences Condition="'$(GenerateBindingRedirectsOutputType)' == 'true' and '$(AutoGenerateBindingRedirects)' != 'true'">false</AutoUnifyAssemblyReferences>
   </PropertyGroup>
   <PropertyGroup>
+    <BaseIntermediateOutputPath Condition=" '$(BaseIntermediateOutputPath)' == '' ">obj\</BaseIntermediateOutputPath>
```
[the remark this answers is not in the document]
Nirmal4G: Our Build env has a customized common props that doesn't have `BaseIntermediateOutputPath` defined. There are cases where we do not use common props at all and no project extensions import (i.e. no NuGet).

In these cases, the build errors out. I can disable the validation via `EnableBaseIntermediateOutputPathMismatchWarning` but the default value is not specified at this point. That's why I've added the missing definition.

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: Similar to `OutDir` and `OutDirWasSpecified`...?

We can do that. I'll try it and see how it goes!

Eventually, We should ignore `OutputPath` as a means of checking invalid Configuration/Platforms.

src/Tasks/Microsoft.Common.CurrentVersion.targets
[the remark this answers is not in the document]
rainersigwald: I still don't understand. At the top of this file, there's

https://github.com/microsoft/msbuild/blob/b449a942dcf3f11a8f245d06a84a41024a9a4afb/src/Tasks/Microsoft.Common.CurrentVersion.targets#L19

Are you saying that you have an environment where that is explicitly disabled?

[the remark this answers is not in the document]
Nirmal4G: Yes

Nirmal4G:
Did a little bit of refactoring to place all the output logic in one place. Other than that, only changes include what you have suggested: Using `*OutputPath` properties to fallback to legacy Configuration/Platfrom check. That's the last commit.

Should I add tests for the `*OutputPath` behaviors? If so, then, Where are the tests for the `OutputPath`?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,24 +127,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
-    <OutputPath Condition=" '$(Platform)'=='' and '$(Configuration)'=='' and '$(OutputPath)'=='' ">bin\Debug\</OutputPath>
-    <_OriginalConfiguration>$(Configuration)</_OriginalConfiguration>
+    <!-- Example, AnyCPU -->
     <_OriginalPlatform>$(Platform)</_OriginalPlatform>
-    <Configuration Condition=" '$(Configuration)'=='' ">Debug</Configuration>
-    <ConfigurationName Condition=" '$(ConfigurationName)' == '' ">$(Configuration)</ConfigurationName>
     <!-- Example, Debug -->
-    <Platform Condition=" '$(Platform)'=='' ">AnyCPU</Platform>
+    <_OriginalConfiguration>$(Configuration)</_OriginalConfiguration>
+    <!-- Check whether OutputPath was specified for valid Configuration/Platform combination -->
```
rainersigwald, line 140: What does "valid" mean here? Is this the same check as before where if `OutputPath` is explicitly set, that means that it's a valid combination?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -161,7 +185,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <ProduceReferenceAssembly Condition="'$(ProduceReferenceAssembly)' == ''">false</ProduceReferenceAssembly>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(OutputPath)' == '' ">
+  <PropertyGroup Condition="'$(BaseOutputPathWasSpecified)' != 'true' and '$(_OutputPathWasMissing)' == 'true'">
     <!--
         A blank OutputPath at this point means that the user passed in an invalid Configuration/Platform
         combination.  Whether this is considered an error or a warning depends on the value of
```
rainersigwald, line 191: Can you please update this comment?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,24 +127,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+  <PropertyGroup>
+    <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
+    <PlatformName Condition="'$(PlatformName)' == ''">$(Platform)</PlatformName>
+    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
+    <ConfigurationName Condition="'$(ConfigurationName)' == ''">$(Configuration)</ConfigurationName>
+
+    <BaseOutputPath>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(BaseOutputPath)', 'bin'))))</BaseOutputPath>
+    <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' == 'AnyCPU'">$(BaseOutputPath)$(Configuration)\</OutputPath>
+    <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' != 'AnyCPU'">$(BaseOutputPath)$(PlatformName)\$(Configuration)\</OutputPath>
```
rainersigwald, line 154: Should this be
```suggestion
    <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' != 'AnyCPU'">$([MSBuild]::NormalizeDirectory('$(BaseOutputPath)', '$(PlatformName)', '$(Configuration)'))</OutputPath>
```

?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -374,12 +389,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <PropertyGroup Condition="'$(_DebugSymbolsProduced)' == 'true' and '$(OutputType)' == 'winmdobj'">
     <WinMDExpOutputPdb Condition="'$(WinMDExpOutputPdb)' == ''">$(IntermediateOutputPath)$(TargetName).pdb</WinMDExpOutputPdb>
-    <_WinMDDebugSymbolsOutputPath>$([System.IO.Path]::Combine('$(OutDir)', $([System.IO.Path]::GetFileName('$(WinMDExpOutputPdb)'))))</_WinMDDebugSymbolsOutputPath>
+    <_WinMDDebugSymbolsOutputPath>$(OutDir)$([System.IO.Path]::GetFileName('$(WinMDExpOutputPdb)'))</_WinMDDebugSymbolsOutputPath>
```
rainersigwald, line 392: I don't think I prefer the concatenation to the `Combine` call, even though it's generally safe. Do you think it's strongly preferable? If so, why?

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: Yes

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: Sure. Just wanted to get a review from you before I finalize with comments and tests

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: It's fine by me.

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: The only thing `Combine` does (_in this context_) is join the parts with a path separator which we can be sure, will be present in `OutDir` property.

So, if we want to be extra safe, sure I can add it. But in this case we shouldn't need to.

Also, Does using unnecessary property function calls impact perf?

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: In general we don't worry about calling property functions if they're a better expression of intent and meaning. If we observe a perf problem later we can always change it (and comment why).

rainersigwald:
I think this looks good to go. Would you mind resolving conflicts?

Nirmal4G:
When I tested the above logic with and without Common targets, there's a noticeable difference of around 700ms-1s delay when using property functions. This is observed with running the test project again and again in Command Prompt and WSL manually. Is there any guide/tool to properly profile MSBuild builds?

I have tested the above logic with

1. `\`, `/`
2. `Path.Join`
3. `Path.Combine`
4. `NormalizeDirectory`

The `NormalizeDirectory`/`NormalizePath` doesn't always generate paths under `MSBuildProjectDirectory`. Is this Intentional?

<details>
<summary><em>Behaviour of </em><code>NormalizeDirectory</code></summary>

```console
D:\Projects\Work\Repros>msbuild Test.proj -p:Configuration=C:Debug

Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Framework

Build started 18-07-2020 08:30:16 PM.
Project "D:\Projects\Work\Repros\Test.proj" on node 1 (default targets).
Test:
  MSBuildToolsVersion: Current
  MSBuildToolsPath:    C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin
  MSBuildProjectDirectory: D:\Projects\Work\Repros
  BaseOutputPath: bin\
  BaseIntermediateOutputPath: obj\
  OutputPath: C:\Program Files (x86)\Microsoft Visual Studio\Installer\Debug\
  IntermediateOutputPath: C:\Program Files (x86)\Microsoft Visual Studio\Installer\Debug\
Done Building Project "D:\Projects\Work\Repros\Test.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.61
```

</details>

<details>
<summary><em>Behaviour of </em><code>Path.Combine</code></summary>

```console
D:\Projects\Work\Repros>msbuild Test.proj -p:Configuration=C:Debug

Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Framework

Build started 18-07-2020 08:33:25 PM.
Project "D:\Projects\Work\Repros\Test.proj" on node 1 (default targets).
Test:
  MSBuildToolsVersion: Current
  MSBuildToolsPath:    C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin
  MSBuildProjectDirectory: D:\Projects\Work\Repros
  BaseOutputPath: bin\
  BaseIntermediateOutputPath: obj\
  OutputPath: C:Debug\
  IntermediateOutputPath: C:Debug\
Done Building Project "D:\Projects\Work\Repros\Test.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.63
```

</details>

<details>
<summary><em>Behaviour of </em><code>Path.Join</code></summary>

```console
D:\Projects\Work\Repros>dotnet msbuild Test.proj -p:Configuration=C:Debug -v:n

Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Core

Build started 18-07-2020 08:36:31 PM.
     1>Project "D:\Projects\Work\Repros\Test.proj" on node 1 (default targets).
     1>Test:
         MSBuildToolsVersion: Current
         MSBuildToolsPath:    C:\Program Files\dotnet\sdk\3.1.302
         MSBuildProjectDirectory: D:\Projects\Work\Repros
         BaseOutputPath: bin\
         BaseIntermediateOutputPath: obj\
         OutputPath: bin\C:Debug\
         IntermediateOutputPath: obj\C:Debug\
     1>Done Building Project "D:\Projects\Work\Repros\Test.proj" (default targets).

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.21
```

</details>

Out of all the 3, which one do we want to have?

So far `Path.Join` is my favourite but it's not available to MSBuild for .NET Framework!

rainersigwald:
> there's a noticeable difference of around 700ms-1s delay when using property functions

That's a shockingly large difference. Can you share exactly how you were testing?

> Is there any guide/tool to properly profile MSBuild builds?

We have a fairly detailed [evaluation-time profiler](https://github.com/microsoft/msbuild/blob/master/documentation/evaluation-profiling.md), and can also use our [ETW events](https://github.com/microsoft/msbuild/blob/master/documentation/specs/event-source.md).

> The `NormalizeDirectory`/`NormalizePath` doesn't always generate paths under `MSBuildProjectDirectory`. Is this Intentional?

The results you shared don't make any sense to me, no. Again, can you describe exactly how you got them?

Nirmal4G:
Here's the test projects with different variations of OutputPath definitions: [MSBuild-5238-Repro.zip](https://github.com/dotnet/msbuild/files/5029301/MSBuild-5238-Repro.zip)

With both Desktop and Core MSBuild run the test projects with `Configuration=C:Debug` or `PlatformName=C:ARM` (i.e. prefix them with Windows Drive notation) and observe the differences.
They are of course invalid `Configuration` and `Platform` but Except `/`,`\` and `Path.Join` all others have different behaviors with relative path inserted into the above properties.
I don't know what sort of problems they will bring down the road but I observed it by accident and so, I post here for clarification.

As you have suggested, we can also profile them with or without Common targets using `ImportCommonTargets` property. I will also profile them (both the test projects and PR patch) and post them here soon.

Nirmal4G:
And sorry for the delayed reply.

Nirmal4G:
@rainersigwald

Nirmal4G:
@rainersigwald @Forgind 🔔

Nirmal4G:
@rainersigwald @Forgind 🔔 Finally, everything green!

src/Shared/FileUtilities.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 #if !CLR2COMPATIBILITY
 using System.Collections.Concurrent;
+#else
+using Microsoft.Build.Shared.Concurrent;
```
rainersigwald, line 8: Can you please revert this unrelated change?

src/Tasks.UnitTests/GenerateBindingRedirects_Tests.cs
```diff
@@ -17,7 +17,7 @@
 using Xunit;
 using Xunit.Abstractions;
 
-namespace Microsoft.Build.Tasks.Unittest
+namespace Microsoft.Build.Tasks.UnitTests
```
rainersigwald, line 20: Likewise please revert.

src/Shared/FileUtilities.cs
Nirmal4G: Moved to a separate PR.

src/Tasks.UnitTests/GenerateBindingRedirects_Tests.cs
Nirmal4G: Moved to a separate PR.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,25 +129,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     <!-- Example, Debug -->
-    <Platform Condition=" '$(Platform)'=='' ">AnyCPU</Platform>
+    <_OriginalConfiguration>$(Configuration)</_OriginalConfiguration>
+    <!-- Check whether OutputPath was specified for valid Configuration/Platform combination -->
+    <_OutputPathWasMissing Condition="'$(_OriginalPlatform)' != '' and '$(_OriginalConfiguration)' != '' and '$(OutputPath)' == ''">true</_OutputPathWasMissing>
+    <!-- Check whether BaseOutputPath was specified -->
+    <BaseOutputPathWasSpecified Condition="'$(BaseOutputPath)' != ''">true</BaseOutputPathWasSpecified>
+  </PropertyGroup>
+
+  <PropertyGroup>
```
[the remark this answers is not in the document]
Nirmal4G: I did inline changes but there was no clear separation of logic. It was ugly and thus the move.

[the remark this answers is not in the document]
Nirmal4G: I did inline changes but there was no clear separation of logic. Hence, a separate property group.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -191,16 +221,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
-    <OutDir Condition="'$(OutDir)' != '' and !HasTrailingSlash('$(OutDir)')">$(OutDir)\</OutDir>
+    <OutDir>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(OutDir)', '$(OutputPath)'))))</OutDir>
     <ProjectName Condition=" '$(ProjectName)' == '' ">$(MSBuildProjectName)</ProjectName>
     <!-- Example, MyProject -->
 
     <!-- For projects that generate app packages or ones that want a per-project output directory, update OutDir to include the project name -->
-    <OutDir Condition="'$(OutDir)' != '' and '$(OutDirWasSpecified)' == 'true' and (('$(WindowsAppContainer)' == 'true' and '$(GenerateProjectSpecificOutputFolder)' != 'false') or '$(GenerateProjectSpecificOutputFolder)' == 'true')">$(OutDir)$(ProjectName)\</OutDir>
+    <OutDir Condition="'$(OutDir)' != '' and '$(OutDirWasSpecified)' == 'true' and (('$(WindowsAppContainer)' == 'true' and '$(GenerateProjectSpecificOutputFolder)' != 'false') or '$(GenerateProjectSpecificOutputFolder)' == 'true')">$([MSBuild]::EnsureTrailingSlash('$(OutDir)$(ProjectName)'))</OutDir>
 
+    <AssemblyName Condition=" '$(AssemblyName)'=='' ">$(MSBuildProjectName)</AssemblyName>
```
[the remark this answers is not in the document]
Nirmal4G: I moved this near to a place where it is actually used. I placed it in a logical group with all the project properties.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -68,6 +68,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Import Project="Microsoft.NET.props" Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' or '$(TargetFrameworkIdentifier)' == '.NETCoreApp'"/>
 
   <PropertyGroup>
+    <!-- Generates full paths for the 'File' property in errors, warnings and messages in many targets -->
+    <GenerateFullPaths Condition="'$(GenerateFullPaths)' == ''">true</GenerateFullPaths>
```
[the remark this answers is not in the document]
Nirmal4G: This property was added after my PR. When resolving conflict, this was the best place, I could add it in.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,25 +129,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     <_OriginalPlatform>$(Platform)</_OriginalPlatform>
-    <Configuration Condition=" '$(Configuration)'=='' ">Debug</Configuration>
-    <ConfigurationName Condition=" '$(ConfigurationName)' == '' ">$(Configuration)</ConfigurationName>
     <!-- Example, Debug -->
-    <Platform Condition=" '$(Platform)'=='' ">AnyCPU</Platform>
+    <_OriginalConfiguration>$(Configuration)</_OriginalConfiguration>
+    <!-- Check whether OutputPath was specified for valid Configuration/Platform combination -->
+    <_OutputPathWasMissing Condition="'$(_OriginalPlatform)' != '' and '$(_OriginalConfiguration)' != '' and '$(OutputPath)' == ''">true</_OutputPathWasMissing>
+    <!-- Check whether BaseOutputPath was specified -->
+    <BaseOutputPathWasSpecified Condition="'$(BaseOutputPath)' != ''">true</BaseOutputPathWasSpecified>
```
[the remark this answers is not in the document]
Nirmal4G: Could be useful to end-user projects. This is similar to `OutDirWasSpecified` property. Since, it was a user-facing property, I have made this one the same.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -119,25 +129,40 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+  <PropertyGroup>
+    <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
+    <PlatformName Condition="'$(PlatformName)' == ''">$(Platform)</PlatformName>
+    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
+    <ConfigurationName Condition="'$(ConfigurationName)' == ''">$(Configuration)</ConfigurationName>
+
+    <BaseOutputPath>$([MSBuild]::EnsureTrailingSlash($([MSBuild]::ValueOrDefault('$(BaseOutputPath)', 'bin'))))</BaseOutputPath>
+    <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' == 'AnyCPU'">$([System.IO.Path]::Combine('$(BaseOutputPath)', '$(Configuration)'))</OutputPath>
+    <OutputPath Condition="'$(OutputPath)' == '' and '$(PlatformName)' != 'AnyCPU'">$([System.IO.Path]::Combine('$(BaseOutputPath)', '$(PlatformName)', '$(Configuration)'))</OutputPath>
```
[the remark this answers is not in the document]
Nirmal4G: Let's keep the `Combine` call for now. I'll create a new issue to investigate it further. On top of my head, I think the issue is with `Path.GetFullPath(...)` and the way MSBuild chooses/gets the active directory in Windows.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -162,12 +187,17 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <ProduceReferenceAssembly Condition="'$(ProduceReferenceAssembly)' == ''">false</ProduceReferenceAssembly>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(OutputPath)' == '' ">
+  <!--
+    For non-Sdk style projects, the validation of an invalid Configuration/Platform is determined by the non-empty
```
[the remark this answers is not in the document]
Nirmal4G: Sure. I recommend that we should not use `.NET Core`, as it is being rebranded into `.NET`. Also, I think, it would be better to say the actual project system used instead of specifically mentioning `.NET SDK`. Since, CPS (_is not tied to `.NET` only, right?!_) might replace old project system altogether by supporting the rest of the project system flavours.

```suggestion
    For projects loaded with MPFProj (Legacy project system), the validation of an invalid Configuration/Platform is determined by the non-empty
```

Nirmal4G:
@rainersigwald Sorry for late replies!!

Nirmal4G:
@rainersigwald I've responded to your comments. Also, Sorry for the late replies!

Nirmal4G:
@Forgind
> Without having looked too carefully, does this also resolve #2308?

No, I don't think so. That requires changes from both C++ and .NET targets.

Nirmal4G:
@rainersigwald

Nirmal4G:
@rainersigwald @Forgind 🔔

Nirmal4G:
@rainersigwald @Forgind 🔔

Nirmal4G:
@rainersigwald @Forgind 🔔

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -109,7 +111,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
 
+    BaseOutputPath:
+    This is the top level folder where all configuration specific output folders will be created.
+    Default value is bin\
+
     OutputPath:
+    This is the full Output Path, and is derived from BaseOutputPath, if none specified
+    (eg. bin\Debug). If this property is overridden, then setting BaseOutputPath has no effect.
+
+    For projects loaded with MPFProj (Legacy project system):
```
rainersigwald, line 122: What is MPFProj?

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -162,12 +188,19 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <ProduceReferenceAssembly Condition="'$(ProduceReferenceAssembly)' == ''">false</ProduceReferenceAssembly>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(OutputPath)' == '' ">
+  <!--
+    For projects loaded with MPFProj (Legacy project system) that defines the properties per Configuration/Platform
+    combination by the IDE itself, the validation of an invalid Configuration/Platform is determined by the non-empty
+    value of the 'OutputPath' property specified under the IDE generated Configuration/Platform 'PropertyGroup' section.
```
rainersigwald, line 194: I don't understand this comment. What does the project system have to do with the validity computation--isn't that entirely in MSBuild logic?

rainersigwald (commented):
I need to look at the tests but other than the comments about "MPFProj" that I need clarification on this is looking good. Thanks for your patience!

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: Isn't that a common name for legacy project system for `.NET` (`csproj.dll` and `msvbprj.dll`)!?

If not, then, what's the correct term?

src/Tasks/Microsoft.Common.CurrentVersion.targets
Nirmal4G: Nothing. It's just that the project uses this logic typically load in legacy project system and not on CPS.

Also see: https://github.com/dotnet/msbuild/pull/5238#discussion_r503230153

As I said in the above comment I'm open to a better alternative term.

Nirmal4G:
@rainersigwald 🔔

Nirmal4G:
@rainersigwald 🔔

src/Tasks/Microsoft.Common.CurrentVersion.targets
rainersigwald: How do you feel about the wording in c1a2ff8c68?

Nirmal4G: Not bad but I do prefer the previous but with changes like...

`For projects loaded by Legacy project system with only common targets...`

Either we could use this sentence directly or we could get a name for it. I prefer the former as it is on point!

Nirmal4G: I have mixed the wording from both suggestions (check latest commit). Is that okay with you?

rainersigwald (approved):
Awesome, thanks!

Nirmal4G:
@rainersigwald Sorry for the late reply!

> Went to celebrate Diwali and came back with a bad case of cold.

Forgind (approved):
Looks good to me! Thanks for this!

> Went to celebrate Diwali and came back with a bad case of cold.

I'm sorry to hear that. Glad you're feeling better!

Nirmal4G:
> I'm sorry to hear that. Glad you're feeling better!

I'm fine now. Thanks!

Nirmal4G:
Finally! 🥳

PR #5242, opened 2020-04-07
Remove unnecessary NuGet package references.

teo-tsirpanis:
Some NuGet packages (such as `System.Net.Http`) do not target .NET Standard 2.0 and reference tons of packages from older .NET Standard versions. The types they contain are already included in .NET Standard 2.0, so they were removed to unburden the package reference graph.

### DIFF
--- a/src/Build/Microsoft.Build.csproj
+++ b/src/Build/Microsoft.Build.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="..\Shared\FileSystemSources.proj" />
   <Import Project="..\Shared\DebuggingSources.proj" />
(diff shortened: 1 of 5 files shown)


src/Build/Microsoft.Build.csproj
```diff
@@ -45,10 +45,8 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
-    <PackageReference Include="System.Diagnostics.TraceSource" />
     <PackageReference Include="System.Reflection.Metadata" />
-    <PackageReference Include="System.Reflection.TypeExtensions" />
-    <PackageReference Include="System.Runtime.Loader" />
+    <PackageReference Include="System.Runtime.Loader" Condition="'$(TargetFramework)' != 'netcoreapp2.1'" />
```
rainersigwald, line 49: Can you remove `System.Runtime.Loader` entirely? I don't think this condition will ever fire because this project is built only for `netcoreapp2.1` and `net472`.

teo-tsirpanis: Done. With the same reasoning, I removed the condition on `System.Security.Principal.Windows`, as it would always fire.

rainersigwald:
Thanks, @teo-tsirpanis!

PR #5257, opened 2020-04-08
Replace Framework references

Forgind:


rrelyea:
should you just make local constants (in this assembly?)

Forgind:
I wanted a working version first. Then I can minimize the scale of the change and clean it up.

rrelyea:
We should ask Phil Hodgson on the internal thread on how to best validate

PR #5260, opened 2020-04-08
Copy satellites in deployment script

cdmihai:
Extracted this from #5222 

When I tried the deployment script msbuild failed not finding the satellite assemblies. Apparently this happens when you're testing msbuild bits with new resources.

### DIFF
--- a/scripts/Deploy-MSBuild.ps1
+++ b/scripts/Deploy-MSBuild.ps1
@@ -8,15 +8,30 @@ Param(
   [string] $runtime = "Desktop"
 )
 
+Set-StrictMode -Version "Latest"
+$ErrorActionPreference = "Stop"
+
 function Copy-WithBackup ($origin) {
-    $destinationPath = Join-Path -Path $destination -ChildPath (Split-Path $origin -leaf)
+    $directoryPart = Join-Path -Path $destination $origin.IntermediaryDirectories 
+    $destinationPath = Join-Path -Path $directoryPart (Split-Path $origin.SourceFile -leaf)
 
     if (Test-Path $destinationPath -PathType Leaf) {
         # Back up previous copy of the file
         Copy-Item $destinationPath $BackupFolder -ErrorAction Stop
     }
 
-    Copy-Item $origin $destinationPath -ErrorAction Stop
+    if (!(Test-Path $directoryPart)) {
+        [system.io.directory]::CreateDirectory($directoryPart)
+    }
+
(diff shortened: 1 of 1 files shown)


scripts/Deploy-MSBuild.ps1
```diff
@@ -36,58 +51,65 @@ if ($runtime -eq "Desktop") {
...
-    "artifacts\bin\MSBuild.Bootstrap\$configuration\$targetFramework\Microsoft.VisualBasic.targets"
+    FileToCopy "$bootstrapBinDirectory\Microsoft.Build.dll"
+    FileToCopy "$bootstrapBinDirectory\Microsoft.Build.Framework.dll"
+    FileToCopy "$bootstrapBinDirectory\Microsoft.Build.Tasks.Core.dll"
+    FileToCopy "$bootstrapBinDirectory\Microsoft.Build.Utilities.Core.dll"
+
+    FileToCopy "$bootstrapBinDirectory\en\Microsoft.Build.resources.dll" "en"
+    FileToCopy "$bootstrapBinDirectory\en\Microsoft.Build.Tasks.Core.resources.dll" "en"
+    FileToCopy "$bootstrapBinDirectory\en\Microsoft.Build.Utilities.Core.resources.dll" "en"
+    FileToCopy "$bootstrapBinDirectory\en\MSBuild.resources.dll" "en"
```
rainersigwald, line 65: Do we need en resources? I thought they were always embedded as default?

cdmihai: That was my understanding as well until I got missing resource failures. I also deployed release bits, maybe those behave differently. I vaguely remember at some point, some email thread, where somebody suggested we don't treat en resources any differently to capture more possible bugs, so maybe I actually implemented that but forgot :)

scripts/Deploy-MSBuild.ps1
```diff
@@ -8,15 +8,30 @@ Param(
   [string] $runtime = "Desktop"
 )
 
+Set-StrictMode -Version "Latest"
+$ErrorActionPreference = "Stop"
```
BenVillalobos, line 12: Wishlist item that may be too hard to implement: Building out the error functionality to delete the folder that was generated for this particular build, or essentially "reset" to the original binaries.

cdmihai: I restore original binaries by adding the VS msbuild bin dir under git. If I were to implement restoration of original binaries, that's how I would do it. But I'd rather not do it in this PR :)

PR #5262, opened 2020-04-09
Adding Shutdown Message to Worker Nodes

BenVillalobos:
Making debugging easier, one step at a time

### DIFF
--- a/src/Build/BackEnd/Node/OutOfProcNode.cs
+++ b/src/Build/BackEnd/Node/OutOfProcNode.cs
@@ -265,7 +265,6 @@ namespace Microsoft.Build.Execution
                 {
                     case 0:
                         NodeEngineShutdownReason shutdownReason = HandleShutdown(out shutdownException);
-                        CommunicationsUtilities.Trace("Shutting down node MSBuild" + currentProcID);
                         return shutdownReason;
 
                     case 1:
@@ -429,10 +428,7 @@ namespace Microsoft.Build.Execution
         {
             if (_debugCommunications)
             {
-                using (StreamWriter writer = File.CreateText(String.Format(CultureInfo.CurrentCulture, Path.Combine(Path.GetTempPath(), @"MSBuild_NodeShutdown_{0}.txt"), Process.GetCurrentProcess().Id)))
-                {
-                    writer.WriteLine("Node shutting down with reason {0} and exception: {1}", _shutdownReason, _shutdownException);
-                }
+                CommunicationsUtilities.Trace("Shutting down with reason: {1},  and exception: {2}.", Process.GetCurrentProcess().Id, _shutdownReason, _shutdownException);
             }
 
             // Clean up the engine
@@ -515,6 +511,11 @@ namespace Microsoft.Build.Execution
 
                 _nodeEndpoint.Disconnect();
                 CleanupCaches();
+            }
+
+            if (_debugCommunications)
+            {
+                CommunicationsUtilities.Trace("Shut down complete.", Process.GetCurrentProcess().Id);
             }
 
             return _shutdownReason;
--- a/src/Build/BackEnd/Node/OutOfProcNode.cs
+++ b/src/Build/BackEnd/Node/OutOfProcNode.cs
@@ -248,6 +248,7 @@ namespace Microsoft.Build.Execution
         public NodeEngineShutdownReason Run(bool enableReuse, bool lowPriority, out Exception shutdownException)
         {
             // Console.WriteLine("Run called at {0}", DateTime.Now);
+            int currentProcID = Process.GetCurrentProcess().Id;
             string pipeName = NamedPipeUtil.GetPipeNameOrPath("MSBuild" + Process.GetCurrentProcess().Id);
 
             _nodeEndpoint = new NodeEndpointOutOfProc(pipeName, this, enableReuse, lowPriority);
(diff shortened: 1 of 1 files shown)


Forgind:
How is this an improvement on [the line immediately before](https://github.com/microsoft/msbuild/blob/master/src/Build/BackEnd/Node/OutOfProcNode.cs#L432)?

BenVillalobos:
The short answer is I didn't notice it, mainly because I never saw that log message or the files it's generating. Interesting that it creates a fresh file only to write that single line.

Looking at this now I think the logic should actually replace the line you highlighted, so instead of creating and writing to a file it just does a `CommunicationsUtilities.Trace` call.

Forgind:
I'll admit that's a little weird. Maybe replace their file write with CommunicationsUtilities.Trace? I'd like to keep the `if (_debugCommunications)` part.

BenVillalobos:
Agreed, the if statement needs to stay and I'll replace the streamwriter logic with the `Trace` call.

src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -248,7 +248,8 @@ public NodeEngineShutdownReason Run(bool enableReuse, out Exception shutdownExce
         public NodeEngineShutdownReason Run(bool enableReuse, bool lowPriority, out Exception shutdownException)
         {
             // Console.WriteLine("Run called at {0}", DateTime.Now);
-            string pipeName = NamedPipeUtil.GetPipeNameOrPath("MSBuild" + Process.GetCurrentProcess().Id);
+            int currentProcID = Process.GetCurrentProcess().Id;
```
Forgind, line 251: May as well revert this part.

src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -427,10 +428,7 @@ private NodeEngineShutdownReason HandleShutdown(out Exception exception)
         {
             if (_debugCommunications)
             {
-                using (StreamWriter writer = File.CreateText(String.Format(CultureInfo.CurrentCulture, Path.Combine(Path.GetTempPath(), @"MSBuild_NodeShutdown_{0}.txt"), Process.GetCurrentProcess().Id)))
-                {
-                    writer.WriteLine("Node shutting down with reason {0} and exception: {1}", _shutdownReason, _shutdownException);
-                }
+                CommunicationsUtilities.Trace("Shutting down with reason: {1},  and exception: {2}.", Process.GetCurrentProcess().Id, _shutdownReason, _shutdownException);
```
Forgind, line 431: Actually, have you looked at `CommunicationsUtilities.Trace`? It does the same thing, just writing to a different spot:
https://github.com/microsoft/msbuild/blob/master/src/Shared/CommunicationsUtilities.cs#L632-L639

src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -515,6 +513,11 @@ private NodeEngineShutdownReason HandleShutdown(out Exception exception)
                 CleanupCaches();
             }
 
+            if (_debugCommunications)
+            {
+                CommunicationsUtilities.Trace("Shut down complete.", Process.GetCurrentProcess().Id);
```
Forgind, line 518: For both of these, I think the order of your arguments is reversed. See [here](https://github.com/microsoft/msbuild/blob/master/src/Shared/CommunicationsUtilities.cs#L604)

src/Build/BackEnd/Node/OutOfProcNode.cs
BenVillalobos: I figured as much. At least with `Trace` it'll be consistent with how we expect to debug node communication.

src/Build/BackEnd/Node/OutOfProcNode.cs
BenVillalobos: I'm using the overload [above that one](https://github.com/microsoft/msbuild/blob/master/src/Shared/CommunicationsUtilities.cs#L596). I actually mistakenly left the second argument in there. We don't need to pass the current process ID so I'll get rid of that.

src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -427,10 +427,7 @@ private NodeEngineShutdownReason HandleShutdown(out Exception exception)
         {
             if (_debugCommunications)
```
rainersigwald, line 428: Should we preserve this check? Trace does it inside.

src/Build/BackEnd/Node/OutOfProcNode.cs
```diff
@@ -427,10 +427,7 @@ private NodeEngineShutdownReason HandleShutdown(out Exception exception)
...
             {
-                using (StreamWriter writer = File.CreateText(String.Format(CultureInfo.CurrentCulture, Path.Combine(Path.GetTempPath(), @"MSBuild_NodeShutdown_{0}.txt"), Process.GetCurrentProcess().Id)))
-                {
-                    writer.WriteLine("Node shutting down with reason {0} and exception: {1}", _shutdownReason, _shutdownException);
-                }
+                CommunicationsUtilities.Trace("Shutting down with reason: {1},  and exception: {2}.", Process.GetCurrentProcess().Id, _shutdownReason, _shutdownException);
```
rainersigwald, line 430:
```suggestion
                CommunicationsUtilities.Trace("Shutting down with reason: {0},  and exception: {1}.", _shutdownReason, _shutdownException);
```

PR #5265, opened 2020-04-10
Updated branding for 16.7

Forgind:


README.md
```diff
@@ -6,11 +6,11 @@ For more information on MSBuild, see the [MSBuild documentation](https://docs.mi
...
 
-The current development branch is `master`. Changes in `master` will go into a future update of MSBuild, which will release with Visual Studio 16.6 and a corresponding version of the .NET Core SDK.
+The current development branch is `master`. Changes in `master` will go into a future update of MSBuild, which will release with Visual Studio 16.7 and a corresponding version of the .NET Core SDK.
 
 [![Build Status](https://dev.azure.com/dnceng/public/_apis/build/status/Microsoft/msbuild/msbuild-pr?branchName=master)](https://dev.azure.com/dnceng/public/_build/latest?definitionId=86&branchName=master)
 
-We have forked for MSBuild 16.5 in the branch [`vs16.5`](https://github.com/Microsoft/msbuild/tree/vs16.5). Changes to that branch need special approval.
+We have forked for MSBuild 16.6 in the branch [`vs16.6`](https://github.com/Microsoft/msbuild/tree/vs16.6). Changes to that branch need special approval.
 
 [![Build Status](https://dev.azure.com/dnceng/public/_apis/build/status/Microsoft/msbuild/msbuild-pr?branchName=vs16.5)](https://dev.azure.com/dnceng/public/_build/latest?definitionId=86&branchName=vs16.5)
```
rainersigwald, line 15:
```suggestion
[![Build Status](https://dev.azure.com/dnceng/public/_apis/build/status/Microsoft/msbuild/msbuild-pr?branchName=vs16.6)](https://dev.azure.com/dnceng/public/_build/latest?definitionId=86&branchName=vs16.6)
```

rainersigwald:
@Forgind, the last line in the test log is

```
Executing [D:\a\1\s\artifacts\bin\Microsoft.Build.CommandLine.UnitTests\Release\net472\MSBuild.exe "C:\Users\VssAdministrator\AppData\Local\Temp\vp2wz2rg.0p4\Temporarya13948ca67ed49e58bbda8aa937b12a9\build.proj" /low  /nr:false /v:diag ]
```

Which has `/low` which is suspicious. Can you try to chase that down? Feel free to rerun for this PR after getting all the logs you want.

Forgind:
@rainersigwald, I ran it twice this morning (once with `build.cmd -test` and once with CIBuild.cmd), and it passed both times. I also note that it passed before you updated the readme and failed afterwards, so I suspect it was just noise that happened to point to a change I made recently. I restarted it.

rainersigwald:
> I suspect it was just noise that happened to point to a change I made recently.

I agree. The noise is what I would like you to chase down.

Forgind:
The point of a low-priority build is that if there's something else that's higher priority for the machine to do, it will do that instead of this, so that it happened to point to low-priority makes sense to me. The machine was busy --> it saw something low-priority, and it didn't bubble to the top of the priority queue within an hour. In other words, this was built into the test. The only way to get rid of it is to get rid of the test.

rainersigwald:
> The machine was busy --> it saw something low-priority --> it ignored it --> it didn't bubble to the top of the priority queue within an hour --> the build failed.

What evidence do you have that this is what happened? Hosted build machines should be idle except for our workload, which should have much less than 60 minutes of higher-pri tasks, especially given Windows [priority inversion avoidance behavior](https://docs.microsoft.com/en-us/windows/win32/procthread/priority-inversion).

> In other words, this was built into the test. The only way to get rid of it is to get rid of the test.

Tests **must not hang**. It wastes machine resources and developer time when we have to rerun to get around hangs.

Forgind:
Why should build machines be idle? I would've imagined they're shared with other groups.

I don't have bulletproof evidence. What I have is that it sounds reasonable, and it passed repeatedly when there was no competition (locally x2 and this morning) but failed once when the PR was sent at a peak time.

Also, how does the priority inversion avoidance behavior work if there's a normal priority thread blocked by a low priority thread with other unblocked normal priority threads? I imaging the blocked thread's priority matters, in which case that might mean this is blocked until the machine is idle.

Forgind:
Also, I'm thinking through in what cases something could hang. It could deadlock/livelock if the processes can't figure out who should be making progress, there could be an infinite loop/recursion, or this thread could just be starved for over an hour. It can only get into this case under certain execution orders or circumstances, so given how simple the test is, I'd eliminate infinite loop/recursion. It doesn't rely on anything external to run, so any deadlock/livelock would have to be internal, and again because of how simple the test is, I'd eliminate those, too. Are there cases other than starvation that you can think of as to why it might hang?

PR #5269, opened 2020-04-11
Sdk resolver updates

### LINKED ISSUE #5239: Add features to MSBuild SDK Resolver interface to support optional SDK workloads
[Workloads overview spec](https://github.com/dotnet/designs/pull/100)

Implementation details: [Workload resolution in MSBuild](https://github.com/dotnet/designs/pull/104)

We will update the SDK Resolver interface so that the result can:

- Return any number of SDK paths (zero, one, or many).  It will be possible to return a successful result that returns no SDK paths.
- Return MSBuild items and properties to add to the evaluation result

dsplaisted:
Add more abilities to SDK resolvers

- Return any number of paths (zero, one, or many) in a successful result
- Return MSBuild items and properties to add to the evaluation result

Resolves #5239


wli3:
test?

dsplaisted:
> test?

It's still a draft :-)

dsplaisted:
@rainersigwald @Forgind @BenVillalobos I think this is basically ready for review.  I think I still need to update the XML documentation for the new and refactored APIs, but otherwise this should be ready.

src/Framework/Sdk/SdkResultPathAndVersion.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            return obj is SdkResultPathAndVersion version &&
+                   Path == version.Path &&
+                   Version == version.Version;
+        }
+
+        public override int GetHashCode()
+        {
+            int hashCode = 2040984829;
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Path);
```
rainersigwald, line 40: Is a mismatch on case on non-case-sensitive filesystems going to be a problem here?

src/Framework/Sdk/SdkResultPathAndVersion.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.Build.Framework
+{
+    /// <summary>
+    /// Encapsulates both the path to an SDK returned by an SDK resolver, and its version
+    /// </summary>
+    public sealed class SdkResultPathAndVersion
+    {
+        public string Path { get; }
+
+        public string Version { get; }
```
rainersigwald, line 16: I'm not totally clear on why version is interesting here. Can you elaborate in a doc comment?

src/Shared/ITranslator.cs
```diff
@@ -288,40 +284,40 @@ void TranslateArray<T>(ref T[] array, NodePacketValueFactory<T> factory)
...
 
         /// <summary>
         /// Translates a dictionary of { string, T } for dictionaries with public parameterless constructors.
         /// </summary>
         /// <typeparam name="D">The reference type for the dictionary.</typeparam>
         /// <typeparam name="T">The reference type for values in the dictionary.</typeparam>
         /// <param name="dictionary">The dictionary to be translated.</param>
-        /// <param name="valueFactory">The factory used to instantiate values in the dictionary.</param>
-        void TranslateDictionary<D, T>(ref D dictionary, NodePacketValueFactory<T> valueFactory)
+        /// <param name="objectTranslator">The translator to use for the values in the dictionary</param>
```
rainersigwald, line 305:
```suggestion
        /// <param name="objectTranslator">The translator to use for the values in the dictionary.</param>
```

src/Framework/Sdk/SdkResultItem.cs
```diff
@@ -0,0 +1,72 @@
...
+        public override int GetHashCode()
+        {
+            int hashCode = -849885975;
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(ItemSpec);
+
+            if (Metadata != null)
+            {
+                foreach (var kvp in Metadata)
+                {
+                    hashCode = hashCode * 1521134295 + kvp.Key.GetHashCode();
```
rainersigwald, line 64: This is negative everywhere else. Is the positive here intentional?

Also, this copies elsewhere but I'm just noticing it now: should this be an explicitly unchecked operation? Couldn't this cause overflow?

src/Framework/Sdk/SdkResultItem.cs
```diff
@@ -0,0 +1,72 @@
...
+        public SdkResultItem()
+        {
+            Metadata = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// Creates an <see cref="SdkResultItem"/>
+        /// </summary>
+        /// <param name="itemSpec">The value (itemspec) for the item</param>
+        /// <param name="metadata">A dictionary of item metadata.  This should be created with <see cref="StringComparer.OrdinalIgnoreCase"/> for the comparer.</param>
```
rainersigwald, line 26: Not `MSBuildNameIgnoreCaseComparer`?

rainersigwald: Debug-only assert for this invariant?

src/Build/BackEnd/Components/SdkResolution/SdkResult.cs
```diff
@@ -33,30 +33,61 @@ public SdkResult(SdkReference sdkReference, string path, string version, IEnumer
...
+            Success = true;
+            SdkReference = sdkReference;
+            if (paths != null)
+            {
+                var firstPathAndVersion = paths.FirstOrDefault();
+                if (firstPathAndVersion != null)
+                {
+                    Path = firstPathAndVersion.Path;
+                    Version = firstPathAndVersion.Version;
+                }
```
rainersigwald, line 58: Can you explain why this pattern rather than allowing Path to be null if multiple-paths are defined?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1817,11 +1817,122 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+            }
+        }
+
+        //  Creates a project to set the properties and include the items from an SdkResult
+        private ProjectRootElement CreateProjectForSdkResult(SdkResult sdkResult)
+        {
+            int propertiesAndItemsHash;
+
+#if NETCOREAPP
+            HashCode hash = new HashCode();
```
rainersigwald, line 1862: I want this everywhere, but why are you using it only here? Is this the only one that doesn't need to cross processes?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1817,11 +1817,122 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+#endif
+
+                }
+            }
+
+#if NETCOREAPP
+            propertiesAndItemsHash = hash.ToHashCode();
+#endif
+
+            string projectPath = _projectRootElement.FullPath + ".SdkResolver." + (propertiesAndItemsHash++) + ".proj";
```
rainersigwald, line 1899: Why the increment?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1817,11 +1817,122 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                        propertyGroup.AddProperty(propertyNameAndValue.Key, EscapingUtilities.Escape(propertyNameAndValue.Value));
+                    }
+                }
+
+                if (sdkResult.ItemsToAdd != null && sdkResult.ItemsToAdd.Any())
+                {
+                    var itemGroup = project.AddItemGroup();
+                    foreach (var item in sdkResult.ItemsToAdd)
+                    {
+                        //  TODO: Does the itemspec or metadata need to be escaped?
```
rainersigwald, line 1920: Can you add a test for this where the resolver produces some horrible special-character nonsense?

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1817,11 +1817,122 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+
+                _projectRootElementCache.AddEntry(project);
+
+                return project;
             }
 
-            ExpandAndLoadImportsFromUnescapedImportExpression(directoryOfImportingFile, importElement, project,
-                throwOnFileNotExistsError, out projects);
+            //  TODO: Is it worth using the cache for these?
+            return _projectRootElementCache.Get(
```
rainersigwald, line 1931: Probably? It should reduce memory pressure a tiny bit for a solution with multiple projects with the same SDK, right?

src/Framework/Sdk/SdkResultPathAndVersion.cs
dsplaisted: Before this change, the `SdkResult` had `Path` and `Version` properties, which were the path and optionally a version for the resolved Sdk.

Now that it can include multiple paths, I thought it made sense to support specifying a version for each path.  It would be a lot simpler if we didn't have this though, we could get rid of the whole `SdkResultPathAndVersion` type.

It looks like the versions returned in the SdkResult are only used for generating warnings such as:

- MSB4241: The SDK reference "{0}" version "{1}" was resolved to version "{2}" instead.  You could be using a different version than expected if you do not update the referenced version to match.
- MSB4240: Multiple versions of the same SDK "{0}" cannot be specified. The SDK version "{1}" already specified by "{2}" will be used and the version "{3}" will be ignored.

I would be in favor of saying that if you are returning multiple paths from an SDK resolver, you can't give a different version to each of them.  We only have one scenario for returning multiple paths and in that scenario we don't need to specify what version each path has.

src/Build/Evaluation/Evaluator.cs
dsplaisted: Are there any existing tests for horrible special characters that I can get the nonsense from? :-)

src/Build/BackEnd/Components/SdkResolution/SdkResult.cs
dsplaisted: Are you asking why if there are multiple paths, it doesn't just use a List to hold them and leave singular `Path` property set to null?

It's fairly arbitrary.  The SdkResult should always be created with the Factory method that takes a list of paths, so the way the multiple paths are stored isn't really exposed outside of MSBuild.

src/Framework/Sdk/SdkResultItem.cs
dsplaisted: I wasn't aware of that comparer, but it's internal and not in the Microsoft.Build.Framework assembly, so we can't tell people to use it.

src/Framework/Sdk/SdkResultItem.cs
dsplaisted: I used the VS code helper to generate these methods, and then modified as appropriate for loops, etc.  So I think the missing negative was a copy/paste error.  I don't know whether they should be explicitly unchecked, but if they should be, I would expect the VS code helper to include that.

src/Build/Evaluation/Evaluator.cs
dsplaisted: I used it here simply because I was writing this code more manually as opposed to letting the code helper write a GetHashCode() method.  I wasn't thinking about cross process issues, but I think that we don't send hash codes across processes, so I think in general it would be OK to use the HashCode type.  The problem is that it's only on .NET Core (or would require adding a dependency in .NET Framework, which can be risky), so it's probably not worth #if'ing most of the GetHashCode() implementations.

src/Build/Evaluation/Evaluator.cs
dsplaisted: I didn't think it was a good idea to have multiple virtual files with the same path but different contents.  It might cache the contents somewhere, or make it harder to debug.

src/Framework/Sdk/SdkResultPathAndVersion.cs
dsplaisted: I've made a change that hopefully will make these comparisons consistently case insensitive.

src/Build.UnitTests/Collections/OMcollections_tests.cs
```diff
@@ -16,6 +16,7 @@
 using Shouldly;
 using ObjectModel = System.Collections.ObjectModel;
 using Xunit;
+using Microsoft.Build.BackEnd;
```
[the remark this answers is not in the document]
dsplaisted: They bring in the extension methods that allow existing code to work with the refactored translation APIs.

src/Framework/Sdk/SdkResultFactory.cs
```diff
@@ -20,6 +21,26 @@ public abstract class SdkResultFactory
         /// <returns></returns>
         public abstract SdkResult IndicateSuccess(string path, string version, IEnumerable<string> warnings = null);
 
+        /// <summary>
+        ///     Create an <see cref="SdkResolver" /> object indicating success.
+        /// </summary>
+        /// <remarks>
+        /// This overload allows any number (zero, one, or many) of SDK paths to be returned.  This means allows a "successful" result
+        /// to not resolve to any SDKs.  The resolver can also supply properties or items to communicate information to the build.  This
```
rainersigwald, line 29:
```suggestion
        /// This overload allows any number (zero, one, or many) of SDK paths to be returned.  This means a "successful" result
        /// may not resolve to any SDKs.  The resolver can also supply properties or items to communicate information to the build.  This
```

src/Framework/Sdk/SdkReference.cs
```diff
@@ -102,8 +103,8 @@ public override int GetHashCode()
             unchecked
             {
                 var hashCode = (Name != null ? Name.GetHashCode() : 0);
```
rainersigwald, line 105: Should this not be lowercased too? You use a case-insensitive comparison for equality above.

src/Build/Evaluation/Evaluator.cs
rainersigwald: Can you comment that?

src/Build/Evaluation/Evaluator.cs
rainersigwald: `src\Build.UnitTests\EscapingInProjects_Tests.cs` has a bunch related to tasks.

Returning properties/items/metadata with `%`-followed-by-two-digits and validating it doesn't get unescaped is probably sufficient.

Directory.Build.targets
```diff
@@ -15,4 +15,16 @@
   </PropertyGroup>
   <Import Project="Sdk.targets" Sdk="Microsoft.Build.CentralPackageVersions" />
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+
+  <Target Name="DeleteDevPackage" AfterTargets="GenerateNuspec">
+    <!-- If package just built was already in global packages folder, delete it.  This helps support a local dev cycle where you are consuming
+         a package from another repo without having to update the package version each time. -->
+    
+    <PropertyGroup>
+      <_PackageFolderInGlobalPackages>$(NuGetPackageRoot)$(PackageId.ToLowerInvariant())\$(PackageVersion)</_PackageFolderInGlobalPackages>
```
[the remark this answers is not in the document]
dsplaisted: This is for deleting the cache of the files when you aren't updating the version between local builds.

src/Build.UnitTests/BackEnd/SdkResolverService_Tests.cs
```diff
@@ -217,6 +217,230 @@ public void CachingWrapperShouldWarnWhenMultipleVersionsAreReferenced()
             resolver.ResolvedCalls.Count.ShouldBe(1);
         }
 
+        private void CreateMockSdkResultPropertiesAndItems(out Dictionary<string, string> propertiesToAdd, out Dictionary<string, SdkResultItem> itemsToAdd)
+        {
+            propertiesToAdd = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```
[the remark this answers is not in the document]
dsplaisted: MSBuild does case insensitive comparison in lots of cases even if the underlying file system might be case sensitive.  If you have files that differ only by case in your project, MSBuild probably isn't going to support that well.

src/Build.UnitTests/BackEnd/SdkResolverService_Tests.cs
```diff
@@ -217,6 +217,230 @@ public void CachingWrapperShouldWarnWhenMultipleVersionsAreReferenced()
...
+            ValidateExpectedPropertiesAndItems(includePropertiesAndItems, result);
+
+            _logger.WarningCount.ShouldBe(0);
+        }
+
+        [Theory]
+        [InlineData(true, false)]
+        [InlineData(false, true)]
+        [InlineData(true, true)]
+        public void AssertResolutionWarnsIfResolvedVersionIsDifferentFromReferencedVersionWithMultipleReturnPaths(bool mismatch1, bool mismatch2)
```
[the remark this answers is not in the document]
dsplaisted: This tests that there's a warning generated when there's a mismatch between the requested SDK version and the one that's returned.  Now that each returned path doesn't have a separate version, the test is simplified - it's just a single `Fact` instead of a `Theory`.

src/Build/Evaluation/Evaluator.cs
dsplaisted: Never mind, the increment was a holdover from a previous version of the code where I wasn't using the hash, just incrementing a number each time it needed to generate a new file.  The increment doesn't make any sense here at all.

dsplaisted:
@rainersigwald @Forgind I think I've addressed all the feedback and this is ready to merge.  I'd appreciate a signoff.  Thanks!

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1917,8 +1917,18 @@ ProjectRootElement InnerCreate(string _, ProjectRootElementCacheBase __)
                     var itemGroup = project.AddItemGroup();
                     foreach (var item in sdkResult.ItemsToAdd)
                     {
-                        //  TODO: Does the itemspec or metadata need to be escaped?
-                        itemGroup.AddItem(item.Key, item.Value.ItemSpec, item.Value.Metadata);
+                        Dictionary<string, string> escapedMetadata = null;
+
+                        if (item.Value.Metadata != null)
+                        {
+                            escapedMetadata = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
rainersigwald, line 1925: Use the overload that does a capacity reservation?

src/Build.UnitTests/Evaluation/SdkResultEvaluation_Tests.cs
```diff
@@ -402,6 +402,64 @@ public void SdkResolverCanReturnTheSamePropertiesAndItemsMultipleTimes()
...
+            File.WriteAllText(projectPath, projectContent);
+
+            var project = CreateProject(projectPath, projectOptions);
+
+            project.GetPropertyValue("PropertyName").ShouldBe("PropertyValue" + specialString);
+
+            var itemsFromResolver = project.GetItems("ItemName");
+            var item = itemsFromResolver.ShouldHaveSingleItem();
+            item.EvaluatedInclude.ShouldBe("ItemValue" + specialString);
+            //item.UnevaluatedInclude.ShouldBe("ItemValue" + specialString);
```
rainersigwald, line 457: Delete?

src/Framework/Sdk/SdkResultFactory.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+
```
rainersigwald, line 4: super nit: delete blank line

PR #5270, opened 2020-04-11
Remove usages of Tuples

KirillOsenkov:
We allocate a lot of Tuples.

Hopefully switching to ValueTuples will simplify the code and reduce allocations.

One concern is that CopyOnWriteDictionary had the generic constraint on V to be a reference type. I don't know why that is, if there's an underlying reason then we need to think what to do there.

Also it feels like (string, ElementLocation) deserves to be its own named type. But we can leave that for later.

### DIFF
--- a/src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
+++ b/src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
@@ -105,14 +105,14 @@ namespace Microsoft.Build.Engine.UnitTests.TestData
 
         public static ProjectTaskInstance CreateTargetTask(
             int? counter = null,
-            IDictionary<string, Tuple<string, ElementLocation>> parameters = null,
+            IDictionary<string, (string, ElementLocation)> parameters = null,
             List<ProjectTaskInstanceChild> outputs = null)
         {
             var stringCounter = CounterToString(counter);
 
             var readonlyParameters = parameters != null
-                ? new CopyOnWriteDictionary<string, Tuple<string, ElementLocation>>(parameters)
-                : new CopyOnWriteDictionary<string, Tuple<string, ElementLocation>>();
+                ? new CopyOnWriteDictionary<string, (string, ElementLocation)>(parameters)
+                : new CopyOnWriteDictionary<string, (string, ElementLocation)>();
 
             outputs = outputs ?? new List<ProjectTaskInstanceChild>();
 
--- a/src/Build/Definition/Project.cs
+++ b/src/Build/Definition/Project.cs
@@ -2686,37 +2686,37 @@ namespace Microsoft.Build.Evaluation
                     }
 
                     var matchOccurrences = ItemMatchesInItemSpecString(itemToMatch, itemSpec, elementLocation, itemElement.ContainingProject.DirectoryPath, _data.Expander, out Provenance provenance);
-                    Tuple<Provenance, int> result = matchOccurrences > 0 ? Tuple.Create(provenance, matchOccurrences) : null;
-
-                    return result?.Item2 > 0
-                        ? new ProvenanceResult(itemElement, operation, result.Item1, result.Item2)
-                        : null;
-                }
-
-                Func<ProvenanceResult>[] provenanceProviders =
-                {
-                // provenance provider for include item elements
-                () =>
-                {
-                    var includeResult = SingleItemSpecProvenance(itemElement.Include, itemElement.IncludeLocation, Operation.Include);
-                    if (includeResult == null)
+                    if (matchOccurrences > 0)
                     {
-                        return null;
+                        return new ProvenanceResult(itemElement, operation, provenance, matchOccurrences);
                     }
 
+                    return null;
+                }
+
+                var includeResult = SingleItemSpecProvenance(itemElement.Include, itemElement.IncludeLocation, Operation.Include);
+                if (includeResult != null)
+                {
                     var excludeResult = SingleItemSpecProvenance(itemElement.Exclude, itemElement.ExcludeLocation, Operation.Exclude);
+                    if (excludeResult != null)
+                    {
+                        return excludeResult;
+                    }
 
-                    return excludeResult ?? includeResult;
-                },
+                    if (includeResult != null)
+                    {
+                        return includeResult;
+                    }
+                }
 
-                // provenance provider for update item elements
-                () => SingleItemSpecProvenance(itemElement.Update, itemElement.UpdateLocation, Operation.Update),
-                
-                // provenance provider for remove item elements
-                () => SingleItemSpecProvenance(itemElement.Remove, itemElement.RemoveLocation, Operation.Remove)
-            };
+                var result = SingleItemSpecProvenance(itemElement.Update, itemElement.UpdateLocation, Operation.Update);
+                if (result != null)
+                {
+                    return result;
+                }
 
-                return provenanceProviders.Select(provider => provider()).FirstOrDefault(provenanceResult => provenanceResult != null);
+                result = SingleItemSpecProvenance(itemElement.Remove, itemElement.RemoveLocation, Operation.Remove);
+                return result;
             }
 
             /// <summary>
--- a/src/Shared/FileMatcher.cs
+++ b/src/Shared/FileMatcher.cs
@@ -1465,7 +1465,7 @@ namespace Microsoft.Build.Shared
                 : null;
         }
 
-        internal delegate Tuple<string, string, string> FixupParts(
+        internal delegate (string fixedDirectoryPart, string recursiveDirectoryPart, string fileNamePart) FixupParts(
             string fixedDirectoryPart,
             string recursiveDirectoryPart,
             string filenamePart);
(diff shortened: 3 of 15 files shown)


KirillOsenkov:
FYI @svetkereMS this commit simplifies the logic in ComputeProvenanceResult. A Tuple was allocated unnecessarily and an array of lambdas with closures was also allocated unnecessarily. Hope I got the logic right.

src/Build/Definition/Project.cs
```diff
@@ -2686,37 +2686,37 @@ ProvenanceResult SingleItemSpecProvenance(string itemSpec, IElementLocation elem
                     }
 
                     var matchOccurrences = ItemMatchesInItemSpecString(itemToMatch, itemSpec, elementLocation, itemElement.ContainingProject.DirectoryPath, _data.Expander, out Provenance provenance);
-                    Tuple<Provenance, int> result = matchOccurrences > 0 ? Tuple.Create(provenance, matchOccurrences) : null;
```
Forgind, line 2689: Grrr...this was my part of my pet project for when I don't know what I should be working on. (It is related to an SLA bug I was assigned.)

I cleaned it up more. See [here](https://github.com/Forgind/msbuild/blob/get-item-provenance/src/Build/Definition/Project.cs#L2678-L2736). I don't think you broke anything, though.

src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
```diff
@@ -105,14 +105,14 @@ public static ProjectOnErrorInstance CreateTargetOnError(int? counter = null)
 
         public static ProjectTaskInstance CreateTargetTask(
             int? counter = null,
-            IDictionary<string, Tuple<string, ElementLocation>> parameters = null,
+            IDictionary<string, (string, ElementLocation)> parameters = null,
```
Forgind, line 108: Does putting ValueTuples in a Dictionary mean they get copied when they're added in or accessed and stored elsewhere? I don't know how Dictionary works very well, but I think this is the point:
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L652

src/Shared/FileMatcher.cs
```diff
@@ -1465,7 +1465,7 @@ private static int LastIndexOfDirectoryOrRecursiveSequence(string str, int start
                 : null;
         }
 
-        internal delegate Tuple<string, string, string> FixupParts(
+        internal delegate (string fixedDirectoryPart, string recursiveDirectoryPart, string fileNamePart) FixupParts(
```
Forgind, line 1468: I like this part.
👍

src/Shared/FrameworkLocationHelper.cs
```diff
@@ -308,28 +308,28 @@ internal static class FrameworkLocationHelper
         /// The rules are maintained in a 2-dimensions array. Each row defines a rule. The first column
         /// defines the trigger condition. The second column defines the fallback .net and VS versions.
         /// </remarks>
-        private static readonly Tuple<Version, Version>[,] s_explicitFallbackRulesForPathToDotNetFrameworkSdkTools =
+        private static readonly (Version, Version)[,] s_explicitFallbackRulesForPathToDotNetFrameworkSdkTools =
```
Forgind, line 311: This only seems to be used in one for loop in this class. It would be messier, but do you think it would be reasonable to move this there so we don't always allocate it? I imagine that whole code path isn't used too often. At minimum, it could be cleaned up by replacing Item1, Item2 with better names.

src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
KirillOsenkov: Think of a Dictionary conceptually like any other collection (basically array).

For ValueTypes, the structs are inlined, there are no separate objects anywhere. Reads and writes copy the entire struct back and forth.

For reference types, the objects exist elsewhere, with 8 (or 16) bytes overhead per object (not including the payload), and the array stores references (4 or 8 bytes). So in this case having a reference type is very wasteful.

src/Shared/FrameworkLocationHelper.cs
KirillOsenkov: We can clean up ad infinitum, so one step at a time ;) I deliberately refrained from further cleanup, to make this conceptually easier to review. One semantic change per commit, one refactoring pass per PR ;)

src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
KirillOsenkov: Also keep in mind that with "inline" structs that are lined up in memory you reap the benefits of cache locality. The CPU prefetches a chunk of memory into the cache and blazes through it. For reference types it has potentially to fetch a page for each reference because it is stored elsewhere in the heap.

src/Build/Definition/Project.cs
KirillOsenkov: let me know if you'd like me to take this commit out to not cause you conflicts

src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
Forgind: I was with you up to the last line. The reference type does one allocation (and the value type does 0), but when reading or writing, the reference type only has to copy a small number of bytes, so everything after that is almost free, whereas the value type has to copy a lot of information back and forth, which takes CPU time. Having one allocation sounds better to me. Can you explain further?

Forgind: Re: locality, if they're stored at roughly the same time, I imagine they'd have fairly good locality either way.

src/Build/Definition/Project.cs
Forgind: Don't worry about it. I don't have a timeline for when my commits might turn into a PR, and this is definitely a huge improvement over what was there.

src/Build.UnitTests/TestData/ProjectInstanceTestObjects.cs
KirillOsenkov: Remember I'm talking in general statements, one needs to profile for each specific case. But the problem is that you don't have just one allocation, you have potentially hundreds of thousands, and they all add GC pressure and huge memory overhead. Allocations in .NET is a big problem, so I'd rather opt into less allocations by default than optimize struct copying. Yes, structs will be copied back and forth but honestly in my experience I never remember that being a perf problem. It's always the allocations that kill you. CPU copies memory around all the time and it's insanely fast. Copying 16 bytes around is nothing compared to the work done when doubling an array to the next power of two when growing a list, and then effectively doing O(n^2) copies, and we never even talk about those costs.

And soon with ref structs etc. we'll even be able to avoid copying the structs, we'll be just passing a reference to the struct that stays in place.

KirillOsenkov: @Forgind you've nerd-sniped me and I made an actual benchmark :) https://github.com/KirillOsenkov/Benchmarks/blob/fbc760fec7f6b310c223e50f9f564ff0135917c7/Benchmarks/Tests/DictionaryOfValueVsReferenceTypes.cs#L10-L11

```
|         Method |     Mean |   Error |  StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |---------:|--------:|--------:|-------:|------:|------:|----------:|
|     ValueTypes | 303.1 ns | 4.62 ns | 4.32 ns | 0.0572 |     - |     - |     240 B |
| ReferenceTypes | 303.4 ns | 3.58 ns | 3.17 ns | 0.0658 |     - |     - |     276 B |
```

Basically there's no perf overhead from struct copying in this case, just 36 bytes less allocations ;) Free allocation reduction!

Forgind: Thanks! I wonder if part of why it's so efficient is that the items in the structs are tiny, so copying back and forth is no worse than copying a few pointers in the form of "what's in the struct."

That (and your comment about allocations) convinces me this is a good change. Don't feel the need to check this, but I'd be curious about it if it were a massive struct, i.e., a struct with a massive non-reference type, vs. a class. I'm not sure what that would be, though, since you can't use a class or a string or any other reference types, and a struct has minimal (0?) overhead. Maybe a struct with a large number of things in it would work, but then you start to wonder if anyone has ever done that.

...and then it would become fully irrelevant with ref structs.

src/Shared/FileMatcher.cs
```diff
@@ -1491,7 +1491,6 @@ private static int LastIndexOfDirectoryOrRecursiveSequence(string str, int start
             out bool isLegalFileSpec,
             FixupParts fixupParts = null)
         {
-            isLegalFileSpec = true;
```
cdmihai, line 1494: Did you check all the writes to `isLegalFileSpec` and make sure they do set it to true, and not just to false?

KirillOsenkov: Roslyn told me that this assignment is redundant and can be deleted. Roslyn is correct, on all filepaths the true value was overwritten.

PR #5273, opened 2020-04-13
Final branding_16.6

Forgind:


Forgind:
I think the final loc pass got included in the NuGet change, and I didn't notice, so just final branding. I set up the PR [here](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/244411).

Issue #5229, opened 2020-04-03
Intermittent hang in PR builds in CommandLine tests


rainersigwald:
See https://dev.azure.com/dnceng/public/_build/results?buildId=587250&view=logs&j=20125c81-7edb-551b-693a-61efae016b74&s=96ac2280-8cb4-5df5-99de-dd2da759617d&t=7ce25265-aabe-59d3-d73c-06619cfcc3c4&l=191

With [Microsoft.Build.CommandLine.UnitTests_net472_x86.log](https://github.com/microsoft/msbuild/files/4427170/Microsoft.Build.CommandLine.UnitTests_net472_x86.log)

Unfortunately I don't think there's enough info to figure out _which_ test hung.

```
Executing [D:\a\1\s\artifacts\bin\Microsoft.Build.CommandLine.UnitTests\Debug\net472\MSBuild.exe "C:\Users\VssAdministrator\AppData\Local\Temp\3x1sftec.xxx\Temporary034568f450ab433eb07d00131f651913\build.proj"  /nr:false /v:diag ]
```

Forgind:
Related:
Most recent hang was in CancelledBuild (ironic 😄)
[Microsoft.Build.Engine.UnitTests_netcoreapp2.1_x64.log](https://github.com/microsoft/msbuild/files/4488672/Microsoft.Build.Engine.UnitTests_netcoreapp2.1_x64.log)

rainersigwald:
We're either getting lucky and this went away or #5423 is hiding it. Closing in the hopes of the former 🤞

rainersigwald:
Got another instance in https://dev.azure.com/dnceng/public/_build/results?buildId=729007

Forgind:
The hang from @ostorc's PR was in BinaryLogContainsImportedFiles. @BenVillalobos (author), any thoughts?

YuliiaKovalova:
Close it as a low priority.
If hang happens again we will create a new ticket to track it.

PR #5277, opened 2020-04-13
Test if timeouts are from starvation

Forgind:
Every 10 seconds, print out both the total CPU time a process has used as well as its priority class. If it repeatedly prints a low value, that suggests a process is being starved.

This is designed to identify if the problem identified [here](https://github.com/microsoft/msbuild/pull/5265#issuecomment-612870526) was due to running at below normal priority.

@rainersigwald

### DIFF
--- a/src/MSBuild.UnitTests/XMake_Tests.cs
+++ b/src/MSBuild.UnitTests/XMake_Tests.cs
@@ -1218,6 +1218,7 @@ namespace Microsoft.Build.UnitTests
         [Fact]
         public void LowPriorityBuild()
         {
+            _output.WriteLine("Low priority build starting");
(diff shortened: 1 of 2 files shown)


rainersigwald:
I don't think this necessarily proves what you hope it to. For one thing, you're tracking only the entry point `msbuild.exe`, but the tests force NOINPROCNODE to create a worker. Would you expect significant CPU time in the scheduler for the one-project build?

Another thing to consider is adding a timeout to `RunnerUtilities.ExecMSBuild`. Then we could gracefully and quickly fail the test on a hang and possibly collect more information.

One other thing to note: based on the spaces in the command line, it looks like #5229 is happening in the `NormalPriorityBuild` test.

Forgind:
I haven't seen this problem crop up a lot recently.

PR #5279, opened 2020-04-14
Reimplement CopyOnWriteDictionary to wrap ImmutableDictionary

rainersigwald:
Draft because there's Midnight Coding™ in here that I haven't even self-reviewed yet.


src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -0,0 +1,173 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace System.Collections.Immutable
```
Forgind, line 6: What's wrong with [System.Collections.Immutable.ImmutableDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2?view=netcore-3.1)?

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -34,25 +36,13 @@ namespace Microsoft.Build.Collections
     /// be run in a separate appdomain.
     /// </comment>
     [Serializable]
-    internal class CopyOnWriteDictionary<K, V> : IDictionary<K, V>, IDictionary where V : class
+    internal class CopyOnWriteDictionary<K, V> : IDictionary<K, V>, IDictionary, ISerializable where V : class
```
Forgind, line 39: This class is internal; can you just delete it in favor of the ImmutableDictionary? If you do this and use System.Collections.Immutable.ImmutableDictionary, you can also get rid of the tests.

src/Utilities.UnitTests/TaskItem_Tests.cs
```diff
@@ -378,15 +378,24 @@ public void RemoteTaskItem()
                     itemsInThisAppDomain[i] = new TaskItem(creator.CreatedTaskItems[i]);
 
                     itemsInThisAppDomain[i].ItemSpec.ShouldBe(creator.CreatedTaskItems[i].ItemSpec);
-                    itemsInThisAppDomain[i].MetadataCount.ShouldBe(creator.CreatedTaskItems[i].MetadataCount + 1);
+                    //itemsInThisAppDomain[i].MetadataCount.ShouldBe(creator.CreatedTaskItems[i].MetadataCount + 1);
```
Forgind, line 381: How is this related to the CopyOnWriteDictionary vs. ImmutableDictionary problem?

Also, why did you comment this out instead of just deleting it if you don't need it?

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
rainersigwald: Nothing that a time machine wouldn't fix! This PR uses it where it's available, but it is not available in .NET 3.5. Which we use in MSBuildTaskHost.exe, to execute 3.5-only tasks.

So this is an API-compatible (at least what we use) reimplementation. It's inefficient, but I don't care: update your project to .NET 4. It came out in 2012!

We've used this strategy for types in `System.Collections.Concurrent`, too, though I think those might be full source copies.

src/Shared/CopyOnWriteDictionary.cs
rainersigwald: No, because of the serializability constraint. `ImmutableDictionary` isn't, so I can either encapsulate it and define the serialization here once, or push it to all classes that have a CopyOnWriteDictionary.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
Forgind: That makes sense, but it deserves 😞

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -80,54 +65,48 @@ internal CopyOnWriteDictionary(IEqualityComparer<K> keyComparer)
...
-                this[pair.Key] = pair.Value;
-            }
+            backing = dictionary.ToImmutableDictionary();
         }
 
         /// <summary>
         /// Returns the collection of keys in the dictionary.
         /// </summary>
-        public ICollection<K> Keys => ReadOperation.Keys;
+        public ICollection<K> Keys => ((IDictionary<K, V>)ReadOperation).Keys;
```
Forgind, line 104: Why is Keys returning an IDictionary? Same for Values. Seems like it should be a set rather than a map.

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -200,53 +179,37 @@ internal bool IsDummy
...
                 {
-                    backing = new CopyOnWriteBackingDictionary<K, V>(capacity, Comparer);
-                }
-                else
-                {
-                    lock (((ICollection)backing).SyncRoot)
-                    {
-                        backing = backing.CloneForWriteIfNecessary();
-                    }
+                    backing = ImmutableDictionary.Create<K,V>(Comparer);
```
Forgind, line 212: Does this keep all the mappings?

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -318,16 +272,11 @@ public bool ContainsKey(K key)
...
-                {
-                    return WriteOperation.Remove(key);
-                }
-            }
+            ImmutableDictionary<K, V> initial = backing;
 
-            return false;
+            backing = backing.Remove(key);
+
+            return initial != backing; // whether the removal occured
```
Forgind, line 279: Should we be cloning?

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -34,31 +30,20 @@ namespace Microsoft.Build.Collections
     /// be run in a separate appdomain.
     /// </comment>
     [Serializable]
-    internal class CopyOnWriteDictionary<K, V> : IDictionary<K, V>, IDictionary
+    internal class CopyOnWriteDictionary<K, V> : IDictionary<K, V>, IDictionary, ISerializable
```
Forgind, line 33: I didn't notice any places where this significantly diverged from ImmutableDictionary other than with the serialization part. Is there a problem with extending ImmutableDictionary and not overriding most of its methods?

ladipro: `ImmutableDictionary` is sealed.

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -34,31 +30,20 @@ namespace Microsoft.Build.Collections
...
-        /// The default capacity.
-        /// </summary>
-        private readonly int capacity;
-
         /// <summary>
         /// The backing dictionary.
         /// Lazily created.
         /// </summary>
-        private CopyOnWriteBackingDictionary<K, V> backing;
+        private ImmutableDictionary<K, V> backing;
```
ladipro, line 39: nit: `_backing` so it's super clear that it's a field. I wouldn't nit-pick on it but here seeing that the write methods mutate the object is critical to understanding the logic.

src/Shared/CopyOnWriteDictionary.cs
BenVillalobos: I was thinking the same thing, but calling `Remove` actually does the cloning and returns a new `ImmutableDictionary`

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -345,7 +294,7 @@ public void Add(KeyValuePair<K, V> item)
         {
             if (!IsDummy)
             {
-                ((IDictionary<K, V>)WriteOperation).Add(item);
+                backing = backing.SetItem(item.Key, item.Value);
```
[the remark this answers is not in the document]
ladipro: I have removed `Dummy` as it was unused, and null checks as they were redundant (`backing` is guaranteed non-null). Consequently `IsDummy` would never return true and both `ReadOperation` and `WriteOperation` became equivalent to just reading the `backing` field.

src/Shared/CopyOnWriteDictionary.cs
ladipro: This code was unreachable due to `backing` being non-nullable. The entire `WriteOperation` has been deleted.

src/Shared/CopyOnWriteDictionary.cs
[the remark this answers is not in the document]
ladipro: Fixed to returning `_backing.Keys`. No null checks are needed as the field is declared non-nullable and enforced as such by the compiler.

ladipro: (facepalm) It compiled because the compiler doesn't know if the object implements the interface or not unless its type is sealed. This would have failed at run-time in task host. I've fixed it by making the simple task host version of `ImmutableDictionary<K, V>` implement `IDictionary<K, V>`.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -0,0 +1,176 @@
...
+            return new ImmutableDictionary<K, V>(comparer);
+        }
+    }
+
+    /// <summary>
+    /// Inefficient ImmutableDictionary implementation: keep a mutable dictionary and wrap all operations.
+    /// </summary>
+    /// <typeparam name="K"></typeparam>
+    /// <typeparam name="V"></typeparam>
+    class ImmutableDictionary<K, V>
```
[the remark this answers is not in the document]
ladipro: The compiler doesn't do any sophisticated flow analysis and assumes that if a type is not `sealed`, a variable typed as such can hold an instance of its subclass that implements the interface we're casting to. All understood and fixed now.

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -80,54 +65,48 @@ internal CopyOnWriteDictionary(IEqualityComparer<K> keyComparer)
         /// <summary>
         /// Constructor taking a specified comparer for the keys and an initial capacity
         /// </summary>
-        internal CopyOnWriteDictionary(int capacity, IEqualityComparer<K> keyComparer)
+        internal CopyOnWriteDictionary(int capacity, IEqualityComparer<K>? keyComparer)
```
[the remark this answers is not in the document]
ladipro: Just removing the `?` wouldn't compile because this constructor is explicitly passed `null` a few lines above and `ImmutableDictionary.Create` actually does accept `null`. I'm not sure if nullable reference types, as introduced in C# 8.0, have already made it into official docs and all APIs are annotated accordingly. Judging by this occurrence it is probably not the case.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -0,0 +1,176 @@
...
+{
+    static class ImmutableExtensions
+    {
+        public static ImmutableDictionary<K,V> ToImmutableDictionary<K,V>(this IDictionary<K,V> dictionary)
+        {
+            return new ImmutableDictionary<K, V>(dictionary);
+        }
+    }
+
+    class ImmutableDictionary
```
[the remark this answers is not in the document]
ladipro: Done.

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -80,54 +65,48 @@ internal CopyOnWriteDictionary(IEqualityComparer<K> keyComparer)
...
-            }
         }
 
         public CopyOnWriteDictionary(IDictionary<K, V> dictionary)
         {
-            foreach (KeyValuePair<K, V> pair in dictionary)
-            {
-                this[pair.Key] = pair.Value;
-            }
+            backing = dictionary.ToImmutableDictionary();
```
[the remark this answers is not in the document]
ladipro: The parameter is not declared as nullable and the file is under `#nullable enable` so it should be guaranteed non-null. With the caveat that this is a public ctor so it depends on whether nullable reference type checks are enable in the calling code. However, there is only one place in the source tree calling this ctor and it is correctly checking for null so I'd say no need to add an extra check, one that would become redundant once the entire code base is compiled with nullable enabled.

src/Shared/CopyOnWriteDictionary.cs
Forgind: I had to read that several times to even understand what you were saying. Very nice catch!

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return _backing.GetEnumerator();
+        }
+
+        void ICollection<KeyValuePair<K, V>>.CopyTo(KeyValuePair<K, V>[] array, int arrayIndex)
+        {
+            foreach (var item in this)
+            {
```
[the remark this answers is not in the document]
ladipro: Done, thanks! And I really think that this code should be covered with tests before merging. We don't care about perf but functionally it should be solid as we don't want to risk breaking anything.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        //
+        // WRITE OPERATIONS
+        //
+
+        internal ImmutableDictionary<K, V> SetItem(K key, V value)
+        {
+            var clone = new ImmutableDictionary<K, V>(_backing);
+            clone._backing.Add(key, value);
```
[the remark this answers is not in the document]
ladipro: We could make the check in `Remove` but for `SetItem` we would have to have the concept of value comparer. From `ImmutableDictionary` docs:

_If the specified key/value pair already exists in the dictionary, this method returns the existing instance of the dictionary. If the key already exists but has a different value, this method returns a new instance of the dictionary with the overwritten value._

Also, fixed to call the indexer instead of `Add` on the backing dictionary to comply with the second sentence above.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -0,0 +1,247 @@
...
+        internal static ImmutableDictionary<K, V> Empty
+        {
+            get
+            {
+                return new ImmutableDictionary<K, V>();
+            }
+        }
+
+        public IEqualityComparer<K> KeyComparer { get => _backing.Comparer; internal set => throw new NotSupportedException(); }
+        public IEqualityComparer<K> ValueComparer { get => _backing.Comparer; internal set => throw new NotSupportedException(); }
```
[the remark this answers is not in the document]
ladipro: Great catch! We were calling `ValueComparer` from `CopyOnWriteDictionary` but it was redundant. I have removed this prop and its caller, thanks.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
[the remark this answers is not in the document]
Forgind: I think that even if the values are identical for all normal purposes, ImmutableDictionary still considers them different if they're different instances, so we could just use `!=`.

The way it currently is, if you rely on having the exact same object returned from SetItem before and after you try to add the same key with the same value, it'll work for the real implementation and fail here. That's such a trivial case that I don't think it's worth fixing, but it's worth bringing up. In contrast, changing from Add to the indexer changes whether it throws an error, which was worth it.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -101,12 +102,29 @@ IEnumerator IEnumerable.GetEnumerator()
...
             {
                 array.SetValue(new DictionaryEntry(item.Key, item.Value), arrayIndex++);
             }
         }
 
+        private void CheckCopyToArguments(Array array, int arrayIndex)
+        {
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
```
BenVillalobos, line 116: nit: We should eventually use `ErrorUtilities` here. Something like `ErrorUtilities.VerifyThrowArgumentNull` and some equivalent for the other checks if possible.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
```diff
@@ -172,5 +187,61 @@ internal ImmutableDictionary(IDictionary<K, V> source)
             var n = new Dictionary<K, V>(_backing, keyComparer);
             return new ImmutableDictionary<K, V>(n);
         }
+
+        //
+        // UNSUPPORTED OPERATIONS
```
BenVillalobos, line 209: super nit: region this section

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
Forgind: What are the rules around when we should use ErrorUtilities vs. not? ErrorUtilities when doing something MSBuild-specific, but I would've thought that in making a DS, we should try to match the real one as closely as possible.

BenVillalobos: I'm not sure, and the summary of the `ErrorUtilities` class doesn't say much: `This class contains methods that are useful for error checking and validation.` My assumption was that we should generally use those methods. I don't think using them hurts in this case, but as long as we're _specifically choosing_ to match the inherited class then I've no problem with that.

src/Shared/CopyOnWriteDictionary.cs
```diff
@@ -174,112 +153,29 @@ public CopyOnWriteDictionary(IDictionary<K, V> dictionary)
...
-                    else
-                    {
-                        // Try to avoid a clone if it already is present with the same value
-                        if (!ReadOperation.TryGetValue(key, out V existingValue) || !EqualityComparer<V>.Default.Equals(existingValue, value))
-                        {
-                            WriteOperation[key] = value;
-                        }
-                    }
-                }
+                _backing = _backing.SetItem(key, value);
```
cdmihai, line 174: Isn't this a data race on the contents of the `_backing` reference? The old implementation [used a lock](https://github.com/dotnet/msbuild/blob/master/src/Shared/CopyOnWriteDictionary.cs#L246) whenever it cloned the backing dict, but the new implementation does not lock anymore when it clones the backing dict. This means that multiple writers can race, right? The comments on the class also specify that multiple writers are made thread safe by locking the writes.

Forgind: _backing is immutable, so you should get a whole new dictionary every time you do this, preventing races.

cdmihai: I was referring to the `_backing` reference, not the ImmutableDictionary object it refers to.

Here's the scenario:

- You have two reader threads: T1, and T2
- Both threads have a reference to the same CopyOnWriteDictionary instance: COWD1. COWD1._backing points to ImmutableDictionary ID1
- ID1 has the following items in it: `{p1: foo, p2: bar}`
- T1 and T2 both decide to become writer threads
  - T1 calls COWD1[p3] = p3Val
  - T2 calls COWD1[p4] = p4Val
- The following interleaving happens
  - T1 setter sets COWD1._backing to ID2: `{p1: foo, p2: bar, p3: p3Val}`
  - T2 setter sets COWD1._backing to ID3: `{p1: foo, p2: bar, p4: p4Val}`

Now we have a situation in which T1's update got lost. The reference to ID2 gets lost.

Forgind: Ah, misunderstood. That sounds legit.

ladipro: I don't believe that the old implementation provided this guarantee. From a comment on the old class:

```C#
    /// Thread safety: for all users, this class is as thread safe as the underlying Dictionary implementation, that is,
    /// safe for concurrent readers or one writer from EACH user. It achieves this by locking itself and cloning before
    /// any write, if it is being shared - i.e., stopping sharing before any writes occur.
```

Note the emphasis on EACH user. The way I read this, you are not safe to write to the data structure from multiple threads concurrently unless each thread operates on its own clone.

This seems to be supported by the implementation. If multiple threads attempt the operation as described above and the COW dictionary has not been cloned, they both attempt to write to the underlying `Dictionary<K, V>` which sure is not thread-safe. In fact, the first write access will even race on this assignment:

```C#
                if (backing == null)
                {
                    backing = new CopyOnWriteBackingDictionary<K, V>(capacity, Comparer);
                }
```

so the exact issue you outlined above where one of the writes may be lost can occur here as well.

ladipro: I've put together this test code and it indeed throws against the old implementation:

```C#
            Random random = new Random();
            CopyOnWriteDictionary<int, int> dd = new CopyOnWriteDictionary<int, int>();
            ThreadStart func = () =>
            {
                while (true)
                {
                    int key = random.Next();
                    dd[key] = random.Next();
                    if (!dd.ContainsKey(key))
                    {
                        throw new Exception("Race!");
                    }
                }
            };
            Thread th1 = new Thread(func);
            Thread th2 = new Thread(func);

            th1.Start(); th2.Start();
            th1.Join(); th2.Join();
```

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
ladipro: Done, thanks!

ladipro:
> Looks good! Should we consider tossing this into an exp/ branch?

Pushed to exp/immutable-cow

BenVillalobos:
We're seeing failures in the pipeline `exp/` builds for this.

[First error](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3956950&view=results):
```
C:\A\_work\24\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.20365.6\tools\Sign.proj(81,5): error MSB4018: The "Microsoft.DotNet.SignTool.SignToolTask" task failed unexpectedly.
C:\A\_work\24\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.20365.6\tools\Sign.proj(81,5): error MSB4018: System.IO.IOException: The process cannot access the file 'C:\A\_work\24\s\artifacts\tmp\Release\ContainerSigning\270\Contents\MSBuild\Current\Bin\zh-Hans\Microsoft.Build.Tasks.Core.resources.dll' because it is being used by another process.
```

[Second error](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3956192&view=results):
```
C:\A\_work\15\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.20365.6\tools\OptimizationData.targets(110,5): error : No optimization data found for assemblies: C:\A\_work\15\s\artifacts\obj\Microsoft.Build.Framework\Release\net472\Microsoft.Build.Framework.dll [C:\A\_work\15\s\src\Framework\Microsoft.Build.Framework.csproj]
##[error].packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.20365.6\tools\OptimizationData.targets(110,5): error : No optimization data found for assemblies: C:\A\_work\15\s\artifacts\obj\Microsoft.Build.Framework\Release\net472\Microsoft.Build.Framework.dll
```

Forgind:
Did you kick off a new optprof run?

BenVillalobos:
@forgind I just did. I'm going to keep a log of the investigation here: https://github.com/dotnet/msbuild/issues/5627 (for future generations)

src/Shared/CopyOnWriteDictionary.cs
cdmihai: Aha, so an instance of COWD must be confined to one writer thread. Multiple writer threads must not use the same COWD instance. I guess I didn't make that connection from reading the class comments, I confused user with thread, and it wasn't clear whether "this dictionary" refers to the COWD type, or the wrapped dictionary :)

Under these constraints, yup, all looks good.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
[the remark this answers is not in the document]
ladipro: Tests added.

src/MSBuildTaskHost/Immutable/ImmutableDictionary.cs
ladipro: I think in this case using the bare `throw new ...` is OK as we're not exposing the data structure publicly and want to match the behavior of the real .NET 4.5 implementation.

marcpopMSFT:
Ladi will be doing an experimental insertion to collect data before merging this.

Forgind:
Experimental insertion looks good:
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/267887

PR #5280, opened 2020-04-14
Add additional parameters to referenceassemblyresolver

erickulcyk:
Referenceassemblyresolver target does not contain properties to control finding of extra files.  This results in extra file accesses being done when they are not wanted simply because the file does not exist.  We would like the ability to turn this functionality off without setting BuildingProject to false as that has other implications for build logic.


erickulcyk:
Does anyone know how to requeue the msbuild-pr task?  It seems to have timeout out without being my fault.

erickulcyk:
Does anyone know where the microsoft.common.targets file from the msbuild.corext package comes from?  It doesn't seem to be the same as the one in the msbuild repo.

PR #5286, opened 2020-04-15
Updating .NET Core branding to .NET

sfoslund:
Fixes 1/2 in dotnet/sdk#11217


rainersigwald:
Timing on this is pretty awkward, since we expect several releases before 5.0 RTM. But it’s the way the wind is blowing so sure.

PR #5287, opened 2020-04-16
[Static graph] Support transitive project references in static graph

cdmihai:
Transitive project references are a thing now. Added support to static graph, so that buildxl and qb can avoid adding the transitive refs.
This PR is independent of #5222. Ideally, review this one first, as QB has a stronger dependency on this PR than on #5222.

Design
- transitive references are opt-in, per project evaluation
- once a project opts-in, transitivity is applied for all ProjectReference items
- a project opt-ins by setting the property `AddTransitiveProjectReferencesInStaticGraph` to true. The sdk does this automatically in Microsoft.Managed.After.Targets.
- interaction with crosstargeting: transitive refs are added only to inner builds, not the outer builds. This mimics vanilla msbuild.


src/Build.UnitTests/Graph/GraphTestingUtilities.cs
```diff
@@ -205,5 +215,32 @@ internal static string GetProjectPath(ProjectGraphNode node)
                 defaultTargets ?? "Build",
                 extraContent);
         }
+
+        internal static IEnumerable<ProjectGraphNode> ComputeClosure(ProjectGraphNode node)
```
Forgind, line 219: This doesn't seem to have a way to detect cyclical references, and there could easily be two projects that reference a single project, thus adding all its references etc. twice. Maybe add an "I've seen this before" escape?

cdmihai: This is a test utility I use to assert transitive reference test results, not the real thing, so I'd rather keep it simple. Also, cycles are not allowed in static graph, so I don't concern myself with them.

src/Tasks/Microsoft.Managed.After.targets
```diff
@@ -20,13 +20,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
 
    <PropertyGroup Condition="'$(IsGraphBuild)' == 'true'">
-      <!-- Microsoft.Net.Sdk adds all transitive project references to ProjectReference. It does this at build time
-           via a target. This fails the undeclared msbuild reference /isolate constraint, so disable transitive
-           references.
-           Do not disable transitive references if BuildProjectReferences is false. If building references is false,
-           then MSBuild won't try to build the transitive references, which won't trigger undeclared reference constraints.
-          -->
-      <DisableTransitiveProjectReferences Condition="'$(BuildProjectReferences)' != 'false' and '$(DisableTransitiveProjectReferences)' == ''">true</DisableTransitiveProjectReferences>
+      <AddTransitiveProjectReferencesInStaticGraph Condition="'$(UsingMicrosoftNETSdk)' == 'true' and '$(DisableTransitiveProjectReferences)' != 'true'">true</AddTransitiveProjectReferencesInStaticGraph>
```
rainersigwald, line 23:
```suggestion
      <AddTransitiveProjectReferencesInStaticGraph Condition="'$(AddTransitiveProjectReferencesInStaticGraph)' == '' and '$(UsingMicrosoftNETSdk)' == 'true' and '$(DisableTransitiveProjectReferences)' != 'true'">true</AddTransitiveProjectReferencesInStaticGraph>
```

src/Build.UnitTests/Graph/GraphTestingUtilities.cs
```diff
@@ -17,20 +17,30 @@ internal static class GraphTestingUtilities
         public static readonly ImmutableDictionary<string, string> EmptyGlobalProperties = new Dictionary<string, string> {{PropertyNames.IsGraphBuild, "true"}}.ToImmutableDictionary();
 
         public static readonly string InnerBuildPropertyName = "InnerBuild";
+        public static readonly string InnerBuildPropertiesName = "InnerBuildProperties";
 
-        public static readonly string MultitargetingSpecification = $@"<PropertyGroup>
+        public static readonly string CrosstargetingSpecificationPropertyGroup = $@"<PropertyGroup>
```
rainersigwald, line 22: I thought we had moved everything to "multitargeting"? Why the change back?

src/Build/Graph/GraphBuilder.cs
```diff
@@ -110,28 +111,74 @@ private static IReadOnlyCollection<ProjectGraphNode> GetGraphRoots(IReadOnlyColl
...
+                                    directMetadata: null,
+                                    definingFileEscaped: currentNode.ProjectInstance.FullPath
+                                ),
+                                edges);
+                        }
+                    }
+                }
+            }
+
+            IEnumerable<ProjectGraphNode> GetTransitiveProjectReferences(ParsedProject parsedProject)
```
rainersigwald, line 163:
```suggestion
            List<ProjectGraphNode> GetTransitiveProjectReferences(ParsedProject parsedProject)
```

So that traversing it multiple times doesn't get slow?

src/Build/Graph/GraphBuilder.cs
```diff
@@ -110,28 +111,74 @@ private static IReadOnlyCollection<ProjectGraphNode> GetGraphRoots(IReadOnlyColl
...
+                    return cachedTransitiveReferences;
+                }
+                else
+                {
+                    var transitiveReferences = new List<ProjectGraphNode>();
+
+                    foreach (var referenceInfo in parsedProject.ReferenceInfos)
+                    {
+                        transitiveReferences.Add(allParsedProjects[referenceInfo.ReferenceConfiguration].GraphNode);
+                        transitiveReferences.AddRange(GetTransitiveProjectReferences(allParsedProjects[referenceInfo.ReferenceConfiguration]));
```
rainersigwald, line 176: Worth worrying about stack depth here? I hope not . . .

src/Build/Graph/ProjectInterpretation.cs
```diff
@@ -420,5 +436,26 @@ public ImmutableList<string> GetApplicableTargetsForReference(ProjectInstance re
...
+            // migrated to new enough Visual Studio versions whose Microsoft.Managed.After.Targets enable transitive references.
+            if (projectInstance.GetPropertyValue("UsingMicrosoftNETSdk").Equals("true", StringComparison.OrdinalIgnoreCase) &&
+                !projectInstance.GetPropertyValue("DisableTransitiveProjectReferences").Equals("true", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            return projectInstance
+                       .GetPropertyValue(AddTransitiveProjectReferencesInStaticGraphPropertyName)
+                       .Equals("true", StringComparison.OrdinalIgnoreCase);
```
rainersigwald, line 458: Should this and all the other ones like it use `ConversionUtilities.ConvertStringToBool` instead? So evil people can put `!no` instead of `tRuE`?

src/Build.UnitTests/Graph/GraphTestingUtilities.cs
cdmihai: I thought multitargeting was changed to crosstargeting :) What's the official name?

src/Build/Graph/GraphBuilder.cs
cdmihai: GraphBuilder.DetectCycles.VisitNode traverses the entire graph and is already recursive, so if that one's recursive, then this new one won't increase the risk of stack overflows.
If we ever find a graph that busts the stack, the temporary workaround for that repo would be to temporarily increase the stack size with `editbin.exe` while we change all recursive traversals to iterative.

If you really want to go iterative now, I'd do it as a separate PR and not block this one on it.

src/Build.UnitTests/Graph/GraphTestingUtilities.cs
Nirmal4G: I read sometime back that multitargeting was the official name but kept crosstargeting in the msbuild property names like `IsCrossTargetingBuild` and `Custom*MicrosoftCommonCrossTargetingTargets` because of compat.

rainersigwald: Crosstargeting is the old name, supported for compat. But for new property names, call them Multitargeting.

src/Build/Graph/GraphBuilder.cs
rainersigwald: Nah, "not worse than before" is ok.

src/Build.UnitTests/Graph/GraphTestingUtilities.cs
cdmihai: Alright, moved the naming back :)

PR #5290, opened 2020-04-16
Introduce timeout to prevent a hang in CancelledBuild

Forgind:
Works towards #5229.

### DIFF
--- a/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
@@ -1517,7 +1517,9 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             asyncResult.ExecuteAsync(null, null);
             _buildManager.CancelAllSubmissions();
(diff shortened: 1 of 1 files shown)


src/Build.UnitTests/BackEnd/BuildManager_Tests.cs
```diff
@@ -1517,7 +1517,8 @@ public void CancelledBuild()
 
             asyncResult.ExecuteAsync(null, null);
             _buildManager.CancelAllSubmissions();
-            asyncResult.WaitHandle.WaitOne();
+            asyncResult.WaitHandle.WaitOne(10000);
```
rainersigwald, line 1520:
```suggestion
            asyncResult.WaitHandle.WaitOne(TimeSpan.FromSeconds(10));
```

I always prefer the explicit overload, so you don't have to remember units.

rainersigwald, line 1520: Comment the arbitrary timeout please.

PR #5292, opened 2020-04-17
Eliminate boxing from OpportunisticIntern string interning paths

ladipro:
This change is addressing unnecessary allocations on the string interning path. The `IInternable` implementations (except `ReuseableStringBuilder`) are struct but they end up on the managed heap anyways because of the implicit boxing taking place at the point where they are passed as an `IInternable` argument. Also, accessing the structures via the interface is relatively expensive and prevents inlining.

Fixed in this PR by genericizing the relevant methods, changing their signatures from taking `IInternable` to taking the `T` type parameter with an interface constraint.

Other changes:
- Replaced the unrolled by-character checks for hardcoded interned strings with a helper call. The old code was problematic especially given the fact that each character read was an interface call. With the changes in this PR it is not that burning but it's better to keep the generic methods small to avoid unnecessary code bloat. We let the runtime handle inlining and loop unrolling. Not to mention readability.

- Renamed `IsOrdinalEqualToStringOfSameLength` to `StartsWithStringByOrdinalComparison` and changed the semantics accordingly to support the detection of `MSBuildDummyGlobalPropertyHeader`.

- Made `InternWithStatistics` run under a lock because it's accessing non-thread-safe data structures and stat gathering (via MSBUILDDUMPOPPORTUNISTICINTERNSTATS) was failing with exceptions in parallel builds.

The net effects of this change are:
- No GC allocations in `OpportunisticIntern` steady state operation, like for example interning strings with `InternStringIfPossible()`.
- &gt;10% faster `InternableToString()` as measured with a micro-benchmark passing a mix of strings, both hardcoded (e.g. "True") and random.

### DIFF
--- a/src/Shared/OpportunisticIntern.cs
+++ b/src/Shared/OpportunisticIntern.cs
@@ -715,6 +715,8 @@ namespace Microsoft.Build
 
             private bool TryInternHardcodedString<T>(T candidate, string str, ref string interned) where T : IInternable
             {
+                Debug.Assert(candidate.Length == str.Length);
+
                 if (candidate.StartsWithStringByOrdinalComparison(str))
                 {
                     interned = str;
--- a/src/Shared/OpportunisticIntern.cs
+++ b/src/Shared/OpportunisticIntern.cs
@@ -116,9 +116,9 @@ namespace Microsoft.Build
             string ExpensiveConvertToString();
 
             /// <summary>
-            /// Compare target to string. Assumes lengths are equal.
+            /// Compare target to string. Assumes string is of equal or smaller length than target.
             /// </summary>
-            bool IsOrdinalEqualToStringOfSameLength(string other);
+            bool StartsWithStringByOrdinalComparison(string other);
 
             /// <summary>
             /// Reference compare target to string. If target is non-string this should return false.
@@ -160,7 +160,7 @@ namespace Microsoft.Build
         /// <summary>
         /// Intern the given internable.
         /// </summary>
-        internal static string InternableToString(IInternable candidate)
+        internal static string InternableToString<T>(T candidate) where T : IInternable
         {
             if (s_whatIfInfinite != null)
             {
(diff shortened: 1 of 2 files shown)


src/Shared/OpportunisticIntern.cs
```diff
@@ -713,15 +713,26 @@ internal void ReportStatistics(string heading)
                 Console.WriteLine("##########Top Rejected Strings: \n{0} ", string.Join("\n==============\n", topRejectedString.ToArray()));
             }
 
+            private bool TryInternHardcodedString<T>(T candidate, string str, ref string interned) where T : IInternable
+            {
+                if (candidate.StartsWithStringByOrdinalComparison(str))
```
rainersigwald, line 718: Should there be a debug assert here that the lengths match too? When I read it I thought "this can't be right" but it's fine because you're checking the length outside the call to this method.

src/Shared/OpportunisticIntern.cs
```diff
@@ -752,147 +763,56 @@ internal void ReportStatistics(string heading)
...
-                            interned = "true";
-                            return true;
-                        }
-
-                        if (candidate[0] == 'v' && candidate[1] == '4' && candidate[2] == '.' && candidate[3] == '0')
+                        if (TryInternHardcodedString(candidate, "TRUE", ref interned) ||
+                            TryInternHardcodedString(candidate, "True", ref interned) ||
+                            TryInternHardcodedString(candidate, "Copy", ref interned) ||
+                            TryInternHardcodedString(candidate, "true", ref interned) ||
+                            TryInternHardcodedString(candidate, "v4.0", ref interned))
```
rainersigwald, line 770: Amazing how much clearer this is. And it makes it obvious that we should think about adding some new common strings like `netstandard2.0` and `net48` and friends.

src/Shared/OpportunisticIntern.cs
ladipro: Done, thanks!

src/Shared/OpportunisticIntern.cs
ladipro: Yup, new strings are coming soon.

PR #5297, opened 2020-04-17
[Static graph] Support skipping nonexisting targets

cdmihai:
Depends on #5222. Review that one first.

In msbuild, one can call targets which do not exist by setting `SkipNonexistentTargets` on the MSBuild task. This PR adds support for nonexistent targets in the static graph [target protocol](https://github.com/microsoft/msbuild/blob/master/documentation/specs/static-graph.md#inferring-which-targets-to-run-for-a-project-within-the-graph).

To mark a target as optional, users can now define the following: `<ProjectReferenceTargets Include='Build' Targets='GetTargetFrameworks' SkipNonexistentTargets='true'>`.

Implementation overview: `ProjectGraph.GetTargetLists()` removes nonexistent targets. Since nonexistent targets do not have a BuildResult, the Scheduler now needs to know about this case when it checks whether cache misses are allowed. This required saving a project's targets in a new `BuildRequestConfiguration.ProjectTargets`, and a corresponding field in `BuildResult.Targets` which ferries the targets from the out of proc nodes to the build manager node.

### DIFF
--- a/src/Build/BackEnd/Components/Scheduler/Scheduler.cs
+++ b/src/Build/BackEnd/Components/Scheduler/Scheduler.cs
@@ -1586,6 +1586,11 @@ namespace Microsoft.Build.BackEnd
                 {
                     emitNonErrorLogs(_componentHost.LoggingService);
 
+                    if (request.SkipStaticGraphIsolationConstraints)
+                    {
+                        _configCache[request.ConfigurationId].SkippedFromStaticGraphIsolationConstraints = true;
+                    }
+
                     // Ensure there is no affinity mismatch between this request and a previous request of the same configuration.
                     NodeAffinity requestAffinity = GetNodeAffinityForRequest(request);
                     NodeAffinity existingRequestAffinity = NodeAffinity.Any;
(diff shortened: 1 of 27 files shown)


src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1893,6 +1924,39 @@ string ConcatenateGlobalProperties(BuildRequestConfiguration configuration)
...
+                {
+                    return false;
+                }
+
+                // Non-existing targets do not have results.
+                // We must differentiate targets that do not exist in the reference, from targets that exist but have no results.
+                // In isolated builds, cache misses from the former are accepted for requests that skip non-existing targets.
+                // Cache misses from the latter break the isolation constraint of no cache misses, even when non-existing targets get skipped.
+                bool any = false;
+                foreach (var target in request.Targets)
```
BartoszKlonowski, line 1948: Perhaps `request.Targets.Any()` would be easier to read?

BartoszKlonowski (commented):
A lot of hard work in this PR.
The only disadvantage of making a PR this size is the difficulty to review it in details. Perhaps this is the reason for this PR to be still unmerged?
I've made some minor comments about the code but there's still the functionality to be checked.
Did you compare unit tests results before and after these changes?

cdmihai:
Closing until this problem becomes more pressing.

PR #5302, opened 2020-04-21
Override default Arcade Xunit configuration

cdmihai:


src/Directory.Build.targets
```diff
@@ -17,7 +17,10 @@
     <!-- Set output type of test projects to EXE.  This will allow us to run tests on .NET Core via F5 or "dotnet run" -->
     <OutputType>Exe</OutputType>
     <GenerateProgramFile>false</GenerateProgramFile>
-
+    
+    <!-- Makes xunit print all test names in stdout -->
+    <XunitOptions>$(XunitOptions) -diagnostics</XunitOptions>
```
cdmihai, line 22: Preventing arcade from overriding the xunit json makes this line useless. But I guess it doesn't hurt, in case other random things overwrite the json in the future.

src/Directory.Build.targets
```diff
@@ -38,6 +41,10 @@
     <TestRunnerAdditionalArguments>$(XunitOptions)</TestRunnerAdditionalArguments>
     
     <PackageOutputPath Condition="'$(IsVisualStudioInsertionPackage)' == 'true'">$(DevDivPackagesDir)</PackageOutputPath>
+
+    <!-- Arcade sdk also carries an xunit.runner.json which sometimes overrides the one in this repo. Assign a value to the arcade properties XUnitDesktopSettingsFile and XUnitCoreSettingsFile to prevent the arcade version of the file being added. -->
+    <XUnitDesktopSettingsFile>$(MSBuildThisFileDirectory)Shared\UnitTests\xunit.runner.json</XUnitDesktopSettingsFile>
```
cdmihai, line 46: @Forgind AFAIK you were investigating test hangs. If you copy this diff, I have a suspicion long running tests will start getting reported.

MSBuild.Dev.sln
```diff
@@ -3,6 +3,12 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.27004.2009
 MinimumVisualStudioVersion = 10.0.40219.1
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{4900B3B8-4310-4D5B-B1F7-2FDF9199765F}"
```
cdmihai, line 6: This snippet is in MSBuild.sln but not in MSBuild.dev.sln. So I copied it over. No idea what it does, but why not? :)

rainersigwald: It makes these things appear in the sidebar:

image

src/Directory.Build.targets
```diff
@@ -38,6 +41,10 @@
     <TestRunnerAdditionalArguments>$(XunitOptions)</TestRunnerAdditionalArguments>
     
     <PackageOutputPath Condition="'$(IsVisualStudioInsertionPackage)' == 'true'">$(DevDivPackagesDir)</PackageOutputPath>
+
+    <!-- Arcade sdk also carries an xunit.runner.json which sometimes overrides the one in this repo. Assign a value to the arcade properties XUnitDesktopSettingsFile and XUnitCoreSettingsFile to prevent the arcade version of the file being added. -->
```
rainersigwald, line 45: can you put some color on "sometimes"? Should I go yell at them?

src/Directory.Build.targets
rainersigwald: image

cdmihai: Also, from what I read, the long running tests option does not make xunit kill long running tests, it just reports them.

MSBuild.Dev.sln
cdmihai: That's the first degree effect, but does it lead to those files getting picked up by anything else in VS? Forever a mystery.

src/Directory.Build.targets
cdmihai: To be honest I didn't understand why it only happens on `Microsoft.Build.Engine.UnitTests.csproj` and not the rest. When I compare the logs to other projects, it all looks the same until the copy tasks does otherwise on that one project. So I just went with this fix.

src/Directory.Build.targets
Forgind: What's odd is that it, at least sometimes, already works. If you look at the log I attached to the issue you mentioned, it reports that a test was taking a long time. Then it reports it again. And again. And a number of other times. But that made it easy for me to insert an explicit timeout on that test, so that test should be slightly less of a problem in the future.

MSBuild.Dev.sln
rainersigwald: I don't think it should but I'm not willing to say it with that much confidence.

src/Directory.Build.targets
cdmihai: > What's odd is that it, at least sometimes, already works.

I'd guess it's nondeterministic whether the arcade xunit json overwrites ours, at least that's what the logs suggest.

PR #5303, opened 2020-04-21
Compute hashes in parallel in GetFileHash

pranavkm:
Based on investigations here: https://github.com/dotnet/aspnetcore/pull/21021. 

Hashes produced by GetFileHash can be calculated in parallel. This scales better for larger number of files. Some before and after with this change:

### Before
image

### After
image

### Before
image

### After
image

In addition, this PR has a few more cleanups as suggested here by @bartonjs https://github.com/dotnet/aspnetcore/pull/21021#discussion_r411714968

### DIFF
--- a/src/Tasks/FileIO/GetFileHash.cs
+++ b/src/Tasks/FileIO/GetFileHash.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Build.Tasks
 
                 lock (writeLock)
                 {
-                    // We cannot guaranteee Files instances are uniques. Write to it inside a lock to
+                    // We cannot guarantee Files instances are unique. Write to it inside a lock to
                     // avoid concurrent edits.
                     file.SetMetadata("FileHashAlgorithm", Algorithm);
                     file.SetMetadata(MetadataName, encodedHash);
--- a/src/Tasks/FileIO/GetFileHash.cs
+++ b/src/Tasks/FileIO/GetFileHash.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Security.Cryptography;
+using System.Threading.Tasks;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Shared.FileSystem;
@@ -19,6 +20,13 @@ namespace Microsoft.Build.Tasks
         internal const string _defaultFileHashAlgorithm = "SHA256";
         internal const string _hashEncodingHex = "hex";
         internal const string _hashEncodingBase64 = "base64";
+        internal static readonly Dictionary<string, Func<HashAlgorithm>> SupportedAlgorithms
+            = new Dictionary<string, Func<HashAlgorithm>>(StringComparer.OrdinalIgnoreCase)
+            {
+                ["SHA256"] = SHA256.Create,
+                ["SHA384"] = SHA384.Create,
+                ["SHA512"] = SHA512.Create,
+            };
 
         /// <summary>
         /// The files to be hashed.
@@ -55,7 +63,7 @@ namespace Microsoft.Build.Tasks
 
         public override bool Execute()
         {
-            if (!SupportsAlgorithm(Algorithm))
+            if (!SupportedAlgorithms.TryGetValue(Algorithm, out var algorithmFactory))
             {
                 Log.LogErrorWithCodeFromResources("FileHash.UnrecognizedHashAlgorithm", Algorithm);
                 return false;
@@ -67,25 +75,27 @@ namespace Microsoft.Build.Tasks
                 return false;
             }
 
-            foreach (var file in Files)
+            Parallel.For(0, Files.Length, index =>
             {
+                var file = Files[index];
                 if (!FileSystems.Default.FileExists(file.ItemSpec))
                 {
                     Log.LogErrorWithCodeFromResources("FileHash.FileNotFound", file.ItemSpec);
                 }
-            }
+            });
 
             if (Log.HasLoggedErrors)
             {
                 return false;
             }
 
-            foreach (var file in Files)
+            Parallel.For(0, Files.Length, index =>
             {
-                var hash = ComputeHash(Algorithm, file.ItemSpec);
+                var file = Files[index];
+                var hash = ComputeHash(algorithmFactory, file.ItemSpec);
                 file.SetMetadata("FileHashAlgorithm", Algorithm);
                 file.SetMetadata(MetadataName, EncodeHash(encoding, hash));
-            }
+            });
 
             Items = Files;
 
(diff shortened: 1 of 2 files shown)


src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -67,25 +75,27 @@ public override bool Execute()
...
                 return false;
             }
 
-            foreach (var file in Files)
+            Parallel.For(0, Files.Length, index =>
             {
-                var hash = ComputeHash(Algorithm, file.ItemSpec);
+                var file = Files[index];
+                var hash = ComputeHash(algorithmFactory, file.ItemSpec);
                 file.SetMetadata("FileHashAlgorithm", Algorithm);
```
ladipro, line 96: This call doesn't appear to be thread safe. You should do it under a lock or defer setting the metadata to after the parallel loop.

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -67,25 +75,27 @@ public override bool Execute()
                 return false;
             }
 
-            foreach (var file in Files)
+            Parallel.For(0, Files.Length, index =>
```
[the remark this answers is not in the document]
ladipro: Thank you for the insight! A little bit of readability vs. a little bit of perf, so looks like a wash. No strong preference.

src/Tasks/FileIO/GetFileHash.cs
ladipro: I was thinking about duplicate items in the `Files` array but haven't succeeded in reproducing this. We may always wrap items with unique instances in which case this would indeed be safe. But it sure feels like an unnecessarily brittle invariant to rely on.

ladipro:
As for testing I/O vs. CPU, I'd try building a simple project that just computes hashes of large files and compare the cold (freshly booted machine) and hot (project already built a couple of times, files are cached) build times. Thanks!

src/Tasks/FileIO/GetFileHash.cs
cdmihai, line 78: Why not merge the two for loops? Stop the loop and task and error on first file that's not found.

src/Tasks/FileIO/GetFileHash.cs
cdmihai: It wouldn't be incorrect to take out duplicates right?

src/Tasks/FileIO/GetFileHash.cs
Forgind: You'd waste computational resources calculating hashes, setting metadata, etc. for all the files that didn't fail. It's also messy and error-prone to stop a parallel process in the middle and make all the other parallel processes stop at the same time.

On the other hand, that would reduce startup costs. Might be an opportunity to get some telemetry and see which is better.

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -113,39 +123,13 @@ internal static string EncodeHash(HashEncoding encoding, byte[] hash)
...
 
-        internal static bool SupportsAlgorithm(string algorithmName)
-            => _supportedAlgorithms.Contains(algorithmName);
-
-        internal static byte[] ComputeHash(string algorithmName, string filePath)
+        internal static byte[] ComputeHash(Func<HashAlgorithm> algorithmFactory, string filePath)
         {
             using (var stream = File.OpenRead(filePath))
-            using (var algorithm = CreateAlgorithm(algorithmName))
+            using (var algorithm = algorithmFactory())
```
Forgind, line 129: Can you replace these with `return algorithmFactory().ComputerHash(File.OpenRead(filePath));`? If so, then you might consider inlining it, since it's only used twice.

Forgind:
(Pending performance checks.)

pranavkm:
I've been a little busy this week, I'll try and get some numbers over the weekend.

src/Tasks/FileIO/GetFileHash.cs
pranavkm: I'd intended to keep the original behavior as-is, but it shouldn't be too hard to combine these. Shouldn't be incredibly difficult to compare how the two perform when everything is present, versus when some inputs are missing

src/Tasks/FileIO/GetFileHash.cs
pranavkm: Should be fairly easily to extract changing the array out if you think it's necessary. Let me know and I'll update accordingly.

src/Tasks/FileIO/GetFileHash.cs
cdmihai: The common case is to not fail, so the potential perf gains of merging the loops would outweigh the perf hit during failures.

src/Tasks/FileIO/GetFileHash.cs
rainersigwald: Team triage: Ladi has successfully scared Rainer about this. Please do extract the SetMetadata outside the Parallel.For.

src/Tasks/FileIO/GetFileHash.cs
pranavkm: Wouldn't you need to dispose the algorithm and the file stream instance? Changing it to your suggestion changes the semantic of that code.

Forgind: You're right.

pranavkm:
Updated. @ladipro I tried your suggestion for cold versus warm builds, but I see very little variance in the numbers once a project gets built. I tested this on a disk drive (not SSD), my guess is that IO is cached after an initial read. Some more numbers from my local tests in case they're interesting:

## File does not exist

### Before (100 files, 4k each)

File at start:
        47 ms  GetFileHash                                1 calls

File at end:
        48 ms  GetFileHash                                1 calls
        

### After (100 files, 4k each)

File at start:
    50 ms  GetFileHash                                1 calls
       
File at end:
    55 ms  GetFileHash                                1 calls
       

## All files exist (5 files, 4kb each)

### Before
42 ms  GetFileHash                                1 calls

### After
43 ms  GetFileHash                                1 calls

## All files exist (100 files, 4kb each)

### Before
49 ms  GetFileHash                                1 calls

### After
46 ms  GetFileHash                                1 calls

## All files exist (5 files, 200 kb each)

### Before
47 ms  GetFileHash                                1 calls

### After
44 ms  GetFileHash                                1 calls

## All files exist (100 files, 200 kb each)

### Before
132 ms  GetFileHash                                1 calls

### After
61 ms  GetFileHash                                1 calls

I tested this on an MSBuild built locally in Release configuration with and without my changes. Using the one in the SDK gives very different results (including time to start being quicker). I suspect that's because it's been ngened

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -67,26 +75,33 @@ public override bool Execute()
...
                 {
                     Log.LogErrorWithCodeFromResources("FileHash.FileNotFound", file.ItemSpec);
+                    return;
                 }
-            }
+
+                var hash = ComputeHash(algorithmFactory, file.ItemSpec);
+                var encodedHash = EncodeHash(encoding, hash);
+
+                lock (file)
```
ladipro, line 91: nit: Locking on public objects should generally be avoided.

pranavkm: Fixed

src/Tasks/FileIO/GetFileHash.cs
```diff
@@ -67,26 +75,34 @@ public override bool Execute()
...
+                    return;
                 }
-            }
+
+                var hash = ComputeHash(algorithmFactory, file.ItemSpec);
+                var encodedHash = EncodeHash(encoding, hash);
+
+                lock (writeLock)
+                {
+                    // We cannot guaranteee Files instances are uniques. Write to it inside a lock to
```
rainersigwald, line 94:
```suggestion
                    // We cannot guarantee Files instances are unique. Write to it inside a lock to
```

Forgind:
Thank you!

PR #5304, opened 2020-04-21
Add evaluation shortcuts for commonly used String ctors

### LINKED ISSUE #5275: New string with empty metadata value chooses wrong constructor
When trying to create a new string and passing in a metadata value, MSBuild tries to bind the the constructor that takes no arguments.  Unfortunately `String` does not have a parameterless constructor and so you get an error.  MSBuild should be using `string.Empty` for these values but for some reason it goes for the wrong overload.

The workaround is to surround the expression with quotes to ensure that the correct overload is chosen.

The below hack is in place because its the only way to call instance methods on metadata values. 
(Related https://github.com/microsoft/msbuild/issues/1155)

### Steps to reproduce

Either include a project sample, attach a zipped project, or provide IDE / CLI steps to create the project and repro the behaviour. Example of a project sample:

Project file
```xml
<Project>
  <ItemGroup>
    <None Include="SomeFolder\**"
          Link="$([System.String]::new(%(RecursiveDir)).Replace(`foo`, `bar`))%(Filename)%(Extension)" />
  </ItemGroup>

  <Target Name="Build">
    <Message Text="%(None.Identity)" Importance="High" />
  </Target>
</Project>
```

Directory contents:
```
│   test.proj
│
└───SomeFolder
        test.txt
```

Command line
```
msbuild /bl
```
### Expected behavior
MSBuild creates a string and calls `.Replace()`.

### Actual behavior
```
error MSB4184: The expression "[System.String]::new()" cannot be evaluated. Attempted to access a missing method.
```

### Workaround
You can places quotes around `%(RecursiveDir)` to force MSBuild to use the overload of `new String()` that accepts a string.
```diff
- $([System.String]::new(%(RecursiveDir))
+ $([System.String]::new(`%(RecursiveDir)`)
```
### Environment data
`msbuild /version` output:

OS info:

If applicable, version of the tool that invokes MSBuild (Visual Studio, dotnet CLI, etc):

ladipro:
```$([System.String]::new(%(SomeVar))``` seems to be a commonly used pattern worth optimizing. Without this change, if SomeVar is not empty, it is handled by the general late binding logic coercing the argument to ```char[]``` and calling ```String(char[])```. That is a redundant conversion from string to char array and then from char array back to string. If the argument is empty, it fails to find a matching constructor, which is technically correct but not what the user expects.

Fixes #5275

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -2489,7 +2489,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
         /// Expand property function that is defined (on CoreFX) in an assembly named after its full namespace.
         /// </summary>
         [Fact]
-        public void PropertyStaticFunctioLocatedFromAssemblyWithNamespaceName()
+        public void PropertyStaticFunctionLocatedFromAssemblyWithNamespaceName()
         {
             PropertyDictionary<ProjectPropertyInstance> pg = new PropertyDictionary<ProjectPropertyInstance>();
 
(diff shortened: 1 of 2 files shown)


ladipro:
> Thanks. You should add a unit test that shows that it no longer fails when a passed in metadatum isn’t defined.

Added, thank you!

PR #5307, opened 2020-04-21
[Don't Merge] Exp/resource management

BenVillalobos:
Creating this PR to get a clean diff and inspect all changes in one place.

(if there's a better way to do this without creating a PR, please enlighten me 🙂)

Edit: Now that I think about it, I could've NOT submitted the PR and looked at the files changed from the Create Pull Request page.


Forgind:
So...do you want to keep this? I don't know if it's building towards something or not.

BenVillalobos:
@Forgind for the time being yes. I'll close it when we're done with this cpp resource investigation

danmosemsft:
> (if there's a better way to do this without creating a PR, please enlighten me 🙂)

You can make PR's in your own fork and direct people to them.

yuehuang010:
IMO. I don't see a world where RequestCores(1) is not always 1.
RequestCores should have a keep trying mode with a cancellation.  If the requester got 0, then most of them will just retry?  Might as well implement it for them.  I used a volatile bool instead of CTS because it is quicker.

PR #5318, opened 2020-04-27
Enable handshake timeout on Mono

ladipro:
I am seeing consistent test hangs on macOS Mono. A node is getting into a bad state and is failing to respond to handshakes. While I do not yet understand the root cause, it is clear that having a timeout on the handshake operation mitigates the issue. The test is now failing but does not hang, saving developer time as well as test pipeline resources.

### DIFF
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -369,13 +369,13 @@ namespace Microsoft.Build.Internal
         /// Extension method to read a series of bytes from a stream
         /// </summary>
         internal static long ReadLongForHandshake(this PipeStream stream
-#if NETCOREAPP2_1
+#if NETCOREAPP2_1 || MONO
             , int handshakeReadTimeout
 #endif
             )
         {
             return stream.ReadLongForHandshake((byte[])null, 0
-#if NETCOREAPP2_1
+#if NETCOREAPP2_1 || MONO
                 , handshakeReadTimeout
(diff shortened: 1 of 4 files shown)


src/Shared/CommunicationsUtilities.cs
```diff
@@ -387,14 +387,14 @@ internal static void WriteLongForHandshake(this PipeStream stream, long value)
...
+#if NETCOREAPP2_1 || MONO
             , int timeout
 #endif
             )
         {
             byte[] bytes = new byte[8];
 
-#if NETCOREAPP2_1
+#if NETCOREAPP2_1 || MONO
             if (!NativeMethodsShared.IsWindows)
```
ladipro, line 398: Out of curiosity:
1. Why was this enabled only for Core 2.1 and not for all non-Framework builds?
2. Why don't we use a/the timeout in the else branch below? It looks like `stream.ReadByte()` may also hang forever if the node is broken.

rainersigwald: 1. We're pretty inconsistent across the codebase with our ifdefs. This was (*checks history to make sure it was me*) probably just sloppiness and forgetting about Mono. Unfortunately, since Mono presents as Framework, we can't just do `#if !NETFRAMEWORK`
2. I think that was mostly "don't change legacy code without good reason" + "no async or timeout-enabled `PipeStream.ReadByte()`. But I don't really remember :(

PR #5322, opened 2020-04-28
Fix ProjectRootElement.SdkReference

### LINKED ISSUE #5320: ProjectImportElement.ParseSdkReference has wrong value for projects created with Construction API
### Steps to reproduce

Use the following code:
```cs
ProjectRootElement projectRootElement = ProjectRootElement.Create(NewProjectFileOptions.None);

ProjectImportElement importElement = projectRootElement.AddImport("Sdk.props");
importElement.Sdk = "Microsoft.Build.NoTargets";
importElement.Version = "1.0.53";

projectRootElement.Save("Test.csproj");

Project project = Project.FromProjectRootElement(projectRootElement, new ProjectOptions());
```
### Expected behavior
The project is successfully evaluated

### Actual behavior
The ParsedSdkReference is missing the value of Version and so the NuGet SDK resolver does not resolve the SDK.

The bug appears to be here: https://github.com/microsoft/msbuild/blob/master/src/Build/Construction/ProjectImportElement.cs#L192
The calculation is coming from the element before it's set which can lead to this condition.
image

The logic was introduced here: https://github.com/microsoft/msbuild/pull/2038

jeffkl:
The logic around whether or not to mark the project dirty when changing one of the three properties for an SDK on an import was not taking into account that a new value might be coming from a setter.  The code now accepts these values and calculates whether or not the set operation affects the import element.

I also renamed the property to make it more clear what it represents.

I added a unit test before fixing the code to verify that this is correctly verified.

Fixes #5320

### DIFF
--- a/src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
@@ -511,9 +511,9 @@ namespace Microsoft.Build.UnitTests.OM.Construction
                 var import = imports[i];
                 var importingElement = import.ImportingElement;
                 importingElement.Sdk.ShouldBe(SdkName + $"/{version}");
-                importingElement.ParsedSdkReference.Name.ShouldBe(SdkName);
-                importingElement.ParsedSdkReference.Version.ShouldBe(expectedVersion);
-                importingElement.ParsedSdkReference.MinimumVersion.ShouldBe(expectedMinimumVersion);
+                importingElement.SdkReference.Name.ShouldBe(SdkName);
+                importingElement.SdkReference.Version.ShouldBe(expectedVersion);
+                importingElement.SdkReference.MinimumVersion.ShouldBe(expectedMinimumVersion);
                 importingElement.SdkLocation.ShouldBe(ElementLocation.EmptyLocation);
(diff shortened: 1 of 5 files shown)


src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
```diff
@@ -554,7 +554,7 @@ private void VerifyPropertyFromImplicitImport(Project project, string propertyNa
 
         private SdkReference GetParsedSdk(ProjectImportElement element)
         {
-            PropertyInfo parsedSdkInfo = typeof(ProjectImportElement).GetProperty("ParsedSdkReference", BindingFlags.Instance | BindingFlags.NonPublic);
+            PropertyInfo parsedSdkInfo = typeof(ProjectImportElement).GetProperty("SdkReference", BindingFlags.Instance | BindingFlags.NonPublic);
```
rainersigwald, line 557: I like the rename, but it makes the more interesting change in this PR hard to review. Can you separate it into its own commit?

jeffkl: Done!

jeffkl:
I need to check out the scenario when you set the value to `null`, don't merge until I've replied about that please 😄

jeffkl:
I have updated the logic to allow setting Version and MinimumVersion to `null`.  Ready for final review!

PR #5327, opened 2020-05-01
Prevent invalidfilenamechars when globbing Fixes #5198

### LINKED ISSUE #5198: GetAllGlobs should filter out items which are clearly not file path
Add this block to the project, and GetAllGlobs will include it as a potential glob item, although it contains invalid path characters. It becomes a problem, when later code doesn't expect this kind of items. If msbuild can filter them out, it will reduce the chance to run into this problem.

  <ItemGroup>
    <PropsFileHeader Include="&lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;" />
  </ItemGroup>

Forgind:
This should fix #5198, but it would have perf implications, so we maybe shouldn't take it.

### DIFF
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -3744,11 +3744,6 @@ namespace Microsoft.Build.UnitTests.OM.Definition
             new[] {"ab", "de"},
             new[] {"bc", "b|c", "b", "c"}
             )]
-        [InlineData(
-            @"<A Include=`ab*;bc*;:de*`/>",
-            new[] { "ab", "bc" },
-            new[] { "de", ":de" }
-            )]
         public void GetAllGlobsShouldProduceGlobThatMatches(string itemContents, string[] stringsThatShouldMatch, string[] stringsThatShouldNotMatch)
         {
             var projectTemplate =
--- a/src/Build.OM.UnitTests/Definition/Project_Tests.cs
+++ b/src/Build.OM.UnitTests/Definition/Project_Tests.cs
@@ -3740,14 +3740,9 @@ namespace Microsoft.Build.UnitTests.OM.Definition
         new[] {"b", "c"}
         )]
         [InlineData(
-            @"<A Include=`ab*;bc|*;de*`/>",
+            @"<A Include=`ab*;b|c*;de*`/>",
             new[] {"ab", "de"},
-            new[] {"bc", "bc|"}
-            )]
-        [InlineData(
-            @"<A Include=`\ab*;bc*;de*`/>",
-            new[] { "bc", "de" },
-            new[] { "ab", "\\ab" }
+            new[] {"bc", "b|c", "b", "c"}
             )]
         [InlineData(
             @"<A Include=`ab*;bc*;:de*`/>",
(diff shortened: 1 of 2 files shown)


src/Build/Definition/Project.cs
```diff
@@ -2479,6 +2479,10 @@ private List<GlobResult> GetAllGlobs(List<ProjectItemElement> projectItemElement
 
                     if (!string.IsNullOrEmpty(itemElement.Include))
                     {
+                        if (NativeMethodsShared.IsWindows && FileUtilities.InvalidFileNameChars.Any(c => itemElement.Include.Contains(c)))
```
danmosemsft, line 2482: Can you avoid Linq, since it's perf sensitive here? This is less overhead-
` if (NativeMethodsShared.IsWindows && itemElement.Include.IndexOfAny( FileUtilities.InvalidFileNameChars) != -1`

Forgind: I probably should have sent this as a draft. I'm planning to use it as a starting point for "here's how we might implement this if we want to do this at all" tomorrow.

cdmihai:
I think this will break the API for Include and Remove elements containing a semicolon delimited list of items with one fragment as a literal invalid character, and other fragments as valid globs: `Include=*.cs;&lt`. Definitely needs tests for all the possible combinations.

src/Build/Definition/Project.cs
lifengl, line 2482: I think it is a wrong location to scan the invalid path. It will end up to scan lots of itemSpec which are not globbing candidate at all. It should be moved into BuildGlobResultFrom???Item method, after it picks up glob fragments. For two reasons:
1, there are far less globbing items than none globbing items, it is much more efficient to filter out other items first, instead of filtering invalid path first (which yield far less positive hit.)
2, when one include contains multiple items, maybe one of them is invalid, but this would exclude all of them. It is way too aggressive to be correct.

Forgind: You mean about [here](https://github.com/microsoft/msbuild/blob/master/src/Build/Definition/Project.cs#L2509)?

src/Build.OM.UnitTests/Definition/Project_Tests.cs
```diff
@@ -3739,6 +3739,21 @@ public void GetAllGlobsShouldFindAllExcludesAndRemoves()
         new[] {"aa", "bb", "cc"},
         new[] {"b", "c"}
         )]
+        [InlineData(
+            @"<A Include=`ab*;bc|*;de*`/>",
+            new[] {"ab", "de"},
+            new[] {"bc", "bc|"}
+            )]
+        [InlineData(
+            @"<A Include=`\ab*;bc*;de*`/>",
```
rainersigwald, line 3748: This one worries me: `\ab*` is a valid glob (all files in the root of the current drive that start with `ab`).

src/Build/Definition/Project.cs
```diff
@@ -57,7 +57,7 @@ public class Project : ILinkableObject
         /// <summary>
         /// * and ? are invalid file name characters, but they occur in globs as wild cards.
         /// </summary>
-        private static readonly char[] s_invalidGlobChars = FileUtilities.InvalidFileNameChars.Where(c => c != '*' && c != '?').ToArray();
+        private static readonly char[] s_invalidGlobChars = FileUtilities.InvalidFileNameChars.Where(c => c != '*' && c != '?' && c!= '/' && c != '\\').ToArray();
```
rainersigwald, line 60: From our discussion: `:`?

PR #5329, opened 2020-05-02
Fix Comment at target ResolveComReferences

Nirmal4G:
It's actually `InteropOutputPath`

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -2667,6 +2667,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         [IN]
         @(COMReference) - The list of COM references
         $(InteropOutputPath) - The output directory in which to generate wrapper assemblies
+                               When $(InteropOutputPath) is not set, then it defaults to $(IntermediateOutputPath).
 
         [OUT]
         @(ReferencePath) - Paths to referenced wrappers.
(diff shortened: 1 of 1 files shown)


Nirmal4G:
We really need `[skip ci]`, `[no ci]` or something similar for these kinds of PR!

Ref: microsoft/azure-pipelines-agent#2944

Nirmal4G:
@rainersigwald Split from #5238

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -2666,7 +2666,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
         [IN]
         @(COMReference) - The list of COM references
-        $(BaseIntermediateOutputPath) - The output directory in which to generate wrapper assemblies
+        $(InteropOutputPath) - The output directory in which to generate wrapper assemblies
```
rainersigwald, line 2669:
```suggestion
        $(InteropOutputPath) - The output directory in which to generate wrapper assemblies
        $(IntermediateOutputPath) - Default InteropOutputPath
```

?

Nirmal4G:
```suggestion
        $(InteropOutputPath) - The output directory in which to generate wrapper assemblies
                               When $(InteropOutputPath) is not set, then it defaults to $(IntermediateOutputPath).
```

How is this?

rainersigwald: Yeah, that looks good.

PR #5330, opened 2020-05-02
Always use 'Configuration' property

Nirmal4G:
If we set `ConfigurationName` to something else, the debug builds break!


Nirmal4G:
@rainersigwald Split from #5238

Nirmal4G:
> For instance, our own `Debug-MONO` configuration. You can then set `ConfigurationName` to `Debug` for that to get the default "like `Debug` behavior".

This should be in the docs. I was using it for that exact purpose but other way around. Thanks for clarifying.

PR #5339, opened 2020-05-07
Avoid errors on empty glob

Forgind:


src/Build/Evaluation/LazyItemEvaluator.IncludeOperation.cs
```diff
@@ -100,7 +100,7 @@ protected override ImmutableList<I> SelectItems(ImmutableList<ItemData>.Builder
                         string[] includeSplitFilesEscaped;
                         if (MSBuildEventSource.Log.IsEnabled())
                         {
-                            MSBuildEventSource.Log.ExpandGlobStart(_rootDirectory, glob, excludePatternsForGlobs.ToList().Aggregate((f, s) => f + ", " + s));
+                            MSBuildEventSource.Log.ExpandGlobStart(_rootDirectory, glob, string.Join(", ", excludePatternsForGlobs.ToList()));
```
ladipro, line 103: The ```ToList()``` looks redundant, ```excludePatternsForGlobs``` can be passed to ```Join()``` as is.

Forgind: Thanks for picking up on these! I'm surprised this didn't fail building MSBuild, but I guess I was lucky. Will fix.

Forgind: Ah, having looked at the second case, it can never be empty—if you don't specify targets, MSBuild will just build the default targets. I'll change it anyway.

PR #5342, opened 2020-05-08
Add String.IndexOf to well known functions

cdmihai:
This one gets hit a lot in a quickbuild repo I've been debugging through and the Expander exception was getting annoying :)

### DIFF
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -4259,7 +4259,7 @@ namespace Microsoft.Build.Evaluation
                 }
 
                 // Allow fully-qualified enum, e.g. "System.StringComparison.OrdinalIgnoreCase"
-                if (comparisonType.Contains("."))
+                if (comparisonType.Contains('.'))
                 {
                     comparisonType = comparisonType.Replace("System.StringComparison.", "").Replace("StringComparison.", "");
                 }
--- a/src/Build/Evaluation/Expander.cs
+++ b/src/Build/Evaluation/Expander.cs
@@ -3537,6 +3537,14 @@ namespace Microsoft.Build.Evaluation
                             return true;
                         }
                     }
+                    else if (string.Equals(_methodMethodName, nameof(string.IndexOf), StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (TryGetArg(args, out string arg0, out StringComparison arg1))
+                        {
+                            returnVal = text.IndexOf(arg0, arg1);
+                            return true;
+                        }
+                    }
                     else if (string.Equals(_methodMethodName, nameof(string.IndexOfAny), StringComparison.OrdinalIgnoreCase))
                     {
                         if (TryGetArg(args, out string arg0))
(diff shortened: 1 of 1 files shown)


src/Build/Evaluation/Expander.cs
```diff
@@ -4242,6 +4250,44 @@ private static bool TryGetArg(object[] args, out string arg0)
...
+            private static readonly Dictionary<string, StringComparison> StringComparisonNames =
+                new Dictionary<string, StringComparison>(StringComparer.Ordinal)
+                {
+                    {"System.StringComparison.OrdinalIgnoreCase", StringComparison.OrdinalIgnoreCase},
+                    {"System.StringComparison.Ordinal", StringComparison.Ordinal},
+                    {"System.StringComparison.CurrentCulture", StringComparison.CurrentCulture},
+                    {"System.StringComparison.CurrentCultureIgnoreCase", StringComparison.CurrentCultureIgnoreCase},
+                    {"System.StringComparison.InvariantCulture", StringComparison.InvariantCulture},
+                    {"System.StringComparison.InvariantCultureIgnoreCase", StringComparison.InvariantCultureIgnoreCase},
+                    {"OrdinalIgnoreCase", StringComparison.OrdinalIgnoreCase},
```
rainersigwald, line 4262: Is it legal to do, e.g. `StringComparison.OrdinalIgnoreCase`?

rainersigwald: Or see/abstract https://github.com/microsoft/msbuild/pull/5342/files#diff-dee96bf621e4ed93f0db318ca71d4d35R3572

src/Build/Evaluation/Expander.cs
```diff
@@ -4242,6 +4250,44 @@ private static bool TryGetArg(object[] args, out string arg0)
                 return arg0 != null;
             }
 
+            private static readonly Dictionary<string, StringComparison> StringComparisonNames =
```
ladipro, line 4253: A drop in an ocean but still, it's wasteful to create the dictionary on each class init when it's used only for evaluating one specific function. Can you please initialize it lazily? Also, shouldn't it be named something like ```s_stringComparisonNames``` to match the convention in other private static data in this class?

cdmihai: It happens once per process, probably not noticeable. But, I replaced the dictionary with another approach that was already being used in the same set of overloads.

src/Build/Evaluation/Expander.cs
```diff
@@ -4242,6 +4240,33 @@ private static bool TryGetArg(object[] args, out string arg0)
...
+                arg0 = args[0] as string;
+
+                if (arg0 == null || !(args[1] is string comparisonType))
+                {
+                    arg1 = default;
+                    return false;
+                }
+
+                // Allow fully-qualified enum, e.g. "System.StringComparison.OrdinalIgnoreCase"
+                if (comparisonType.Contains("."))
```
ladipro, line 4262: super-nit: ```'.'``` instead of ```"."```

cdmihai:
```suggestion
                if (comparisonType.Contains('.'))
```

PR #5343, opened 2020-05-08
Fix "Array dimensions exceeded supported range" exception

cdmihai:
Hitting this on a quickbuild repo with ~550 projects and ~19K edges (due to transitive edges).

The problem was using an intermediary list to aggregate intermediary results, which exploded exponentially in size due to duplicates. Fix is to use a hash set.
The failure can't really be tested (without stress tests) because it appears in an intermediary list which finally gets dumped into a set, so there's no observable trace of the problem at the end of graph construction.

This fix avoids the exception and takes graph construction down from 5 minutes to 30 seconds on that specific repo.

### DIFF
--- a/src/Build.UnitTests/Graph/GraphTestingUtilities.cs
+++ b/src/Build.UnitTests/Graph/GraphTestingUtilities.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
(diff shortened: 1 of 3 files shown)


src/Build.UnitTests/Graph/GraphTestingUtilities.cs
```diff
@@ -218,13 +219,18 @@ internal static string GetProjectPath(ProjectGraphNode node)
 
         internal static IEnumerable<ProjectGraphNode> ComputeClosure(ProjectGraphNode node)
         {
-            foreach (var reference in node.ProjectReferences)
-            {
-                yield return reference;
+            return ComputeClosureRecursive(node).ToHashSet();
```
cdmihai, line 222: This is just a test helper method, real code is down below.

PR #5346, opened 2020-05-10
Update 'WinFx' -> 'WinFX' casing

Nirmal4G:
As the title suggests, This PR renames **`WinFx`** to **`WinFX`** where possible.

I've separated out the white-space changes to a separate commit.
Only files that were touched with `WinFX` casing changes are updated with formatting changes.
If it's not possible to merge and/or if you want the line & space changes applied to the whole repo, please let me know, I'll separate these into a new PR.

**Related**: dotnet/wpf#2975

> The casing could also be updated for WPF (.NET Framework) targets and sources.


Nirmal4G:
@rainersigwald 

.NET Framework is Windows only and Case in-sensitive. Even though .NET Framework has high bar for compat, Why is this a breaking change?

rainersigwald:
I don't know why it changed from .NET 3.5 to .NET 4.something. I agree that it shouldn't have--but at this point changing it back is not worth much and might break someone.

I took a look around and this file is present in Mono installs on macOS (at least), so it can be present on a case-sensitive filesystem. Combined with the low impact of the change I don't think we should take this.

Thanks for trying to clean this up!

Nirmal4G:
@rainersigwald
> ... might break someone.
> I took a look around and this file is present in Mono installs on macOS (at least), so it can be present on a case-sensitive filesystem.

Does MSBuild (mono on MacOS) change `MSBuildToolsPath` change based on requested tools version? Even so, it's probably safe _as it is an internal import_. This might not even be a breaking change unless they are importing the targets file directly. If that's the case, then we can add a workaround to import both versions just like in F# targets.

Nirmal4G:
@rainersigwald
> ... might break someone.
> I took a look around and this file is present in Mono installs on macOS (at least), so it can be present on a case-sensitive filesystem.

Observe that the common targets use the correct case but the actual file on disk is wrong. That will lead to file not found error. I'm surprised why no one reported it.

I've add a workaround to import either versions just like in F# targets. This should fix the regression.

Nirmal4G:
@rainersigwald Can you reopen this PR?

> ... might break someone.
> I took a look around and this file is present in Mono installs on macOS (at least), so it can be present on a case-sensitive filesystem.

Observe that the common targets use the correct case but the actual file on disk is wrong. That will lead to file not found error. I'm surprised why no one reported it.

I've add a workaround to import either versions just like in F# targets. This should fix the regression.

Reference: https://github.com/microsoft/msbuild/commit/7ae67524ab2454269430377e7a1fb58f49685427

PR #5350, opened 2020-05-12
Make Remove-all-items O(1)

### LINKED ISSUE #2314: Update and Remove should optimize on identity matches

rainersigwald:
In the special case where a remove operation removes all items like

```xml
<Compile Remove="@(Compile)" />
```

We currently have poor behavior because we match ("everything matches")
and then remove items one at a time. Instead, we can just empty out the
list.

Fixes #2314. Encouraged to finally do this by @maartenba's blog post https://blog.jetbrains.com/dotnet/2020/05/11/story-csproj-large-solutions-memory-usage/.

### DIFF
--- a/src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
@@ -23,9 +23,14 @@ namespace Microsoft.Build.Evaluation
                 _matchOnMetadataOptions = builder.MatchOnMetadataOptions;
             }
 
-            // todo port the self referencing matching optimization (e.g. <I Remove="@(I)">) from Update to Remove as well. Ideally make one mechanism for both. https://github.com/Microsoft/msbuild/issues/2314
-            // todo Perf: do not match against the globs: https://github.com/Microsoft/msbuild/issues/2329
-            protected override ImmutableList<I> SelectItems(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
+            /// <summary>
+            /// Apply the Remove operation.
+            /// </summary>
+            /// <remarks>
+            /// This operation is mostly implemented in terms of the default <see cref="LazyItemOperation.ApplyImpl(ImmutableList{ItemData}.Builder, ImmutableHashSet{string})"/>.
+            /// This override exists to apply the removing-everything short-circuit.
+            /// </remarks>
+            protected override void ApplyImpl(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
             {
                 var matchOnMetadataValid = !_matchOnMetadata.IsEmpty && _itemSpec.Fragments.Count == 1
                     && _itemSpec.Fragments.First() is ItemSpec<ProjectProperty, ProjectItem>.ItemExpressionFragment;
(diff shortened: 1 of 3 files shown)


src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
```diff
@@ -23,9 +23,7 @@ public RemoveOperation(RemoveOperationBuilder builder, LazyItemEvaluator<P, I, M
                 _matchOnMetadataOptions = builder.MatchOnMetadataOptions;
             }
 
-            // todo port the self referencing matching optimization (e.g. <I Remove="@(I)">) from Update to Remove as well. Ideally make one mechanism for both. https://github.com/Microsoft/msbuild/issues/2314
-            // todo Perf: do not match against the globs: https://github.com/Microsoft/msbuild/issues/2329
-            protected override ImmutableList<I> SelectItems(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
+            protected override void ApplyImpl(ImmutableList<ItemData>.Builder listBuilder, ImmutableHashSet<string> globsToIgnore)
```
BenVillalobos, line 26: "For future generations" nit: ApplyImpl isn't exactly a self-documenting function name (like `ItemspecContainsASingleItemReference`), so I'd suggest writing up a short summary describing its purpose. Or adding that summary to the base function while you have the info in your head.

src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
```diff
@@ -34,6 +32,20 @@ protected override ImmutableList<I> SelectItems(ImmutableList<ItemData>.Builder
                     new BuildEventFileInfo(string.Empty),
                     "OM_MatchOnMetadataIsRestrictedToOnlyOneReferencedItem");
 
+                if (_matchOnMetadata.IsEmpty && UpdateOperation.ItemspecContainsASingleItemReference(_itemSpec, _itemElement.ItemType))
```
cdmihai, line 35:
```suggestion
                if (_matchOnMetadata.IsEmpty && ItemspecContainsASingleItemReference(_itemSpec, _itemElement.ItemType))
```

src/Build/Evaluation/LazyItemEvaluator.RemoveOperation.cs
```diff
@@ -34,6 +39,20 @@ protected override ImmutableList<I> SelectItems(ImmutableList<ItemData>.Builder
                     new BuildEventFileInfo(string.Empty),
                     "OM_MatchOnMetadataIsRestrictedToOnlyOneReferencedItem");
 
+                if (_matchOnMetadata.IsEmpty && ItemspecContainsASingleItemReference(_itemSpec, _itemElement.ItemType))
+                {
+                    // Perf optimization: If the Remove operation references itself (e.g. <I Remove="@(I)"/>)
+                    // then all items are removed and matching is not necessary
+                    listBuilder.RemoveAll((_) => true);
```
ladipro, line 46: nit:
```suggestion
                    listBuilder.Clear();
```

rainersigwald: 🤦‍♂️ I swear I used and/or looked for this before. Thanks!

Issue #4172, opened 2019-02-13
Many duplicate strings held by Microsoft.Build.Execution.ProjectMetadataInstance._escapedValue


drewnoakes:
Analysis via [StringTheory](https://github.com/drewnoakes/string-theory) of a VS2019 devenv.exe dump shows over 1.5 million duplicated strings held by `Microsoft.Build.Execution.ProjectMetadataInstance._escapedValue` (2.4% of the heap).

The loaded solution is generated, having 300 class library projects, each with 200 generated class files with a single class each. There are 300 P2P references.

- `_escapedValue` references 1,563,768 string objects (unique addresses), but only 4,326 unique string values (1.2% unique)
- The 1,563,768 strings occupy 43.3 MB bytes, 98.8% of which is wasted, meaning pooling here will reclaim 42.8 MB (2.4% of the ~1.8 GB managed heap)

//cc: @davkean @lifengl

Here are some samples of the duplicated strings. There appears to be several distinct groups of value, so I'll include a chunk from each.

| WastedBytes | Count | Length | String |
|------------:|------:|-------:|--------|
| 256,360 | 885 | 138 | C:\Users\drnoakes\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\System.Runtime.InteropServices.RuntimeInformation.dll |
| 249,288 | 885 | 134 | C:\Users\drnoakes\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\System.Security.Cryptography.X509Certificates.dll |
| 241,500 | 876 | 131 | C:\Users\drnoakes\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\System.Diagnostics.TextWriterTraceListener.dll |
| 239,490 | 888 | 128 | C:\Users\drnoakes\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\System.Runtime.CompilerServices.VisualC.dll |
| 238,680 | 885 | 128 | C:\Users\drnoakes\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\System.Runtime.Serialization.Formatters.dll |
| 238,680 | 885 | 128 | C:\Users\drnoakes\.nuget\packages\netstandard.library\2.0.3\build\netstandard2.0\ref\System.Runtime.Serialization.Primitives.dll |

| WastedBytes | Count | Length | String |
|------------:|------:|-------:|--------|
[...]

PR #5353, opened 2020-05-14
Rewrite OpportunisticIntern using weak GC handles

ladipro:
This PR contains a rewrite of the OpportunisticIntern logic, addressing scalability concerns with the original implementation.

The changes are structured as follows:
- The `IInternable` interface and its implementations are factored out of OpportunisticIntern.cs into a separate file src/Shared/IInternable.cs with no changes other than making them top-level types.
- A new data structure for maintaining weak GC handles to strings is introduced in a new file src/Shared/WeakStringCache.cs.
- `PrioritizedStringList` and friends is removed from OpportunisticIntern.cs as it's no longer used. The file now contains only common hard-coded strings, the logic to report statistics, and simple wrappers.

Motivation:
- The original implementation centered around MRU lists exposes many parameters to trade speed for memory as well as memory for memory. Make the lists too small and important strings will be falling out of the cache and won't be getting de-duplicated (#4172). Make the list too large and the linear search will take too long to find the string.
- The original implementation keeps strong references to the cached strings. This means that strings may be taking up memory even though they are not actually used anymore. By the cache that is supposed to save memory, ironically.

The new implementation:
- Does not expose any parameters. All strings are cached, regardless of length and prior state of the cache.
- Keeps only weak references to strings so it's not preventing them from being collected after they're not used anymore.

Performance:
- The memory win depends on the characteristics of the projects, it tends to be more significant the larger the project/solution is. For example, compared to the original implementation, `Microsoft.Build.Execution.ProjectInstance` instances use ~11% less memory in the Visual Studio process with a 300 project solution loaded.
- Time-wise, the new implementation is
  - Faster by 30% when interning hard-coded strings (result of the code becoming slightly simpler as the hard-coded string logic hasn't changed)
  - Faster when there is a cache hit, unless the hit is in the first few strings in the old MRU list (previously: linear search, depends on when the string is found; now: hash table lookup, fast regardless of size but requiring two passes over the string, one to hash it and one to compare it). The worst case where the hit is in the most recent item on the MRU list makes the new implementation about 2.5x slower (depends on the length of the string).
  - Typically faster when there is a cache miss (old implementation has to compare the string to all items of the same length). Depends on the size of the old MRU list and lengths of its strings.
- % of time spent in GC hasn't shown any measurable change. A common concern with GC handles is that they make the GC work harder.
- The memory overhead of `WeakStringCache` is the internal overhead of `Dictionary` + 16 bytes per string on 32-bit platforms / 28 bytes per string on 64-bit platforms.
  - 32-bit integer (hash code).
  - Pointer sized `GCHandle`.
  - Pointer sized overflow area for additional handles.
  - The runtime handle itself (pointer sized).

### DIFF
--- a/src/Build.UnitTests/OpportunisticIntern_Tests.cs
+++ b/src/Build.UnitTests/OpportunisticIntern_Tests.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Build.UnitTests
 {
     public class OpportunisticIntern_Tests
     {
-        private static bool IsInternable(OpportunisticIntern.IInternable internable)
+        private static bool IsInternable(IInternable internable)
         {
             string i1 = OpportunisticIntern.InternableToString(internable);
             string i2 = OpportunisticIntern.InternableToString(internable);
@@ -20,19 +20,19 @@ namespace Microsoft.Build.UnitTests
             return Object.ReferenceEquals(i1, i2);
         }
 
-        private static void AssertInternable(OpportunisticIntern.IInternable internable)
+        private static void AssertInternable(IInternable internable)
         {
             Assert.True(IsInternable(internable));
         }
 
         private static void AssertInternable(StringBuilder sb)
         {
-            AssertInternable(new OpportunisticIntern.StringBuilderInternTarget(sb));
+            AssertInternable(new StringBuilderInternTarget(sb));
         }
 
         private static string AssertInternable(char[] ch, int startIndex, int count)
         {
-            var target = new OpportunisticIntern.CharArrayInternTarget(ch, startIndex, count);
+            var target = new CharArrayInternTarget(ch, startIndex, count);
             AssertInternable(target);
             Assert.Equal(target.Length, count);
 
@@ -45,19 +45,19 @@ namespace Microsoft.Build.UnitTests
             AssertInternable(value.ToCharArray(), 0, value.ToCharArray().Length);
         }
 
-        private static void AssertNotInternable(OpportunisticIntern.IInternable internable)
+        private static void AssertNotInternable(IInternable internable)
         {
             Assert.False(IsInternable(internable));
         }
 
         private static void AssertNotInternable(StringBuilder sb)
         {
-            AssertNotInternable(new OpportunisticIntern.StringBuilderInternTarget(sb));
+            AssertNotInternable(new StringBuilderInternTarget(sb));
         }
 
         private static void AssertNotInternable(char[] ch)
         {
-            AssertNotInternable(new OpportunisticIntern.CharArrayInternTarget(ch, ch.Length));
+            AssertNotInternable(new CharArrayInternTarget(ch, ch.Length));
         }
 
         private static void AssertNotInternable(string value)
@@ -86,24 +86,6 @@ namespace Microsoft.Build.UnitTests
             var result = AssertInternable(new char[] { 'a', 't', 'r', 'u', 'e', 'x' }, 1, 4);
 
             Assert.Equal("true", result);
-        }
-
-        /// <summary>
-        /// Test a single know-to-intern tiny string to verify the mechanism.
-        /// </summary>
-        [Fact]
-        public void InternableTinyString()
-        {
-            AssertInternable("true");
-        }
-
-        /// <summary>
-        /// Test a single known-to-not-intern tiny string to verify the mechanism.
-        /// </summary>
-        [Fact]
-        public void NonInternableTinyString()
-        {
-            AssertNotInternable("1234");
         }
 
         /// <summary>
--- a/src/Build.UnitTests/WeakStringCache_Tests.cs
+++ b/src/Build.UnitTests/WeakStringCache_Tests.cs
@@ -1,0 +1,195 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text;
+using System.IO;
+using Microsoft.Build;
+using Microsoft.Build.Shared;
+using Xunit;
+using Shouldly;
+using System.Runtime.CompilerServices;
+using System.Diagnostics;
+
+namespace Microsoft.Build.UnitTests
+{
+    public class WeakStringCache_Tests : IDisposable
+    {
+        /// <summary>
+        /// The weak string cache under test.
+        /// </summary>
+        private WeakStringCache _cache = new WeakStringCache();
+
+        public void Dispose()
+        {
+            _cache.Dispose();
+        }
+
+        /// <summary>
+        /// Adds a string to the cache under test.
+        /// </summary>
+        /// <param name="strPart1">Part one of the string (split to prevent runtime interning and unintended GC roots).</param>
+        /// <param name="strPart2">Part two of the string (split to prevent runtime interning and unintended GC roots).</param>
+        /// <param name="callbackToRunWithTheStringAlive">Callback to be invoked after the string has been added but before the strong GC ref is released.</param>
+        /// <returns>The hash code of the string as calculated by WeakStringCache.</returns>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private int AddString(string strPart1, string strPart2, Action<string> callbackToRunWithTheStringAlive)
+        {
+            // Compose the string with SB so it doesn't get interned by the runtime.
+            string testString = new StringBuilder(strPart1).Append(strPart2).ToString();
+            StringInternTarget testStringTarget = new StringInternTarget(testString);
+
+            int hashCode = WeakStringCache.GetInternableHashCode(testStringTarget);
+
+            string cachedString = _cache.GetOrCreateEntry(testStringTarget, out bool cacheHit);
+            cacheHit.ShouldBeFalse();
+            cachedString.ShouldBeSameAs(testString);
+
+            callbackToRunWithTheStringAlive(cachedString);
+
+            // Verify that the string is really in the cache and the cache returns the interned instance.
+            string testStringCopy = new StringBuilder(strPart1).Append(strPart2).ToString();
+            cachedString = _cache.GetOrCreateEntry(new StringInternTarget(testStringCopy), out cacheHit);
+            cacheHit.ShouldBeTrue();
+            cachedString.ShouldBeSameAs(testString);
+
+            // Trigger full GC and verify that nothing has changed since we're still keeping testString alive.
+            GC.Collect();
+
+            callbackToRunWithTheStringAlive(cachedString);
+
+            cachedString = _cache.GetOrCreateEntry(new StringInternTarget(testStringCopy), out cacheHit);
+            cacheHit.ShouldBeTrue();
+            cachedString.ShouldBeSameAs(testString);
+
+            return hashCode;
+        }
+
+        /// <summary>
+        /// Adds two strings that are known to have a hash code collision to the cache under test.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void AddTwoStringsWithHashCollision()
+        {
+            string firstCachedString = null;
+
+            // Add strings to the cache using a non-inlinable method to make sure they're not reachable from a GC root.
+            int hashCode1 = AddString("Random string ", "1", (string cachedString) =>
+            {
+                _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+                {
+                    UsedBucketCount = 1,
+                    UnusedBucketCount = 0,
+                    LiveStringCount = 1,
+                    CollectedStringCount = 0,
+                    HashCollisionCount = 0
+                });
+                firstCachedString = cachedString;
+            });
+
+            int hashCode2 = AddString("Random string ", "14858396876", (string cachedString) =>
+            {
+                _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+                {
+                    UsedBucketCount = 1,
+                    UnusedBucketCount = 0,
+                    LiveStringCount = 2,
+                    CollectedStringCount = 0,
+                    HashCollisionCount = 1
+                });
+
+                string firstCachedStringFromCache = _cache.GetOrCreateEntry(new StringInternTarget(firstCachedString), out bool cacheHit);
+                cacheHit.ShouldBeTrue();
+                firstCachedStringFromCache.ShouldBeSameAs(firstCachedString);
+            });
+
+            // The two string have been carefully chosen to have the same hash code.
+            hashCode2.ShouldBe(hashCode1);
+        }
+
+        /// <summary>
+        /// Simple test case to verify that:
+        /// 1. A string added to the cache stays in the cache as long as it's alive.
+        /// 2. The string is no longer retrievable after all strong GC refs are gone.
+        /// 3. The cache completely removes the bucket with the string after calling Scavenge on it.
+        /// </summary>
+        [Fact]
+        public void RetainsStringUntilCollected()
+        {
+            // Add a string to the cache using a non-inlinable method to make sure it's not reachable from a GC root.
+            AddString("Random string ", "test", (string cachedString) =>
+            {
+                _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+                {
+                    UsedBucketCount = 1,
+                    UnusedBucketCount = 0,
+                    LiveStringCount = 1,
+                    CollectedStringCount = 0,
+                    HashCollisionCount = 0
+                });
+            });
+
+            // Trigger full GC.
+            GC.Collect();
+
+            // The bucket is still in the cache but it's unused now as the string has been collected.
+            _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+            {
+                UsedBucketCount = 0,
+                UnusedBucketCount = 1,
+                LiveStringCount = 0,
+                CollectedStringCount = 1,
+                HashCollisionCount = 0
+            });
+
+            // Ask the cache to get rid of unused buckets.
+            _cache.Scavenge();
+
+            // The cache should be empty now.
+            _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+            {
+                UsedBucketCount = 0,
+                UnusedBucketCount = 0,
+                LiveStringCount = 0,
+                CollectedStringCount = 0,
+                HashCollisionCount = 0
+            });
+        }
+
+        /// <summary>
+        /// Same as RetainsStringUntilCollected but with two strings with the same hash code. Verifies that the bucket overflow area
+        /// works correctly.
+        /// </summary>
+        [Fact]
+        public void RetainsStringsWithHashCollisions()
+        {
+            AddTwoStringsWithHashCollision();
+
+            // Trigger full GC.
+            GC.Collect();
+
+            // The bucket is still in the cache but it's unused now as the strings have been collected.
+            _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+            {
+                UsedBucketCount = 0,
+                UnusedBucketCount = 1,
+                LiveStringCount = 0,
+                CollectedStringCount = 2,
+                HashCollisionCount = 1
+            });
+
+            // Ask the cache to get rid of unused buckets.
+            _cache.Scavenge();
+
+            // The cache should be empty now.
+            _cache.GetDebugInfo().ShouldBe(new WeakStringCache.DebugInfo()
+            {
+                UsedBucketCount = 0,
+                UnusedBucketCount = 0,
+                LiveStringCount = 0,
+                CollectedStringCount = 0,
+                HashCollisionCount = 0
+            });
+        }
+    }
+}
--- a/src/Build/Microsoft.Build.csproj
+++ b/src/Build/Microsoft.Build.csproj
@@ -351,6 +351,8 @@
     <Compile Include="..\Shared\CollectionHelpers.cs" />
     <Compile Include="Collections\ConvertingEnumerable.cs" />
     <Compile Include="Collections\CopyOnReadEnumerable.cs" />
+    <Compile Include="..\Shared\IInternable.cs" />
+    <Compile Include="..\Shared\WeakStringCache.cs" />
     <Compile Include="..\Shared\OpportunisticIntern.cs" />
     <Compile Include="..\Shared\CopyOnWriteDictionary.cs">
       <Link>Collections\CopyOnWriteDictionary.cs</Link>
--- a/src/Shared/IInternable.cs
+++ b/src/Shared/IInternable.cs
@@ -1,0 +1,346 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text;
+using Microsoft.Build.Shared;
+
+namespace Microsoft.Build
+{
+    #region IInternable
+    /// <summary>
+    /// Define the methods needed to intern something.
+    /// </summary>
+    internal interface IInternable
+    {
+        /// <summary>
+        /// The length of the target.
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Indexer into the target. Presumed to be fast.
+        /// </summary>
+        char this[int index] { get; }
+
+        /// <summary>
+        /// Convert target to string. Presumed to be slow (and will be called just once).
+        /// </summary>
+        string ExpensiveConvertToString();
+
+        /// <summary>
+        /// Compare target to string. Assumes string is of equal or smaller length than target.
+        /// </summary>
+        bool StartsWithStringByOrdinalComparison(string other);
+
+        /// <summary>
+        /// Reference compare target to string. If target is non-string this should return false.
+        /// </summary>
+        bool ReferenceEquals(string other);
+    }
+    #endregion
+
+
+    #region IInternable Implementations
+    /// <summary>
+    /// A wrapper over StringBuilder.
+    /// </summary>
+    internal struct StringBuilderInternTarget : IInternable
+    {
+        /// <summary>
+        /// The held StringBuilder
+        /// </summary>
+        private readonly StringBuilder _target;
+
+        /// <summary>
+        /// Pointless comment about constructor.
+        /// </summary>
+        internal StringBuilderInternTarget(StringBuilder target)
+        {
+            _target = target;
+        }
+
+        /// <summary>
+        /// The length of the target.
+        /// </summary>
+        public int Length => _target.Length;
+
+        /// <summary>
+        /// Indexer into the target. Presumed to be fast.
+        /// </summary>
+        public char this[int index] => _target[index];
+
+        /// <summary>
+        /// Never reference equals to string.
+        /// </summary>
+        public bool ReferenceEquals(string other) => false;
+
+        /// <summary>
+        /// Convert target to string. Presumed to be slow (and will be called just once).
+        /// </summary>
+        public string ExpensiveConvertToString()
+        {
+            // PERF NOTE: This will be an allocation hot-spot because the StringBuilder is finally determined to
+            // not be internable. There is still only one conversion of StringBuilder into string it has just
+            // moved into this single spot.
+            return _target.ToString();
+        }
+
+        /// <summary>
+        /// Compare target to string. Assumes string is of equal or smaller length than target.
+        /// </summary>
+        public bool StartsWithStringByOrdinalComparison(string other)
+        {
+#if DEBUG
+            ErrorUtilities.VerifyThrow(other.Length <= _target.Length, "should be at most as long as target");
+#endif
+            int length = other.Length;
+
+            // Backwards because the end of the string is (by observation of Australian Government build) more likely to be different earlier in the loop.
+            // For example, C:\project1, C:\project2
+            for (int i = length - 1; i >= 0; --i)
+            {
+                if (_target[i] != other[i])
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Don't use this function. Use ExpensiveConvertToString
+        /// </summary>
+        public override string ToString() => throw new InvalidOperationException();
+    }
+
+    /// <summary>
+    /// A wrapper over char[].
+    /// </summary>
+    internal struct CharArrayInternTarget : IInternable
+    {
+        /// <summary>
+        /// Start index for the string
+        /// </summary>
+        private readonly int _startIndex;
+
+        /// <summary>
+        /// The held array
+        /// </summary>
+        private readonly char[] _target;
+
+        /// <summary>
+        /// Pointless comment about constructor.
+        /// </summary>
+        internal CharArrayInternTarget(char[] target, int count)
+            : this(target, 0, count)
+        {
+        }
+
+        /// <summary>
+        /// Pointless comment about constructor.
+        /// </summary>
+        internal CharArrayInternTarget(char[] target, int startIndex, int count)
+        {
+#if DEBUG
+            if (startIndex + count > target.Length)
+            {
+                ErrorUtilities.ThrowInternalError("wrong length");
+            }
+#endif
+            _target = target;
+            _startIndex = startIndex;
+            Length = count;
+        }
+
+        /// <summary>
+        /// The length of the target.
+        /// </summary>
+        public int Length { get; }
+
+        /// <summary>
+        /// Indexer into the target. Presumed to be fast.
+        /// </summary>
+        public char this[int index]
+        {
+            get
+            {
+                if (index > _startIndex + Length - 1 || index < 0)
+                {
+                    ErrorUtilities.ThrowInternalError("past end");
+                }
+
+                return _target[index + _startIndex];
+            }
+        }
+
+        /// <summary>
+        /// Convert target to string. Presumed to be slow (and will be called just once).
+        /// </summary>
+        public bool ReferenceEquals(string other)
+        {
+            return false;
+        }
+
+        /// <summary>
+        /// Convert target to string. Presumed to be slow (and will be called just once).
+        /// </summary>
+        public string ExpensiveConvertToString()
+        {
+            // PERF NOTE: This will be an allocation hot-spot because the char[] is finally determined to
+            // not be internable. There is still only one conversion of char[] into string it has just
+            // moved into this single spot.
+            return new string(_target, _startIndex, Length);
+        }
+
+        /// <summary>
+        /// Compare target to string. Assumes string is of equal or smaller length than target.
+        /// </summary>
+        public bool StartsWithStringByOrdinalComparison(string other)
+        {
+#if DEBUG
+            ErrorUtilities.VerifyThrow(other.Length <= Length, "should be at most as long as target");
+#endif
+            // Backwards because the end of the string is (by observation of Australian Government build) more likely to be different earlier in the loop.
+            // For example, C:\project1, C:\project2
+            for (int i = other.Length - 1; i >= 0; --i)
+            {
+                if (_target[i + _startIndex] != other[i])
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Don't use this function. Use ExpensiveConvertToString
+        /// </summary>
+        public override string ToString()
+        {
+            throw new InvalidOperationException();
+        }
+    }
+
+    /// <summary>
+    /// Wrapper over a string.
+    /// </summary>
+    internal struct StringInternTarget : IInternable
+    {
+        /// <summary>
+        /// Stores the wrapped string.
+        /// </summary>
+        private readonly string _target;
+
+        /// <summary>
+        /// Constructor of the class
+        /// </summary>
+        /// <param name="target">The string to wrap</param>
+        internal StringInternTarget(string target)
+        {
+            ErrorUtilities.VerifyThrowArgumentLength(target, nameof(target));
+            _target = target;
+        }
+
+        /// <summary>
+        /// Gets the length of the target string.
+        /// </summary>
+        public int Length => _target.Length;
+
+        /// <summary>
+        /// Gets the n character in the target string.
+        /// </summary>
+        /// <param name="index">Index of the character to gather.</param>
+        /// <returns>The character in the position marked by index.</returns>
+        public char this[int index] => _target[index];
+
+        /// <summary>
+        /// Returns the target which is already a string.
+        /// </summary>
+        /// <returns>The target string.</returns>
+        public string ExpensiveConvertToString() => _target;
+
+        /// <summary>
+        /// Compare target to string. Assumes string is of equal or smaller length than target.
+        /// </summary>
+        /// <param name="other">The string to compare with the target.</param>
+        /// <returns>True if target starts with <paramref name="other"/>, false otherwise.</returns>
+        public bool StartsWithStringByOrdinalComparison(string other) => _target.StartsWith(other, StringComparison.Ordinal);
+
+        /// <summary>
+        /// Verifies if the reference of the target string is the same of the given string.
+        /// </summary>
+        /// <param name="other">The string reference to compare to.</param>
+        /// <returns>True if both references are equal, false otherwise.</returns>
+        public bool ReferenceEquals(string other) => ReferenceEquals(_target, other);
+    }
+
+    /// <summary>
+    /// Wrapper over a substring of a string.
+    /// </summary>
+    internal struct SubstringInternTarget : IInternable
+    {
+        /// <summary>
+        /// Stores the wrapped string.
+        /// </summary>
+        private readonly string _target;
+
+        /// <summary>
+        /// Start index of the substring within the wrapped string.
+        /// </summary>
+        private readonly int _startIndex;
+
+        /// <summary>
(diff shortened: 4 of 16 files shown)


src/Shared/IInternable.cs
```diff
@@ -0,0 +1,346 @@
...
+
+        /// <summary>
+        /// Compare target to string. Assumes string is of equal or smaller length than target.
+        /// </summary>
+        public bool StartsWithStringByOrdinalComparison(string other)
+        {
+#if DEBUG
+            ErrorUtilities.VerifyThrow(other.Length <= Length, "should be at most as long as target");
+#endif
+            // Backwards because the end of the string is (by observation of Australian Government build) more likely to be different earlier in the loop.
```
danmosemsft, line 205: I don't think there is any need to refer to the Aussie Govt benchmark anymore. Anyway, it's fairly intuitive

src/Build.UnitTests/WeakStringCache_Tests.cs
```diff
@@ -0,0 +1,195 @@
...
+        /// <summary>
+        /// Adds two strings that are known to have a hash code collision to the cache under test.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void AddTwoStringsWithHashCollision()
+        {
+            string firstCachedString = null;
+
+            // Add strings to the cache using a non-inlinable method to make sure they're not reachable from a GC root.
+            int hashCode1 = AddString("Random string ", "1", (string cachedString) =>
```
danmosemsft, line 77: given it's a list it might be good to add more than 1 entry to it.

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,346 @@
...
+            _target = target;
+        }
+
+        /// <summary>
+        /// The length of the target.
+        /// </summary>
+        public int Length => _target.Length;
+
+        /// <summary>
+        /// Indexer into the target. Presumed to be fast.
```
danmosemsft, line 69: Warning--  indexing into a StringBuilder is NOT necessarily fast -- and looping through it (as one might through an array or string) can be n^2. (It depends on whether the StringBuilder is fragmented, which depends on how it was created and modified.) This is in the docs now, IIRC because MSBuild was doing this somewhere else.

Can indexing into a StringBuilder be avoided entirely?

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,346 @@
...
+        public bool ReferenceEquals(string other) => ReferenceEquals(_target, other);
+    }
+
+    /// <summary>
+    /// Wrapper over a substring of a string.
+    /// </summary>
+    internal struct SubstringInternTarget : IInternable
+    {
+        /// <summary>
+        /// Stores the wrapped string.
```
danmosemsft, line 286: This seems like maybe it should be rewritten in terms of `ReadOnlySpan<char>`'s. I see that MSBuild already has taken a dependency on System.Memory so it is available.

src/Shared/IInternable.cs
danmosemsft: Here's context - https://github.com/Microsoft/msbuild/issues/1593

BTW, to compare to string -- on .NET Core there is a `StringBuilder.Equals(ReadOnlySpan<Char>)`. On .NET FRamework, you have to create a StringBuilder from the string, or vice versa, and compare like with like.

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,346 @@
...
+        public int Length { get; }
+
+        /// <summary>
+        /// Indexer into the target. Presumed to be fast.
+        /// </summary>
+        public char this[int index]
+        {
+            get
+            {
+                if (index > _startIndex + Length - 1 || index < 0)
```
danmosemsft, line 169: Suggest removing this check. It's not adding any value and just makes the codepath longer.

danmosemsft: and SubstringInternTarget  doesn't have it

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,346 @@
...
+    /// </summary>
+    internal interface IInternable
+    {
+        /// <summary>
+        /// The length of the target.
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Indexer into the target. Presumed to be fast.
```
danmosemsft, line 22: Since as pointed out below it's not necessarily fast, it might be best to remove this from the interface and replace with `bool EqualsOrdinal(IInternable)` and on the concrete types have specialized overloads for the other concrete types.

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,346 @@
...
+#if DEBUG
+            ErrorUtilities.VerifyThrow(other.Length <= _target.Length, "should be at most as long as target");
+#endif
+            int length = other.Length;
+
+            // Backwards because the end of the string is (by observation of Australian Government build) more likely to be different earlier in the loop.
+            // For example, C:\project1, C:\project2
+            for (int i = length - 1; i >= 0; --i)
+            {
+                if (_target[i] != other[i])
```
danmosemsft, line 103: This is indexing into a string builder.

src/Shared/OpportunisticIntern.cs
```diff
@@ -21,156 +17,83 @@ namespace Microsoft.Build
...
-            s_whatIfInfinite = new BucketedPrioritizedStringList(/*gatherStatistics*/ true, int.MaxValue, int.MaxValue, int.MaxValue, s_smallMruThreshold, s_largeMruThreshold, s_hugeMruThreshold, s_ginormousThreshold, s_useSimpleConcurrency);
-            s_whatIfDoubled = new BucketedPrioritizedStringList(/*gatherStatistics*/ true, s_smallMruSize * 2, s_largeMruSize * 2, s_hugeMruSize * 2, s_smallMruThreshold, s_largeMruThreshold, s_hugeMruThreshold, s_ginormousThreshold, s_useSimpleConcurrency);
-            s_whatIfHalved = new BucketedPrioritizedStringList(/*gatherStatistics*/ true, s_smallMruSize / 2, s_largeMruSize / 2, s_hugeMruSize / 2, s_smallMruThreshold, s_largeMruThreshold, s_hugeMruThreshold, s_ginormousThreshold, s_useSimpleConcurrency);
-            s_whatIfZero = new BucketedPrioritizedStringList(/*gatherStatistics*/ true, 0, 0, 0, s_smallMruThreshold, s_largeMruThreshold, s_hugeMruThreshold, s_ginormousThreshold, s_useSimpleConcurrency);
+            s_missedHardcodedStrings = new Dictionary<string, int>();
+            s_gatherStatistics = true;
```
danmosemsft, line 88: Do you guys ever need to gather stats on customer machines? Assuming not -- and this is just for tuning the interning, which you've made need much less tuning than before -- I suggest to put all the stats gathering code behind #if DEBUG and stop shipping it to customers.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// </summary>
+        public void Scavenge()
+        {
+            List<int> keysToRemove = null;
+            foreach (KeyValuePair<int, StringBucket> entry in _stringsByHashCode)
+            {
+                if (!entry.Value.IsUsed)
+                {
+                    entry.Value.Free();
+                    keysToRemove ??= new List<int>();
```
danmosemsft, line 296: Aside: in .NET Core you can safely remove keys from a `Dictionary<K,V>` while enumerating it. That' doesn't help you today of course.

src/Shared/IInternable.cs
danmosemsft: Hm, I see you need a hashcode as well. Maybe string builders should hash only their first N chars?

danmosemsft:
Glad this is getting freshened up! Sorry to see us lose references to `ginormous` though.

Forgind:
How many strings do you expect to see? If it's a lot, it might be reasonable to replace the current hash with a set of n distinct hashes where n depends on the number of strings, and a match is defined by matching all the hashes.

src/Shared/IInternable.cs
ladipro: Removed, thanks!

src/Build.UnitTests/WeakStringCache_Tests.cs
ladipro: Totally. I've generalized the helper method and now testing with 3 strings: 1 base + 2 in the overflow list.

src/Shared/IInternable.cs
ladipro: Thank you. I'd like to keep the PR focused only on the interner itself. This code is just being copied verbatim from `OpportunisticIntern.cs`. I will address this separately later if you don't mind.

src/Shared/IInternable.cs
ladipro: Yes, having our `StringBuilder` wrapper implements the interface throws a wrench into what would otherwise be a straightforward refactor. Will be addressed in a separate PR.

src/Shared/IInternable.cs
ladipro: Will be addressed in a separate PR.

src/Shared/OpportunisticIntern.cs
ladipro: It's a valid point but I think I'll keep it in for now.

In the last couple of commits I brought the old implementation back as we agreed that it would be safer to provide an "escape hatch" in case the new one doesn't perform well, at least for a few releases. I've actually tried to put it all under DEBUG and it turned out that it would be either messy or require even more refactoring of code which we want to kill soon. Also, one of the numbers included in the statistics is "Time Spent Interning" which is only useful in release builds (although it's debatable whether it's useful at all given its implementation).

src/Shared/IInternable.cs
ladipro: Great point, removed!

src/Shared/WeakStringCache.cs
ladipro: Nice, I didn't know! We could take advantage of that when building for Core but aside from the added complexity with `#ifdef`s I'm afraid that we currently target 2.1 and according to the documentation this is a 3.0+ feature. Maybe moving to 3.0 is OK at this point but it's going to be a separate change for sure.

ladipro:
> How many strings do you expect to see? If it's a lot, it might be reasonable to replace the current hash with a set of n distinct hashes where n depends on the number of strings, and a match is defined by matching all the hashes.

The goal is for the implementation to work reasonably well regardless of the number of strings. It can be as high as tens of thousands. @Forgind are you suggesting to use multiple hash tables in order to reduce the number of hash collisions? In my testing I haven't seen more than a couple of collisions even in really large workloads. This appears to be supported by the theory as well. Assuming the hash function produces codes uniformly covering the 32-bit integer space (and my calculation is correct :p), the expected number of collisions for 100k strings is just over 1. Meaning that on average all but two strings are expected to have distinct hash codes. Of course the reality is going to be worse but it shouldn't be orders of magnitude worse.

ladipro:
@danmosemsft thank you for reviewing the changes. I have added a few more commits. Most importantly I had to bring the old implementation back as the team has agreed to have an "escape hatch" in case something goes wrong. I've been struggling producing readable and "follows good practices" code while avoiding extensive refactoring, especially as we expect the old implementation to be deleted some time soon. Some of the refactoring I just had to do (e.g. made `OpportunisticIntern` into a singleton for testability reasons) and the diff is admittedly ugly.

@all I'm wondering if I should maybe split the refactoring changes into a separate PR. Let me know if you'd prefer that.

danmosemsft:
> the expected number of collisions for 100k strings is just over 1

If that's the case could you just drop the overflow list? If there are two strings with the same hash you just don't intern both.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,355 @@
...
```
Forgind, line 296: I'm guessing you put this in the loop so you wouldn't allocate unnecessarily if none of the entries are used. What if you walked through this with a while loop, just checking for not unused entries and stopped on the first one, then had a second while loop that continued to the end doing this logic after allocating? This is only one line, but it's more complex than normal.

src/Shared/OpportunisticIntern.cs
```diff
@@ -21,116 +21,138 @@ namespace Microsoft.Build
...
+    /// The legacy implementation uses heuristics to decide whether it will be efficient to intern a string or not. There is no
     /// guarantee that a string will intern.
     ///
     /// The thresholds and sizes were determined by experimentation to give the best number of bytes saved
     /// at reasonable elapsed time cost.
+    ///
+    /// The new implementation interns all strings but maintains only weak references so it doesn't keep the strings alive.
     /// </summary>
-    internal static class OpportunisticIntern
+    internal sealed class OpportunisticIntern
```
Forgind, line 35: Why sealed?

src/Shared/Traits.cs
```diff
@@ -43,7 +43,12 @@ public Traits()
         public readonly bool CacheFileExistence = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MsBuildCacheFileExistence"));
 
         /// <summary>
-        /// Eliminate locking in OpportunisticIntern at the expense of memory
+        /// Use the legacy string interning implementation based on MRU lists.
+        /// </summary>
+        public readonly bool UseLegacyStringInterner = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MsBuildUseLegacyStringInterner"));
```
Forgind, line 48: tiny nit:
```suggestion
        public readonly bool UseLegacyStringInterner = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MSBuildUseLegacyStringInterner"));
```

src/Shared/OpportunisticIntern.cs
```diff
@@ -21,116 +21,138 @@ namespace Microsoft.Build
...
+            void ReportStatistics(string heading);
+        }
+
+        /// <summary>
+        /// The singleton instance of OpportunisticIntern.
+        /// </summary>
+        private static OpportunisticIntern _instance = new OpportunisticIntern();
+        internal static OpportunisticIntern Instance => _instance;
+
+        private readonly bool _useLegacyInterner = Traits.Instance.UseLegacyStringInterner;
```
Forgind, line 61: Do you have to store this here rather than just using Traits.Instance.UseLegacyStringInterner a few times?

src/Shared/WeakStringCache.cs
danmosemsft: @rainersigwald do you still bits out of this branch destined to run on .NET Core 2.1?

src/Shared/IInternable.cs
danmosemsft: Might be worth opening an issue if there isn't one already.

src/Shared/IInternable.cs
danmosemsft: > To make the hash a bit better, maybe first n/3 characters, last n/3 characters, and middle n/3 characters?

Whatever you do should be based on the perf characteristics of StringBuilders. It's an implementation detail, but unlikely to change: if the SB is fragmented, the SB object you hold contains the buffer of the last fragment, and a pointer to the next previous. Indexing into the SB requires walking along the chain - there is no "index" and there is no "cached last index". So perhaps I should have said the last N characters; jumping around in the SB is what is costly. It's essentially a character buffer designed to be cheaply modified at the very end, then turned into a contiguous buffer or string and thrown away.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// </summary>
+        /// <param name="internable">The internable describing the string we're looking for.</param>
+        /// <returns>A string matching the given internable.</returns>
+        public string GetOrCreateEntry<T>(T internable, out bool cacheHit) where T : IInternable
+        {
+            int hashCode = GetInternableHashCode(internable);
+
+            StringBucket bucket;
+            string result;
+            lock (_stringsByHashCode)
```
drewnoakes, line 229: Would an upgradeable read/write lock be a good fit here?

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,355 @@
...
+            }
+            return hashCode;
+        }
+
+        /// <summary>
+        /// Iterates over the cache and removes unused buckets, i.e. buckets that don't reference live strings.
+        /// This is expensive so try to call such that the cost is amortized to O(1) per GetOrCreateEntry() invocation.
+        /// Assumes exclusive access to the dictionary, i.e. the lock is taken.
+        /// </summary>
+        public void Scavenge()
```
drewnoakes, line 288: Should this be public? How can an external caller guarantee the lock on `_stringsByHashCode` is held?

src/Build.UnitTests/OpportunisticIntern_Tests.cs
```diff
@@ -173,4 +168,61 @@ public void DoubleDoubleQuotes()
...
+    {
+        public OpportunisticInternLegacy_Tests(ITestOutputHelper testOutput)
+            : base(testOutput)
+        {
+            _env.SetEnvironmentVariable("MsBuildUseLegacyStringInterner", "1");
+            OpportunisticIntern.ResetForTests();
+        }
+
+        /// <summary>
+        /// Test a single know-to-intern tiny string to verify the mechanism.
```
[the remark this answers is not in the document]
ladipro: Fixed by deleting this method as we're already testing the same string in `KnownInternableTinyStrings` in the base class. Tweaked the comment on the other method which stays. Thanks!

src/Shared/OpportunisticIntern.cs
ladipro: If `Traits` explicitly guaranteed that the traits never change during execution it would make sense. I think I'd rather leave it like this (same as the pre-existing `_useSimpleConcurrency`) for consistency and a little bit of extra defensive safety.

src/Shared/Traits.cs
ladipro: Done, thanks.

src/Shared/OpportunisticIntern.cs
ladipro: If a class is not intended to have subclasses it is a good practice to seal it by default. Enables certain codegen optimizations - e.g. calling a virtual method can be faster because the runtime knows that the method cannot be overriden. It likely makes zero difference here but doing it anyway as a good practice.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,355 @@
...
+            /// </summary>
+            public bool IsUsed
+            {
+                get
+                {
+                    if (WeakHandle.Target != null)
+                    {
+                        return true;
+                    }
+                    if (WeakHandleOverflow != null)
```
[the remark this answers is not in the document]
ladipro: I have deleted this code together with everything `WeakHandleOverflow`-related on Dan's recommendation.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,355 @@
...
+                {
+                    int liveCount = 0;
+                    if (WeakHandle.Target != null)
+                    {
+                        liveCount++;
+                    }
+                    if (WeakHandleOverflow != null)
+                    {
+                        for (int i = 0; i < WeakHandleOverflow.Count; i++)
+                        {
```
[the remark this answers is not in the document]
ladipro: `WeakHandleOverflow` is gone.

src/Build/Microsoft.Build.csproj
```diff
@@ -351,6 +351,8 @@
     <Compile Include="..\Shared\CollectionHelpers.cs" />
     <Compile Include="Collections\ConvertingEnumerable.cs" />
     <Compile Include="Collections\CopyOnReadEnumerable.cs" />
+    <Compile Include="..\Shared\IInternable.cs" />
+    <Compile Include="..\Shared\WeakStringCache.cs" />
```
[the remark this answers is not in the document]
ladipro: I was confused by seeing `OpportunisticIntern.cs` referenced in so many project files and then I remembered your work of eliminating the duplication and hitting obscure issues where customers have taken dependency on certain way MSBuild assemblies reference each other. I agree that it's probably not worth it to try to solve it now.

src/Shared/WeakStringCache.cs
ladipro: This pattern has been shamelessly stolen from `WeakValueDictionary`. It's just a simple `if` and while I can measure a difference if the loop is doing nothing but iterating over the dictionary, in reality it's going to be dominated by the collection and handle operations. I lean towards not complicating the code in this case. Interesting idea, though!

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// <summary>
+        /// Don't use this function. Use ExpensiveConvertToString
+        /// </summary>
+        public override string ToString() => throw new InvalidOperationException();
+    }
+
+    /// <summary>
+    /// A wrapper over char[].
+    /// </summary>
+    internal struct CharArrayInternTarget : IInternable
```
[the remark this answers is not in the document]
ladipro: Thanks, done. (You made the same comment on another PR and I feel stupid for not remembering it.)

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,341 @@
...
+        public override string ToString()
+        {
+            throw new InvalidOperationException();
+        }
+    }
+
+    /// <summary>
+    /// Wrapper over a string.
+    /// </summary>
+    internal struct StringInternTarget : IInternable
```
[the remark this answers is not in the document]
ladipro: Done, thanks!

src/Shared/IInternable.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// </summary>
+        /// <param name="other">The string reference to compare to.</param>
+        /// <returns>True if both references are equal, false otherwise.</returns>
+        public bool ReferenceEquals(string other) => ReferenceEquals(_target, other);
+    }
+
+    /// <summary>
+    /// Wrapper over a substring of a string.
+    /// </summary>
+    internal struct SubstringInternTarget : IInternable
```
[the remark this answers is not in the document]
ladipro: Done, thanks!

src/Shared/OpportunisticIntern.cs
```diff
@@ -21,116 +21,138 @@ namespace Microsoft.Build
...
         /// </summary>
-        private static readonly int s_ginormousThreshold = AssignViaEnvironment("MSBUILDGINORMOUSINTERNTHRESHOLD", 8000);
+        private readonly int _ginormousThreshold;
 
         /// <summary>
-        /// Manages the separate MRU lists.
+        /// The interner implementation in use.
         /// </summary>
-        private static BucketedPrioritizedStringList s_si = new BucketedPrioritizedStringList(/*gatherStatistics*/ false, s_smallMruSize, s_largeMruSize, s_hugeMruSize, s_smallMruThreshold, s_largeMruThreshold, s_hugeMruThreshold, s_ginormousThreshold, s_useSimpleConcurrency);
+        private IInternerImplementation s_si;
```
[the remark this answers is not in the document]
ladipro: Yes, this one should have been renamed. Done!

src/Shared/WeakStringCache.cs
ladipro: Great point, thanks. It's used only by tests so far but that's definitely a bug. Fixed by splitting to a public `Scavenge` and non-public `ScavengeUnderLock`.

src/Shared/WeakStringCache.cs
ladipro: The workload is mostly single-threaded and `ReaderWriterLockSlim` has almost twice the overhead if there's no contention compared to a simple lock. So not a good fit in my opinion as the benefits of R/W lock will likely not materialize here.

ladipro:
> > the expected number of collisions for 100k strings is just over 1
> 
> If that's the case could you just drop the overflow list? If there are two strings with the same hash you just don't intern both.

This is an awesome suggestion @danmosemsft. I guess I was too fixated on having a "perfect" implementation. Just to be sure I've re-measured the scenario and ~10k strings generated 6 collisions. So yes, totally reasonable to reduce the overhead (by ~15% as expected) and simplify the code at the expense of these very rare cases where a new string boots an existing string out of the cache.

danmosemsft:
> I've re-measured the scenario and ~10k strings generated 6 collisions.

Curiously high for a 32 bit hash -- it should take ~77K strings to have a single collision, on average. https://en.wikipedia.org/wiki/Birthday_attack  (Doesn't matter - just an aside)

Forgind:
> > I've re-measured the scenario and ~10k strings generated 6 collisions.
> 
> Curiously high for a 32 bit hash -- it should take ~77K strings to have a single collision, on average. https://en.wikipedia.org/wiki/Birthday_attack (Doesn't matter - just an aside)

That's an expected number and assumes a perfect hash. It should actually be 0 more often than anything above 1, but it's notoriously hard to create a perfect hashing algorithm for strings, and this is one manifestation of that.

src/Build.UnitTests/WeakStringCache_Tests.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        /// <summary>
+        /// Triggers full GC that reliably collects all objects without strong GC roots and also nulls out weak GC handles.
+        /// </summary>
+        /// <remarks>
+        /// While a simple GC.Collect() would be enough when running on .NET Framework or .NET Core, the Mono runtime appears to have
+        /// GC handles implemented differently and requires a bit more effort. We don't care when exactly our handles are nulled out
+        /// as long as it eventually happens for otherwise unreachable strings.
+        /// </remarks>
+        private void RunGC()
```
Forgind, line 34: Current mono failures all refer to RunGC, and all tests that refer to it seem to have failed.

danmosemsft:
> That's an expected number and assumes a perfect hash. It should actually be 0 more often than anything above 1, but it's notoriously hard to create a perfect hashing algorithm for strings, and this is one manifestation of that.

Ah, perhaps this is using the custom hashing in this PR. .NET (at least .NET Core) uses Marvin, a cryptographically secure hash, for string.GetHashCode(). Except in special cases for performance (like in `Dictionary<string,V>`) where we use an [internal non randomized hash](https://github.com/dotnet/runtime/blob/6b5850fa7cf476b66be9d01a3ed05f4484b01d3a/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs#L783) but we are careful to revert to a secure hash if the collisions build up.

Unless all input is trusted, the interning cache should be bounded, unless using a secure hash or counting collisions in this way. If a non secure hash is OK I wonder whether the one linked above that CLR uses is interesting. It's supposed to be fast.

src/Build.UnitTests/WeakStringCache_Tests.cs
ladipro: Aha! I finally figured out why these tests have been problematic on Mac Mono. From https://www.mono-project.com/docs/advanced/garbage-collector/sgen/#precise-stack-marking:

_"Both Boehm GC, and SGen on macOS scan threads stacks conservatively. This means that they treated every value found on a stack as a potential pointer. This is problematic for several reasons:_

_- The possibility that some objects remain alive by mistake, because a value which points to them is found on the stack."_

Conservative GC scanning makes it very hard to write a reliable test where we expect an object to get collected. I could try to hack around it but I think we can just disable these tests on Mono to save us from future pain.

rainersigwald: Sounds fine to me. Maybe link to that page for posterity :)

ladipro:
> > I've re-measured the scenario and ~10k strings generated 6 collisions.
> 
> Curiously high for a 32 bit hash -- it should take ~77K strings to have a single collision, on average. https://en.wikipedia.org/wiki/Birthday_attack (Doesn't matter - just an aside)
> 

I took a closer look and the number reported by `GetDebugInfo` was actually not calculated only from currently "live" strings but from all buckets in the dictionary. Meaning that those 6 collisions have happened some time in the past and maybe even at different times as the cache was holding strings that got later collected.

The true number of collisions for the ~10k strings in my scenario is indeed 0, matching the expected number.

> Unless all input is trusted, the interning cache should be bounded, unless using a secure hash or counting collisions in this way. If a non secure hash is OK I wonder whether the one linked above that CLR uses is interesting. It's supposed to be fast.

I don't think that we have to be using a secure hash. The user can DoS MSBuild in so many ways by crafting project files, and hash collisions in a cache that dedups strings does not seem to be at the top of the list honestly.

The non-randomized implementation you linked looks almost equivalent to what's currently in this PR. The differences are: 1) The math is done per character and not per dword, 2) Using addition vs. xor. I'll switch to xor for good measure. If CLR is using it chances are it's been proven better. Or it's at least a bit easier on the transistors :)

danmosemsft:
> The non-randomized implementation you linked looks almost equivalent to what's currently in this PR. The differences are: 1) The math is done per character and not per dword, 2) Using addition vs. xor. I'll switch to xor for good measure. If CLR is using it chances are it's been proven better. Or it's at least a bit easier on the transistors :)

Are we looking at the same thing? These two things look pretty different 😄 
https://github.com/microsoft/msbuild/pull/5353/commits/e97f3f1acc75a655428ca0164b10ab7174ef8b18#diff-8d1206ab0c9dbc43fac981181079c854R154-L165
https://github.com/dotnet/runtime/blob/6b5850fa7cf476b66be9d01a3ed05f4484b01d3a/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs#L783-L812

I'm not familiar with djb2, but what I found on the web has `h * 33 + c`  -- if your xor version isn't an existing well characterized flavor, you'd have to do some original research into its properties which might be time consuming.

Another option is to just paste in the CLR one. That way you have some guarantee of its qualities, and if there's an issue, they'll fix it. You would have to replace `BitOperations.RotateLeft` (which is .NET Core specific):
```c#
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static uint RotateLeft(uint value, int offset)
            => (value << offset) | (value >> (32 - offset));
```

This isn't my codebase (just my very old code) so this is just my 2c.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// </summary>
+        /// <param name="internable">The internable describing the string we're looking for.</param>
+        /// <returns>A string matching the given internable.</returns>
+        public string GetOrCreateEntry<T>(T internable, out bool cacheHit) where T : IInternable
+        {
+            int hashCode = GetInternableHashCode(internable);
+
+            StringWeakHandle handle;
+            string result;
+            lock (_stringsByHashCode)
```
cdmihai, line 114: Quickbuild does parallel evaluations. As far as I remember, lock contention in the opportunistic string cache was showing as a bottleneck so they added the ConcurrentDictionary implementation to relieve it. This means QB will most likely have to keep using the ConcurrentDictionary approach.

cdmihai: Actually, you could use a ConcurrentDictionary and only fully lock when scavenging right?

---
In reply to: [428400003](https://github.com/microsoft/msbuild/pull/5353#discussion_r428400003) [](ancestors = 428400003)

drewnoakes:
> Using addition vs. xor. I'll switch to xor for good measure. If CLR is using it chances are it's been proven better. Or it's at least a bit easier on the transistors :)

My understanding is that XOR is favoured as ADD biases towards having more 1 bits due to carries.

ladipro:
> Are we looking at the same thing? These two things look pretty different 😄

Aren't

```C#
(BitOperations.RotateLeft(hash1, 5) + hash1)
```

and

```C#
hash1 * 33
```

almost the same things? It looks like the CLR implementation is also based on djb2 but chose to rotate instead of shift, probably because this way it can process multiple characters at the same time kind of symmetrically.

There seem to be surprisingly few resources on the web on these hash functions. FWIW, the xor variant of djb2 is mentioned [here](http://www.cse.yorku.ca/~oz/hash.html) as "now favored by the author".

After we figure out the issue with `StringBuilder` and can represent all inputs to the interner with `ReadOnlySpan<char>` I think it would make sense to adopt the CLR implementation.

src/Shared/WeakStringCache.cs
ladipro: I was considering `ConcurrentDictionary` but it's not available on downlevel so it didn't seem worth it. If parallel evaluations are a thing, it definitely changes the priorities, thank you!

I have switched `WeakStringCache` to use `ConcurrentDictionary` where available. Not relying on `AddOrUpdate` because it doesn't have much benefits in this use case and has about the same perf as `TryGetValue` followed by `TryAdd`. I found places where we could leak handles, though, and fixed them.

How difficult would it be to test these changes in Quickbuild before or shortly after merging?

danmosemsft: `ConcurrentDictionary` is available back to .NET Framework 4.5. How far do you guys support downlevel?

https://apisof.net/catalog/System.Collections.Concurrent.ConcurrentDictionary%3CTKey,TValue%3E

rainersigwald: .NET 3.5 for MSBuildTaskHost :(

I'm happy to have "correct enough but horrible performance" implementations for that codepath, though. That's what I'm doing for `System.Collections.Immutable` in #5279.

src/Shared/WeakStringCache.cs
```diff
@@ -86,16 +89,24 @@ public void Free()
         /// <summary>
         /// Improvised capacity for the scavenging heuristics.
         /// </summary>
-        private int _capacity = 100;
+        private int _capacity = 101;
 
         /// <summary>
         /// A dictionary mapping string hash codes to weak GC handles.
         /// </summary>
+#if CLR2COMPATIBILITY
```
cdmihai, line 97: I find ifdeffed types really hard to read and understand. I suggest making an interface or abstract class out of OpportunisticIntern with two implementations, a super simple, short, and non-optimized used just by MSBuildTaskHost, and the WeakStringCache implementation for the rest. MSBuildTaskHost.csproj would have a reference to the IOpportunisticIntern.cs interface / abstract class, and provide its own simple implementation.

danmosemsft: In the core libraries we use partial classes wherever possible. Sometimes with partial methods, sometimes with a private "XXXCore()" method with divergent implementations that the single entrypoint API calls into. Usually for OS specific impl but sometimes for target framework. They seem to be easier to read and of course have zero runtime impact. Possibly an option here.

cdmihai: True, that's another way of separating common interface from varying implementation, as long as it actually achieves a good compromise between cognitive separation of concerns and runtime perf.

ladipro: I have split WeakStringCache.cs into three files and used partial classes instead of ifdefs. It is definitely cleaner now and should be easy for an otherwise uninitiated reader to understand. At least after they realize that only one of WeakStringCache.Locking.cs and WeakStringCache.Concurrent.cs is always included in the build. I'm wondering if there's a better way to name the files to make it more obvious.

I have also tweaked the Concurrent implementation and convinced myself that it is now free of race conditions. It is complex enough that maybe writing a small stress test would be worthwhile. Please review this code carefully.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,143 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Build
+{
+    /// <summary>
+    /// A cache of weak GC handles (equivalent to weak references) pointing to strings. As long as a string has an ordinary strong GC root
```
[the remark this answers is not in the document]
ladipro: Done!

src/Shared/WeakStringCache.Locking.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Build
+{
+    /// <summary>
+    /// Implements the WeakStringCache functionality on .NET Framework 3.5 where ConcurrentDictionary is not available.
+    /// </summary>
+    internal sealed partial class WeakStringCache : IDisposable
+    {
+        private Dictionary<int, StringWeakHandle> _stringsByHashCode;
```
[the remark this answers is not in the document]
ladipro: Done. Here and in WeakStringCache.Concurrent.cs.

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,143 @@
...
+            /// Frees the GC handle.
+            /// </summary>
+            public void Free()
+            {
+                WeakHandle.Free();
+            }
+        }
+
+        /// <summary>
+        /// Improvised capacity for the scavenging heuristics.
```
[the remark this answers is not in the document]
ladipro: It's just a fairly random estimate of the number of interned strings. I have beefed up the comment and increased it a bit. FWIW, the legacy dictionary as used with "simple concurrency" is created with the initial capacity of 2053 but the use case there is QuickBuild so presumably larger solutions.

src/Build.UnitTests/WeakStringCache_Tests.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// </summary>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void AddStringsWithSameHashCode(int numberOfStrings)
+        {
+            string[] cachedStrings = new string[numberOfStrings];
+            int[] hashCodes = new int[numberOfStrings];
+
+            for (int i = 0; i < numberOfStrings; i++)
+            {
+                string strPart2 = "1" + String.Concat(Enumerable.Repeat("4428939786", i));
```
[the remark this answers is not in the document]
ladipro: Brute force :)

src/Build.UnitTests/WeakStringCache_Tests.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// 1. A string added to the cache stays in the cache as long as it's alive.
+        /// 2. The string is no longer retrievable after all strong GC refs are gone.
+        /// 3. The cache completely removes the handle after calling Scavenge on it.
+        /// </summary>
+        /// <remarks>
+        /// Disabled on MacOS Mono because it doesn't play well with conservative GC scanning.
+        /// https://www.mono-project.com/docs/advanced/garbage-collector/sgen/#precise-stack-marking
+        /// </remarks>
+        [Fact]
+        [Trait("Category", "mono-osx-failing")]
```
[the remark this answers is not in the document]
ladipro: Done, thanks.

src/Shared/WeakStringCache.cs
cdmihai: Looks good. Maybe change the filename to signify which is the implementation specific for the taskhost, like `WeakStringCache.Locking.cs` -> `WeakStringCache.Concurrent.MSBuildTaskHost.cs`. Maybe that makes it more clear which is the swapable logic. But even as it is, I find it pretty clear :)

src/Shared/WeakStringCache.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                WeakHandle.Free();
+            }
+        }
+
+        /// <summary>
+        /// Improvised capacity for the scavenging heuristics. We use this number as the initial capacity of the underlying collection
+        /// and then as the maximum size we let the collection grow before scavenging unused entries.
+        /// </summary>
+        private int _capacity = 503;
```
cdmihai, line 91: nit: is it worth renaming this to something like `_scavengeThreshold`?

src/Shared/WeakStringCache.cs
cdmihai: > I have also tweaked the Concurrent implementation and convinced myself that it is now free of race conditions. It is complex enough that maybe writing a small stress test would be worthwhile. Please review this code carefully.

Yea, I have a hard time reasoning whether there are any incorrect thread interleavings. On the other hand, as far as I can tell, the worst case interleaving just busts some of the weak handles with the cost of having to materialize extra IInternables right after a scavenge with the incorrect interleavings. The alternative would be to put a readers/writers lock around scavenging. Scavenging takes the writers lock, and `GetOrCreateEntry` takes the readers lock.

src/Shared/WeakStringCache.cs
ladipro: Yes, this was confusing. I have added a new const `_initialCapacity` and renamed the r/w field to `_scavengeThreshold` as suggested. Thank you!

src/Shared/WeakStringCache.cs
ladipro: >Another option to make it more clear which implementation serves which purpose is to move WeakStringCache.Locking.cs from the Shared directory to the msbuildtaskhost assembly

Great point, done!

> The alternative would be to put a readers/writers lock around scavenging. Scavenging takes the writers lock, and GetOrCreateEntry takes the readers lock.

I wanted to avoid this. Acquiring the read lock has its overhead even if there's no contention and in highly parallel scenarios (QuickBuild evaluation) scavenging would basically stop all other threads and that's not desirable.

ladipro: > The stress test would help :)

Stress test passed :)

```C#
            for (int i = 0; i < Environment.ProcessorCount; i++)
            {
                Thread thread = new Thread(() =>
                {
                    Random rnd = new Random();
                    while (true)
                    {
                        OpportunisticIntern.InternStringIfPossible(rnd.Next().ToString());
                    }
                });
                thread.Start();
            }
```

I don't think stress should be part of unit tests due to its long-running nature. I'll think about running something like this elsewhere.

src/Shared/IInternable.cs
ladipro: This is now tracked with https://github.com/microsoft/msbuild/issues/5438.

PR #5357, opened 2020-05-18
Get past first hurdle loading a project from \\wsl$

rainersigwald:
WSL2 uses a 9p share exposed as `\\wsl$` to allow the Linux filesystem
to be seen by Windows and win32 processes. This was failing with a
`System.UriFormatException` because `wsl$` is not a valid URI hostname.

Instead, just don't treat these files as read-only.

### DIFF
--- a/src/Build/ElementLocation/XmlDocumentWithLocation.cs
+++ b/src/Build/ElementLocation/XmlDocumentWithLocation.cs
@@ -139,7 +139,14 @@ namespace Microsoft.Build.Construction
         {
             if (reader.BaseURI.Length > 0)
             {
-                DetermineWhetherToLoadReadOnly(new Uri(reader.BaseURI).LocalPath);
+                string adjustedLocalPath = null;
+
+                if (Uri.TryCreate(reader.BaseURI, UriKind.RelativeOrAbsolute, out Uri uri))
+                {
+                    adjustedLocalPath = uri.LocalPath;
+                }
+
+                DetermineWhetherToLoadReadOnly(adjustedLocalPath);
             }
 
             // Set the line info source if it is available given the specific implementation of XmlReader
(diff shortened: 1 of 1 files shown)


src/Build/ElementLocation/XmlDocumentWithLocation.cs
```diff
@@ -340,7 +347,7 @@ private void DetermineWhetherToLoadReadOnly(string fullPath)
                 {
                     _loadAsReadOnly = true;
                 }
-                else if (s_readOnlyFlags == ReadOnlyLoadFlags.LoadReadOnlyIfAppropriate)
+                else if (s_readOnlyFlags == ReadOnlyLoadFlags.LoadReadOnlyIfAppropriate && fullPath is object)
```
ladipro, line 350: nit: That's an unusual way to check for null. First I was like 🤔 and then 🤦 and 🙃

rainersigwald: I blame @jaredpar: https://twitter.com/jaredpar/status/1171474157667643400

jaredpar: This is **the** way to check for non-null 😉

jaredpar: This comment string was an emotional roller coaster for me. First an email from a MSBuild PR where @rainersigwald said "i blame @jaredpar" which is downright terrifying.  To actually looking at the PR and seeing proper `null` check code :smile:

ladipro: I guess I haven't been keeping up with the latest development in C#. The way to check for non-null for some, operator abuse for others 😛 

Is it part of a larger effort to make C# read like English? Trying to help VB developers switch the language? Genuinely curious, thank you!

ladipro: Found the answer on Twitter. Cool stuff!

Issue #5360, opened 2020-05-20
IndexOf(Char, StringComparison) is used instead of IndexOf(Char, Int32), resulting in build errors.


auott:
### Steps to reproduce

Add the following to a project file or targets file that gets included when building:

```xml
  <PropertyGroup>
    <Revision>$([System.DateTime]::Now.ToString(yyyyMMdd)).99</Revision>
    <TheIndex>$(Revision.IndexOf('.', 10))</TheIndex>
  </PropertyGroup>
```

### Expected behavior
The project can build successfully and IndexOf(Char, Int32) is invoked on Revision.

### Actual behavior
An error is produced:
```
error MSB4184: The expression ""20200519.99".IndexOf(., 10)" cannot be evaluated. The string comparison type passed in is currently not supported. Parameter name: comparisonType
```

It appears that it's only attempting to use IndexOf(Char, StringComparison) and 10 isn't
a valid value for the StringComparison enum. However, it should still be using IndexOf(Char, Int32)
### Environment data
`msbuild /version` output:

Microsoft (R) Build Engine version 16.7.0-preview-20265-07+31a1a34cb for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

16.7.0.26507

OS info:
Windows - 10.0.18362

If applicable, version of the tool that invokes MSBuild (Visual Studio, dotnet CLI, etc):
Occurs opening the project in VS or when trying to build the project from a VS command prompt.

andrewcrawley:
Note that this worked fine until recently.  Here's a working version of msbuild:

```
D:\>msbuild /version
Microsoft (R) Build Engine version 16.7.0-preview-20228-06+9571f7584 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

16.7.0.22806
```

rainersigwald:
Probably introduced by #5342. cc @cdmihai.

Let's try to revert before shipping 16.7preview2.

rainersigwald:
(Thanks for the report @auott!)

Issue #2217, opened 2017-06-14
MSBuild throws thousands of first-chance exceptions evaluating projects with certain types of property functions


davkean:
This makes debugging VS extremely painful. 

Opening Roslyn.sln, MSBuild threw over > 21,000 first chance exceptions opening Roslyn.sln, with stacks similar to:

<details>

        MissingMethodException: Attempted to access a missing member.
 	mscorlib.dll!System.DefaultBinder.BindToMethod(System.Reflection.BindingFlags bindingAttr, System.Reflection.MethodBase[] match, ref object[] args, System.Reflection.ParameterModifier[] modifiers, System.Globalization.CultureInfo cultureInfo, string[] names, out object state)	Unknown
 	Microsoft.Build.dll!Microsoft.Build.Evaluation.Expander<Microsoft.Build.Evaluation.ProjectProperty, Microsoft.Build.Evaluation.ProjectItem>.Function<Microsoft.Build.Evaluation.ProjectProperty>.Execute(object objectInstance, Microsoft.Build.Evaluation.IPropertyProvider<Microsoft.Build.Evaluation.ProjectProperty> properties, Microsoft.Build.Evaluation.ExpanderOptions options, Microsoft.Build.Shared.IElementLocation elementLocation)	Unknown
 	Microsoft.Build.dll!Microsoft.Build.Evaluation.Expander<Microsoft.Build.Evaluation.ProjectProperty, Microsoft.Build.Evaluation.ProjectItem>.PropertyExpander<Microsoft.Build.Evaluation.ProjectProperty>.ExpandPropertyBody(string propertyBody, object propertyValue, Microsoft.Build.Evaluation.IPropertyProvider<Microsoft.Build.Evaluation.ProjectProperty> properties, Microsoft.Build.Evaluation.ExpanderOptions options, Microsoft.Build.Shared.IElementLocation elementLocation, Microsoft.Build.Evaluation.UsedUninitializedProperties usedUninitializedProperties)	Unknown
[...]

PR #5363, opened 2020-05-21
A better addition of String.IndexOf to well known functions

cdmihai:
While not regressing #5360

I guess the lesson here is to always add unit tests for touched code .... 🙄

### DIFF
--- a/src/Build.UnitTests/Evaluation/Expander_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Expander_Tests.cs
@@ -2777,6 +2777,26 @@ namespace Microsoft.Build.UnitTests.Evaluation
             Assert.Equal(expectedExpansion, result);
         }
 
+        [Theory]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('x', 1))", "3")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('x', 1, 4))", "3")]
+        // 9 is not a valid StringComparison enum value
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('x', 9))", "10")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('X', 'StringComparison.Ordinal'))", "-1")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('X', 'StringComparison.OrdinalIgnoreCase'))", "0")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('X4', 'StringComparison.OrdinalIgnoreCase'))", "3")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('X4', 1, 'StringComparison.OrdinalIgnoreCase'))", "3")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('X', 1, 3, 'StringComparison.OrdinalIgnoreCase'))", "3")]
+        public void StringIndexOfTests(string propertyName, string properyValue, string propertyFunction, string expectedExpansion)
+        {
+            var pg = new PropertyDictionary<ProjectPropertyInstance>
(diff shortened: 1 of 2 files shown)


src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2777,6 +2777,26 @@ public void PropertyFunctionRuntimeInformation(string propertyFunction, string e
             Assert.Equal(expectedExpansion, result);
         }
 
+        [Theory]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('x', 1))", "3")]
+        [InlineData("AString", "x12x456789x11", "$(AString.IndexOf('x', 1, 4))", "3")]
```
rainersigwald, line 2782: Nit: multi-character string as first arg with no StringComparison specified.

src/Build/Evaluation/Expander.cs
```diff
@@ -4242,6 +4240,47 @@ private static bool TryGetArg(object[] args, out string arg0)
...
+
+                // Allow fully-qualified enum, e.g. "System.StringComparison.OrdinalIgnoreCase"
+                if (comparisonTypeName.Contains('.'))
+                {
+                    comparisonTypeName = comparisonTypeName.Replace("System.StringComparison.", "").Replace("StringComparison.", "");
+                }
+
+                return Enum.TryParse(comparisonTypeName, out arg1);
+
+                bool ContainsDigits(string aString)
```
Forgind, line 4270: Instead of writing this yourself, you could use:
`comparisonTypeName.Any(c => char.IsDigit(c))` above.

src/Build/Evaluation/Expander.cs
```diff
@@ -3537,6 +3537,14 @@ private bool TryExecuteWellKnownFunction(out object returnVal, object objectInst
                             return true;
                         }
                     }
+                    else if (string.Equals(_methodMethodName, nameof(string.IndexOf), StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (TryGetArgs(args, out string arg0, out StringComparison arg1))
```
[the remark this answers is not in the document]
cdmihai: Not worth it, gets complicated to guess what the real binder would do.

src/Build/Evaluation/Expander.cs
cdmihai: While not a super hot path, it's still called pretty often so I wanted to avoid linq.

ladipro: The Linq version is indeed significantly slower. I understand the intention here is to see if the argument is a number as opposed to an identifier. Noting that identifiers may have digits in them (luckily not `StringComparison` fields, at least in its current version), I'm wondering if `int.TryParse` wouldn't be a more robust check. And if not, maybe it would make sense to move this helper to Shared\StringExtensions.cs.

src/Build/Evaluation/Expander.cs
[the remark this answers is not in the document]
Forgind: If it's empty, can't we just call a different IndexOf method? Then we don't have to guess.

src/Build/Evaluation/Expander.cs
rainersigwald: I like using `int.TryParse`. Does that work?

aolszowka:
Will resolve at least one of our issues https://github.com/microsoft/msbuild/issues/2217#issuecomment-517021765 thanks for the change, when will this drop?

Forgind:
> Will resolve at least one of our issues [#2217 (comment)](https://github.com/microsoft/msbuild/issues/2217#issuecomment-517021765) thanks for the change, when will this drop?

Probably 16.7 preview 3.

src/Build/Evaluation/Expander.cs
cdmihai: Yup, changed it.

PR #5369, opened 2020-05-22
Disable hanging tests

Forgind:
NormalPriorityBuild has hung several times. I haven't been able to figure out why it would hang and other tests that use the same path (ExecuteMSBuildExeExpectSuccess) wouldn't.

### DIFF
--- a/src/MSBuild.UnitTests/XMake_Tests.cs
+++ b/src/MSBuild.UnitTests/XMake_Tests.cs
@@ -1212,6 +1212,7 @@ namespace Microsoft.Build.UnitTests
             }
         }
 
+        /* These tests hung repeatedly. See https://github.com/microsoft/msbuild/issues/5368
         /// <summary>
         /// Test that low priority builds actually execute with low priority.
         /// </summary>
@@ -1250,7 +1251,6 @@ namespace Microsoft.Build.UnitTests
             IDictionary<string, string> environmentVars = new Dictionary<string, string>
             {
                 { "MSBUILDNOINPROCNODE", "1"},
-                { "DISABLECONSOLECOLOR", "1"},
                 { "MSBUILDENABLEALLPROPERTYFUNCTIONS", "1" },
             };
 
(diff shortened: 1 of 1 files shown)


src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -1250,7 +1251,6 @@ private void RunPriorityBuildTest(ProcessPriorityClass expectedPrority, params s
             IDictionary<string, string> environmentVars = new Dictionary<string, string>
             {
                 { "MSBUILDNOINPROCNODE", "1"},
-                { "DISABLECONSOLECOLOR", "1"},
```
rainersigwald, line 1253: Why this change?

src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -1212,6 +1212,7 @@ public void ResponseFileSupportsThisFileDirectory()
             }
         }
 
+        /* These tests hung repeatedly. See https://github.com/microsoft/msbuild/issues/5368
```
rainersigwald, line 1215: Don't comment tests; instead, use 

```C#
[Fact(Skip = "url of bug")]
```

src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -1259,8 +1259,9 @@ private void RunPriorityBuildTest(ProcessPriorityClass expectedPrority, params s
             string expected = string.Format(@"Task priority is '{0}'", expectedPrority);
             logContents.ShouldContain(expected, () => logContents);
         }
+        */
 
-#region IgnoreProjectExtensionTests
+        #region IgnoreProjectExtensionTests
```
rainersigwald, line 1264: Revert whitespace change please.

src/MSBuild.UnitTests/XMake_Tests.cs
Forgind: This was part of trying to figure out why it was hanging, but I forgot to revert it. Will fix.

PR #5370, opened 2020-05-22
Log graph construction metrics

cdmihai:
I always add something like this when debugging static graph, so I decided to just send a PR.


src/Build/BackEnd/BuildManager/BuildManager.cs
```diff
@@ -2499,6 +2506,13 @@ private bool ReuseOldCaches(string[] inputCacheFiles)
             }
         }
 
+        private void LogMessage(string message)
+        {
+            var loggingService = ((IBuildComponentHost)this).LoggingService;
+
+            loggingService?.LogCommentFromText(BuildEventContext.Invalid, MessageImportance.High, message);
```
cdmihai, line 2513: Should we create a new BuildEventArg type for logs generated by the engine (by the Scheduler, BuildManager, BuildRequestEngine, etc)?

src/Build/Graph/ProjectGraph.cs
```diff
@@ -396,7 +415,9 @@ public ProjectGraph(ProjectGraphEntryPoint entryPoint, ProjectCollection project
         {
             ErrorUtilities.VerifyThrowArgumentNull(projectCollection, nameof(projectCollection));
 
-            projectInstanceFactory = projectInstanceFactory ?? DefaultProjectInstanceFactory;
+            var timer = Stopwatch.StartNew();
```
Forgind, line 418: Should the timer be here in debug mode only?

src/Build/BackEnd/BuildManager/BuildManager.cs
Forgind: I'd really like to have BuildEventArgs structured differently so we can avoid problems like [this](https://github.com/microsoft/msbuild/issues/5019). That being said, we already have a lot, and adding a new one would make this logging a bit cleaner...just also make changing them all to use a visitor pattern a little harder.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
```diff
@@ -64,6 +64,21 @@ public void ConstructWithSingleNode()
             }
         }
 
+        [Fact]
```
Forgind, line 67: I'm a little confused—what does this test have to do with the rest of the PR?

cdmihai: Asserts that metrics are actually computed.

src/Build/Graph/ProjectGraph.cs
cdmihai: The intent is to always log graph construction metrics.

ladipro: Can you please log Start/Stop ETW events as well? (ref: src/Framework/MSBuildEventSource.cs)

This will be super useful for automated perf runs to make sure that the graph building and evaluation performance hasn't regressed.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
Forgind: That still feels perpendicular to this PR, since it isn't about computing anything but rather logging how long it took.

src/Build/Graph/ProjectGraph.cs
Forgind: You could theoretically get this from grepping through a log, right? It's just messier As far as I'm concerned, either approach is fine, and ETW events would be lighter, but this form of logging it would run every time, whereas ETW events only fire when they're enabled.

src/Build.UnitTests/Graph/ProjectGraph_Tests.cs
rainersigwald: I don't understand your concern @Forgind. This PR introduces a new public field, and this is a lightweight test that the field is populated.

src/Build/BackEnd/BuildManager/BuildManager.cs
rainersigwald: I like more-specific events but am fine punting this. I think it mostly depends on if you want it in a logger in BuildXL or another application.

src/Build/Graph/ProjectGraph.cs
rainersigwald: Agree with @ladipro: also add ETW.

src/Build/BackEnd/BuildManager/BuildManager.cs
cdmihai: BXL and QB would use the API to get the information, not loggers. Let's leave it like this and upgrade to typed event if need to.

src/Framework/MSBuildEventSource.cs
```diff
@@ -398,6 +398,18 @@ public void ExecuteTaskReacquireStop(string taskName, int taskID)
         {
             WriteEvent(52, taskName, taskID);
         }
+
+        [Event(53)]
+        public void ProjectGraphConstructionStart()
```
cdmihai, line 403: Decided to not send any data in. The only relevant data that I could send in would be a super long string concatenating either the graph entry point project files, or the graph entry point project files together with their global properties, but this can be inferred from the evaluation events between the graph construction start / end.

src/Build/Graph/ProjectGraph.cs
```diff
@@ -435,6 +437,7 @@ public ProjectGraph(ProjectGraphEntryPoint entryPoint, ProjectCollection project
 
             _projectNodesTopologicallySorted = new Lazy<IReadOnlyCollection<ProjectGraphNode>>(() => TopologicalSort(GraphRoots, ProjectNodes));
 
+            MSBuildEventSource.Log.ProjectGraphConstructionStop();
```
Forgind, line 440: Do you want to log what you constructed?

src/Framework/MSBuildEventSource.cs
Forgind: Just what I was about to ask about 👍

Forgind: Although you should be able to send a variable number of strings rather than concatenating everything, so it should be performant anyway.

cdmihai: The project path strings are wrapped in other objects, so it requires constructing a collection to put them in (not really that perf impacting, really, only happens once). Global properties for each entry project would be more expensive to format into strings, but again, probably not that impactful since it runs once. On the other hand, all this information is available from the evaluation events that happen between graph construction start/stop, so I'm unsure whether it's worth putting on. Let me know which option you prefer (project paths, project paths + global properties for each one, or nothing) and I'll add it in

Forgind: If you wrap it in `MSBuildEventSource.Log.IsEnabled()` calls, you'll only do the calculations when someone is actually interested in the data, so it would have ~no effect on perf. That also means that my tendency would be to log everything everywhere—no point in hiding information. I don't feel that strongly, though, personally.

cdmihai: Cool, pushed a commit with the changes.

Forgind:
Last week, I broke this. This week, the long running test (for Windows core, at least) was:
Microsoft.Build.Graph.UnitTests.GraphLoadedFromSolutionTests.SolutionBasedGraphCanMatchProjectSpecificConfigurations
so you might want to take a look.

PR #5371, opened 2020-05-23
Update Binary-Log.md

shargon:
Update current file format from documentation

### DIFF
--- a/documentation/wiki/Binary-Log.md
+++ b/documentation/wiki/Binary-Log.md
@@ -82,7 +82,7 @@ It is a `GZipStream`-compressed binary stream of serialized `BuildEventArgs` obj
 
 ## Incrementing the file format
 
-Every .binlog file has the first three bytes that indicate the file version. The current file format version is 2 (`00 00 02`).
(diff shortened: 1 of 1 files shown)


documentation/wiki/Binary-Log.md
```diff
@@ -82,7 +82,7 @@ It is a `GZipStream`-compressed binary stream of serialized `BuildEventArgs` obj
 
 ## Incrementing the file format
 
-Every .binlog file has the first three bytes that indicate the file version. The current file format version is 2 (`00 00 02`).
+Every .binlog file has the first three bytes that indicate the file version. The current file format version is 7 (`00 00 07`).
```
rainersigwald, line 85:
```suggestion
Every .binlog file has the first three bytes that indicate the file version. The current file format version is 9 (`00 00 09`).
```

https://github.com/microsoft/msbuild/blob/b0d72851842af6f6b50ad62e6e2077790ee5de29/src/Build/Logging/BinaryLogger/BinaryLogger.cs#L36-L42

Right?

KirillOsenkov: Right.

rainersigwald:
Thanks @shargon!

Issue #5150, opened 2020-02-28
Unlogged error in RAR (?)


rainersigwald:
@sharwell reported offline an IDE build that suddenly failed without any error messages. He was able to capture an exception using PerfView

```
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Reflection.AssemblyName.nGetFileInformation(String s)
   at System.Reflection.AssemblyName.GetAssemblyName(String assemblyFile)
   at Microsoft.Build.Shared.AssemblyNameExtension.GetAssemblyNameEx(String path)
   at Microsoft.Build.Tasks.SystemState.GetAssemblyName(String path)
   at Microsoft.Build.Tasks.ReferenceTable.SetPrimaryAssemblyReferenceItem(ITaskItem referenceAssemblyName)
   at Microsoft.Build.Tasks.ReferenceTable.SetPrimaryItems(ITaskItem[] referenceAssemblyFiles, ITaskItem[] referenceAssemblyNames, List`1 exceptions)
   at Microsoft.Build.Tasks.ReferenceTable.ComputeClosure(IEnumerable`1 remappedAssembliesValue, ITaskItem[] referenceAssemblyFiles, ITaskItem[] referenceAssemblyNames, List`1 exceptions)
   at Microsoft.Build.Tasks.ResolveAssemblyReference.Execute(FileExists fileExists, DirectoryExists directoryExists, GetDirectories getDirectories, GetAssemblyName getAssemblyName, GetAssemblyMetadata getAssemblyMetadata, GetRegistrySubKeyNames getRegistrySubKeyNames, GetRegistrySubKeyDefaultValue getRegistrySubKeyDefaultValue, GetLastWriteTime getLastWriteTime, GetAssemblyRuntimeVersion getRuntimeVersion, OpenBaseKey openBaseKey, GetAssemblyPathInGac getAssemblyPathInGac, IsWinMDFile isWinMDFile, ReadMachineTypeFromPEHeader readMachineTypeFromPEHeader)
   at Microsoft.Build.Tasks.ResolveAssemblyReference.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #5377, opened 2020-05-26
Ensure OutOfMemoryException gets reported as a build failure

sharwell:
See #5150


Forgind:
I tried inserting this, and it still seems to say:
```
Build FAILED.
    0 Warning(s)
    0 Error(s)
```

although without the popups asking if I want to abort.

Issue #5349, opened 2020-05-12
MSBuild SDKs version property is not expanded


andrew-boyarshin:
### Steps to reproduce

Directory.Build.props file:
```xml
<Project>

  <PropertyGroup>
    <LocalPackageVersion>2.0.0-local</LocalPackageVersion>
  </PropertyGroup>

  <Import Project="Sdk.props" Sdk="SharpGenTools.Sdk" Version="$(LocalPackageVersion)" Condition="'$(MSBuildProjectExtension)' == '.csproj'" />

</Project>
```

### Expected behavior
`SharpGenTools.Sdk/2.0.0-local` is resolved and imported.

### Actual behavior
`SharpGenTools.Sdk/$(LocalPackageVersion)` fails to be resolved by NuGet resolver.

Log:
```
Resolving SDK 'SharpGenTools.Sdk/$(LocalPackageVersion)'...
```

### Environment data
```
Process = "C:\Program Files\dotnet\dotnet.exe"
MSBuild executable path = "C:\Program Files\dotnet\sdk\5.0.100-preview.5.20261.11\MSBuild.dll"
Command line arguments = ""C:\Program Files\dotnet\sdk\5.0.100-preview.5.20261.11\MSBuild.dll" -maxcpucount -verbosity:m /t:Restore;Build /bl "-distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,C:\Program Files\dotnet\sdk\5.0.100-preview.5.20261.11\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,C:\Program Files\dotnet\sdk\5.0.100-preview.5.20261.11\dotnet.dll""
MSBuild version = "16.7.0-preview-20258-02+26f6d1d87"
```

### Source
[`ProjectImportElement.ParsedSdkReference`](https://github.com/Microsoft/msbuild/blob/1adb5c56a1cb00ec44986c5dfd2bed81e05212a0/src/Build/Construction/ProjectImportElement.cs#L135) contains unexpanded properties from XML element, but [`Evaluator.ExpandAndLoadImportsFromUnescapedImportExpressionConditioned`](https://github.com/Microsoft/msbuild/blob/6f41a4fa3c6957d430783b7d31fe16e6a13b6e8e/src/Build/Evaluation/Evaluator.cs#L1738) doesn't expand them before passing to SDK resolvers.

---

[...]

PR #5378, opened 2020-05-26
Rework SdkReference computation

andrew-boyarshin:
* Fix `<Import />` `ProjectImportElement.CreateImplicit` setting Sdk to an invalid value (stringified `SdkReference`)
* Provide means to get `SdkReference` and XML attribute locations for the values `SdkReference` was created with.
* Improve support for implicit `<Import />` (so, for instance, if one changes the value on `<Sdk />` element, it will reflect on `<Import />` elements created for this element)
* Drop `ProjectImportElement.SdkReference` cache to reduce data duplication (XML tree is now the authority)
* To compensate for the previous point, be careful about memory allocations (make newly-introduced `internal` types `struct`, `readonly struct`, `readonly ref struct`, pass arguments with `in` when the resulting IL is better) and not doing extra work
* Nullable Reference Types annotations for modified code ranges

Note: this is the preparation part of resolving #5349 _without_ the actual change.

### DIFF
--- a/src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
+++ b/src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
@@ -177,7 +177,7 @@ namespace Microsoft.Build.BackEnd.SdkResolution
         }
 
         /// <summary>
-        /// Used for unit tests only.  This is currently only called through reflection in Microsoft.Build.Engine.UnitTests.TransientSdkResolution.CallResetForTests
+        /// Used for unit tests only.
         /// </summary>
         /// <param name="resolverLoader">An <see cref="SdkResolverLoader"/> to use for loading SDK resolvers.</param>
         /// <param name="resolvers">Explicit set of SdkResolvers to use for all SDK resolution.</param>
--- a/src/Build/Construction/ProjectImportElement.cs
+++ b/src/Build/Construction/ProjectImportElement.cs
@@ -1,7 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#nullable enable
+
+using System;
 using System.Diagnostics;
+using System.Xml;
 using Microsoft.Build.Framework;
 using Microsoft.Build.ObjectModelRemoting;
 using Microsoft.Build.Shared;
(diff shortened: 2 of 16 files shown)


src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs
```diff
@@ -177,7 +177,7 @@ public virtual SdkResult ResolveSdk(int submissionId, SdkReference sdk, LoggingC
         }
 
         /// <summary>
-        /// Used for unit tests only.  This is currently only called through reflection in Microsoft.Build.Engine.UnitTests.TransientSdkResolution.CallResetForTests
+        /// Used for unit tests only.
```
andrew-boyarshin, line 180: The test class is now gone, and this method is now used in many tests.

src/Build/Construction/ProjectImportElement.cs
```diff
@@ -60,144 +95,341 @@ public string Project
...
+#nullable enable
+
         /// <summary>
-        /// Gets or sets the SDK that contains the import.
+        ///     Gets or sets the SDK that contains the import.
         /// </summary>
-        public string Sdk
+        public string? Sdk
         {
-            get => FileUtilities.FixFilePath(GetAttributeValue(XMakeAttributes.sdk));
```
andrew-boyarshin, line 72: Behavioral change there. I feel like this was just copy-pasted from `Project` property. It makes no sense to call `FixFilePath` here, since `Sdk` is not a file path (it can't even contain any slashes).

src/Build/Construction/ProjectRootElement.cs
```diff
@@ -1852,53 +1852,60 @@ internal void MarkAsExplicitlyLoaded()
...
+                var origin = new SdkReferenceOrigin(location, location, location);
+
+                foreach (var sdk in sdkAttribute.Split(MSBuildConstants.SemicolonChar).Select(i => i.Trim()))
                 {
-                    nodes.Add(ProjectImportElement.CreateImplicit("Sdk.props", currentProjectOrImport, ImplicitImportLocation.Top, referencedSdk, this));
-                    nodes.Add(ProjectImportElement.CreateImplicit("Sdk.targets", currentProjectOrImport, ImplicitImportLocation.Bottom, referencedSdk, this));
+                    if (!SdkReference.TryParse(sdk, out var reference))
+                    {
+                        ProjectErrorUtilities.ThrowInvalidProject(location, "InvalidSdkFormat", sdkAttribute);
+                        break;
```
andrew-boyarshin, line 1878: Explicit (unreachable) break to make control flow a bit more correct for the code analysis tools. I don't like these asserts in MSBuild, but I kept them in existing code.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1814,10 +1816,14 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
-                    ProjectErrorUtilities.ThrowInvalidProject(importElement.SdkLocation, "CouldNotResolveSdk", importElement.SdkReference.ToString());
+                    ProjectErrorUtilities.ThrowInvalidProject(
+                        sdkReferenceWithOrigin.Origin.Name,
+                        "CouldNotResolveSdk",
+                        sdkReference.ToString()
+                    );
                 }
 
-                project = Path.Combine(sdkResult.Path, project);
+                project = Path.Combine(sdkResult!.Path, project);
```
andrew-boyarshin, line 1826: Gross, but that's the MSBuild asserts for you.

src/Build/Construction/ProjectSdkElement.cs
```diff
@@ -35,45 +52,46 @@ internal ProjectSdkElement(ProjectElementContainerLink link)
...
+        {
+        }
 
         /// <summary>
         /// Gets or sets the name of the SDK.
         /// </summary>
         public string Name
         {
-            get => GetAttributeValue(XMakeAttributes.sdkName);
+            get => GetAttributeValue(XMakeAttributes.sdkName, true);
```
andrew-boyarshin, line 63: I've decided to annotate this property as non-nullable, but thanks to `GetAttributeValue` not being NRT-annotated, I don't even need to put a `!` here.

src/Build/Construction/ProjectSdkElement.cs
```diff
@@ -35,45 +52,46 @@ internal ProjectSdkElement(ProjectElementContainerLink link)
...
             set
             {
                 ErrorUtilities.VerifyThrowArgumentLength(value, XMakeAttributes.sdkName);
-                SetOrRemoveAttribute(XMakeAttributes.sdkName, value, $"Set SDK Name to {value}", XMakeAttributes.sdkName);
```
andrew-boyarshin, line 49: Fix reason messages here and in the following 2 properties (`param` was not used, inconsistent with `ProjectImportElement`).

src/Build/Construction/ProjectImportElement.cs
```diff
@@ -60,144 +95,341 @@ public string Project
...
             {
                 Project = project,
-                Sdk = sdkReference.ToString(),
                 ImplicitImportLocation = implicitImportLocation,
-                SdkReference = sdkReference,
                 OriginalElement = originalElement
             };
         }
 
+#nullable restore
```
andrew-boyarshin, line 285: Since I don't know the proper nullability of the following 2 methods, I've decided to skip annotating these (the same goes for the ones in `ProjectSdkElement`).

Nirmal4G:
Why not have a `Version` as range instead of having two (`Version` and `MinimumVersion`) attributes, Like specifying NuGet Package version ranges?

andrew-boyarshin:
@Nirmal4G I didn't change the existing way of presenting the data. I only refactored the existing code to enable getting precise locations of each `SdkReference` component. I think this question should be moved into a separate issue.

P.S. It would be great to receive initial feedback on this issue.

rainersigwald:
Team triage: Thank you for your contribution! We continue to be skeptical of #5349. This PR is quite large and reviewing it has gotten stalled a few times. Can you break it up into smaller PRs that are easier to review?

andrew-boyarshin:
@rainersigwald sure, I'll think of something.

rainersigwald:
Closing this due to age. If you'd like to bring it back we can, just let us know.

andrew-boyarshin:
Yeah, I think it was a cool (from technical standpoint), but unnecessary idea (it complicated things for little benefit), since MSBuild data model isn't really used that much when modifying the project files, and when it is, it's not that much of a hassle to process all occurrences of SDK reference, instead of the SDK reference origin.

PR #5379, opened 2020-05-27
Use the file version instead of timestamp in handshake

KirillOsenkov:
Fixes https://github.com/microsoft/msbuild/issues/5375. A timestamp can be different between Microsoft.Build.dll in the toolset and the copy in the amd64 subfolder.

Getting the hash of AssemblyInformationalVersion gets the SHA of the actual file.

### DIFF
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Build.BackEnd
             // Mask out the first byte. That's because old
             // builds used a single, non zero initial byte,
             // and we don't want to risk communicating with them
-            return (Constants.AssemblyTimestamp ^ Int64.MaxValue) & 0x00FFFFFFFFFFFFFF;
+            return (CommunicationsUtilities.FileVersionHash ^ Int64.MaxValue) & 0x00FFFFFFFFFFFFFF;
         }
 
         /// <summary>
--- a/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
+++ b/src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
@@ -80,12 +80,9 @@ namespace Microsoft.Build.BackEnd
         /// Magic number sent by the client to the host during the handshake.
         /// Munged version of the host handshake.
         /// </summary>
-        internal static long GetClientHandshake()
+        internal static long GetClientHandshake(bool enableNodeReuse, bool enableLowPriority)
         {
-            // Mask out the first byte. That's because old
-            // builds used a single, non zero initial byte,
-            // and we don't want to risk communicating with them
-            return (CommunicationsUtilities.FileVersionHash ^ Int64.MaxValue) & 0x00FFFFFFFFFFFFFF;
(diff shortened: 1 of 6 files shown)


src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
```diff
@@ -85,7 +85,7 @@ internal static long GetClientHandshake()
             // Mask out the first byte. That's because old
             // builds used a single, non zero initial byte,
             // and we don't want to risk communicating with them
-            return (Constants.AssemblyTimestamp ^ Int64.MaxValue) & 0x00FFFFFFFFFFFFFF;
+            return (CommunicationsUtilities.FileVersionHash ^ Int64.MaxValue) & 0x00FFFFFFFFFFFFFF;
```
Forgind, line 88: I think it's better to just call CommunicationsUtilities's [GetClientHandshake ](https://github.com/microsoft/msbuild/blob/master/src/Shared/CommunicationsUtilities.cs#L347)(and pass in its [GetHandshakeOptions](https://github.com/microsoft/msbuild/blob/master/src/Shared/CommunicationsUtilities.cs#L496) method). If that doesn't work, this is certainly an improvement.

KirillOsenkov: Sounds good! Just keep in mind that we still need to zero the first byte. Looks like GetClientHandshake is already derived from FileVersionHash and a few other deterministic values so we should be good.

Also I just realized we should probably delete Constants.AssemblyTimestamp, so if you're in that area...

Also CI is failing and I have no idea why.

KirillOsenkov: Yeah go ahead it's highly unlikely that it's going to be a problem.

Forgind: I had a chance to look at some of the failures. They seem to have failed for three separate reasons, each involving failing to download a nuget package, and it's very odd that Windows Core passed. If any one of them had failed, I would have assumed it was just a race condition. Seeing that pattern, noting that it's a very small change that should not require downloading new NuGet packages, and noting that all tests passed when I ran them locally, my best guess is still that it's just flakiness, just impressively unfortunate flakiness.

Forgind: I made the change [here](https://github.com/Forgind/msbuild/commit/ed7540c76378c54e00fdf6a29e96a2ff136c8775). If you think that looks good, feel free to cherry-pick it onto your change.

KirillOsenkov: Pushed thanks. Feel free to also push directly to my branch in the future (I always set "allow commits from maintainers").

KirillOsenkov:
Should we also delete Constants.AssemblyTimestamp?

Forgind:
> Should we also delete Constants.AssemblyTimestamp?

Sounds good to me!

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProc.cs
Forgind: If it's a small change, I would do that, but I avoid doing it for larger changes as long as the original author is still engaged. In theory, I shouldn't really be approving of something that I mostly wrote, so it's good to have at least one other pair of eyes on it.

Issue #5372, opened 2020-05-24
Linux "dotnet run" fails with Win32Exception when launching at low process priority


Foritus:
Hello, this has recently regressed and caused my build system to fall over. We manually invoke `dotnet build` with `nice -n20 ionice -c3` to ensure the machine remains responsive during a large build, but this now causes msbuild to crash in recent builds.

It looks like the bug is here: https://github.com/microsoft/msbuild/blob/236f48ecb51e413c13a16d55b6712df4d0a3ca9f/src/MSBuild/XMake.cs#L630
MSBuild needs to store the process' current priority instead of assuming it is running at "Normal".

Commit that introduced the regression: https://github.com/microsoft/msbuild/commit/b111470ae61eba02c6102374c2b7d62aebe45f5b#diff-3099880c51224257bef6905bc2dfbf3d

### Steps to reproduce

dotnet new webapi
nice -n20 dotnet build

Observe crash.
MSBUILD : error MSB1025: An internal failure occurred while running MSBuild.
System.ComponentModel.Win32Exception (13): Permission denied
   at System.Diagnostics.Process.set_PriorityClassCore(ProcessPriorityClass value)
   at System.Diagnostics.Process.set_PriorityClass(ProcessPriorityClass value)
   at Microsoft.Build.CommandLine.MSBuildApp.Execute(String[] commandLine)
Unhandled exception. System.ComponentModel.Win32Exception (13): Permission denied
   at System.Diagnostics.Process.set_PriorityClassCore(ProcessPriorityClass value)
   at System.Diagnostics.Process.set_PriorityClass(ProcessPriorityClass value)
   at Microsoft.Build.CommandLine.MSBuildApp.Execute(String[] commandLine)
   at Microsoft.Build.CommandLine.MSBuildApp.Main(String[] args)

### Expected behavior
MSBuild doesn't crash.

### Actual behavior
It crashes.

### Environment data
dotnet --version
3.1.300

OS info:
Ubuntu Server 18 lts

Foritus:
I made a small change on a fork to see what the fix might look like: https://github.com/microsoft/msbuild/compare/master...Foritus:master

cbricart:
related: #5365

rainersigwald:
Thanks for the report! @Forgind, can you look at this? I think it might be more complete than #5365.

rainersigwald:
Duplicate of #5365

Forgind:
[...]

PR #5381, opened 2020-05-27
Respect current priority Fixes #5365

### LINKED ISSUE #5365: Update 3.4.1 breaks building Apache Thrift (permission issue?) on Ubuntu 18.04
This issue has been reported originally as https://github.com/dotnet/core/issues/4719, and @dagood has done a great job to isolate a potential cause. I'll summarize the original issue and dagood's findings below, in case this is _indeed_ an msbuild issue. If its not related to msbuild, please forgive.

@emmenlau's original report:
I've been using dotnet on Ubuntu Linux 18.04 x86_64 (installed as in https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-1804) for almost a year successfully to build https://github.com/apache/thrift. Yesterday I've installed system updates (which may have included an update of `dotnet`). Usually I install updates at least once every one or two months.

Since today, the build of Thrift fails with error:
```
[...]
cd /home/user/thrift/lib/netstd/Thrift && /usr/bin/dotnet msbuild Thrift.csproj /nologo /restore /target:Rebuild /property:Configuration=Debug /property:OutDir=/home/user/Debug/thrift/lib/netstd/Thrift/bin /property:BaseIntermediateOutputPath=/home/user/Debug/thrift/lib/netstd/Thrift/obj/
FAILED: lib/netstd/Thrift/bin/Thrift.dll 
cd /home/user/thrift/lib/netstd/Thrift && /usr/bin/dotnet msbuild Thrift.csproj /nologo /restore /target:Rebuild /property:Configuration=Debug /property:OutDir=/home/user/Debug/thrift/lib/netstd/Thrift/bin /property:BaseIntermediateOutputPath=/home/user/Debug/thrift/lib/netstd/Thrift/obj/
MSBUILD : error MSB1025: An internal failure occurred while running MSBuild.
System.ComponentModel.Win32Exception (13): Permission denied
   at System.Diagnostics.Process.set_PriorityClassCore(ProcessPriorityClass value)
   at System.Diagnostics.Process.set_PriorityClass(ProcessPriorityClass value)
   at Microsoft.Build.CommandLine.MSBuildApp.Execute(String[] commandLine)
Unhandled exception. System.ComponentModel.Win32Exception (13): Permission denied
   at System.Diagnostics.Process.set_PriorityClassCore(ProcessPriorityClass value)
   at System.Diagnostics.Process.set_PriorityClass(ProcessPriorityClass value)
   at Microsoft.Build.CommandLine.MSBuildApp.Execute(String[] commandLine)
   at Microsoft.Build.CommandLine.MSBuildApp.Main(String[] args)
```

I've tried to isolate the cause and could find so far:
 - No useful tips or workarounds with Google worked for me, or at least there was nothing that I could understand or map to this problem
 - When running `dotnet` with sudo permissions, the build works. This is not a usable workaround for me because it breaks my build environment in various places
 - My user account has read permissions on all files in `/usr/share/dotnet`. This was tested with reading all files with `grep -rn "Win32Exception" /usr/share/dotnet/ > /dev/zero` which reports no errors
 - Trying to upgrade `dotnet` lists no newer packages available as of today
 - Downgrading dotnet to the previous latest version 3.1.3 solves the problem and the build works. I've downgraded with:
```
sudo apt-get install dotnet-apphost-pack-3.1=3.1.3-1 dotnet-host=3.1.3-1 \
    dotnet-hostfxr-3.1=3.1.3-1 dotnet-runtime-3.1=3.1.3-1 dotnet-sdk-3.1=3.1.201-1 \
    aspnetcore-runtime-3.1=3.1.3-1 aspnetcore-targeting-pack-3.1=3.1.3-1 \
    dotnet-runtime-deps-3.1=3.1.3-1
```

This is an acceptable workaround for me, but sooner or later a fix would be appreciated.

@emmenlau then adds: One thing that might be related: My default process priority on Linux is not zero (as one would typically assume on Linux). I've set it slightly lower (higher integers give lower priority) to `5`. This gives better overall user experience on a multi-user machine.

@dagood then adds:
That stack trace goes (well, appears to go) here in particular:

https://github.com/microsoft/msbuild/blob/7869acfe0be8e983da5de514b896cca6f38fd645/src/MSBuild/XMake.cs#L628-L631

The priority values per enum value seem to be here (not sure where in docs):

https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/Common/src/Interop/Unix/System.Native/Interop.GetSetPriority.cs#L57-L58

So it does appear to be relying on that exact assumption you're talking about for default priority. As far as I can tell (I'm not familiar with this change, just found this issue interesting) it's trying to raise it back to "normal" 0. I'd suggest raising this issue in the MSBuild repo.

Forgind:
This otherwise breaks in linux-like environments when starting MSBuild at below normal priority unless also run with administrator privileges.

Fixes #5365

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -627,7 +627,7 @@ namespace Microsoft.Build.CommandLine
 
                     // Honor the low priority flag, we place our selves below normal
                     // priority and let sub processes inherit that priority.
-                    ProcessPriorityClass priority = lowPriority ? ProcessPriorityClass.BelowNormal : ProcessPriorityClass.Normal;
+                    ProcessPriorityClass priority = lowPriority ? ProcessPriorityClass.BelowNormal : Process.GetCurrentProcess().PriorityClass;
                     Process.GetCurrentProcess().PriorityClass = priority;
 
                     DateTime t1 = DateTime.Now;
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -625,8 +625,9 @@ namespace Microsoft.Build.CommandLine
                         Environment.SetEnvironmentVariable("MSBUILDLOADALLFILESASWRITEABLE", "1");
                     }
 
-                    // Honor the low priority flag, we place our selves below normal
-                    // priority and let sub processes inherit that priority.
+                    // Honor the low priority flag, we place our selves below normal priority and let sub processes inherit
+                    // that priority. Idle priority would prevent the build from proceeding as the user does normal actions.
+                    // We avoid increasing priority because that causes failures on mac/linux.
                     if (lowPriority && Process.GetCurrentProcess().PriorityClass != ProcessPriorityClass.Idle)
                     {
                         Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -627,8 +627,10 @@ namespace Microsoft.Build.CommandLine
 
                     // Honor the low priority flag, we place our selves below normal
                     // priority and let sub processes inherit that priority.
-                    ProcessPriorityClass priority = lowPriority ? ProcessPriorityClass.BelowNormal : Process.GetCurrentProcess().PriorityClass;
-                    Process.GetCurrentProcess().PriorityClass = priority;
+                    if (lowPriority && Process.GetCurrentProcess().PriorityClass != ProcessPriorityClass.Idle)
+                    {
+                        Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
+                    }
 
                     DateTime t1 = DateTime.Now;
 


src/MSBuild/XMake.cs
```diff
@@ -627,7 +627,7 @@ string commandLine
 
                     // Honor the low priority flag, we place our selves below normal
                     // priority and let sub processes inherit that priority.
-                    ProcessPriorityClass priority = lowPriority ? ProcessPriorityClass.BelowNormal : ProcessPriorityClass.Normal;
+                    ProcessPriorityClass priority = lowPriority ? ProcessPriorityClass.BelowNormal : Process.GetCurrentProcess().PriorityClass;
```
bgianfo, line 630: Should we fetch the process object once? (we do it again immediately below?) I imagine this is another TLS lookup for no reason.

bgianfo: It's also odd that we even bother to set the process priority at all if we are only going to set it to it's current value.

Forgind: Principal reason is that I'm a big fan of ternary operators, but I agree it doesn't make sense here. 😉

bgianfo, line 630: Should we also check if we are already running at a priority below ProcessPriorityClass.BelowNormal?

Forgind: That's what I was just looking at...it would be nice if the enum had a monotonic ordering from idle to normal to realtime, but unfortunately that isn't true. Are you familiar with a way to compare priorities between these?

src/MSBuild/XMake.cs
```diff
@@ -627,7 +627,7 @@ string commandLine
...
                     Process.GetCurrentProcess().PriorityClass = priority;
```
bgianfo, line 631: We really shouldn't touch priority unless the lowPriority flag was passed... it  just makes the code ripe for these kinds of bugs.

src/MSBuild/XMake.cs
Forgind: I think BasePriority is monotonic, but it's also readonly and (I think?) system-dependent to some extent, so I think it's best to avoid that.

bgianfo:
Seems like the change @Foritus mentioned in #5372 is more in line with what should be done here?
https://github.com/microsoft/msbuild/compare/master...Foritus:master

Forgind:
> Seems like the change @Foritus mentioned in #5372 is more in line with what should be done here?
> [master...Foritus:master](https://github.com/microsoft/msbuild/compare/master...Foritus:master)

That's almost right and very similar to what I have locally. As I mentioned above, I'm trying to find a way to _compare_ two PriorityClasses. Since it's an enum, people can add values not currently covered, so it's more robust if I can find a clean comparator.

src/MSBuild/XMake.cs
bgianfo: I think you might have to build your own comparator.

bgianfo: You can then add a reflection based test to catch new entries in the enum?

Forgind: There can be unsupported values for the enum that still have effects, so I'd prefer to have an official comparator if one exists. If not, comparing with Idle is best. I'd be hesitant to add a test that would break because of something external. That would be extremely confusing for whichever dev first encounters it unless I add detailed comments.

Forgind: Of course, when you first see a test failure, you don't necessarily look at comments first, so even detailed comments might not really be sufficient.

src/MSBuild/XMake.cs
```diff
@@ -627,8 +627,10 @@ string commandLine
 
                     // Honor the low priority flag, we place our selves below normal
                     // priority and let sub processes inherit that priority.
-                    ProcessPriorityClass priority = lowPriority ? ProcessPriorityClass.BelowNormal : ProcessPriorityClass.Normal;
-                    Process.GetCurrentProcess().PriorityClass = priority;
+                    if (lowPriority && Process.GetCurrentProcess().PriorityClass != ProcessPriorityClass.Idle)
```
rainersigwald, line 630: Explain check with comment please.

tianon:
I think the discussion in https://github.com/dotnet/msbuild/pull/5381#discussion_r431447074 needs to be revisited (and maybe https://github.com/dotnet/msbuild/issues/5365 reopened, as it's not completely fixed by this change?).

As written, this tests whether the `PriorityClass` is not set to exactly `ProcessPriorityClass.Idle`, and only sets it to `BelowNormal` if so, but I think it actually needs to compare against *every* possible `PriorityClass` value that's lower than `BelowNormal`, right?  So (currently), it needs to also take `Lowest` into consideration, since `Lowest` is *not* `Idle`, but *is* lower than `BelowNormal`, so this would still attempt to raise the process priority.

Another alternative could be to instead capture the "permission denied" error and explicitly ignore it?

(I originally commented this with more detail at https://github.com/dotnet/core/issues/4719#issuecomment-663216884, but was redirected here as that issue should've been closed. :sweat_smile:)

Forgind:
Priorities in ProcessPriorityClass do not necessarily increase or decrease with higher priority, so `<` and `>` are out. As far as `Lowest`, I think you were looking at [ThreadPriorityLevel](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.threadprioritylevel?view=netcore-3.1) instead of [ProcessPriorityClass](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processpriorityclass?view=netcore-3.1). ProcessPriorityClass doesn't have a "Lowest" setting currently.

If you're familiar with a good way to compare against every possible lower priority, I'd be happy to incorporate that. (See [here](https://github.com/dotnet/msbuild/pull/5381#discussion_r431447827).) I'd like to avoid explicitly catching the permission denied error, but if this is a pain point, that might be the best choice.

mihsoft:
We've observed this on Ubuntu1804 agents when using VMSS in Azure - https://github.com/actions/virtual-environments/issues/1287

Forgind:
I put a change in #5547.

PR #5386, opened 2020-05-31
Update broken link to RAR target in Microsoft.Common.CurrentVersion.targets

vritant24:


documentation/wiki/ResolveAssemblyReference.md
```diff
@@ -9,7 +9,7 @@ If you notice the ordering, ResolveAssemblyReferences is happening before Compil
 
 ## Source Code
 You can browse Microsoft's MSBuild targets online at:
-http://source.roslyn.io/#MSBuildFiles/C/ProgramFiles(x86)/MSBuild/14.0/bin_/amd64/Microsoft.Common.CurrentVersion.targets,1820
+https://referencesource.microsoft.com/#MSBuildFiles/C/ProgramFiles(x86)/MSBuild/14.0/bin_/amd64/Microsoft.Common.CurrentVersion.targets,1820
```
rainersigwald, line 12: This is the most straightforward fix, but I think we should just point it to the current implementation:

```suggestion
https://github.com/microsoft/msbuild/blob/a936b97e30679dcea4d99c362efa6f732c9d3587/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1991-L2140
```

vritant24:
Looks like I don't have an option to merge. Would you do it?

PR #5391, opened 2020-06-02
Adding condition for default TargetPlatform properties

sfoslund:
Adding a condition for setting default TargetPlatform properties. `_EnableDefaultWindowsPlatform` will be set from the SDK in a separate PR

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -86,8 +86,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetPlatformIdentifier Condition="'$(TargetPlatformIdentifier)' == ''">Windows</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition="'$(TargetPlatformVersion)' == ''">7.0</TargetPlatformVersion>
+    <TargetPlatformIdentifier Condition="'$(TargetPlatformIdentifier)' == '' and '$(_EnableDefaultWindowsPlatform)' != 'false'">Windows</TargetPlatformIdentifier>
+    <TargetPlatformVersion Condition="'$(TargetPlatformVersion)' == '' and '$(_EnableDefaultWindowsPlatform)' != 'false'">7.0</TargetPlatformVersion>
     <TargetPlatformSdkPath Condition="'$(TargetPlatformSdkPath)' == '' and '$(TargetPlatformSdkRootOverride)' != ''">$(TargetPlatformSdkRootOverride)\</TargetPlatformSdkPath>
     <TargetPlatformSdkPath Condition="'$(TargetPlatformSdkPath)' == '' and '$(TargetPlatformIdentifier)' == 'Windows' and '$(OS)' == 'Windows_NT' and '$(MSBuildRuntimeType)' != 'Core'">$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\Software\Microsoft\Microsoft SDKs\Windows\v$(TargetPlatformVersion)', InstallationFolder, null, RegistryView.Registry32, RegistryView.Default))</TargetPlatformSdkPath>
     <TargetPlatformSdkPath Condition="'$(TargetPlatformSdkPath)' == ''">$([Microsoft.Build.Utilities.ToolLocationHelper]::GetPlatformSDKLocation($(TargetPlatformIdentifier), $(TargetPlatformVersion)))</TargetPlatformSdkPath>


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -86,8 +86,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <TargetPlatformIdentifier Condition="'$(TargetPlatformIdentifier)' == '' and '$(_EnableDefaultWindowsPlatform)' != 'false'">Windows</TargetPlatformIdentifier>
+    <TargetPlatformVersion Condition="'$(TargetPlatformVersion)' == '' and '$(_EnableDefaultWindowsPlatform)' != 'false'">7.0</TargetPlatformVersion>
     <TargetPlatformSdkPath Condition="'$(TargetPlatformSdkPath)' == '' and '$(TargetPlatformSdkRootOverride)' != ''">$(TargetPlatformSdkRootOverride)\</TargetPlatformSdkPath>
     <TargetPlatformSdkPath Condition="'$(TargetPlatformSdkPath)' == '' and '$(TargetPlatformIdentifier)' == 'Windows' and '$(OS)' == 'Windows_NT' and '$(MSBuildRuntimeType)' != 'Core'">$([MSBuild]::GetRegistryValueFromView('HKEY_LOCAL_MACHINE\Software\Microsoft\Microsoft SDKs\Windows\v$(TargetPlatformVersion)', InstallationFolder, null, RegistryView.Registry32, RegistryView.Default))</TargetPlatformSdkPath>
     <TargetPlatformSdkPath Condition="'$(TargetPlatformSdkPath)' == ''">$([Microsoft.Build.Utilities.ToolLocationHelper]::GetPlatformSDKLocation($(TargetPlatformIdentifier), $(TargetPlatformVersion)))</TargetPlatformSdkPath>
```
dsplaisted, line 93: Will `GetPlatformSDKLocation` and `GetPlatformSDKDisplayName` (on line 100) handle empty parameters correctly?  Did you do a smoke test that you can still build .NET core projects with these changes and `_EnableDefaultWindowsPlatform` set to false?

dsplaisted: We may want/need to disable setting `UseOSWinMdReferences` on line 97.  We won't be using WinMDs at all on .NET Core.

dsplaisted:
I'd also be interested in hearing if @rainersigwald has a better suggestion for the name than `_EnableDefaultWindowsPlatform`.

src/Tasks/Microsoft.Common.CurrentVersion.targets
sfoslund: Yes, `TargetPlatformDisplayName` just defaults to an empty string and I'm able to build a .NET core console app. 

Would you suggest using the same property (`_EnableDefaultWindowsPlatform`) to disable `UseOSWinMdReferences`?

dsplaisted: Actually, I think we should probably just set `UseOSWinMdReferences` to false in the .NET SDK (possibly conditioned on .NET 5 or higher).

rainersigwald:
I think you could drop the underscore and make this "public", but I don't feel super strongly about it. My biggest concern is unintended consequences. Can you diff the evaluated properties with and without this set (and no other changes) and see if there are any interesting knock-on effects?

sfoslund:
@rainersigwald sure, here's the diff when building a console app:


**With defaults:**
```
TargetPlatformDisplayName = Windows 7.0
TargetPlatformIdentifier = Windows 
TargetPlatformMoniker = Windows,Version=7.0 
TargetPlatformRegistryBase = Software\Microsoft\Microsoft SDKs\Windows
TargetPlatformVersion = 7.0
```


**Without defaults (`_EnableDefaultWindowsPlatform` set to false):**
```
TargetPlatformDisplayName =
TargetPlatformMoniker = ,Version=
TargetPlatformRegistryBase = Software\Microsoft\Microsoft SDKs\
```
`ProjectCapability` item group does not contain `BuildWindowsDesktopTarget`

dsplaisted:
> @dsplaisted do you have strong feelings on dropping or keeping the underscore?

No, I don't feel strongly.  Personally, I would keep it.

Nirmal4G:
@sfoslund @rainersigwald @zivkan

The `TargetPlatformMoniker` should not be set when Id or version is empty. The empty `,Version=` causes failure during build and/or restore in a heavily customised build.

I have recently encountered both NuGet/Home#10423 and `invalid platform version`...

```log
D:\Projects\Work\MSBuild\BuildTasks
> msbuild -v:m BuildTasks.sln
  Determining projects to restore...
C:\Program Files\Microsoft Visual Studio\2022\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(131,5): error : Invalid platform version '.0'. [D:\Projects\Work\MSBuild\BuildTasks\BuildTasks.sln]
```

The code paths that are affected by this empty moniker value are in NuGet's parsing logic...

https://github.com/NuGet/NuGet.Client/blob/8b3be1734d595aa591557ab710cad5c4ce661810/src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs#L225-L231

https://github.com/NuGet/NuGet.Client/blob/8b3be1734d595aa591557ab710cad5c4ce661810/src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs#L266-L272

**Either we should fix the parsing logic to account for empty versions or we should make sure `TPV` and `TFV` shouldn't be empty.**

I can do a PR but I need consensus from the MSBuild/NuGet/SDK teams.

ViktorHofer:
Sorry for replying to a one year old PR but what was the reasoning for adding a default TargetPlatformIdentifier i.e. to netstandard2.0?

PR #5414, opened 2020-06-08
Enable logging instance properties in default text loggers

laurenprinn:
Edits console logger such that various properties can be output in an error or warning message. The properties are determined by an item "outputProperties" defined by the project. Ex:
Program.cs(14,13): error CS0103: The name 'Consol' does not exist in the current context [C:\Users\t-laprin\source\repos\tester_app\tester_app\tester_app.csproj SchemaVersion:2.0 TargetFramework:netcoreapp3.0 ]

I don't have write permission to Link an issue to this, but this would be the issue it should link to: https://github.com/microsoft/msbuild/issues/3015


src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -471,6 +537,17 @@ public override void ErrorHandler(object sender, BuildErrorEventArgs e)
...
             ShowDeferredMessages();
             setColor(ConsoleColor.Red);
+
+            //determine the mapping of properties to output
+            Dictionary<string, string> outputProperties = new Dictionary<string, string>();
+            if (e.BuildEventContext != null)
+            {
+                int nodeId = e.BuildEventContext.NodeId;
+                int projectContextId = e.BuildEventContext.ProjectContextId;
+                TargetFramework_mapping.TryGetValue((nodeId, projectContextId), out outputProperties);
```
Forgind, line 547:
```suggestion
                TargetFramework_mapping.TryGetValue((nodeId, projectContextId), out e.outputProperties);
```

If you set it here like this, you can set it below with:
`e.outputProperties ??= new Dictionary<string, string>();`
and avoid allocating when e.BuildEventContext != null. Also, when is e.BuildEventContext null? Similar examples above.

src/Build/Logging/BaseConsoleLogger.cs
```diff
@@ -29,6 +30,13 @@ internal abstract class BaseConsoleLogger : INodeLogger
         /// </summary>
         private static bool _supportReadingBackgroundColor = true;
 
+        ///<summary>
+        /// TargetFramework_mapping is the map between a (nodeID and project_context_id) to a target framework
+        /// TargetFramework_errorwarning is the map between a target framwork and the number of errors and warnings
+        /// These are used to give more direction as to where an error is happening when multi targetted
+        ///</summary>
+        protected Dictionary<(int nodeId, int contextId), Dictionary<string, string>> TargetFramework_mapping = new Dictionary<(int nodeId, int contextId), Dictionary<string, string>> ();
```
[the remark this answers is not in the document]
laurenprinn: Originally it was necessary because each thread has multiple properties, since it's not just the TargetFrameworks we are printing and each thread could be using different properties. I can just skip the second mapping and build the string originally to simplify it to one map

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -425,13 +428,15 @@ private void ShowErrorWarningSummary(IEnumerable<BuildEventArgs> listToProcess)
...
                 if (!groupByProjectEntryPoint.ContainsKey(key))
-                {
-                    // If there is no bucket create a new one which contains a list of all the errors which
-                    // happened for a given buildEventContext / target
-                    var errorWarningEventListByTarget = new List<BuildEventArgs>();
-                    groupByProjectEntryPoint.Add(key, errorWarningEventListByTarget);
-                }
+                    groupByProjectEntryPoint.Add(key, new List<BuildEventArgs>());
+#else
+                        groupByProjectEntryPoint.TryAdd(key, new List<BuildEventArgs>());
```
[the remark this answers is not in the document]
laurenprinn: TryAdd does not exist on Core for Dictionaries that are serializable. To make it easier to read though I can just have all frameworks do what the Core would need

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +540,66 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            }
+            //creating the value to be added to the TargetFramework_mapping
+            Dictionary<string, string> propertyOutputs = new Dictionary<string, string>();
+            if (e.BuildEventContext != null && e.Items != null)
+            {
+                foreach (DictionaryEntry item in e.Items)
+                {
+                    ITaskItem itemVal = (ITaskItem)item.Value;
+                    //finding if the outputProperties item has been used
+                    if ((string)item.Key == "outputProperties")
```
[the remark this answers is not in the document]
laurenprinn: Because of string interning this would actually do the same thing as the .Equals method because I am not creating two instances, I'm essentially referencing the same instance. I can change it for clarity and consistency with other code though

src/Build/Logging/SerialConsoleLogger.cs
laurenprinn: I hadn't realized that BuildEventContext could be null and then when I had tried running the tests some of the tests had it null which I assume is a rare case but still possible. TryGetValue oddly enough can't take a property or indexer. Otherwise this would be a great idea!

src/Build/Logging/BaseConsoleLogger.cs
```diff
@@ -29,6 +30,13 @@ internal abstract class BaseConsoleLogger : INodeLogger
         /// </summary>
         private static bool _supportReadingBackgroundColor = true;
 
+        ///<summary>
+        /// TargetFramework_mapping is the map between a (nodeID and project_context_id) to a target framework
+        /// TargetFramework_errorwarning is the map between a target framwork and the number of errors and warnings
+        /// These are used to give more direction as to where an error is happening when multi targetted
+        ///</summary>
+        protected Dictionary<(int nodeId, int contextId), StringBuilder> TargetFramework_mapping = new Dictionary<(int nodeId, int contextId), StringBuilder> ();
```
Nirmal4G, line 38:
```suggestion
        protected Dictionary<(int nodeId, int contextId), StringBuilder> propertyOutputMap = new Dictionary<(int nodeId, int contextId), StringBuilder>();
```

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -91,6 +91,7 @@ internal override void ResetConsoleLoggerState()
             errorCount = 0;
             warningCount = 0;
 
+            propertyOutputMap = new Dictionary<(int nodeId, int contextId), StringBuilder>();
```
rainersigwald, line 94: I might go so far as to say that if you're using the serial console logger you should just not get the new behavior. Is that too conservative?

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -323,6 +323,7 @@ public partial class LazyFormattedBuildEventArgs : Microsoft.Build.Framework.Bui
         protected LazyFormattedBuildEventArgs() { }
         public LazyFormattedBuildEventArgs(string message, string helpKeyword, string senderName) { }
         public LazyFormattedBuildEventArgs(string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
+        public System.Text.StringBuilder LogOutputProperties { get { throw null; } set { } }
```
rainersigwald, line 326: Exposing a StringBuilder as public API is a bit funky. What's the goal of this new field? Can you get the same output without it (I expected that based on the `(nodeId, contextId)` lookup).

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +532,58 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            if (e.BuildEventContext != null)
+            {
+                node = e.BuildEventContext.NodeId;
+                projectContextId = e.BuildEventContext.ProjectContextId;
+            }
+            //creating the value to be added to the propertyOutputMap
+            StringBuilder LogOutputProperties = new StringBuilder();
+            if (e.BuildEventContext != null && e.Items != null)
+            {
+                foreach (DictionaryEntry item in e.Items)
```
rainersigwald, line 548: Do you have to foreach here instead of doing a dictionary lookup like `e.Items["LogOutputProperties"]`?

rainersigwald: Ah, I see you do because it's a non-generic `IEnumerable` 😱

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +532,58 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+                        {
+                            LogOutputProperties.Append(itemVal.ItemSpec).Append(":").Append(value).Append(" ");
+                        }
+                    }
+                }
+            }
+            //adding the finished dictionary to propertyOutputMap
+            //this creates a mapping of a specific project/node to a dictionary of property values
+            if (e.BuildEventContext != null)
+                if (!propertyOutputMap.ContainsKey((node, projectContextId)))
```
rainersigwald, line 585: Under what circumstances does `propertyOutputMap` already contain something for this?

src/Shared/EventArgsFormatting.cs
```diff
@@ -257,11 +258,19 @@ int threadId
...
-
+;
             // If the project file was specified, tack that onto the very end.
+            // Check for additional properties that should be outputted as well
             if (projectFile != null && !String.Equals(projectFile, file))
             {
-                format.Append(" [{10}]");
+                if (LogOutputProperties != null && LogOutputProperties.Length > 0)
+                {
+                    format.Append(" [{10}> ").Append(LogOutputProperties).Append("]");
```
rainersigwald, line 268: It feels a bit weird to actually expand something here when everything else is just creating holes for the final `Format` call. Is there an advantage to doing it this way?

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +532,58 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+                    {
+                        //looking for the property value associated with the property key
+                        //Note: the property key is the item value
+                        string value = null;
+                        bool foundProperty = e.GlobalProperties.TryGetValue(itemVal.ItemSpec, out value);
+     
+                        //looking for the property in local properties if it wasn't found in global properties
+                        if (!foundProperty)
+                        {
+                            foreach (DictionaryEntry prop in e.Properties)
```
rainersigwald, line 562: Should the item itself be key-value pairs of things to log, instead of an indirection to a property name, requiring this second lookup?

src/Build/Logging/SerialConsoleLogger.cs
laurenprinn: I think using the serial console logger you could still be multitargetted or using various properties that could help you debug a build. Do you think it would not be as helpful to people or more that it would break something for people?

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
laurenprinn: I can make it internal and it should still be okay. It's necessary though as it stores the properties that get printed inside the BuildEventargs so it wasn't necessary to change any parameters when printing the properties.

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
laurenprinn: the original idea was to make it as user friendly as possible. Do you think that would make it tedious for people who are trying to output multiple properties?

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
laurenprinn: That's a good point. It's an unnecessary check I can remove

src/Shared/EventArgsFormatting.cs
laurenprinn: No, I can change it to be more consistent. My bad.

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +532,57 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+
+            //node and project context ids for the propertyOutputMap key
+            int node = -1;
+            int projectContextId = -1;
+            if (e.BuildEventContext != null)
+            {
+                node = e.BuildEventContext.NodeId;
+                projectContextId = e.BuildEventContext.ProjectContextId;
+            }
+            //creating the value to be added to the propertyOutputMap
```
danmosemsft, line 544: nit: house style is a space after //

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -277,6 +279,58 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+                        //Note: the property key is the item value
+                        string value = null;
+                        bool foundProperties = e.GlobalProperties.TryGetValue(itemVal.ItemSpec, out value);
+
+                        //looking for the property in local properties if it wasn't found in global properties
+                        if (!foundProperties)
+                        {
+                            foreach (DictionaryEntry prop in e.Properties)
+                            {
+
```
danmosemsft, line 312: blank line

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -488,6 +553,17 @@ public override void WarningHandler(object sender, BuildWarningEventArgs e)
             SetErrorsOrWarningsOnCurrentFrame();
             ShowDeferredMessages();
             setColor(ConsoleColor.Yellow);
+
+            //determine the mapping of properties to output
+            StringBuilder LogOutputProperties = new StringBuilder();
```
danmosemsft, line 558: you've repeated these few lines numerous times, can the duplication be eliminated?

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -277,6 +279,58 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            if (e.BuildEventContext != null && e.Items != null)
+            {
+                foreach (DictionaryEntry item in e.Items)
+                {
+                    ITaskItem itemVal = (ITaskItem)item.Value;
+
+                    //finding if the outputProperties item has been used
+                    if (string.Equals((string)item.Key, "LogOutputProperties", StringComparison.OrdinalIgnoreCase))
+                    {
+                        //looking for the property value associated with the property key
```
danmosemsft, line 302: Nit, house comment style is generally imperative ("// Look for ..") rather than descriptive ("// Looking for..")

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +532,57 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            StringBuilder LogOutputProperties = new StringBuilder();
```
danmosemsft, line 545: Not sure of the preference in this repo, but often var is used if the RHS type is unambiguous (specifically if it is a constructor or a cast). So eg., `var logOutputProperties = new StringBuilder();`. It saves reading.

src/Build/Logging/BaseConsoleLogger.cs
```diff
@@ -29,6 +30,11 @@ internal abstract class BaseConsoleLogger : INodeLogger
         /// </summary>
         private static bool _supportReadingBackgroundColor = true;
 
+        ///<summary>
+        /// propertyOutputMap is the map between a (nodeID and project_context_id) to a target framework
+        ///</summary>
+        protected Dictionary<(int nodeId, int contextId), StringBuilder> propertyOutputMap = new Dictionary<(int nodeId, int contextId), StringBuilder> ();
```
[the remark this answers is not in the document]
danmosemsft: I agree, StringBuilder is a temporary helper. It is not good for storage (it is fat for one thing)> It does not belong in collections. Also if you use string, it may be possible to deduplicate (I don't have full context here, but I assume there's few "target framework" strings, so you certainly don't want many copies of the object representing them.

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
Forgind: Personally, I prefer not to use var for anything so I always know where to look if I want to know what type something is, but I know some disagree. When to use it has been on my radar for "things to talk about in a style meeting if we ever have a style meeting."

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +532,57 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+                                    foundProperty = true;
+                                    break;
+                                }
+                            }
+                        }
+
+                        //adding the property key and value pair to the propertyOutputs
+                        if (e.BuildEventContext != null && foundProperty)
+                        {
+                            LogOutputProperties.Append(itemVal.ItemSpec).Append(":").Append(value).Append(" ");
```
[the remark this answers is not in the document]
laurenprinn: It is currently printing everything on one line. Do you mean you think .AppendLine() makes more sense?

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
laurenprinn: Thanks I didn't know this I'll change all of the comments

src/Build/Logging/SerialConsoleLogger.cs
laurenprinn: Yep I'll make a helper method and call that instead

src/Build/Logging/SerialConsoleLogger.cs
laurenprinn: Awesome I didn't know this either, I'll change that in the other comments as well

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
laurenprinn: I haven't seen var being used very often in this repo so I think I may keep it as explicitly defined

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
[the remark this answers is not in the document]
Forgind: Right, for the last one, so
LogOutputProperties.Append(itemVal.ItemSpec).Append(":").AppendLine(value);

Looks cleaner to me

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -277,6 +279,54 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
                     WriteItems(itemList);
                 }
             }
+            // The node and project context ids for the propertyOutputMap key
```
Forgind, line 282: This comment seems misplaced. Down a few lines?

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -277,6 +279,54 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            if (e.BuildEventContext == null || e.Items == null)
```
Forgind, line 283: nit:
```suggestion
            if (e.BuildEventContext == null || e.Items == null || e.Items.Count == 0)
```
?

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -462,6 +512,22 @@ public override void TaskFinishedHandler(object sender, TaskFinishedEventArgs e)
             }
         }
 
+        /// <summary>
+        /// Finds the LogOutPropterty string to be printed in messages
+        /// </summary>
+        /// <param name="e"> the build event where the LogOutPutProperty string will be added</param>
+        internal void PropertyMapping(LazyFormattedBuildEventArgs e)
+        {
+            string logOutputProperties = "";
```
Forgind, line 521: tiny nit:
```suggestion
            string logOutputProperties = string.Empty;
```

danmosemsft: Just a style thing? As these generate the same code.

Forgind: I believe "" allocates a new empty string, whereas all string.Empty calls use the same empty string, so it also saves (a tiny amount of) space.

danmosemsft: I [don't think that's the case](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxNMAfAAgJgIwCwAUFgMwAEu5AwuQN7HlOUVYAs5AsgBQCU9jZkKx4ADOSh5yAXnIBlGAgCWAOwDmAOgCiYAA4wAngG5BQpiICc3SbxNEhAX1PlnZSh044+A+2fNiJKVkAImC7P38rG3DmJyIHIA===)

At the time I worked in this code, we did believe it was the case, and used String.Empty consistently. It may or may not have been true then (~2005). I think it is now a matter of style only.

Forgind: That seems convincing to me. Looking around online, someone suggested that "" could hide a zero-length space, but I don't think we have to worry about that.😄 

It would still be good to match original style, though.

danmosemsft: Yes, for sure. I was used to String.Empty, but in my current project we use "" (although I see it's not in the [style guide](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md))

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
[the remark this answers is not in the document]
laurenprinn: So in the end we actually want all the properties on the same line as it is being printed along with the error/warning message. While .AppendLine looks like cleaner code I think it is easier for the user to read when it is a space

src/Build/Logging/BaseConsoleLogger.cs
```diff
@@ -15,6 +15,7 @@
 using ColorResetter = Microsoft.Build.Logging.ColorResetter;
 using WriteHandler = Microsoft.Build.Logging.WriteHandler;
 using Microsoft.Build.Exceptions;
+using System.Collections.Concurrent;
```
rainersigwald, line 18: ?

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -425,14 +425,13 @@ private void ShowErrorWarningSummary(IEnumerable<BuildEventArgs> listToProcess)
...
                 if (!groupByProjectEntryPoint.ContainsKey(key))
                 {
-                    // If there is no bucket create a new one which contains a list of all the errors which
-                    // happened for a given buildEventContext / target
-                    var errorWarningEventListByTarget = new List<BuildEventArgs>();
-                    groupByProjectEntryPoint.Add(key, errorWarningEventListByTarget);
+                    groupByProjectEntryPoint.Add(key, new List<BuildEventArgs>());
                 }
-
+
```
rainersigwald, line 434: nit:

```suggestion

```

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +534,55 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
                     WriteItems(e, e.Items);
                 }
             }
+
+            if (e.BuildEventContext == null || e.Items == null)
+            {
+                return;
+            }
+            // node and project context ids for the propertyOutputMap key
+            int nodeID = -1;
```
rainersigwald, line 543: These are redundant now, so can you collapse definition/assignment?

But also there's a named constant for this:

```suggestion
            int nodeID = BuildEventContext.InvalidNodeId;
```

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -923,6 +987,9 @@ public override void ErrorHandler(object sender, BuildErrorEventArgs e)
             // Keep track of the number of error events raised 
             errorCount++;
 
+            // Determine the mapping of properties to output
+            PropertyMapping(e);
```
rainersigwald, line 991: If you change the name of the method, it can be self-documenting without the comment:

```suggestion
            PopulatePropertyMap(e);
```

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -914,6 +962,22 @@ public override void TaskFinishedHandler(object sender, TaskFinishedEventArgs e)
...
+        /// </summary>
+        /// <param name="e"> the build event where the LogOutPutProperty string will be added</param>
+        internal void PropertyMapping(LazyFormattedBuildEventArgs e)
+        {
+            string logOutputProperties = "";
+            if (e.BuildEventContext != null)
+            {
+                int nodeId = e.BuildEventContext.NodeId;
+                int projectContextId = e.BuildEventContext.ProjectContextId;
+                propertyOutputMap.TryGetValue((nodeId, projectContextId), out logOutputProperties);
```
rainersigwald, line 976: Under what circumstance would this not return a value? Should that be an internal error instead of silent `string.Empty`?

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -914,6 +962,22 @@ public override void TaskFinishedHandler(object sender, TaskFinishedEventArgs e)
             }
         }
 
+        /// <summary>
+        /// Finds the LogOutPropterty string to be printed in messages
+        /// </summary>
+        /// <param name="e"> the build event where the LogOutPutProperty string will be added</param>
+        internal void PropertyMapping(LazyFormattedBuildEventArgs e)
+        {
+            string logOutputProperties = "";
```
rainersigwald, line 971: nit:

```suggestion
            string logOutputProperties = string.Empty;
```

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +534,55 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+                ITaskItem itemVal = (ITaskItem)item.Value;
+                // Determine if the LogOutputProperties item has been used
+                if (string.Equals((string)item.Key, "LogOutputProperties", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Look for the property value associated with the property key
+                    // Note: the property key is the item value
+                    string value = null;
+                    bool foundProperty = e.GlobalProperties.TryGetValue(itemVal.ItemSpec, out value);
+     
+                    // Look for the property in local properties if it wasn't found in global properties
```
rainersigwald, line 562: Do we want this? Wouldn't it then apply to single-targeted SDK projects, where it doesn't really make sense?

src/Framework/LazyFormattedBuildEventArgs.cs
```diff
@@ -179,6 +180,8 @@ internal override void CreateFromStream(BinaryReader reader, Int32 version)
             }
         }
 
+        internal string LogOutputProperties{ get;  set; }
```
rainersigwald, line 183: I think I'd like to see this called something like `DistinguishingProperties`?

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -153,6 +154,7 @@ public override void BuildFinishedHandler(object sender, BuildFinishedEventArgs
                     //     1 Warning(s)
                     //     0 Error(s)
                     // Don't color the line if it's zero. (Per Whidbey behavior.)
+                    // separate counts by framework
```
rainersigwald, line 157: Is this still done?

src/Framework/LazyFormattedBuildEventArgs.cs
rainersigwald, line 183: I'm not a fan of mutating the event object mid-logging. It looks like it would be possible to keep this in the logger itself and pass it to formatting calls, right? Could we do that instead of adding a mutable field to a public type, even if the field is internal?

laurenprinn: Ya we can. I don't want to change the paramters of the current format methods because that could cause potential breaks elsewhere but I can add another overload method that passes it.

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
laurenprinn: Looking at it now it will always return a value because it is populated in the projectstarted method

src/Framework/LazyFormattedBuildEventArgs.cs
laurenprinn: This property was removed and changed to a parameter that is passed. So this should be good now

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +539,30 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            int projectContextId = e.BuildEventContext.ProjectContextId;
+
+            // Create the value to be added to the propertyOutputMap
+            StringBuilder logOutputProperties = new StringBuilder();
+
+            foreach (DictionaryEntry item in e.Items)
+            {
+                ITaskItem itemVal = (ITaskItem)item.Value;
+                // Determine if the LogOutputProperties item has been used
+                if (string.Equals((string)item.Key, "LogOutputProperties", StringComparison.OrdinalIgnoreCase))
```
cdmihai, line 558: Put item type names as constants in here: [https://github.com/cdmihai/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Shared/Constants.cs#L120](https://github.com/cdmihai/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Shared/Constants.cs#L120)

cdmihai, line 558: `LogOutputProperties` sounds a bit vague, it does not really nail down the property's purpose. I like @rainersigwald's suggestion of `MSBuildProjectConfigurationDescription`

cdmihai:
Wouldn't it be better if the engine computed this string, rather than a specific logger? If loggers are responsible for computing it, then everybody who writes a logger has to duplicate this logic. If the engine computes it, then it's in a single place.

`ProjectStartedEventArgs` can hold the computed string for the project instance it represents. The `LoggingService` can hold the string cache to avoid recomputations and [LogProjectStarted](https://github.com/cdmihai/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Build/BackEnd/Components/Logging/LoggingServiceLogMethods.cs#L554) would compute it.

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +539,30 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+                {
+                    // Add the item value to the string to be printed in error/warning messages
+                    logOutputProperties.Append(" ").Append(itemVal.ItemSpec);
```
cdmihai, line 561: It might be good to allow configuring the separator via a property (like `MSBuildProjectConfigurationDescriptionSeparator`). Default to space if none is set.

rainersigwald: That makes parsing harder so I think I'm against it.

rainersigwald:
> `ProjectStartedEventArgs` can hold the computed string for the project instance it represents. The `LoggingService` can hold the string cache to avoid recomputations and [LogProjectStarted](https://github.com/cdmihai/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Build/BackEnd/Components/Logging/LoggingServiceLogMethods.cs#L554) would compute it.

This is a nice design, I think.

> Taking it even further, it could even be a property of `ProjectInstance` and / or `Project`. This would help all the tools out there to provide this info in their UIs. `ProjectStartedEventArgs` would enable loggers to display it, `ProjectInstance`/`Project` would enable msbuild API users to display it (Visual Studio, QuickBuild, dotnet CLI?, etc).

Maybe, but right now I think I prefer leaving it at the logging level (in the event).

rainersigwald:
Talked this over with @laurenprinn and others. I like pushing it to the LoggingService, but that doesn't solve the whole problem, because it's only easy to do that for the `ProjectStarted` event. When logging, though, we'll really want this for errors/warnings/messages, too. So the logger would still have to preserve/restore it.

We could add it to a deep base class like `BuildEventContext` or `LazyFormattedBuildEventArgs`, but that would potentially spend a fair amount of memory for minimal benefit when using other loggers (like the VS logger).

So let's go with this implementation for now and we can potentially improve it further in the future.

src/Shared/EventArgsFormatting.cs
```diff
@@ -259,9 +305,17 @@ int threadId
             }
 
             // If the project file was specified, tack that onto the very end.
+            // Check for additional properties that should be outputted as well
             if (projectFile != null && !String.Equals(projectFile, file))
             {
-                format.Append(" [{10}]");
+                if (LogOutputProperties != null && LogOutputProperties.Length > 0)
+                {
+                    format.Append(" [{10}>{11}]");
```
rainersigwald, line 313: Still dithering on the `>`, but still haven't come up with a concrete proposal :(

src/Shared/EventArgsFormatting.cs
```diff
@@ -168,7 +213,8 @@ internal static string FormatEventMessage
             int endLineNumber,
             int columnNumber,
             int endColumnNumber,
-            int threadId
+            int threadId,
+            string LogOutputProperties
```
rainersigwald, line 217: Style nit: local variables and method arguments get lower-case names

```suggestion
            string logOutputProperties
```

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -425,14 +431,12 @@ private void ShowErrorWarningSummary(IEnumerable<BuildEventArgs> listToProcess)
...
                 // Check to see if there is a bucket for the warning
+                // If there is no bucket create a new one which contains a list of all the errors which
+                // happened for a given buildEventContext / target
                 if (!groupByProjectEntryPoint.ContainsKey(key))
                 {
-                    // If there is no bucket create a new one which contains a list of all the errors which
-                    // happened for a given buildEventContext / target
-                    var errorWarningEventListByTarget = new List<BuildEventArgs>();
-                    groupByProjectEntryPoint.Add(key, errorWarningEventListByTarget);
+                    groupByProjectEntryPoint.Add(key, new List<BuildEventArgs>());
```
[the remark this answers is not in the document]
laurenprinn: Ya it does. I can revert changes

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -471,11 +475,11 @@ private void ShowErrorWarningSummary(IEnumerable<BuildEventArgs> listToProcess)
                 {
                     if (errorWarningEvent is BuildErrorEventArgs)
                     {
-                        WriteMessageAligned("  " + EventArgsFormatting.FormatEventMessage(errorWarningEvent as BuildErrorEventArgs, showProjectFile), false);
+                        WriteMessageAligned("  " + EventArgsFormatting.FormatEventMessage(errorWarningEvent as BuildErrorEventArgs, showProjectFile, FindLogOutputProperties((BuildErrorEventArgs)errorWarningEvent)), false);
```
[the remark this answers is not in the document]
laurenprinn: I can change the parameters of FindLogOutputProperties so it's not necessary to cast

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -471,11 +475,11 @@ private void ShowErrorWarningSummary(IEnumerable<BuildEventArgs> listToProcess)
                 {
                     if (errorWarningEvent is BuildErrorEventArgs)
```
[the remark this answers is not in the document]
laurenprinn: I changed the parameter of the method to remove the casting as I realized there was no need for the parameter to be as specific as it was

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -535,6 +539,30 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
...
+            if (e.BuildEventContext == null || e.Items == null)
+            {
+                return;
+            }
+            // node and project context ids for the propertyOutputMap key
+            int nodeID = e.BuildEventContext.NodeId;
+            int projectContextId = e.BuildEventContext.ProjectContextId;
+
+            // Create the value to be added to the propertyOutputMap
+            StringBuilder msBuildProjectConfigurationDescription = new StringBuilder();
```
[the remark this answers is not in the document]
laurenprinn: Doesn't seem it's available in netcoreapp2.1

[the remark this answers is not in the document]
rainersigwald: Hmm, works on my machine™:

```suggestion
            using var msBuildProjectConfigurationDescription = new ReuseableStringBuilder();
```

rainersigwald, line 552:
```suggestion
            using var projectInstanceDescription = new ReuseableStringBuilder();
```

src/Build/Logging/SerialConsoleLogger.cs
```diff
@@ -224,12 +224,12 @@ public override void ProjectStartedHandler(object sender, ProjectStartedEventArg
                 this.VerifyStack(contextStack.Peek().type == FrameType.Target, "Bad stack -- Top is project {0}", contextStack.Peek().ID);
             }
 
-            // if verbosity is normal, detailed or diagnostic
+            // If verbosity is normal, detailed or diagnostic
```
rainersigwald, line 227: Revert file (whitespace/casing changes only, no longer touched in current PR).

src/MSBuild.UnitTests/XMake_Tests.cs
```diff
@@ -2169,9 +2169,7 @@ public void BinaryLogContainsImportedFiles()
                 string binLogLocation = testEnvironment.DefaultTestDirectory.Path;
 
                 string output = RunnerUtilities.ExecMSBuild($"\"{testProject.Path}\" \"/bl:{binLogLocation}/output.binlog\"", out var success, _output);
-
                 success.ShouldBeTrue(output);
-
```
rainersigwald, line 2174: Revert file (whitespace changes only, not needed in current PR)

src/Shared/EventArgsFormatting.cs
```diff
@@ -15,6 +15,51 @@ namespace Microsoft.Build.Shared
...
     {
+        /// <summary>
+        /// Format the error event message and all the other event data into
+        /// a single string.
+        /// </summary>
+        /// <param name="e">Error to format</param>
+        /// <param name="showProjectFile"><code>true</code> to show the project file which issued the event, otherwise <code>false</code>.</param>
+        /// <param name="msBuildProjectConfigurationDescription">Properties to Print along with message</param>
+        /// <returns>The formatted message string.</returns>
+        internal static string FormatEventMessage(BuildErrorEventArgs e, bool showProjectFile, string msBuildProjectConfigurationDescription)
```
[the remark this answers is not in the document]
rainersigwald: (and for all overloads)

src/Shared/EventArgsFormatting.cs
```diff
@@ -259,9 +305,17 @@ int threadId
             }
 
             // If the project file was specified, tack that onto the very end.
+            // Check for additional properties that should be outputted as well
```
rainersigwald, line 308:
```suggestion
            // Check for additional properties that should be output as well
```

PR #5415, opened 2020-06-09
Do not check first byte before handshake

Forgind:
Related to dotnet/runtime#37333. This is an alternate (and, perhaps, better) way to resolve the problem.

### DIFF
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -377,7 +377,7 @@ namespace Microsoft.Build.Internal
 #endif
             )
         {
-            return stream.ReadLongForHandshake((byte[])null, 0
+            return stream.ReadLongForHandshake(0
 #if NETCOREAPP2_1 || MONO
                 , handshakeReadTimeout
 #endif
@@ -388,7 +388,7 @@ namespace Microsoft.Build.Internal
         /// Extension method to read a series of bytes from a stream.
         /// If specified, leading byte matches one in the supplied array if any, returns rejection byte and throws IOException.
         /// </summary>
-        internal static long ReadLongForHandshake(this PipeStream stream, byte[] leadingBytesToReject,
+        internal static long ReadLongForHandshake(this PipeStream stream,
             byte rejectionByteToReturn
 #if NETCOREAPP2_1 || MONO
             , int timeout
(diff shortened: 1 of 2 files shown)


src/Shared/NodeEndpointOutOfProcBase.cs
```diff
@@ -384,23 +384,9 @@ private void PacketPumpProc()
...
-                    // The other way around is not a problem, because the old client would immediately return the (wrong)
-                    // byte on receiving the first byte of the long sent by the new host, and the new host would disconnect.
-                    // To avoid the hang, special case here:
-                    // Make sure our handshakes always start with 00.
-                    // If we received ONLY one byte AND it's 0x5F or 0x60, return 0xFF (it doesn't matter what as long as
-                    // it will cause the host to reject us; new hosts expect 00 and old hosts expect F5 or 06).
                     try
                     {
-                        long handshake = localReadPipe.ReadLongForHandshake(/* reject these leads */ new byte[] { 0x5F, 0x60 }, 0xFF /* this will disconnect the host; it expects leading 00 or F5 or 06 */
+                        long handshake = localReadPipe.ReadLongForHandshake(0xFF /* this will disconnect the host; it expects leading 00 or F5 or 06 */
```
rainersigwald, line 389:
```suggestion
                        long handshake = localReadPipe.ReadLongForHandshake(0xFF /* this will disconnect a < 4.5 host; it expects leading 00 or F5 or 06 */
```

src/Shared/CommunicationsUtilities.cs
```diff
@@ -388,7 +384,7 @@ internal static void WriteLongForHandshake(this PipeStream stream, long value)
         /// Extension method to read a series of bytes from a stream.
         /// If specified, leading byte matches one in the supplied array if any, returns rejection byte and throws IOException.
         /// </summary>
-        internal static long ReadLongForHandshake(this PipeStream stream, byte[] leadingBytesToReject,
+        internal static long ReadLongForHandshake(this PipeStream stream,
```
danmosemsft, line 387: Since this handshake has caused some pain, maybe consider in the exceptions (like the one below) including the hex form of the bytes that were received (and possibly what was expected, if it was different)

rainersigwald:
Is this still a good idea or should we just skip it in favor of the "just send components" approach we discussed in #5439?

PR #5416, opened 2020-06-09
Switch default for GenerateFullPaths Fixes #5358

### LINKED ISSUE #5358: GenerateFullPaths should be on default for all projects
### Steps to reproduce

`GenerateFullPaths` property controls whether a variety of targets generate full paths to for the `File` property in their errors, warnings and messages.

Both C#/VB projects (legacy and CPS) work around the fact that this property is set to false, by _after the fact_ modification of the event args to set the full path of the error before reporting to it the Error List/Output windows:

For example, the following code is duplicated 3 times for error, warning and message:

``` C#
        private BuildWarningEventArgs MakeRooted(BuildWarningEventArgs e)
        {
            if (this.TryMakeRooted(e.ProjectFile, e.File, out string rootedPath))
            {
                return new BuildWarningEventArgs(
                    e.Subcategory,
                    e.Code,
                    rootedPath,
                    e.LineNumber,
                    e.ColumnNumber,
                    e.EndLineNumber,
                    e.EndColumnNumber,
                    e.Message,
                    e.HelpKeyword,
                    e.SenderName,
                    e.Timestamp)
                {
                    BuildEventContext = e.BuildEventContext,
                    ProjectFile = e.ProjectFile
                };
            }

            return e;
        }

        private bool TryMakeRooted(string projectFile, string file, out string result)
        {
            if (!string.IsNullOrEmpty(projectFile) && !string.IsNullOrEmpty(file))
            {
                try
                {
                    string basePath = Path.GetDirectoryName(projectFile);

                    result = Path.Combine(basePath, file);
                    return true;
                }
                catch (ArgumentException)
                {
                }
                catch (PathTooLongException)
                {
                }
            }

            result = null;
            return false;
        }
```

This code is problematic because 

1) It increases the number of allocations per message
2) It assumes that msbuild will not derive from, or add any new fields to these events.

This has been attempted to be fixed a number of times from a VS perspective and in AzureDevOps, can we please centralize the fix in MSBuild by flipping this property to true by default both via command-line and VS. 

See: https://github.com/dotnet/project-system/issues/5588 for more context.

Forgind:
Fixes #5358

### DIFF
--- a/src/Tasks/Microsoft.Common.props
+++ b/src/Tasks/Microsoft.Common.props
@@ -175,10 +175,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </PropertyGroup>
   <Import Condition="Exists('$(NuGetPropsFile)')" Project="$(NuGetPropsFile)" />
 
-  <PropertyGroup Condition="'$(GenerateFullPaths)' == ''">
-    <GenerateFullPaths>true</GenerateFullPaths>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(MSBuildLogVerboseTaskParameters)' != 'true' ">
     <DisableLogTaskParameter_ConvertToAbsolutePath_Path>true</DisableLogTaskParameter_ConvertToAbsolutePath_Path>
     <DisableLogTaskParameter_FindUnderPath_OutOfPath>true</DisableLogTaskParameter_FindUnderPath_OutOfPath>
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.Common.props
```diff
@@ -175,6 +175,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </PropertyGroup>
   <Import Condition="Exists('$(NuGetPropsFile)')" Project="$(NuGetPropsFile)" />
 
+  <PropertyGroup Condition="'$(GenerateFullPaths)' == ''">
+    <GenerateFullPaths>true</GenerateFullPaths>
+  </PropertyGroup>
```
rainersigwald, line 180: Please move this to Common.CurrentVersion.targets: that's where we set the defaults for most things like this, and it makes the "user set an explicit value in the project file" concept slightly easier.

rainersigwald:
@davkean can this wait for 16.8? I'm a bit concerned about impact changing it in a late preview.

PR #5421, opened 2020-06-11
Optimize metadata storage for items in immutable ProjectInstance

panopticoncentral:
Although ProjectInstance enforces the immutability flag, it doesn't actually do a lot of optimization of the underlying storage it uses. This is an issue for CPS which heavily uses immutable ProjectInstances to hold snapshots of projects. In large solutions, ProjectInstance snapshots can take significant portions of the heap. In one test solution of 755 projects, 10% of the managed heap was ProjectInstance snapshots.

This tackles one part of the problem. ProjectItemInstances hold the metadata on the item in a copy-on-write dictionary to optimize memory for the sparse write case. However, in an immutable ProjectInstance, there will never be any writing of metadata at all, so the overhead of the copy-on-write scheme is wasted. This PR generalizes the storage of the metadata to hold it in an ImmutableDictionary if the item is immutable. In the above example of 755 projects, this reduces the size of the ProjectInstances in memory by 20% and saves 2% of the overall heap.

This also fixes a bug where the project items weren't being marked as immutable. This is because when the ProjectInstance is being created, it doesn't set the immutable flag until after it has cloned everything. That _seemed_ wrong, but I don't know enough about ProjectInstance creation to change it, so instead I explicitly pass the flag through right now. Would be happy for any comments people who are more familiar with ProjectInstance.

Note that this is just one step along a bigger journey -- there are lots of other places in ProjectInstance where we can likely optimize storage for the immutable case, and we'll continue to work on those.

### DIFF
--- a/src/Build/Collections/CopyOnWritePropertyDictionary.cs
+++ b/src/Build/Collections/CopyOnWritePropertyDictionary.cs
@@ -68,6 +68,26 @@ namespace Microsoft.Build.Collections
         }
 
         /// <summary>
+        /// Creates a dictionary from another dictionary.
+        /// </summary>
+        /// <param name="dictionary">The other dictionary.</param>
+        public CopyOnWritePropertyDictionary(IDictionary<string, T> dictionary)
+        {
+            if (dictionary == null)
+            {
+                _properties = new CopyOnWriteDictionary<string, T>(MSBuildNameIgnoreCaseComparer.Default);
+            }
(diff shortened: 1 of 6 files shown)


panopticoncentral:
Actually, I think this needs to be a bit bigger, so closing for the moment.

rainersigwald:
Hey, just wanted to make sure you're aware of #5279 that I push along every now and then.

PR #5425, opened 2020-06-13
Fix WinFX casing and targets import

Nirmal4G:
Supersedes #5346 

https://github.com/microsoft/msbuild/blob/26f6d1d87700faa8b38d215dbde5968e6b8381f6/src/Tasks/Microsoft.NETFramework.CurrentVersion.targets#L129

https://github.com/microsoft/msbuild/blob/26f6d1d87700faa8b38d215dbde5968e6b8381f6/src/Tasks/Microsoft.Common.CurrentVersion.targets#L6151

These targets point to `WinFX` casing but has `WinFx` casing in the `MSBuildToolsPath`. This will result in file not found, if the targets get imported in a case-sensitive file-system. Since, no one used WinFX targets on Mono, they have not been reported, I think.

Also NETFX 2 has `WinFX` casing but NETFX 4 has `WinFx` casing. So, In order to cope with both `WinFX` and `WinFx`, I renamed the redirection targets to the correct case and then added a workaround which F# team uses to import either ones on a case-sensitive file system.


Nirmal4G:
All Related PRs:
- dotnet/wpf#2975
- dotnet/wpf#2976
- dotnet/sdk#11606

Nirmal4G:
@rainersigwald /cc @ryalanms @vatsan-madhavan @marcpopMSFT

PR #5426, opened 2020-06-14
Log CurrentUICulture in binlog.

KirillOsenkov:
This will be useful to be able to open localized binlogs. If we know the culture of the log we can fetch the right resources from the MSBuild .dlls.

### DIFF
--- a/src/Build/Logging/BinaryLogger/BinaryLogger.cs
+++ b/src/Build/Logging/BinaryLogger/BinaryLogger.cs
@@ -154,6 +154,7 @@ namespace Microsoft.Build.Logging
         private void LogInitialInfo()
         {
             LogMessage("BinLogFilePath=" + FilePath);
+            LogMessage("CurrentUICulture=" + System.Globalization.CultureInfo.CurrentUICulture.Name);
         }
 
         private void LogMessage(string text)


KirillOsenkov:
image

Nirmal4G:
If we are going to have `CurrentUICulture`, why not have `CurrentCulture` too?

KirillOsenkov:
Honestly I don’t know enough about this to know what’s the difference. Do you have a specific usage in mind where knowing CurrentCulture is required?

Nirmal4G:
Same as yours but what happens if the `CurrentUICulture` is not found, we have to fall back to some sensible default. We also have `DefaultThread*` props but those are bit overkill. So, falling back to `CurrentCulture` may not be a bad idea.

> Really what my answer boils down to is consistency and completeness!!

KirillOsenkov:
In my opinion we should be logging enough and just enough. Logging needs a purpose. We've been aggressively trying to *log less* to reduce binlog sizes and cognitive load. I realize it's a small thing, but I'd say let's add this when we need it. For now I need the `CurrentUICulture` to know the language of the binlog so I can properly interpret localized binlogs.

Nirmal4G:
So, will you have `en-US` as the default instead of `CurrentCulture`?

KirillOsenkov:
I'm sorry, perhaps I'm misunderstanding. I have no plans to use CurrentCulture. I only need CurrentUICulture for my purposes.

If you feel like I'm missing your point could you please explain in more detail? What scenario will not work as desired?

Nirmal4G:
What I'm asking is, Will `CurrentUICulture` change depending on OS (Windows Display) language settings. For Example, I have mine in `ta-IN`, but there are no resources for `ta-IN`. So, you have to load `en-US`/`en` by default, right?

KirillOsenkov:
Yes, if the logged culture is not among those supported by MSBuild I will fall back to en-US, because I assume that's what MSBuild would do.

KirillOsenkov:
And yes, the CurrentUICulture updates based on your OS language, and MSBuild uses it to retrieve the resources in that locale:
https://source.dot.net/#Microsoft.Build/Resources/AssemblyResources.cs,76

Nirmal4G:
The `CurrentCulture` is guaranteed to be available even if `UI` is not, since full windows language packages are available for those. So can you fall-back to `CurrentCulture` instead of just `en-US`?

That's (mostly) why I've asked for the `CurrentCulture` to be added as a fall-back.

KirillOsenkov:
I think maybe you're misunderstanding what I'm logging this property for. I don't need to know what CurrentCulture was set to in the MSBuild process. I only need to know what CurrentUICulture was set to, if anything. I'm not interested in any fallbacks. I'm not going to show any UI myself.

The only purpose that I need this information for is what language the strings are in the .binlog. They can only be in one of the languages supported by MSBuild and that's what I care about.

Nirmal4G:
I see. Then what about opening the binlogs in systems where the specified UI language resource was not available?

KirillOsenkov:
MSBuild always ships with a fixed set of languages, on all systems. It doesn't matter on which OS the log viewer is running, it will always have all languages supported by MSBuild available, for example for `de` you will always have the German resources.dll, regardless of the OS:
`C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\de\Microsoft.Build.resources.dll`

Nirmal4G:
We have an internal fork which has `en`, `ta` and other indic languages.

KirillOsenkov:
@Forgind I'm not sure how to test it but according to https://referencesource.microsoft.com/#mscorlib/system/globalization/cultureinfo.cs,811 it can't be null.

Forgind:
I was thinking of trying to set it to a language MSBuild doesn't recognize and seeing what happens, but that's convincing enough for me.

PR #5428, opened 2020-06-15
Add logging for ItemTransformContainingSemicolon_InTaskHost

Forgind:


src/Shared/NodePacketFactory.cs
```diff
@@ -54,7 +59,11 @@ public void DeserializeAndRoutePacket(int nodeId, NodePacketType packetType, ITr
             // PERF: Not using VerifyThrow to avoid boxing of packetType in the non-error case
             if (!_packetFactories.ContainsKey(packetType))
             {
-                ErrorUtilities.ThrowInternalError("No packet handler for type {0}", packetType);
```
rainersigwald, line 57: Can you add code to dump the known types? That might give another clue.

PR #5429, opened 2020-06-15
Target Platform Intrinsic Functions

sfoslund:


sfoslund:
@rainersigwald I've updated the NuGet dependency to get the latest version, which has the target platform changes. Is [this](https://github.com/microsoft/msbuild/pull/5429/files#diff-9876d4d5be1d691af5a98591d78e3d1fR4) the right way to go about it?

sfoslund:
Looks like the test failures are unrelated to this PR. 

@rainersigwald what's the timeline with working to 16.7?

rainersigwald:
Test failures should have been fixed by #5439.

wli3:
All tests passed. Good to go?

sfoslund:
@wli3 see https://github.com/microsoft/msbuild/pull/5429#pullrequestreview-432715609, we're waiting until the fork for 16.7

wli3:
i see. looks like we are in a lock for releases dates. cswinrt need to be in 5.0.0-preivew 7. I will just communicate that it won't be the final form. as in you can just use net5.0-windows. You need to set `<TargetPlatformVersion>` separately.

wli3:
Looks like we need to rush a bit after the branch cut. cswinrt wants to be done in preview 8. 7/15 is the branch cut date. After this is merged, we need to get on the task of parsing out $(TargetPlatformVersion) quickly. Although, it is possible for you to write to code for parsing out $(TargetPlatformVersion) now? So we can just merge the code after 7/2.

wli3:
Get a build of this local msbuild and then swap the content inside stage 0 for you local SDK might work. We can try together

sfoslund:
@wli3, I have the skeleton of the code written and I already did a little bit of testing with this MSBuild change in my stage 0. If you're going to be blocked I can clean it up and test some more tomorrow and put out a draft PR so you can review.

wli3:
That's great! Let's do that

sfoslund:
Once NuGet/NuGet.Client#3479 is merged we'll want to update the NuGet version number to point to the updated implementation, since the one we're using at the moment has known bugs.

wli3:
We could merge it today?

sfoslund:
@wli3 We're going to want to update the NuGet version number once NuGet/NuGet.Client#3479 is merged, so we're still blocked on that. Right now this is implemented on top of the buggy target platform work.

rainersigwald:
@sfoslund I converted this to a draft so we don't look at it in triage until the NuGet thing is resolved. Can you promote it to "ready for review" when that's done and consumed here?

sfoslund:
@rainersigwald I've updated the nuget version so this should be ready to go.

wli3:
@rainersigwald ?

eng/Packages.props
```diff
@@ -18,7 +18,7 @@
     <PackageReference Update="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="1.16.30" />
     <PackageReference Update="Microsoft.Win32.Registry" Version="4.3.0" />
     <PackageReference Update="NuGet.Build.Tasks" Version="$(NuGetBuildTasksVersion)" />
-    <PackageReference Update="NuGet.Frameworks" Version="$(NuGetBuildTasksVersion)" />
+    <PackageReference Update="NuGet.Frameworks" Version="5.7.0-rtm.6710" />
```
Forgind, line 21: What does rtm stand for? And will this always be available?

sfoslund: I think it means release to manufacturing?... (@wli3 is that right?) I don't see why it wouldn't be avaliable.

wli3: ya, "release to manufacturing". In our case it is mostly mean this version is ready for try. And it should be stable.
although remember to remove line once nuget proper insertion is in.

sfoslund: What do you mean proper nuget insertion? We're using reflection here so we won't need insertions into MSBuild.

wli3: ah, i forgot that. you are right

Forgind: What's the fallback if it can't find 5.7.0-rtm-6710?

sfoslund: I don't know that there is a fallback, but this dependency is [only for testing](https://github.com/microsoft/msbuild/pull/5234/files#diff-ea54c5eb95cb47249db3854a7632151eR22). Otherwise we use reflection and try to load the dll from its VS install location or next to the MSBuild dll. This logic is [here](https://github.com/microsoft/msbuild/pull/5234/files#diff-5af5289df4605c1a0a27f11cb43a8d23R27) in the last intrinsic functions PR I made.

rainersigwald, line 21: Can you just update the master NuGet reference on line 4 instead of overriding just this package?

sfoslund:
>  error MSB3243: (NETCORE_ENGINEERING_TELEMETRY=Build) No way to resolve conflict between "Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "Microsoft.Build, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". Choosing "Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" arbitrarily.

@rainersigwald updating the master nuget reference seems to create some version mismatches. Is there a simple way to solve this?

rainersigwald:
@sfoslund add it here:

https://github.com/microsoft/msbuild/blob/fdddb27ebec8f6a51513b6e869f039a32d3a8c39/eng/BootStrapMSBuild.targets#L47-L61

Issue #5423, opened 2020-06-12
ItemTransformContainingSemicolon_InTaskHost test failure


rainersigwald:
Mysterious failure in the test `ItemTransformContainingSemicolon_InTaskHost`

Forgind:
It appears that this failure came about because TaskHostNodes were able to connect to other OutOfProcNodes—their handshakes came to the same value. To make this more confusing, it had to do with the file version hash sometimes having a bit that masked the bit used to indicate if a node was a task host or not, so this only appeared if the file version hash happened to have a 1 in a specific position. Additionally, it only happened if a TaskHost tried to connect to a non-TaskHost node in the first place, so this consistently succeeded in some cases and normally but not always failed in other cases.

PR #5435, opened 2020-06-16
Fix API breaking change from adding SdkResolver IndicateSuccess overload

dsplaisted:
#5269 introduced an API breaking change by adding an overload to the `IndicateSuccess` method of the `SdkResultFactory` abstract class.

This class should only be implemented by MSBuild itself and by implementors of SDK resolvers when they want to mock this class for testing.  So the impact of the break is probably low, but it still seems like a good idea to fix.


dsplaisted:
@Forgind @rainersigwald Any idea why there are test failures for ItemTransformContainingSemicolon_InTaskHost in this PR?

Forgind:
#5423 
The test started failing a lot recently, and we don't know why, although we have a few ideas. Sadly, it often seems to start working when we add logging. I'd be happy to show you what I have so far if you're interested.

dsplaisted:
@Forgind So do you typically just re-run it?

Forgind:
That, wait for me to figure out what's wrong, or help me figure out what's wrong.

PR #5439, opened 2020-06-17
Prevent Task Hosts from communicating with other nodes Fixes #5423

### LINKED ISSUE #5423: ItemTransformContainingSemicolon_InTaskHost test failure

Forgind:
TaskHost bit was sometimes being overwritten by a FileVersionHash with a 1 in its eighth bit. This prevents that from being a problem by switching to xor instead of or.

Fixes #5423.

### DIFF
--- a/src/Build.UnitTests/Utilities_Tests.cs
+++ b/src/Build.UnitTests/Utilities_Tests.cs
@@ -3,19 +3,13 @@
 
 using System;
 using System.Collections;
-using System.Collections.Specialized;
-using System.Text.RegularExpressions;
 
-using Microsoft.Build.Framework;
 using Microsoft.Build.Shared;
-
-
 
 using CommunicationsUtilities = Microsoft.Build.Internal.CommunicationsUtilities;
 using InternalUtilities = Microsoft.Build.Internal.Utilities;
 using InvalidProjectFileException = Microsoft.Build.Exceptions.InvalidProjectFileException;
 using MSBuildApp = Microsoft.Build.CommandLine.MSBuildApp;
-using Project = Microsoft.Build.Evaluation.Project;
 using ProjectCollection = Microsoft.Build.Evaluation.ProjectCollection;
 
 using Toolset = Microsoft.Build.Evaluation.Toolset;
@@ -27,14 +21,21 @@ using XmlElementWithLocation = Microsoft.Build.Construction.XmlElementWithLocati
 using Xunit;
 using System.Collections.Generic;
 using System.IO;
+using Microsoft.Build.Internal;
+using Shouldly;
+using System.Linq;
+using Xunit.Abstractions;
 
 namespace Microsoft.Build.UnitTests
 {
     public class UtilitiesTestStandard : UtilitiesTest
     {
-        public UtilitiesTestStandard()
+        private readonly ITestOutputHelper _output;
+
+        public UtilitiesTestStandard(ITestOutputHelper output)
         {
             this.loadAsReadOnly = false;
+            _output = output;
         }
 
         [Fact]
@@ -62,6 +63,26 @@ namespace Microsoft.Build.UnitTests
             string xmlContents = GetXmlContents(xmlText);
             // Should get XML; note space after x added
             Assert.Equal("<!-- bar; baz; --><!-- bar --><x />", xmlContents);
+        }
+
+        [Fact]
+        public void HandshakesDiffer()
+        {
+            int numHandshakeOptions = (int)Math.Pow(2, Enum.GetNames(typeof(HandshakeOptions)).Length - 1);
+            Dictionary<long, int> handshakes = new Dictionary<long, int>();
+            for (int i = 0; i < numHandshakeOptions; i++)
+            {
+                long nextKey = CommunicationsUtilities.GetHostHandshake((HandshakeOptions)i);
+                if (handshakes.TryGetValue(nextKey, out int collision))
+                {
+                    _output.WriteLine("There was a collision between {0} and {1}.", collision, i);
+                }
+                else
+                {
+                    handshakes.Add(nextKey, i);
+                }
+            }
+            handshakes.Count.ShouldBe(numHandshakeOptions, "two or more combinations of handshake options hashed to the same value");
         }
     }
 
--- a/src/Shared/CommunicationsUtilities.cs
+++ b/src/Shared/CommunicationsUtilities.cs
@@ -330,14 +330,16 @@ namespace Microsoft.Build.Internal
 
             Trace("MSBUILDNODEHANDSHAKESALT=\"{0}\", msbuildDirectory=\"{1}\", nodeType={2}, FileVersionHash={3}", salt, toolsDirectory, nodeType, FileVersionHash);
 
-            //FileVersionHash (32 bits) is shifted 8 bits to avoid session ID collision
-            //nodeType (4 bits) is shifted just after the FileVersionHash
-            //nodeHandshakeSalt (32 bits) is shifted just after hostContext
-            //the most significant byte (leftmost 8 bits) will get zero'd out to avoid connecting to older builds.
-            //| masked out | nodeHandshakeSalt | hostContext |              fileVersionHash             | SessionID
-            //  0000 0000     0000 0000 0000        0000        0000 0000 0000 0000 0000 0000 0000 0000   0000 0000
-            long baseHandshake = ((long)nodeHandshakeSalt << 44) | ((long)nodeType << 40) | ((long)FileVersionHash << 8);
-            return GenerateHostHandshakeFromBase(baseHandshake);
+            // FileVersionHash (32 bits) is shifted 8 bits to avoid session ID collision
+            // HandshakeOptions (5 bits) is shifted just after the FileVersionHash
+            // remaining bits of nodeHandshakeSalt (32 bits truncated to 11) are shifted next
+            //      nodeHandshakeSalt    | HandshakeOptions |             fileVersionHash           | SessionID
+            //  0000 0000 0000 0000 000        0 0000        0000 0000 0000 0000 0000 0000 0000 0000  0000 0000
+            unchecked
+            {
+                ulong baseHandshake = ((ulong)(uint)nodeHandshakeSalt << 45) | ((ulong)(uint)nodeType << 40) | ((ulong)(uint)FileVersionHash << 8);
+                return GenerateHostHandshakeFromBase((long)baseHandshake);
+            }
         }
 
         /// <summary>


src/Build.UnitTests/BackEnd/NodeEndpointInProc_Tests.cs
```diff
@@ -180,6 +182,18 @@ public void ConstructionWithValidHost()
...
         }
 
+        [Fact]
+        public void HandshakesDiffer()
+        {
+            for (int i = 0; i < 32; i++)
+            {
+                for (int j = i + 1; j < 32; j++)
+                {
+                    CommunicationsUtilities.GetHostHandshake((HandshakeOptions)i).ShouldNotBe(CommunicationsUtilities.GetHostHandshake((HandshakeOptions)j));
```
rainersigwald, line 192: nit: pull the `i` computation out of this loop

src/Build.UnitTests/BackEnd/NodeEndpointInProc_Tests.cs
```diff
@@ -180,6 +182,18 @@ public void ConstructionWithValidHost()
                     NodeEndpointInProc.EndpointMode.Asynchronous, _host);
         }
 
+        [Fact]
+        public void HandshakesDiffer()
```
rainersigwald, line 186: Why this file? I would expect something more like `src\Build.UnitTests\Utilities_Tests.cs`

src/Shared/CommunicationsUtilities.cs
```diff
@@ -330,13 +330,12 @@ internal static long GetHostHandshake(HandshakeOptions nodeType)
...
-            //the most significant byte (leftmost 8 bits) will get zero'd out to avoid connecting to older builds.
-            //| masked out | nodeHandshakeSalt | hostContext |              fileVersionHash             | SessionID
-            //  0000 0000     0000 0000 0000        0000        0000 0000 0000 0000 0000 0000 0000 0000   0000 0000
-            long baseHandshake = ((long)nodeHandshakeSalt << 44) | ((long)nodeType << 40) | ((long)FileVersionHash << 8);
+            // FileVersionHash (32 bits) is shifted 8 bits to avoid session ID collision
+            // HandshakeOptions (4 bits) is shifted just after the FileVersionHash
+            // nodeHandshakeSalt (32 bits) is shifted just after HandshakeOptions
+            //     nodeHandshakeSalt    | HandshakeOptions |             fileVersionHash           | SessionID
+            //  0000 0000 0000 0000 0000        0000        0000 0000 0000 0000 0000 0000 0000 0000  0000 0000
+            long baseHandshake = (nodeHandshakeSalt << 44) ^ ((short)nodeType << 40) ^ (FileVersionHash << 8);
```
ladipro, line 338: I'm afraid that the casts to `long` in the original code were important. The result of shifting an `int` left by N bits is still an `int` and all but the lowest 5 bits of N are discarded.

E.g.

```
(nodeHandshakeSalt << 44)
```

is the same as

```
(nodeHandshakeSalt << 12)
```

here. The fact that it's being assigned to a `long` doesn't make a difference.

src/Build.UnitTests/Utilities_Tests.cs
```diff
@@ -361,6 +358,14 @@ public void GetEnvVars()
             Helpers.AssertCollectionsValueEqual(envVars, referenceVars2);
         }
 
+        [Fact]
+        public void HandshakesDiffer()
+        {
+            int numHandshakeOptions = (int)Math.Pow(2, Enum.GetNames(HandshakeOptions.None.GetType()).Length - 1);
+            HashSet<long> handshakeOptions = Enumerable.Range(0, numHandshakeOptions).Select(i => CommunicationsUtilities.GetHostHandshake((HandshakeOptions)i)).ToHashSet();
+            handshakeOptions.Count.ShouldBe(numHandshakeOptions);
```
rainersigwald, line 366:
```suggestion
            handshakeOptions.Count.ShouldBe(numHandshakeOptions, "two or more combinations of handshake options hashed to the same value");
```

?

src/Shared/CommunicationsUtilities.cs
```diff
@@ -330,14 +330,17 @@ internal static long GetHostHandshake(HandshakeOptions nodeType)
...
-            return GenerateHostHandshakeFromBase(baseHandshake);
+            // FileVersionHash (32 bits) is shifted 8 bits to avoid session ID collision
+            // HandshakeOptions (4 bits) is shifted just after the FileVersionHash
+            // nodeHandshakeSalt (32 bits) is xored with the result
+            //  HandshakeOptions |             fileVersionHash           | SessionID
+            //       0000        0000 0000 0000 0000 0000 0000 0000 0000  0000 0000
+            unchecked
+            {
+                ulong baseHandshake = ((ulong)nodeType << 40) | ((ulong)FileVersionHash << 8);
+                baseHandshake ^= (ulong)nodeHandshakeSalt << 32 + nodeHandshakeSalt; // shift and add to give real values for the whole long
```
ladipro, line 341: `+` has higher precedence than `<<` so this is not doing the right thing.

Also:

The reason you're xor'ing the result with the salt is that you can use the full 32 bits of the salt now, is that correct? It introduces potential collisions between the salt the other fields, however. For example:

`nodeType = NodeType.None, FileVersionHash = 0, nodeHandshakeSalt = 256`

and

`nodeType = NodeType.TaskHost, FileVersionHash = 1, nodeHandshakeSalt = 0`

both yield the same baseHandshake of 1099511628032 so this has the same problem as just encapsulating all these integers in a struct and computing its hash. I'd expect this to be undesired and would rather be open to more collisions within the handshake salt but keeping the guarantee that a task host will never be mistaken for a non-task host. What do you think?

Forgind: I should have checked that, but I'd thought all bitwise operators were higher precedence than arithmetic ones. Thanks!

I may have misunderstood what we wanted to do, but I thought that was actually what we were trying for. The salt can still be used to prevent nodes from communicating as long as you don't choose one of the few salts that would cause a collision, and this means that if we have a mistake in the hash computation, we can ask users to set the salt to a specific value, and their problems may magically disappear pending a real fix.

src/Build.UnitTests/Utilities_Tests.cs
```diff
@@ -63,6 +60,23 @@ public void GetTextFromTextNodeWithXmlComment7()
             // Should get XML; note space after x added
             Assert.Equal("<!-- bar; baz; --><!-- bar --><x />", xmlContents);
         }
+
+        [Fact]
+        public void HandshakesDiffer()
+        {
+            int numHandshakeOptions = (int)Math.Pow(2, Enum.GetNames(HandshakeOptions.None.GetType()).Length - 1);
```
rainersigwald, line 67: You can do this at compile time instead

```suggestion
            int numHandshakeOptions = (int)Math.Pow(2, Enum.GetNames(typeof(HandshakeOptions)).Length - 1);
```

src/Build.UnitTests/Utilities_Tests.cs
```diff
@@ -63,6 +60,23 @@ public void GetTextFromTextNodeWithXmlComment7()
...
+            for (int i = 0; i < numHandshakeOptions; i++)
+            {
+                long nextKey = CommunicationsUtilities.GetHostHandshake((HandshakeOptions)i);
+                if (handshakes.TryGetValue(nextKey, out int collision))
+                {
+                    CommunicationsUtilities.Trace("There was a collision between {0} and {1}.", collision, i);
+                }
+                handshakes.Add(nextKey, i);
+            }
+            handshakes.Count.ShouldBe(numHandshakeOptions, "two or more combinations of handshake options hashed to the same value");
```
rainersigwald, line 78: Collect into an array so you can use the helper function with a nice assert?

```suggestion
            handshakes.ShouldBeUnique("two or more combinations of handshake options hashed to the same value");
```

src/Shared/CommunicationsUtilities.cs
rainersigwald: Thinking about this with Monday Brain, I think I'm back on "salt only the non-options bits".

I was thinking that it'd be nice to be able to drop particular bits out of the hash computation. But that doesn't make sense in practice, because the environment variable will be inherited to child processes. So suppose we decided to mask `TaskHost`--we'd need to have one salt to set it up right in the scheduler process, and another to set it up in the TaskHost workers, or we'd get the same kind of cross-node-type communication we're getting with the current buggy code.

Sorry for the waffling here @Forgind.

src/Build.UnitTests/Utilities_Tests.cs
```diff
@@ -63,6 +60,23 @@ public void GetTextFromTextNodeWithXmlComment7()
...
+        public void HandshakesDiffer()
+        {
+            int numHandshakeOptions = (int)Math.Pow(2, Enum.GetNames(HandshakeOptions.None.GetType()).Length - 1);
+            Dictionary<long, int> handshakes = new Dictionary<long, int>();
+            for (int i = 0; i < numHandshakeOptions; i++)
+            {
+                long nextKey = CommunicationsUtilities.GetHostHandshake((HandshakeOptions)i);
+                if (handshakes.TryGetValue(nextKey, out int collision))
+                {
+                    CommunicationsUtilities.Trace("There was a collision between {0} and {1}.", collision, i);
```
rainersigwald, line 74: Why is this `CommunicationsUtilities.Trace` instead of an xunit logging trace? Or for that matter just a failed assertion?

Forgind: I can switch to _output.WriteLine. I don't want a failed assertion because if there are multiple collisions, I want to see them all.

src/Build.UnitTests/Utilities_Tests.cs
Forgind: If I leave the current tracing code, making this a real check would mean duplicating that work, but the current tracing is cleaner in my opinion for reasons outlined below. The real purpose of this check is just to make sure the test fails if there are duplicates so someone knows to look into it.

The various handshakeoptions are obviously going to be unique, including if I made the items in the array tuples, so I'd have to make it an array of the handshakes, and those are hard to understand. You could still reverse-engineer which pair(s) are colliding by looking at the index of the two that are the same, but beyond identifying which pairs are colliding when there can be very similar pairs of handshakes, you'd have to count to the indices to see which handshakeoptions they represent, and that's already dicey with as many options as there are and will get worse if we add more. I think it's doable, but it would make it harder to use.

src/Shared/CommunicationsUtilities.cs
```diff
@@ -330,14 +330,16 @@ internal static long GetHostHandshake(HandshakeOptions nodeType)
...
-            //nodeType (4 bits) is shifted just after the FileVersionHash
-            //nodeHandshakeSalt (32 bits) is shifted just after hostContext
-            //the most significant byte (leftmost 8 bits) will get zero'd out to avoid connecting to older builds.
-            //| masked out | nodeHandshakeSalt | hostContext |              fileVersionHash             | SessionID
-            //  0000 0000     0000 0000 0000        0000        0000 0000 0000 0000 0000 0000 0000 0000   0000 0000
-            long baseHandshake = ((long)nodeHandshakeSalt << 44) | ((long)nodeType << 40) | ((long)FileVersionHash << 8);
-            return GenerateHostHandshakeFromBase(baseHandshake);
+            // FileVersionHash (32 bits) is shifted 8 bits to avoid session ID collision
+            // HandshakeOptions (4 bits) is shifted just after the FileVersionHash
+            // nodeHandshakeSalt (32 bits) is xored with the result
```
rainersigwald, line 335:
```suggestion
            // remaining bits of nodeHandshakeSalt (32 bits truncated to 20) are shifted next
```

?

src/Shared/CommunicationsUtilities.cs
```diff
@@ -330,14 +330,16 @@ internal static long GetHostHandshake(HandshakeOptions nodeType)
...
-            //nodeType (4 bits) is shifted just after the FileVersionHash
-            //nodeHandshakeSalt (32 bits) is shifted just after hostContext
-            //the most significant byte (leftmost 8 bits) will get zero'd out to avoid connecting to older builds.
-            //| masked out | nodeHandshakeSalt | hostContext |              fileVersionHash             | SessionID
-            //  0000 0000     0000 0000 0000        0000        0000 0000 0000 0000 0000 0000 0000 0000   0000 0000
-            long baseHandshake = ((long)nodeHandshakeSalt << 44) | ((long)nodeType << 40) | ((long)FileVersionHash << 8);
-            return GenerateHostHandshakeFromBase(baseHandshake);
+            // FileVersionHash (32 bits) is shifted 8 bits to avoid session ID collision
+            // HandshakeOptions (5 bits) is shifted just after the FileVersionHash
+            // remaining bits of nodeHandshakeSalt (32 bits truncated to 19) are shifted next
```
ladipro, line 335: nit: It's effectively truncated to 11 bits because the most significant byte is masked out in `GenerateHostHandshakeFromBase()`. And I don't think the session ID is necessarily an 8-bit value either.

Considering the above and taking a step back, I'm wondering why we insist on encoding everything in a single `long` instead of defining a proper back- and forward-compatible structure.

```C#
struct Handshake
{
  byte zeroForBackCompat = 0;
  byte lengthForForwardCompat = sizeof(Handshake);
  NodeType nodeType;
  int handshakeSalt;
  int versionHash;
  int sessionId;
  // ...
}
```

rainersigwald: That does make logging on the rejecting side easier too. Should we push this to unblock VS insertions and other PRs and then switch to that?

ladipro: I think it would make sense to merge this PR now and track making further improvements as a lower-pri work item.

rainersigwald: I'd even think about dropping the version _hash_ and just sending the 4 parts of AssemblyFileVersion?

Forgind: @rainersigwald, @ladipro

I put my first draft [here](https://github.com/microsoft/msbuild/compare/master...Forgind:struct-handshake?expand=1).

There are two parts that I think still need fixing:
1) fileVersion is currently a string because the four parts were returned together. @rainersigwald, is there a way to get the parts individually and turn them into ints if necessary, is there a premade way to do that, or do I have to roll my own (probably regex) parser?
2) It's claiming I need to import ValueTuple before I can have `internal IEnumerable<int part, int index)> RetrieveHandshakeComponents()`. Then it's making me say `handshake.RetrieveHandshakeComponents().First().Item1` instead of `handshake.RetrieveHandshakeComponents().First().part` https://github.com/microsoft/msbuild/pull/5270/ made it look like I could just have the valueTuples, and it would work. What do I need to do differently?

ladipro: > fileVersion is currently a string because the four parts were returned together

You should be able to call `Assembly.GetName().Version` to get to the four version parts but it looks like in the current code we use `AssemblyInformationalVersion` which could be an arbitrary string and not necessarily just integers e.g. "16.7.0-dev-20324-01+ee1c9fd0c7e0e43174785a4e2ed177f14f5856a0". Is it maybe too restrictive to have the handshake consist only of integers?

For the new handshake code please consider all these scenarios, both from the client and host perspective:
- The code you are writing trying to connect to a very old MSBuild with an 8-bit handshake.
- The code you are writing trying to connect to a not-so-old MSBuild with a 64-bit handshake.
- The code you are writing trying to connect to a newer MSBuild with an updated `Handshake` structure.

and make sure that the handshake will reliably fail without hanging.

PR #5440, opened 2020-06-17
Fix visual studio applying nodereuse despite MSBUILDDISABLENODEREUSE being set

### LINKED ISSUE #5221: Visual Studio reuses MSBuild process although MSBuild node reuse is disabled
MSBuild Node reuse is disabled via Environment variable (MSBUILDDISABLENODEREUSE=1), but Visual Studio starts the MSBuild child process with /nodeReuse:true.

As we are developing an MSBuild task that is compiled and used in the same project, we are getting file locking issues, because of the kept alive msbuild processes.

-----

Issue created from feedback ticket: [AB#1055938](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1055938)

BenVillalobos:
Fixes #5221 

See https://github.com/microsoft/msbuild/issues/5221#issuecomment-645026616 for details

> After some investigation it looks like MSBuild is launched differently for Design Time Builds compared to regular builds. These builds don't respect the environment variable MSBUILDDISABLENODEREUSE because they explicitly set nodeReuse to true.

### DIFF
--- a/src/Build/BackEnd/BuildManager/BuildParameters.cs
+++ b/src/Build/BackEnd/BuildManager/BuildParameters.cs
@@ -16,6 +16,7 @@ using Microsoft.Build.Internal;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Shared.FileSystem;
 using ForwardingLoggerRecord = Microsoft.Build.Logging.ForwardingLoggerRecord;
+using Microsoft.Build.Utilities;
 
 namespace Microsoft.Build.Execution
 {
@@ -394,7 +395,7 @@ namespace Microsoft.Build.Execution
         public bool EnableNodeReuse
         {
             get => _enableNodeReuse;
-            set => _enableNodeReuse = Environment.GetEnvironmentVariable("MSBUILDDISABLENODEREUSE") == "1" ? false : value;
+            set => _enableNodeReuse = Traits.Instance.DisableNodeReuse ? false : value;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Forgind:
Might someone who does almost exclusively DTBs and wants them to reuse nodes/go fast but wants full builds to not reuse nodes see this as a breaking change? If so, is that a group we want to consider?

src/Build/BackEnd/BuildManager/BuildParameters.cs
```diff
@@ -394,7 +394,7 @@ public bool ResetCaches
         public bool EnableNodeReuse
         {
             get => _enableNodeReuse;
-            set => _enableNodeReuse = value;
+            set => _enableNodeReuse = Environment.GetEnvironmentVariable("MSBUILDDISABLENODEREUSE") == "1" ? false : value;
```
rainersigwald, line 397: Please move `MSBUILDDISABLENODEREUSE` to `shared/traits.cs` and reference it from there, that way we don't have to do an actual environment variable lookup in this setter.

src/Shared/Traits.cs
```diff
@@ -94,6 +94,11 @@ public Traits()
         /// </summary>
         public readonly int LogPropertyTracking = ParseIntFromEnvironmentVariableOrDefault("MsBuildLogPropertyTracking", 0); // Default to logging nothing via the property tracker.
 
+        /// <summary>
+        /// Setting this environment variable to 1 disables the node reuse feature.
+        /// </summary>
+        public readonly bool DisableNodeReuse = Environment.GetEnvironmentVariable("MSBUILDDISABLENODEREUSE") == "1";
```
ladipro, line 100: nit: Would it make sense to call this from `BuildParameters.Initialize()` (line ~909 in BuildParameters.cs) so there's only one place where we read the environment variable?

BenVillalobos: Good call. I'll go ahead and replace every manual check for this flag with this new property.

PR #5447, opened 2020-06-18
Respect transforms in lazy eval

### LINKED ISSUE #5445: Lazy evaluation optimizations don't respect item transforms
`Update` and `Remove` don't respect item transforms at evaluation time; they apply to all items instead of just the ones that were selected by the item transform.

This was the case for `Update` for a long time; `Remove` was working correctly until #5350.

### Steps to reproduce

Project file
```xml
<Project>
  <ItemGroup>
    <Foo Include="C:\*" />
    <Foo Include="x" />

    <!-- <Foo Update="@(Foo->WithMetadataValue('Filename', 'x'))" Bar="Baz" /> -->
    <Foo Remove="@(Foo->WithMetadataValue('Filename', 'x'))" />
  </ItemGroup>

  <Target Name="Go">
    <Message Importance="High"
             Text="Foo=@(Foo->'%(Filename) %(Bar)')" />
  </Target>
</Project>
```

### Expected behavior

Only the `x` item removed/updated

### Actual behavior

```
Go:
  Foo=
```

### Environment data
`msbuild /version` output: `16.7.0-preview-20309-02+d6862eb21`

rainersigwald:
Fixes #5445 by checking to see if an item function is invoked (the captured
expression has subcaptures) before optimizing operations on same-item
captures.

Added regression tests and improved some asserts that didn't really help me when they fired.

fyi @cdmihai, @safern

### DIFF
--- a/src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
@@ -121,6 +121,43 @@ namespace Microsoft.Build.UnitTests.Evaluation
         }
 
         [Fact]
+        public void RemoveRespectsItemTransform()
+        {
+            var content = @"
+                            <i Include='a;b;c' />
+
+                            <i Remove='@(i->WithMetadataValue(`Identity`, `b`))' />
+                            <i Remove='@(i->`%(Extension)`)' /> <!-- should do nothing -->";
+
+            IList<ProjectItem> items = ObjectModelHelpers.GetItemsFromFragment(content, allItems: true);
+
+            ObjectModelHelpers.AssertItems(new[] { "a", "c" }, items);
+        }
+
+        [Fact]
+        public void UpdateRespectsItemTransform()
+        {
+            var content = @"
+                            <i Include='a;b;c' />
+
+                            <i Update='@(i->WithMetadataValue(`Identity`, `b`))'>
+                                <m1>m1_updated</m1>
+                            </i>
+                            <i Update=`@(i->'%(Extension)')`> <!-- should do nothing -->
+                                <m2>m2_updated</m2>
+                            </i>";
+
+            IList<ProjectItem> items = ObjectModelHelpers.GetItemsFromFragment(content, allItems: true);
+
+            ObjectModelHelpers.AssertItems(new[] { "a", "b", "c" }, items,
+                new[] {
+                    new Dictionary<string, string>(),
+                    new Dictionary<string, string> { ["m1"] = "m1_updated" },
+                    new Dictionary<string, string>(),
+                });
+        }
+
+        [Fact]
         public void UpdateShouldPreserveIntermediaryReferences()
         {
             var content = @"
--- a/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
@@ -328,7 +328,7 @@ namespace Microsoft.Build.Evaluation
             /// Is this spec a single reference to a specific item?
             /// </summary>
             /// <returns>True if the item is a simple reference to the referenced item type.</returns>
-            protected static bool ItemspecContainsASingleItemReference(ItemSpec<P, I> itemSpec, string referencedItemType)
+            protected static bool ItemspecContainsASingleBareItemReference(ItemSpec<P, I> itemSpec, string referencedItemType)
             {
                 if (itemSpec.Fragments.Count != 1)
                 {
--- a/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
+++ b/src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
@@ -324,6 +324,10 @@ namespace Microsoft.Build.Evaluation
                 return needToExpandMetadataForEachItem;
             }
 
+            /// <summary>
+            /// Is this spec a single reference to a specific item?
+            /// </summary>
+            /// <returns>True if the item is a simple reference to the referenced item type.</returns>
             protected static bool ItemspecContainsASingleItemReference(ItemSpec<P, I> itemSpec, string referencedItemType)
             {
                 if (itemSpec.Fragments.Count != 1)
@@ -338,6 +342,14 @@ namespace Microsoft.Build.Evaluation
                 }
 
                 if (!itemExpressionFragment.Capture.ItemType.Equals(referencedItemType, StringComparison.OrdinalIgnoreCase))
+                {
+                    return false;
+                }
+
+                // If the itemSpec is a single call to an item function, like @(X->Something(...)), it may get this
+                // far, but shouldn't be treated as a single reference: the item function may return entirely
+                // different results from a bare reference like @(X).
+                if (itemExpressionFragment.Capture.Captures is object)
                 {
                     return false;
                 }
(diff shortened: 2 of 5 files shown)


src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -120,6 +120,39 @@ public void RemoveShouldPreserveIntermediaryReferences(string content)
             ObjectModelHelpers.AssertItems(new string[0], itemsForI2);
         }
 
+        [Fact]
+        public void RemoveRespectsItemTransform()
+        {
+            var content = @"
+                            <i Include='a;b;c' />
+
+                            <i Remove='@(i->WithMetadataValue(`Filename`, `b`))' />";
```
safern, line 129: Curious, why use `Filename` instead of `Identity`?

src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -342,6 +346,14 @@ protected static bool ItemspecContainsASingleItemReference(ItemSpec<P, I> itemSp
                     return false;
                 }
 
+                // If the itemSpec is a single call to an item function, like @(X->Something(...)), it may get this
```
safern, line 349: I guess only item function transforms are impacted by this since I can't do `@(X-> '%(Identity)')` in lazy evaluations?

cdmihai: That's still an item transform. It should be supported in lazy evaluations.
The safest is for this function to reject all transforms, since there's no guarantee the transform would produce the same result as the original item.

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -120,6 +120,39 @@ public void RemoveShouldPreserveIntermediaryReferences(string content)
             ObjectModelHelpers.AssertItems(new string[0], itemsForI2);
         }
 
+        [Fact]
```
cdmihai, line 123: I'd add tests for both types of item transforms:
- item functions `@(foo->function())`
- string templatey thing: `@(foo->'%(filename)')`

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
rainersigwald: I was copy/pasting from the real repro :) `Identity` does make more sense.

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -120,6 +120,43 @@ public void RemoveShouldPreserveIntermediaryReferences(string content)
...
+        {
+            var content = @"
+                            <i Include='a;b;c' />
+
+                            <i Update='@(i->WithMetadataValue(`Identity`, `b`))'>
+                                <m1>m1_updated</m1>
+                            </i>
+                            <i Update=`@(i->'%(Identity)')`>
+                                <m2>m2_updated</m2>
+                            </i>";
```
rainersigwald, line 148: This is the wrong direction; I should make sure it _doesn't_ apply when it shouldn't

ericstj:
cc @mmitche who suggested getting this into Preview7

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -120,6 +120,43 @@ public void RemoveShouldPreserveIntermediaryReferences(string content)
...
         }
 
+        [Fact]
+        public void RemoveRespectsItemTransform()
+        {
+            var content = @"
+                            <i Include='a;b;c' />
+
+                            <i Remove='@(i->WithMetadataValue(`Identity`, `b`))' />
+                            <i Remove='@(i->`%(Extension)`)' /><!-- should do nothing -->";
```
Forgind, line 130: nit:
```suggestion
                            <i Remove='@(i->`%(Extension)`)' /> <!-- should do nothing -->";
```

src/Build.UnitTests/Evaluation/ItemEvaluation_Tests.cs
```diff
@@ -120,6 +120,43 @@ public void RemoveShouldPreserveIntermediaryReferences(string content)
...
+        [Fact]
+        public void UpdateRespectsItemTransform()
+        {
+            var content = @"
+                            <i Include='a;b;c' />
+
+                            <i Update='@(i->WithMetadataValue(`Identity`, `b`))'>
+                                <m1>m1_updated</m1>
+                            </i>
+                            <i Update=`@(i->'%(Extension)')`><!-- should do nothing -->
```
Forgind, line 146: nit:
```suggestion
                            <i Update=`@(i->'%(Extension)')`> <!-- should do nothing -->
```

src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
```diff
@@ -342,6 +346,14 @@ protected static bool ItemspecContainsASingleItemReference(ItemSpec<P, I> itemSp
...
+                // far, but shouldn't be treated as a single reference: the item function may return entirely
+                // different results from a bare reference like @(X).
+                if (itemExpressionFragment.Capture.Captures is object)
+                {
+                    return false;
```
Forgind, line 354: Should this always return false or should it occasionally return true? I think this is only used for performance, so it doesn't matter if it has false negatives now, but I'd like to see if I understand it correctly. Also could be relevant if we start using it elsewhere.

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -243,7 +243,10 @@ public static void AssertItems(string[] expectedItems, IList<TestItem> items, Di
                 expectedItems.ShouldNotBeEmpty();
             }
 
-            for (var i = 0; i < expectedItems.Length; i++)
+            // iterate to the minimum length; if the lengths don't match but there's a prefix match the count assertion below will trigger
```
Forgind, line 246: Could we just move the assert up? Seems simpler.

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -243,7 +243,10 @@ public static void AssertItems(string[] expectedItems, IList<TestItem> items, Di
...
+            int minimumLength = Math.Min(expectedItems.Length, items.Count());
```
Forgind, line 247:
```suggestion
            int minimumLength = Math.Min(expectedItems.Length, items.Count);
```

src/Shared/UnitTests/ObjectModelHelpers.cs
rainersigwald: That was my initial implementation, but it loses a bunch of information in the error message--it's nicer to see the item-level diff if there is one. That helped me chase down a couple of errors in this implementation.

src/Build/Evaluation/LazyItemEvaluator.LazyItemOperation.cs
rainersigwald: It should always return `false`. The question this method answers is "can I just remove/update every item in the group, or do I need to expand the value to match against existing items?"

Renamed it for a bit more clarity there.

Issue #5442, opened 2020-06-17
System.Runtime.InteropServices.COMException during build


kunalspathak:
### Steps to reproduce

Either include a project sample, attach a zipped project, or provide IDE / CLI steps to create the project and repro the behaviour. Example of a project sample:

Project file
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp5.0</TargetFramework>
  </PropertyGroup>
  <PropertyGroup>
    <DebugType>None</DebugType>
    <Optimize>True</Optimize>
  </PropertyGroup>
<PropertyGroup>
    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
  </PropertyGroup>
</Project>

```

Directory contents:
```
/
- helloworld.cs
```

Command line
```
dotnet build -c Release
```
### Expected behavior
Pass

### Actual behavior
```
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166: UNHANDLED EXCEPTIONS FROM PROCESS 11460:
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166: =====================
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166: 6/17/2020 10:06:51 AM
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166: System.Runtime.InteropServices.COMException (0x80004005): Unspecified error (0x80004005 (E_FAIL))
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166:    at System.Runtime.InteropServices.GCHandle.InternalSet(IntPtr handle, Object value)
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166:    at System.Runtime.InteropServices.GCHandle.set_Target(Object value)
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166:    at Microsoft.Build.WeakStringCache.GetOrCreateEntry[T](T internable, Boolean& cacheHit)
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166:    at Microsoft.Build.OpportunisticIntern.WeakStringCacheInterner.TryIntern[T](T candidate, String& interned)
[2020/06/17 10:06:51][INFO] MSBUILD : error MSB4166:    at Microsoft.Build.OpportunisticIntern.WeakStringCacheInterner.InterningToString[T](T candidate)
[...]

PR #5455, opened 2020-06-24
Fix GCHandle race in WeakStringCache.Concurrent

ladipro:
I haven't been able to reproduce #5442 on ARM64 but code inspection revealed a potential race between one thread dereferencing a `GCHandle` and another thread freeing it. The exception in #5442 does not match what I'd expect to see in such scenario but lacking other cues there's a good chance that fixing it will resolve the issue.

The concurrent implementation of `WeakStringCache` runs scavenging in parallel with other threads accessing the cache so we must never assume that an entry fetched from the cache is valid unless it's removed from the cache at the same time. 

Changes in this PR:
- `GetOrCreateEntry()` now always performs two operations on the cache. Previously it was one to three operations depending on whether it was a cache hit or cache miss. With this change hits become slower. One alternative would be wrapping `Scavenge()` with an R/W lock which would have negative impact on concurrency (scavenging would be an exclusive operation and all other threads would wait for it to finish). Another alternative would be per-entry locking which would add memory overhead. We can still do millions of cache hits per second per core with this new code so it's very unlikely to be a bottleneck.
- `_scavengeThreshold` is set to a high number before scavenging starts in order to prevent multiple threads scavenging at the same time. It is not a true lock so races are possible and we are OK with that.

### DIFF
--- a/src/Shared/WeakStringCache.Concurrent.cs
+++ b/src/Shared/WeakStringCache.Concurrent.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Build
             handle.SetString(result);
             if (!_stringsByHashCode.TryAdd(hashCode, handle))
             {
-                // If somebody beat us to it and the handle has not been added, free it.
+                // Another thread has managed to add a handle for the same hash code, so the one we got can be freed.
                 handle.Free();
             }
 
--- a/src/Shared/WeakStringCache.Concurrent.cs
+++ b/src/Shared/WeakStringCache.Concurrent.cs
@@ -28,16 +28,12 @@ namespace Microsoft.Build
         /// <param name="internable">The internable describing the string we're looking for.</param>
         /// <returns>A string matching the given internable.</returns>
         /// <remarks>
-        /// This method performs one operation on the underlying ConcurrentDictionary on cache hit, and two or three operations on cache miss.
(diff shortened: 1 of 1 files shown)


src/Shared/WeakStringCache.Concurrent.cs
```diff
@@ -65,39 +69,37 @@ public WeakStringCache()
...
             result = internable.ExpensiveConvertToString();
 
-            // If the handle is new, we have to add it to the cache. We do it after removing unused handles if our heuristic
-            // indicates that it would be productive. Note that the _capacity field accesses are not protected from races. Being
-            // atomic (as guaranteed by the 32-bit data type) is enough here.
+            // Set the handle to reference the new string and put it in the cache.
+            handle.SetString(result);
+            if (!_stringsByHashCode.TryAdd(hashCode, handle))
+            {
+                // If somebody beat us to it and the handle has not been added, free it.
```
Forgind, line 76: tiny nit:
has now been added?

ladipro: Thanks, yes, the sentence is confusing. I have changed it to
```C#
// Another thread has managed to add a handle for the same hash code, so the one we got can be freed.
```
same as a few lines above where the same pattern is repeated.

PR #5457, opened 2020-06-25
Removed some unnecessary using statements

Forgind:


src/Shared/INodePacketFactory.cs
```diff
@@ -1,10 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
```
ladipro, line 3: nit: Is it intentional to have no empty line between the copyright header and the namespace declaration?

src/Shared/INodePacketHandler.cs
```diff
@@ -1,9 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
```
ladipro, line 3: nit: No empty line between header and `namespace`.

src/MSBuild/NodeEndpointOutOfProcTaskHost.cs
```diff
@@ -1,9 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
```
ladipro, line 3: nit: No empty line between header and `using`.

PR #5458, opened 2020-06-25
Update bindingRedirect versions of Workflow build tasks dlls (dev16)

HongGit:
Fix the issue with the retail version of msbuild.exe.config, so it has a binding redirect for v15 xamlbuildtask.dll, as it does for v4.0.

### DIFF
--- a/src/MSBuild/app.amd64.config
+++ b/src/MSBuild/app.amd64.config
@@ -60,6 +60,7 @@
         <dependentAssembly>
           <assemblyIdentity name="XamlBuildTask" culture="neutral" publicKeyToken="31bf3856ad364e35" />
           <bindingRedirect oldVersion="4.0.0.0" newVersion="16.0.0.0" />
+          <bindingRedirect oldVersion="15.0.0.0" newVersion="16.0.0.0" />
         </dependentAssembly>
 
         <!-- Workaround for crash in C++ CodeAnalysis scenarios due to https://github.com/Microsoft/msbuild/issues/1675 -->
--- a/src/MSBuild/app.amd64.config
+++ b/src/MSBuild/app.amd64.config
@@ -59,8 +59,7 @@
         </dependentAssembly>
         <dependentAssembly>
           <assemblyIdentity name="XamlBuildTask" culture="neutral" publicKeyToken="31bf3856ad364e35" />
-          <bindingRedirect oldVersion="4.0.0.0" newVersion="16.0.0.0" />
-          <bindingRedirect oldVersion="15.0.0.0" newVersion="16.0.0.0" />
+          <bindingRedirect oldVersion="4.0.0.0-16.0.0.0" newVersion="16.0.0.0" />
         </dependentAssembly>
 
         <!-- Workaround for crash in C++ CodeAnalysis scenarios due to https://github.com/Microsoft/msbuild/issues/1675 -->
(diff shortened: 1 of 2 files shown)


BenVillalobos:
@rainersigwald I forget, do we need to run this through an `exp/` branch to make absolutely certain the change works, or is a CI build enough?

src/MSBuild/app.config
```diff
@@ -50,6 +50,7 @@
         <dependentAssembly>
           <assemblyIdentity name="XamlBuildTask" culture="neutral" publicKeyToken="31bf3856ad364e35" />
           <bindingRedirect oldVersion="4.0.0.0" newVersion="16.0.0.0" />
+          <bindingRedirect oldVersion="15.0.0.0" newVersion="16.0.0.0" />
```
rainersigwald, line 53: Should we collapse this to a range, so we remember to update it again in the future?

```suggestion
          <bindingRedirect oldVersion="4.0.0.0-15.0.0.0" newVersion="16.0.0.0" />
```

? Or were there some intermediate ones that shouldn't be redirected?

rainersigwald:
fyi @sphibbs

src/MSBuild/app.config
HongGit: @rainersigwald If I understand this correctly, the 4.0.0.0 applies to .NET framework version, and 15.0.0.0 implies VS version.  I am not sure what the intrinsic implication is if we change that into a range.   

On the newVesion side, do you have any suggestion about that range?  Could we use "16.0.0.0 - *"?  I am concerned that people will forget to update this with newer version of Visual Studio.

Forgind: We also don't want to downgrade someone from, say, 17.x to 16.0.0.0. I'd at least extend it to 15.0.0.0-16.0.0.0, but I think it's better not to extend into the future.

I believe things will break in RPS if they rely on a new version of XamlBuildTask and ignore this, so as long as they don't push it in over RPS failures, I think we're ok.

rainersigwald: `newVersion` can only have a single value, not a range. It needs to be the version that's currently shipping, so leaving it as 16.0.0.0 makes sense.

We prefer this, which will be easy to update and clearly shows "redirect existing versions to the current version". This isn't super duper relevant now, but will make a lot more sense when we bump to `17.0.0.0`.

```suggestion
          <bindingRedirect oldVersion="4.0.0.0-16.0.0.0" newVersion="16.0.0.0" />
```

HongGit:
@rainersigwald could you please approve the PR so I could merge?

Thanks.
Hong

src/MSBuild/app.amd64.config
```diff
@@ -60,6 +60,7 @@
         <dependentAssembly>
           <assemblyIdentity name="XamlBuildTask" culture="neutral" publicKeyToken="31bf3856ad364e35" />
           <bindingRedirect oldVersion="4.0.0.0" newVersion="16.0.0.0" />
+          <bindingRedirect oldVersion="15.0.0.0" newVersion="16.0.0.0" />
```
Forgind, line 63: I'd do the same thing here as you did for the other one.

HongGit: done.

HongGit:
@Forgind @rainersigwald could one of you please help me merge it?  It does not seem that I have the permission to merge it.

Thanks.
Hong

PR #5461, opened 2020-06-25
Use a struct to pass the node handshake to allow for a more complete check

Forgind:
See comments [here](https://github.com/microsoft/msbuild/pull/5439).


src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -351,19 +351,36 @@ private Stream TryConnectToProcess(int nodeProcessId, int timeout, long hostHand
...
+                    throw new InvalidOperationException();
+                }
+
+                // This indicates that we have finished all the parts of our handshake; hopefully the endpoint has as well.
+                nodeStream.WriteIntForHandshake(-0x2a2a2a2a);
+                if (nodeStream.ReadIntForHandshake(0x12
+#if NETCOREAPP2_1 || MONO
+                   , timeout
+#endif
+                    ) != 0x12812812)
```
BenVillalobos, line 381: Same as above

src/Shared/CommunicationsUtilities.cs
```diff
@@ -437,24 +313,18 @@ byte rejectionByteToReturn
...
-                            }
-                        }
-                    }
-
                     bytes[i] = Convert.ToByte(read);
                 }
+
+                if (byteToAccept != 0x00 && byteToAccept != bytes[0])
+                {
+                    stream.WriteIntForHandshake(0x0F0F0F0F);
```
BenVillalobos, line 321: Same as above

src/Shared/CommunicationsUtilities.cs
```diff
@@ -647,4 +527,48 @@ internal static int GetHandshakeHashCode(string fileVersion)
...
+        readonly int fileVersionMinor;
+        readonly int fileVersionBuild;
+        readonly int fileVersionPrivate;
+        readonly int sessionId;
+
+        internal Handshake(HandshakeOptions nodeType)
+        {
+            // We currently use 6 bits of this 32-bit integer. Very old builds will instantly reject any handshake that does not start with F5 or 06; slightly old builds always lead with 00.
+            // This indicates in the first byte that we are a modern build.
+            options = (int)nodeType | 0x01000000;
```
BenVillalobos, line 545: No suggestion here, but this comment is a good example of context.

src/Shared/NodeEndpointOutOfProcBase.cs
```diff
@@ -378,53 +371,67 @@ private void PacketPumpProc()
...
-                        // We will only talk to a host that was started by the same user as us.  Even though the pipe access is set to only allow this user, we want to ensure they
-                        // haven't attempted to change those permissions out from under us.  This ensures that the only way they can truly gain access is to be impersonating the
-                        // user we were started by.
-                        WindowsIdentity clientIdentity = null;
-                        localPipeServer.RunAsClient(delegate () { clientIdentity = WindowsIdentity.GetCurrent(true); });
+                        CommunicationsUtilities.Trace("Writing accept code to parent");
+                        localWritePipe.WriteIntForHandshake(0x39393939);
 
-                        if (clientIdentity == null || !String.Equals(clientIdentity.Name, currentIdentity.Name, StringComparison.OrdinalIgnoreCase))
+                        // To ensure that our handshake and theirs have the same number of bytes, send and receive a magic number indicating EOS. This has to be different from the
```
BenVillalobos, line 400: Thinking out loud: I wonder if it would be worth having "SendEOS" or "sendAccept" style functions that simply writes the magic number, but the name of the function gives it context instead of commenting in multiple places.

Forgind: That's clean and pretty easy. Thanks!

src/Shared/CommunicationsUtilities.cs
```diff
@@ -646,5 +610,10 @@ internal static int GetHandshakeHashCode(string fileVersion)
                 }
             }
         }
+
+        internal static int AvoidEndOfHandshakeSignal(int x)
+        {
+            return x == EndOfHandshakeSignal ? ~x : x;
```
ladipro, line 616: This is clever! Is there any possibility that a non-end-of-handshake value that should legitimately be equal to `EndOfHandshakeSignal` could be confused for another value that should legitimately be equal to `~EndOfHandshakeSignal`?

You are effectively projecting the 2^32 values of an int to 2^32-1 values, reserving one for the end-of-handshake marker. Can it not introduce collisions where two different handshakes would look the same on the wire?

src/Shared/NodeEndpointOutOfProcBase.cs
```diff
@@ -374,77 +365,85 @@ private void PacketPumpProc()
...
+                            int handshakePart = _pipeServer.ReadIntForHandshake(index == 1 ? 0x01 : 0x00 /* this will disconnect a < 16.8 host; it expects leading 00 or F5 or 06. 0x00 is a wildcard */
 #if NETCOREAPP2_1 || MONO
                             , ClientConnectTimeout /* wait a long time for the handshake from this side */
 #endif
                             );
 
-#if FEATURE_SECURITY_PERMISSIONS
-                        WindowsIdentity currentIdentity = WindowsIdentity.GetCurrent();
-#endif
+                            if (handshakePart != part)
```
ladipro, line 384: Is there any chance the various ways how a handshake can fail can be tested with a unit test? Client sends a shorter handshake and it will be detected here, client sends a longer handshake and it will be detected in `ReadEndOfHandshakeSignal` below, etc.

src/Shared/CommunicationsUtilities.cs
Forgind: That's possible, but I think we have to accept that. Most of the values in the handshake (options, fileVersionMajor, fileVersionMinor, etc.) are explicitly non-negative and reasonably small even though they're all ints as far as the Handshake is concerned, so these won't collide in the near future. Also, if I did my math right, the probability of a collision only increases by 2^-63 for each component with a base probability of erroneous collision of 2^-32, which is pretty small.

The bigger worry is if we add a new component that uses all the bits, and it happens to equal the EndOfHandshakeSignal, since then we could have a hang or an erroneous connection. That's what I was trying to avoid.

src/Shared/NodeEndpointOutOfProcBase.cs
Forgind: I think...maybe? The hardest part, I think, would be mocking a pipe. If I can do that, I don't actually have to make handshakes; I can just send my customized ints and see when they disconnect/throw errors.

Of course, that only tests part of what really happens with a handshake. It would be ideal if I can call TryConnectToProcess with a MockProcess, but that sounds like it would be...complicated.

Forgind:
@rainersigwald, since you brought up not connecting with non-administrator processes this morning, and this was very non-obvious, I should mention that I made "is an administrator" a flag in the handshake. I assumed there's no issue with always including the SessionId in the handshake—it's currently only used if if you have admin privileges.

src/Shared/NodeEndpointOutOfProcBase.cs
ladipro: Got it, thanks. I guess it would be nice if the logic of de/serializing the handshake and detecting a match was all abstracted away into its own helper class that would be unit-testable without elaborate mocking but I understand that it's not super easy. It's just a bit unfortunate that the only change made to test code in this PR is a test case deletion.

src/Shared/CommunicationsUtilities.cs
```diff
@@ -365,39 +318,49 @@ internal static void WriteLongForHandshake(this PipeStream stream, long value)
...
+            }
         }
 
         /// <summary>
         /// Extension method to read a series of bytes from a stream.
         /// If specified, leading byte matches one in the supplied array if any, returns rejection byte and throws IOException.
         /// </summary>
-        internal static long ReadLongForHandshake(this PipeStream stream, byte[] leadingBytesToReject,
-            byte rejectionByteToReturn
+        internal static int ReadIntForHandshake(this PipeStream stream, int byteToAccept
```
ladipro, line 357: nit: Typing `byteToAccept` as `int?` would make it slightly more readable (no special meaning of 0x00). Also, the caller already handles the general case of a handshake component mismatch so maybe it should handle the leading byte mismatch as well? It could be done through a helper on `struct Handshake` to keep the two sides together - OR'ing `options` with 0x01... and checking that it is there.

ladipro: Never mind, the leading byte needs to be handled here so we don't wait for more bytes on mismatch 🤦

src/Build/BackEnd/Components/Communications/NodeEndpointOutOfProc.cs
```diff
@@ -50,17 +51,9 @@ internal class NodeEndpointOutOfProc : NodeEndpointOutOfProcBase
...
-            return NodeProviderOutOfProc.GetHostHandshake(_enableReuse, _lowPriority);
-        }
-
-        /// <summary>
-        /// Returns the client handshake for this node endpoint
-        /// </summary>
-        protected override long GetClientHandshake()
-        {
-            return NodeProviderOutOfProc.GetClientHandshake(_enableReuse, _lowPriority);
+            return new Handshake(CommunicationsUtilities.GetHandshakeOptions(taskHost: false, is64Bit: EnvironmentUtilities.Is64BitProcess, nodeReuse: _enableReuse, lowPriority: _lowPriority));
```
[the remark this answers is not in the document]
Forgind: How can you differentiate between a caller that explicitly passes EnvironmentUtilities.Is64BitProcess and a caller that lets it be false?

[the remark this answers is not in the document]
ladipro: All callers that don't explicitly pass `is64Bit` pass `taskHost: true`, which already makes the method handle `is64Bit` internally. I'm simply suggesting the same be done for the `taskHost: false` case:

```
            // ...
            if (taskHost && clrVersion == 0)
            {
                if (taskHostParameters == null)
                {
                    is64Bit = XMakeAttributes.GetCurrentMSBuildArchitecture().Equals(XMakeAttributes.MSBuildArchitectureValues.x64);
                }
                else
                {
                    is64Bit = taskHostParameters[XMakeAttributes.architecture].Equals(XMakeAttributes.MSBuildArchitectureValues.x64);
                }
            }
            else
            {
                    is64Bit = EnvironmentUtilities.Is64BitProcess;
            }
```

BTW, nobody passes `clrVersion`, the parameter can be removed.

src/Shared/CommunicationsUtilities.cs
```diff
@@ -116,6 +116,11 @@ static internal class CommunicationsUtilities
         /// </summary>
         private const int EndOfHandshakeSignal = -0x2a2a2a2a;
 
+        /// <summary>
+        /// The version of the handshake. This should be updated each time the handshake structure is altered.
+        /// </summary>
+        internal const int handshakeVersion = 0x01000000;
```
ladipro, line 122: nit: Would be slightly more elegant to define `handshakeVersion = 0x01` and shift left in the `Handshake` ctor. The way it's written now, if you say "bump the version", someone might do `handshakeVersion = 0x01000001`.

src/Build/BackEnd/Components/Communications/NodeEndpointOutOfProc.cs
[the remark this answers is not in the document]
Forgind: MSBuildTaskHost.csproj and MSBuild.csproj compile CommunicationsUtilities, and neither has access to EnvironmentUtilities, so I'd have to either make a separate class just to provide Is64BitProcess, change EnvironmentUtilities's visibility, make internals visible, or leave it as-is. I don't like options 2 or 3, and 4 is cleaner than 1, so I'll leave it as-is. I can still remove clrVersion, though.

Forgind:
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/264045 succeeded.

PR #5465, opened 2020-06-26
Updating bootstrapper package locator to include a fixed location under MSBuild extensions folder

MSLukeWest:
Right now bootstrapper packages (used by ClickOnce and VS Installer Projects) are discovered in one of two ways:
 - In a default path under "Program Files (x86)\Microsoft SDKs\ClickOnce Bootstrapper\Packages"
 - In customizable paths set under the "Software\Microsoft\GenericBootstrapper\AdditionalPackagePaths" reg key

Both of these involve paths/keys outside the VS install root.  For the new .NET Core packages we want to install them per instance under the MSBuild extensions path.  So this change is adding an entry to the list of folders used during package discovery to include "MSBuild\Microsoft\VisualStudio\BootstrapperPackages".

### DIFF
--- a/src/Tasks/BootstrapperUtil/Util.cs
+++ b/src/Tasks/BootstrapperUtil/Util.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using Microsoft.Build.Shared;
 using Microsoft.Win32;
 
 namespace Microsoft.Build.Tasks.Deployment.Bootstrapper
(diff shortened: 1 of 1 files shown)


MSLukeWest:
@rainersigwald @SvetBonev Please take a look

MSLukeWest:
> Change looks fine to me. Can you confirm that this works with your other changes before we merge? If you need help applying this MSBuild change on top of a VS with your other changes, feel free to reach out.

Yes, I've confirmed this change works combined with my other changes.  I tested by just patching the Microsoft.Build.Tasks.Core assembly.  I tested using both ClickOnce and installer projects, and I tested both when this folder is empty/doesn't exist and when it contains packages.

PR #5470, opened 2020-06-29
Deduplicate and share Regex objects for globbing

### LINKED ISSUE #3668: GlobInfo uses excessive amount of memory
Based on some dump files collected by Svet, we noticed that GlobInfo we gathered from msbuild is bigger than expected, and in some traces, it uses 1.2% of the total heap.

The majority of those objects are RegEx objects.  CPS gathers GlobInfo from every projects, and uses it to watch directory changes.  Because the majority of the .Net Core project uses exactly the same globbing pattern, we hope those RegEx can be reused (or maybe a more efficient globbing matching logic can be used to replace RegEx), and we can have a much reduced memory usage for that structure.

ladipro:
`Regex` is immutable and can be shared between `MSBuildGlob` instances. Analyzing the glob patterns used by real-world projects, it is clear that such sharing would have significant impact on memory. For example, loading the OrchardCore solution in Visual Studio creates ~12k instances of `MSBuildGlob` with only ~100 unique match file expressions.

The PR adds a simple process-wide cache of unique `Regex` objects that have been handed out. Reusing these expensive objects removes the memory pressure (at most one `Regex` per match file expression) and also saves non-trivial time compiling them.

Ballpark numbers:
- The memory cost of one compiled `Regex`: ~30 kB
- The time cost of compiling one `Regex`: ~5 ms

Fixes #3668

### DIFF
--- a/src/Build/Globbing/MSBuildGlob.cs
+++ b/src/Build/Globbing/MSBuildGlob.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Build.Globbing
     /// </summary>
     public class MSBuildGlob : IMSBuildGlob
     {
-        private struct GlobState
+        private readonly struct GlobState
         {
             public string GlobRoot { get; }
             public string FileSpec { get; }
@@ -222,7 +222,7 @@ namespace Microsoft.Build.Globbing
                                 s_regexCache[matchFileExpression] = newRegex;
                             }
                         }
-                        regex = regex ?? newRegex;
+                        regex ??= newRegex;
                     }
                 }
                 return new GlobState(globRoot, fileSpec, isLegalFileSpec, fixedDirectoryPart, wildcardDirectoryPart, filenamePart, matchFileExpression, needsRecursion, regex);
--- a/src/Build/Globbing/MSBuildGlob.cs
+++ b/src/Build/Globbing/MSBuildGlob.cs
@@ -5,6 +5,7 @@
 using System;
 using System.IO;
 using System.Text.RegularExpressions;
+using Microsoft.Build.Collections;
 using Microsoft.Build.Shared;
 
 namespace Microsoft.Build.Globbing
@@ -44,11 +45,15 @@ namespace Microsoft.Build.Globbing
             }
         }
 
+        // Cache of Regex objects that we have created and are still alive.
+        private static WeakValueDictionary<string, Regex> s_regexCache = new WeakValueDictionary<string, Regex>();
+
         private readonly Lazy<GlobState> _state;
 
         internal string TestOnlyGlobRoot => _state.Value.GlobRoot;
         internal string TestOnlyFileSpec => _state.Value.FileSpec;
         internal bool TestOnlyNeedsRecursion => _state.Value.NeedsRecursion;
+        internal Regex TestOnlyRegex => _state.Value.Regex;
 
         /// <summary>
         ///     The fixed directory part.
@@ -198,11 +203,28 @@ namespace Microsoft.Build.Globbing
                         return (normalizedFixedPart, wildcardDirPart, filePart);
                     });
 
-                // compile the regex since it's expected to be used multiple times
-                var regex = isLegalFileSpec
-                    ? new Regex(matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled)
-                    : null;
+                Regex regex = null;
+                if (isLegalFileSpec)
+                {
+                    lock (s_regexCache)
+                    {
+                        s_regexCache.TryGetValue(matchFileExpression, out regex);
+                    }
 
+                    if (regex == null)
+                    {
+                        // compile the regex since it's expected to be used multiple times
+                        Regex newRegex = new Regex(matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled);
+                        lock (s_regexCache)
+                        {
+                            if (!s_regexCache.TryGetValue(matchFileExpression, out regex))
+                            {
+                                s_regexCache[matchFileExpression] = newRegex;
+                            }
+                        }
+                        regex = regex ?? newRegex;
+                    }
+                }
                 return new GlobState(globRoot, fileSpec, isLegalFileSpec, fixedDirectoryPart, wildcardDirectoryPart, filenamePart, matchFileExpression, needsRecursion, regex);
             },
             true);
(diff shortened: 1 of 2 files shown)


src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -198,11 +203,28 @@ public static MSBuildGlob Parse(string globRoot, string fileSpec)
...
+                    }
 
+                    if (regex == null)
+                    {
+                        // compile the regex since it's expected to be used multiple times
+                        Regex newRegex = new Regex(matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled);
+                        lock (s_regexCache)
+                        {
+                            if (!s_regexCache.TryGetValue(matchFileExpression, out regex))
+                            {
```
Forgind, line 221: I continue to be sad about the lack of TryAdd.

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -198,11 +203,28 @@ public static MSBuildGlob Parse(string globRoot, string fileSpec)
...
+                                s_regexCache[matchFileExpression] = newRegex;
+                            }
+                        }
+                        regex = regex ?? newRegex;
```
Forgind, line 225: tiny nit:
```suggestion
                        regex ??= newRegex;
```

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -198,11 +203,28 @@ public static MSBuildGlob Parse(string globRoot, string fileSpec)
...
+                {
+                    lock (s_regexCache)
+                    {
+                        s_regexCache.TryGetValue(matchFileExpression, out regex);
+                    }
 
+                    if (regex == null)
+                    {
+                        // compile the regex since it's expected to be used multiple times
+                        Regex newRegex = new Regex(matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled);
```
Forgind, line 217: How much pressure do you expect to be on the lock? If not too much, you could move the Regex creation into the lock and avoid creating it when it turns out to not be necessary (an edge case, admittedly).

ladipro: This is a public API and the usage pattern can be anything, really, so I wanted to err on the side of caution. Creating `Regex` when it's not really necessary is something that's been happening until now at a large scale so nothing should break if there's still a rare occurrence. On the other hand, holding the lock while compiling the regex could hypothetically regress concurrent scenarios. Btw, similar reason why I'm caching `Regex` and not the entire `MSBuildGlob` even though the latter is also immutable. Saving the small object just doesn't seem worth the potential break. Let me know if I'm overthinking it .. when I see the `public` modifier I tend to expect the worst.

src/Build/Globbing/MSBuildGlob.cs
ladipro: Done, thanks!

src/Build/Globbing/MSBuildGlob.cs
ladipro: In this case we own the dictionary type so we could emulate it but I understand that you're worried not only about convenience but also about performance, i.e. doing it with one dictionary lookup.

src/Build/Globbing/MSBuildGlob.cs
Forgind: This is probably overoptimizing, but rather than locking the DS, you could lock parts of the DS. Looking at our [ConcurrentDictionary](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1900), I see that each bucket has an associated lock, so that dramatically reduces pressure from parallelism. I'm curious why they didn't just make one lock per bucket, but they probably wanted to minimize memory pressure, and this is a reasonable compromise.

In any case, if you were to make a ConcurrentWeakValueDictionary based on ConcurrentDictionary and use that instead, you don't have to lock the whole ds, and you probably wouldn't have to worry about the time it takes to compile a regex or maybe even the whole glob. What do you think?

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -198,11 +203,28 @@ public static MSBuildGlob Parse(string globRoot, string fileSpec)
...
+                if (isLegalFileSpec)
+                {
+                    lock (s_regexCache)
+                    {
+                        s_regexCache.TryGetValue(matchFileExpression, out regex);
+                    }
 
+                    if (regex == null)
+                    {
+                        // compile the regex since it's expected to be used multiple times
```
siprbaum, line 216: As someone complete unfamiliar with the code and just a drive-by-reviewer, I am wondering why this code block inside the `if` doesn't simply create the regex and assigns it uncoditionally to the WeakValueDictionary. From what I undestood, the goal here is to share the regex and IMHO it shouldn't matter which of the diffrent regex instances are used, as long as they both match the same regex pattern and have the same options.

If my understading is correct, this much simpler version for the body of the `if (regex == null)` would also work.

```
Regex regex = new Regex(matchFileExpression, FileMatcher.DefaultRegexOptions | RegexOptions.Compiled);
lock (s_regexCache) {
    s_regexCache[matchFileExpression] = regex;
}
```

I must be missing something obvious. Care to explain?

ladipro: It is true that functionally the code you are proposing would work just as well. Performance-wise, though, if multiple threads create equivalent `Regex` instances, it is better if the first one wins and is used by all threads as opposed to overwriting the dictionary entry (i.e. last one wins) which, in the worst case, could end up with a different instance on each thread, so effectively zero sharing. With the code in this PR, if a thread creates a `Regex` and then finds that it's lost the race, the new `Regex` won't be JIT-compiled and likely collected in a gen0 GC so pretty cheap.

In summary, I believe that the cost of the extra dictionary lookup is negligible compared to the cost of newing up and fully compiling the `Regex` so worth doing here to guarantee that there is at most one `Regex` instance for the given match file expressions.

src/Build/Globbing/MSBuildGlob.cs
ladipro: What you're suggesting makes sense and would indeed help parallelize the code. Looking at the usage pattern of the API in Visual Studio (most calls are made on the same thread and the total number of `Regex` objects is in the order of 100s, even for very large solutions), I don't believe that there is a need for it, though, and making a ConcurrentWeakValueDictionary would not be justified.

src/Build/Globbing/MSBuildGlob.cs
siprbaum: @ladipro Thx for the explanation, it clears things up and I agree with your reasoning.

PR #5483, opened 2020-07-03
[vs16.8 snap] Bring vs16.7 up to date with current master

ladipro:
I don't seem to have permissions to directly fast-forward the branch so opening a PR.


rainersigwald:
Fast-forwarded to this commit, but not before merging to master, so this won't auto-close. Closing manually.

PR #5484, opened 2020-07-03
[vs16.8 snap] Updated branding for 16.8

ladipro:
Same changes as in the last update for 16.7.

Grepping the tree for "16.7", I'm wondering if we still need this override in eng\Versions.props:

```
    <!-- Override Arcade's default VSSDK version with one that supports client enablement.
         Can be removed after Arcade moves up. -->
    <MicrosoftVSSDKBuildToolsVersion>16.7.13</MicrosoftVSSDKBuildToolsVersion>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,13 @@ For more information on MSBuild, see the [MSBuild documentation](https://docs.mi
 
 ### Build Status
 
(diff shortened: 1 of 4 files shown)


ladipro:
Thanks, I'll let you guys merge when the branch opens for 16.8 commits.

rainersigwald:
Arcade still seems to be on an old VSSDK so I think leaving the override is correct.

PR #5488, opened 2020-07-06
Add ability to set a help link for errors/warnings

jmezach:
As discussed in #4029 (see this [comment](https://github.com/microsoft/msbuild/issues/4029#issuecomment-652131756)) I've added the ability to set a help link for both the Error and Warning tasks. This would allow authors of MSBuild targets to issue warnings and/or errors and point users to some page describing more details about the error/warning. Obviously these would need to show up in the IDE somewhere, but I'm hoping that @davkean can take care of that. I think I've done everything that needs to be done on the MSBuild side of things, but this is my first PR here so I'm not entirely sure.

### DIFF
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Build.Framework
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
-        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, string helpLink, params object[] messageArgs) { }
+        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, string helpLink, System.DateTime eventTimestamp, params object[] messageArgs) { }
         public string Code { get { throw null; } }
         public int ColumnNumber { get { throw null; } }
         public int EndColumnNumber { get { throw null; } }
@@ -120,7 +120,7 @@ namespace Microsoft.Build.Framework
         public BuildWarningEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName) { }
         public BuildWarningEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp) { }
         public BuildWarningEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
-        public BuildWarningEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string helpLink, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
+        public BuildWarningEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, string helpLink, System.DateTime eventTimestamp, params object[] messageArgs) { }
         public string Code { get { throw null; } }
         public int ColumnNumber { get { throw null; } }
         public int EndColumnNumber { get { throw null; } }
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Build.Framework
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
-        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string helpLink, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
+        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, string helpLink, params object[] messageArgs) { }
         public string Code { get { throw null; } }
         public int ColumnNumber { get { throw null; } }
         public int EndColumnNumber { get { throw null; } }
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -16,11 +16,13 @@ namespace Microsoft.Build.Framework
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
+        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string helpLink, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
         public string Code { get { throw null; } }
         public int ColumnNumber { get { throw null; } }
         public int EndColumnNumber { get { throw null; } }
         public int EndLineNumber { get { throw null; } }
         public string File { get { throw null; } }
+        public string HelpLink { get { throw null; } }
         public int LineNumber { get { throw null; } }
         public string ProjectFile { get { throw null; } set { } }
         public string Subcategory { get { throw null; } }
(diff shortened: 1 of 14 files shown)


jmezach:
Looks like the macOS build on Mono is failing due to a single test failure. Looking at the test it doesn't seem related to my changes, so not sure what's going on there. The Windows Full Release build seems to be failing due to mismatches with the reference assemblies, which makes sense since I've expanded the existing API. What would be the right course of action here though? Should we also changes the reference assemblies?

src/Tasks/Error.cs
```diff
@@ -31,6 +31,11 @@ public sealed class Error : TaskExtension
         /// </summary>
         public string HelpKeyword { get; set; }
 
+        /// <summary>
+        /// A link pointing to more information about the error
+        /// </summary>
+        public string HelpLink { get; set; }
```
rainersigwald, line 37: Shouldn't these be passed through in the call to `Log.LogError`?

src/Framework/BuildErrorEventArgs.cs
```diff
@@ -199,25 +243,30 @@ public string ProjectFile
...
         /// <summary>
-        /// Ending column number of interest in associated file. 
+        /// Ending column number of interest in associated file.
         /// </summary>
         public int EndColumnNumber => endColumnNumber;
 
+        /// <summary>
+        /// A link pointing to more information about the error.
+        /// </summary>
+        public string HelpLink => helpLink;
```
rainersigwald, line 268: I would prefer this be named `Url` or `HelpUrl`, but it sounds like `HelpLink` is already in use at the project system layer (https://github.com/microsoft/msbuild/issues/4029#issuecomment-450045271), so it's probably better to be consistent. So 👍

src/Shared/TaskLoggingHelper.cs
```diff
@@ -739,14 +774,14 @@ public void LogErrorWithCodeFromResources(string messageResourceName, params obj
...
-        /// Logs an error using the specified resource string and other error details. 
+        /// Logs an error using the specified resource string and other error details.
         /// If the message has an error code prefixed, the code is extracted and logged with the message. If a
         /// help keyword prefix has been provided, a help keyword for the host IDE is also logged with the message. The help
         /// keyword is composed by appending the error message resource string name to the prefix.
-        /// 
+        ///
         /// A task can provide a help keyword prefix either via the Task (or TaskMarshalByRef) base class constructor, or the
         /// Task.HelpKeywordPrefix (or AppDomainIsolatedTask.HelpKeywordPrefix) property.
-        ///
```
rainersigwald, line 749: Would it be easy for you to revert these unrelated whitespace fixes?

src/Tasks/Warning.cs
```diff
@@ -31,6 +31,11 @@ public sealed class Warning : TaskExtension
         /// </summary>
         public string HelpKeyword { get; set; }
 
+        /// <summary>
+        /// A link pointing to more information about the warning
+        /// </summary>
+        public string HelpLink { get; set; }
```
rainersigwald, line 37: As above, should this be used?

src/Framework/BuildErrorEventArgs.cs
```diff
@@ -264,6 +315,8 @@ internal override void CreateFromStream(BinaryReader reader, int version)
             columnNumber = reader.ReadInt32();
             endLineNumber = reader.ReadInt32();
             endColumnNumber = reader.ReadInt32();
+
+            helpLink = reader.ReadByte() == 0 ? null : reader.ReadString();
```
rainersigwald, line 319: This should be guarded by a `version` check.

The version is derived here

https://github.com/microsoft/msbuild/blob/d98fde4795add51613433cb5def9c7cbd09f4f15/src/Shared/LogMessagePacketBase.cs#L105-L119

And so we'll have 3 possibilities:

1. .NET Core (some high version, all versions guaranteed to match, can assume it's in the packet)
2. .NET 4.7.2 (version `40`, latest code, can assume it's in the packet)
3. .NET 3.5 (version `20`, message was serialized by _MSBuild assemblies from the .NET Framework_, will not have the field in the packet)

So the deserialization code (here) needs to check `version` before reading the byte, as is done with `projectFile`.

rainersigwald:
> The Windows Full Release build seems to be failing due to mismatches with the reference assemblies, which makes sense since I've expanded the existing API. What would be the right course of action here though? Should we also changes the reference assemblies?

If you do a build locally, it should update the reference source with the additions you've made. Add those to your PR.

src/Tasks/Error.cs
jmezach: Yes, I already thought that I missed something somewhere. Fixed that now :).

src/Framework/BuildErrorEventArgs.cs
jmezach: Yeah, I picked that name because it was mentioned in that issue. I can change it of course, but it made sense to me to use what was suggested, rather than re-inventing the wheel ;).

src/Tasks/Warning.cs
jmezach: You're absolutely right ;).

src/Shared/TaskLoggingHelper.cs
jmezach: Looks like my editor (VS Code) is actively removing these. Not entirely sure how to fix that to be honest, but I'll check.

src/Framework/BuildErrorEventArgs.cs
jmezach: I've added a `if (version => 40)` check which seems to be the best approach based on the above information.

jmezach:
I did a local build, but I'm not seeing any local changes after that. Could it be that this only works on Windows perhaps? I'm no a Mac. I do have an Azure VM ready to go though, so I'll give it a shot there.

src/Shared/TaskLoggingHelper.cs
jmezach: Since I was on Windows with Visual Studio I managed to fix these. Let me know if this is okay now.

jmezach:
Looks like the build still breaks, although I'm not sure it is actually related to my changes. Any idea what's going on here?

rainersigwald:
> Looks like the reference assemblies are only updated when building on Windows.

Sorry, I thought we had fixed that! Thanks for switching OSes :)

> Looks like the build still breaks, although I'm not sure it is actually related to my changes. Any idea what's going on here?

```
Assert.Equal() Failure\nExpected: Success\nActual: Failure
   at Microsoft.Build.UnitTests.BackEnd.BuildManager_Tests.OutOfProcProjectInstanceBasedBuildDoesNotReloadFromDisk(Boolean shouldSerializeEntireState) in /home/vsts/work/1/s/src/Build.UnitTests/BackEnd/BuildManager_Tests.cs:line 3847
```

With the build log

```
Build started.
Project "tmpN9NolS.tmp" (Foo target(s)):
Building with tools version "Current".
Target "Foo" in file "/tmp/h345wyeo.ceo/Temporaryd7adbbaa40ea4788bbecc7aa53d1c216/import" from project "/tmp/h345wyeo.ceo/tmpN9NolS.tmp" (entry point):
Using "Message" task from assembly "Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
Task "Message"
Bar
Done executing task "Message".
Done building target "Foo" in project "tmpN9NolS.tmp".
Done building project "tmpN9NolS.tmp".
Build succeeded.
Build started.
(0,0):  error MSB4223: A node of the required type OutOfProc could not be created.
Build succeeded.
```

Should definitely not be you. I'm rerunning that leg. @Forgind, recognize this?

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -16,11 +16,13 @@ public partial class BuildErrorEventArgs : Microsoft.Build.Framework.LazyFormatt
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp) { }
         public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
+        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string helpLink, string senderName, System.DateTime eventTimestamp, params object[] messageArgs) { }
```
rainersigwald, line 19: I'm torn between the current approach (which is most logical IMO) and this, which would be more likely to keep existing code working if someone adds the link. Anyone else have an opinion?

```suggestion
        public BuildErrorEventArgs(string subcategory, string code, string file, int lineNumber, int columnNumber, int endLineNumber, int endColumnNumber, string message, string helpKeyword, string senderName, string helpLink, System.DateTime eventTimestamp, params object[] messageArgs) { }
```

jmezach:
Added the tests as instructed.

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
Forgind: I'm not sure I understand why flipping the order changes whether something will keep working. For either case, someone would add a new parameter, just before the senderName or after it, so the user would have to make sure they get it right. Unless I'm misunderstanding the problem, I'd have a weak preference for keeping helpKeyword and helpLink together.

jmezach: I think I'm with @Forgind on this. I don't see much value in swapping the arguments, especially since they are both `string`'s. As a user of this API you need to be aware of the parameter order anyway.

davkean: Putting API review hat on, @rainersigwald suggestion is exactly what we would say, you should maintain parameter position. Refactoring by adding/removing an argument would result in a very hard to see bug. We maintain parameter order across APIs in the framework.

jmezach: Fair enough, I can understand that. I guess that applies to the changes in `TaskLoggingHelper`as well then? I won't mind changing this.

jmezach: I do think it has to be after `eventTimestamp` though, since `messageArgs` is a params parameter and has to come last right?

davkean: Yes, params last.

jmezach: I've made the changes for Error now. Just want to make sure that this is the right direction before I change Warning as well. It feels weird to me, but maybe I'm just missing something ;).

davkean: It feels weird because?

jmezach: Because something that used to be part of the messageArgs array now becomes the helpLink as far as I can tell.

Forgind: > part of the messageArgs array now becomes the helpLink as far as I can tell.

Can messageArgs start with a string? If so, wouldn't this silently change how any code with a string as the first messageArg works?

rainersigwald: @terrajobst can we borrow a few minutes for a micro–API-change review here?

My read on the `params` binding is that @Forgind is right to worry ([sharplab demo of binding to the with-a-`string` overload over the `params` overload](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyDMABGgQMIEDeOB1RhyALAQPIAUYAdsAQB5QEDKwAE4cA5gQDOfAA4BDIbIC2E5gCMAVgFMAxsADaAXQIB7CQEpyAXyo18RRqw5deBOQuVqtuwyfNWb1AG09gQAsixmFEE0zCwADGYA3NE0KdSsqHwARACexllJOGkEAPQlrOjZeVnZEsaKmjUEWW5KEgXJ2DExQdbYlkA===)). But that wouldn't be a problem with putting it at the end of the `string`s.

jmezach:
Is there anything I can do to move this along?

Forgind:
> Is there anything I can do to move this along?

I think we're currently waiting for @terrajobst to chime in. I can ping him offline.

jmezach:
Is this still waiting for an API review from @terrajobst?

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
terrajobst: Your interpretation is correct. C# rules for `params` goes something like this: if there is an applicable overload without expanding params, it will be called. I suggest keeping `eventTimestamp` last as a tie breaker:

```diff
 public BuildErrorEventArgs(/* ..., */ string senderName);
 public BuildErrorEventArgs(/* ..., */ string senderName, System.DateTime eventTimestamp);
 public BuildErrorEventArgs(/* ..., */ string senderName, System.DateTime eventTimestamp, params object[] messageArgs);
+public BuildErrorEventArgs(/* ..., */ string senderName, string helpLink, System.DateTime eventTimestamp, params object[] messageArgs);
```

Yes, this means that `eventTimestamp` changes position, but in my opinion that change isn't all that different of overloads were we keep `ref` last, for instance.

jmezach: Okay, I'll make the necessary changes some time today hopefully ;).

jmezach:
I think this is ready to go now, but feel free to review it again. Seems like the build isn't triggering again. Not sure why that is ;).

src/Framework/BuildErrorEventArgs.cs
```diff
@@ -264,6 +315,15 @@ internal override void CreateFromStream(BinaryReader reader, int version)
             columnNumber = reader.ReadInt32();
             endLineNumber = reader.ReadInt32();
             endColumnNumber = reader.ReadInt32();
+
+            if (version >= 40)
+            {
+                helpLink = reader.ReadByte() == 0 ? null : reader.ReadString();
```
Forgind, line 321: nit:
Any reason not to crunch the conditionals together? Like:
`helpLink = version >= 40 && reader.ReadByte() != 0 ? reader.ReadString() : null;`
or
```
if (version >= 40 && reader.ReadByte() != 0)
{
    helpLink = reader.ReadString();
}
else
{
    helpLink = null;
}
```

? Same below.

jmezach: Not really, just tried to keep it in line with a similar construct a little higher up for the projectFile field. I don't mind changing it though, but I figured keeping in the same style was more important ;).

jmezach:
Are we ready to merge this?

davkean:
It looks like BuildMessageEventArgs/CriticalBuildMessageEventArgs didn't get a HelpLink. Was that intentional?

jmezach:
@davkean Not really. I just approached it from the `Error` and `Warning` tasks and took things from there. What are those types used for exactly? I did think of adding an `Info` or `Information` task so that you could issues those messages from MSBuild targets as well, but I thought it would be better to do that as a separate PR if there's interest in such a thing.

davkean:
@jmezach Critical Messages can be logged inside a task without needing a specific Info task. We treat error/warning/messages the same for HelpKeyword, and had to handle HelpLink differently because its not on the base.

jmezach:
@davkean I don't mind moving things to a base class, but as far as I can see the only common base class for all those types (`BuildWarningEventArgs`, `BuildErrorEventArgs`, `BuildMessageEventArgs` and `CriticalBuildMessageEventArgs`) is `LazyFormattedBuildEventArgs`. Should we add the HelpLink there?

davkean:
Filed the following documentation bugs to get this tracked:

https://github.com/MicrosoftDocs/visualstudio-docs/issues/5662
https://github.com/MicrosoftDocs/visualstudio-docs/issues/5663

PR #5490, opened 2020-07-06
Port fix for Kudu to not call PriorityClass unless low priority build is specified

marcpopMSFT:
Customer scenario

Kudu and Linux scenarios where the api to set the priority is disabled are blocked from using msbuild.  When lowpriority build is not set, we were still previously setting priority to Normal which was not allowed on some system.

Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal

Bugs this fixes

https://github.com/microsoft/msbuild/issues/5365

Workarounds, if any

None

Risk

Low, we moved the set within a check for is lowpriority build enabled.  The fix is NOT required for VS.

Performance impact

Low

Root cause analysis

We didn't realize this win32 api was blocked on some systems

How was the bug found?

https://github.com/dotnet/core/issues/4719

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -34,6 +34,7 @@
   <!-- Toolset Dependencies -->
   <PropertyGroup>
     <DotNetCliVersion>3.1.100</DotNetCliVersion>
+    <DropAppVersion>18.165.29912-buildid11693003</DropAppVersion>
   </PropertyGroup>
   <Target Name="OverrideArcadeFileVersion" AfterTargets="_InitializeAssemblyVersion">
     <!-- See https://github.com/dotnet/arcade/issues/3386
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -15,10 +15,9 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
   <PropertyGroup>
-    <VersionPrefix>16.6.0</VersionPrefix>
+    <VersionPrefix>16.6.1</VersionPrefix>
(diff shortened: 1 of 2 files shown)


eng/Versions.props
```diff
@@ -15,10 +15,9 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
   <PropertyGroup>
-    <VersionPrefix>16.6.0</VersionPrefix>
+    <VersionPrefix>16.6.1</VersionPrefix>
     <AssemblyVersion>15.1.0.0</AssemblyVersion>
-    <DotNetFinalVersionKind>release</DotNetFinalVersionKind>
```
rainersigwald, line 20: Put this line back before merging

eng/Versions.props
```diff
@@ -15,10 +15,9 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
   <PropertyGroup>
-    <VersionPrefix>16.6.0</VersionPrefix>
+    <VersionPrefix>16.6.1</VersionPrefix>
```
marcpopMSFT, line 18: @rainersigwald isn't this technically 16.6.2 since June would have been 16.6.1 of VS?

rainersigwald: We usually diverge our versions from VS: VS 16.5.5 shipped MSBuild 16.5.1, for instance.

rainersigwald:
Based on the current plan, I'm going to close this in favor of a second version bump so that we can ship the QFE version and then roll a new official version for later.

PR #5499, opened 2020-07-09
Introduce $(IncrementalCleanDependsOn)

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/4905
Fixes: https://github.com/microsoft/msbuild/issues/3916

Xamarin.Android runs MSBuild targets after `CoreCompile` and `Build`
that generates files in `$(IntermediateOutputPath)`. If you want to
add these files to `@(FileWrites)`, the build-in functionality from the
`IncrementalClean` target doesn't work unless you do something like:

    <Target Name="Foo" BeforeTargets="_CleanGetCurrentAndPriorFileWrites">

The problem here being `_CleanGetCurrentAndPriorFileWrites` is "private".

Or modify `$(CoreBuildDependsOn)`:

    <CoreBuildDependsOn>
      $([MSBuild]::Unescape($(CoreBuildDependsOn.Replace('IncrementalClean;', 'Foo;IncrementalClean;'))))
    </CoreBuildDependsOn>

This seems weird to me, and it could also break if a target was named
`FooIncrementalClean` or semi-colons moved around...

If we introduced `$(IncrementalCleanDependsOn)`, we could just do this
instead:

    <IncrementalCleanDependsOn>
      Foo;
      $(IncrementalCleanDependsOn);
    </IncrementalCleanDependsOn>

### DIFF
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -4922,7 +4922,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     -->
   <Target
        Name="IncrementalClean"
-       DependsOnTargets="_CleanGetCurrentAndPriorFileWrites">
+       DependsOnTargets="$(IncrementalCleanDependsOn);_CleanGetCurrentAndPriorFileWrites">
 
     <!-- Subtract list of files produced in prior builds from list of files produced in this build. -->
     <ItemGroup>


src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4920,9 +4920,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     Leave the Clean cache file containing only the files produced in the current build.
     ============================================================
     -->
+  <PropertyGroup>
+    <IncrementalCleanDependsOn>_CleanGetCurrentAndPriorFileWrites</IncrementalCleanDependsOn>
```
Forgind, line 4924: Do you think it would be better to have an if-not-defined condition here? Makes it easier to hook into.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -4920,9 +4920,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+  </PropertyGroup>
   <Target
        Name="IncrementalClean"
-       DependsOnTargets="_CleanGetCurrentAndPriorFileWrites">
+       DependsOnTargets="$(IncrementalCleanDependsOn)">
```
rainersigwald, line 4928: Another option would be to make this

```suggestion
       DependsOnTargets="$(IncrementalCleanDependsOn);_CleanGetCurrentAndPriorFileWrites">
```

So that it's purely additive and we don't have to deal with the possibility of someone accidentally emptying/overwriting `$(IncrementalCleanDependsOn)`. I think I prefer that. Anybody want to talk me out of it?

Forgind: Only downside I see is if they already have _CleanGetCurrentAndPriorFileWrites in the list or if they don't want it there for some reason. In any case: sounds good to me.

PR #5503, opened 2020-07-10
Remove TargetResult compression

### LINKED ISSUE #5443: Consider disabling compression of target results
The code in `TargetResults.CompressItems()` and `TargetResults.DecompressItems()` may be hurting performance rather than helping. On the compression side we translate all task items into a `MemoryStream`, convert the stream to a `byte` array, then compress the `byte` array using a `DeflateStream`, then convert this stream to a `byte` array to finally subject it to the rest of the serialization logic to send it cross process. Similarly on the receiving side, we make a bunch of allocations and burn CPU to get to the original decompressed items. It's very likely that it would be faster and less memory intensive to simply send the uncompressed items.

We should:
- Measure the performance impact of compression with a microbenchmark.
- Remove it if it doesn't prove helpful.

ostorc:
Fixes #5443

For reason discussed in the linked issue this PR removes the compression from TargetResult.

I have removed entire nested ItemStore class, because it's only purpose was to encapsulate the logic behind compression. Without the compression it only wrapped an array of TaskItems.


Forgind:
Thanks for doing this! Do you have any perf results?

ostorc:
I haven’t measured impact of deletion of this code, but it should be similar to the perf result in the issue. But measuring it again, seems like a good idea. Will do it

Forgind:
I actually missed that you'd mentioned it in the issue, but that looks good to me, assuming the test you did with no compression in the issue was equivalent to this change.

ostorc:
So the performance measurement results are same or slightly better than these in the issue (as expected, because I deleted some additional logic)
```
COMPRESSION - 100 items
---------
AVG: 17.8602 ms

COMPRESSION - 1000 items
---------
AVG: 22.7986 ms

COMPRESSION - 10000 items
---------
AVG: 47.2003 ms

NO COMPRESSION - 100 items
---------
AVG: 15.6751 ms

NO COMPRESSION - 1000 items
---------
AVG: 19.1399 ms

NO COMPRESSION - 10000 items
---------
AVG: 51.6685 ms

NO COMPRESSION - ACTUAL - 100 items
---------
AVG: 15.7737 ms

NO COMPRESSION - ACTUAL - 1000 items
---------
AVG: 17.1248 ms

NO COMPRESSION - ACTUAL - 10000 items
---------
AVG: 51.1904 ms
```

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -887,6 +888,7 @@ private void HeaderReadComplete(IAsyncResult result)
                 }
 
                 int packetLength = BitConverter.ToInt32(_headerByte, 1);
+                MSBuildEventSource.Log.PacketReadSize(packetLength);
 
                 byte[] packetData;
                 if (packetLength < _smallReadBuffer.Length)
```
ladipro, line 894: nit/pre-existing: This should be `<=` as both are lengths so no need to subtract one.

ostorc:
Compressed
- 100 items - 374 bytes 
- 1000 items - 2 703 bytes 
- 10000 items - 25 766 bytes 

Without compression
- 100 items - 7 975 bytes 
- 1000 items - 81 076 bytes
- 1000 items - 830 077 bytes 

Note that my items are all same (except the name), so the compression here should be super effective. If needed will try it with more "natural" files

src/Build/Instance/ProjectItemInstance.cs
```diff
@@ -1596,16 +1596,34 @@ internal static TaskItem FactoryForDeserialization(ITranslator translator, Looka
...
+            }
+
             /// <summary>
             /// Reads or writes the task item to the translator using an interner for metadata.
             /// </summary>
             internal void TranslateWithInterning(ITranslator translator, LookasideStringInterner interner)
             {
-                translator.Translate(ref _includeEscaped);
-                translator.Translate(ref _includeBeforeWildcardExpansionEscaped);
+                translator.Translate(ref _itemDefinitions, ProjectItemDefinitionInstance.FactoryForDeserialization);
```
ladipro, line 1615: Great catch! It's surprising that not translating these fields wasn't causing any issues.

src/Build/BackEnd/Shared/TargetResult.cs
```diff
@@ -284,6 +279,46 @@ private void RetrieveItemsFromCache()
...
+                using var itemsStream = new MemoryStream(capacity);
+                var itemTranslator = BinaryTranslator.GetWriteTranslator(itemsStream);
+
+                foreach (var taskItem in _items)
+                {
+                    taskItem.TranslateWithInterning(itemTranslator, interner);
+                }
+
+                interner.Translate(translator);
+                var itemsBuffer = itemsStream.ToArray();
```
ladipro, line 317: nit: It should be possible to eliminate this `byte[]` allocation. For example by calling `GetBuffer()` instead of `ToArray()` and passing the length of the array to `Translate` explicitly.

ostorc: Had to add new method on interface, to support this. Done

src/Build/Instance/ProjectItemInstance.cs
```diff
@@ -1624,6 +1642,14 @@ internal void TranslateWithInterning(ITranslator translator, LookasideStringInte
                 }
                 else
                 {
+                    string fileName = string.Empty, directory = string.Empty;
+                    ReadInternString(translator, interner, ref fileName);
+                    ReadInternString(translator, interner, ref directory);
+
+                    _includeEscaped = Path.Combine(directory, fileName);
```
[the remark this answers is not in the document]
ostorc: I see. Reverted this change.

src/Build/Instance/ProjectItemInstance.cs
ostorc: Tests were passing because the previous implementation was modifying the original instance. It looks like these fields are not really used...

src/Build/BackEnd/Shared/TargetResult.cs
```diff
@@ -284,6 +279,46 @@ private void RetrieveItemsFromCache()
...
+                var itemTranslator = BinaryTranslator.GetWriteTranslator(itemsStream);
+
+                foreach (var taskItem in _items)
+                {
+                    taskItem.TranslateWithInterning(itemTranslator, interner);
+                }
+
+                interner.Translate(translator);
+                var itemsBuffer = itemsStream.GetBuffer();
+                translator.Translate(ref itemsBuffer, itemsBuffer.Length);
```
ladipro, line 318: Should be `itemsStream.Length` instead of `itemsBuffer.Length`.

src/Shared/BinaryTranslator.cs
```diff
@@ -201,6 +201,20 @@ public void Translate(ref byte[] byteArray)
                 }
             }
 
+            /// <summary>
+            /// Translates a byte array
+            /// </summary>
+            /// <param name="byteArray">The array to be translated.</param>
+            /// <param name="length">The length of array which will be used in translation. This parameter is not used when reading</param>
+            /// <remarks>
+            /// This method should be used only for serilization, not for reading from stream.
```
ladipro, line 210: nit: typo `serilization` (multiple occurrences)

ostorc: Spelling, my nemesis.

src/Shared/ITranslator.cs
```diff
@@ -259,6 +259,18 @@ void Translate<T>(ref T value)
...
+        /// <summary>
+        /// Translates a byte array
+        /// </summary>
+        /// <param name="byteArray">The array to be translated.</param>
+        /// <param name="length">The length of array which will be used in translation</param>
+        /// <remarks>
+        /// This method should be used only for serialization, not for reading from stream.
+        /// Typical usage will be for example in combination with <see cref="MemoryStream.GetBuffer"> when we get array, but we want to use only part of it.
+        /// </remarks>
+        void Translate(ref byte[] byteArray, int length);
```
ladipro, line 272: Just thinking out loud: Maybe if the signature was:

```
void Translate(ref byte[] byteArray, ref int length);
```

you wouldn't have to comment that `length` is used only for writing. Reading would populate the argument with the length of the array (not super useful, I know, but eliminates this inelegant pattern where the interface method has a parameter that we know is going to be unused in half of its implementations). What do you think?

src/Build/Instance/ProjectItemInstance.cs
```diff
@@ -1596,16 +1596,31 @@ internal static TaskItem FactoryForDeserialization(ITranslator translator, Looka
                 return new TaskItem(translator, interner);
             }
 
+            private void WriteInternString(ITranslator translator, LookasideStringInterner interner, ref string str)
+            {
+                var key = interner.Intern(str);
+                translator.Writer.Write(key);
+            }
```
ladipro, line 1603: nit: Missing empty line between methods.

src/Build/BackEnd/Shared/TargetResult.cs
```diff
@@ -284,6 +280,68 @@ private void RetrieveItemsFromCache()
...
+                var interner = new LookasideStringInterner(StringComparer.Ordinal, _items.Length);
+                foreach (TaskItem t in _items)
+                {
+                    t.TranslateWithInterning(itemTranslator, interner);
+                }
+
+                interner.Translate(translator);
+                if (itemsStream.TryGetBuffer(out var buffer))
+                {
+                    var offset = buffer.Offset;
```
ladipro, line 314: Is this really needed? We know how we created the `MemoryStream` so `Offset` is guaranteed to be 0.

src/Build/BackEnd/Shared/TargetResult.cs
```diff
@@ -284,6 +280,68 @@ private void RetrieveItemsFromCache()
...
+            translator.Translate(ref itemsCount);
+
+            if (translator.Mode == TranslationDirection.WriteToStream)
+            {
+                // We will just calculate a very rough starting buffer size for the memory stream based on the number of items and a
+                // rough guess for an average number of bytes needed to store them compressed.  This doesn't have to be accurate, just
+                // big enough to avoid unnecessary buffer reallocations in most cases.
+                var defaultCompressedBufferCapacity = _items.Length * 64;
+                // Again, a rough calculation of buffer size, this time for an uncompressed buffer.  We assume compression 
+                // will give us 2:1, as it's all text.
```
ladipro, line 295: nit: It looks like the comment applied to the `DeflateStream` compression in the old code and is no longer relevant. Also, the variable is not used for anything else than calculating the uncompressed buffer capacity. Suggestion: simplify by eliminating this local and also remove the "compressed"/"uncompress" from identifier names.

src/Build/BackEnd/Components/Communications/NodeProviderOutOfProcBase.cs
```diff
@@ -887,6 +888,7 @@ private void HeaderReadComplete(IAsyncResult result)
                 }
 
                 int packetLength = BitConverter.ToInt32(_headerByte, 1);
+                MSBuildEventSource.Log.PacketReadSize(packetLength);
```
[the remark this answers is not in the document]
ostorc: Reverted

src/Build/Instance/ProjectItemInstance.cs
ostorc: Fixed

src/Build/BackEnd/Shared/TargetResult.cs
ostorc: Removed

src/Build/BackEnd/Shared/TargetResult.cs
ostorc: Fixed

src/Shared/BinaryTranslator.cs
```diff
@@ -997,17 +1011,31 @@ public void Translate<T>(ref T value)
             /// </summary>
             /// <param name="byteArray">The byte array to be translated</param>
             public void Translate(ref byte[] byteArray)
+            {
+                Translate(ref byteArray, byteArray?.Length ?? 0);
+            }
+
```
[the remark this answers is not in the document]
ostorc: Fixed

PR #5514, opened 2020-07-14
First stab at caching more RAR info

Forgind:
This gives RAR the ability to save off a modified state file and load it for later use.

@rainersigwald, would you mind looking at this at some point to let me know if you think it's going in the direction you expected? This is certainly incomplete—proposed future steps below.

Note: As far as I can tell, the only file path stored and serialized (currently) is the key—that is, FileStates didn't have any paths in them—but the sample solutions I tried didn't have scatter files, so I was unsure about those. As far as I could tell, they're never used? That felt wrong, but it's internal, and there were very few references to it, and they seemed to be assigning to it with one exception that could be using it in a very indirect way.

Still to do:
 - [x] Verify that the precomputed cache read in is real.
 - [x] Change logging such that it will log messages instead of errors or warnings if there are problems with deserializing the precomputed cache
 - [x] Comments at relevant points
 - [x] Switch serialization mode to another form (json?)
 - [x] Validation + performance tests
 - [ ] Have SDK opt in


src/Tasks/SystemState.cs
```diff
@@ -547,13 +552,58 @@ out fileState.frameworkName
...
+                foreach (string s in sfBase.instanceLocalFileStateCache.Keys)
+                {
+                    if (!retVal.instanceLocalFileStateCache.ContainsKey(s))
+                    {
+                        FileState fileState = (FileState)sfBase.instanceLocalFileStateCache[s];
+                        // Verify that the assembly is correct
+                        if (File.Exists(s) && Assembly.LoadFile(s).ManifestModule.ModuleVersionId.Equals(fileState.ModuleVersionID))
+                        {
+                            // Correct file path and timestamp
+                            string fullPath = Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), stateFile));
```
rainersigwald, line 575: directory of stateFile, not current directory, right?

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -1845,23 +1849,36 @@ private void LogConflict(Reference reference, string fusionName)
         /// <summary>
         /// Reads the state file (if present) into the cache.
         /// </summary>
-        private void ReadStateFile()
+        private void ReadStateFile(GetLastWriteTime getLastWriteTime, AssemblyTableInfo[] installedAssemblyTableInfo)
```
rainersigwald, line 1852: Why this change?

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -441,6 +441,10 @@ public string TargetedRuntimeVersion
             set { _targetedRuntimeVersionRawValue = value; }
         }
 
+        public string PreComputedCacheOutputPath { get; set; }
```
rainersigwald, line 444: I don't like this much better, would like to think on a name for these.

```suggestion
        public string OutputCache { get; set; }
```

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -441,6 +441,10 @@ public string TargetedRuntimeVersion
...
+
+        public string[] PreComputedCacheFileList { get; set; }
```
rainersigwald, line 446:
```suggestion
        public string[] AssemblyInformationCache { get; set; }
```

?

src/Tasks/SystemState.cs
```diff
@@ -547,13 +552,58 @@ out fileState.frameworkName
...
+
+            return retVal;
+        }
+
+        internal void SerializePrecomputedCache(string stateFile, TaskLoggingHelper log)
+        {
+            foreach (string path in instanceLocalFileStateCache.Keys)
+            {
+                FileState fileState = (FileState)instanceLocalFileStateCache[path];
+                fileState.ModuleVersionID = Assembly.LoadFrom(path).ManifestModule.ModuleVersionId;
```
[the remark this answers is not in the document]
Forgind: Does `Assembly.ReflectionOnlyLoadFrom(path)` also work? It seems clearer to me.

src/Tasks/SystemState.cs
Forgind: Yes—directory of stateFile, and the relative path is s. (I should probably change that name, too.)

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
Forgind: I know you said GetLastWriteTime was only provided for testing, but this is in case that ever changes. I wanted to update the `LastModified` field of each fileState to what it would be if it's queried later, and since it technically uses getLastWriteTime, I wanted to use the appropriate getLastWriteTime, which meant passing it through to DeserializePrecomputedCaches and using it to adjust LastModified rather than using the real timestamp.

Forgind: I added installedAssemblyTableInfo because they were set together, and it seemed wrong to split them. That part wasn't actually needed, though.

src/Tasks/SystemState.cs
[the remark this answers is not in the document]
rainersigwald: It does not. That also locks the assembly.

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
Forgind: How about CacheOutputPath, Relative(Path)CacheOutputPath, or OutputCachePath?

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
Forgind: "AssemblyInformation" just sounds like "related to RAR" to me. It would probably be best to match this with the above name, so the equivalents of my latest suggestions would be (Secondary)CacheInputPath, Relative(Path)CacheInputPath, or InputCachePath. Do any of those sound reasonable to you?

wli3:
Don't block on my approval. I am not very familiar with MSBuild. Although I will read this

src/Tasks/SystemState.cs
```diff
@@ -192,22 +196,24 @@ public void GetObjectData(SerializationInfo info, StreamingContext context)
...
                 }
+                info.AddValue("mvid", ModuleVersionID);
             }
 
             /// <summary>
             /// Gets the last modified date.
             /// </summary>
             /// <value></value>
-            internal DateTime LastModified
+            public DateTime LastModified
```
Forgind, line 206: These were made public as part of trying to switch to json. It seemed to be serializing/deserializing only public properties in my testing console even if it had access, but I think I'm still missing a piece.

BenVillalobos:
PR nit: 

- [ ] Verify that the precomputed cache read in is real.
- [ ] Change logging such that it will log messages instead of errors or warnings if there are problems with deserializing the precomputed cache
- [ ] Comments at relevant points
- [ ] Switch serialization mode to another form (json?)
- [ ] Validation + performance tests
- [ ] Have SDK opt in

Checkboxes! 😉

wli3:
Looks like we are still in explore phase, trying out it will help the perf or not. But once we have are sure about this route, we should have unit test coverage.

src/Tasks/SystemState.cs
```diff
@@ -115,7 +118,7 @@ private sealed class FileState : ISerializable
             /// <summary>
             /// The last modified time for this file.
             /// </summary>
-            private DateTime lastModified;
+            internal DateTime lastModified;
```
ladipro, line 121: nit: This change shouldn't be needed as JSON serialization ignores fields.

src/Tasks/SystemState.cs
```diff
@@ -548,12 +559,85 @@ out fileState.frameworkName
...
+
+            return retVal;
+        }
+
+        /// <summary>
+        /// Modifies this object to be more portable across machines, then writes it to stateFile.
+        /// </summary>
+        internal void SerializePrecomputedCache(string stateFile, TaskLoggingHelper log)
+        {
+            Hashtable newInstanceLocalFileStateCache = new Hashtable();
```
ladipro, line 611:
```suggestion
            Hashtable newInstanceLocalFileStateCache = new Hashtable(instanceLocalFileStateCache.Count);
```

src/Tasks/SystemState.cs
```diff
@@ -548,12 +559,85 @@ out fileState.frameworkName
...
+            foreach (string path in instanceLocalFileStateCache.Keys)
+            {
+                // Add MVID to allow us to verify that we are using the same assembly later
+                FileState fileState = (FileState)instanceLocalFileStateCache[path];
```
ladipro, line 615: nit: Enumerate `instanceLocalFileStateCache` and not `instanceLocalFileStateCache.Keys` to avoid another hashtable lookup here.

ladipro (commented):
Have you had a chance to measure how long it takes to read an assembly MVID versus read & parse its references?

src/Tasks/StateFileBase.cs
```diff
@@ -64,9 +65,10 @@ internal virtual void SerializeCache(string stateFile, TaskLoggingHelper log)
         /// <summary>
         /// Reads the specified file from disk into a StateFileBase derived object.
         /// </summary>
-        internal static StateFileBase DeserializeCache(string stateFile, TaskLoggingHelper log, Type requiredReturnType)
+        internal static StateFileBase DeserializeCache(string stateFile, TaskLoggingHelper log, Type requiredReturnType, bool logWarnings = true)
```
cdmihai, line 68: Looking at the usages of this method, wouldn't it look nicer as a generic method?

src/Tasks/SystemState.cs
```diff
@@ -548,12 +559,91 @@ out fileState.frameworkName
...
+
+            foreach (string stateFile in stateFiles)
+            {
+                // Verify that it's a real stateFile; log message but do not error if not
+                SystemState sysState = (SystemState)DeserializeCache(stateFile, log, requiredReturnType, false);
+                if (sysState == null)
+                {
+                    continue;
+                }
+                foreach (DictionaryEntry kvp in sysState.instanceLocalFileStateCache)
```
cdmihai, line 580: nit: make instanceLocalFileStateCache generic.

src/Tasks/SystemState.cs
```diff
@@ -548,12 +559,91 @@ out fileState.frameworkName
...
+                        Guid mvid;
+                        string fullPath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(stateFile), relativePath));
+                        if (File.Exists(fullPath))
+                        {
+                            using (var reader = new PEReader(File.OpenRead(fullPath)))
+                            {
+                                var metadataReader = reader.GetMetadataReader();
+                                mvid = metadataReader.GetGuid(metadataReader.GetModuleDefinition().Mvid);
+                            }
+                            if (mvid.Equals(fileState.ModuleVersionID))
```
cdmihai, line 596: Why is the mvid check required? What can go wrong if the check didn't exist?

src/Tasks/SystemState.cs
```diff
@@ -548,12 +559,91 @@ out fileState.frameworkName
...
+                foreach (DictionaryEntry kvp in sysState.instanceLocalFileStateCache)
+                {
+                    string relativePath = (string)kvp.Key;
+                    if (!assembliesFound.Contains(relativePath))
+                    {
+                        FileState fileState = (FileState)kvp.Value;
+                        // Verify that the assembly is correct
+                        Guid mvid;
+                        string fullPath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(stateFile), relativePath));
+                        if (File.Exists(fullPath))
```
cdmihai, line 589: use `FileSystems.Default`

src/Tasks/SystemState.cs
Forgind: This cache is expected to come from another machine, and you could have a different version of the assembly.

src/Tasks/SystemState.cs
Forgind: I don't understand. Do you mean to make the type of instanceLocalFileStateCache more generic, change DictionaryEntry to var, make it so this can pull from something other than a system state by making the input to this loop instanceLocalFileStateCache instead of sysState, or something else?

src/Tasks/SystemState.cs
cdmihai: I see, and the mvid is the only ID you found that's guaranteed to change, as people can lie about the assembly version (like we do). I wonder if using the PEReader to parse the assembly mvid is faster than just hashing the assembly contents as a blob.

src/Tasks/SystemState.cs
cdmihai: Make it a generic dictionary.

src/Tasks/SystemState.cs
Forgind: Right. I imagine which is faster would depend heavily on how big the assembly is—if it's tiny, you may as well hash the whole thing. If it's huge, mvids are the way to go.

cdmihai: Yup. Curious what the median BCL assembly size is.

Forgind:
Did my first performance test, comparing building MSBuild.sln (with no caches) to building MSBuild.sln with only the precomputed caches. Here are the times for RAR:
No caches:
image
Only precomputed caches:
image

Although there's always a nontrivial chance that I did something wrong, and these results are invalid, they look promising!

src/Tasks/SystemState.cs
```diff
@@ -544,12 +556,90 @@ out fileState.frameworkName
...
+        }
+
+        /// <summary>
+        /// Modifies this object to be more portable across machines, then writes it to stateFile.
+        /// </summary>
+        internal void SerializePrecomputedCache(string stateFile, TaskLoggingHelper log, Func<string, Guid> calculateMvid)
+        {
+            Dictionary<string, FileState> newInstanceLocalFileStateCache = new Dictionary<string, FileState>(instanceLocalFileStateCache.Count);
+            calculateMvid ??= CalculateMvid;
+            foreach (KeyValuePair<string, FileState> kvp in instanceLocalFileStateCache)
```
BenVillalobos, line 608: How large does the `instanceLocalFileStateCache` dictionary get on average?

Thinking out loud: I wonder what difference it would make to `foreach` over the keys and look up the values vs iterating over the kvp's.

src/Tasks/StateFileBase.cs
```diff
@@ -77,32 +78,35 @@ internal static StateFileBase DeserializeCache(string stateFile, TaskLoggingHelp
                     using (FileStream s = new FileStream(stateFile, FileMode.Open))
                     {
                         var formatter = new BinaryFormatter();
```
BenVillalobos, line 80: The commit history shows you reverted back to a binaryformatter over json serialization. What was the reason for this again?

src/Tasks/SystemState.cs
```diff
@@ -544,12 +556,90 @@ out fileState.frameworkName
...
+                // Verify that it's a real stateFile; log message but do not error if not
+                SystemState sysState = DeserializeCache<SystemState>(stateFile.ToString(), log, false);
+                if (sysState == null)
+                {
+                    continue;
+                }
+                foreach (KeyValuePair<string, FileState> kvp in sysState.instanceLocalFileStateCache)
+                {
+                    string relativePath = kvp.Key;
+                    if (!assembliesFound.Contains(relativePath))
```
BenVillalobos, line 582: Is this check necessary? You're iterating through a dictionary, there will be no duplicate keys.

BenVillalobos: Ah, I see there may be duplicates between files. Will order matter here? Will there be some future file that has more information than a previous one, and therefore that info won't get logged?

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+                    new TaskItem(precomputedCachePath)
+                };
+
+                u.ReadStateFile(File.GetLastWriteTime, Array.Empty<AssemblyTableInfo>(), calculateMvid, p => true);
+                u._cache.instanceLocalFileStateCache.ShouldNotContainKey(dllName);
+                File.Delete(standardCache.Path);
+                u._cache = null;
+                u.ReadStateFile(File.GetLastWriteTime, Array.Empty<AssemblyTableInfo>(), calculateMvid, p => true);
+                u._cache.instanceLocalFileStateCache.ShouldContainKey(dllName);
+                SystemState.FileState a3 = u._cache.instanceLocalFileStateCache[dllName];
```
BenVillalobos, line 95: nit: `t`, `u`, and `a3` aren't great variable names.

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+                t._cache = new SystemState();
+                t._cache.instanceLocalFileStateCache = new Dictionary<string, SystemState.FileState>() {
+                    { Path.Combine(standardCache.Path, "assembly1"), new SystemState.FileState(DateTime.Now) },
+                    { Path.Combine(standardCache.Path, "assembly2"), new SystemState.FileState(DateTime.Now) { Assembly = new Shared.AssemblyNameExtension("hi") } } };
+                t._cache.IsDirty = true;
+                t.StateFile = standardCache.Path;
+                t.WriteStateFile(calculateMvid);
+                int standardLen = File.ReadAllText(standardCache.Path).Length;
+                File.Delete(standardCache.Path);
+                standardLen.ShouldBeGreaterThan(0);
```
BenVillalobos, line 43: Is this a strong enough check? Should we directly compare `File.ReadAllText` to exactly what we think the output should be?

BenVillalobos (approved):
LGTM, my comments aren't merge-blocking.

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
Forgind: I'd love to, but unfortunately, that output would differ depending on serialization method.

src/Tasks/SystemState.cs
Forgind: No because any particular key with a valid mvid will be identical to any identical key/mvid pair.

src/Tasks/StateFileBase.cs
Forgind: Had trouble getting it to work. Made it work in #5704.

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
cdmihai: +1

---
In reply to: [505966901](https://github.com/dotnet/msbuild/pull/5514#discussion_r505966901) [](ancestors = 505966901)

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    new TaskItem(precomputedCachePath)
+                };
+
+                u.ReadStateFile(File.GetLastWriteTime, Array.Empty<AssemblyTableInfo>(), calculateMvid, p => true);
+                u._cache.instanceLocalFileStateCache.ShouldNotContainKey(dllName);
+                File.Delete(standardCache.Path);
+                u._cache = null;
+                u.ReadStateFile(File.GetLastWriteTime, Array.Empty<AssemblyTableInfo>(), calculateMvid, p => true);
+                u._cache.instanceLocalFileStateCache.ShouldContainKey(dllName);
```
cdmihai, line 94: Leave a comment saying why the entry exists here but not at line 90.

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
```diff
@@ -1855,25 +1855,33 @@ private void LogConflict(Reference reference, string fusionName)
...
         /// </summary>
-        private void WriteStateFile()
+        internal void WriteStateFile(Func<string, Guid> calculateMvid = null)
         {
-            if (!string.IsNullOrEmpty(_stateFile) && _cache.IsDirty)
+            if (!string.IsNullOrEmpty(AssemblyInformationCacheOutputPath))
+            {
+                _cache.SerializePrecomputedCache(AssemblyInformationCacheOutputPath, Log, calculateMvid);
+            }
+            else if (!string.IsNullOrEmpty(_stateFile) && _cache.IsDirty)
```
cdmihai, line 1884: Does it make any sense to run the else branch even when `AssemblyInformationCacheOutputPath` is set?

Forgind: I don't think there's any need to. It wouldn't be a bad idea to switch the order of these; the first is expected to almost never be true, and if it is, you're creating a cache to send to users, and since that would likely only happen on a remote build machine, you probably don't care about local state.

src/Tasks/SystemState.cs
```diff
@@ -31,7 +35,7 @@ internal sealed class SystemState : StateFileBase, ISerializable
         /// <summary>
         /// Cache at the SystemState instance level. It is serialized and reused between instances.
         /// </summary>
-        private Hashtable instanceLocalFileStateCache = new Hashtable(StringComparer.OrdinalIgnoreCase);
+        internal Dictionary<string, FileState> instanceLocalFileStateCache = new Dictionary<string, FileState>(StringComparer.OrdinalIgnoreCase);
```
cdmihai, line 38: Nice, nice.

src/Tasks/Resources/Strings.resx
```diff
@@ -451,6 +451,10 @@
     <value>MSB3101: Could not write state file "{0}". {1}</value>
     <comment>{StrBegin="MSB3101: "}</comment>
   </data>
+  <data name="General.StateFileAlreadyPresent">
+    <value>MSB3101: There is already a file at "{0}". If you are trying to create a precomputed cache, ensure that you are building a single project that depends on your assemblies rather than building your assemblies themselves. If you are running RAR normally, do not set the "_CacheOutputPath" field of RAR.</value>
```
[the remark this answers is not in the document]
cdmihai: Yea, make RAR explicit.

---
In reply to: [477323537](https://github.com/dotnet/msbuild/pull/5514#discussion_r477323537) [](ancestors = 477323537)

cdmihai (approved):

src/Tasks/SystemState.cs
cdmihai, line 582: What about conflicts? You could either error (as a sanity check that conflicts don't happen) or you could keep the newest one.

Forgind: If there are conflicts, the first one was already verified, so I skipped checking later ones that might be invalid to save a little time. I could easily imagine collisions, but if they both have the right mvid, they should be identical.

cdmihai: What if two files contain the same relative path but with different mvids?

Forgind:
I build OrchardCore once with no caches of any kind and once with the perfect precomputed caches. Results:
NoCaches:
Mean: 85838.3843
Median: 82582.754
PrecomputedCaches:
Mean: 61150.18
Median: 58261.5

Looking at the full data, I noticed more powerfully a trend hinted at by the data with building MSBuild: precomputed caches make it a little slower if RAR would be fast either way but much faster if it would actually have to calculate something. This make sense, since deserializing and validating the cache takes some time. More precisely, having no caches made RAR faster for the fastest forty (out of 159) RAR executions, ranging up to about 30 seconds. After that, however, the time it took increased much faster, so whereas the difference between the fastest RAR executions for the two was about 100ms, the difference between having precomputed caches and not having them was over a minute for the slowest execution.

src/Tasks/SystemState.cs
```diff
@@ -544,12 +549,73 @@ out fileState.frameworkName
...
+                {
+                    string relativePath = kvp.Key;
+                    if (!assembliesFound.Contains(relativePath))
+                    {
+                        FileState fileState = kvp.Value;
+                        string fullPath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(stateFile.ToString()), relativePath));
+                        if (fileExists(fullPath))
+                        {
+                            // Correct file path and timestamp
+                            fileState.LastModified = retVal.getLastWriteTime(fullPath);
```
rainersigwald, line 581: Is this the right timestamp? Doesn't it mean that we would potentially get incorrect information if the user updated one of the cached assemblies before a clean build?

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
rainersigwald: I don't think I understand the purpose of this test. Isn't it a subset of the other one that does a round trip?

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+                    new TaskItem(precomputedCachePath)
+                };
+
+                // At this point, we should have created two cache files: one "normal" one and one "precomputed" one.
+                // When we read the state file the first time, it should read from the caches produced in a normal
+                // build, partially because we can read it faster. If that cache does not exist, as with the second
+                // time we try to read the state file, it defaults to reading the "precomputed" cache. In this case,
+                // the normal cache does not have dll.dll, whereas the precomputed cache does, so it should not be
+                // present when we read the first time but should be present the second time. Then we verify that the
+                // information contained in that cache matches what we'd expect.
```
rainersigwald, line 88: I find this very confusing. Can you split it into two tests:

1. "Normal project-level StateFile cache takes precedence over AssemblyInformationCachePaths", and
2. "Writing and then reading AssemblyInformationCachePaths preserves all relevant information"?

src/Tasks/SystemState.cs
```diff
@@ -544,12 +549,73 @@ out fileState.frameworkName
...
+            {
+                string relativePath = FileUtilities.MakeRelative(Path.GetDirectoryName(stateFile), kvp.Key);
+                newInstanceLocalFileStateCache[relativePath] = kvp.Value;
+            }
+            instanceLocalFileStateCache = newInstanceLocalFileStateCache;
+
+            if (FileUtilities.FileExistsNoThrow(stateFile))
+            {
+                log.LogWarningWithCodeFromResources("General.StateFileAlreadyPresent", stateFile);
+            }
```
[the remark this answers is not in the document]
Forgind: We decided this in a meeting a while back. Most people won't use this (_making_ a cache) at all, and those who do may be confused as to whether they should be building their solution to create the cache or building a project that uses all the assemblies from their solution to create the cache. This attempts to push for the correct behavior.

src/Tasks/SystemState.cs
```diff
@@ -544,12 +549,73 @@ out fileState.frameworkName
...
+
+        /// <summary>
+        /// Modifies this object to be more portable across machines, then writes it to stateFile.
+        /// </summary>
+        internal void SerializePrecomputedCache(string stateFile, TaskLoggingHelper log)
+        {
+            Dictionary<string, FileState> newInstanceLocalFileStateCache = new Dictionary<string, FileState>(instanceLocalFileStateCache.Count);
+            foreach (KeyValuePair<string, FileState> kvp in instanceLocalFileStateCache)
+            {
+                string relativePath = FileUtilities.MakeRelative(Path.GetDirectoryName(stateFile), kvp.Key);
```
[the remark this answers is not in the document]
Forgind: I believe so. The cache will be serialized to a particular location relative to the SDK assemblies on the computer that builds the SDK, and when the user downloads both those assemblies and this cache, they will be in the same place relative to each other.

src/Tasks/SystemState.cs
```diff
@@ -282,6 +287,7 @@ public void GetObjectData(SerializationInfo info, StreamingContext context)
         internal bool IsDirty
         {
             get { return isDirty; }
+            set { isDirty = value; }
```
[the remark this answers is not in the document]
Forgind: Yes

src/Tasks/SystemState.cs
```diff
@@ -544,12 +549,73 @@ out fileState.frameworkName
...
+            retVal.SetGetLastWriteTime(getLastWriteTime);
+            retVal.SetInstalledAssemblyInformation(installedAssemblyTableInfo);
+            retVal.isDirty = stateFiles.Length > 0;
+            HashSet<string> assembliesFound = new HashSet<string>();
+            fileExists ??= FileSystems.Default.FileExists;
+
+            foreach (ITaskItem stateFile in stateFiles)
+            {
+                // Verify that it's a real stateFile; log message but do not error if not
+                SystemState sysState = DeserializeCache<SystemState>(stateFile.ToString(), log, false);
```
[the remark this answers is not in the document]
Forgind: I hadn't been thinking about that. The way I've been using them is passing in full paths, and users could get those from property + relative path if necessary. It's certainly limiting to make this assumption, but since it's new functionality, no one will be broken by it being restricted in that way, and it would certainly be faster to just know that it's a full path to the cache. I could go either way. It sounds like you'd prefer it account for globs and paths from other projects? We might need to talk about that if so, since I don't think I know all of how to do that.

src/Tasks/SystemState.cs
```diff
@@ -544,12 +549,73 @@ out fileState.frameworkName
...
+            SystemState retVal = new SystemState();
+            retVal.SetGetLastWriteTime(getLastWriteTime);
+            retVal.SetInstalledAssemblyInformation(installedAssemblyTableInfo);
+            retVal.isDirty = stateFiles.Length > 0;
+            HashSet<string> assembliesFound = new HashSet<string>();
+            fileExists ??= FileSystems.Default.FileExists;
+
+            foreach (ITaskItem stateFile in stateFiles)
+            {
+                // Verify that it's a real stateFile; log message but do not error if not
```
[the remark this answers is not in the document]
Forgind: I think it was a compat concern. Old state file uses serializationMethod_1, but user is using new version of MSBuild that expected seralizationMethod_2 --> we can't read the file, but we shouldn't error. It's better to have one slow build (then revert to normal speed) than to send an error and force the user to try to figure it out. This will hopefully be an edge case anyway, since a lot of the use of these caches is with new users.

src/Tasks/SystemState.cs
Forgind: Yes! Thank you for pointing that out. I removed that line entirely, and it would have made the cache (often) wrong. This was good behavior when I was checking for mvids but is no longer correct, and I didn't remove it.

src/Tasks/AssemblyDependency/ResolveAssemblyReference.cs
Forgind: I guess my question would be "why are you creating a precomputed cache if you're just doing an incremental build"? The precomputed caches are for sending to customers, and they're only useful when building from an otherwise cache-free state. Note that if I remember correctly, serializing caches took ~7% of RAR's total time, and serializing two caches would double the time spent on it.

src/Tasks.UnitTests/RARPrecomputedCache_Tests.cs
Forgind: Kinda, yeah. I still favor leaving it in because 1) it's already written, and 2) if there's a problem here (nontrivial possibility, since this is new), having a simpler test will more readily provide insight as to what exactly is broken.

PR #5527, opened 2020-07-17
Add support for specifying publicsign property for SGen task

Therzok:
Fixes https://github.com/microsoft/msbuild/issues/5526

### DIFF
--- a/src/Tasks/SGen.cs
+++ b/src/Tasks/SGen.cs
@@ -41,6 +41,8 @@ namespace Microsoft.Build.Tasks
 
         bool DelaySign { get; set; }
 
+        bool PublicSign { get; set; }
+
         [Output]
         ITaskItem[] SerializationAssembly { get; set; }
 
@@ -87,6 +89,8 @@ namespace Microsoft.Build.Tasks
         public string KeyFile { get; set; }
 
         public bool DelaySign { get; set; }
+
+        public bool PublicSign { get; set; }
 
         [Output]
         public ITaskItem[] SerializationAssembly { get; set; }
@@ -217,6 +221,12 @@ namespace Microsoft.Build.Tasks
         {
             set => Bag[nameof(DelaySign)] = value;
             get => GetBoolParameterWithDefault(nameof(DelaySign), false);
+        }
+
+        public bool PublicSign
+        {
+            set => Bag[nameof(PublicSign)] = value;
+            get => GetBoolParameterWithDefault(nameof(PublicSign), false);
         }
 
         [Output]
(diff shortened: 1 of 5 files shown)


src/Tasks/SGen.cs
```diff
@@ -404,6 +414,8 @@ protected override string GenerateCommandLineCommands()
 
                 commandLineBuilder.AppendPlusOrMinusSwitch("/compiler:/delaysign", Bag, "DelaySign");
 
+                commandLineBuilder.AppendPlusOrMinusSwitch("/compiler:/publicsign", Bag, "PublicSign");
```
Therzok, line 417: Is there anything special that needs to be done here?

`publicsign+` and `delaysign+` can't be both set at the same time, but `publicsign+` and `delaysign-` can.

I'm assuming it would fail earlier in the build process.

rainersigwald:
I don't think the compiler that sgen uses supports this option:

```sh-session
$ sgen artifacts\obj\Microsoft.Build.Framework\Debug\net472\Microsoft.Build.Framework.dll /compiler:/publicsign+
Microsoft (R) Xml Serialization support utility
[Microsoft (R) .NET Framework, Version 4.8.3928.0]
Error: Unable to generate a temporary class (result=1).
error CS2007: Unrecognized option: '/publicsign+'


If you would like more help, please type "sgen /?".
```

Therzok:
That's interesting, because it worked on Mono!

rainersigwald:
I bet Mono has an up-to-date Roslyn-based `csc.exe`, while Windows desktop .NET Framework's `sgen.exe` has the ancient C++ `csc.exe` that lives in the .NET Framework directory.

Therzok:
What would be an elegant way to fix this? I suppose it's not possible with .NET Framework constraints.

In any case, I'm not blocked on this, luckily the SGen task could be subclassed so I could implement this without depending on the original targets running.

rainersigwald:
I don't think there's a great way out of this. I thought of making it conditional on `#if MONO` but that leads to some confusion, especially around adding it to the default call to this task in Common.targets, which uses the standard `$(PublicSign)` and might work forward.

Is there an expected long future for `sgen` on Mono? For .NET Core, we advise replacing its use with the `Microsoft.XmlSerializer.Generator` package.

Therzok:
> Is there an expected long future for sgen on Mono? For .NET Core, we advise replacing its use with the Microsoft.XmlSerializer.Generator package.

Does that work on .netframework?

Therzok:
Closing this PR. I'll prototype using Microsoft.XmlSerializer.Generator. Thanks for the direction!

PR #5531, opened 2020-07-20
Sync XSD "IntelliSense" with VS repo automatically on insertion

rainersigwald:
We've been manually managing this and naturally forgotten it a few times. That's not cool. Let's let a robot do it.

This requires change to the releases: https://dev.azure.com/devdiv/DevDiv/_releaseProgress?releaseId=749728&_a=release-history but it will be harmless to do this first and then change them.

Example internal PR: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/262150?_a=files

### DIFF
--- a/src/MSBuild/MSBuild.csproj
+++ b/src/MSBuild/MSBuild.csproj
@@ -216,16 +216,9 @@
     <Content Include="MSBuild.rsp" Pack="false">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <None Include="MSBuild\Microsoft.Build.CommonTypes.xsd">
-      <SubType>Designer</SubType>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
-    <None Include="MSBuild\Microsoft.Build.Core.xsd">
-      <SubType>Designer</SubType>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
-    <None Include="Microsoft.Build.xsd">
-      <SubType>Designer</SubType>
+    <XsdsForVS Include="**\*.xsd" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <XsdsForVS Include="Update-MSBuildXsds.ps1" />
+    <None Include="@(XsdsForVS)">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
@@ -246,6 +239,14 @@
     <ProjectReference Include="..\Framework\Microsoft.Build.Framework.csproj" PrivateAssets="All" />
     <ProjectReference Include="..\Tasks\Microsoft.Build.Tasks.csproj" />
   </ItemGroup>
+
+  <Target Name="CopyXsds"
+          BeforeTargets="AfterBuild">
+    <Copy SourceFiles="@(XsdsForVS)"
+          DestinationFiles="@(XsdsForVS->'$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'xsd'))%(RecursiveDir)%(Filename)%(Extension)')"
+          SkipUnchangedFiles="$(SkipCopyUnchangedFiles)"
+          />
+  </Target>
 
   <!-- Include MSBuild.deps.json and MSBuild.runtimeconfig.json in ContentWithTargetPath so they will be copied to the output folder of projects
        that reference this one. -->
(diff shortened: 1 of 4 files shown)


src/MSBuild/MSBuild/Microsoft.Build.CommonTypes.xsd
```diff
@@ -5624,7 +5624,7 @@ elementFormDefault="qualified">
             </xs:complexContent>
         </xs:complexType>
     </xs:element>
-
+<!-- a change made on GitHub that should flow -->
```
rainersigwald, line 5627: I should of course revert this proof that it worked.

src/MSBuild/Update-MSBuildXsds.ps1
```diff
@@ -0,0 +1,9 @@
+$ErrorActionPreference = 'Stop'
+
+cmd /c set
```
rainersigwald, line 3: remove

src/MSBuild/MSBuild.csproj
```diff
@@ -247,6 +240,14 @@
     <ProjectReference Include="..\Tasks\Microsoft.Build.Tasks.csproj" />
   </ItemGroup>
 
+  <Target Name="CopyXsds"
+          BeforeTargets="AfterBuild">
+    <Copy SourceFiles="@(XsdsForVS)"
+          DestinationFiles="@(XsdsForVS->'$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'xsd'))%(RecursiveDir)%(Filename)%(Extension)')"
```
Forgind, line 246: Why does this work? I thought you didn't assign it metadata.

src/MSBuild/MSBuild.csproj
```diff
@@ -216,16 +216,9 @@
...
-      <SubType>Designer</SubType>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
-    <None Include="MSBuild\Microsoft.Build.Core.xsd">
-      <SubType>Designer</SubType>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
-    <None Include="Microsoft.Build.xsd">
-      <SubType>Designer</SubType>
+    <XsdsForVS Include="**\*.xsd" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
```
Forgind, line 219: This hits extra xsds and sometimes multiple copies of one xsd, right? I don't think that's a problem, although it might hit performance slightly, but I wanted to check if I understood.

src/MSBuild/MSBuild.csproj
rainersigwald: I don't understand the question. Can you rephrase? Are you asking about the [well-known metadata](https://docs.microsoft.com/visualstudio/msbuild/msbuild-well-known-item-metadata?view=vs-2019)?

src/MSBuild/MSBuild.csproj
rainersigwald: I don't understand the question. Can you rephrase? Specifically what do you mean by "extra xsds" and by "multiple copies of one xsd"?

Forgind: I was thinking that if it's looking for anything.xsd, it would also find RuleSet.xsd, CustomDictionary.xsd, etc. and a different copy for each build configuration, but it occurred to me that that's all under the artifacts folder, so I think I was wrong.

src/MSBuild/MSBuild.csproj
Forgind: I'd forgotten about those. Thanks!

src/MSBuild/MSBuild.csproj
BenVillalobos, line 219: There are two "Default exclude" items here,  so presumably "DefaultItemExcludes" doesn't include the "DefaultExcludesInProjectFolder". How does one know or find out how many default exclude items there are?

Forgind: There's probably a better way, but I build a project that excludes those, catch it in a debugger, break on the code that handles that in `Project.cs`, and inspect what it gets turned into.

rainersigwald: I'm following the SDK's approaches here:

https://github.com/dotnet/sdk/blob/e053f40806bf1e50b01271cbddc57759d24f0b0b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets#L22-L42

https://github.com/dotnet/sdk/blob/062d9a036ea335833e472d3b3c17174ba62c5c19/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#L37

No real way to know other than looking for analogies there that I know of.

PR #5532, opened 2020-07-20
SignFile: Treat empty TimestampUrl as null

MatthewSteeples:
Unless I'm missing something, it's quite difficult to set a Property to be null in MSBuild once it's been declared as having a value. An empty string is as close as it gets.

So, if you're making the TimestampUrl a configurable property as part of a build pipeline you can only specify an empty string. This causes this task to throw an exception because `new Uri("")` produces the following

> System.UriFormatException: Invalid URI: The URI is empty.

### DIFF
--- a/src/Tasks/SignFile.cs
+++ b/src/Tasks/SignFile.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Build.Tasks
             try
             {
                 SecurityUtilities.SignFile(CertificateThumbprint,
-                TimestampUrl == null ? null : new Uri(TimestampUrl),
+                string.IsNullOrEmpty(TimestampUrl) ? null : new Uri(TimestampUrl),
                 SigningTarget.ItemSpec, TargetFrameworkVersion);
                 return true;
             }


rainersigwald:
Can you give an example project that causes this to fail? The change looks good (and thanks!) but I think there are ways to deal with the situation you're describing without it that may be helpful to you.

MatthewSteeples:
I may have jumped the gun slightly here. After some digging, I'd misunderstood what the problem was, and the variable was actually being set elsewhere. In summary, one of the timestamp servers that we were using was offline the other week, and we were trying to remove our dependency on it entirely (as we ship often enough that it was no longer considered beneficial to use it). 

When we stopped setting the variable we still received an error, so I assumed that it was down to the value being null, which is when I put this pull request together. It turns out that we'd missed one place where the value was being set, so it was still trying to use a timestamp server, which was returning an error (due to it not being available).

In summary, this PR is probably not needed (although I've been struggling to find any information on how MSBuild deals with nulls in general if you have some pointers for that?)

Forgind:
Closing since it sounds like you don't need this anymore.

It's a little arbitrary as far as how it deals with nulls. Are there specific cases you were wondering about?

MatthewSteeples:
@Forgind No scenarios in particular. Our problem was purely diagnostic rather than new development

PR #5538, opened 2020-07-23
5513 refactor createmanifestresourcename

### LINKED ISSUE #5513: Refactor CreateManifestResourceName to reduce string builder creation
MakeValidEverettIdentifier calls MakeValidEverettFolderIdentifier in a for loop, which creates a new string builder for each invocation. 

MakeValidEverettFolderIdentifier then, also in a for loop, calls MakeValidEverettSubFolderIdentifier. Which again creates a new string builder for each invocation.

I suggest MakeValidEverettIdentifier creates the string builder and it is passed through the calls.

https://github.com/microsoft/msbuild/blob/ff5daef6923623fe685528f3c22ab94d99277cc6/src/Tasks/CreateManifestResourceName.cs#L296-L404

elachlan:
fixes #5513

### DIFF
--- a/src/Tasks/CreateCSharpManifestResourceName.cs
+++ b/src/Tasks/CreateCSharpManifestResourceName.cs
@@ -156,15 +156,13 @@ namespace Microsoft.Build.Tasks
                         string.Equals(sourceExtension, ".resources", StringComparison.OrdinalIgnoreCase)
                     )
                 {
-
                     if (!string.IsNullOrEmpty(directoryName))
                     {
-                        manifestName.Append('.').Append(Path.GetFileNameWithoutExtension(info.cultureNeutralFilename));
+                        manifestName.Append('.');
                     }
-                    else
-                    {
-                        manifestName.Append(Path.GetFileNameWithoutExtension(info.cultureNeutralFilename));
-                    }
+
+                    // append the file name without extension
+                    manifestName.Append(Path.GetFileNameWithoutExtension(info.cultureNeutralFilename));
 
                     // Replace all '\' with '.'
                     manifestName.Replace(Path.DirectorySeparatorChar, '.');
--- a/src/Tasks/CreateCSharpManifestResourceName.cs
+++ b/src/Tasks/CreateCSharpManifestResourceName.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Build.Tasks
                 // Empty namespaces are allowed.
                 if (!string.IsNullOrEmpty(rootNamespace))
                 {
-                    manifestName.Append(rootNamespace).Append(".");
+                    manifestName.Append(rootNamespace).Append('.');
                 }
 
                 // only strip extension for .resx and .restext files
@@ -173,7 +173,7 @@ namespace Microsoft.Build.Tasks
                     // Append the culture if there is one.        
                     if (!string.IsNullOrEmpty(info.culture))
                     {
-                        manifestName.Append(".").Append(info.culture);
+                        manifestName.Append('.').Append(info.culture);
                     }
 
                     // If the original extension was .resources, add it back
--- a/src/Tasks/CreateCSharpManifestResourceName.cs
+++ b/src/Tasks/CreateCSharpManifestResourceName.cs
@@ -219,7 +219,7 @@ namespace Microsoft.Build.Tasks
         protected override bool IsSourceFile(string fileName)
         {
             string extension = Path.GetExtension(fileName);
-            return (String.Compare(extension, ".cs", StringComparison.OrdinalIgnoreCase) == 0);
+            return (String.Compare(extension, SourceFileExtension, StringComparison.OrdinalIgnoreCase) == 0);
         }
     }
 }
--- a/src/Tasks/CreateCSharpManifestResourceName.cs
+++ b/src/Tasks/CreateCSharpManifestResourceName.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Build.Tasks
             Culture.ItemCultureInfo info = Culture.GetItemCultureInfo(embeddedFileName, dependentUponFileName);
 
             // If the item has a culture override, respect that. 
-            if (!String.IsNullOrEmpty(culture))
+            if (!string.IsNullOrEmpty(culture))
             {
                 info.culture = culture;
             }
@@ -144,18 +144,21 @@ namespace Microsoft.Build.Tasks
 
                 // only strip extension for .resx and .restext files
                 string sourceExtension = Path.GetExtension(info.cultureNeutralFilename);
+                string directoryName = Path.GetDirectoryName(info.cultureNeutralFilename);
+
+                // append the directory name
+                MakeValidEverettIdentifier(manifestName, directoryName);
                 if (
-                        (String.Equals(sourceExtension, ".resx", StringComparison.OrdinalIgnoreCase))
+                        string.Equals(sourceExtension, ".resx", StringComparison.OrdinalIgnoreCase)
                         ||
-                        (String.Equals(sourceExtension, ".restext", StringComparison.OrdinalIgnoreCase))
+                        string.Equals(sourceExtension, ".restext", StringComparison.OrdinalIgnoreCase)
                         ||
-                        (String.Equals(sourceExtension, ".resources", StringComparison.OrdinalIgnoreCase))
+                        string.Equals(sourceExtension, ".resources", StringComparison.OrdinalIgnoreCase)
                     )
                 {
-                    var directoryName = Path.GetDirectoryName(info.cultureNeutralFilename);
+
                     if (!string.IsNullOrEmpty(directoryName))
                     {
-                        MakeValidEverettIdentifier(manifestName, directoryName);
                         manifestName.Append('.').Append(Path.GetFileNameWithoutExtension(info.cultureNeutralFilename));
                     }
                     else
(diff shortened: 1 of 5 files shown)


elachlan:
@Forgind I have gone and implemented the changes we talked about. Are you able to have a quick look?

src/Tasks/CreateManifestResourceName.cs
```diff
@@ -375,32 +373,31 @@ internal static string MakeValidEverettFolderIdentifier(string name)
...
             ErrorUtilities.VerifyThrowArgumentNull(name, nameof(name));
 
-            var everettId = new StringBuilder(name.Length);
-
             // split the name into folder names
             string[] subNames = name.Split(MSBuildConstants.ForwardSlashBackslash);
 
             // convert every folder name
-            everettId.Append(MakeValidEverettFolderIdentifier(subNames[0]));
+            if (!string.IsNullOrEmpty(subNames[0]))
```
BenVillalobos, line 391: should we just have a base case that returns on an empty `name`?

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/msbuild?pullRequest=5538) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/msbuild?pullRequest=5538) before we can accept your contribution.<br/><br/>:x: elachlan [sign now](https://cla.dotnetfoundation.org/dotnet/msbuild?pullRequest=5538)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/msbuild?pullRequest=5538) it.</sub>

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/msbuild?pullRequest=5538) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/msbuild?pullRequest=5538) before we can accept your contribution.<br/><br/>:x: elachlan [sign now](https://cla.dotnetfoundation.org/dotnet/msbuild?pullRequest=5538)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/msbuild?pullRequest=5538) it.</sub>

elachlan:
@BenVillalobos have you got any ideas on MakeValidEverettFolderIdentifier? I wasn't able to remove its string builder. As it checks its value at the bottom for a single underscore. I think we would have to get a substring to do it, which might be worse performance than just having the string builder.

elachlan:
Also are there any guidelines on when to use MSBuildConstants? there are dots and underscore characters everywhere and I am wondering if they should use the constants.

elachlan:
Also maybe it would help to have constants for ".resx" and the like?

src/Tasks/CreateManifestResourceName.cs
```diff
@@ -296,111 +296,121 @@ private static bool IsValidEverettIdChar(char c)
...
             ErrorUtilities.VerifyThrowArgumentNull(name, nameof(name));
 
-            // give string length to avoid reallocations; +1 since the resulting string may be one char longer than the
-            // original - if the name is a single underscore we add another underscore to it
-            var everettId = new StringBuilder(name.Length + 1);
+            if (!string.IsNullOrEmpty(name))
+            {
+                // give string length to avoid reallocations; +1 since the resulting string may be one char longer than the
+                // original - if the name is a single underscore we add another underscore to it
+                var everettId = new StringBuilder(name.Length + 1);
```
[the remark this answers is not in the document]
elachlan: Doing this breaks the tests, as its not comparing the whole path, but just the specific folder.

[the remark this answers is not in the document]
elachlan: I think we can grab the length as the start and then use a substring or something.

elachlan:
I think the build failures are caused by the Public API not matching the methods in CreateManifestResourceName.

src/Tasks/CreateCSharpManifestResourceName.cs
```diff
@@ -166,18 +171,25 @@ TaskLoggingHelper log
...
                 {
-                    manifestName.Append(Path.Combine(everettCompatibleDirectoryName, Path.GetFileName(info.cultureNeutralFilename)));
+                    if (!string.IsNullOrEmpty(directoryName))
+                    {
+                        manifestName.Append('.').Append(Path.GetFileName(info.cultureNeutralFilename));
+                    }
+                    else
+                    {
+                        manifestName.Append(Path.GetFileName(info.cultureNeutralFilename));
+                    }
```
Forgind, line 192: last nit:
```suggestion
                    if (!string.IsNullOrEmpty(directoryName))
                    {
                        manifestName.Append('.');
                    }
                    
                    manifestName.Append(Path.GetFileName(info.cultureNeutralFilename));
```

src/Tasks/CreateManifestResourceName.cs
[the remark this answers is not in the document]
Forgind: Good thought

src/Tasks/CreateCSharpManifestResourceName.cs
elachlan: Thanks. I have made the change.

elachlan:
One change we should maybe look at is Execute:
https://github.com/dotnet/msbuild/blob/e8338f5ad91d42c3196218648859e9e837099842/src/Tasks/CreateManifestResourceName.cs#L114-L256

It has a for each loop and makes the call to CreateManifestName, which in turn calls the associated CreateManifestNameImpl. I think it would be more performant if we can pass a string builder down through there and avoid a string builder creation within the for loop. We would just call StringBuilder.Clear() at the end of each iteration.

Forgind:
I don't think we can change the signature of CreateManifestName, since that could theoretically be a breaking change, although I find it hard to imagine anyone actually relies on that. It might be safer (and easier) to define a constant StringBuilder at the top of the file and reuse it between calls. On the other hand, that makes me start to wonder if it's overoptimizing—do you know how often this code runs? Meaning how much would reusing the StringBuilder improve overall build performance for the average user?

elachlan:
> I don't think we can change the signature of CreateManifestName, since that could theoretically be a breaking change, although I find it hard to imagine anyone actually relies on that. It might be safer (and easier) to define a constant StringBuilder at the top of the file and reuse it between calls. On the other hand, that makes me start to wonder if it's overoptimizing—do you know how often this code runs? Meaning how much would reusing the StringBuilder improve overall build performance for the average user?

I think you are correct on over-optimising. I'd imagine the code gets run once per embedded resource, and that wouldn't make much difference as the majority of the work would be in compilation or something along those lines.

Maybe it can be optimised at a later date. There are probably other areas that need it more.

elachlan:
Should the const values be marked as internal? That way they don't change the Public API?

Forgind:
> Should the const values be marked as internal? That way they don't change the Public API?

Yes, that would be good.

src/Tasks/CreateManifestResourceName.cs
```diff
@@ -296,86 +299,77 @@ private static bool IsValidEverettIdChar(char c)
...
         /// </summary>
-        internal static string MakeValidEverettFolderIdentifier(string name)
+        internal static void MakeValidEverettFolderIdentifier(StringBuilder builder, string name)
         {
             ErrorUtilities.VerifyThrowArgumentNull(name, nameof(name));
 
-            // give string length to avoid reallocations; +1 since the resulting string may be one char longer than the
-            // original - if the name is a single underscore we add another underscore to it
-            var everettId = new StringBuilder(name.Length + 1);
+            if (!string.IsNullOrEmpty(name))
```
BenVillalobos, line 349: Code styling nit: I couldn't find anything specific over at the [coding style document](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md) we refer to in our contributing guide, but I suggest turning this into something like `if(string.IsNullOrEmpty(name)) { return; }` to avoid nesting.

src/Tasks/CreateManifestResourceName.cs
```diff
@@ -296,86 +299,77 @@ private static bool IsValidEverettIdChar(char c)
         /// <summary>
         /// Make a folder subname into an Everett-compatible identifier 
         /// </summary>
-        private static string MakeValidEverettSubFolderIdentifier(string subName)
+        private static void MakeValidEverettSubFolderIdentifier(StringBuilder builder, string subName)
         {
             ErrorUtilities.VerifyThrowArgumentNull(subName, nameof(subName));
 
-            if (subName.Length == 0)
+            if (!string.IsNullOrEmpty(subName))
```
BenVillalobos, line 306: Same as below.

src/Tasks/CreateManifestResourceName.cs
```diff
@@ -385,22 +379,28 @@ internal static string MakeValidEverettFolderIdentifier(string name)
         public static string MakeValidEverettIdentifier(string name)
         {
             ErrorUtilities.VerifyThrowArgumentNull(name, nameof(name));
+            if (!string.IsNullOrEmpty(name))
```
BenVillalobos, line 382: with respect to my other nits, have our "base case" come first and return. Followed by all code that would run under normal circumstances

```suggestion
            if (string.IsNullOrEmpty(name))
            {
                   return;
            }
            
            <rest of code goes here with no else>
```

elachlan:
> Thank you @elachlan for the PR! I have added a couple of comments inline.
> 
> Do you think it would make sense to add a unit test to verify that `MakeValidEverettIdentifier()` behaves the same with your changes?

I have made the changes you asked for. I don't know much about msbuild and how its unit tested to be honest. I might have to leave the tests to someone more knowledgeable than myself.

ladipro:
> I have made the changes you asked for. I don't know much about msbuild and how its unit tested to be honest. I might have to leave the tests to someone more knowledgeable than myself.

Thank you. The code you have refactored appears to be well covered by `CreateCSharpManifestResourceName_Tests` and `CreateVisualBasicManifestResourceName_Tests`, specifically with `Regress311473()` in the former. So a new unit test is not required.

Forgind:
Thanks!

PR #5544, opened 2020-07-24
Use UnhandledException handler only in MSBuild.exe

### LINKED ISSUE #5543: MSBuild adds an AppDomain.UnhandledException handler even when running hosted
It is unusual to be logging unhandled exceptions in somebody else's process/AD just because they have called an MSBuild API. MSBuild may not be on the stack at all when an exception goes unhandled and the work the handler is doing may not be safe in all processes.

We are seeing issues with the UnhandledException handler when running inside Visual Studio. Specifically deadlocks at this call stack:

```
clr.dll!AwareLock::Enter
clr.dll!JITutil_MonReliableContention
microsoft.build.dll!Microsoft.Build.Shared.ExceptionHandling.DumpExceptionToFile
microsoft.build.dll!Microsoft.Build.Shared.ExceptionHandling.UnhandledExceptionHandler
clr.dll!CallDescrWorkerWithHandler
clr.dll!DispatchCallSimple
clr.dll!ExceptionNotifications::DeliverExceptionNotification
clr.dll!InvokeUnhandledSwallowing
clr.dll!DistributeUnhandledExceptionReliably
clr.dll!AppDomain::RaiseUnhandledExceptionEvent
```

This is the code installing the handler (in a static ctor, no less):
https://github.com/dotnet/msbuild/blob/e8338f5ad91d42c3196218648859e9e837099842/src/Build/Definition/ProjectCollection.cs#L235

ladipro:
This change moves adding the UnhandledException handler to the current AppDomain from various places in the MSBuild library to one location in MSBuild.exe. This ensures that it is not done when MSBuild is running hosted in another process where such handler with its side-effects is considered problematic.

Fixes #5543

### DIFF
--- a/src/Build/BackEnd/Node/OutOfProcNode.cs
+++ b/src/Build/BackEnd/Node/OutOfProcNode.cs
@@ -140,10 +140,6 @@ namespace Microsoft.Build.Execution
         {
             s_isOutOfProcNode = true;
 
-#if FEATURE_APPDOMAIN_UNHANDLED_EXCEPTION
-            AppDomain.CurrentDomain.UnhandledException += ExceptionHandling.UnhandledExceptionHandler;
-#endif
-
             _debugCommunications = (Environment.GetEnvironmentVariable("MSBUILDDEBUGCOMM") == "1");
 
             _receivedPackets = new ConcurrentQueue<INodePacket>();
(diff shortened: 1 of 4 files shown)


rainersigwald:
We went into ancient history and discovered that this was first introduced in a commit labeled (in its entirety)

```
Changeset 966904: Some <Choose> related fixes to help unblock Jay.
```

So: sorry, Jay. Maybe you're blocked again? Let us know.

PR #5546, opened 2020-07-25
Share file system caches via the EvaluationContext

cdmihai:
- Made IFileSystem public. Added some extra methods to it from QuickBuild's file system interface. This interface will also be passed to the future project cache interface so that caches can reuse IO from evaluation and between themselves.
- EvaluationContext can take an IFileSystem. This enables QuickBuild and VisualStudio to share their file system caches with msbuild evaluations.

Possible Next steps (future PRs):
- Pass the interface to sdk resolvers in a non breaking manner. AFAIK they do a lot of IO
- I saw some evaluation bits that do IO without passing through the interface, so propagating the interface to those places as well seems like a potential perf gain.

### DIFF
--- a/src/Build.UnitTests/Definition/ProjectEvaluationContext_Tests.cs
+++ b/src/Build.UnitTests/Definition/ProjectEvaluationContext_Tests.cs
@@ -86,6 +86,51 @@ namespace Microsoft.Build.UnitTests.Definition
             }
         }
 
+        [Fact]
+        public void PassedInFileSystemShouldBeReusedInSharedContext()
+        {
+            var projectFiles = new[]
+            {
+                _env.CreateFile("1.proj", @"<Project> <PropertyGroup Condition=`Exists('1.file')`></PropertyGroup> </Project>".Cleanup()).Path,
+                _env.CreateFile("2.proj", @"<Project> <PropertyGroup Condition=`Exists('2.file')`></PropertyGroup> </Project>".Cleanup()).Path
+            };
+
+            var projectCollection = _env.CreateProjectCollection().Collection;
+            var fileSystem = new Helpers.LoggingFileSystem();
+            var evaluationContext = EvaluationContext.Create(EvaluationContext.SharingPolicy.Shared, fileSystem);
+
+            foreach (var projectFile in projectFiles)
+            {
+                Project.FromFile(
+                    projectFile,
+                    new ProjectOptions
+                    {
+                        ProjectCollection = projectCollection,
+                        EvaluationContext = evaluationContext
+                    }
+                );
+            }
+
+            fileSystem.ExistenceChecks.OrderBy(kvp => kvp.Key)
+                .ShouldBe(
+                    new Dictionary<string, int>
+                    {
+                        {Path.Combine(_env.DefaultTestDirectory.Path, "1.file"), 1},
+                        {Path.Combine(_env.DefaultTestDirectory.Path, "2.file"), 1}
+                    }.OrderBy(kvp => kvp.Key));
+
+            fileSystem.DirectoryEntryExistsCalls.ShouldBe(2);
+        }
+
+        [Fact]
+        public void IsolatedContextShouldNotSupportBeingPassedAFileSystem()
+        {
+            _env.DoNotLaunchDebugger();
+
+            var fileSystem = new Helpers.LoggingFileSystem();
+            Should.Throw<ArgumentException>(() => EvaluationContext.Create(EvaluationContext.SharingPolicy.Isolated, fileSystem));
+        }
+
         [Theory]
         [InlineData(EvaluationContext.SharingPolicy.Shared)]
         [InlineData(EvaluationContext.SharingPolicy.Isolated)]
--- a/src/Build/Evaluation/Context/EvaluationContext.cs
+++ b/src/Build/Evaluation/Context/EvaluationContext.cs
@@ -6,6 +6,7 @@ using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using System.Threading;
 using Microsoft.Build.BackEnd.SdkResolution;
+using Microsoft.Build.FileSystem;
 using Microsoft.Build.Internal;
 using Microsoft.Build.Shared;
 using Microsoft.Build.Shared.FileSystem;
@@ -17,13 +18,20 @@ namespace Microsoft.Build.Evaluation.Context
     ///     evaluations).
     ///     The caller should throw away the context when the environment changes (IO, environment variables, SDK resolution
     ///     inputs, etc).
-    ///     This class and it's closure needs to be thread safe since API users can do evaluations in parallel.
+    ///     This class and its closure needs to be thread safe since API users can do evaluations in parallel.
     /// </summary>
     public class EvaluationContext
     {
         public enum SharingPolicy
         {
+            /// <summary>
+            /// Instructs the <see cref="EvaluationContext"/> to reuse state between the different project evaluations that use it.
+            /// </summary>
             Shared,
+
+            /// <summary>
+            /// Instructs the <see cref="EvaluationContext"/> not to reuse state between the different project evaluations that use it.
+            /// </summary>
             Isolated
         }
 
@@ -40,15 +48,21 @@ namespace Microsoft.Build.Evaluation.Context
         /// <summary>
         /// Key to file entry list. Example usages: cache glob expansion and intermediary directory expansions during glob expansion.
         /// </summary>
-        internal ConcurrentDictionary<string, ImmutableArray<string>> FileEntryExpansionCache { get; }
+        private ConcurrentDictionary<string, ImmutableArray<string>> FileEntryExpansionCache { get; }
 
-        internal EvaluationContext(SharingPolicy policy)
+        private EvaluationContext(SharingPolicy policy, IFileSystem fileSystem)
         {
+            // Unsupported case: isolated context with non null file system.
+            // Isolated means caches aren't reused, but the given file system might cache.
+            ErrorUtilities.VerifyThrowArgument(
+                policy == SharingPolicy.Shared || fileSystem == null,
+                "IsolatedContextDoesNotSupportFileSystem");
+
             Policy = policy;
 
             SdkResolverService = new CachingSdkResolverService();
             FileEntryExpansionCache = new ConcurrentDictionary<string, ImmutableArray<string>>();
-            FileSystem = new CachingFileSystemWrapper(FileSystems.Default);
+            FileSystem = fileSystem ?? new CachingFileSystemWrapper(FileSystems.Default);
             EngineFileUtilities = new EngineFileUtilities(new FileMatcher(FileSystem, FileEntryExpansionCache));
         }
 
@@ -57,7 +71,28 @@ namespace Microsoft.Build.Evaluation.Context
         /// </summary>
         public static EvaluationContext Create(SharingPolicy policy)
         {
-            var context = new EvaluationContext(policy);
+            
+            // ReSharper disable once IntroduceOptionalParameters.Global
+            // do not remove this method to avoid breaking binary compatibility
+            return Create(policy, fileSystem: null);
+        }
+
+        /// <summary>
+        ///     Factory for <see cref="EvaluationContext" />
+        /// </summary>
+        /// <param name="policy"> The <see cref="SharingPolicy"/> to use.</param>
+        /// <param name="fileSystem">The <see cref="IFileSystem"/> to use.
+        ///     This parameter is compatible only with <see cref="SharingPolicy.Shared"/>.
+        ///     The method throws if a file system is used with <see cref="SharingPolicy.Isolated"/>.
+        ///     The reasoning is that <see cref="SharingPolicy.Isolated"/> means not reusing any caches between evaluations,
+        ///     and the passed in <paramref name="fileSystem"/> might cache state.
+        /// </param>
+        public static EvaluationContext Create(SharingPolicy policy, MSBuildFileSystemBase fileSystem)
+        {
+            var context = new EvaluationContext(
+                policy,
+                fileSystem == null ? null : new MSBuildFileSystemAdapter(fileSystem));
+
             TestOnlyHookOnCreate?.Invoke(context);
 
             return context;
--- a/src/Build/FileSystem/MSBuildFileSystemAdapter.cs
+++ b/src/Build/FileSystem/MSBuildFileSystemAdapter.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Build.Shared.FileSystem;
+
+namespace Microsoft.Build.FileSystem
+{
+     internal class MSBuildFileSystemAdapter : IFileSystem
+    {
+        private readonly MSBuildFileSystemBase _msbuildFileSystem;
+        public MSBuildFileSystemAdapter(MSBuildFileSystemBase msbuildFileSystem)
+        {
+            _msbuildFileSystem = msbuildFileSystem;
+        }
+        public TextReader ReadFile(string path) => _msbuildFileSystem.ReadFile(path);
+
+        public Stream GetFileStream(string path, FileMode mode, FileAccess access, FileShare share) => _msbuildFileSystem.GetFileStream(path, mode, access, share);
+
+        public string ReadFileAllText(string path) => _msbuildFileSystem.ReadFileAllText(path);
+
+        public byte[] ReadFileAllBytes(string path) => _msbuildFileSystem.ReadFileAllBytes(path);
+
+        public IEnumerable<string> EnumerateFiles(string path, string searchPattern = "*", SearchOption searchOption = SearchOption.TopDirectoryOnly)
+        {
+            return _msbuildFileSystem.EnumerateFiles(path, searchPattern, searchOption);
+        }
+
+        public IEnumerable<string> EnumerateDirectories(string path, string searchPattern = "*", SearchOption searchOption = SearchOption.TopDirectoryOnly)
+        {
+            return _msbuildFileSystem.EnumerateDirectories(path, searchPattern, searchOption);
+        }
+
+        public IEnumerable<string> EnumerateFileSystemEntries(
+            string path,
+            string searchPattern = "*",
+            SearchOption searchOption = SearchOption.TopDirectoryOnly)
+        {
+            return _msbuildFileSystem.EnumerateFileSystemEntries(path, searchPattern, searchOption);
+        }
+
+        public FileAttributes GetAttributes(string path) => _msbuildFileSystem.GetAttributes(path);
+
+        public DateTime GetLastWriteTimeUtc(string path) => _msbuildFileSystem.GetLastWriteTimeUtc(path);
+
+        public bool DirectoryExists(string path) => _msbuildFileSystem.DirectoryExists(path);
+
+        public bool FileExists(string path) => _msbuildFileSystem.FileExists(path);
+
+        public bool DirectoryEntryExists(string path) => _msbuildFileSystem.FileOrDirectoryExists(path);
+    }
+}
--- a/src/Build/Resources/Strings.resx
+++ b/src/Build/Resources/Strings.resx
@@ -1831,4 +1831,7 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
   <data name="StaticGraphConstructionMetrics" xml:space="preserve">
     <value>"Static graph loaded in {0} seconds: {1} nodes, {2} edges"</value>
   </data>
+  <data name="IsolatedContextDoesNotSupportFileSystem" xml:space="preserve">
+    <value>"EvaluationContext objects created with SharingPolicy.Isolated do not support being passed an MSBuildFileSystemBase file system."</value>
+  </data>
 </root>
--- a/src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
+++ b/src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
@@ -21,6 +21,16 @@ namespace Microsoft.Build.Shared.FileSystem
             return NativeMethodsShared.FileOrDirectoryExists(path);
         }
 
+        public FileAttributes GetAttributes(string path)
+        {
+            return File.GetAttributes(path);
+        }
+
+        public DateTime GetLastWriteTimeUtc(string path)
+        {
+            return File.GetLastWriteTimeUtc(path);
+        }
+
         public bool DirectoryExists(string path)
         {
             return NativeMethodsShared.DirectoryExists(path);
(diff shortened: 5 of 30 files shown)


cdmihai:
@lifengl @dfederm

cdmihai:
@Forgind, @rainersigwald 
The PR fails this curious looking test: https://github.com/dotnet/msbuild/blob/e8338f5ad91d42c3196218648859e9e837099842/src/Build.OM.UnitTests/NugetRestoreTests.cs#L21-L26
What's the idea behind the test? We can't ever break that buggy old nuget version? :)

The change that breaks it is that on master, IFileSystem is included in every MSBuild dll, but in this PR I am making it public, so I have to put it in only a single place as opposed to inserting it in every dll, otherwise C# fails with ambiguous typing errors. This breaks the old nuget version, but the latest nuget works.

How should I go around it?

src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
```diff
@@ -21,6 +21,26 @@ public bool DirectoryEntryExists(string path)
             return NativeMethodsShared.FileOrDirectoryExists(path);
         }
 
+        public void ClearCaches() { }
```
dfederm, line 24: Why is this part of the interface if the impl is empty and it's never called?

Same with the write state method below

src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
```diff
@@ -21,6 +21,26 @@ public bool DirectoryEntryExists(string path)
...
+        public void WriteStatistics(TextWriter writer) { }
+
+        public FileAttributes GetAttributesNoMissingException(string path)
```
dfederm, line 27: This seems like an awkward API to expose

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -242,6 +242,22 @@ public partial interface IEventSource3 : Microsoft.Build.Framework.IEventSource,
         void IncludeEvaluationProfiles();
         void IncludeTaskInputs();
     }
+    public partial interface IFileSystem
+    {
+        void ClearCaches();
+        bool DirectoryEntryExists(string path);
```
dfederm, line 248: What's a "directory entry" vs just a directory?

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
```diff
@@ -242,6 +242,22 @@ public partial interface IEventSource3 : Microsoft.Build.Framework.IEventSource,
...
+        bool DirectoryExists(string path);
+        System.Collections.Generic.IEnumerable<string> EnumerateDirectories(string path, string searchPattern="*", System.IO.SearchOption searchOption=(System.IO.SearchOption)(0));
+        System.Collections.Generic.IEnumerable<string> EnumerateFiles(string path, string searchPattern="*", System.IO.SearchOption searchOption=(System.IO.SearchOption)(0));
+        System.Collections.Generic.IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern="*", System.IO.SearchOption searchOption=(System.IO.SearchOption)(0));
+        bool FileExists(string path);
+        System.IO.FileAttributes GetAttributesNoMissingException(string path);
+        System.IO.Stream GetFileStream(string path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share);
+        System.IO.TextReader ReadFile(string path);
+        byte[] ReadFileAllBytes(string path);
```
dfederm, line 257: This doesn't seem like something MSBuild needs?

dfederm:
In general you seem to just have taken Quickbuild's impl even the parts which make no sense for MSBuild.

You should cater MSBuild's interface for MSBuild alone. Adapters will be used to bridge the gap.

Forgind:
> @Forgind Nathan Mytelka FTE, @rainersigwald Rainer Sigwald FTE
> The PR fails this curious looking test:
> 
> https://github.com/dotnet/msbuild/blob/e8338f5ad91d42c3196218648859e9e837099842/src/Build.OM.UnitTests/NugetRestoreTests.cs#L21-L26
> 
> 
> What's the idea behind the test? We can't ever break that buggy old nuget version? :)

In short, yes. Some enterprise customers rely on an older version of NuGet that can't follow dependencies as effectively. They use reflection to load a solution as part of restore, so even if building works, it will throw an error before it gets there.

> The change that breaks it is that on master, IFileSystem is included in every MSBuild dll, but in this PR I am making it public, so I have to put it in only a single place as opposed to inserting it in every dll, otherwise C# fails with ambiguous typing errors. This breaks the old nuget version, but the latest nuget works.
> 
> How should I go around it? Can I just drop a binding redirect file next to the old nuget.exe to make it bind all msbuild assemblies to the correct ones (copying msbuild's binding redirect next to it makes it work)?

The old NuGet version didn't know that it needed to load Framework at all (for the restore—it clearly knows how to load it for building), so a binding redirect won't help. There can't be any reference to anything in Framework as part of loading a solution.

Forgind:
Nothing jumped out at me as obviously the point at which restore needed IFileSystem, but it's used in so many places, it's probably indirect. I was surprised FileUtilities isn't directly called from SolutionFile, but there's probably somewhere in ProjectInstance or SolutionFile that indirectly calls a relevant FileUtilities method or similar.

I don't know nearly enough about this to know if this even makes sense, but is there some way you can leave IFileSystem where it is but pass as a parameter to its constructor a delegate with the file system cache from visual studio/quickbuild baked into it?

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
cdmihai: An entry is either a file or a directory. Took the naming from [this BCL method](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefilesystementries?view=netcore-3.1). Is `FileOrDirectoryExists` better? Longer but explicit.

src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
cdmihai: Yea, it's pretty funky. I'll remove the part with the exception.

cdmihai:
@dfederm 
>In general you seem to just have taken Quickbuild's impl even the parts which make no sense for MSBuild.
You should cater MSBuild's interface for MSBuild alone. Adapters will be used to bridge the gap.

I think having a reasonably complete interface (enumerations, existence checks, reading, file attributes) would be good here:
- The interface won't be used by MSBuild alone. The new project cache plugins will use it, and I was thinking to also pass it to sdk resolvers which do a lot of IO.
- IFileSystem is relatively new to MSBuild. My intent was / is to slowly move as much msbuild IO as possible under this interface. Initially I mostly did evaluation (the convenient parts, and the profiling hot paths). So even if MSBuild does not use it much today, my intent is to have it use it more and more.

But I agree there's a fine line between what's useful and what's not.

src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
cdmihai: I also wasn't sure about this one, so left it for review comments :)

Having a method to clear caches would be useful in situations where replacing the entire IFileSsytem object is not possible (the msbuild xml cache suffers from this pain). But this is speculative, since no particular situation comes to mind.

ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
cdmihai: But it does seem useful for sdk resolvers and cache plugins which read files.

cdmihai:
@Forgind 

> Some enterprise customers rely on an older version of NuGet that can't follow dependencies as effectively. They use reflection to load a solution as part of restore, so even if building works, it will throw an error before it gets there.

The simplest solution here is to just break the things that reflect into non-public APIs :)
Is there some telemetry or other data we can use to learn how many users still use the old nuget? If they are inside Microsoft, we could push them to update. If they are outside, maybe they've updated in the meantime? Do you know a good contact to make such a query?

>The old NuGet version didn't know that it needed to load Framework at all (for the restore—it clearly knows how to load it for building), so a binding redirect won't help. There can't be any reference to anything in Framework as part of loading a solution.

The sdk resolver interfaces were put in Framework as well. They should have also broken the old nuget scenario. I say it's a good precedent to break it again :)

Possible ways to workaround this:
- try putting the interface in Utilities
- rename the public interface to something else (`IMSBuildFileSystem`) and write an adapter to the now legacy `IFileSystem`
- drop / postpone the PR. This isn't a blocker to project cache plugins as far as I can tell. But would have been a nice thing to have.

>is there some way you can leave IFileSystem where it is but pass as a parameter to its constructor a delegate with the file system cache from visual studio/quickbuild baked into it?

The delegate would have to reference a new type from Framework right? And the old nuget requirement is that one cannot add new apis to Framework.

Forgind:
> Is there some telemetry or other data we can use to learn how many users still use the old nuget? If they are inside Microsoft, we could push them to update. If they are outside, maybe they've updated in the meantime? Do you know a good contact to make such a query?

I think they were external, but @rrelyea would know more.

> The sdk resolver interfaces were put in Framework as well. They should have also broken the old nuget scenario. I say it's a good precedent to break it again :)

If they weren't required for restore, and they weren't referenced in a method required by restore, I think it was ok. There are certainly some things that can be in Framework and used by other parts of MSBuild that don't affect it.

> Possible ways to workaround this:
> 
> * try putting the interface in Utilities
> * rename the public interface to something else (`IMSBuildFileSystem`) and write an adapter to the now legacy `IFileSystem`
> * drop / postpone the PR. This isn't a blocker to project cache plugins as far as I can tell. But would have been a nice thing to have.

I don't remember if Utilities was loaded by the old NuGet, but it's worth a try. Option 2 is probably the easiest and most likely to work, but it's also a bit messy...

> The delegate would have to reference a new type from Framework right? And the old nuget requirement is that one cannot add new apis to Framework.

That isn't necessarily problematic, but I imagine you could use something like string -> object and cast the output appropriately. That should work without needing to refer to any MSBuild types in relevant code paths.

cdmihai:
I tried option 1 but that still broke the old nuget.exe, so I did option 2.

>If they weren't required for restore, and they weren't referenced in a method required by restore, I think it was ok. There are certainly some things that can be in Framework and used by other parts of MSBuild that don't affect it.

Sdk resolvers are required during restore because they are called during evaluation.

>something like string -> object and cast the output appropriately. That should work without needing to refer to any MSBuild types in relevant code paths.

IMHO that would leak too much implementation detail, raising maintenance pain in the future.

rrelyea:
@skofman and her team run nuget.org and could give data about how many users use different nuget.exe versions.

rainersigwald:
It's not just "how many people use old NuGet.exe". It's also "who are those people and do they know they're doing it?". The bug that caused that test to be written caused an Azure DevOps outage, because AzDO defaulted to an old version of NuGet. In addition, several customers had slightly-funky workflows like "restore all packages online, then disable the network and build stuff" that broke and didn't have an easy way to update NuGet versions.

I don't think we can break that version of NuGet/that test. Maybe for 17.0.

cdmihai:
>  It's also "who are those people and do they know they're doing it?"

Well, at some point they are going to break no matter what, so I'd rather have a slow trickle of breakages to keep the systems fit, well tested, and proactive. It's for their own good :). Joke aside, the workaround turned out to not be that horrible, as long as people follow the instructions in the interface comments (which realistically means that given enough time the interfaces will diverge, but that's a problem for future us).

cdmihai:
@dfederm Removed the obviously awkward APIs. Remaining ones seem reasonable to me for a file system API.

Forgind:
@cdmihai, our CIBuild isn't working for some reason for any PR. @BenVillalobos is looking into it.

BenVillalobos:

src/Build.UnitTests/Evaluation/ItemSpec_Tests.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using Microsoft.Build.Collections;
 using Microsoft.Build.Execution;
+using Microsoft.Build.Framework;
```
Forgind, line 8: This (and similar) isn't necessary anymore, right?

src/Build/Evaluation/Context/EvaluationContext.cs
```diff
@@ -17,13 +19,20 @@ namespace Microsoft.Build.Evaluation.Context
...
     {
         public enum SharingPolicy
         {
+            /// <summary>
+            /// Instructs the <see cref="EvaluationContext"/> to reuse state between the different project evaluations that use it.
+            /// </summary>
             Shared,
+
+            /// <summary>
+            /// Instructs the <see cref="EvaluationContext"/> to not reuse state between the different project evaluations that use it.
```
Forgind, line 34: nit:
```suggestion
            /// Instructs the <see cref="EvaluationContext"/> not to reuse state between the different project evaluations that use it.
```

src/Build/Evaluation/Context/EvaluationContext.cs
```diff
@@ -57,7 +66,29 @@ internal EvaluationContext(SharingPolicy policy)
         /// </summary>
         public static EvaluationContext Create(SharingPolicy policy)
         {
-            var context = new EvaluationContext(policy);
+            
+            // ReSharper disable once IntroduceOptionalParameters.Global
```
Forgind, line 70: I don't understand this comment. Disable ReSharper once IntroduceOptionalParameters.Global...something?

src/Build/FileSystem/MSBuildFileSystemAdapter.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Build.Shared.FileSystem;
+
+namespace Microsoft.Build.FileSystem
+{
+     public class MSBuildFileSystemAdapter : IFileSystem
```
Forgind, line 10: Should this implement IMSBuildFileSystem?

src/MSBuildTaskHost/FileSystem/MSBuildTaskHostFileSystem.cs
Forgind: I'm fine with leaving it or removing it, but if you leave it, I'd prefer for it to be implemented. I don't know if the intention is for a user to be able to call this via IFileSystem, but if so, they would be confused that it doesn't do anything. (And I would be, too, if I saw the method existed and tried to call it but couldn't figure out what exactly it did until I looked at it.)

src/Shared/FileSystem/IFileSystem.cs
```diff
@@ -6,39 +6,39 @@
 
 namespace Microsoft.Build.Shared.FileSystem
 {
-    /// <summary>
-    /// Abstracts away some file system operations
-    /// </summary>
+
+    /*
+     * This is a clone of Microsoft.Build.FileSystem.IMSBuildFileSystem.
```
Forgind, line 11: I noticed a reference to IFileSystem earlier. Made me wonder if it would be worth it to have something that detected references to "deprecated" APIs and gave a warning about them.

src/Shared/FileSystem/MSBuildOnWindowsFileSystem.cs
```diff
@@ -3,56 +3,78 @@
...
 using System.IO;
+using System.Reflection;
+using Microsoft.Build.Framework;
 
 namespace Microsoft.Build.Shared.FileSystem
 {
     /// <summary>
-    /// Implementation of file system operations directly over the dot net managed layer
+    /// Implementation of file system operations on windows. Combination of native and managed implementations.
+    /// TODO Remove this class and replace with WindowsFileSystem. Test perf to ensure no regressions.
```
Forgind, line 13: Looks like you modified WindowsFileSystem. What more has to be done to replace this?

src/Build/Evaluation/Context/EvaluationContext.cs
cdmihai: It disables a Resharper rule that wants to collapse the two Create methods into a single one with a null default for the file system. But that would break binary compat so I thought it would be safer to just have it disabled so other people don't run it at some later point in time.

src/Build/FileSystem/MSBuildFileSystemAdapter.cs
cdmihai: MSBuild uses IFileSystem internally (to avoid breaking the archaic yet oldie but goldie nuget version), so this adapter takes an IMSBuildFileSystem and makes it look like an IFileSystem, hence it extends IFileSystem.

src/Shared/FileSystem/IFileSystem.cs
cdmihai: IFileSystem will remain the internal interface that MSBuild uses until we can break that old nuget version.

src/Shared/FileSystem/MSBuildOnWindowsFileSystem.cs
cdmihai: Wait until nobody uses Windows anymore :)

Ideally this class should be implemented with Windows specific fast native APIs and used on Windows, instead of the presumably slower ManagedFileSystem. So somebody needs to make native implementations for the rest of the interface, test and prove better perf, and then delete `MSBuildOnWindowsFileSystem` in favor of just `WindowsFileSystem`.

src/Build/Evaluation/Context/EvaluationContext.cs
Forgind: I wanted to do something similar for a different method at one point. Someone had to explain to me why I couldn't. I was sad.

cdmihai: Such is the way of seeing sharp.

src/Shared/FileSystem/WindowsFileSystem.cs
```diff
@@ -22,50 +22,43 @@ internal enum FileArtifactType : byte
     }
 
     /// <summary>
-    /// Windows-specific implementation of file system operations using Windows native invocations
+    /// Windows-specific implementation of file system operations using Windows native invocations.
+    /// TODO For potential extra perf gains, provide native implementations for all IFileSystem methods and stop inheriting from ManagedFileSystem
     /// </summary>
-    internal class WindowsFileSystem : IFileSystem
+    internal class WindowsFileSystem : ManagedFileSystem
```
ladipro, line 28: nit: Should this class be sealed?

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1936,5 +1939,126 @@ public void Dispose()
...
+            }
+
+            private void IncrementCalls(ref int incremented)
+            {
+                Interlocked.Increment(ref incremented);
+            }
+
+            private void IncrementExistenceChecks(string path)
+            {
+                ExistenceChecks.AddOrUpdate(path, p => 1, (p, c) => c + 1);
```
ladipro, line 2060: Nice!

src/Build/FileSystem/MSBuildFileSystemAdapter.cs
ladipro: Why does the adapter class have to be public? Thanks!

cdmihai: Good catch! It shouldn't be.

cdmihai:
> How likely do you think it is that in the future we'll find ourselves wishing the interface had more and/or different methods? If such a need arises, is the versioning plan simply to add IMSBuildFileSystem2, IMSBuildFileSystem3, ...?

@ladipro I am hoping we'll be able to take advantage of the new C# default interface implementations. If that's not feasible (@rainersigwald?), then I would rather have them be abstract classes instead of interfaces.

src/Shared/FileSystem/WindowsFileSystem.cs
cdmihai: `WindowsFileSystem` extends it to override some implementations with native ones.

src/Shared/UnitTests/ObjectModelHelpers.cs
cdmihai: Not as cute as Python's [Counter](https://docs.python.org/3/library/collections.html#collections.Counter), but at least it's one short readable line.

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1936,5 +1939,126 @@ public void Dispose()
...
         }
+
+        internal class LoggingFileSystem : IMSBuildFileSystem
+        {
+            private readonly IFileSystem _wrappingFileSystem;
+            private int _fileSystemCalls;
+
+            public int FileSystemCalls => _fileSystemCalls;
+
+            public ConcurrentDictionary<string, int> ExistenceChecks { get; } = new ConcurrentDictionary<string, int>();
```
BenVillalobos, line 1950: I didn't know this was a thing 😮

src/Shared/UnitTests/ObjectModelHelpers.cs
```diff
@@ -1936,5 +1939,126 @@ public void Dispose()
                 _buildManager.Dispose();
             }
         }
+
+        internal class LoggingFileSystem : IMSBuildFileSystem
+        {
+            private readonly IFileSystem _wrappingFileSystem;
+            private int _fileSystemCalls;
```
BenVillalobos, line 1946: Out of curiosity, why are we only curious (for now) about how many file system calls there have been?

Should there be a test for verifying this increments properly?

cdmihai: This class is a mock class for the actual tests, no used in production. It would be a first in the repo to have tests for test helpers, so I'd rather not :)

src/Shared/UnitTests/ObjectModelHelpers.cs
cdmihai: Not to be confused with `=>`. Been there done that :)

src/Shared/UnitTests/ObjectModelHelpers.cs
cdmihai: If our issue count dips bellow 500 its time to rewrite the engine to make it saner.

ladipro (approved):
Looks good! We also talked about adding a method/methods to `MSBuildFileSystemBase` for retrieving file timestamps. Do you think it's worth doing now?

src/Shared/FileSystem/MSBuildOnWindowsFileSystem.cs
```diff
@@ -59,6 +60,11 @@ public FileAttributes GetAttributes(string path)
             return ManagedFileSystem.Singleton().GetAttributes(path);
         }
 
+        public DateTime GetLastWriteTimeUtc(string path)
+        {
+            return ManagedFileSystem.Singleton().GetLastWriteTimeUtc(path);
```
cdmihai, line 65: I'll replace this with WindowsFileSystem in a later PR when I actually hook this method up in the evaluator

src/Build.UnitTests/Definition/ProjectEvaluationContext_Tests.cs
```diff
@@ -86,6 +86,61 @@ public void SharedContextShouldGetReusedWhereasIsolatedContextShouldNot(Evaluati
...
+                            new Dictionary<string, int>
+                            {
+                                {Path.Combine(_env.DefaultTestDirectory.Path, "1.file"), 1},
+                                {Path.Combine(_env.DefaultTestDirectory.Path, "2.file"), 1}
+                            }.OrderBy(kvp => kvp.Key));
+                    fileSystem.DirectoryEntryExistsCalls.ShouldBe(2);
+                    break;
+
+                case EvaluationContext.SharingPolicy.Isolated:
+                    // in the isolated policy, only the first usage uses the given file system
```
rainersigwald, line 130: Discussed offline but for posterity:

It feels wrong to pass a filesystem to the context and pass the context in multiple places and have the filesystem used in only one of them. Can we instead error?

src/Build/Resources/Strings.resx
```diff
@@ -1831,4 +1831,7 @@ Utilization:          {0} Average Utilization: {1:###.0}</value>
   <data name="StaticGraphConstructionMetrics" xml:space="preserve">
     <value>"Static graph loaded in {0} seconds: {1} nodes, {2} edges"</value>
   </data>
+  <data name="IsolatedContextDoesNotSupportFileSystem" xml:space="preserve">
+    <value>"EvaluationContext objects created with SharingPolicy.Isolated do not support being passed a MSBuildFileSystemBase file system."</value>
```
rainersigwald, line 1835:
```suggestion
    <value>"EvaluationContext objects created with SharingPolicy.Isolated do not support being passed an MSBuildFileSystemBase file system."</value>
```

PR #5547, opened 2020-07-25
prevent exception for increasing priority

Forgind:
#5381 did not account for any possible priority class, and apparently customers are using priorities that are below BelowNormal but not Idle. This will more permanently resolve the issue.

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.ComponentModel;
 #if FEATURE_SYSTEM_CONFIGURATION
 using System.Configuration;
 #endif
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -627,11 +627,18 @@ namespace Microsoft.Build.CommandLine
 
                     // Honor the low priority flag, we place our selves below normal priority and let sub processes inherit
                     // that priority. Idle priority would prevent the build from proceeding as the user does normal actions.
-                    // We avoid increasing priority because that causes failures on mac/linux.
-                    if (lowPriority && Process.GetCurrentProcess().PriorityClass != ProcessPriorityClass.Idle)
+                    try
                     {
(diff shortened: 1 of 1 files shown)


src/MSBuild/XMake.cs
```diff
@@ -627,11 +627,18 @@ string commandLine
...
-                        Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
+                        if (lowPriority && Process.GetCurrentProcess().PriorityClass != ProcessPriorityClass.Idle)
+                        {
+                            Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
+                        }
                     }
+                    // We avoid increasing priority because that causes failures on mac/linux, but there is no good way to
+                    // verify that a particular priority is lower than "BelowNormal." If the error appears, ignore it and
+                    // leave priority where it was.
+                    catch (Exception) { }
```
BenVillalobos, line 640: Since this is a very specific case, can we catch the specific exception that's thrown? Otherwise there's the potential for missing some new exception that we didn't consider.

Forgind, line 640:
```suggestion
                    catch (Win32Exception) { }
```

src/MSBuild/XMake.cs
```diff
@@ -627,11 +628,18 @@ string commandLine
...
-                        Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
+                        if (lowPriority && Process.GetCurrentProcess().PriorityClass != ProcessPriorityClass.Idle)
+                        {
+                            Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
+                        }
                     }
+                    // We avoid increasing priority because that causes failures on mac/linux, but there is no good way to
+                    // verify that a particular priority is lower than "BelowNormal." If the error appears, ignore it and
+                    // leave priority where it was.
+                    catch (Win32Exception) { }
```
cdmihai, line 641: Might be a stupid question, but is it still a `win32` exception on non windows?

Forgind:
https://github.com/actions/virtual-environments/issues/1287 is one example of a win32 exception on ubuntu. I should actually make this PR close that issue.

PR #5549, opened 2020-07-27
Converting target framework intrinsic functions to use min version parts

sfoslund:
MSBuild side of https://github.com/dotnet/sdk/issues/12614


src/Build.UnitTests/Evaluation/Expander_Tests.cs
```diff
@@ -2890,7 +2891,8 @@ public void PropertyFunctionTargetFrameworkVersionMultipartParsing(string tfm, i
 
         [Theory]
         [InlineData("net5.0-windows10.1.2.3", 4, "10.1.2.3")]
-        [InlineData("net5.0-windows10.1.2.3", 2, "10.1")]
+        [InlineData("net5.0-windows10.1.2.3", 2, "10.1.2.3")]
+        [InlineData("net5.0-windows0.0.0.3", 2, "0.0.0.3")]
```
Forgind, line 2895: I think you account for it properly, but it would be nice to see a test with a 0 in the middle like 10.0.2.3.

PR #5552, opened 2020-07-28
Expand properties in SdkReference on evaluation

### LINKED ISSUE #5349: MSBuild SDKs version property is not expanded

andrew-boyarshin:
Fix #5349.

Unlike #5378 (which didn't fix #5349 but introduced a lot of refactorings around `SdkReference`), this doesn't do anything but implement the fix for #5349. #5378 is absolutely separate from this PR and the feature request #5349 in question.

I emphasize the simplicity of adding this feature, and the flexibility it provides.

### DIFF
--- a/src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
@@ -75,6 +75,7 @@ namespace Microsoft.Build.UnitTests.OM.Construction
         private const string SdkNameProperty = "$(" + SdkNamePropertyName + ")";
         private const string SdkVersionPropertyName = "MyTestSdkVersion";
         private const string SdkVersionProperty = "$(" + SdkVersionPropertyName + ")";
+        private const string SdkExpectedVersion = "42.42.42-local";
 
         public ProjectSdkImplicitImport_Tests()
         {
@@ -540,16 +541,81 @@ namespace Microsoft.Build.UnitTests.OM.Construction
         }
 
         [Theory]
-        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkName, MockExpandedSdkResolver.ExpectedVersion, false)]
-        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkName, SdkVersionProperty, true)]
-        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkVersionProperty, true)]
-        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkName, MockExpandedSdkResolver.ExpectedVersion, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkName, SdkExpectedVersion, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkName, SdkVersionProperty, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkVersionProperty, false)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkName, SdkExpectedVersion, true)]
         [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkName, SdkVersionProperty, true)]
         [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkNameProperty, SdkVersionProperty, true)]
-        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkName, MockExpandedSdkResolver.ExpectedVersion, false)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkName, SdkExpectedVersion, true)]
         [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkName, SdkVersionProperty, true)]
         [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkNameProperty, SdkVersionProperty, true)]
-        public void SdkPropertiesAreExpandedBeforeResolution(string projectTemplate, string sdkName, string sdkVersion, bool shouldLog4264)
+        public void SdkPropertiesAreExpandedBeforeResolution(string projectTemplate, string sdkName, string sdkVersion,
+            bool minimumVersionIsNonNull)
+        {
+            using var xmlReader = PrepareForSdkPropertiesAreExpandedTests(projectTemplate, sdkName, sdkVersion,
+                out var projectOptions);
+
+            var project = Project.FromXmlReader(xmlReader, projectOptions);
+
+            var expectedSdkPath = Path.GetDirectoryName(_sdkPropsPath);
+
+            // self-consistency check
+            expectedSdkPath.ShouldBe(Path.GetDirectoryName(_sdkTargetsPath));
+
+            var minimumVersion = minimumVersionIsNonNull ? string.Empty : null;
+            var expectedSdkReferenceRaw = new SdkReference(sdkName, sdkVersion, minimumVersion);
+            var expectedSdkReference = new SdkReference(SdkName, SdkExpectedVersion, minimumVersion);
+
+            project.Imports.Count.ShouldBe(2);
+
+            foreach (var import in project.Imports)
+            {
+                import.ImportingElement.SdkReference.ShouldBe(expectedSdkReferenceRaw);
+                import.SdkResult.Success.ShouldBeTrue();
+                import.SdkResult.SdkReference.ShouldBe(expectedSdkReference);
+                import.SdkResult.Path.ShouldBe(expectedSdkPath);
+                import.SdkResult.Version.ShouldBe(expectedSdkReference.Version);
+            }
+        }
+
+        [Theory]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkVersionProperty, false, true)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkVersionProperty, true, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkVersionProperty, false, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkName, SdkVersionProperty, true, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkName, SdkVersionProperty, false, false)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkExpectedVersion, false, true)]
+        [InlineData(ProjectTemplateSdkAsAttributeWithVersion, SdkNameProperty, SdkExpectedVersion, false, false)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkNameProperty, SdkVersionProperty, false, true)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkNameProperty, SdkVersionProperty, true, false)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkNameProperty, SdkVersionProperty, false, false)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkName, SdkVersionProperty, true, false)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkName, SdkVersionProperty, false, false)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkNameProperty, SdkExpectedVersion, false, true)]
+        [InlineData(ProjectTemplateSdkAsElementWithVersion, SdkNameProperty, SdkExpectedVersion, false, false)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkNameProperty, SdkVersionProperty, false, true)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkNameProperty, SdkVersionProperty, true, false)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkNameProperty, SdkVersionProperty, false, false)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkName, SdkVersionProperty, true, false)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkName, SdkVersionProperty, false, false)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkNameProperty, SdkExpectedVersion, false, true)]
+        [InlineData(ProjectTemplateSdkAsExplicitImportWithVersion, SdkNameProperty, SdkExpectedVersion, false, false)]
+        public void SdkPropertiesAreExpandedBeforeResolutionFailWhenPropertyNotSet(string projectTemplate,
+            string sdkName, string sdkVersion, bool setName, bool setVersion)
+        {
+            using var xmlReader = PrepareForSdkPropertiesAreExpandedTests(projectTemplate, sdkName, sdkVersion,
+                out var projectOptions, setName, setVersion);
+
+            projectOptions.LoadSettings |= ProjectLoadSettings.IgnoreMissingImports;
+
+            var project = Project.FromXmlReader(xmlReader, projectOptions);
+
+            project.Imports.Count.ShouldBe(0);
+        }
+
+        private XmlReader PrepareForSdkPropertiesAreExpandedTests(string projectTemplate, string sdkName, string sdkVersion,
+            out ProjectOptions projectOptions, bool setName = true, bool setVersion = true)
         {
             _env.SetEnvironmentVariable("MSBuildSDKsPath", _testSdkRoot);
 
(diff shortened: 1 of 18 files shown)


andrew-boyarshin:
Uhm... My added test (`Microsoft.Build.UnitTests.OM.Construction.ProjectSdkImplicitImport_Tests.SdkPropertiesAreExpandedBeforeResolution`) is failing on CI, but passing locally.
image
image

Where do the extra messages come from? How can I debug this? Trying to add more logging now...

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1790,8 +1791,53 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+
+                    var expandedSdkReference = new SdkReference(
+                        EvaluateProperty(sdkReference.Name, sdkReferenceOrigin, _expander),
+                        EvaluateProperty(sdkReference.Version, sdkReferenceOrigin, _expander),
+                        EvaluateProperty(sdkReference.MinimumVersion, sdkReferenceOrigin, _expander)
+                    );
+
+                    if (!Equals(expandedSdkReference, sdkReference))
+                    {
+                        var eventArgs = new ProjectImportedEventArgs(
```
cdmihai, line 1820: AFAIK msbuild does not log expansions, in which case it's probably best to be consistent and not log this.s

andrew-boyarshin: But MSBuild logs e.g. property overrides, which really helps debug cases when something expected is overridden with sth unexpected. I can remove this, of course (that would also drop the new string in resx), but I tried to make this a bit more debuggable.

andrew-boyarshin:
I've rebased my branch on latest master and applied suggestion from @cdmihai (removed logging). I've also fixed confirmation biased tests by introducing tests that actually test failure cases.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1790,8 +1791,33 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                    static string EvaluateProperty(string value, IElementLocation location, Expander<P, I> expander) =>
+                        value != null
+                            ? expander.ExpandIntoStringAndUnescape(value,
+                                                                   ExpanderOptions.ExpandProperties,
+                                                                   location)
+                            : null;
+
+                    var sdkReferenceOrigin = importElement.SdkLocation;
+
+                    sdkReference = new SdkReference(
```
cdmihai, line 1812: Before you had a check to only create a `new SdkReference` if expansion changed the contents of the original sdk reference string. I think that's a sensible check to keep.

andrew-boyarshin: The check below that I've removed only influenced logging the event. It didn't influence overriding `sdkReference` variable. When `sdkReference` contains something that can theoretically be expanded (`$(` heuristic above), it is always created anew from the result of the expansion.

src/Build/Evaluation/Evaluator.cs
cdmihai: I agree with the debuggability part, however the following arguments make me not want the log event for this case:
- other expansions are not logged. If we decide to do something else, which I agree might improve debuggability, I'd rather we have a unified approach that logs all expansions. That's a bigger project for a different PR, and needs careful design to avoid perf issues or log size inflation.
- the expanded sdk reference information already kinda gets indirectly logged when the import on the expanded sdk reference gets done. There is room for improvement here: the ProjectImportedEventArgs does not contain information on whether the import was an sdk import and if it was, what the SdkReference contents was. It's a somewhat unrelated change, if you do it you can open a new PR with it, rebased on this current PR, and we'll merge them in cadence. But I think it would be super useful for debugging to be able to search the binlog for all the sdk based imports.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1779,7 +1779,8 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
 
             string project = importElement.Project;
 
-            if (importElement.SdkReference != null)
+            var sdkReference = importElement.SdkReference;
```
Forgind, line 1782: nit:
Real type

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1790,8 +1791,33 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
                 if (solutionPath == "*Undefined*") solutionPath = null;
                 var projectPath = _data.GetProperty(ReservedPropertyNames.projectFullPath)?.EvaluatedValue;
 
+                var compareInfo = CultureInfo.InvariantCulture.CompareInfo;
```
Forgind, line 1794: nit:
real type

src/Build/Evaluation/Evaluator.cs
cdmihai: Ah, yes, good point.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1790,8 +1791,33 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                static bool HasProperty(string value, CompareInfo compareInfo) =>
+                    value != null && compareInfo.IndexOf(value, "$(") != -1;
+
+                if (HasProperty(sdkReference.Name, compareInfo) ||
+                    HasProperty(sdkReference.Version, compareInfo) ||
+                    HasProperty(sdkReference.MinimumVersion, compareInfo))
+                {
+                    static string EvaluateProperty(string value, IElementLocation location, Expander<P, I> expander) =>
+                        value != null
+                            ? expander.ExpandIntoStringAndUnescape(value,
```
Forgind, line 1805: I could somewhat imagine this being problematic if you want your name to include a %, and it wasn't escaping it previously. If your name includes $(, you're probably asking for it if you don't expect it to be expanded. Is it feasible for it to be a path intended for use as a key at some point? Because ExpandIntoStringAndUnescape also fixes file paths.

andrew-boyarshin: You propose using `ExpandIntoStringLeaveEscaped` instead, right? So that it would preserve special characters, like `%`, not treat them like a part of escaped character sequence. I'm not sure about this part, I leave it for MSBuild team to decide. If you think it's better to use the other option, I'll gladly make the change.

cdmihai: @Forgind, trying to summarize the failure cases. Let me know if I've missed any:
- user has a `$(` which they don't want expanded, but now it will get expanded. The traditional approach here is for the user to escape any msbuild syntax when they want it as a literal, which will still work here.
- on non windows systems, the `$(` is [used as a guard to prevent adjusting path like strings](https://github.com/cdmihai/msbuild/blob/9c2d922959e80ba72fefb7addec21ede227639c3/src/Shared/FileUtilities.cs#L454). So unix based users with strings that contained both `$(` and backslashes in their sdk references, will now get their slashes changed. I think this is still fine:
  - first, `$(` and slashes would constitute illegal nuget names and versions, so their builds would have probably crashed anyway had they tried to put these in nuget based sdks.
  - second, even though backslashes are valid unix filename characters, AFAIK their usage is not common. And even if the user did have backslashes in their name / version, they can escape them. If this gets to be a common case (and thus make escaping bad for usability) we could then add an option to not fix the file path when expanding the sdk reference (my guess is it won't happen).

Forgind: These are certainly all things the user can get out of pretty easily; I'm just trying to avoid having code that works with one MSBuild version, then fails with the next. I was also thinking of the failure case @andrew-boyarshin mentioned with %. The most odd part to me is if you have, say, a name with an escape character but no properties. Then you add a property to the version, and the name is escaped. That also applies to paths, so you noted that slashes would be illegal—are quotes? Since MaybeAdjustFilePath also removes quotes if present. I think that was all I was thinking. I'm not too worried about most properties being expanded, but it did occur to me that if someone uses the guard against paths being adjusted, it might look like a property and be converted to "" when it can't be found. That's the worst case I can imagine.

cdmihai: > trying to avoid having code that works with one MSBuild version, then fails with the next

If it does break anybody, it's probably in their own interest to fix it, since the correct behaviour from the very start should have been to expand the sdk reference. So they'd be migrating off of relying on a bug.

However, in case it does break anybody, it probably makes sense to enable the expansion by default but provide an escape hatch so if anybody does break, they can revert to the old behaviour. @andrew-boyarshin, can you please add an escape hatch [here](https://github.com/dotnet/msbuild/blob/f365aa332f58cf43e78d76491438f15fc1988f27/src/Shared/Traits.cs#L110), and disable sdk reference expansion if the escape hatch is enabled?

>MaybeAdjustFilePath also removes quotes if present

I don't think it does. The results from RemoveQuotes does not seem to flow into the return of MaybeAdjustFilePath

>a name with an escape character but no properties. Then you add a property to the version, and the name is escaped.

This breaking scenario [is constrained ](https://github.com/cdmihai/msbuild/blob/9c2d922959e80ba72fefb7addec21ede227639c3/src/Deprecated/Engine/Shared/EscapingUtilities.cs#L65-L66) only to strings that have `%` followed by two digits. In which case (beyond the fact nuget would probably fail on it), the user can use the escape hatch.

andrew-boyarshin:
```xml
<Project Sdk="MSBuild.Sdk.Extras/$(MySdkVersion)">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
    <RootNamespace>msbuild_test</RootNamespace>
  </PropertyGroup>

</Project>
```

image
image
image
image
image
image
image

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1790,8 +1792,58 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+
+                            const ExpanderOptions Options = ExpanderOptions.ExpandProperties;
+
+                            switch (mode)
+                            {
+                                case SdkReferencePropertyExpansionMode.ExpandUnescape:
+                                    return expander.ExpandIntoStringAndUnescape(value, Options, location);
+                                case SdkReferencePropertyExpansionMode.ExpandLeaveEscaped:
+                                    return expander.ExpandIntoStringLeaveEscaped(value, Options, location);
+                                case SdkReferencePropertyExpansionMode.NoExpansion:
```
Forgind, line 1827: Can you add test cases for the new modes? In particular, without a break in here, I believe it will fall through on the `NoExpansion` and `DefaultExpand` cases to throw an error.

src/Build/Evaluation/Evaluator.cs
```diff
@@ -1790,8 +1792,58 @@ private List<ProjectRootElement> ExpandAndLoadImports(string directoryOfImportin
...
+                                case SdkReferencePropertyExpansionMode.DefaultExpand:
+                                default:
+                                    ErrorUtilities.ThrowArgumentOutOfRange(nameof(mode));
```
Forgind, line 1830: This should never happen, right? Maybe throw an internal error rather than ArgumentOutOfRange.

src/Build/Evaluation/Evaluator.cs
andrew-boyarshin: But that's intentional. Both cases are unreachable here, and must cause an internal assertion error. I've listed all enum values to make it explicit and make it more future-proof in case new enum values are added.

andrew-boyarshin: Additionally, I'm not sure I can test it. EscapeHatches do the read only once in a process lifetime, so one must make sure that each test restarts the test host. Is there any existing example in MSBuild tests on how to do it?

src/Build/Evaluation/Evaluator.cs
andrew-boyarshin: Initially I had a `switch` expression that was throwing `ArgumentOutOfRangeException`. After I've discovered there is a matching `ErrorUtilities` method, I've rewritten it to be a `switch` statement and use that. Rationale: however unlikely case this is, it's better to have an opportunity to prevent throwing and use a fallback.

src/Build/Evaluation/Evaluator.cs
andrew-boyarshin: Okay, it seems xUnit unloads and reloads the assemblies for each subtest, so the EscapeHatches takes the correct value on each subtest..

andrew-boyarshin:
I've greatly rewritten tests so that both success and failure cases are tested within a single parametrized test method. I've also increased the number of tested value combinations: from 30 manually specified cases up to 240 automatically-generated ones.

src/Build/Evaluation/Evaluator.cs
Forgind: Oh, missed that. You're right.

src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
```diff
@@ -534,6 +542,257 @@ public void IgnoreMissingImportsSdkNotFoundDoesNotThrow()
...
+                if (SetVersionProperty)
+                    result.Append(", SetVersion");
+                if (ExpectedSuccess)
+                    result.Append(", ExpectedSuccess");
+
+                return result.ToString();
+            }
+        }
+
+        [Theory]
```
Forgind, line 609: nit:
Does this still need to be a theory?

andrew-boyarshin: Well, since it is a parametrized test, it has to be a `Theory`, not a `Fact`.

Forgind: I was thinking there's only one input used with the test, so you could inline it.

andrew-boyarshin: You propose inlining the whole set of subtests from `SdkPropertiesAreExpandedData` into this method? I'm not sure why is that better than having xUnit do the full stats per-subtest.

andrew-boyarshin: I'm also not sure what you mean by "one input". That's 240 different dynamically-generated inputs.

Forgind: I should have said the data come from one place. That means that converting it to a Fact would mean a single for loop, which isn't too messy. I normally think theories are better if I would need to specify a lot of data in the test to keep it a Fact. In this case, you could remove the `new object[]` part in SdkPropertiesAreExpandedData and make it a simple loop through all the Cases, which isn't ugly.

Forgind: In any case, it isn't bothering me too much, so you can ignore me if you want.

andrew-boyarshin: 1. Will setting the environment variable still work? The value in `EscapeHatches` is set on first access, it can't be changed later, so each change in Mode requires at least reloading the assemblies (if not restarting the test host altogether).
2. Why reimplement the xUnit parametrized test? What's the expected behavior if one subtest fails? Assert will stop the iteration and the log will contain only the fact that something has failed, not the data on which it failed or the amount of successful/failed subtests.

I strongly oppose the suggested inlining.

Forgind: I actually don't know about (1), so if it doesn't, you're probably right. I would have thought `var project = Project.FromXmlReader(xmlReader, projectOptions);` is the point at which the environment variable needs to have changed, since that simulates making a new project.
For (2), I tend to debug into them the same, so it doesn't really matter to me.

You feel far more strongly than I do, so it's fine to leave it as-is. I already approved.

src/Build.UnitTests/Evaluation/ProjectSdkImplicitImport_Tests.cs
```diff
@@ -534,6 +542,257 @@ public void IgnoreMissingImportsSdkNotFoundDoesNotThrow()
...
+                static IEnumerable<SdkReferencePropertyExpansionMode?> Modes()
+                {
+                    yield return null;
+                    yield return SdkReferencePropertyExpansionMode.NoExpansion;
+                    yield return SdkReferencePropertyExpansionMode.DefaultExpand;
+                    yield return SdkReferencePropertyExpansionMode.ExpandUnescape;
+                    yield return SdkReferencePropertyExpansionMode.ExpandLeaveEscaped;
+                }
+
+                static IEnumerable<(SdkPropertiesAreExpandedDataTemplate, bool, bool)> Templates()
```
cdmihai, line 696: nit: name the tuple elements to self document the two bools

andrew-boyarshin: I had them initially documented, but then the hard wrap at 120 characters made the function definition look bad, and I decided since this method is used only once (and can only be used within this property), and the return value is immediately deconstructed with meaningful variable names, this can be left with default tuple field names. I can add the tuple field names, but this'll make method definition look worse. Your call.

cdmihai: As middle ground you can only name the two bools (`setName, setVersion`, not that long) and leave the first tuple element unnamed, since the type is expressive already.

BenVillalobos:

PR #5561, opened 2020-07-29
Added untriaged labels to all new issues

BenVillalobos:
This should help with our bug triage meeting.


Forgind:
Do you think it might be worth it to add a "doesn't fit any other category" type of issue?

Issue #5454, opened 2020-06-23
Detect macOS in Big Sur-compatible way


rainersigwald:
https://github.com/microsoft/msbuild/blob/6019483f9105a83f8a5bf1fb781ff4a5091f1df5/src/Shared/NativeMethodsShared.cs#L636-L647

This will stop working in the next version of macOS:

> New in macOS Big Sur 11 beta, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to `dlopen()` the path, which will correctly check for the library in the cache. (62986286)

(@akoeplinger pointed this out offline)

akoeplinger:
I just tried on an macOS 11 installation and it looks like `/usr/lib/libc.dylib` is still there, even though with a broken symlink:
```
/usr/lib/libc.dylib -> /usr/lib/libSystem.dylib -> libSystem.B.dylib (doesn't exist)
```

The File.Exists() check still returns true though so we should be good for now. There's still the possibility that Apple just didn't remove this .dylib in the first Beta though so we need to keep an eye on it.

rainersigwald:
We should probably just call [`RuntimeInformation.IsOSPlatform(OSPlatform.OSX)`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=netcore-3.1) anyway; I don't think there's a need to roll it our own way any more.

akoeplinger:
Agreed.

dalexsoto:
@rainersigwald / @akoeplinger  looks like `libc` is now gone for good in macOS 11.0 Beta 3 and it is causing issues like these https://gist.github.com/dalexsoto/9e8f2c804ec7ba2adac36360d1059ea4

> alex@iDxBS ~ % ls -l /usr/lib/libc.dylib
> ls: /usr/lib/libc.dylib: No such file or directory

akoeplinger:
Yeah, confirmed it is gone on my macOS 11 beta 3 too...

PR #5568, opened 2020-07-30
Mono: use RuntimeInformation.IsOSPlatform to check for macOS

akoeplinger:
Follows what we already do for Windows.

Fixes https://github.com/dotnet/msbuild/issues/5454

### DIFF
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -627,28 +627,25 @@ namespace Microsoft.Build.Shared
 #endif
         }
 
-#if MONO
+#if !CLR2COMPATIBILITY
         private static bool? _isOSX;
 #endif
+
         /// <summary>
         /// Gets a flag indicating if we are running under Mac OSX
         /// </summary>
         internal static bool IsOSX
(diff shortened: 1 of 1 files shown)


src/Shared/NativeMethodsShared.cs
```diff
@@ -627,28 +627,25 @@ internal static bool IsWindows
...
+#if !CLR2COMPATIBILITY
         private static bool? _isOSX;
 #endif
+
         /// <summary>
         /// Gets a flag indicating if we are running under Mac OSX
         /// </summary>
         internal static bool IsOSX
         {
-#if MONO
```
Forgind, line 638: Can Mono never CLR2 compatible? If it can be, might need an extra case for this.

cdmihai: The compat check is there because that old CLR didn't have nullable structs, the diff makes sense.

cdmihai: The CLR2 check is there because that version does not have RuntimeInformation, switching from Mono to that makes sense.

Issue #4906, opened 2019-11-09
Nodemode failure


Forgind:
There seems to be a problem with nodemode (an internal flag) without an argument.
I set MSBUILDDEBUGONSTART to 1 and built a project that referenced two other libraries using the command:
`<path to MSBuild.exe> <path to project file> /m /nodemode /nodereuse:false`

Expected behavior: useful error message about how nodemode works.
Actual behavior: unexpected bug with a message asking me to file a bug on GitHub.

This should not be an important issue because is has minimal customer impact, but it still might be nice to have a better error message.

abc516:
Hi @Forgind I'm willing to take this on. Any tips on how to approach?

Forgind:
Thanks @abc516!

It turns out this was resolved by https://github.com/dotnet/msbuild/pull/5569, but we didn't realize it until I tried to reproduce this issue just now. Let me know if there are any others you'd like to try!

PR #5569, opened 2020-07-30
Fix watson crashes in StartLocalNode.  

marcpopMSFT:
Customer may be trying to call this without setting a node or setting a value of 1 both which will trigger an index out of range exception.
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1157058

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -2624,7 +2624,8 @@ namespace Microsoft.Build.CommandLine
 
                         // If FEATURE_NODE_REUSE is OFF, just validates that the switch is OK, and always returns False
                         bool nodeReuse = ProcessNodeReuseSwitch(commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.NodeReuse]);
-                        bool lowpriority = commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.LowPriority][0].Equals("true");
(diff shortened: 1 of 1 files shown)


src/MSBuild/XMake.cs
```diff
@@ -2624,7 +2624,7 @@ private static void StartLocalNode(CommandLineSwitches commandLineSwitches)
 
                         // If FEATURE_NODE_REUSE is OFF, just validates that the switch is OK, and always returns False
                         bool nodeReuse = ProcessNodeReuseSwitch(commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.NodeReuse]);
-                        bool lowpriority = commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.LowPriority][0].Equals("true");
+                        bool lowpriority = commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.LowPriority].Length > 0 ? commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.LowPriority][0].Equals("true") : false;
```
Forgind, line 2627: Might be good to factor out commandLineSwitches[CommandLineSwitches.ParameterizedSwitch.LowPriority]. Also, if they use it, but they use it wrong, do you think it's more likely they were trying to turn it on or off?

Forgind:
Turns out this resolved #4906, too. Thanks!

PR #5571, opened 2020-07-31
Revert "Add ability to set a help link for errors/warnings (#5488)"

BenVillalobos:
This reverts commit 8343d638eaad09f38fa145b3fc98a2a316d64512.

Which is very clearly causing cloudbuild failures:
```
Q:\cmd\b\src\venus\project\webapp\tasks\GetSilverlightItemsFromProperty.cs(114,21): error CS0121: The call is ambiguous between the following methods or properties: 'TaskLoggingHelper.LogError(string, string, string, string, int, int, int, int, string, params object[])' and 'TaskLoggingHelper.LogError(string, string, string, string, int, int, int, int, string, string, params object[])' 
Q:\cmd\b\src\venus\project\webapp\tasks\CopyFilesToFolders.cs(148,25): error CS0121: The call is ambiguous between the following methods or properties: 'TaskLoggingHelper.LogError(string, string, string, string, int, int, int, int, string, params object[])' and 'TaskLoggingHelper.LogError(string, string, string, string, int, int, int, int, string, string, params object[])' 
Log location: \\BN01APB8F60E51D\d$\dbs\sh\ddvsm\0731_134247_0\src\venus\project\webapp\tasks\Logs\Retail\X86 

```


Forgind:
Fortunately, we don't need this anymore.

PR #5574, opened 2020-08-01
use nameof instead of string

### LINKED ISSUE #5573: Use nameof instead of string
There are many locations where inline strings are used when using nameof would be more appropriate and less error prone.

elachlan:
Fixes #5573.


BenVillalobos:

PR #5585, opened 2020-08-02
Use 'is' operator instead of 'as' operator

### LINKED ISSUE #5584: Use 'is' operator instead of 'as' operator
There are two that I can find.

found using Roslynator
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1172.md

elachlan:
Fixes #5584

### DIFF
--- a/src/Build/Instance/TaskFactories/TaskHostTask.cs
+++ b/src/Build/Instance/TaskFactories/TaskHostTask.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Build.BackEnd
 
                 // If we returned an exception, then we want to throw it when we 
                 // do the get.  
-                if (value != null && value is Exception)
+                if (value is Exception)
                 {
                     throw (Exception)value;
                 }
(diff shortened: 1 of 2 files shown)


src/Build/Instance/TaskFactories/TaskHostTask.cs
```diff
@@ -206,7 +206,7 @@ public object GetPropertyValue(TaskPropertyInfo property)
 
                 // If we returned an exception, then we want to throw it when we 
                 // do the get.  
-                if (value != null && (value as Exception) != null)
+                if (value != null && value is Exception)
```
Forgind, line 209: nit:
```suggestion
                if (value is Exception)
```

src/Build/Instance/TaskFactories/TaskHostTask.cs
```diff
@@ -206,7 +206,7 @@ public object GetPropertyValue(TaskPropertyInfo property)
 
                 // If we returned an exception, then we want to throw it when we 
                 // do the get.  
-                if (value != null && (value as Exception) != null)
+                if (value is Exception)
                 {
                     throw (Exception)value;
                 }
```
ladipro, line 212: nit: I know it's already merged. Just in case you're planning to submit more refactoring PRs, this could have been:

```C#
                if (value is Exception exception)
                {
                    throw exception;
                }
```

Thank you for the contributions!

elachlan: I plan on going through every refactor suggestion. This is one of them that is in the works :)

PR #5588, opened 2020-08-02
Use 'string.IsNullOrEmpty' method

### LINKED ISSUE #5587: Use 'string.IsNullOrEmpty' method
There are several areas in code where a null check and a length check are done, where 'string.IsNullOrEmpty' would be more appropriate and less verbose.

found using Roslynator
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1113.md

elachlan:
Fixes #5587


elachlan:
@Forgind this needs test re-run.

PR #5590, opened 2020-08-02
'null' should be on the right side of a binary expression.

### LINKED ISSUE #5589: 'null' should be on the right side of a binary expression.
For better readability null should be on the right side of a binary expression.

https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1098.md

elachlan:
Fixes #5589


elachlan:
One thing i'd like to point out is that readability has been reduced on some of the tests where the left nulls were aligned. So it isn't as easy to gather whats going on at a glance. But it does make more sense with this change.

elachlan:
@Forgind it would be great to get this merged before I start on the parentheses PR.

Forgind:
Team triage:
We talked about this PR with regard to the normal benefit of readability and the negative of making `git blame` harder and general churn plus the readability you noted with alignment in tests. If you find a rule in dotnet/runtime's coding styles (either by finding one that exists or adding a new one) that says this is beneficial, we would take it.

elachlan:
> Team triage:
> We talked about this PR with regard to the normal benefit of readability and the negative of making `git blame` harder and general churn plus the readability you noted with alignment in tests. If you find a rule in dotnet/runtime's coding styles (either by finding one that exists or adding a new one) that says this is beneficial, we would take it.

I have no skin in the game, so its really up to the team if they want it or not.

This might help:
https://github.com/JosefPihrt/Roslynator/issues/183

elachlan:
I have created an issue in dotnet/runtime asking about the code-style change.
https://github.com/dotnet/runtime/issues/40561

BenVillalobos:

rainersigwald:
Thanks, @elachlan!

PR #5602, opened 2020-08-02
Convert switch statement to switch expression

elachlan:
Refactor to convert switch statements to switch expressions to simplify code.

### DIFF
--- a/src/Build/BackEnd/Components/Scheduler/Scheduler.cs
+++ b/src/Build/BackEnd/Components/Scheduler/Scheduler.cs
@@ -1287,8 +1287,7 @@ namespace Microsoft.Build.BackEnd
                 return false;
             }
 
-            int limit = 0;
-            limit = _componentHost.BuildParameters.MaxNodeCount switch
+            int limit = _componentHost.BuildParameters.MaxNodeCount switch
             {
                 1 => 1,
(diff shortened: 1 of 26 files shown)


src/Build/BackEnd/Components/Scheduler/Scheduler.cs
```diff
@@ -1288,20 +1288,12 @@ private bool AtSchedulingLimit()
             }
 
             int limit = 0;
-            switch (_componentHost.BuildParameters.MaxNodeCount)
+            limit = _componentHost.BuildParameters.MaxNodeCount switch
```
Forgind, line 1291: Can you also say `int limit = _componentHost...` and avoid the definition above?

src/Build/BackEnd/Components/SdkResolution/SdkResolverManifest.cs
```diff
@@ -31,21 +31,15 @@ internal static SdkResolverManifest Load(string filePath)
             {
                 while (reader.Read())
                 {
-                    switch (reader.NodeType)
+                    return reader.NodeType switch
```
Forgind, line 34: I feel like this should have been an if-else in the first place.

BenVillalobos:

PR #5606, opened 2020-08-02
Simplify boolean comparison

elachlan:
Refactor to simplify boolean comparison using Roslynator.

https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1049.md


elachlan:
> I believe none of these can secretly be bool? (and thus necessary) because then it wouldn't even compile, correct?

That is my understanding.

PR #5609, opened 2020-08-03
Enable manifest generation in a .NET Core process

NikolaMilosavljevic:
This PR enables manifest generation in a .NET (Core) process.

Some APIs, classes and concepts are not available on .NET Core - new code had to be written to support existing code paths.

Examples:
- Partial and custom trust are not supported on .NET Core
- Permission sets are not supported on .NET Core
- Metadata reader now uses native .NET (Core) support - old PInvokes are now missing
- Minor updates in Cryptography and COM types to match changes in .NET Core
- While most of the manifest generation code is now cross-plat, some codepaths still require Windows platform, i.e. manifest signing


eng/Packages.props
```diff
@@ -35,9 +35,12 @@
     <PackageReference Update="System.Reflection.TypeExtensions" Version="4.1.0" />
     <PackageReference Update="System.Resources.Extensions" Version="4.6.0" />
     <PackageReference Update="System.Resources.Writer" Version="4.0.0" />
+    <PackageReference Update="System.Runtime.InteropServices" Version="4.3.0" />
```
cdmihai, line 38: @forgind do these versions have to be synced with nuget, the sdk, or any other tool in the .net core toolchain?

cdmihai:
@NikolaMilosavljevic, as far as I can tell there are no tests for these tasks in the msbuild repo. Are there tests anywhere else that exercise these tasks **under .net core**? If there are, did you run them and they pass? In general, where do they live and when do they get automatically run?

eng/Packages.props
Forgind: @wli3? I don't think we have to worry about the .net core toolchain if we can insert into VS, but we should make sure that works. If someone has already used a different version, it's easiest to switch to theirs to avoid a mismatch if possible; if no one has used it, we'd have to add it, which can be tough.

NikolaMilosavljevic:
> @NikolaMilosavljevic, as far as I can tell there are no tests for these tasks in the msbuild repo. Are there tests anywhere else that exercise these tasks **under .net core**? If there are, did you run them and they pass? In general, where do they live and when do they get automatically run?

I don't believe there are tests - @sujitnayak do you know? Are there plans to add them?

Visual Studio uses .NET FX version of these tasks/APIs. .NET FX used the same version in the old versions of manifest generation tool. New tool will run on .NET 5 and is the only one that will use the newly ported tasks/APIs.

I have run an extensive batch of manual tests, diffing older outputs with new ones and ensuring new scenarios work as expected. There are differences in support due to some dependent types and concepts that are missing on .NET (Core).

Changes I'm making have zero impact on .NET FX version of these tasks.

eng/Packages.props
NikolaMilosavljevic: Version of System.Runtime.InteropServices (4.3.0) is the latest available, built in 2016 - all projects seem to reference that one.

System.Security.Cryptography.Pkcs and System.Security.Cryptography.Xml - versions 4.7.0 are from .NET Core 3.1 release (LTS).

I see references to these two versions for several other packages.

cdmihai:
Sounds good. I was trying to understand the landscape of current and future regressions, and the chances of detecting them before they hit the users. Sounds like:
- it won't break existing scenarios
- manual tests for new scenarios
- not clear how to catch any regressions the MSBuild team might accidentally cause to the click once tasks, but I guess that has been the case up until now too.

Forgind:
[Insertion into VS](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/265308) (internal link) passed RPS.

BenVillalobos:

Issue #5620, opened 2020-08-07
MSBuild repo doesn't build without 3.5 SP1 Targeting Pack


KirillOsenkov:
On a clean Windows machine with VS 2019 16.6 where 3.5 SP1 targeting pack was never installed, this very MSBuild repo fails to build with:

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\
Bin\Microsoft.Common.CurrentVersion.targets(1177,5): error MSB3645: .NET Framew
ork v3.5 Service Pack 1 was not found. In order to target ".NETFramework,Versio
n=v3.5", .NET Framework v3.5 Service Pack 1 or later must be installed. [C:\MSB
uild\src\MSBuildTaskHost\MSBuildTaskHost.csproj]
```

When trying to add https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/1.0.0 I've found out that it omits net35 from the list of frameworks it supports. See https://github.com/dotnet/installer/issues/2022 for details.

Fortunately, our friend @jnm2 went out of his way to fill this gap and created this awesome package: https://www.nuget.org/packages/jnm2.ReferenceAssemblies.net35

However I found that it still doesn't work because of this check that checks for the "sentinel" System.Data.Entity.dll being installed in the GAC:
https://github.com/dotnet/msbuild/blob/e8c17c128ed847d4acd278b42a02df9621bd1652/src/Tasks/GetReferenceAssemblyPaths.cs#L160-L183

KirillOsenkov:
Aha, I see they left us an escape hatch! Setting BypassFrameworkInstallChecks lets us bypass that entire block!

PR #5621, opened 2020-08-07
Enable MSBuild building without 3.5 SP1 targeting pack installed.

KirillOsenkov:
Fixes https://github.com/dotnet/msbuild/issues/5620.

Unfortunately the Microsoft.NETFramework.ReferencesAssemblies does not include net35, so we're using https://www.nuget.org/packages/jnm2.ReferenceAssemblies.net35 combined with a workaround to bypass looking for System.Data.Entity.dll 3.5.0.0 in the GAC.

### DIFF
--- a/eng/Packages.props
+++ b/eng/Packages.props
@@ -8,6 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Update="jnm2.ReferenceAssemblies.net35" Version="1.0.1" />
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
@jnm2 FYI, also thanks for saving the day.

jnm2:
@KirillOsenkov msbuildlog.com saved my day a hundred times, so I'm just glad to be able to make a dent in the opposite direction 😊

BenVillalobos:

PR #5625, opened 2020-08-07
Fix cpu core count detection

### LINKED ISSUE #435: MSBuild detects wrong core count on multi-proc machines
If you use `msbuild /m` on a machine with more than 32 cores, we will only launch up to 32 processes. This appears to be because we generally run as a 32-bit process, and 32-bit Windows had a 32-core limit. But we should be able to use as many processes as cores, so we need to call an API that gives us "the real truth" about `NUMBER_OF_PROCESSORS`.

mfkl:
Attempt to fix #435

- I'm not sure whether similar code should be needed for Unix support.
- I don't have a machine with > 32 cores to actually test this.

Potentially relevant:
- https://github.com/dotnet/corefx/pull/25630

### DIFF
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,12 +1725,16 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-                                if (!Environment.Is64BitProcess && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
+#if !RUNTIME_TYPE_NETCORE
+                                // .NET framework has a core count limit (32/64 depending on process bitness), 
+                                // so check with GetLogicalProcessorInformationEx in that case.
+                                if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
                                 {
                                     var result = NativeMethodsShared.GetPhysicalCoreCount();
                                     if(result != -1)
                                         numberOfCpus = result;
                                 }
+#endif
                                 switchParameters = ":" + numberOfCpus;
                             }
                             else if (String.Equals(switchName, "bl", StringComparison.OrdinalIgnoreCase) ||
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,7 +1725,7 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-                                if(numberOfCpus == 32 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
+                                if (numberOfCpus == 32 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
                                 {
                                     var result = NativeMethodsShared.GetPhysicalCoreCount();
                                     if(result != -1)
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1726,7 +1726,7 @@ namespace Microsoft.Build.CommandLine
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
 #if NETFRAMEWORK && !MONO
-                                // .NET framework has a core count limit (32/64 depending on process bitness), 
+                                // .NET framework calls Windows APIs that have a core count limit (32/64 depending on process bitness), 
                                 // so check with GetLogicalProcessorInformationEx in that case.
                                 if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
                                 {
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,9 +1725,11 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-#if NETFRAMEWORK && !MONO
-                                // .NET framework calls Windows APIs that have a core count limit (32/64 depending on process bitness), 
-                                // so check with GetLogicalProcessorInformationEx in that case.
+#if !MONO
+                                // .NET framework calls Windows APIs that have a core count limit (32/64 depending on process bitness).
+                                // .NET Core on Windows returns a core count limited to the current NUMA node
+                                //     https://github.com/dotnet/runtime/issues/41902
+                                // So if we get a high core count on Windows, double-check it.
                                 if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
                                 {
                                     var result = NativeMethodsShared.GetPhysicalCoreCount();
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,6 +1725,12 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
+                                if(numberOfCpus == 32 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
+                                {
+                                    var result = NativeMethodsShared.GetPhysicalCoreCount();
+                                    if(result != -1)
+                                        numberOfCpus = result;
+                                }
                                 switchParameters = ":" + numberOfCpus;
                             }
                             else if (String.Equals(switchName, "bl", StringComparison.OrdinalIgnoreCase) ||
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,7 +1725,7 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-                                if (numberOfCpus == 32 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
+                                if (IntPtr.Size == 4 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
                                 {
                                     var result = NativeMethodsShared.GetPhysicalCoreCount();
                                     if(result != -1)
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,7 +1725,7 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-#if !RUNTIME_TYPE_NETCORE
+#if NETFRAMEWORK && !MONO
                                 // .NET framework has a core count limit (32/64 depending on process bitness), 
                                 // so check with GetLogicalProcessorInformationEx in that case.
                                 if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1725,7 +1725,7 @@ namespace Microsoft.Build.CommandLine
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-                                if (IntPtr.Size == 4 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
+                                if (!Environment.Is64BitProcess && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
                                 {
                                     var result = NativeMethodsShared.GetPhysicalCoreCount();
                                     if(result != -1)
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1726,11 +1726,16 @@ namespace Microsoft.Build.CommandLine
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
 #if !MONO
-                                // .NET framework calls Windows APIs that have a core count limit (32/64 depending on process bitness).
                                 // .NET Core on Windows returns a core count limited to the current NUMA node
-                                //     https://github.com/dotnet/runtime/issues/41902
-                                // So if we get a high core count on Windows, double-check it.
-                                if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
+                                //     https://github.com/dotnet/runtime/issues/29686
+                                // so always double-check it.
+                                if (NativeMethodsShared.IsWindows
+#if NETFRAMEWORK
+                                     // .NET Framework calls Windows APIs that have a core count limit (32/64 depending on process bitness).
+                                     // So if we get a high core count on full framework, double-check it.
+                                     && (numberOfCpus >= 32)
+#endif
+                                    )
                                 {
                                     var result = NativeMethodsShared.GetLogicalCoreCount();
                                     if(result != -1)
--- a/src/MSBuild/XMake.cs
+++ b/src/MSBuild/XMake.cs
@@ -1732,7 +1732,7 @@ namespace Microsoft.Build.CommandLine
                                 // So if we get a high core count on Windows, double-check it.
                                 if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
                                 {
-                                    var result = NativeMethodsShared.GetPhysicalCoreCount();
+                                    var result = NativeMethodsShared.GetLogicalCoreCount();
                                     if(result != -1)
                                         numberOfCpus = result;
                                 }
--- a/src/Shared/NativeMethodsShared.cs
+++ b/src/Shared/NativeMethodsShared.cs
@@ -150,21 +150,16 @@ namespace Microsoft.Build.Shared
         {
             public LOGICAL_PROCESSOR_RELATIONSHIP Relationship;
             public uint Size;
-            public GROUP_RELATIONSHIP Group;
+            public PROCESSOR_RELATIONSHIP Processor;
         }
-        internal unsafe struct GROUP_RELATIONSHIP
+        [StructLayout(LayoutKind.Sequential)]
+        internal unsafe struct PROCESSOR_RELATIONSHIP
         {
-            private ushort MaximumGroupCount;
-            public ushort ActiveGroupCount;
+            public byte Flags;
+            private byte EfficiencyClass;
             private fixed byte Reserved[20];
-            public PROCESSOR_GROUP_INFO GroupInfo;
-        }
-        internal unsafe struct PROCESSOR_GROUP_INFO
-        {
-            public byte MaximumProcessorCount;
-            public byte ActiveProcessorCount;
-            public fixed byte Reserved[38];
-            public IntPtr ActiveProcessorMask;
+            public ushort GroupCount;
+            public IntPtr GroupInfo;
         }
 #pragma warning restore 0169, 0149
 
@@ -512,7 +507,7 @@ namespace Microsoft.Build.Shared
             uint len = 0;
             const int ERROR_INSUFFICIENT_BUFFER = 122;
 
-            if (!GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationGroup, IntPtr.Zero, ref len) &&
+            if (!GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore, IntPtr.Zero, ref len) &&
                 Marshal.GetLastWin32Error() == ERROR_INSUFFICIENT_BUFFER)
             {
                 // Allocate that much space
@@ -520,7 +515,7 @@ namespace Microsoft.Build.Shared
                 fixed (byte* bufferPtr = buffer)
                 {
                     // Call GetLogicalProcessorInformationEx with the allocated buffer
-                    if (GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationGroup, (IntPtr)bufferPtr, ref len))
+                    if (GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore, (IntPtr)bufferPtr, ref len))
                     {
                         // Walk each SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX in the buffer, where the Size of each dictates how
                         // much space it's consuming.  For each group relation, count the number of active processors in each of its group infos.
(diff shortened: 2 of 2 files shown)


src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +480,42 @@ public SystemInformationData()
...
+            const int ERROR_INSUFFICIENT_BUFFER = 122;
+
+			// Determine the required buffer size to store the processor information
+			uint ReturnLength = 0;
+			if(!GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore, IntPtr.Zero, ref ReturnLength) && Marshal.GetLastWin32Error() == ERROR_INSUFFICIENT_BUFFER)
+			{
+                IntPtr ptr = IntPtr.Zero;
+				try
+				{
+                    ptr = Marshal.AllocHGlobal((int)ReturnLength);
```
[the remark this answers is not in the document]
mfkl: I usually put `Marshal.AllocHGlobal` inside try/catch statements to catch potential OOM exceptions. Calling free with `IntPtr.Zero` is safe. Can change it, if you prefer.

src/MSBuild/XMake.cs
```diff
@@ -1725,6 +1725,12 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
+                                if(numberOfCpus == 32 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
```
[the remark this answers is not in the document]
mfkl: done

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +480,42 @@ public SystemInformationData()
...
+					if (GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore, ptr, ref ReturnLength))
+					{
+						int count = 0;
+						for(int pos = 0; pos < ReturnLength; )
+						{
+							LOGICAL_PROCESSOR_RELATIONSHIP Type = (LOGICAL_PROCESSOR_RELATIONSHIP)Marshal.ReadInt16(ptr, pos);
+							if(Type == LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore)
+							{
+								count++;
```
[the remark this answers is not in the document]
mfkl: Ok, reused previous corefx code, count should be more accurate now.

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +480,42 @@ public SystemInformationData()
...
+                IntPtr ptr = IntPtr.Zero;
+				try
+				{
+                    ptr = Marshal.AllocHGlobal((int)ReturnLength);
+					if (GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore, ptr, ref ReturnLength))
+					{
+						int count = 0;
+						for(int pos = 0; pos < ReturnLength; )
+						{
+							LOGICAL_PROCESSOR_RELATIONSHIP Type = (LOGICAL_PROCESSOR_RELATIONSHIP)Marshal.ReadInt16(ptr, pos);
```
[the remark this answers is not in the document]
mfkl: Yes

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +480,42 @@ public SystemInformationData()
...
+							if(Type == LOGICAL_PROCESSOR_RELATIONSHIP.RelationProcessorCore)
+							{
+								count++;
+							}
+							pos += Marshal.ReadInt32(ptr, pos + 4);
```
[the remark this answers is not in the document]
mfkl: See new version.

mfkl:
> I imagine we do care about linux, but I don't know whether the same API works.

Don't think so as `GetLogicalProcessorInformationEx` is a win32 API. There are several options for Linux/macOS, unsure which one you'd prefer.

Also, shouldn't this cpu checking code be located in https://github.com/dotnet/runtime/blob/96f178d32b7ba62485917ac46ef1edcfd3c2d10d/src/coreclr/src/classlibnative/bcltype/system.cpp#L327 instead?

src/Shared/NativeMethodsShared.cs
```diff
@@ -137,6 +137,39 @@ private enum eDesiredAccess : int
...
+        internal struct SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX
+        {
+            public LOGICAL_PROCESSOR_RELATIONSHIP Relationship;
+            public uint Size;
+            public GROUP_RELATIONSHIP Group;
+        }
+
+        internal unsafe struct GROUP_RELATIONSHIP
+        {
+            private byte MaximumGroupCount;
```
ladipro, line 159: nit: This is defined as
```C++
WORD                 MaximumGroupCount;
```
in the unmanaged world so should be an `ushort` here.

src/Shared/NativeMethodsShared.cs
```diff
@@ -137,6 +137,39 @@ private enum eDesiredAccess : int
             PROCESS_ALL_ACCESS = SYNCHRONIZE | 0xFFF
         }
 
+        internal enum LOGICAL_PROCESSOR_RELATIONSHIP
+		{
```
ladipro, line 141: nit: Bad indentation.

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +503,47 @@ public SystemInformationData()
             }
         }
 
+        public unsafe static int GetPhysicalCoreCount()
```
ladipro, line 506: Can you please add a comment with a short `<summary>`?

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +503,47 @@ public SystemInformationData()
...
+                        while (ptr < endPtr)
+                        {
+                            var current = (SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX*)ptr;
+                            if (current->Relationship == LOGICAL_PROCESSOR_RELATIONSHIP.RelationGroup)
+                            {
+                                PROCESSOR_GROUP_INFO* groupInfo = &current->Group.GroupInfo;
+                                int groupCount = current->Group.ActiveGroupCount;
+                                for (int i = 0; i < groupCount; i++)
+                                {
+                                    processorCount += (groupInfo + i)->ActiveProcessorCount;
```
ladipro, line 534: Add a check that this dereference is not overrunning the containing `buffer`, just in case.

src/MSBuild/XMake.cs
```diff
@@ -1725,6 +1725,12 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
+                                if (numberOfCpus == 32 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
```
ladipro, line 1728: I think it would be better if either:
- This code ran always, i.e. remove the `numberOfCpus == 32` condition. It would help test the new code on more machines. Unless calling the Win32 API has a measurable perf impact.
- Or this code ran only if the process is actually 32-bit. Users may choose to run 64-bit msbuild.exe and my understanding is that this fix is not needed there.

BenVillalobos:

src/Shared/NativeMethodsShared.cs
elachlan, line 159: There was a build error related to the MaximumGroupCount:
`error CS0169: The field 'NativeMethodsShared.GROUP_RELATIONSHIP.MaximumGroupCount' is never used`

src/Shared/NativeMethodsShared.cs
mfkl: I'm not quite sure about the kind of check you have in mind here?
```
byte* ptr = bufferPtr, endPtr = bufferPtr + len;
while (ptr < endPtr)
```
This seems correct from what I can tell.
What check would you add to `processorCount += (groupInfo + i)->ActiveProcessorCount;`?

src/MSBuild/XMake.cs
```diff
@@ -1725,6 +1725,12 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
+                                if (IntPtr.Size == 4 && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
```
Forgind, line 1728: Is there a more intuitive way to get that it's 32-bit? If not, this is fine. I was only confused until I read the comment.

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +501,52 @@ public SystemInformationData()
...
+                var buffer = new byte[len];
+                fixed (byte* bufferPtr = buffer)
+                {
+                    // Call GetLogicalProcessorInformationEx with the allocated buffer
+                    if (GetLogicalProcessorInformationEx(LOGICAL_PROCESSOR_RELATIONSHIP.RelationGroup, (IntPtr)bufferPtr, ref len))
+                    {
+                        // Walk each SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX in the buffer, where the Size of each dictates how
+                        // much space it's consuming.  For each group relation, count the number of active processors in each of its group infos.
+                        int processorCount = 0;
+                        byte* ptr = bufferPtr, endPtr = bufferPtr + len;
```
Forgind, line 527: nit:
split onto two lines

I also kinda liked your for loop better, but this is fine.

src/MSBuild/XMake.cs
BenVillalobos: If there isn't another way, can we comment with a link to https://docs.microsoft.com/en-us/dotnet/api/system.intptr.size?view=netcore-3.1, or the excerpt from that page
`The size of a pointer or handle in this process, measured in bytes. The value of this property is 4 in a 32-bit process, and 8 in a 64-bit process. You can define the process type by setting the /platform switch when you compile your code with the C# and Visual Basic compilers.`

src/Shared/NativeMethodsShared.cs
ladipro: If `groupCount` is a bogus number, `(groupInfo + i)` could be pointing outside of the `bufferPtr` <-> `endPtr` block. I believe it's a good security practice to not trust the data and check that all memory accesses are within the bounds when using unsafe C# code.

Strictly speaking, this should also include

src/MSBuild/XMake.cs
Forgind: To be clear, I understood immediately after reading the comment. Although I'd count this as somewhat subtle but definitely simple and functional.

elachlan: Maybe extract it to a readonly static property is32bit? Or use an enum?

src/Shared/NativeMethodsShared.cs
```diff
@@ -136,6 +136,37 @@ private enum eDesiredAccess : int
...
+        internal enum LOGICAL_PROCESSOR_RELATIONSHIP
+        {
+            RelationProcessorCore,
+            RelationNumaNode,
+            RelationCache,
+            RelationProcessorPackage,
+            RelationGroup,
+            RelationAll = 0xffff
+        }
+        internal struct SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX
```
BenVillalobos, line 149: Question for myself. I noticed this struct has more members: https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-system_logical_processor_information_ex.

Can you simply not define all members and it will populate only what you have defined without raising some sort of error?

src/MSBuild/XMake.cs
Forgind: Good point! You reminded me of https://github.com/dotnet/msbuild/blob/master/src/Shared/EnvironmentUtilities.cs#L11

We should follow this model.

elachlan: Just use this but test for the negation?

elachlan: Also it looks like there is a static bool in environment.
https://github.com/dotnet/runtime/blob/ea2b09beea535abf960651c59359e40f99f2fecf/src/libraries/System.Private.CoreLib/src/System/Environment.cs#L138

Forgind: Either way works, but I'd make a new `Is32Bit` version. We probably don't have to worry about 16-bit machines, and we don't currently have to worry about 128-bit machines, but just in case this code lasts a while...

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +501,52 @@ public SystemInformationData()
             }
         }
 
+        /// <summary>
+        /// Get the exact physical core count on Windows
+        /// Useful for getting the exact core count in 32 bits processes,
+        /// as Environment.ProcessorCount has a 32-core limit in that case. 
+        /// </summary>
+        public unsafe static int GetPhysicalCoreCount()
```
cdmihai, line 509: Ideally this should be copy pasted code from a trustworthy repo (dotnet runtime, corefx, etc) and we just copy paste it and add a comment citing the source. If that's the case, can you please cite the source? (github file line permalink)

Well, ideally ideally it should be in a shared library of some sorts, but let's not get too ideal :)

src/MSBuild/XMake.cs
mfkl: `EnvironmentUtilities` is internal to this assembly thus cannot be referenced without adding an InternalsVisibleTo which I don't think we want to do. Used `!Environment.Is64BitProcess` instead.

src/Shared/NativeMethodsShared.cs
mfkl: >  split onto two lines

done

src/Shared/NativeMethodsShared.cs
mfkl: > Can you simply not define all members and it will populate only what you have defined without raising some sort of error?

Fair question. I think it depends on what you need to marshal. In case of union, this way of doing things seems to be fine and the original writer had a comment about it: https://github.com/dotnet/runtime/blob/3ab9c0b79300a5b3d954421aa4da15a9c2ca1271/src/libraries/Common/src/Interop/Windows/kernel32/Interop.GetLogicalProcessorInformationEx.cs#L27

src/Shared/NativeMethodsShared.cs
mfkl: Done https://github.com/dotnet/msbuild/pull/5625/commits/70d41ae6e695ad78acc73543dde8093285ae93b2

src/Shared/NativeMethodsShared.cs
mfkl: Since this code was directly taken from dotnet/runtime unmodified, and has been pushed to production, I'm inclined to believe it works and does not need additional checks. I'll add the checks since you prefer though.

mfkl: Added some checks https://github.com/dotnet/msbuild/pull/5625/commits/4223cc2e0c875c527afaf5f67a255c2995c4ebb1, please feel free to let me know if that's not what you had in mind.

src/Shared/NativeMethodsShared.cs
```diff
@@ -470,6 +501,61 @@ public SystemInformationData()
...
+                        while (ptr < endPtr)
+                        {
+                            var current = (SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX*)ptr;
+                            if (current != null && current->Relationship == LOGICAL_PROCESSOR_RELATIONSHIP.RelationGroup)
+                            {
+                                PROCESSOR_GROUP_INFO* groupInfo = &current->Group.GroupInfo;
+                                if (groupInfo != null)
+                                {
+                                    int groupCount = current->Group.ActiveGroupCount;
+                                    for (int i = 0; i < groupCount; i++)
```
Forgind, line 539: I may have misunderstood, but I thought @ladipro was thinking something more like this?

```suggestion
                                    for (int i = 0; i < groupCount && groupInfo + i < endPtr; i++)
```

src/MSBuild/XMake.cs
```diff
@@ -1725,6 +1725,12 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
+                                if (!Environment.Is64BitProcess && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
```
rainersigwald, line 1728: 64-bit Windows had a 64-core limit, too. I think this should be called when we're running on .NET Framework regardless of bitness (on .NET Core/.NET 5, we can rely on `Environment.ProcessorCount` to do the trick for us, right?).

src/Shared/NativeMethodsShared.cs
ladipro: Thanks, yes, I was thinking something like this, not the `null` checks added in 4223cc2e0c875c527afaf5f67a255c2995c4ebb1. It would have to be:

```C#
for (int i = 0; i < groupCount && groupInfo + i + 1 <= endPtr; i++)
```

i.e. the entire structure pointed to by `(groupInfo + i)` fits in the buffer.

I mentioned this to @danmosemsft offline and he doesn't consider this kind of hardening necessary when calling a trusted API. @mfkl, feel free to revert the 4223cc2e0c875c527afaf5f67a255c2995c4ebb1. Apologies for randomizing you.

danmosemsft: Yeah, clearly we don't do it in the code this came from in dotnet/runtime and I think it would be simply "testing the OS has no bugs". But if there's a security context, don't take my word for it - we should ask others.

src/MSBuild/XMake.cs
mfkl: Ok

src/Shared/NativeMethodsShared.cs
mfkl: All good, reverted.

src/MSBuild/XMake.cs
```diff
@@ -1725,12 +1725,16 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
-                                if (!Environment.Is64BitProcess && NativeMethodsShared.IsWindows) // 32-bit process, 32-bit Windows had a 32-core limit
+#if !RUNTIME_TYPE_NETCORE
```
rainersigwald, line 1728:
```suggestion
#if NETFRAMEWORK && !MONO
```

I think this is slightly clearer and preserves the intent: do this only on full .NET Framework.

src/MSBuild/XMake.cs
```diff
@@ -1725,12 +1725,16 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
...
+                                // .NET framework has a core count limit (32/64 depending on process bitness),
```
rainersigwald, line 1729:
```suggestion
                                // .NET framework calls Windows APIs that have a core count limit (32/64 depending on process bitness), 
```

src/MSBuild/XMake.cs
elachlan: > 64-bit Windows had a 64-core limit, too. I think this should be called when we're running on .NET Framework regardless of bitness (on .NET Core/.NET 5, we can rely on `Environment.ProcessorCount` to do the trick for us, right?).

.NET 5 `Environment.ProcessorCount` calls the native GetProcessorCount function. So I think its all good. 
https://github.com/dotnet/runtime/blob/ea2b09beea535abf960651c59359e40f99f2fecf/src/libraries/System.Private.CoreLib/src/System/Environment.cs#L13

https://github.com/dotnet/runtime/blob/60eff3f3766631bd4e7ee256ca17619fec90e9e6/src/coreclr/src/System.Private.CoreLib/src/System/Environment.CoreCLR.cs#L83-L84

https://github.com/dotnet/runtime/blob/96f178d32b7ba62485917ac46ef1edcfd3c2d10d/src/coreclr/src/classlibnative/bcltype/system.cpp#L327-L364

stackedsax:
> I think this looks good to me. I'm trying to pull some strings internally to borrow a machine big enough to test it on (looks like Azure will charge me merely $50,000/month for a big enough VM!).

@rainersigwald I reached out to some people within the dotnet foundation to see if there might be funds for something like this.  I'll let you know if I hear anything.

elachlan:
@rainersigwald I emailed AMD and they have offered to test on their 64-core ThreadRipper 3990X processor.

rainersigwald:
Current status:

I got access to a [Standard_F72s_v2](https://docs.microsoft.com/en-us/azure/virtual-machines/fsv2-series?toc=/azure/virtual-machines/linux/toc.json&bc=/azure/virtual-machines/linux/breadcrumb/toc.json) VM, installed a private Visual Studio with this fix and tested it out.

It almost works! In 64-bit everything looks great. In 32-bit, the new code correctly identifies 2 processor groups (2 NUMA nodes), but Windows lies to us _again_ and claims only 32 cores in each, when really there are 36--so we're artificially limited to 64 MSBuild nodes.

This is in [the docs](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformationex#remarks):

> In addition, the sum of all per-group ActiveProcessorCount and MaximumProcessorCount values reported in PROCESSOR_GROUP_INFO structures may exclude some active logical processors.

I'll start a thread up with some Windows folks internally to see if this is the best we can do.

rainersigwald:
Latest push seems to work, at least until [AMD blows things up](https://www.tweaktown.com/news/67801/amds-next-gen-zen-3-rumor-4-threads-per-core-enables-64c-256t-chips/index.html).

rainersigwald:
However, depending on how https://github.com/dotnet/runtime/issues/41902 is resolved, we may need to do this on .NET Core/Windows too.

rainersigwald (approved):
Tested my latest commit on the 72-core machine in `dotnet build` and `msbuild.exe` scenarios and both successfully started and used 72 nodes. Looks good to me!

src/MSBuild/XMake.cs
```diff
@@ -1725,6 +1725,18 @@ internal static void GatherCommandLineSwitches(ArrayList commandLineArgs, Comman
                                 String.Equals(switchName, "maxcpucount", StringComparison.OrdinalIgnoreCase))
                             {
                                 int numberOfCpus = Environment.ProcessorCount;
+#if !MONO
+                                // .NET framework calls Windows APIs that have a core count limit (32/64 depending on process bitness).
+                                // .NET Core on Windows returns a core count limited to the current NUMA node
+                                //     https://github.com/dotnet/runtime/issues/41902
+                                // So if we get a high core count on Windows, double-check it.
+                                if (numberOfCpus >= 32 && NativeMethodsShared.IsWindows)
```
Forgind, line 1733: Were we not thinking of expanding this for cases like, say, if you have 2 * 24 CPUs?

Forgind (approved):

src/MSBuild/XMake.cs
rainersigwald: Ah, indeed we were. Back to the drawing board.

rainersigwald: Ok, latest push looks good again.

rainersigwald:
Thank you @mfkl! This should be available starting in 16.8 preview 4.

mfkl:
Thank you for picking it up and finishing this :-)

rainersigwald:
By the way, GitHub somehow mangled the squash/merge commit--I did not intentionally erase you from the commit history. Sorry about that!

mfkl:
No worries.

src/Shared/NativeMethodsShared.cs
GSPP: These checks will now just ignore invalid data which is, to our understanding, logically impossible. Should they not be asserting or throwing?

dmex:
@rainersigwald  @mfkl 

This PR for the MSBuild cpu core count detection has an issue with debugging processor groups. I've created a new issue here and included some extra information for debugging without physical hardware: https://github.com/dotnet/msbuild/issues/7943

Issue #5183, opened 2020-03-19
Remove MSBuild assemblies from the GAC at VSIX install time.


rainersigwald:
Customers repeatedly hit problems caused by having stale MSBuild assemblies in their Global Assembly Cache.

For example,

- https://developercommunity.visualstudio.com/content/problem/208631/unable-to-create-or-open-solution.html
- https://developercommunity.visualstudio.com/content/problem/953754/after-update-to-1650-no-longer-able-to-loadcreate.html
- many more

We have long considered finding a way to ensure that MSBuild is not in the GAC when the package is updated, perhaps by running

```
gacutil /u "Microsoft.Build.Utilities.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL"
gacutil /u "Microsoft.Build.Task.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL"
gacutil /u "Microsoft.Build.Conversion.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL"
gacutil /u "Microsoft.Build.Engine, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL"
gacutil /u "Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL"
```

in a related VS installer package.

PR #5626, opened 2020-08-07
UnGAC MSBuild Assemblies on Install or Update

BenVillalobos:
Fixes https://github.com/dotnet/msbuild/issues/5183


src/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+                // Throw away the first line - it's the command we just used
+                output.ReadLine();
+                string s = output.ReadLine();
+
+                // Store all instances of gacutil for now.
+                while (!String.IsNullOrEmpty(s))
+                {
+                    allInstancesOfGACUtil.Add(s);
+                    s = output.ReadLine();
+                }
```
Forgind, line 59:
```suggestion
                // Store all instances of gacutil for now.
                for (string loc = output.ReadLine(); !String.IsNullOrEmpty(loc); loc = output.ReadLine())
                {
                    allInstancesOfGACUtil.Add(s);
                }
```

Does this terminate appropriately? If so, I think there's a ReadAllLines function that might do the same. (I might have made that up 😄.)

Also, do we really care about any gacutils after the first?

src/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+                    proc.Dispose();
+                    Console.WriteLine("Could not use where.exe to find gacutil.exe. Exiting...");
+                    return;
+                }
+
+                // Use where.exe to find gacutil.exe
+                input.WriteLine("where /F /R \"C:\\Program Files (x86)\\Microsoft SDKs\\Windows\" gacutil.exe");
+
+                // Throw away the first line - it's the command we just used
+                output.ReadLine();
```
Forgind, line 51: This would probably make it messier rather than cleaner, but you should be able to set echo to off to suppress this.

src/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        static void Main(string[] args)
+        {
+            try
+            {
+                List<string> allInstancesOfGACUtil = new List<string>();
+                StreamReader output;
+                StreamWriter input;
+
+                Process proc = new Process
```
Forgind, line 22: This seems heavy. Is there a way to run it in-proc?

src/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+                // Output everything gacutil returned.
+                while (s != null)
+                {
+                    Console.WriteLine(s);
+                    s = output.ReadLine();
+                }
+
+                proc.Close();
+                proc.Dispose();
+                Console.ReadKey();
```
Forgind, line 103: What is this?

src/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.ExceptionServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Build.UnGAC
+{
+    class Program
```
Forgind, line 12: Does this work, and is this the right place for this? I'm surprised first that this runs at install time rather than running every time the user builds and second that you had to make a new project to make something that runs automatically. Could you explain that a little more?

src/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: > Does this terminate appropriately? If so, I think there's a ReadAllLines function that might do the same. (I might have made that up 😄.)
It does and has always, which kind of worries me. I would do ReadAllLines but have experienced lockups from that as well. Perhaps I can clear the buffer (as I do below) so I can safely pull everything.

> Also, do we really care about any gacutils after the first?
Honestly I don't think so. I did it because "why not" while I had them there. The final version will almost certainly not store all instances.

src/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: I'm not aware of one. I considered using an `exec` task but we need this to be an exe. If someone knows I'd love to know!

src/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: The readkey is temporary so I can see the console output before the window closes.

Forgind: 😕 --> 😄 --> 👍

src/Microsoft.Build.UnGAC/Program.cs
Forgind: @ladipro?
He knows everything.

BenVillalobos:
Another issue I've been seeing since adding the `Microsoft.Build.UnGAC` project:
image
I've tried removing the nuget packages, but can't seem to from this project.

Looks like they're taken from `src/Directory.Build.targets`
```
  <ItemGroup Condition="'$(IsUnitTestProject)' == 'true' And '$(TargetFrameworkIdentifier)' != '.NETFramework' ">
    <PackageReference Include="xunit.console" />
  </ItemGroup>
```

Presumably I shouldn't have this project under `src`, or somehow override that import for this particular project.

src/Microsoft.Build.UnGAC/Program.cs
ladipro: You can access the GAC via the Fusion interface, for example:
https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/fusion/iassemblycache-uninstallassembly-method

Not sure if a COM interop wrapper exists or you'll have to build one yourself. I guess a big reason to go this route would be if gacutil.exe is not guaranteed to be present on the machine. Otherwise, if you know for sure there is gacutil.exe and you can find it, it may be easier to just execute it. I'm assuming we don't care about performance here.

src/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: What will allow it to run at install time is the `exe` being pointed to from `src/Package/MSBuild.VSSetup/exe.swr`, which `src/Package/MSBuild.VSSetup/files.swr` will point at. And `files.swr` is already being picked up by microbuild.

src/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: Good idea. There are a few that exist, but either the links to them are dead or they're not supported anymore. I found https://github.com/EasyHook/EasyHook
http://blog.wibeck.org/2008/10/working-with-gac-from-code/ points to [an API wrapper](https://docs.microsoft.com/en-us/archive/blogs/junfeng/articles/229649) that unfortunately 404's today.
There's also https://www.codeproject.com/Articles/8285/GAC-API-Interface

rainersigwald: This is likely of interest: https://github.com/dotnet/msbuild/blob/f2c4bfd563f559daca27ea1cd8ae40db24e2e7cd/src/Tasks/NativeMethods.cs#L344-L351

I don't think we can depend on `gacutil.exe` existing: the VS install may be what installs it. We could maybe express a dependency on "some .NET SDK" and use that, but it's probably easier to use the COM object.

> it's probably easier to use the COM object

😲

BenVillalobos: I just spent a good chunk of time trying to get https://docs.microsoft.com/en-us/dotnet/api/system.enterpriseservices.internal.publish.gacremove?view=netframework-4.8 working to no avail (passing in file path, assembly name, pointing directly to dll, etc). MattCav suggested it was preferable to use some sort of API and keep the code simple so that's the route I'll go.

BenVillalobos: Reusing `NativeMethods` was the way to go. The program does need to be running as admin for it to work, however.

MSBuild.Dev.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2009
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30320.27
```
rainersigwald, line 4: Nit: Go ahead and revert this entirely since there's no meaningful change.

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
```
rainersigwald, line 1: Copyright header please.

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <!-- The project MUST be net45. -->
```
rainersigwald, line 4:
```suggestion
    <!-- Setup runs on net45 and may not have installed newer yet. -->
```

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
```
rainersigwald, line 1:
```suggestion
```

Apache is wrong in the first place, and we don't usually banner our proj files.

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
```diff
@@ -0,0 +1,29 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <!-- The project MUST be net45. -->
+    <TargetFramework>net45</TargetFramework>
+    <!-- Set as an exe because this project publishes its own output. -->
+    <OutputType>Exe</OutputType>
+    <!-- Forcing the 'vsix' output to be json output. -->
```
rainersigwald, line 8: Add

> ... workaround for dotnet/arcade#xxxx

(and file the "support exe packages" bug there ;))

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
```diff
@@ -0,0 +1,29 @@
...
+
+  <ItemGroup>
+    <_SwixArgs Include="PackageType=manifest"/>
+    <!-- If we don't set TargetExt to .json, arcade defaults to generating a .vsix -->
+    <_SwixArgs Include="TargetExt=.json"/>
+  </ItemGroup>
+
+  <!-- SwrProperty is an arcade concept here. -->
+  <ItemGroup>
+    <!-- BinDir is set via SwrProperty so arcade can see it. OutDir is the path to the exe. See exe.swr to see how this is used. -->
```
rainersigwald, line 22:
```suggestion
    <!-- BinDir is set via SwrProperty to be passed to the swr compiler. OutDir is the path to the bin folder. See exe.swr to see how this is used. -->
```

Personally I'd cut out the last two sentences but no real harm in leaving them in.

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <SwrProperty Include="BinDir=$(OutDir)" />
+    <!-- Version must constantly update in order for our exe package to run on install and update. -->
+    <SwrProperty Include="Version=$(VsixVersion)" />
+    <!-- our swr file must be picked up, this is how we set that. -->
+    <SwrFile Include="exe.swr" Condition=" '$(MSBuildRuntimeType)' == 'Full' " />
```
rainersigwald, line 27:
```suggestion

    <!-- our swr file must be picked up, this is how we set that. -->
    <SwrFile Include="exe.swr" Condition=" '$(MSBuildRuntimeType)' == 'Full' " />
```

src/Package/MSBuild.UnGAC.Setup/NativeMethods.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Build.UnGAC
+{
+    [ComImport, Guid("E707DCDE-D1CD-11D2-BAB9-00C04F8ECEAE"), InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    internal interface IAssemblyCache
+    {
+        [PreserveSig]
+        uint UninstallAssembly(uint dwFlags, [MarshalAs(UnmanagedType.LPWStr)] string pszAssemblyName, IntPtr refData, int pulDisposition);
```
rainersigwald, line 13:
```suggestion
        uint UninstallAssembly(uint dwFlags, [MarshalAs(UnmanagedType.LPWStr)] string pszAssemblyName, IntPtr refData, ref int pulDisposition);
```

? it's defined as `[out, optional] ULONG *pulDisposition`. Though since we're still doing best-effort we don't need to check it.

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.Build.UnGAC
+{
+    class Program
+    {
+        static void Main(string[] args)
```
rainersigwald, line 7: Needs comment describing functionality/motivation.

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                    "Microsoft.Build.Conversion.Core, Version=15.1.0.0"
+                };
+
+                uint hresult = NativeMethods.CreateAssemblyCache(out IAssemblyCache assemblyCache, 0);
+
+                // Most significant bit is set, meaning there was an error in the Hresult.
+                if ((hresult >> 31) == 1)
+                {
+                    Console.WriteLine($"Could not successfully call CreateAssemblyCache. HResult: {hresult}");
+                    Console.WriteLine("Exiting without removing assemblies from the GAC...");
```
rainersigwald, line 27: Can you confirm with the Setup folks that this is the best way to log? Or is there something where we could collect telemetry or something?

(I'd also love to have telemetry for just "how many people had something in the GAC?")

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                    return;
+                }
+
+                for (int i = 0; i < assembliesToUnGAC.Length; i++)
```
rainersigwald, line 31: why not a `foreach`?

src/Package/MSBuild.UnGAC.Setup/NativeMethods.cs
rainersigwald: extreme nit: could you be more consistent about dropping or not dropping the Hungarian prefixes?

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    hresult = assemblyCache.UninstallAssembly(0, assembliesToUnGAC[i], IntPtr.Zero, 0);
```
rainersigwald, line 33: include arg names to make the magic constants clearer?
```suggestion
                    hresult = assemblyCache.UninstallAssembly(dwFlags: 0, pszAssemblyName: assembliesToUnGAC[i], refData: IntPtr.Zero, pulDisposition: 0);
```

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+
+                    // If we hit an error with an assembly, keep trying the others.
+                    if ((hresult >> 31) == 1)
+                    {
+                        Console.WriteLine($"Could not remove {assembliesToUnGAC[i]} from the GAC. HResult: {hresult}");
+                    }
+                }
+
+                Console.ReadKey();
```
rainersigwald, line 42: 👀

src/Package/MSBuild.UnGAC.Setup/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine($"Caught an exception! We don't want to throw because we want MSBuild to install." +
+                                    $"Message: {e.Message}" +
+                                    $"Inner Exception: {e.InnerException}" +
+                                    $"Stack Trace: {e.StackTrace}");
```
rainersigwald, line 49: `Exception.ToString()` should give you all of this for free. Printing `Exception.Message` is a code smell.

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
```diff
@@ -0,0 +1,29 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <!-- The project MUST be net45. -->
+    <TargetFramework>net45</TargetFramework>
+    <!-- Set as an exe because this project publishes its own output. -->
+    <OutputType>Exe</OutputType>
+    <!-- Forcing the 'vsix' output to be json output. -->
+    <TargetVsixContainerName>MSBuild.UnGAC.Setup.json</TargetVsixContainerName>
+    <!-- VS Insertion -->
+    <VisualStudioInsertionComponent>Microsoft.Build.UnGAC.Setup</VisualStudioInsertionComponent>
```
rainersigwald, line 11: Can you make the exe name and namespace and folder and package name consistent? I'd say `Microsoft.Build.UnGAC`, but wouldn't argue if you want the `.Setup`.

src/Package/MSBuild.UnGAC.Setup/Program.cs
BenVillalobos: Oops! 😅

src/Package/MSBuild.UnGAC.Setup/Program.cs
BenVillalobos: Force of habit for the _slight_ performance difference.

ladipro: C# is smart enough to emit a for loop even if you write a foreach:

https://sharplab.io/#v2:EYLgdgpgLgZgHgHwAICYCMBYAUASD37JAZgAJUy0B2bAb2x1RXrqxLfbaTQAYBtAXRIBDAM4iIAW2AAbAJYQRAFQD2AVTABxAIIBhEgF5sHDi2NnzbAEQBZWQGMATspHKYUAHQAhAK6zpAEwAaEgA1CAcRWWUwfTQAVnc0d25ky0CjC0yrW0dnVw8fP393AFEwAHNZSGCwiKiY+MTk1PTWLMybeycXNy9fAPcAMQchCQgAd2UHAGsa8Mjo2ISklO40jPbzTtyegv7ipAA2EXcdKYg5usXGlZaNzeNt7vy+ovdVKD9ZT4VT88uFg1ls01q0Hlscs9eoUBmcwAA3eb1P4OC6hJHXYGrSz3DgAXwA3PQcMSYOchHYABYkAAU8KEDmEYkkMgAniQqkzxFI5AoVOptDoAJTElj4LgAThpom5bKFRNwODx9GVWDxQA

BenVillalobos: 😲 I'm convinced!

src/Package/MSBuild.UnGAC.Setup/MSBuild.UnGAC.Setup.csproj
BenVillalobos: Filed: https://github.com/dotnet/arcade/issues/6120

src/Package/Microsoft.Build.UnGAC/Microsoft.Build.UnGAC.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <!-- Setup runs on net45 and may not have installed newer yet. -->
+    <TargetFramework>net45</TargetFramework>
```
Forgind, line 4: Does that mean it doesn't work on mac/linux? Is there a way we can make it work on mac/linux?

src/Package/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Microsoft.Build.UnGAC
+{
+    /// <summary>
+    /// Original Issue: https://github.com/dotnet/msbuild/issues/5183
+    /// This tool was created to help prevent customers from putting MSBuild assemblies in the Global Assembly Cache.
+    /// It runs at VS install-time as well as repair-time.
+    /// It is intended to run as best effort. Meaning that if it fails, we avoid throwing and instead log it.
+    /// </summary>
+    class Program
```
Forgind, line 14: nit: generic name

src/Package/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        static void Main(string[] args)
+        {
+            try
+            {
+                string[] assembliesToUnGAC =
+                {
+                    "Microsoft.Build, Version=15.1.0.0",
```
Forgind, line 22: Are these the only versions people GAC?

src/Package/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+
+                foreach (string assembly in assembliesToUnGAC)
+                {
+                    hresult = assemblyCache.UninstallAssembly(dwFlags: 0, pszAssemblyName: assembly, pRefData: IntPtr.Zero, pulDisposition: 0);
+
+                    // If we hit an error with an assembly, keep trying the others.
+                    if ((hresult >> 31) == 1)
+                    {
+                        Console.WriteLine($"Could not remove {assembly} from the GAC. HResult: {hresult}");
```
Forgind, line 47: Does it write this if the assembly just wasn't in the GAC?

Forgind (approved):
Looking good!

src/Package/Microsoft.Build.UnGAC/Microsoft.Build.UnGAC.csproj
rainersigwald: The GAC is a thing only on .NET Framework (on Windows). Things do not break because of the GAC on Mac/Linux, because there's no GAC.

src/Package/Microsoft.Build.UnGAC/Program.cs
rainersigwald: They're the only versions that we use, so they're the only ones that need to be removed from the GAC. Prior to VS 15, our assemblies were GACed by design; after they're not and they're all `15.1.0.0`.

eng/Signing.props
```diff
@@ -2,7 +2,7 @@
     <ItemGroup>
 		<FileSignInfo Include="Nerdbank.Streams.dll" CertificateName="3PartySHA2" /> 
 		<FileSignInfo Include="MessagePack.dll" CertificateName="3PartySHA2" /> 
-		<FileSignInfo Include="MessagePack.Annotations.dll" CertificateName="3PartySHA2" /> 
-		<ItemsToSign Include="$(VisualStudioSetupOutputPath)DevDivPackages\*.nupkg" />
+		<FileSignInfo Include="MessagePack.Annotations.dll" CertificateName="3PartySHA2" />
```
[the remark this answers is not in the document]
rainersigwald: We didn't sign it in the first place; this line was just in error.

src/Package/Microsoft.Build.UnGAC/NativeMethods.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        [PreserveSig]
+        uint UninstallAssembly(uint dwFlags, [MarshalAs(UnmanagedType.LPWStr)] string pszAssemblyName, IntPtr pRefData, ref ulong pulDisposition);
+    }
+
+    public static class NativeMethods
+    {
+        [DllImport("fusion.dll")]
+        internal static extern uint CreateAssemblyCache(out IAssemblyCache ppAsmCache, int dwReserved);
+
```
rainersigwald, line 22: nit: 
```suggestion
```

src/Package/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+                    "Microsoft.Build.Utilities.Core, Version=15.1.0.0",
+                    "Microsoft.Build.Conversion.Core, Version=15.1.0.0"
+                };
+
+                uint hresult = NativeMethods.CreateAssemblyCache(out IAssemblyCache assemblyCache, 0);
+
+                // Most significant bit is set, meaning there was an error in the Hresult.
+                if ((hresult >> 31) == 1)
+                {
+                    Console.WriteLine($"Could not successfully call CreateAssemblyCache. HResult: {hresult}");
```
rainersigwald, line 35: Format HRESULTs as 
```suggestion
                    Console.WriteLine($"Could not successfully call CreateAssemblyCache. HResult: 0x{hresult:X8}");
```

src/Package/Microsoft.Build.UnGAC/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+                    Console.WriteLine("Exiting without removing assemblies from the GAC...");
+                    return;
+                }
+
+                foreach (string assembly in assembliesToUnGAC)
+                {
+                    hresult = assemblyCache.UninstallAssembly(dwFlags: 0, pszAssemblyName: assembly, pRefData: IntPtr.Zero, pulDisposition: 0);
+
+                    // If we hit an error with an assembly, keep trying the others.
+                    if ((hresult >> 31) == 1)
```
rainersigwald, line 45: Should we just not bother with this `if`?

> Tried to remove {assembly} from the GAC. HR: 0x{hresult:X8}

?

rainersigwald (commented):
I'd like to see your insertion-automation changes before signing off.

src/Package/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: I'm actually not opposed to this idea. I'll do this.

src/Package/Microsoft.Build.UnGAC/Program.cs
BenVillalobos: It should, but I'm modifying this now to not check whether it was successful. It will just print out that it tried, and what the result was.

rainersigwald (approved):

BenVillalobos:
@rainersigwald you can see it here: https://dev.azure.com/devdiv/DevDiv/_releaseProgress?releaseId=815300&_a=release-pipeline-progress
The only change was under Variables/InsertJsonValues. Changed from `Microsoft.Build.vsman{$(MSBuild_ExtApisPackageVersion)+$(Build.SourceVersion)}=https://vsdrop.corp.microsoft.com/file/v1/Products/DevDiv/dotnet/msbuild/$(Build.SourceBranchName)/$(Build.BuildNumber);Microsoft.Build.vsman`
to
`Microsoft.Build.vsman{$(MSBuild_ExtApisPackageVersion)+$(Build.SourceVersion)}=https://vsdrop.corp.microsoft.com/file/v1/Products/DevDiv/dotnet/msbuild/$(Build.SourceBranchName)/$(Build.BuildNumber);Microsoft.Build.vsman,Microsoft.Build.UnGAC.vsman=https://vsdrop.corp.microsoft.com/file/v1/Products/DevDiv/dotnet/msbuild/$(Build.SourceBranchName)/$(Build.BuildNumber);Microsoft.Build.UnGAC.vsman`

Note that I removed the version from the ungac vsman, and there comma separating `Microsoft.Build.vsman` and `Microsoft.Build.UnGAC`. If you modify the Insert VS Payload step in the release and find the `Component.json values` field, there's a little guide there (see Example 2) that shows you can do this.

rainersigwald:
@BenVillalobos that change looks good. I think this is good to go. Please change the real insertion job after you merge it.

PR #5629, opened 2020-08-07
Optimize StringBuilder.Append/AppendLine call

elachlan:
The change optimises the usage of stringbuilder to avoid string operations and instead use stringbuilder.
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1197.md


BenVillalobos:

elachlan:
> I think that when you replace things like `AppendLine($"\t\t{dependency.Guid} = {dependency.Guid}"` with `Append("\t\t").Append(dependency.Guid).Append(" = ").AppendLine(dependency.Guid)`, it allocates the same total size of strings but does it in small pieces, making it actually less efficient, but I'm not an expert. Is that true?

Good catch, I'll switch the interpolated strings to AppendFormat with AppendLine.

elachlan:
@Forgind Could you check with someone smarter than myself to see if they should be left alone or if StringBuilder.Format should be used? I am not confident it will be better.

Forgind:
I'd highly recommend taking a look at https://benchmarkdotnet.org/
I think you'd really enjoy playing with the numbers for this sort of thing. 😃 

The results were pretty decisive. My recommendation allocated almost twice as much and ran about 20% slower, so...I retract my suggestion.

elachlan:
> The results were pretty decisive. My recommendation allocated almost twice as much and ran about 20% slower, so...I retract my suggestion.

Did you test String.format vs the individual stings?

Forgind:
Nope! Forgot. I tested the original strings vs. appending them individually. I just tested StringBuilder.AppendFormat, and it was intermediate—closer to how it is now in speed but closer to appending them individually as far as allocations.

elachlan:
> Nope! Forgot. I tested the original strings vs. appending them individually. I just tested StringBuilder.AppendFormat, and it was intermediate—closer to how it is now in speed but closer to appending them individually as far as allocations.

So basically AppendFormat is slower than the individual appends and has similar allocations? So I guess we stick with the individual appends?

Forgind:
Yes

elachlan:
That is surprising, but it kind of makes sense. So I guess there isn't any more changes then?

Forgind:
I'm content with it, at least. I'll still probably run it by people to make sure they're ok with it, too, before merging.

src/Build.OM.UnitTests/Definition/Project_Tests.cs
```diff
@@ -2866,7 +2866,7 @@ public void GetItemProvenanceResultsShouldBeInItemElementOrder()
             var sb = new StringBuilder();
             for (int i = 0; i < itemElements; i++)
             {
-                sb.AppendLine($"<i_{i} Include=\"a\"/>");
+                sb.Append("<i_").Append(i).AppendLine(" Include=\"a\"/>");
```
BenVillalobos, line 2869: Is the goal here specifically for memory improvement? I'm not particularly great at reading IL, but it [looks like](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyAMABMugHQAqApgB7ADcOyAzEagQMIEDeOBvRzyACwEAsgAoAlFx59ZAZWAAnMADsA5gCEArpBgVFBAM4AjAgF4CKigHcCC5eu279k+tlmzVwAmHME8bh58JiQAggAO4RQqMAAyqhRiACQARAA8YAD6nGAAvgQAkioAxhBaemYAOikAhtUA9AB8KRKBQUbGYZHRMGLpWS1dUTFiYBJDPfFWfYUlZRXVdSlNLW28uTi5QA) there's an allocation that happens in the `AppendLine()` format with the integer, whereas the `Append().Append().AppendLine()` format has more instructions but no `box` being called and therefore no allocation.

Forgind: The benchmarking I mentioned [above ](https://github.com/dotnet/msbuild/pull/5629#issuecomment-671649072) was for this specifically:
```     [Benchmark]
        public StringBuilder AtOnce()
        {
            StringBuilder sb = new StringBuilder();
            for (int i = 0; i < NumStrings; i++)
            sb.AppendLine($"hi {i} bye {DateTime.Now.ToString()} done");
            return sb;
        }
        [Benchmark]
        public StringBuilder WithFormat()
        {
            StringBuilder sb = new StringBuilder();
            for (int i = 0; i < NumStrings; i++)
                sb.AppendFormat("hi {0} bye {1} done", i, DateTime.Now.ToString());
            return sb;
        }
        [Benchmark]
        public StringBuilder InParts()
        {
            StringBuilder sb = new StringBuilder();
            for (int i = 0; i < NumStrings; i++)
                sb.Append("hi ").Append(i).Append(" bye ").Append(DateTime.Now.ToString()).AppendLine(" done");
            return sb;
        }
```

And it was pretty decisive. It does allocate but not nearly as much.

Forgind: (Although I think it allocates on the stack?)

Forgind: In any case:
image

PR #5631, opened 2020-08-08
change empty string comparison

elachlan:
Found some string.empty comparisons using Rosylnator and converted them to string.IsNullOrEmpty

https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1156.md

### DIFF
--- a/src/Build/Globbing/MSBuildGlob.cs
+++ b/src/Build/Globbing/MSBuildGlob.cs
@@ -144,7 +144,7 @@ namespace Microsoft.Build.Globbing
 
             // Degenerate case when the string to match is empty.
             // Ensure trailing slash because the fixed directory part has a trailing slash.
-            if (stringToMatch == string.Empty)
+            if (string.IsNullOrEmpty(stringToMatch))
             {
                 normalizedInput += Path.DirectorySeparatorChar;
             }
@@ -171,7 +171,7 @@ namespace Microsoft.Build.Globbing
             ErrorUtilities.VerifyThrowArgumentNull(fileSpec, nameof(fileSpec));
             ErrorUtilities.VerifyThrowArgumentInvalidPath(globRoot, nameof(globRoot));
 
-            if (globRoot == string.Empty)
+            if (string.IsNullOrEmpty(globRoot))
             {
                 globRoot = Directory.GetCurrentDirectory();
             }
--- a/src/Build/Globbing/MSBuildGlob.cs
+++ b/src/Build/Globbing/MSBuildGlob.cs
@@ -171,7 +171,7 @@ namespace Microsoft.Build.Globbing
             ErrorUtilities.VerifyThrowArgumentNull(fileSpec, nameof(fileSpec));
             ErrorUtilities.VerifyThrowArgumentInvalidPath(globRoot, nameof(globRoot));
 
-            if (string.IsNullOrEmpty(globRoot))
+            if (globRoot.Length == 0)
             {
                 globRoot = Directory.GetCurrentDirectory();
             }
--- a/src/Build/Instance/ProjectTaskInstance.cs
+++ b/src/Build/Instance/ProjectTaskInstance.cs
@@ -141,10 +141,10 @@ namespace Microsoft.Build.Execution
             new CopyOnWriteDictionary<string, (string, ElementLocation)>(8, StringComparer.OrdinalIgnoreCase),
             new List<ProjectTaskInstanceChild>(),
             location,
-            condition == string.Empty ? null : ElementLocation.EmptyLocation,
-            continueOnError == string.Empty ? null : ElementLocation.EmptyLocation,
-            msbuildRuntime == string.Empty ? null : ElementLocation.EmptyLocation,
-            msbuildArchitecture == string.Empty ? null : ElementLocation.EmptyLocation)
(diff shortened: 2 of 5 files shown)


BenVillalobos:

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -171,7 +171,7 @@ public static MSBuildGlob Parse(string globRoot, string fileSpec)
             ErrorUtilities.VerifyThrowArgumentNull(fileSpec, nameof(fileSpec));
             ErrorUtilities.VerifyThrowArgumentInvalidPath(globRoot, nameof(globRoot));
 
-            if (globRoot == string.Empty)
+            if (string.IsNullOrEmpty(globRoot))
```
BenVillalobos, line 174: So globRoot is guaranteed to not be null as `ErrorUtilities.VerifyThrowArgumentNull` is called just before this. The change isn't necessary and adds an extra stack frame to what can stay as a simple string check. If anything this should (and all other uses of `X == string.Empty`) be replaced with `globRoot.Length == 0`

src/Build/Globbing/MSBuildGlob.cs
```diff
@@ -144,7 +144,7 @@ private string NormalizeMatchInput(string stringToMatch)
 
             // Degenerate case when the string to match is empty.
             // Ensure trailing slash because the fixed directory part has a trailing slash.
-            if (stringToMatch == string.Empty)
+            if (string.IsNullOrEmpty(stringToMatch))
```
BenVillalobos, line 147: Both callers of `NormalizeMatchInput` guarantee `stringToMatch` ensure it isn't null. I suppose one could make the argument that we may call this function in the future and potentially pass null in.

src/Build/Instance/ProjectTaskInstance.cs
```diff
@@ -141,10 +141,10 @@ string msbuildArchitecture
             new CopyOnWriteDictionary<string, (string, ElementLocation)>(8, StringComparer.OrdinalIgnoreCase),
             new List<ProjectTaskInstanceChild>(),
             location,
-            condition == string.Empty ? null : ElementLocation.EmptyLocation,
-            continueOnError == string.Empty ? null : ElementLocation.EmptyLocation,
-            msbuildRuntime == string.Empty ? null : ElementLocation.EmptyLocation,
-            msbuildArchitecture == string.Empty ? null : ElementLocation.EmptyLocation)
+            string.IsNullOrEmpty(condition) ? null : ElementLocation.EmptyLocation,
```
BenVillalobos, line 144: This one technically changes the code path. If `null` were passed before, `ElementLocation.EmptyLocation` would be used. If `null` were passed with this change, it would pass null into the base constructor. I'm not sure what that realistically means for this code though.

Looks like these values would never be null either.
```
        internal ProjectTaskInstance AddTask(string taskName, string condition, string continueOnError, string msbuildRuntime, string msbuildArchitecture)
        {
            ErrorUtilities.VerifyThrowInternalLength(taskName, "taskName");
            ProjectTaskInstance task = new ProjectTaskInstance(taskName, _location, condition ?? String.Empty, continueOnError ?? String.Empty, msbuildRuntime ?? String.Empty, msbuildArchitecture ?? String.Empty);
            this.AddProjectTargetInstanceChild(task);
            return task;
        }
```

src/Build/Globbing/MSBuildGlob.cs
BenVillalobos: Had to verify for myself. `IsNullOrEmpty` checks `str == null || str.Length == 0`. Since we know, some of these couldn't possibly be null, just checking length would be better.

src/Build/Instance/TaskRegistry.cs
```diff
@@ -379,9 +379,8 @@ IFileSystem fileSystem
             if ((runtime != String.Empty) || (architecture != String.Empty))
             {
                 taskFactoryParameters = CreateTaskFactoryParametersDictionary();
-
-                taskFactoryParameters.Add(XMakeAttributes.runtime, runtime == String.Empty ? XMakeAttributes.MSBuildRuntimeValues.any : runtime);
-                taskFactoryParameters.Add(XMakeAttributes.architecture, architecture == String.Empty ? XMakeAttributes.MSBuildArchitectureValues.any : architecture);
+                taskFactoryParameters.Add(XMakeAttributes.runtime, string.IsNullOrEmpty(runtime) ? XMakeAttributes.MSBuildRuntimeValues.any : runtime);
```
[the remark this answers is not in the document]
elachlan: We also check for an empty string above. Very strange.

src/Build/Instance/ProjectTaskInstance.cs
elachlan: What would you like to do? I can revert the change if you would like.

BenVillalobos: I lean toward reverting this particular change. In the PR review meeting tomorrow we'll determine the direction we want to go here. Thanks for the quick responses by the way!

Forgind:
Team triage:
Since Ben found that there wasn't a danger of a null pointer exception, and these are roughly equal as far as style is concerned, we'd prefer not to take this one.

Issue #5618, opened 2020-08-06
Add test hook for SDK resolvers


sfoslund:
See original comment here: https://github.com/dotnet/sdk/pull/12633#discussion_r466053772

The SDK would like to be able to run end to end tests for the SDK resolver, which would require adding the ability to specify an addition place to look for SDK resolvers in addition to the SdkResolvers folder in the MSBuild bin directory.

PR #5640, opened 2020-08-10
Adding test hook for SDK resolver end to end tests

sfoslund:
Fixes https://github.com/dotnet/msbuild/issues/5618


BenVillalobos:
This was closed/reopened because it ran into https://github.com/dotnet/msbuild/issues/5646.

sfoslund:
@dsplaisted I've updated this PR so it now uses the resolvers from `RootFolder` unless there is one with a matching name under the directory pointed to by the new environment variable, in which case the resolver from the environment variable wins.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -78,6 +83,29 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder, Elem
...
+                subfolders = rootFolderInfo.GetDirectories();
+            }
+
+            if (additionalResolversFolder != null)
+            {
+                var resolversDirInfo = new DirectoryInfo(additionalResolversFolder);
+                if (resolversDirInfo.Exists)
+                {
+                    var overrideFolders = resolversDirInfo.GetDirectories();
+                    var overriddenFolders = subfolders.Where(subfolder => overrideFolders.Any(overrideFolder => overrideFolder.Name.Equals(subfolder.Name)));
```
dsplaisted, line 101: Probably the name comparison should be case insensitive.

Forgind, line 101: This would be much prettier if you made an EqualityComparer for this, wrapped it in a HashSet, added all the elements of overrideFolders to that HashSet, and did:
overrideFolders.Concat(subfolders.Where(e => set.Add(e)));

This is also very inefficient as-is. Make sure to use StringComparer.OrdinalIgnoreCase if on a case-insensitive OS.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -23,8 +23,9 @@ internal class SdkResolverLoader
         internal virtual IList<SdkResolver> LoadResolvers(LoggingContext loggingContext,
             ElementLocation location)
         {
-            // Always add the default resolver
-            var resolvers = new List<SdkResolver> {new DefaultSdkResolver()};
+            var resolvers = Environment.GetEnvironmentVariable("MSBUILDINCLUDEDEFAULTSDKRESOLVER")?.ToLowerInvariant() != "false" ? 
+                new List<SdkResolver> {new DefaultSdkResolver()}
+                : new List<SdkResolver>();
```
dsplaisted, line 28: I'm now realizing this does something different than what I originally thought.  It disables the default SDK resolver, which is different than disabling loading resolvers from the default Sdk folder.  I can't remember what you said you added this for, are you sure it is going to do what you meant?

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -78,6 +83,29 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder, Elem
             return assembliesList;
         }
 
+        private DirectoryInfo[] GetSubfolders(string rootFolder, string additionalResolversFolder)
+        {
+            DirectoryInfo[] subfolders = null;
+            var rootFolderInfo = new DirectoryInfo(rootFolder);
```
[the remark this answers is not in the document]
sfoslund: I don't think we have to worry about this case, rootFolder should never be null and we didn't have a null check before this change either.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
sfoslund: I included this so that I could ensure SDK resolution fails when we can't find the global.json specified version (before my change). When I was testing this test hook along with my SDK tests ([branch here](https://github.com/dotnet/sdk/compare/master...sfoslund:ResolverTests?expand=1)), SDK resolution would fail with the MSBuildSDK resolver so it would fall back to the default resolver which would actually succeed. I wanted to have a test case to ensure that outside of VS SDK resolution really does fail, which requires disabling the default SDK.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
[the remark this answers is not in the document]
Forgind: https://github.com/dotnet/msbuild/blob/master/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs#L56-L59 is a null check.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
sfoslund: I'm not fully convinced that this would be any cleaner, it seems like a waste to allocate a hashset and create a comparer when there are other alternatives.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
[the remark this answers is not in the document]
sfoslund: Oh you're right, I missed that thank you :)

sfoslund:
> If it is not obvious what the type of a variable is, do not use var. That just makes it harder to read.

@Forgind is there anywhere in particular you'd prefer I not use var?

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
dsplaisted: Note that this code should only be running when the test hook is enabled, so the perf is not as critical.

Generally a HashSet is probably better, but I'm not really sure if there are usually only 3 or so items in the sets.

Forgind:
My preference is to never use var unless you actually don't know the type of the variable. Most people are a little more lax and say you can/should use var if the right side of the equation is "new ___", since that's pretty obvious. If you assign a new variable to the return value of a function, and you know the type, I can't think of any reason to use var.

dsplaisted:
I think there are strong opinions on both side of the `var` debate.  I tend to like using it, especially for generic types which can get long.

sfoslund:
@Forgind personally I'm with Daniel, I like using it pretty universally. That also seems to be the pattern in this file (there are very few places where it's not used) so I was trying to stick with precedence.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
sfoslund: That's fine with me, I've updated the PR.

sfoslund:
@dsplaisted does any of your current resolver work depend on this? In case it does, I've pushed the SDK side of this change (https://github.com/dotnet/sdk/pull/12875) which should work once this flows. That way you won't be blocked in case this doesn't go in this week (as I'll be OOF mon-weds next week).

dsplaisted:
> @dsplaisted does any of your current resolver work depend on this? In case it does, I've pushed the SDK side of this change ([dotnet/sdk#12875](https://github.com/dotnet/sdk/pull/12875)) which should work once this flows. That way you won't be blocked in case this doesn't go in this week (as I'll be OOF mon-weds next week).

It will be helpful to have this for my resolver work.  I need the test hook for automated tests on full Framework MSBuild, and the code to locate workloads is different on Full Framework, so it would be very good to be able to test it.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -20,11 +20,16 @@ internal class SdkResolverLoader
...
+        private readonly string IncludeDefaultResolver = Environment.GetEnvironmentVariable("MSBUILDINCLUDEDEFAULTSDKRESOLVER");
+
+        private readonly string AdditionalResolversFolder = Environment.GetEnvironmentVariable("MSBUILDADDITIONALSDKRESOLVERSFOLDER");
+
         internal virtual IList<SdkResolver> LoadResolvers(LoggingContext loggingContext,
             ElementLocation location)
         {
-            // Always add the default resolver
-            var resolvers = new List<SdkResolver> {new DefaultSdkResolver()};
+            var resolvers = !String.Equals(IncludeDefaultResolver, "false", StringComparison.OrdinalIgnoreCase) ?
```
Forgind, line 30: Part of what I don't like about var is that this is the third variable named `resolvers`, but it's a different type than the other two. This one isn't the end of the world, though.

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -53,12 +58,14 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder, Elem
...
             var assembliesList = new List<string>();
 
-            if (string.IsNullOrEmpty(rootFolder) || !FileUtilities.DirectoryExistsNoThrow(rootFolder))
+            if ((string.IsNullOrEmpty(rootFolder) || !FileUtilities.DirectoryExistsNoThrow(rootFolder)) && AdditionalResolversFolder == null)
             {
                 return assembliesList;
             }
 
-            foreach (var subfolder in new DirectoryInfo(rootFolder).GetDirectories())
+            var subfolders = GetSubfolders(rootFolder, AdditionalResolversFolder);
```
Forgind, line 66: type

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -78,6 +85,41 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder, Elem
...
+            {
+                subfolders = new DirectoryInfo(rootFolder).GetDirectories();
+            }
+
+            if (additionalResolversFolder != null)
+            {
+                var resolversDirInfo = new DirectoryInfo(additionalResolversFolder);
+                if (resolversDirInfo.Exists)
+                {
+                    var overrideFolders = resolversDirInfo.GetDirectories();
```
Forgind, line 101: type

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -78,6 +85,41 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder, Elem
...
+                    var foldersSet = overrideFolders.ToHashSet(new DirInfoNameComparer());
```
BenVillalobos, line 102: I wonder if we'll ever hit a scenario where someone has a ridiculous number of directories such that this impacts perf. This is more a comment than an actual suggestion.

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -271,6 +271,73 @@ public void SdkResolverLoaderErrorsWhenManifestTargetMissing()
...
+                    File.WriteAllText(resolver2Path, string.Empty);
+                    var resolver3 = "Resolver3";
+                    var resolver3Path = Path.Combine(additionalRoot, resolver3, $"{resolver3}.dll");
+                    Directory.CreateDirectory(Path.Combine(additionalRoot, resolver3));
+                    File.WriteAllText(resolver3Path, string.Empty);
+
+                    Environment.SetEnvironmentVariable("MSBUILDADDITIONALSDKRESOLVERSFOLDER", additionalRoot);
+
+                    SdkResolverLoader loader = new SdkResolverLoader();
+                    var resolvers = loader.FindPotentialSdkResolvers(testRoot, new MockElementLocation("file"));
```
Forgind, line 330: type

src/Build.UnitTests/BackEnd/SdkResolverLoader_Tests.cs
```diff
@@ -271,6 +271,73 @@ public void SdkResolverLoaderErrorsWhenManifestTargetMissing()
...
+                {
+                    var testRoot = env.CreateFolder().Path;
+                    Environment.SetEnvironmentVariable("MSBUILDINCLUDEDEFAULTSDKRESOLVER", "false");
+                    SdkResolverLoader loader = new MockSdkResolverLoader()
+                    {
+                        LoadResolversAction = (resolverPath, loggingContext, location, resolvers) => {
+                            resolvers.Add(new MockSdkResolverWithAssemblyPath(resolverPath));
+                        }
+                    };
+                    var resolvers = loader.LoadResolvers(_loggingContext, new MockElementLocation("file"));
```
Forgind, line 290: type

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
dsplaisted: Note that this would also only happen with the test hook enabled.

Forgind: Also, this is now asymptotically optimal. I don't think it's likely that it _has_ to be that good, but MSBuild users continually surprise me...

BenVillalobos: Considering what Daniel mentioned, I'm fine with this as-is.

cdmihai:
There already exists an environment variable based way of specifying sdk resolver paths: https://github.com/cdmihai/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Shared/BuildEnvironmentHelper.cs#L580

Why add additional mechanisms instead of reusing / enhancing the existing one?

dsplaisted:
> There already exists an environment variable based way of specifying sdk resolver paths: https://github.com/cdmihai/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Shared/BuildEnvironmentHelper.cs#L580
> 
> Why add additional mechanisms instead of reusing / enhancing the existing one? (for enhancing it, we could change the existing env var to contain a semicolon delimited list of paths)

The existing mechanism lets you set a path for SDKs.  We want to be able to override an individual SDK Resolver so that we can specifically test the resolver behavior.  For example, see https://github.com/dotnet/sdk/pull/12875, and we'd also like to use the test hook to test [the workload resolver](https://github.com/dotnet/designs/blob/master/accepted/2020/workloads/workload-resolvers.md).

src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs
```diff
@@ -78,6 +85,41 @@ internal virtual IList<string> FindPotentialSdkResolvers(string rootFolder, Elem
...
+                    return overrideFolders.Concat(subfolders.Where(subfolder => foldersSet.Add(subfolder))).ToArray();
```
cdmihai, line 103: You can simplify this with [HashSet.UnionWith](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.unionwith?view=netframework-4.7.2)

sfoslund:
@BenVillalobos @Forgind I've updated with your feedback, thanks!

PR #5647, opened 2020-08-11
Simplify null checks

elachlan:
Refactoring to simplify null checks.

### DIFF
--- a/src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
+++ b/src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
@@ -1239,7 +1239,7 @@ namespace Microsoft.Build.BackEnd.Logging
                 string targetName = string.Empty;
 
                 // Does the context (Project, Node, Context, Target, NOT task) of the previous event match the current message
-                bool contextAreEqual = s_compareContextNodeIdTargetId.Equals(currentBuildEventContext, _lastDisplayedBuildEventContext ?? null);
+                bool contextAreEqual = s_compareContextNodeIdTargetId.Equals(currentBuildEventContext, _lastDisplayedBuildEventContext);
 
                 TargetStartedEventMinimumFields targetStartedEvent = null;
                 // If the previous event does not have the same target context information, the target name needs to be printed to the console
--- a/src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
+++ b/src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
@@ -605,7 +605,7 @@ namespace Microsoft.Build.BackEnd.Logging
 
                         string projectName = string.Empty;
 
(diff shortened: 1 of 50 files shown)


src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
```diff
@@ -1239,7 +1239,7 @@ private void PrintTargetNamePerMessage(BuildMessageEventArgs e, bool lightenText
                 string targetName = string.Empty;
 
                 // Does the context (Project, Node, Context, Target, NOT task) of the previous event match the current message
-                bool contextAreEqual = s_compareContextNodeIdTargetId.Equals(currentBuildEventContext, _lastDisplayedBuildEventContext == null ? null : _lastDisplayedBuildEventContext);
+                bool contextAreEqual = s_compareContextNodeIdTargetId.Equals(currentBuildEventContext, _lastDisplayedBuildEventContext ?? null);
```
BenVillalobos, line 1242: Is this even worth a null check? Even if not, I'd rather not make the change in this PR so we keep this one focused.

src/Tasks/GenerateResource.cs
```diff
@@ -1833,7 +1833,7 @@ private bool NeedSeparateAppDomain()
                                             (
                                                 MessageImportance.Low,
                                                 "GenerateResource.SeparateAppDomainBecauseOfType",
-                                                (name == null) ? String.Empty : name,
+name ?? String.Empty,
```
BenVillalobos, line 1836: Fix formatting here and in the other 2 changes below.

src/Tasks/GenerateResource.cs
```diff
@@ -1879,7 +1879,7 @@ private bool NeedSeparateAppDomain()
                                             (
                                                 MessageImportance.Low,
                                                 "GenerateResource.SeparateAppDomainBecauseOfMimeType",
-                                                (name == null) ? String.Empty : name,
+name ?? String.Empty,
```
BenVillalobos, line 1882: same as above

src/Tasks/GenerateResource.cs
```diff
@@ -1925,7 +1925,7 @@ private bool NeedSeparateAppDomain()
                                         MessageImportance.Low,
                                         "GenerateResource.SeparateAppDomainBecauseOfErrorDeserializingLineNumber",
                                         source.ItemSpec,
-                                        (name == null) ? String.Empty : name,
+name ?? String.Empty,
```
BenVillalobos, line 1928: same as above

src/Build/Logging/ParallelLogger/ParallelConsoleLogger.cs
elachlan: There was a refactor suggestion to remove it. So I did. I think it comes under a simplification of null checks.

PR #5648, opened 2020-08-11
Avoid nullreferenceexception

elachlan:
Refactoring to avoid NullReferenceException.
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1202.md


Forgind:
I actually don't see how availablePlatformVersions could be null either. I like condensing 8 lines into 1 at the end, although I'd take out all but one `?`. You're right that it probably isn't critical. I'm fine with closing it.

PR #5649, opened 2020-08-11
Remove redundant assignment

elachlan:
Refactoring to remove redundant assignments.
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1212.md

### DIFF
--- a/src/Build.UnitTests/BackEnd/TaskHostTaskCancelled_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TaskHostTaskCancelled_Tests.cs
@@ -35,8 +35,6 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             ((ITranslatable)cancelled).Translate(TranslationHelpers.GetWriteTranslator());
             INodePacket packet = TaskHostTaskCancelled.FactoryForDeserialization(TranslationHelpers.GetReadTranslator());
-
-            TaskHostTaskCancelled deserializedCancelled = packet as TaskHostTaskCancelled;
         }
     }
 }
--- a/src/Build.UnitTests/BackEnd/TaskHostTaskCancelled_Tests.cs
+++ b/src/Build.UnitTests/BackEnd/TaskHostTaskCancelled_Tests.cs
@@ -8,6 +8,7 @@ using System.Linq;
 
 using Microsoft.Build.BackEnd;
 using Xunit;
+using Shouldly;
 
 namespace Microsoft.Build.UnitTests.BackEnd
 {
@@ -35,6 +36,7 @@ namespace Microsoft.Build.UnitTests.BackEnd
 
             ((ITranslatable)cancelled).Translate(TranslationHelpers.GetWriteTranslator());
             INodePacket packet = TaskHostTaskCancelled.FactoryForDeserialization(TranslationHelpers.GetReadTranslator());
+            packet.ShouldBeOfType<TaskHostTaskCancelled>();
         }
     }
 }
--- a/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
+++ b/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
@@ -4914,7 +4914,6 @@ namespace Microsoft.Build.UnitTests.Evaluation
         /// </summary>
         private void VerifyImportTargetRelativePath(string directory, string directory2, string[] imports)
         {
-            string file0 = null;
             string file1 = null;
             string file2 = null;
             string file3 = null;
@@ -4927,7 +4926,7 @@ namespace Microsoft.Build.UnitTests.Evaluation
                     FileUtilities.DeleteWithoutTrailingBackslash(directory);
                 }
 
-                file0 = Path.Combine(directory, "my.proj");
+                string file0 = Path.Combine(directory, "my.proj");
                 file1 = Path.Combine(directory, "1.targets");
                 file2 = Path.Combine(directory2, "2.targets");
                 file3 = Path.Combine(directory2, "3.cpp.targets");
--- a/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
@@ -963,7 +963,7 @@ namespace Microsoft.Build.BackEnd
             string oldestOutput = EscapingUtilities.UnescapeAll(FileUtilities.FixFilePath(outputs[0].ToString()));
             ErrorUtilities.ThrowIfTypeDoesNotImplementToString(outputs[0]);
 
-            DateTime oldestOutputFileTime;
+            DateTime oldestOutputFileTime = DateTime.MinValue;
             try
             {
                 string oldestOutputFullPath = Path.Combine(projectDirectory, oldestOutput);
@@ -988,7 +988,7 @@ namespace Microsoft.Build.BackEnd
             {
                 string candidateOutput = EscapingUtilities.UnescapeAll(FileUtilities.FixFilePath(outputs[i].ToString()));
                 ErrorUtilities.ThrowIfTypeDoesNotImplementToString(outputs[i]);
-                DateTime candidateOutputFileTime;
+                DateTime candidateOutputFileTime = DateTime.MinValue;
                 try
                 {
                     string candidateOutputFullPath = Path.Combine(projectDirectory, candidateOutput);
--- a/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
+++ b/src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
@@ -153,21 +153,13 @@ namespace Microsoft.Build.BackEnd
             }
             else
             {
-                ItemVectorPartitionCollection itemVectorsInTargetInputs;
-                ItemVectorPartitionCollection itemVectorTransformsInTargetInputs;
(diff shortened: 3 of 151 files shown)


src/Deprecated/Engine/Solution/SolutionWrapperProject.cs
```diff
@@ -755,7 +756,7 @@ static private void AddResolveProjectReferenceTasks
                         catch (Exception e)
                         {
                             if (ExceptionHandling.NotExpectedException(e))
-                               throw;
+                                throw;
```
Forgind, line 759: One option that's even nicer:
```suggestion
                         catch (Exception e) when (!ExceptionHandling.NotExpectedException(e))
```

src/Tasks.UnitTests/XslTransformation_Tests.cs
```diff
@@ -1126,15 +1126,13 @@ private void Compile(string inputFile, string outputFile)
...
-            CompilerErrorCollection errors = null;
             try
             {
                 using (XmlReader reader = XmlReader.Create(sourceUri, readerSettings))
                 {
-                    errors = XslCompiledTransform.CompileToType(
-                        reader, xsltSettings, xmlResolver, false, typeBldr, scriptAsmPath
-                    );
+                    CompilerErrorCollection errors = XslCompiledTransform.CompileToType(
+            reader, xsltSettings, xmlResolver, false, typeBldr, scriptAsmPath
```
Forgind, line 1134: indent a bit more?

src/Deprecated/Engine/Solution/SolutionWrapperProject.cs
elachlan: I was going to put in a separate PR for exception filters.

src/Build.UnitTests/BackEnd/TaskHostTaskCancelled_Tests.cs
```diff
@@ -35,8 +35,6 @@ public void TestTranslation()
 
             ((ITranslatable)cancelled).Translate(TranslationHelpers.GetWriteTranslator());
             INodePacket packet = TaskHostTaskCancelled.FactoryForDeserialization(TranslationHelpers.GetReadTranslator());
-
-            TaskHostTaskCancelled deserializedCancelled = packet as TaskHostTaskCancelled;
```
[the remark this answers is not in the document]
elachlan: `Assert.True(packet is TaskHostTaskCancelled);` ?

src/Build.UnitTests/Evaluation/Evaluator_Tests.cs
```diff
@@ -4927,7 +4926,7 @@ private void VerifyImportTargetRelativePath(string directory, string directory2,
                     FileUtilities.DeleteWithoutTrailingBackslash(directory);
                 }
 
-                file0 = Path.Combine(directory, "my.proj");
+                string file0 = Path.Combine(directory, "my.proj");
```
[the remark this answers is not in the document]
elachlan: I am pretty sure its because it is never created. I think it is kept in memory. See the description of ProjectRootElement.Create.

src/Build.UnitTests/BackEnd/TaskHostTaskCancelled_Tests.cs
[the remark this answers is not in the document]
Forgind: packet.ShouldBeOfType<TaskHostTaskCancelled>();

We're gently trying to use Shouldly a little more because it tends to give better error messages.

elachlan:
> Looks good! I'd still like to see the patterns of the form:
> 
> ```
> <type> <name>;
> <functionCall>(<param0>, <param1>, out <name>);
> ```
> 
> or equivalent translated into
> `<functionCall>(<param0>, <param1>, out <type> <name>);`
> But we can save that for later.

I can do another refactor PR for that. There are a lot of changes like that so it will be easier to review in a separate issue.

src/Deprecated/Conversion/ProjectFileConverter.cs
```diff
@@ -2629,7 +2627,6 @@ private static ProjectItemElement ConvertClassicComReference(XmlElementWithLocat
         /// <returns></returns>
         private ProjectItemElement ConvertProjectToProjectReference(XmlElementWithLocation referenceElement, ProjectItemGroupElement referencesItemGroup, string referenceName, ref string referencedProjectGuid)
         {
-            ProjectItemElement newReferenceItem;
```
BenVillalobos, line 2632: This one is unsafe to delete. When this function returns you can just return `referencesItemGroup.AddItem(XMakeProjectStrings.projectReference, ProjectCollection.Escape(value));`

src/Deprecated/Engine/Engine/BuildTask.cs
```diff
@@ -365,11 +365,10 @@ string attributeName
...
 
-            string attributeValue;
-
+            
             // If this is a persisted Task, grab the attribute directly from the
             // task element.
-            attributeValue = taskElement.GetAttribute(attributeName);
+            string attributeValue = taskElement.GetAttribute(attributeName);
 
             return attributeValue ?? String.Empty;
```
BenVillalobos, line 373:
```suggestion
            return taskElement.GetAttribute(attributeName) ?? String.Empty;
```
This type of change matches with other changes made earlier.

Which raises a good question, do we want to make changes like this? It does make things _slightly_ harder to debug.

src/Deprecated/Conversion/ProjectFileConverter.cs
Forgind: This is actually part of why I like this change: mistakes like this cause compile-time errors instead of execution-time errors. It looks like, in this case, elachlan refactored it correctly, but when merging in master, overwrote the change you suggested without overwriting removing the declaration, hence the problem. Should be easy enough to fix.

src/Deprecated/Engine/Engine/BuildTask.cs
Forgind: I definitely vote yes. _Slightly_ harder to debug, maybe, but cleaner and easier to read, definitely.

elachlan:
> Reviewed half and then got tired and gave it up. As long as the class of issues that I commented on get fixed, looks good to me.
> 
> Since these big repetitive changes suck the life out of reviewers, maybe we should have a contribution rule where we restrain these types of changes somehow. Thinking out loud, one strategy would be to only allow them on the files touched by a PR that has a meaningful feature addition / bug fix change, or some other throttling mechanism. We could also define a set of linters to use and/or config files for them (resharper, VS, roslynator, etc)

Thanks for getting so far. I will create an issue to look at copying the Anlyzers and config from the dotnet/runtime codebase.

src/Build/ElementLocation/ElementLocation.cs
```diff
@@ -209,7 +209,7 @@ internal static ElementLocation Create(string file, int line, int column)
         /// </summary>
         private static string GetLocationString(string file, int line, int column)
         {
-            string locationString = String.Empty;
```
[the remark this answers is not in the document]
Forgind: It's guaranteed to be set (without having been read) in one of the three clauses below. If this is changed such that the "else" doesn't exist, for instance, it would throw a compile-time error, so this is actually safer than initializing it to empty.

cdmihai:
>Thanks for getting so far. I will create an issue to look at copying the Anlyzers and config from the dotnet/runtime codebase.

That would cool, thanks! It would also be cool if we had a bot which adds a commit to every PR with the rules applied to the changed files. That would trickle the changes in a more digestible manner. Alternatively, we could also just apply everything in one humongous PR (maybe linting + formatting), and then have a few weeks of stabilization where we check that nothing regressed.

elachlan:
> > Thanks for getting so far. I will create an issue to look at copying the Anlyzers and config from the dotnet/runtime codebase.
> 
> That would cool, thanks! It would also be cool if we had a bot which adds a commit to every PR with the rules applied to the changed files. That would trickle the changes in a more digestible manner. Alternatively, we could also just apply everything in one humongous PR (maybe linting + formatting), and then have a few weeks of stabilization where we check that nothing regressed. And after that have a bot apply them to each PR.

Here is the issue #5655.

Forgind:
Team triage:
Can you resolve @cdmihai's first two comments? Then we'll take it.

We're also thinking that we might use https://github.com/dotnet/msbuild/pull/5656 as a checklist to measure progress towards a nicely refactored repo, in which case you can put the lint ID you're addressing in the title to make it easy to find. We should have a separate team meeting to discuss what rules we want to keep and which we might discard, though, and we haven't had that yet. We'll get back to you when that happens.

src/Build/BackEnd/Components/RequestBuilder/TargetUpToDateChecker.cs
```diff
@@ -979,7 +963,7 @@ internal static bool IsAnyOutOfDate<T>(out DependencyAnalysisLogDetail dependenc
             string oldestOutput = EscapingUtilities.UnescapeAll(FileUtilities.FixFilePath(outputs[0].ToString()));
             ErrorUtilities.ThrowIfTypeDoesNotImplementToString(outputs[0]);
 
-            DateTime oldestOutputFileTime = DateTime.MinValue;
```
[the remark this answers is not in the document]
elachlan: `oldestOutputFileTime = NativeMethodsShared.GetLastWriteFileUtcTime` returns "The last write time of the file, or DateTime.MinValue if the file does not exist". So we don't need to assign DateTime.MinValue. Additionally the try catch also assigns it. Same thing with candidateOutputFileTime. I've added it back in for you though.

PR #5657, opened 2020-08-15
Get target framework info from inner builds in GetTargetFrameworks

dsplaisted:
In order to support implicitly setting the `TargetPlatformVersion` (for example for a `TargetFramework` of `net5.0-windows`), as well as to eventually support [TargetFramework values as aliases](https://github.com/NuGet/Home/issues/5154), we need to update the project reference protocol.  The `GetTargetFrameworks` target needs to return information not just about the `TargetFrameworks`, but also about the `TargetFrameworkMoniker` and `TargetPlatformMoniker` that it parses out to.  This requires dispatching an inner build to get this information.

The new information will be consumed by the `GetReferenceNearestTargetFrameworkTask` task, which is being updated here: https://github.com/NuGet/NuGet.Client/pull/3578

FYI @nkolev92

### DIFF
--- a/src/Tasks/Microsoft.Common.CrossTargeting.targets
+++ b/src/Tasks/Microsoft.Common.CrossTargeting.targets
@@ -21,13 +21,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Import Project="$(CustomBeforeMicrosoftCommonCrossTargetingTargets)" Condition="'$(CustomBeforeMicrosoftCommonCrossTargetingTargets)' != '' and Exists('$(CustomBeforeMicrosoftCommonCrossTargetingTargets)')"/>
 
   <Target Name="GetTargetFrameworks"
+          DependsOnTargets="GetTargetFrameworksWithPlatformFromInnerBuilds"
           Returns="@(_ThisProjectBuildMetadata)">
+
+    <Error Condition="'$(IsCrossTargetingBuild)' != 'true'"
+           Text="Internal MSBuild error: CrossTargeting GetTargetFrameworks target should only be used in cross targeting (outer) build" />
+    
     <ItemGroup>
       <_ThisProjectBuildMetadata Include="$(MSBuildProjectFullPath)">
-        <TargetFrameworks Condition="'$(TargetFrameworks)' != ''">$(TargetFrameworks)</TargetFrameworks>
-        <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">$(TargetFramework)</TargetFrameworks>
-        <HasSingleTargetFramework>true</HasSingleTargetFramework>
-        <HasSingleTargetFramework Condition="'$(IsCrossTargetingBuild)' == 'true'">false</HasSingleTargetFramework>
+        <TargetFrameworks>@(_TargetFrameworkInfo)</TargetFrameworks>
+        <TargetFrameworkMonikers>@(_TargetFrameworkInfo->'%(TargetFrameworkMonikers)')</TargetFrameworkMonikers>
+        <TargetPlatformMonikers>@(_TargetFrameworkInfo->'%(TargetPlatformMonikers)')</TargetPlatformMonikers>
+
+        <HasSingleTargetFramework>false</HasSingleTargetFramework>
+
         <!-- indicate to caller that project is RID agnostic so that a global property RuntimeIdentifier value can be removed -->
         <IsRidAgnostic>false</IsRidAgnostic>
         <IsRidAgnostic Condition=" '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == '' ">true</IsRidAgnostic>
@@ -44,6 +51,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <AdditionalProperties>TargetFramework=%(_TargetFrameworkNormalized.Identity)</AdditionalProperties>
       </_InnerBuildProjects>
     </ItemGroup>
+  </Target>
+
+  <Target Name="GetTargetFrameworksWithPlatformFromInnerBuilds"
+          DependsOnTargets="_ComputeTargetFrameworkItems">
+
+    <MSBuild Projects="@(_InnerBuildProjects)"
+             Condition="'@(_InnerBuildProjects)' != '' "
+             Targets="GetTargetFrameworksWithPlatformForSingleTargetFramework"
+             BuildInParallel="$(BuildInParallel)">
+      <Output ItemName="_TargetFrameworkInfo" TaskParameter="TargetOutputs" />
+    </MSBuild>
+
   </Target>
 
   <!--
--- a/src/Tasks/Microsoft.Common.CurrentVersion.targets
+++ b/src/Tasks/Microsoft.Common.CurrentVersion.targets
@@ -1697,18 +1697,39 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </Target>
 
   <Target Name="GetTargetFrameworks"
+          DependsOnTargets="GetTargetFrameworksWithPlatformForSingleTargetFramework"
           Returns="@(_ThisProjectBuildMetadata)">
+
+    <Error Condition="'$(IsCrossTargetingBuild)' == 'true'"
+           Text="Internal MSBuild error: Non-CrossTargeting GetTargetFrameworks target should not be used in cross targeting (outer) build" />
+
     <ItemGroup>
       <_ThisProjectBuildMetadata Include="$(MSBuildProjectFullPath)">
-        <TargetFrameworks Condition="'$(TargetFrameworks)' != ''">$(TargetFrameworks)</TargetFrameworks>
-        <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">$(TargetFramework)</TargetFrameworks>
+        <TargetFrameworks>@(_TargetFrameworkInfo)</TargetFrameworks>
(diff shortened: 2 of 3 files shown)


src/Tasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -21,13 +21,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Import Project="$(CustomBeforeMicrosoftCommonCrossTargetingTargets)" Condition="'$(CustomBeforeMicrosoftCommonCrossTargetingTargets)' != '' and Exists('$(CustomBeforeMicrosoftCommonCrossTargetingTargets)')"/>
 
   <Target Name="GetTargetFrameworks"
+          DependsOnTargets="GetTargetFrameworksWithPlatformFromInnerBuilds"
           Returns="@(_ThisProjectBuildMetadata)">
+
+    <Error Condition="'$(IsCrossTargetingBuild)' != 'true'"
```
cdmihai, line 27: I like the assertion, we should have them everywhere, but I am morbidly curios how many people it will break :D

src/Tasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -46,6 +53,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
   </Target>
 
+  <Target Name="GetTargetFrameworksWithPlatformFromInnerBuilds"
+          DependsOnTargets="_ComputeTargetFrameworkItems">
+
+    <MSBuild Projects="@(_InnerBuildProjects)"
+             Condition="'@(_InnerBuildProjects)' != '' "
```
cdmihai, line 60: nit: `'@(_InnerBuildProjects->Count())' == '0'` could be faster and avoid having spurious whitespace accidentally trigger the condition.

src/Tasks/Microsoft.Common.CrossTargeting.targets
```diff
@@ -46,6 +53,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
   </Target>
 
+  <Target Name="GetTargetFrameworksWithPlatformFromInnerBuilds"
+          DependsOnTargets="_ComputeTargetFrameworkItems">
+
+    <MSBuild Projects="@(_InnerBuildProjects)"
```
cdmihai, line 59: This will cause extra msbuild calls (but not extra evaluations as far as I can tell) even for non dotnet5 multi targeting projects. Is it worth doing this extra computation only when it's needed? (dotnet5 and above, or when the TF name requires expanding the platform?)

dsplaisted: We want to (eventually) support generic TargetFramework aliasing, so we don't want to restrict this to net5.0 and above.  Also, we don't even know if a TargetFramework is net5.0 or above without the parsing that happens in the inner build (which we could probably work around if we didn't want to eventually support aliasing).

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1697,20 +1712,41 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
       </_ThisProjectBuildMetadata>
     </ItemGroup>
   </Target>
 
+  <Target Name="GetTargetFrameworksWithPlatformForSingleTargetFramework"
+        Returns="@(_TargetFrameworkInfo)">
+
+    <ItemGroup>
+      <_TargetFrameworkInfo Include="$(TargetFramework)">
+        <TargetFrameworks>$(TargetFramework)</TargetFrameworks>
```
rainersigwald, line 1741: Is it intentional that the metadata name is plural but the value is singular?

rainersigwald: Ah: yes, because this is the single-targeting case.

src/Tasks/Microsoft.Common.CrossTargeting.targets
rainersigwald: It's extra evaluations in the not-all-TFs-are-used case, right?

This seems mostly fine. I wonder if it'll cause scheduling/node-locking problems by shifting around the timing for the evaluations/executions/requests. Hopefully not!

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1653,11 +1653,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
 
     <!-- For each reference, get closest match -->
+
+    <!-- Pass the CurrentProjectTargetPlatform parameter to the task only if GetReferenceNearestTargetFrameworkTaskSupportsTargetPlatformParameter is true.  This means
+         that we are using a version of NuGet which supports that parameter on this task. -->
+    
     <GetReferenceNearestTargetFrameworkTask AnnotatedProjectReferences="@(_ProjectReferenceTargetFrameworkPossibilities)"
```
rainersigwald, line 1660: If we call this task the new way with an item that has only the old information, does it still work? I'd like it to, so that anyone who's cobbled together a custom system that currently works continues working.

src/Tasks/Microsoft.Common.CrossTargeting.targets
cdmihai: Evaluation wise it should not be worse than the previous multitargeting scenario, where it also evaluates all innerbuilds but uses one.

cdmihai:
Also add `GetTargetFrameworksWithPlatformForSingleTargetFramework` to the target calling protocol here: https://github.com/dotnet/msbuild/blob/df645299a158c787f9884fca2bcd8c295318193f/src/Tasks/Microsoft.Managed.After.targets#L45-L49

We could also leave it stale to make sure that it breaks the tests that should break in the sdk if the protocol is not declared: https://github.com/dotnet/sdk/blob/4112eb34e3dd71c0888503af53e7e9d7560ad544/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAnAppWithoutTransitiveProjectRefs.cs#L34

src/Tasks/Microsoft.Common.CrossTargeting.targets
Forgind: If people want to call GetTargetFrameworks, I think they can do it from Microsoft.Common.CurrentVersion.targets. That sounds like a more legitimate place to call it if IsCrossTargetingBuild is false.

dsplaisted: I added these checks because previously the body of GetTargetFrameworks was the same in the CrossTargeting and CurrentVersion targets.  I assume that the CrossTargeting targets are only imported for the outer build, and the CurrentVersion targets only in an inner or  non-multi-targeted build.  The checks are basically assertions to verify that is correct.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1653,11 +1653,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
 
     <!-- For each reference, get closest match -->
+
+    <!-- Pass the CurrentProjectTargetPlatform parameter to the task only if GetReferenceNearestTargetFrameworkTaskSupportsTargetPlatformParameter is true.  This means
+         that we are using a version of NuGet which supports that parameter on this task. -->
```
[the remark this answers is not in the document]
dsplaisted: This is just to ease the transition.  Ideally we delete the duplication soon.

src/Tasks/Microsoft.Common.CurrentVersion.targets
dsplaisted: I think you're asking if the new version of the `GetReferenceNearestTargetFrameworkTask` task will continue to work in the old way (ie by parsing the TargetFrameworks) if it gets passed project items without `TargetFrameworkMonikers` and `TargetPlatformMonikers` 

@nkolev92 Is this the case?  Looking at [the NuGet PR](https://github.com/NuGet/NuGet.Client/pull/3578), it looks like probably it is.

dsplaisted:
@cdmihai Does d15de13 address the static graph protocol?  Do you think anything else will need to change?

@rainersigwald Does this look good to merge?

cdmihai:
> @cdmihai Does d15de13 address the static graph protocol? Do you think anything else will need to change?

Looks good for Build. If it ends up being called for Clean as well, you have to update that one as well.

src/Tasks/Microsoft.Common.CurrentVersion.targets
nkolev92: Yep, it will work in both.

src/Tasks/Microsoft.Common.CurrentVersion.targets
```diff
@@ -1697,20 +1712,41 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   </Target>
 
   <Target Name="GetTargetFrameworks"
+          DependsOnTargets="GetTargetFrameworksWithPlatformForSingleTargetFramework"
           Returns="@(_ThisProjectBuildMetadata)">
+
+    <Error Condition="'$(IsCrossTargetingBuild)' == 'true'"
+           Text="Internal MSBuild error: Non-CrossTargeting GetTargetFrameworks target should not be used in cross targeting (outer) build" />
```
Forgind, line 1719: Looking at https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1257284, does that mean this error check shouldn't be here because you can have an inner build that invokes an outer build in a way that it still needs to get the inner build's target frameworks and match them with the new outer build's target frameworks?

Issue #5019, opened 2020-01-01
Long if-else statements


Forgind:
https://github.com/microsoft/msbuild/blob/master/src/Build/Logging/BinaryLogger/BuildEventArgsWriter.cs#L35-L86
is not pretty and can probably be made a lot cleaner using a visitor. Alternatively, if the relevant information to be written to the log can be passed in a form other than the type—a string for example—that might be a cleaner and more efficient solution.

There are probably other similar examples.

szaliszali:
also related to the `BuildEventArgsWriter` mentioned in the issue by @Forgind :

`Microsoft.Build.Logging.EventArgsDispatcher.Dispatch` at https://github.com/dotnet/msbuild/blob/e70a3159d64f9ed6ec3b60253ef863fa883a99b1/src/Build/Logging/BinaryLogger/BuildEventArgsDispatcher.cs#L86-L142

additional very deeply nested if-else's:

`Microsoft.Build.Shared.FileUtilities.ItemSpecModifiers.GetItemSpecModifier` at https://github.com/dotnet/msbuild/blob/bf8629b4cd92c1bd05de8b1e124b4700eb0b7834/src/Shared/Modifiers.cs#L390-L616

`Microsoft.Build.Evaluation.Expander.Function.TryExecuteWellKnownFunction` at https://github.com/dotnet/msbuild/blob/df645299a158c787f9884fca2bcd8c295318193f/src/Build/Evaluation/Expander.cs#L3531-L4148

also, there are several cases where high complexity was introduced for performance optimization - one example: `Microsoft.Build.Shared.FileUtilities.ItemSpecModifiers.IsItemSpecModifier` at  https://github.com/dotnet/msbuild/blob/bf8629b4cd92c1bd05de8b1e124b4700eb0b7834/src/Shared/Modifiers.cs#L86-L253
this structure seems to be a frequently recurring pattern to reduce number of string comparisons, not sure if these are still worth with modern .Net versions of not

_[list above is not exhaustive, don't have good tools to discover all occurrences efficiently]_

PR #5659, opened 2020-08-16
convert ifs to visitor in BuildEventArgsWriter, BuildEventArgsDispatcher

szaliszali:
had a shot at converting the ifs mentioned in issue #5019 to visitor pattern

- does not close the issue because issue aims for a more general scope ("There are probably other similar examples.")
- did not use dynamic dispatch, so it's less pretty than it would be possible with dynamic (but no runtime cost)
- made it better or worse? not sure if really worth the change

### DIFF
--- a/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
+++ b/ref/Microsoft.Build.Framework/net/Microsoft.Build.Framework.cs
@@ -26,6 +26,8 @@ namespace Microsoft.Build.Framework
         public int LineNumber { get { throw null; } }
         public string ProjectFile { get { throw null; } set { } }
         public string Subcategory { get { throw null; } }
+        public override void Visit(Microsoft.Build.Framework.IBuildEventArgsDispatchVisitor buildEventArgsDispatcher) { }
+        public override void Visit(Microsoft.Build.Framework.IBuildEventArgsWriteVisitor buildEventArgsWriter) { }
     }
     public delegate void BuildErrorEventHandler(object sender, Microsoft.Build.Framework.BuildErrorEventArgs e);
     public abstract partial class BuildEventArgs : System.EventArgs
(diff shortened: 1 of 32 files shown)


KirillOsenkov:
Hi, first off thanks for the contribution!

Honestly I have to confess that I'm not sure this is a big improvement worth introducing. It does make some things cleaner but it also adds complexity. The current approach might seem uglier and more primitive, but a lot of ifs in a single method is a kind of complexity that's easy to contain and it doesn't add coupling or new API. The downside of the current approach is that you need to remember to update the method when new event args types are added. But you still have the same problem with visitors.

So one of my concerns is adding unjustified complexity into the core of MSBuild. Currently the binlog-related stuff is "on the side", so the core doesn't know anything about it. Adding two new interfaces there honestly doesn't feel justified. I notice that you could probably collapse them into a single interface and make it a bit more generic, like IBuildEventArgsVisitor, so you would only need to add a single override on each class, and not two. This would also enable adding more visitors in the future in case we need them. But even then honestly I feel like it wouldn't be worth the effort.

Another important concern is performance. We can't merge this change without thorough performance regression testing. This area is very performance sensitive and would be bad to inadvertently make things slower. It may make things faster though, but we won't know until we measure.

I'll let other team mates chime in and see what they think, but my recommendation would be to not merge this PR, even though I feel terrible about declining your work (sorry about that!). Besides just very questionable benefits which are hard to quantify, there's also the "bigger fish to fry" aspect - we have so many other issues to worry about that spending a lot of time on this would mean that real improvements that make life dramatically better for everyone wouldn't get attention.

Thanks again!

Forgind:
As a relevant data point, this seems to have essentially no effect on performance. My test took 49.4 seconds before the changes and 49.5 seconds after it.

rainersigwald:
Closing based on the discussion above.

